PR #1662, opened 2020-07-19
add org.eyecreate.vehiclevoyage

eyecreate:


LeandroStanger:
bot, build org.eyecreate.vehiclevoyage

eyecreate:
bot, build org.eyecreate.vehiclevoyage

eyecreate:
I tested the last build and things look good.

org.eyecreate.vehiclevoyage.json
```diff
@@ -0,0 +1,38 @@
...
+  "command": "vehiclevoyage",
+  "runtime-version": "5.14",
+  "sdk": "org.kde.Sdk",
+  "finish-args": [
+    "--share=ipc",
+    "--share=network",
+    "--socket=x11",
+    "--socket=wayland",
+    "--device=dri",
+    "--filesystem=home",
```
nedrichards, line 13: is there a reason you need access to the whole home directory rather than using the portal or scoping to a specific path or directory?

eyecreate: I don't need it at the moment(I only write/read files to Qt AppDataLocation), but thought I might need to have user saved files later. A quick look at portals shows I'm unfamiliar with everything flatpak offers and shouldn't need whole home permission. I'll update the manifest.

eyecreate:
bot, build org.eyecreate.vehiclevoyage

org.eyecreate.vehiclevoyage.json
nedrichards: If you need a permission later you can add it to the manifest then, but certainly always try and ask for the minimum!

PR #1663, opened 2020-07-22
add co.headsetapp.headset

fcastilloec:


fcastilloec:
bot, build co.headsetapp.headset

fcastilloec:
bot, build

fcastilloec:
bot, build please

fcastilloec:
bot, build

fcastilloec:
bot, build co.headsetapp.headset

fcastilloec:
bot, build co.headsetapp.headset

fcastilloec:
bot, build co.headsetapp.headset

fcastilloec:
bot, build co.headsetapp.headset

fcastilloec:
bot, build co.headsetapp.headset

fcastilloec:
bot, build co.headsetapp.headset

fcastilloec:
bot, build co.headsetapp.headset

fcastilloec:
bot, build co.headsetapp.headset

co.headsetapp.headset.desktop
```diff
@@ -0,0 +1,11 @@
...
+Name=Headset
+Comment=Discover and collect music on YouTube
+GenericName=Music Player
+Exec=headset %U
+Icon=co.headsetapp.headset
+Type=Application
+StartupNotify=true
+Categories=Audio;AudioVideo;Player;Music;
+
+Name[en_US]=co.headsetapp.headset.desktop
```
bilelmoussaoui, line 11:
```suggestion
```

co.headsetapp.headset.metainfo.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
```
bilelmoussaoui, line 2:
```suggestion
<component type="desktop-application">
```

co.headsetapp.headset.metainfo.xml
```diff
@@ -0,0 +1,40 @@
...
+  <id>co.headsetapp.headset</id>
+  <metadata_license>MIT</metadata_license>
+  <name>Headset</name>
```
bilelmoussaoui, line 5: The metainfo/desktop file & the icons should be part of upstream. They are not Flathub specific.

co.headsetapp.headset.yml
```diff
@@ -0,0 +1,45 @@
+id: co.headsetapp.headset
+default-branch: stable
```
bilelmoussaoui, line 2:
```suggestion
```
it's the default

co.headsetapp.headset.metainfo.xml
fcastilloec: I'm trying to create the flatpak from a `.deb` repository that we already have (it's proving quite difficult so far). I can use the `.dekstop` file included in the `.deb` instead of providing one directly in this repo, if that's what you mean. The file is not going to change, and I thought it'll be just better to provide it here, rather than edit it as part of the build process.
For the `metainfo` file, I thought this file was Flathub specific, I've never seen it used anywhere else, at least our project doesn't use it anywhere else.

fcastilloec:
@bilelmoussaoui Thanks for your comments. I have some questions about this process. I've been trying to find some documentation on how to build from a `.deb` file/repository and found multiple approaches. One sets the `.deb` as a source of type `file` and unpacks it as part of the build process, the other uses a source of type `extra-data` together with a `script`. Are there any advantages of one over the other? Which one is recommended?

Headset is an electron app and the official Flathub documentation deals with unpackaged electron apps. There's no mention of zypak anywhere. I've tried to find some help with this, but I've unsuccessful. The majority of PR I've seen, they simply just start the app with `zypak-wrapper` and don't include zypak as a source. When building locally and trying to run using `flatpak-builder` I get the message that `zypak` can't be found. Should I add zypak as a source? How are all these app just using `zypak` without adding it as a source? [Slack](https://github.com/flathub/com.slack.Slack/blob/master/com.slack.Slack.json) is a good example.

co.headsetapp.headset.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
```
TingPing, line 1: These go upstream.

co.headsetapp.headset.metainfo.xml
TingPing: Appstream is a generic standard, it is used by all distributions and modern "app stores" on linux.

co.headsetapp.headset.yml
```diff
@@ -0,0 +1,44 @@
+id: co.headsetapp.headset
+base: org.electronjs.Electron2.BaseApp
+base-version: stable
```
TingPing, line 3:
```suggestion
base-version: '19.08'
```

icons/headset-symbolic.svg
```diff
@@ -0,0 +1,72 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
TingPing, line 1: Should just be upstream also.

TingPing: Resolving as these are upstream in a future release.

barthalion:
bot, build co.headsetapp.headset

fcastilloec:
bot, build co.headsetapp.headset

co.headsetapp.headset.metainfo.xml
fcastilloec: I've switched to using the desktop file from the `.deb`, for the metainfo. I'm going to leave it here for the moment. Flatpak is the only one using it so far.
If I publish the app on other app stores, I'll move the manifest upstream and just reference it from there.

@bilelmoussaoui Regarding the icons, as mentioned somewhere else. The last version of our `.deb` is not using scalable icons, so this is just temporary. Once we release a new version, the icons will also come from the `.deb`. If you prefer, I can use URLs for the icons.

TingPing: Flatpak is not the only thing using. Literally all mainstream distributions use this file.

bilelmoussaoui:
> @bilelmoussaoui Thanks for your comments. I have some questions about this process. I've been trying to find some documentation on how to build from a `.deb` file/repository and found multiple approaches. One sets the `.deb` as a source of type `file` and unpacks it as part of the build process, the other uses a source of type `extra-data` together with a `script`. Are there any advantages of one over the other? Which one is recommended?
> 
> Headset is an electron app and the official Flathub documentation deals with unpackaged electron apps. There's no mention of zypak anywhere. I've tried to find some help with this, but I've unsuccessful. The majority of PR I've seen, they simply just start the app with `zypak-wrapper` and don't include zypak as a source. When building locally and trying to run using `flatpak-builder` I get the message that `zypak` can't be found. Should I add zypak as a source? How are all these app just using `zypak` without adding it as a source? [Slack](https://github.com/flathub/com.slack.Slack/blob/master/com.slack.Slack.json) is a good example.

The extra-data type is used for files that Flathub can not re-distribute for license reasons. The extra-data is downloaded on the user machine and it requires a apply-extra script that uncompress the archive and install it on the proper place. As the download happens on the user's machine during the install, we require a data-check to ensure the URL wasn't updated to something new due to an update which might not work for the users trying to install it right now.

Regarding zypak, it's already available in the electron baseapp

fcastilloec:
bot, build co.headsetapp.headset

co.headsetapp.headset.yml
```diff
@@ -0,0 +1,45 @@
...
+command: headset
+finish-args:
+  - --socket=x11
+  - --share=ipc
+  - --device=dri
+  - --socket=pulseaudio
+  - --share=network
+  - --env=TMPDIR=/var/tmp
+  - --talk-name=org.freedesktop.Notifications
+  - --own-name=org.mpris.MediaPlayer2.headset
```
bilelmoussaoui, line 16: After trying the app, it doesn't seems to use mpris at all or this is supposed to work?

bilelmoussaoui: It seems to work fine with`org.mpris.MediaPlayer2.chromium.instance3` you might need to change that somewhere at the source code of your app. By default your app owns the name of `org.mpris.MediaPlayer2.$app-id`

fcastilloec: I've tried installing the latest test build on Ubuntu 18.04 and 20.04 and everything works without any problem there, including mpris.
When using D-Feet, I can only see `org.mprsi.MediaPlayer2.headset`. The app has explicit code to expose the name I've set in the manifest, and also explicitly blocks any attempt by electron to create additional session busses.
A reliable tool to control mpris interfaces (besides using dbus command line) is [playerctl](https://github.com/altdesktop/playerctl). I've found that some OSs intercept keyboard shortcuts, and mess up mpris. Also using [Mpris Indicator Button](https://extensions.gnome.org/extension/1379/mpris-indicator-button/) is another great way to control mpris players. In fact, that gnome extension has specific code with exceptions to make shuffle and repeat work out of the box.
What distro are you using?

fcastilloec: @bilelmoussaoui actually, now that you mention it, there was a problem that has already been fixed upstream with mpris. As I mentioned, we now have code to block electron from automatically creating its own mpris session, which the current version in flatpak doesn't have. It's the same with the icons, upstream uses scalable icons, but the version in flatpak doesn't, hence why I added them explicitly.

bilelmoussaoui: Sure, it would be nice to use the current(the one set by electron) mpris name as well and remove it once you have published a new release later, lgtm otherwise

fcastilloec: I can't add the electron mpris name to the manifest because electron uses a random number `org.mpris.MediaPlayer2.chromium.instance${random number}` and even if I add it with an asterisk (which probably is not supported) might cause trouble with other electron apps that haven't deal carefully with mpris since I'm basically calling any mpris electron mpris name mine.

bilelmoussaoui: Makes sense. Thanks for the details

PR #1673, opened 2020-07-27
Add org.sugarlabs.MusicBlocks

Clue88:
Flatpak distribution of Music Blocks

@tchx84

### DIFF
--- a/org.sugarlabs.MusicBlocks.yaml
+++ b/org.sugarlabs.MusicBlocks.yaml
@@ -1,0 +1,74 @@
+app-id: org.sugarlabs.MusicBlocks
+branch: stable
+runtime: org.freedesktop.Platform
+runtime-version: '19.08'
+sdk: org.freedesktop.Sdk
+base: "org.electronjs.Electron2.BaseApp"
+base-version: '19.08'
+separate-locales: false
+command: musicblocks-run
+finish-args:
+  - '--share=ipc'
+  - '--socket=x11'
+  - '--socket=pulseaudio'
+  - '--share=network'
+  - '--device=dri'
+  - '--filesystem=home'
+  - '--env=MUSICBLOCKS_PATH=/app/share/musicblocks'
+modules:
+  - name: unappimage
+    buildsystem: simple
+    build-commands:
+      - make -C squashfs-tools -j ${FLATPAK_BUILDER_N_JOBS} install INSTALL_DIR=${FLATPAK_DEST}/bin
+    sources:
+      - type: git
+        url: https://github.com/refi64/unappimage
+        commit: d7f86f2a0d7ec3a69211125207d5f127386b849a
+  
+  - name: mb-data
+    buildsystem: simple
+    build-commands:
+      - python3 setup.py install --prefix=${FLATPAK_DEST}
+
+      - install -d /app/share/appdata
+      - mv /app/share/metadata/org.sugarlabs.MusicBlocks.appdata.xml /app/share/appdata
+      
+      - unappimage MusicBlocks.AppImage
+      - rm MusicBlocks.AppImage
+      - mv squashfs-root /app/bin/musicblocks
+    sources:
+      - type: git
+        url: 'https://github.com/sugarlabs/musicblocks-launcher'
+        tag: 'v1.1.2'
+        commit: 'd96812365a0e91e0ea08ea0d3e72c60acba6f650'
+      - type: file
+        dest-filename: MusicBlocks.AppImage
+        url: https://github.com/sugarlabs/musicblocks-launcher/releases/download/v1.1.2/Music.Blocks-3.2.0-aarch64.AppImage
+        sha256: 79529206250952752dd0e2711ce26349e951478996878ce4b379b3c0299166a9
+        size: 117327933
+        only-arches:
+          - aarch64
+      - type: file
+        dest-filename: MusicBlocks.AppImage
+        url: https://github.com/sugarlabs/musicblocks-launcher/releases/download/v1.1.2/Music.Blocks-3.2.0-arm.AppImage
+        sha256: 8cb33655c758954b2a9f1babff93ad16cc2e65ca2a1f93c232a5c130db133de6
+        size: 131462400
+        only-arches:
+          - arm
+      - type: file
+        dest-filename: MusicBlocks.AppImage
+        url: https://github.com/sugarlabs/musicblocks-launcher/releases/download/v1.1.2/Music.Blocks-3.2.0-x86_64.AppImage
+        sha256: a0956a53c8981fb5696829b81ef9a7059fa80bd86092fcbe419454c2ee351dfc
+        size: 127771791
+        only-arches:
+          - x86_64
+
+  - name: musicblocks
+    buildsystem: simple
+    build-commands:
+      - install -Dm 755 musicblocks-run /app/bin/musicblocks-run
+    sources:
+      - type: script
+        dest-filename: musicblocks-run
+        commands:
+          - exec /app/bin/musicblocks/musicblocks "$@" --no-sandbox


tchx84:
@Clue88 great work!

A few comments:
* add `--device=dri` permission to the manifest, so it can take advantage of GPU acceleration.
* change the developer name to `Sugar Labs Community` to keep it consistent with the other apps.
* fix the "save project" option, it doesn't seem to be using the portal (see image below)

image
.

tchx84:
Another detail I noticed, is that the bundle is way bigger that it should be.

Can you double check which of those directories are not strictly necessary? And remove them? You can do that from the manifest,before or after you install the directory.

```
4.0K	android_chrome_manifest.json
16K	bower_components
4.0K	cache.appcache
1.2M	charts
4.0K	config.xml
36K	COPYING
4.0K	CREDITS.md
52K	css
8.0K	Debugging.md
56K	dist
2.8M	documentation
1.8M	documentation-es
1.7M	documentation-ja
1.7M	documentation-kana
1.9M	documentation-pt
4.3M	examples
12K	FAQ
68K	favicon.ico
828K	fonts
5.6M	guide
2.4M	guide-es
5.6M	guide-ja
3.6M	guide-pt
3.0M	guide-zhCN
4.0K	gulpfile.js
760K	header-icons
1.6M	images
32K	index.html
3.7M	js
7.0M	lib
11M	lilypond
3.5M	loading-animation.gif
1.2M	loading-animation-ja.svg
476K	localization.ini
4.0K	manifest.json
4.0K	manifest.webapp
460K	Minsky.jpg
136K	mouse-art
176K	msie_flashFallback
4.0K	NEWS
4.0K	org.sugarlabs.MusicBlocks.ui
4.0K	package.json
572K	planet
8.0K	pluginify.py
432K	plugins
19M	po
8.0K	README-ja.md
8.0K	README.md
416K	screenshots
4.0K	script.js
4.0K	server.md
4.0K	setup.py
23M	sounds
4.0K	sw.js
97M	usermanual
```

Clue88:
I'll make the changes. Walter said that he would investigate which directories were unnecessary – `po` and `usermanual` stand out to me as large directories that aren't used, so I'll remove them in the cleanup during the build.

org.sugarlabs.MusicBlocks.json
```diff
@@ -10,6 +10,7 @@
         "--socket=fallback-x11",
         "--socket=pulseaudio",
         "--device=dri",
+        "--filesystem=xdg-download",
```
tchx84, line 13: What I think we need is display a proper file chooser, instead of that very unintuitive dialog.

org.sugarlabs.MusicBlocks.json
```diff
@@ -43,6 +43,15 @@
                     "commit": "5a0e5fc244acd8fa806f78a937165cc2a528e68e",
                     "dest": "musicblocks"
                 }
+            ],
+            "cleanup": [
+                "musicblocks/po",
```
tchx84, line 48: This won't work, you need to remove (`rm -rf`) these directories in `build-commands` before you copy the files. Also, take a look a `install` command, instead of `cp`.

nedrichards:
bot, build org.sugarlabs.MusicBlocks

nedrichards:
> builddir/files/share/appdata/org.sugarlabs.MusicBlocks.appdata.xml: FAILED:
• attribute-invalid     : component has invalid type attribute

bilelmoussaoui:
https://github.com/sugarlabs/musicblocks-launcher/blob/master/data/org.sugarlabs.MusicBlocks.appdata.xml#L2
This needs a `type="desktop-application"` along with missing OARS tag

org.sugarlabs.MusicBlocks.json
Clue88: Thank you! I don't think that install will work in this case – unless there's some way to recursively copy the folders? `install -D musicblocks/* app/share/musicblocks` only copies the files outside of directories.

tchx84:
bot, build org.sugarlabs.MusicBlocks

tchx84:
@Clue88 MusicBlocks icon is SVG, therefore you should install it in `share/icons/hicolor/scalable/apps/`, e.g. `/app/share/icons/hicolor/scalable/apps/org.sugarlabs.MusicBlocks.svg.

See https://github.com/sugarlabs/musicblocks-launcher/blob/master/setup.py#L10

org.sugarlabs.MusicBlocks.json
```diff
@@ -24,7 +24,7 @@
                     "type": "git",
                     "url": "https://github.com/sugarlabs/musicblocks-launcher",
                     "tag": "v1.0.4",
```
tchx84, line 26: Read my comment above, and don't forget to bump your tag.

Clue88: Just changed it!

tchx84:
bot, build org.sugarlabs.MusicBlocks

tchx84:
@Clue88 I tested the app, and I am noticing something odd... when I play the "demo/default" song, half of the times the sound is a bit choppy. e.g. some notes do not sound, or they overlap. Have you noticed this?

Clue88:
That’s strange... I haven’t seen it on my end. Maybe some issue with PulseAudio?

tchx84:
> That’s strange... I haven’t seen it on my end. Maybe some issue with PulseAudio?

I doubt it, as I don't this happening in other music apps. Most likely something related to the launcher.

electron-app.patch
```diff
@@ -0,0 +1,37 @@
+diff --git a/main/package.json b/main/package.json
+index c4fd488e30644e247e86f14a0ec12e82d3cf27c1..92d238c4b9b4d639b446c0e7936aa2e2ae456944 100644
+--- a/main/package.json
++++ b/main/package.json
+@@ -4,7 +4,7 @@
+   "description": "Visual Programming Language for Exploring Math, Music, and Programming",
+   "main": "main.js",
+   "scripts": {
+-    "start": "electron .",
++    "start": "electron . --no-sandbox",
```
bilelmoussaoui, line 10: You should use zypak instead of disabling the sandbox, see other electron apps on Flathub for an example.

org.sugarlabs.MusicBlocks.json
```diff
@@ -1,20 +1,50 @@
...
-        "--socket=fallback-x11",
+        "--socket=x11",
         "--socket=pulseaudio",
         "--device=dri",
+        "--filesystem=home",
         "--env=MUSICBLOCKS_PATH=/app/share/musicblocks"
     ],
+    "build-options" : {
+        "cflags": "-O2 -g",
+        "cxxflags": "-O2 -g",
```
bilelmoussaoui, line 21:
```suggestion
```

electron-app.patch
Clue88: Hi, Thank you so much for your help! I was originally having problems running electron because of an issue with the configuration of the SUID sandbox helper binary and was told to just not use the sandbox (because flatpak already has one) as a fix. Will zypak fix the original configuration problem?

Clue88: I'm also not entirely sure how to implement zypak... I'm a little lost looking through the examples that use it. The `run.sh` files that they create have a bunch more commands that I don't understand and it looks pretty different from the `npm start` used to run the app in the Flatpak [electron-sample-app](https://github.com/flathub/electron-sample-app) example I based the app on.

I've also pasted below the console output of what I got when I tried to run the app in the first place without `--no-sandbox`. I did check the file that it referenced and the permissions seem to be correct so perhaps it's a symptom of a different issue (which was fixed when I added `--no-sandbox`. According to some basic Googling it seems like this is a relatively common electron issue with AppImage and Flatpak, but if you know any more about it, any help would be greatly appreciated!

```
The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without
sandboxing I'm aborting now. You need to make sure that
/app/share/musicblocks/node_modules/electron/dist/chrome-sandbox is owned by root and has mode 4755.
```

Thanks so much for your help.
@bilelmoussaoui

nedrichards:
bot, build org.sugarlabs.MusicBlocks

tchx84:
> Build [24889 failed](https://flathub.org/builds/#/builders/32/builds/24889)

@Clue88  I know little-to-nothing about electron, but you should be seeing this same error when building locally:

```
flatpak-builder: Cleanup electron-appError: module electron-app: Can't specify build-args in sandboxed build
```

Clue88:
@tchx84 There’s some process that’s happening in the postinstall for electron that requires network access but that’s disabled by default during the actual build process. So, I enabled it as a build arg for that module, which works but not when you build it in sandboxed mode (where all of the build args are disabled). Though, seeing as all of the other electron apps on Flathub seem to be building fine, there must be some workaround.

org.sugarlabs.MusicBlocks.json
```diff
@@ -0,0 +1,125 @@
...
+                }
+            ]
+        },
+        {
+            "name": "electron-app",
+            "build-options": {
+                "env": {
+                    "electron_config_cache": "/run/build/electron-app/flatpak-node/npm-cache"
+                },
+                "build-args": [ "--share=network" ]
```
barthalion, line 94: That's not going to work, Flathub builds are offline. Look at https://github.com/flatpak/flatpak-builder-tools/blob/master/node/README.md to work around it.

If it's about just some component not being stored in generated-sources.json, @refi64 might know what's going on.

org.sugarlabs.MusicBlocks.json
```diff
@@ -0,0 +1,125 @@
...
+                "/share",
+                "/app/lib/node_modules/npm/changelogs",
+                "/app/lib/node_modules/npm/doc",
+                "/app/lib/node_modules/npm/html",
+                "/app/lib/node_modules/npm/man",
+                "/app/lib/node_modules/npm/scripts"
+            ],
+            "sources": [
+                {
+                    "type": "archive",
```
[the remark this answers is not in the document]
barthalion: (Also for how to use zypak.)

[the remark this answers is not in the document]
Clue88: Thanks! As a point of clarification, Music Blocks is really only a static HTML page rather than a full-fledged electron app – so it should hopefully look more like the simple example here: https://github.com/flatpak/flatpak-builder-tools/tree/master/node/vanilla-quick-start. I'm still having problems with something not being stored that is needed in electron postinstall as well as some SUID sandbox helper binary misconfiguration, but I'll try using node12 to see if that fixes anything.

Clue88:
bot, build org.sugarlabs.MusicBlocks

Clue88:
bot, build org.sugarlabs.MusicBlocks

electron-app.patch
```diff
@@ -0,0 +1,24 @@
...
+     var choice = require('electron').dialog.showMessageBoxSync(this,
+         {
+@@ -32,6 +34,7 @@ function createWindow () {
+          e.preventDefault();
+        }
+   });
++*/
+ 
+   // Open the DevTools.
+   // mainWindow.webContents.openDevTools()
```
tchx84, line 24: Do you still need this patch ?

org.sugarlabs.MusicBlocks.yaml
```diff
@@ -0,0 +1,115 @@
+app-id: org.sugarlabs.Musicblocks
```
tchx84, line 1: The error you see is because the `app-id` doesn't match. It should be `org.sugarlabs.MusicBlocks` capital B. This shouldn't have worked either locally.

electron-app.patch
Clue88: Yes – the `main.js` from `musicblocks-app` has the wrong filepath for `index.html` and there's also a bug with the prompt that's given when the user tries to exit the app which makes it impossible to quit the app without force-quitting.

org.sugarlabs.MusicBlocks.yaml
Clue88: I see, thanks! I'm not sure why it worked locally – must have been something left over from previous builds.

Clue88:
bot, build org.sugarlabs.MusicBlocks

electron-app.patch
refi64: If there's no `chrome-sandbox` binary in the same directory as your Electron app, do something like

```bash
echo 'true' > chrome-sandbox
chmod +x chrome-sandbox
```

which should make it work. This is a deficiency of sorts where Zypak expects the a chrome-sandbox to exist that can be overridden (which is something I plan on changing, not hard but haven't gotten the chance).

org.sugarlabs.MusicBlocks.json
Clue88: @barthalion My main issue is that it seems to be trying to redownload the electron binary during postinstall. The binary is specified in `generated-sources.json`, and the path is stored in `electron_config_cache`, but it still tries to download it again anyway (resulting in `ELIFECYCLE` and `EAI_AGAIN` errors if it's not connected to the network). Is there something I can do to force it to use the cached binary?

barthalion:
Looks like appdata file is installed in the wrong location. This makes it work:

```
diff --git a/org.sugarlabs.MusicBlocks.yaml b/org.sugarlabs.MusicBlocks.yaml
index 88c2135..022e372 100644
--- a/org.sugarlabs.MusicBlocks.yaml
+++ b/org.sugarlabs.MusicBlocks.yaml
@@ -43,11 +43,12 @@ modules:
     build-commands:
       # Install all of the app data in the appropriate locations.
       - 'python3 setup.py install --prefix=${FLATPAK_DEST}'
+      - install -d /app/share/appdata
+      - mv /app/share/metadata/org.sugarlabs.MusicBlocks.appdata.xml /app/share/appdata
     sources:
       - type: git
         url: 'https://github.com/sugarlabs/musicblocks-launcher'
-        tag: 'v1.1.1'
-        commit: '10661f7db183f520d9ce6bd3872d47f31c94e947'
+        commit: '88b1e1272b8b48ce832799bba1160ac1ea3fa268'
 
   # Next is the actual musicblocks code.
   - name: mb-app
```

That being said, releases in appdata need to be sorted by version:

```
org.sugarlabs.MusicBlocks:         AppData problem: tag-invalid : <release> versions are not in order [3.1 before 3.2]
```

Clue88:
bot, build org.sugarlabs.MusicBlocks

tchx84:
@Clue88 I see the following error when I run it:

```
$ flatpak run org.sugarlabs.MusicBlocks//stable

> Musicblocks@3.1.0 start /app/share/musicblocks
> electron .

/app/share/musicblocks/node_modules/electron/index.js:14
    throw new Error('Electron failed to install correctly, please delete node_modules/electron and try installing again')
    ^

Error: Electron failed to install correctly, please delete node_modules/electron and try installing again
    at getElectronPath (/app/share/musicblocks/node_modules/electron/index.js:14:11)
    at Object.<anonymous> (/app/share/musicblocks/node_modules/electron/index.js:18:18)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/app/share/musicblocks/node_modules/electron/cli.js:3:16)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! Musicblocks@3.1.0 start: `electron .`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the Musicblocks@3.1.0 start script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /var/home/tchx84/.npm/_logs/2020-08-12T19_15_47_824Z-debug.log

```

Clue88:
@tchx84 I got that as well. I think there are still issues with the electron binary. In theory, what's supposed to happen during electron's postinstall is that it checks to see if the binary is in cache, and downloads the correct binary from github if it is not. So, if it's in cache everything should be offline. I added a workaround just so it can build to make sure everything else was working, but this version doesn't actually have a working electron binary (so nothing works). For some reason, it's not picking up that there's a cached version of the binary already there during the build process (and I've checked to make sure it's actually there and has the right version). I used the same environment variable to specify the cache location (`electron_config_cache`) that other electron Flatpak apps have used, so I'm not sure why this is the case.

I'm looking into whether it's possible to manually install the cache after all of the other packages are installed. It's also possible this is an issue with Electron 9 (I haven't found any apps that are using electron versions past v8.2.5 so maybe those developers know something I don't know yet). That said, the [source generator script](https://github.com/flatpak/flatpak-builder-tools/blob/master/node/flatpak-node-generator.py) will only include v8.5.0, now that it is the latest version of electron 8.x (even if v8.2.5 is specified explicitly in `package-lock.json`), and I ran into the same binary problems in v8.5.0.

I also tried similar builds with yarn instead of npm, but I ran into the same problems. I'm not sure if v8.2.5 would work, but seeing as I can't get it to use v8.2.5 in the first place with the existing source generator script, it might be a mute point anyway.

Clue88:
I've also tried other tools like [`electron-installer-flatpak`](https://github.com/endlessm/electron-installer-flatpak), but that was also unsuccessful; it returns an error when building the flatpak that I can only assume is the same binary problem as before (though it doesn't log anything from the error, only the error code, which is generic, so I don't know for certain).

Clue88:
bot, build org.sugarlabs.MusicBlocks

Clue88:
bot, build org.sugarlabs.MusicBlocks

Clue88:
bot, build org.sugarlabs.MusicBlocks

Clue88:
bot, build org.sugarlabs.MusicBlocks

barthalion:
```
❯ flatpak run org.sugarlabs.MusicBlocks
/app/bin/MusicBlocks: error while loading shared libraries: libffmpeg.so: cannot open shared object file: No such file or director
```

That's definitely a progress! I don't see this library anywhere in the app though. Usually it's shipped with Electron binary/directory.

Clue88:
Current issues:
When I run the app, I get the following error: `[fake-sandbox: zypak-sandbox] No data could be read (host died?)`. The app executable itself runs fine, however, so it seems like it's definitely an issue with how I set up zypak. If worst comes to worst, I can use `--no-sandbox` instead of zypak which should work.

Clue88:
bot, build org.sugarlabs.MusicBlocks

PR #1674, opened 2020-07-28
Add org.freedesktop.Sdk.Extension.ziglang

protheory8:


org.freedesktop.Sdk.Extension.ziglang.json
```diff
@@ -0,0 +1,90 @@
...
+            "buildsystem": "simple",
+            "build-commands": [
+                "mkdir -p /usr/lib/sdk/ziglang",
+                "cp -r ./* /usr/lib/sdk/ziglang",
+                "chmod a+x /usr/lib/sdk/ziglang/zig"
+            ],
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://ziglang.org/download/0.6.0/zig-linux-x86_64-0.6.0.tar.xz",
```
bilelmoussaoui, line 26: It would be nice to use Flathub's data checker to auto update the SDK, see what we do for Rust stable.

org.freedesktop.Sdk.Extension.ziglang.json
```diff
@@ -0,0 +1,90 @@
...
+                    ]
+                },
+                {
+                    "type": "archive",
+                    "url": "https://ziglang.org/download/0.6.0/zig-linux-i386-0.6.0.tar.xz",
+                    "sha256": "a97a2f9ae21575743cdd763c1917d49400d83fc562ef64582b18bade43eb24ce",
+                    "only-arches": [
+                        "i386"
+                    ]
+                },
```
bilelmoussaoui, line 39: The recent runtimes don't support i386 anymore. So this can be removed
```suggestion
```

org.freedesktop.Sdk.Extension.ziglang.json
protheory8: It seems like `"type": "rotating-url"` and `"type": "html"` are fine, but I haven't seen any URLs/webpages that can be used with those.

org.freedesktop.Sdk.Extension.ziglang.json
protheory8: Sure, but this manifest is for Freedesktop SDK 19.08, so if 19.08 supports i386, I would like this package to support it as well.

protheory8: I'll apply your change later today I guess.

bilelmoussaoui: 19.08 doesn't already, hence my change suggestion

org.freedesktop.Sdk.Extension.ziglang.json
protheory8: Yeah, can't find them. Should I make a HTML page that just reports latest version of Zig and use it?

protheory8:
There are 2 files for `arm` arch, `zig-linux-armv7a-0.6.0.tar.xz` and `zig-linux-armv6kz-0.6.0.tar.xz`. I chose the former one. Did I make a right choice?

protheory8:
So what should I do?

nedrichards:
bot, build org.freedesktop.Sdk.Extension.ziglang

org.freedesktop.Sdk.Extension.ziglang.json
bilelmoussaoui: if there's no way to use the auto-updater for now, it's not a problem. Just make sure to keep it up to date.

PR #1675, opened 2020-07-28
Add org.ksnip.ksnip

LyzardKing:
This is a working declaration for the ksnip flatpak.
The remaining issues are:
- The file portal will not give the proper (executable) access to the selected files.
  Also if the file-chooser is set to select directories it will not show any files/folders in the users home
  A workaround using the `filesystem=host` permission solves this issue.
  `filesystem=home` could be used, simply removing the access to non /home/<user> directories (right?)
- On wayland it will not take the screenshot (or show the indicator icon)

The first issue has been previously reported at https://github.com/flatpak/xdg-desktop-portal/issues/517

### DIFF
--- a/org.ksnip.ksnip.yaml
+++ b/org.ksnip.ksnip.yaml
@@ -7,13 +7,11 @@ rename-icon: ksnip
 rename-appdata-file: ksnip.appdata.xml
 rename-desktop-file: ksnip.desktop
 finish-args:
-  - "--socket=wayland"
-  - "--socket=fallback-x11"
-  - "--share=ipc"
-  - "--device=dri"
-  - "--filesystem=host"
-  - "--share=network"
-  - "--own-name=org.ksnip"
+  - --socket=x11
+  - --share=ipc
+  - --device=dri
+  - --filesystem=host
+  - --share=network
 cleanup:
   - "*.a"
   - "/include"
--- a/org.ksnip.ksnip.yaml
+++ b/org.ksnip.ksnip.yaml
@@ -24,14 +24,18 @@ modules:
     sources:
       - type: git
         url: https://github.com/ksnip/kColorPicker.git
+        tag: v0.1.4
   - name: kImageAnnotator
     buildsystem: cmake-ninja
     sources:
       - type: git
         url: https://github.com/ksnip/kImageAnnotator.git
+        tag: v0.3.2
   - name: ksnip
     buildsystem: cmake-ninja
     sources:
       - type: git
         url: https://github.com/ksnip/ksnip.git
-        #tag: v1.7.2
+        commit: 813b1048b5bfa5fb09a5aa838912e4b26c8b0c9c
+        # Enable the tag when the changes are in a new tag
+        # tag: v1.7.2
--- a/org.ksnip.ksnip.yaml
+++ b/org.ksnip.ksnip.yaml
@@ -13,6 +13,7 @@ finish-args:
   - --filesystem=host
   - --share=network
   - --own-name=org.kde.*
+  - --talk-name=org.freedesktop.Notifications
 cleanup:
   - "*.a"
   - "/include"
--- a/org.ksnip.ksnip.yaml
+++ b/org.ksnip.ksnip.yaml
@@ -4,8 +4,6 @@ runtime-version: '5.14'
 sdk: org.kde.Sdk
 command: ksnip
 rename-icon: ksnip
-rename-appdata-file: ksnip.appdata.xml
-rename-desktop-file: ksnip.desktop
 finish-args:
   - --socket=x11
   - --share=ipc
@@ -33,9 +31,7 @@ modules:
         url: https://github.com/ksnip/kImageAnnotator.git
   - name: ksnip
     buildsystem: cmake-ninja
-    post-install:
-      - install -Dm644 /app/share/pixmaps/ksnip.svg /app/share/icons/hicolor/scalable/apps/ksnip.svg
     sources:
       - type: git
         url: https://github.com/ksnip/ksnip.git
-        tag: v1.7.1
+        #tag: v1.7.2
--- a/org.ksnip.ksnip.yaml
+++ b/org.ksnip.ksnip.yaml
@@ -12,6 +12,7 @@ finish-args:
   - --device=dri
   - --filesystem=host
   - --share=network
+  - --own-name=org.kde.*
 cleanup:
   - "*.a"
   - "/include"
--- a/org.ksnip.ksnip.yaml
+++ b/org.ksnip.ksnip.yaml
@@ -1,0 +1,41 @@
+app-id: org.ksnip.ksnip
+runtime: org.kde.Platform
+runtime-version: '5.14'
+sdk: org.kde.Sdk
+command: ksnip
+rename-icon: ksnip
+rename-appdata-file: ksnip.appdata.xml
+rename-desktop-file: ksnip.desktop
+finish-args:
+  - "--socket=wayland"
+  - "--socket=fallback-x11"
+  - "--share=ipc"
+  - "--device=dri"
+  - "--filesystem=host"
+  - "--share=network"
+  - "--own-name=org.ksnip"
+cleanup:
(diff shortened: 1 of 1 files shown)


nedrichards:
bot, build org.ksnip.ksnip

nedrichards:
To note for other admins who may not be subscribed to the issue: upstream at: https://github.com/ksnip/ksnip/issues/127

LyzardKing:
@nedrichards what is the issue (Check for right id in AppStream xml)?
Locally it builds fine

bilelmoussaoui:
The id in the appdata file here is wrong https://github.com/ksnip/ksnip/blob/master/desktop/ksnip.appdata.xml#L3 it should be `org.ksnip.ksnip`
The launchable tag should be updated as well https://github.com/ksnip/ksnip/blob/master/desktop/ksnip.appdata.xml#L30

Those are not only requirements of Flathub but are the recommendations by the appstream & desktop specs.

org.ksnip.ksnip.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - "/include"
+  - "/lib/cmake"
+  - "/lib/mkspecs"
+  - "/mkspecs"
+modules:
+  - name: kColorPicker
+    buildsystem: cmake-ninja
+    sources:
+      - type: git
+        url: https://github.com/ksnip/kColorPicker.git
```
bilelmoussaoui, line 28: This should be a tagged release or a specific commit at least for builds reproducibility

org.ksnip.ksnip.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - name: kImageAnnotator
+    buildsystem: cmake-ninja
+    sources:
+      - type: git
+        url: https://github.com/ksnip/kImageAnnotator.git
```
bilelmoussaoui, line 33: Same

org.ksnip.ksnip.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - name: ksnip
+    buildsystem: cmake-ninja
+    post-install:
+      - install -Dm644 /app/share/pixmaps/ksnip.svg /app/share/icons/hicolor/scalable/apps/ksnip.svg
```
bilelmoussaoui, line 37: This should be fixed upstream as well. `pixmaps` is a deprecated directory and have been the case for a long time.

org.ksnip.ksnip.yaml
```diff
@@ -0,0 +1,41 @@
+app-id: org.ksnip.ksnip
+runtime: org.kde.Platform
+runtime-version: '5.14'
+sdk: org.kde.Sdk
+command: ksnip
+rename-icon: ksnip
+rename-appdata-file: ksnip.appdata.xml
+rename-desktop-file: ksnip.desktop
```
bilelmoussaoui, line 8: Not a blocker: but this could be renamed upstream as well

LyzardKing:
@DamirPorobic
Do you see any issues with changing the upstream id?
Is setting tags in kColorPicker and kImageAnnotator a problem?
@bilelmoussaoui  Is it not possible to set the source as the "latest" tag? It would simplify having to change them all manually..

bilelmoussaoui:
> @DamirPorobic
> Do you see any issues with changing the upstream id?
> Is setting tags in kColorPicker and kImageAnnotator a problem?
> @bilelmoussaoui Is it not possible to set the source as the "latest" tag? It would simplify having to change them all manually..

If you change the id upstream, please make sure to add a `<provides><id>$theold-id</id></provides>` which will tell to the stores that it's the same app, just with a new ID.

I don't understand what you mean by the latest tag? If you mean a way to automatically update it whenever there's a tagged release, sadly not yet. We have a bot that automatically updates the URL's but only works for tarballs and doesn't support Github yet.

LyzardKing:
Yes, automatically update on a new tagged release.
That would be very useful.
For now I'll change it and manually set the tags.
Is this something that will exist in the future?

org.ksnip.ksnip.yaml
DamirPorobic: What would be here the correct name?  `ksnip.ksnip.appdata.xml`?

DamirPorobic: Would be `org.ksnip.ksnip.appdata.xml`?

DamirPorobic:
@nedrichards 
launchable tag should be `ksnip.ksnip.desktop` or `org.ksnip.ksnip.desktop`?
The provides tag should be `<provides><id>ksnip</id></provides>`?

@LyzardKing I'll take of that, today or tomorrow.

Regarding the tags, would it be possible to tag the last commit with a "continuous" tag and use that tag in the yaml to get always the latest version? The tag could be moved forward for all non tagged commits. We have such a functionality for ksnip but not for the dependencies.

bilelmoussaoui:
> org.ksnip.ksnip.desk

The launchable tag should have the desktop file name with the `.desktop`  extension. It's mostly to tell software centers how to run your application when it's installed.
The provides tag looks good to me.

LyzardKing:
> Regarding the tags, would it be possible to tag the last commit with a "continuous" tag and use that tag in the yaml to get always the latest version? The tag could be moved forward for all non tagged commits. We have such a functionality for ksnip but not for the dependencies.

I suppose that if there's a `continuous` tag it should be fine for the flatpak tool itself..
it it valid from a reproducible build standpoint @bilelmoussaoui ?

nedrichards:
bot, build org.ksnip.ksnip

nedrichards:
@LyzardKing from the [App Requirements](https://github.com/flathub/flathub/wiki/App-Requirements). 

> When building from a git tag, both the tag name and the commit id should be specified, like so:
> 
>    "tag": "1.0.4",
>    "commit": "cdfb19b90587bc0c44404fae30c139f9ec1cca5c"
> 
> This ensures that the build is reproducible, since tag can change their values over time. A commit can also be specified without a tag, however tag names are encouraged as they are better for readability.

We can discuss ways to automate releases in sync with upstream - it'd be great to offer this for apps that build from source although the point of having a PR with an automatically generated testable artifact before releasing is not insignificant.

DamirPorobic:
@nedrichards the last build failed with following errors:
```
```
The first one seems simple, just created the rating with the [OARS Generator](https://hughsie.github.io/oars/generate.html):
`<content_rating type="oars-1.0" /> `

The second is some kind of information about every release? 
Something like this?
```
  <releases>
    <release type="stable" version="0.2.0" date="2020-03-14T00:00:00Z">
      <description>
        <p>This release fixes the following bug:</p>
        <ul>
          <li>The CPU no longer overheats when you hold down spacebar</li>
        </ul>
      </description>
    </release>
    <release type="stable" version="0.1.0" date="2020-01-10T00:00:00Z">
      <description>
        <p>This release adds the following features:</p>
        <ul>
          <li>Now plays a "zap" sound on every character input</li>
        </ul>
      </description>
    </release>
  </releases>
```

DamirPorobic:
I've added the missing tags upstream.

nedrichards:
bot, build org.ksnip.ksnip

For sure, @LyzardKing you can use the command as well if you push an update btw.

nedrichards:
OK - this seems to work fine in local testing and passes the metadata tests. Now all it needs is the specific commits and it's good to merge.

LyzardKing:
bot, build org.ksnip.ksnip

DamirPorobic:
@LyzardKing should I cherry-pick the commits required for building the flatpak into a new `1.7.3` version of ksnip? Currently they're mixed with features that should go into `1.8.0` which is still work in progress, actually just started.

LyzardKing:
No need, I selected the commit directly.
In kColorPicker and kImageAnnotator I selected the latest tags.. are they too old?

nedrichards:
@LyzardKing who else from upstream needs commit rights btw. I assume @DamirPorobic but if there's anyone else we can add them easily too.

LyzardKing:
Just me and @DamirPorobic (the proper upstream) for now I guess...

PR #1679, opened 2020-08-01
Added org.freedesktop.LinuxAudio.Plugins.ArtyFX

hfiguiere:
This will replace org.freedesktop.LinuxAudio.Lv2Plugins.ArtyFX that I will deprecate.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "shared-modules"]
+	path = shared-modules
+	url = https://github.com/flathub/shared-modules.git
--- a/flathub.json
+++ b/flathub.json
@@ -1,0 +1,3 @@
+{
+    "skip-icons-check": true
+}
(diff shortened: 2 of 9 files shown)


hfiguiere:
bot, build org.freedesktop.LinuxAudio.Plugins.ArtyFX

barthalion:
Is extension ID really that big deal that we need to rename it all?

hfiguiere:
After the original design was rolled out, I learned that I could have multiple `merge-dirs` so I could have one flatpak for the same set of plugins in different format. So the idea is to have ONE extension point rule them all. This is the penultimate round of renaming.

But the goal is:

1. reduce the number of individual flatpak without impact to the user.
2. reduce the number of extension points

On overall this simplify the whole thing, given that 20.08 is around the corner and we'll need to branch for 20.08 as well.

PR #1680, opened 2020-08-01
Added org.freedesktop.LinuxAudio.Plugins.AVLDrums

hfiguiere:
This will replace org.freedesktop.LinuxAudio.Lv2Plugins.AVLDrums that I will deprecate


hfiguiere:
bot, build org.freedesktop.LinuxAudio.Plugins.AVLDrums

PR #1681, opened 2020-08-01
Add org.freedesktop.LinuxAudio.Plugins.DrumGizmo

hfiguiere:
Replaces org.freedesktop.LinuxAudio.Lv2Plugins.DrumGizmo that I will deprecate


hfiguiere:
bot, build org.freedesktop.LinuxAudio.Plugins.DrumGizmo

PR #1682, opened 2020-08-01
Add org.freedesktop.LinuxAudio.Plugins.VL1Emulator

hfiguiere:
This replaces org.freedesktop.LinuxAudio.Lv2Plugins.VL1Emulator that I will deprecate.


hfiguiere:
bot, build org.freedesktop.LinuxAudio.Plugins.VL1Emulator

PR #1684, opened 2020-08-01
Add org.freedesktop.LinuxAudio.Plugins.SoSynthLV2

hfiguiere:
This replaces org.freedesktop.LinuxAudio.Lv2Plugins.SoSynthLV2 that I will deprecate.


hfiguiere:
bot, build org.freedesktop.LinuxAudio.Plugins.SoSynthLV2

PR #1685, opened 2020-08-01
Add org.freedesktop.LinuxAudio.Plugins.setBfree

hfiguiere:
This replaces org.freedesktop.LinuxAudio.Lv2Plugins.setBfree that I will deprecate


hfiguiere:
bot, build org.freedesktop.LinuxAudio.Plugins.setBfree

PR #1686, opened 2020-08-01
Add org.freedesktop.LinuxAudio.Plugins.MDA

hfiguiere:
Replaces org.freedesktop.LinuxAudio.Lv2Plugins.MDA that I will deprecate


hfiguiere:
bot, build org.freedesktop.LinuxAudio.Plugins.MDA

PR #1690, opened 2020-08-02
Add org.jitsi.jitsi-meet

ananace:
Upstream has been asked about Flatpak support [[1]][[2]], but are currently not willing to spend their own limited time on setting up and maintaining a second Linux build pipeline (understandably seeing the lack of Flatpak support in electron-builder), so I've here written up a manifest that uses unappimage to repackage their existing AppImage build into a Flatpak.
That way, people who want to use Jitsi Meet's desktop app without manually handling their AppImage will still be able to do so.

[1]: https://github.com/jitsi/jitsi-meet-electron/issues/320
[2]: https://github.com/jitsi/jitsi-meet-electron/issues/426


org.jitsi.jitsi-meet.appdata.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
```
bilelmoussaoui, line 2: This should be upstream.

org.jitsi.jitsi-meet.yaml
```diff
@@ -0,0 +1,60 @@
...
+  - name: unappimage
+    buildsystem: simple
+    build-commands:
+      - make -C squashfs-tools -j ${FLATPAK_BUILDER_N_JOBS} install INSTALL_DIR=/app/bin
+    sources:
+      - type: git
+        url: https://github.com/refi64/unappimage
+        commit: d7f86f2a0d7ec3a69211125207d5f127386b849a
+
+  - name: jitsi-meet
```
bilelmoussaoui, line 28: Have you tried to build it from source?

ananace: Was poking around with that for a few days, but wasn't able to get it all to behave properly, so I decided to get a working flatpak up before I continue poking around with a from-source build.

org.jitsi.jitsi-meet.appdata.xml
ananace: Do you mean a literal `"upstream"` type? Considering the line it's a comment on.
Or that the entire appdata file should be from upstream? (They don't generate one, and have so far not wanted to accept any changes to their code for AppImage-unrelated things)

bilelmoussaoui: Except appstream is not a Flatpak specific thing but a freedesktop standard.

ananace: Yep, and technically AppImages require an appstream file too, but the electron-builder utility seems to ignore that requirement.
And seeing as Jitsi Meet is using electron-builder to build their AppImages, with no special configuration or additional files, only packing what is generated by that tool...

org.jitsi.jitsi-meet.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - install -Dm644 org.jitsi.jitsi-meet.png -t /app/share/icons/hicolor/512x512/apps
+
+      - mv squashfs-root /app/bin/jitsi-meet
+    sources:
+      - type: script
+        dest-filename: jitsi-meet-run
+        commands:
+          - zypak-wrapper /app/bin/jitsi-meet/jitsi-meet "$@"
+      - type: file
+        path: org.jitsi.jitsi-meet.appdata.xml
```
LeandroStanger, line 51: New appdata files should use metainfo instead. https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html

nedrichards:
bot, build org.jitsi.jitsi-meet

org.jitsi.jitsi-meet.metainfo.xml
```diff
@@ -0,0 +1,26 @@
...
+    <id type="desktop">org.jitsi.jitsi-meet</id>
+    <launchable type="desktop-id">org.jitsi.jitsi-meet.desktop</launchable>
+    <name>Jitsi Meet</name>
+    <summary>Jitsi Meet - Secure, Simple and Scalable Video Conferences</summary>
+    <metadata_license>CC0-1.0</metadata_license>
+    <project_license>Apache-2.0</project_license>
+    <developer_name>Jitsi Team</developer_name>
+    <url type="homepage">https://jitsi.org</url>
+    <description>
+        <p>Desktop application for Jitsi Meet built with Electron.</p>
```
nedrichards, line 12: I understand that this is the description from the github readme but it's really bad appstore copy ;-) it'd be much more helpful for users to use the relevant copy from here for example: https://play.google.com/store/apps/details?id=org.jitsi.meet&hl=en_GB

nedrichards:
Appdata screenshot didn't download, but otherwise it's all good!

> • url-not-found         : <screenshot> failed to download (HTTP 500: Internal Server Error) [https://gitcdn.link/repo/jitsi/jitsi-meet-electron/master/screenshot.png]

org.jitsi.jitsi-meet.metainfo.xml
ananace: That's a really good idea, I didn't want to write something myself, but never thought to steal a description from somewhere else. (Disregarding the fact that the play store app is not exactly the same thing)

nedrichards:
bot, build org.jitsi.jitsi-meet

org.jitsi.jitsi-meet.appdata.xml
bilelmoussaoui: meh, electron-builder isn't doing Linux users any good

org.jitsi.jitsi-meet.yaml
bilelmoussaoui: makes sense, it should pretty straightforward in most of the cases. Not a blocker though

org.jitsi.jitsi-meet.yaml
```diff
@@ -0,0 +1,56 @@
+---
+app-id: org.jitsi.jitsi-meet
+base: org.electronjs.Electron2.BaseApp
+base-version: '19.08'
+runtime: org.freedesktop.Platform
+runtime-version: '19.08'
+sdk: org.freedesktop.Sdk
+command: jitsi-meet-run
+separate-locales: false
+tags: []
```
bilelmoussaoui, line 10:
```suggestion

```

org.jitsi.jitsi-meet.yaml
```diff
@@ -0,0 +1,56 @@
...
+        url: https://github.com/refi64/unappimage
+        commit: d7f86f2a0d7ec3a69211125207d5f127386b849a
+
+  - name: jitsi-meet
+    buildsystem: simple
+    build-commands:
+      - unappimage jitsi-meet-x86_64.AppImage
+      - rm jitsi-meet-x86_64.AppImage
+
+      - sed -e 's/Exec=.*/Exec=jitsi-meet-run/'
```
bilelmoussaoui, line 34: You can use desktop-file-edit here btw

org.jitsi.jitsi-meet.metainfo.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
```
bilelmoussaoui, line 2:
```suggestion
<component type="desktop-application">
```
the desktop component shouldn't be used in newly written appdata files

org.jitsi.jitsi-meet.yaml
ananace: Don't think `desktop-file-edit` can change the `Exec` key though?

bilelmoussaoui: It can change the value of any key I think

ananace: It can change the value of any extended key (`X-*`), but for default keys there are only setters for Name, GenericName, Comment, and Icon. (As well as list adders/removers for categories, mime-types, and only show in environments)

So no, it can't change `Exec` - at least the version that's available in org.freedesktop.Sdk.

bilelmoussaoui: `desktop-file-edit --set-key=Exec --set-value=something $desktop-file` works just fine here

bilelmoussaoui:
bot, build org.jitsi.jitsi-meet

org.jitsi.jitsi-meet.yaml
ananace: Yeah, I'm not able to recreate the exact error anymore here either, but the application shouted at me when I tried before. Maybe I managed to get an input error?

PR #1691, opened 2020-08-03
Add com.system76.Popsicle

ids1024:


com.system76.Popsicle.yaml
```diff
@@ -0,0 +1,35 @@
...
+- org.freedesktop.Sdk.Extension.rust-stable
+build-options:
+  append-path: /usr/lib/sdk/rust-stable/bin
+  env:
+    CARGO_HOME: /run/build/popsicle/cargo
+finish-args:
+- --share=ipc
+- --socket=fallback-x11
+- --socket=wayland
+- --device=dri
```
bilelmoussaoui, line 15:
```suggestion
```
already included in device=all.

com.system76.Popsicle.yaml
```diff
@@ -0,0 +1,35 @@
...
+- --device=all
+- --filesystem=home
```
bilelmoussaoui, line 17: I wonder why this permission is necessary, GtkNativeFileChooser can't be used here?

bilelmoussaoui:
bot, build com.system76.Popsicle

nedrichards:
Is there anyone else from upstream who will need access when added?

com.system76.Popsicle.yaml
ids1024: Right. I don't think migrating Popsicle to using `GtkNativeFileChooser` would pose any difficulties...

Until very recently recently Popsicle couldn't run in a sandbox, and I don't believe it's been tested on Windows or Mac. So there wasn't any difference, but it looks like switching to `GtkNativeFileChooser` would be a good idea now.

ids1024:
> Is there anyone else from upstream who will need access when added?

You can add @jackpot51 and @mmstick.

com.system76.Popsicle.yaml
bilelmoussaoui: Not a blocker, but it would be nice to land that in the future. Merged

ids1024: Yep. That, and potentially using UDisks2 to find block devices so `--device=all` isn't needed.

bilelmoussaoui: Ah didn't catch this one

PR #1693, opened 2020-08-04
Add org.freedesktop.sdk.extension.haskell

putinscousin:
Adds the Haskell compiler GHC, cabal and stack as a sdk extension.

[Haskell](https://www.haskell.org/)
[Cabal](https://www.haskell.org/cabal/)
[Stack](https://docs.haskellstack.org/en/stable/README/)

I'm not sure how to handle the different licenses. Where should these be added?
[Haskell license](https://gitlab.haskell.org/ghc/ghc/blob/master/LICENSE)
[Cabal license](https://github.com/haskell/cabal/blob/master/LICENSE)
[Stack license](https://github.com/commercialhaskell/stack/blob/master/LICENSE)

### DIFF
--- a/flathub.json
+++ b/flathub.json
@@ -1,5 +1,4 @@
 {
     "skip-appstream-check": true,
-    "skip-icons-check": true,
-    "only-arches": ["x86_64"]
+    "skip-icons-check": true
 }
--- a/flathub.json
+++ b/flathub.json
@@ -1,0 +1,5 @@
+{
+    "skip-appstream-check": true,
+    "skip-icons-check": true,
+    "only-arches": ["x86_64"]
+}
--- a/org.freedesktop.Sdk.Extension.haskell.json
+++ b/org.freedesktop.Sdk.Extension.haskell.json
@@ -12,7 +12,11 @@
     },
     "modules": [
         {
-            "name": "numa",
+            "name": "libnuma",
+            "buildsystem": "autotools",
+            "make-args": [
+                "install"
+            ],
             "only-arches": [
                 "aarch64"
             ],
--- a/org.freedesktop.Sdk.Extension.haskell.json
+++ b/org.freedesktop.Sdk.Extension.haskell.json
@@ -8,6 +8,7 @@
     "separate-locales": false,
     "appstream-compose": false,
     "build-options": {
+        "prepend-ld-library-path": "/usr/lib/sdk/haskell/lib/",
         "prefix": "/usr/lib/sdk/haskell"
     },
     "modules": [
@@ -22,15 +23,10 @@
             ],
             "sources": [
                 {
-                    "type": "archive",
-                    "url": "https://github.com/numactl/numactl/releases/download/v2.0.13/numactl-2.0.13.tar.gz",
-                    "x-checker-data": {
-                        "type": "html",
-                        "url": "https://github.com/numactl/numactl/releases/latest",
-                        "version-pattern": "/numactl/stack/releases/tag/v((?:\\d+\\.)?(?:\\d+\\.)?(?:\\d+))",
-                        "url-template": "https://github.com/numactl/numactl/releases/download/v$version/stack-$version-linux-x86_64.tar.gz"
-                    },
-                    "sha256": "991e254b867eb5951a44d2ae0bf1996a8ef0209e026911ef6c3ef4caf6f58c9a"
+                    "type": "git",
+                    "url": "https://github.com/numactl/numactl.git",
+                    "tag": "v2.0.13",
+                    "commit": "5d9f16722e3df49dc618a9f361bd482559695db7"
                 }
             ],
             "cleanup": [
--- a/org.freedesktop.Sdk.Extension.haskell.json
+++ b/org.freedesktop.Sdk.Extension.haskell.json
@@ -20,10 +20,18 @@
             "sources": [
                 {
                     "type": "archive",
-                    "only-arches": ["x86_64"],
-                    "url": "https://downloads.haskell.org/~ghc/8.10.1/ghc-8.10.1-x86_64-deb9-linux.tar.xz",
-                    "sha512": "0df7c24e095bc60f761c953a292506c5068d84cef4b78bcf598eccd6d2397dd39b809576c37eca649c192eebfbf8d5b44ca3be06e149ef46bfd20e8690d6024a"
-
+                    "only-arches": [
+                        "x86_64"
+                    ],
+                    "url": "https://downloads.haskell.org/~ghc/8.10.1/ghc-8.10.1-x86_64-fedora27-linux.tar.xz",
+                    "x-checker-data": {
+                        "type": "html",
+                        "url": "https://www.haskell.org/ghc/download.html",
+                        "version-pattern": "((?:\\d+\\.)(?:\\d+\\.)(?:\\d+))",
+                        "url-template": "https://downloads.haskell.org/~ghc/$version/ghc-$version-x86_64-fedora27-linux.tar.xz"
+                    },
+                    "sha256": "3c4cd72b4806045779739e8f5d1658e30e57123d88c2c8966422cdbcae448470",
+                    "size": 214693280
                 }
             ]
         },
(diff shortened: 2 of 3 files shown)


flathub.json
```diff
@@ -0,0 +1,5 @@
+{
+    "skip-appstream-check": true,
+    "skip-icons-check": true,
+    "only-arches": ["x86_64"]
```
barthalion, line 4: I think that ideally it should be build from source for all architectures if possible, or at least provide repackaged binaries for x86_64 and aarch64.

org.freedesktop.Sdk.Extension.haskell.json
```diff
@@ -0,0 +1,96 @@
+{
+    "id": "org.freedesktop.Sdk.Extension.haskell",
```
barthalion, line 2: Wouldn't `ghc` a better name than `haskell`?

org.freedesktop.Sdk.Extension.haskell.json
```diff
@@ -0,0 +1,96 @@
...
+        {
+            "name": "ghc",
+            "buildsystem": "autotools",
+            "make-args": [
+                "install"
+            ],
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://downloads.haskell.org/~ghc/latest/ghc-8.10.1-x86_64-fedora27-linux.tar.xz",
```
barthalion, line 23: Should use flatpak-external-data-checker for detecting when updates are available.

org.freedesktop.Sdk.Extension.haskell.json
```diff
@@ -0,0 +1,96 @@
...
+                }
+            ]
+        },
+        {
+            "name": "cabal",
+            "buildsystem": "simple",
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://downloads.haskell.org/~cabal/cabal-install-3.2.0.0/cabal-install-3.2.0.0-x86_64-unknown-linux.tar.xz",
```
barthalion, line 34: Should use flatpak-external-data-checker for detecting when updates are available.

org.freedesktop.Sdk.Extension.haskell.json
```diff
@@ -0,0 +1,96 @@
...
+                "install -C ./cabal ${FLATPAK_DEST}/bin/"
+            ]
+        },
+        {
+            "name": "stack",
+            "buildsystem": "simple",
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://get.haskellstack.org/stable/linux-x86_64.tar.gz",
```
barthalion, line 49: It should use versioned URL. This way the checksum can change any minute.

org.freedesktop.Sdk.Extension.haskell.json
```diff
@@ -0,0 +1,96 @@
...
+        {
+            "name": "scripts",
+            "buildsystem": "simple",
+            "sources": [
+                {
+                    "type": "script",
+                    "commands": [
+                        ""
+                    ],
+                    "dest-filename": "install.sh"
```
barthalion, line 66: Just drop it, no point in shipping an empty script.

barthalion:
bot, build org.freedesktop.Sdk.Extension.haskell

nedrichards:
bot, build org.freedesktop.Sdk.Extension.haskell

flathub.json
nedrichards: Existing haskell users on flathub are split. These two are stuck on an old unsupported runtime and only support x86: https://github.com/flathub/org.freedesktop.Bustle and  https://github.com/flathub/com.lettier.movie-monad wheras https://github.com/flathub/org.hedgewars.Hedgewars gets more by getting binaries for other architectures.

nedrichards:
fwiw sources say that in principle https://hub.darcs.net/thielema/cabal-flatpak/ supports other architectures.

org.freedesktop.Sdk.Extension.haskell.json
putinscousin: I'll try to add that later today for all the links.

org.freedesktop.Sdk.Extension.haskell.json
putinscousin: I'd rather name it `haskell-platform`, since that includes cabal and stack as well.
Or should we split it up into something like `ghc` and `haskell-plaform`?

nedrichards:
bot, build org.freedesktop.Sdk.Extension.haskell

@putinscousin you can use that command to run test builds when you push new commits.

putinscousin:
I added the arches `arm`, `aarch64` and `i386` for `ghc`.
So we at least have that until I figure out how to compile `cabal` and `stack` from source.

bot, build org.freedesktop.Sdk.Extension.haskell
@nedrichards thanks, I thought only members could do that.

bilelmoussaoui:
> I added the arches `arm`, `aarch64` and `i386` for `ghc`.
> So we at least have that until I figure out how to compile `cabal` and `stack` from source.
> 

> @nedrichards thanks, I thought only members could do that.

i386 & ghc are not needed.

barthalion:
Build failed due to this error: `/usr/lib/sdk/haskell/lib/ghc-8.10.1/bin/ghc-pkg: error while loading shared libraries: libtinfo.so.5: cannot open shared object file: No such file or directory`

ncurses can be built to provide ABI version 5, for example: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=ncurses5-compat-libs

wjt:
> fwiw sources say that in principle https://hub.darcs.net/thielema/cabal-flatpak/ supports other architectures.

Hi, I'm the source. I happen to have been working to update Bustle to a more modern runtime & toolchain using cabal-flatpak in the last few weeks.

https://github.com/flathub/org.freedesktop.Bustle/pull/8

The manifests generated by cabal-flatpak just consist of a `ghc` binary plus a list of source packages to build. So, aarch64 support which includes `ghc` but not `stack` & `cabal` would still be of interest.

That said, having `cabal` available within a Flatpak environment on at least one arch would be very interesting. As detailed in that PR, to get a usable manifest I had to generate the build plan (with `cabal new-build --dry-run`) within the sandbox rather than on my host system, to be sure it sees the same package versions as come with the GHC being used in the Flatpak build. (To bootstrap this I generated the plan on the host system, generated the manifest, ran it to the point where `cabal` had been installed, and then did the whole thing again.)

If this SDK extension were available the process would be much simpler:

1. generate the cabal build plan
2. generate the manifest
3. build & enjoy

wjt:
> Build failed due to this error: `/usr/lib/sdk/haskell/lib/ghc-8.10.1/bin/ghc-pkg: error while loading shared libraries: libtinfo.so.5: cannot open shared object file: No such file or directory`
> 
> ncurses can be built to provide ABI version 5, for example: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=ncurses5-compat-libs

The manifest that cabal-flatpak generates has [`ln -s /usr/lib/x86_64-linux-gnu/libtinfo.so /app/lib/libtinfo.so.5`](https://github.com/flathub/org.freedesktop.Bustle/blob/3d1297c91796d4e06d149f97c17e7026f0d4da57/org.freedesktop.Bustle.json#L52). What's funny about this is that indeed the libtinfo in 19.08 does not have that soname!

```
[📦 org.freedesktop.Sdk ~]$ ls -l /usr/lib/x86_64-linux-gnu/libtinfo.so*
lrwxrwxrwx 1 nfsnobody nfsnobody     13 Jul 17 08:18 /usr/lib/x86_64-linux-gnu/libtinfo.so -> libtinfo.so.6
lrwxrwxrwx 1 nfsnobody nfsnobody     15 Jul 17 08:18 /usr/lib/x86_64-linux-gnu/libtinfo.so.6 -> libtinfo.so.6.1
-rwxr-xr-x 4 nfsnobody nfsnobody 240568 Jan  1  1970 /usr/lib/x86_64-linux-gnu/libtinfo.so.6.1
```

And yet GHC works fine. It does emit lots of warnings like this:

```
/app/lib/ghc-8.6.5/bin/ghc: /app/lib/libtinfo.so.5: no version information available (required by /app/lib/ghc-8.6.5/bin/../haskeline-0.7.4.3/libHShaskeline-0.7.4.3-ghc8.6.5.so)
```

putinscousin:
I think the easiest solution would be the one in the Hedgewars flatpak (symlink `libtinfo.so.6` to `libtinfo.so.5`).

A little off topic but I'm currently on vacation, so I won't be able to spend a lot of time on this until I'm back home.

putinscousin:
bot, build org.freedesktop.Sdk.Extension.haskell

putinscousin:
The aarch64 build is still failing because of `libnuma`, but at least x86_64 works for now.

Is there a way to stop the external data checker from following redirects? The URL it provides for stack is quite useless because the link expires after a while.

org.freedesktop.Sdk.Extension.haskell.json
```diff
@@ -0,0 +1,155 @@
...
+                    "type": "archive",
+                    "only-arches": [
+                        "x86_64"
+                    ],
+                    "url": "https://github.com/commercialhaskell/stack/releases/download/v2.3.3/stack-2.3.3-linux-x86_64.tar.gz",
+                    "x-checker-data": {
+                        "type": "html",
+                        "url": "https://github.com/commercialhaskell/stack/releases/latest",
+                        "version-pattern": "/commercialhaskell/stack/releases/tag/v((?:\\d+\\.)?(?:\\d+\\.)?(?:\\d+))",
+                        "url-template": "https://github.com/commercialhaskell/stack/releases/download/v$version/stack-$version-linux-x86_64.tar.gz"
```
wjt, line 113: Here's the offending line:

https://github.com/flathub/flatpak-external-data-checker/blob/05a99d4cb6aaa924b7b03f31389e6eebde19af77/src/checkers/htmlchecker.py#L102

Some of the other checkers (always or optionally, I can't remember) replace the URL with the original one, you could make this a parameter of the html checker, or just change its default behaviour, though I bet there are now some projects which rely on the current behaviour. It's all rather ad-hoc :(

putinscousin:
bot, build org.freedesktop.Sdk.Extension.haskell

putinscousin:
bot, build org.freedesktop.Sdk.Extension.haskell

putinscousin:
bot, build org.freedesktop.Sdk.Extension.haskell

putinscousin:
The build of `ghc` finally works for both architectures.
Is it possible to merge this pull request for now, so we at least have `ghc` availabe for now?
Of maybe split it up in org.freedesktop.Sdk.Extension.ghc and .haskell-platform.

If anyone wants to help me compile `cabal` and `stack` from source I would highly appreciate that.

PR #1694, opened 2020-08-04
Add io.openrct2.OpenRCT2

takluyver:
I don't have the game data handy at the moment to test this, but this gets [OpenRCT2](https://openrct2.io/) building and launching as far as the prompt that asks for the game data.

This probably isn't ready to merge until someone can actually test it, but I hope that by opening a PR I might save someone else duplicating the work to get this far. If no-one else gets there first, I'll hopefully be able to retrieve my old CD of the game some time in the next few months and open a can of nostalgia. :roller_coaster:


io.openrct2.OpenRCT2.yaml
```diff
@@ -0,0 +1,37 @@
...
+  - --share=network
+  # OpenGL access
+  - --device=dri
+modules:
+  - shared-modules/SDL/SDL-1.2.15.json
+  - shared-modules/SDL/SDL_ttf-2.0.11.json
+  - libzip.yaml
+  - Jansson.yaml
+  - SpeexDSP.yaml
+  - OpenRCT2.yaml
```
bilelmoussaoui, line 32: Any reason for splitting each module on it's own file?

takluyver: I was mostly following the example of [another app](https://github.com/flathub/app.rednotebook.RedNotebook) I've looked at recently. Though I did also think it would make it easy to move some of these (e.g. `libzip`) to `shared-modules` later.

I'm happy enough to change it if the preference is to stick things together in a single file.

bilelmoussaoui: By convention the modules are split into different files if they are generated automatically by a script, but whatever works for you.

takluyver: Ooh, are there handy scripts for generating things like that automatically? That could save a bit of time.

nedrichards: https://github.com/flatpak/flatpak-builder-tools has a bunch of tools for different language buildsystems. patches welcome!

barthalion:
Have you contacted the upstream team if they would be interested in (co-)maintaining the app here?

takluyver:
Good point. I have now, on https://github.com/OpenRCT2/OpenRCT2/issues/6112

no-lstdcxxfs.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/src/openrct2/CMakeLists.txt b/src/openrct2/CMakeLists.txt
+index b0bd3e9..e937b45 100644
+--- a/src/openrct2/CMakeLists.txt
++++ b/src/openrct2/CMakeLists.txt
+@@ -25,7 +25,7 @@ SET_CHECK_CXX_FLAGS(${PROJECT_NAME})
+ # GCC / Clang likes us to pass the -lstdc++fs flag to link C++17 filesystem implementation.
+ if (NOT MINGW AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
+     if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+-        target_link_libraries(${PROJECT_NAME} stdc++fs)
++        #target_link_libraries(${PROJECT_NAME} stdc++fs)
```
janisozaur, line 10: How does it link with the lib removed?

barthalion:
bot, build io.openrct2.OpenRCT2

no-lstdcxxfs.patch
barthalion: Quoting GCC 9 change log:

```
Using the types and functions in <filesystem> does not require linking with -lstdc++fs now. 
```

barthalion:
```
builddir/files/share/appdata/io.openrct2.OpenRCT2.appdata.xml: FAILED:
```

no-lstdcxxfs.patch
takluyver: And in the build environment, it seems that you can't use `-lstdc++fs` - I got an error, which prompted this patch.

takluyver:
The appdata comes from upstream; I'm guessing there's no `<release>` tag because it's just a static file which isn't being modified for each version. Should I patch it here?

no-lstdcxxfs.patch
janisozaur: Interesting, we've been using such compilers without issues.

The quoted wording does not imply it's forbidden to link, so I wonder why the change and how best to detect it. I wonder how e.g. OpenMandriva works.

Can you follow up with patch upstream so we can test it?

janisozaur:
@takluyver if there's such change required, now it's a good time to submit it as we're nearing another release.

bilelmoussaoui:
> The appdata comes from upstream; I'm guessing there's no `<release>` tag because it's just a static file which isn't being modified for each version. Should I patch it here?

It makes sense to send the patch upstream as the release tags are useful for end users (displayed in every store out there that uses appstream)

no-lstdcxxfs.patch
barthalion: I suspect it's dependent on the distro. I think freedesktopsdk team decided to drop `libstdc++fs.a` from their GCC based on that change log entry. Arch still ships it as well; I'll ask around.

takluyver:
OK, I've made a quick upstream PR for that: https://github.com/OpenRCT2/OpenRCT2/pull/12623

no-lstdcxxfs.patch
takluyver: Happy to follow up - is an issue OK? I don't really know how to do anything more sophisticated than patching the line out entirely.

takluyver:
PR merged upstream; I've added the same change as a patch here until it's in a release.

barthalion:
bot, build io.openrct2.OpenRCT2

barthalion:
I bought RC2 on GoG and seems the app works fine. @janisozaur, who besides you should be added to the Flathub repo>

io.openrct2.OpenRCT2.yaml
barthalion: @takluyver  While it's ultimately your choice, I strongly prefer self-containted manifests instead of splitting modules to separate files.

io.openrct2.OpenRCT2.yaml
```diff
@@ -0,0 +1,37 @@
...
+  # Wayland access
+  - --socket=wayland
+  # Pulseaudio access
+  - --socket=pulseaudio
+  # Filesystem access
+  # to access local images and other files
+  - --filesystem=home
+  # Network access
+  # to access remote images
+  - --share=network
```
barthalion, line 23: Does it actually need network access?

io.openrct2.OpenRCT2.yaml
```diff
@@ -0,0 +1,37 @@
...
+sdk: org.freedesktop.Sdk
+command: openrct2
+rename-appdata-file: openrct2.appdata.xml
+rename-desktop-file: openrct2.desktop
+rename-icon: openrct2
+copy-icon: true
+finish-args:
+  # X11 + XShm access
+  - --share=ipc
+  - --socket=x11
```
barthalion, line 13: Should be `--socket=fallback-x11`.

io.openrct2.OpenRCT2.yaml
takluyver: I have no preference, so if you have a strong one, I'll follow that. :-)

io.openrct2.OpenRCT2.yaml
takluyver: I'm not sure - if I remove the flag, do you have time to try a build without it?

barthalion: Yeah, no problem.

janisozaur: In OpenRCT2 we have (compile-time opt-out) multiplayer support and soon to be done update checking

takluyver: Aha, thanks. I'll leave network support in place, then.

How does it work with the game data files - does it copy them to a hidden data directory when you select them, or does it keep using them from the folder you pointed to? This might affect what filesystem access it needs.

janisozaur: The RCT 2 assets are intact and used from their specified location. All the files we create are located inside "user data dir", by default `$XDG_CONFIG_HOME` or, if unset, `~/.config/OpenRCT2`. There's a command line switch which can be used to override this location. The files we store contain configuration, external objects, tracks, etc.

takluyver:
I've trimmed the permissions a bit - read-only access to the home directory and 'fallback' X11 access. Please do check that it still works!

io.openrct2.OpenRCT2.yaml
takluyver: Thanks. I think this should work with read-only access to the user's home directory (it will still have write access to its own XDG cache/config/home dirs).

barthalion:
bot, build io.openrct2.OpenRCT2

io.openrct2.OpenRCT2.yaml
```diff
@@ -0,0 +1,80 @@
...
+  - --socket=wayland
+  # Pulseaudio access
+  - --socket=pulseaudio
+  # Filesystem access - for game files
+  - --filesystem=home:ro
+  # Network access - for multiplayer
+  - --share=network
+  # OpenGL access
+  - --device=dri
+modules:
```
barthalion, line 24: I noticed there's few static libraries lying around. You should add a global or per-module `cleanup` array specifying files to delete. A global one should be something along this:

```
cleanup:
  - *.a
  - /share/cmake
```

takluyver: Done. Should `/lib/cmake` also be cleaned? It's only about 20K, so probably doesn't make much difference.

barthalion: Not like it actually matters, but in general yes as this is an application and not SDK extension/BaseApp.

barthalion:
bot, build io.openrct2.OpenRCT2

barthalion:
bot, build io.openrct2.OpenRCT2

rmnvgr:
I tried it out of curiosity as I attempted to flatpak OpenRCT2 a few months back, but gave up because of the `stdc++fs` thing.

The game works fine, however under Wayland there's some problems with the cursor:
- Cursors are bigger in the OpenRCT2 window than in my desktop
- The "pointer" cursor (when you hover over an object) is not the right one (it's a full hand instead of just a pointed index)
- When you right click and drag to move the view, instead of the view moving with your mouse, the view moves faster as you get further from the click origin.

I don't know if it comes from Wayland or OpenRCT2, but changing `--socket=fallback-x11 `--socket=x11` and removing `--socket=wayland` solves the issues. So maybe for now it should default to X11?

I also made some changes to support the 0.3.0 release, do you want me to make a PR on your branch?

takluyver:
Thanks @rmnvgr!

I can't try it for myself at the moment, but it looks like all of the issues you mention are pretty minor things that won't stop people playing and enjoying the game. I'm inclined to leave it using Wayland by default so that the bugs are visible and there's a chance of fixing them.

I'm hoping this will be merged soon, and then there will be a little separate repository for managing the recipe, and we can collaborate there. But if not, feel free to make a PR against this branch.

rmnvgr:
You're right, that's minor annoyances, except for the right click view dragging that becomes uncontrollable if you move your mouse more than a few pixel.

I'll wait for this PR to be merged to propose my changes for the 0.3.0 release then. I also made some changes to export the additional desktop files for opening scenarios savegames and uris.

barthalion:
/merge @janisozaur 

@janisozaur Please let us know if there's anyone else from the OpenRCT2 team who should be invited to the Flathub repo.

takluyver:
Thanks @barthalion !

@rmnvgr fire away with the PRs: https://github.com/flathub/io.openrct2.OpenRCT2

janisozaur:
I mentioned the flathub effort to team and will reply if anyone wants in as well. For now I suppose you could invite @Gymnasiast, he might be interested.

As for myself I don't know flatpak enough to be useful, but I own most of our cmake stuff and can provide info on packaging.

We usually want our users to be on the CI builds from `develop` branch, is it a use case for flatpak? Should we do anything upstream regarding flatpak support? Is flathub _the_ place to get flatpaks or is would it be useful if we included the file in our release assets as well? Any other things we should consider?

barthalion:
Ah well, I guess these are valid question to answer on our page in the first place…

Flathub is basically an app store for Linux, and most of what applies to Apple Store/Google Play/etc applies here. In the ideal world, applications are maintained by project's authors but not everyone is interested in packaging so we allow "downstream" maintainers as long as they reach out to authors. So it would be nice, but with @takluyver (and possibly @rmnvgr too?) involvement, no need to do anything.

The manifest itself is not very useful per se, however there are projects that do nightly builds after some minor changes or as the part of CI. Flathub offers a beta branch but it should be rather used for betas/release candidates, that being said, it's entirely up to you. There is no single file artifact to add to the release assets; you can just point users to https://flathub.org/apps/details/io.openrct2.OpenRCT2 and it will always serve the version as defined in the manifest.

Flathub is considered "the place" for Flatpaks. Fedora and Elementary have their own remotes but Flathub is the general purpose one where majority of (Flatpak) applications live.

takluyver:
> Flathub is considered "the place" for Flatpaks

But to be clear, a user can have multiple remotes configured, so they can install apps from different places - it's "the place" in a social sense, not as a technical limitation.

janisozaur:
What needs to be done to produce continuous/daily builds in the beta branch? It's it always "beta" or are we free to rename it?

rmnvgr:
@barthalion, yes I'd like to help maintain the OpenRCT2 flatpak manifest. Can you please add me as a collaborator on the io.openrct2.OpenRCT2 repo?

bilelmoussaoui:
> What needs to be done to produce continuous/daily builds in the beta branch? It's it always "beta" or are we free to rename it?

It's just beta, there are no daily/nightlies builds.


> @barthalion, yes I'd like to help maintain the OpenRCT2 flatpak manifest. Can you please add me as a collaborator on the io.openrct2.OpenRCT2 repo?

Done

takluyver:
AIUI, a commit to the `beta` branch will publish a new version in the beta channel. But this is designed for numbered betas/RCs, so there's no official mechanism for doing automatic releases on a schedule. It's probably possible to hack around that with a job that runs elsewhere which pushes a commit every night - I don't know if that's frowned upon.

Alternatively, you could integrate a Flatpak build into OpenRCT2's CI directly, and maintain a separate remote where people can get the development version. I haven't done that myself, but the Flatpak docs have a section on [hosting a repo](https://docs.flatpak.org/en/latest/hosting-a-repository.html).

Gymnasiast:
We always recommend our users use `develop`. It is also what most servers use. So if the `beta` branch can be used in that way, it would be preferable.

PR #1698, opened 2020-08-09
Add io.bit3.WhatsAppQT

tristanlins:
WhatsAppQT is a qt based desktop app for [WhatsApp Web](https://web.whatsapp.com/).

### DIFF
--- a/io.bit3.WhatsAppQT.yml
+++ b/io.bit3.WhatsAppQT.yml
@@ -8,6 +8,7 @@ command: WhatsAppQT
 finish-args:
 - "--share=ipc"
 - "--socket=fallback-x11"
+- "--socket=wayland"
 - "--share=network"
 - "--filesystem=xdg-download"
 - "--device=dri"
--- a/io.bit3.WhatsAppQT.yml
+++ b/io.bit3.WhatsAppQT.yml
@@ -14,6 +14,7 @@ finish-args:
 - "--device=dri"
 - "--env=QTWEBENGINEPROCESS_PATH=/app/bin/QtWebEngineProcess"
 - "--talk-name=org.kde.StatusNotifierWatcher"
+- "--own-name=org.kde.*"
 cleanup-commands:
 - /app/cleanup-BaseApp.sh
 modules:
--- a/io.bit3.WhatsAppQT.yml
+++ b/io.bit3.WhatsAppQT.yml
@@ -29,6 +29,6 @@ modules:
   sources:
   - type: git
     url: https://gitlab.com/bit3/whatsappqt.git
-    tag: v1.0.0
-    commit: 4e955390bbd7163cd0cd4b220896938097a485d8
+    tag: v1.0.1
+    commit: 743e0d17ceb5ed8d05775b8bb6f05f23830c4ad3
 
--- a/io.bit3.WhatsAppQT.yml
+++ b/io.bit3.WhatsAppQT.yml
@@ -7,8 +7,7 @@ base-version: '5.14'
 command: WhatsAppQT
 finish-args:
 - "--share=ipc"
-- "--socket=x11"
-- "--socket=wayland"
+- "--socket=fallback-x11"
 - "--share=network"
 - "--filesystem=home"
 - "--device=dri"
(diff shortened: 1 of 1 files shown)


tristanlins:
bot, build io.bit3.WhatsAppQT

tristanlins:
bot, build io.bit3.WhatsAppQT

io.bit3.WhatsAppQT.yml
```diff
@@ -0,0 +1,34 @@
...
+sdk: org.kde.Sdk
+base: io.qt.qtwebengine.BaseApp
+base-version: '5.14'
+command: WhatsAppQT
+finish-args:
+- "--share=ipc"
+- "--socket=x11"
+- "--socket=wayland"
+- "--share=network"
+- "--filesystem=home"
```
barthalion, line 13: Does it really need full home access?

io.bit3.WhatsAppQT.yml
```diff
@@ -0,0 +1,34 @@
+app-id: io.bit3.WhatsAppQT
+runtime: org.kde.Platform
+runtime-version: '5.14'
+sdk: org.kde.Sdk
+base: io.qt.qtwebengine.BaseApp
+base-version: '5.14'
+command: WhatsAppQT
+finish-args:
+- "--share=ipc"
+- "--socket=x11"
```
barthalion, line 10: Should be `fallback-x11`, assuming it actually works natively on Wayland.

io.bit3.WhatsAppQT.yml
tristanlins: I almost said it was personal preference. I don't like it when I want to send files and have to move them to Downloads first. And vice versa, if I want to download files, I want to save them directly to a certain location and not in Downloads to move them from there.

But I can switch that to "xdg-downloads" if that's preferred. I can still overwrite the access locally.

io.bit3.WhatsAppQT.yml
tristanlins: I will change it and try it out.

tristanlins:
bot, build io.bit3.WhatsAppQT

tristanlins:
@barthalion I have made the desired changes :-)

io.bit3.WhatsAppQT.yml
```diff
@@ -7,8 +7,7 @@ base-version: '5.14'
 command: WhatsAppQT
 finish-args:
 - "--share=ipc"
-- "--socket=x11"
-- "--socket=wayland"
+- "--socket=fallback-x11"
```
barthalion, line 10: Just to clarify, you still need `--socket=wayland` here, same for ThreemaQt.

tristanlins: Oops, I read the [documentation](https://docs.flatpak.org/en/latest/flatpak-command-reference.html?highlight=socket#) and thought at first that the Wayland socket would then be passed through. But the second time I read this I realized that I just misunderstood it. 🙈

tristanlins:
bot, build io.bit3.WhatsAppQT

barthalion:
The way Qt handles tray icons is broken:

```
failed to register service "org.kde.StatusNotifierItem-2-1"
```

You should add `--own-name=org.kde.*` to `finish-args`. It's a huge hole but there's no other way currently. With that fixed, LGTM.

tristanlins:
bot, build io.bit3.WhatsAppQT

tristanlins:
> The way Qt handles tray icons is broken:
> 
> ```
> failed to register service "org.kde.StatusNotifierItem-2-1"
> ```

Crazy, I haven't seen any such error message. But maybe there are also differences between `flatpak install` and` flatpak-builder --install`? I've added `--own-name=org.kde.*`.

barthalion:
You probably would if you tried to run TreemaQt and WhatsAppQt at the same time. Qt/KDE uses PID of the process to build service name used to register the icon and it's usually `2` in Flatpak sandbox, causing things to explode.

/merge

tristanlins:
@barthalion Ah, thx for explanation. 
I actually had double tray icons before. But have not seen a suitable message. Which was probably because I never started both clients at the same time via the terminal 🙊

PR #1749, opened 2020-08-15
Add org.freedesktop.LinuxAudio.Plugins.CAPS

hfiguiere:
This rename/replace org.freedesktop.LinuxAudio.LadspaPlugins.CAPS that I will deprecate


hfiguiere:
bot, build org.freedesktop.LinuxAudio.Plugins.CAPS

PR #1751, opened 2020-08-16
Add com.github.devalien.workspaces.json

DevAlien:
I'm the author of this App. Would be great to have it added to Flathub!

### DIFF
--- a/com.github.devalien.workspaces.json
+++ b/com.github.devalien.workspaces.json
@@ -1,0 +1,42 @@
+{
+    "app-id": "com.github.devalien.workspaces",
+    "runtime": "org.freedesktop.Platform",
+    "runtime-version": "19.08",
+    "sdk": "org.freedesktop.Sdk",
+    "base": "io.elementary.BaseApp",
+    "base-version": "juno-19.08",
+    "command": "com.github.devalien.workspaces",
+    "finish-args": [
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--filesystem=host",
+        "--talk-name=org.freedesktop.Flatpak"
+    ],
+    "cleanup": [
+        "/include",
+        "/lib/pkgconfig",
+        "/man",
+        "/share/doc",
+        "/share/gtk-doc",
+        "/share/man",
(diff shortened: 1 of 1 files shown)


DevAlien:
bot, build com.github.devalien.workspaces

barthalion:
Seems to have some icons missing.

image

I also guess you should enforce Elementary stylesheet as it looks slightly different here on vanilla GNOME than on your screenshots.

com.github.devalien.workspaces.json
```diff
@@ -0,0 +1,42 @@
...
+    "runtime-version": "19.08",
+    "sdk": "org.freedesktop.Sdk",
+    "base": "io.elementary.BaseApp",
+    "base-version": "juno-19.08",
+    "command": "com.github.devalien.workspaces",
+    "finish-args": [
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--filesystem=host",
```
barthalion, line 13: Why does it need full host access?

com.github.devalien.workspaces.json
```diff
@@ -0,0 +1,42 @@
...
+        "--talk-name=org.freedesktop.Flatpak"
```
barthalion, line 14: Just to double check, this is to run `flatpak-spawn --host`?

com.github.devalien.workspaces.json
DevAlien: Yes, besically it is a command launcher and you might need access to different things on your filesystem. But the app by default does not launch any command, the commands are defined by the user and then launched by the user itself

com.github.devalien.workspaces.json
DevAlien: as above

DevAlien:
I think I don't want to force elementary's stylesheet, I want to look more or less ok with anyone's theme.
Strange that is missing that icon, I tested in ubuntu and was ok, I will fix it now.

DevAlien:
bot, build com.github.devalien.workspaces

DevAlien:
bot, build com.github.devalien.workspaces

PR #1753, opened 2020-08-17
Add org.sugarlabs.Paint

tchx84:


tchx84:
bot, build org.sugarlabs.Paint

org.sugarlabs.Paint.json
```diff
@@ -0,0 +1,69 @@
...
+        "--socket=x11",
+        "--socket=pulseaudio",
+        "--share=ipc",
+        "--device=dri",
+        "--env=SUGAR_BUNDLE_ID=org.sugarlabs.Paint",
+        "--env=SUGAR_BUNDLE_PATH=/app/share/sugar/activities/Paint.activity"
+    ],
+    "modules": [
+        {
+            "name": "hello-world",
```
nedrichards, line 20: nitpick but I suspect this isn't *actually* the name ;-)

nedrichards:
When running:

> 1597742692.118042 ERROR root: No valid fill binaries. Using slower python code

I assume that's no big deal. I was also able to get:

> Traceback (most recent call last):
  File "/app/share/sugar/activities/Paint.activity/Area.py", line 1317, in internal_invert
    new_array, cairo.FORMAT_ARGB32, width, height)
TypeError: buffer is not long enough

But that may be "just a bug" ;-)

tchx84:
Good catch! :smile:

tchx84:
> > Traceback (most recent call last):
> > File "/app/share/sugar/activities/Paint.activity/Area.py", line 1317, in internal_invert
> > new_array, cairo.FORMAT_ARGB32, width, height)
> > TypeError: buffer is not long enough
> 
> But that may be "just a bug" ;-)

Can you share whatever image you are using  ? or the specs of the laptop you are using ? I haven't being able to reproduce this.

tchx84:
> When running:
> 
> > 1597742692.118042 ERROR root: No valid fill binaries. Using slower python code
> 

Known issue https://github.com/flathub/flathub/pull/1753#issuecomment-675768918

Shouldn't be a blocker, just means we are using a slower exec path without it.

tchx84:
bot build, org.sugarlabs.Paint

nedrichards:
Sure, image is here temporarily https://nickr.org/linux/test-image.paint then choose "Invert colours"

tchx84:
> Sure, image is here temporarily https://nickr.org/linux/test-image.paint then choose "Invert colours"

I see, funny thing, I don't even have a proper setup to test it :sweat: ... Paint doesn't support scaling, so I can't reproduce it because it simply crops the image to what is available on screen ( https://github.com/sugarlabs/paint-activity/issues/38 ), which I assume fits into that buffer.

I can only imagine that, for some reason, there is a inconsistency between the width and height and the actual pixels data given to the create_with_data method. Maybe it is always inconsistent but the issue doesn't manifest with smaller sizes, or the inconsistency is only triggered with such big images.

nedrichards:
Ah yes, HiDPI issues! The device is an ex-Endless Yoga 900 FWIW.

PR #1755, opened 2020-08-18
Add Gajim plugin: appindicator

a-andre:


a-andre:
bot, build org.gajim.Gajim.Plugin.appindicator

PR #1759, opened 2020-08-21
Add org.x.Warpinator

mtwebster:


mtwebster:
Thank you!

barthalion:
@mtwebster FYI the app-id should be a reverse DNS of a domain you "own", so probably `com.linuxmint.Warpinator`. It might be easier to handle the rename sooner rather than later because migrating users is a little bit wonky in Flatpak. It's enough to submit new PR with fixed ID and I'll handle the rest.

PR #1573, opened 2020-06-08
add org.gnome.Baobab

junjieyuan:
Signed-off-by: Junjie Yuan <junjieraw@gmail.com>

PR #1760, opened 2020-08-21
Add org.gnome.baobab

A6GibKm:
Continuation of #1573.


org.gnome.baobab.json
```diff
@@ -0,0 +1,38 @@
...
+    "sdk" : "org.gnome.Sdk",
+    "command" : "baobab",
+    "finish-args" : [
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--filesystem=host:ro",
+        "--filesystem=~/.var/app:ro",
+        "--talk-name=org.gtk.vfs.*",
+        "--system-talk-name=org.freedesktop.hostname1",
```
barthalion, line 14: What is it for, displaying local hostname?

A6GibKm: I do not know...  As far as I know it was in the previous pr and upstream.

barthalion: Seems so, without it, it just says "Computer". I'd drop it, it looks very uncommon to request it.

A6GibKm: ...but the immersion :disappointed:

A6GibKm:
It seems the app does not work in x11, even with the `--socket=x11` flag.

org.gnome.baobab.json
A6GibKm: I am not sure how I feel about 'just because it is uncommon'.

A6GibKm: Ok, so flatpak is not able to see `\` anyways. I am dropping this.

barthalion:
It works absolutely fine for me on X11.

A6GibKm:
Ok, might been an issue on my side. I will test later using X11.

barthalion:
bot, build org.gnome.baobab

nedrichards:
Yeah, this was working fine for me in x11. I'm still not entirely clear who needs to be added to this when merged? Other than that lgtm (helped me clean up 2gb of storage yesterday!)

PR #1775, opened 2020-08-26
Add com.thincast.client RDP client

akallabeth:
A RDWebFeed and RDP client based on FreeRDP

The build needs openh264 1.8.0 until https://github.com/cisco/openh264/issues/3305 is resolved.
to the admins: Is there a way to download the openh264 binary at installation time?


com.thincast.client.json
```diff
@@ -0,0 +1,292 @@
+{
+    "app-id": "com.thincast.client",
+    "runtime": "org.kde.Platform",
+    "runtime-version": "5.13",
```
LeandroStanger, line 4: Update runtime to 5.14
```suggestion
    "runtime-version": "5.14",
```

LeandroStanger:
bot, build com.thincast.client

bilelmoussaoui:
> A RDWebFeed and RDP client based on FreeRDP
> 
> The build needs openh264 1.8.0 until https://github.com/cisco/openh264/issues/3305 is resolved.
> to the admins: Is there a way to download the openh264 binary at installation time?

By using extra-data module type yes. See how other propriety apps are handled like Spotify.

com.thincast.client.json
akallabeth: Due to a bug in Qt 5.14 regarding NTLM authentication we have to stick with 5.13 for now.

akallabeth:
@bilelmoussaoui thank you, added that to the manifest.
Also (temporarily) added the updated `appdata.xml` (we will add the updated ones to the `tar.gz` files for future releases)

bilelmoussaoui:
bot, build com.thincast.client

(you can talk the bot as well btw)

com.thincast.client.json
```diff
@@ -0,0 +1,337 @@
...
+            "build-commands": [
+                "tar xf webaccess.tar.xz --strip-components=1 -C /app"
+            ],
+            "sources": [
+                {
+                    "type": "file",
+                    "dest-filename": "webaccess.tar.xz",
+                    "url": "https://thincast.com/software/download/artifact/by-hash/45acb93d207f2deaeba0b710547e9750e1d68f385249661104f7146d7e1902e4",
+                    "sha256": "45acb93d207f2deaeba0b710547e9750e1d68f385249661104f7146d7e1902e4",
+                    "only-arches": [
```
barthalion, line 286: No need to specify it here (and in other modules) with `flathub.json` limiting the architectures.

barthalion (changes requested):

com.thincast.client.json
akallabeth: true, but we are preparing other architectures too (they are just not ready yet).
Can remove this for the time being, of course.

barthalion: I think it's cleaner, but it's your choice.

com.thincast.client.json
```diff
@@ -0,0 +1,337 @@
...
+        "--share=ipc",
+        "--socket=x11",
+        "--socket=pcsc",
+        /* Needs to talk to the network */
+        "--share=network",
+        /* Play sounds redirected from guests */
+        "--socket=pulseaudio",
+        /* Allow rw access to download folder */
+        "--device=all",
+        "--filesystem=home",
```
barthalion, line 22: The comment above says _download folder_. We generally prefer filesystem permissions to be narrowed down as they're easy to be modified by users, should they want it.

com.thincast.client.appdata.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
barthalion, line 1: Ideally appdata files should be part of client/webaccess tarballs as this way you can actually maintain them as part of software release cycle but if you think it fits better here, ignore this comment.

barthalion:
bot, build com.thincast.client

akallabeth:
bot, build com.thincast.client

com.thincast.client.appdata.xml
akallabeth: @barthalion will do that for the next version. Need to wait for a new version to be build (that might take a few days until there are enough fixes)

com.thincast.client.json
akallabeth: updated the comment. The access is only used when the user selects a folder to redirect in the app and only allowing access to `Downloads` might be a bit too strict

com.thincast.client.json
akallabeth: I´d prefer to keep it that way.

PR #1779, opened 2020-08-30
Create flatpak app openerp-client (gtk2 desktop application)

rga-odoo:


bilelmoussaoui:
You're not supposed to include in-tree the shared-modules, you should use git submodules instead

org.flatpak.openerp.json
```diff
@@ -0,0 +1,169 @@
+{
+  "app-id": "org.flatpak.openerp",
```
bilelmoussaoui, line 2: Doesn't sound like the right id.

LeandroStanger:
```suggestion
  "app-id": "com.odoo.Openerp",
```

rga-odoo:
@bilelmoussaoui I have added shared modules as git submodules

org.flatpak.openerp.json
```diff
@@ -0,0 +1,169 @@
+{
+  "app-id": "com.odoo.Openerp",
+  "runtime": "org.freedesktop.Platform",
+  "runtime-version": "19.08",
+  "sdk": "org.freedesktop.Sdk",
+  "command": "openerp-client",
+  "finish-args": [
+    "--socket=x11",
```
nedrichards, line 8: you need to add --share=ipc if you use x11

nedrichards:
bot, build com.odoo.Openerp

org.flatpak.openerp.json
```diff
@@ -0,0 +1,169 @@
+{
```
barthalion, line 1: The file should be renamed to match app-id.

Eonfge:
Hey, I've upstreamed a shared module for python, which contains pygtk. Youre project would be an ideal candiate to use pygtk from the /shared-modules/ since it saves us all from double maintenance:

https://github.com/flathub/shared-modules/tree/master/pygtk

PR #1797, opened 2020-09-04
Add com.jetbrains.toolbox-app

x80486:
This is based on "packaged" AppImage format JetBrains provides — similar to `Unity Hub`.

> It builds fine, but it doesn't run quite yet :innocent: because it's missing `libcef.so` (and probably some other dependencies once that one is solved) but I can't find a way to provide it, so I was hoping somebody here more versed would know what dependency is needed for that one.


FakeShemp:
`libcef.so` is some electron library I think. Maybe this is an electron app.

x80486:
> `libcef.so` is some electron library I think. Maybe this is an electron app.

While I was searching for that dependency I read people talking about Chromium, etc. ...and I did try with both: `org.electronjs.Electron2.BaseApp` and `io.atom.electron.BaseApp` (version `19.08`), but it didn't work — same error message. That's why I decided to put here in such way, because I don't know what else to try.

FakeShemp:
I was wrong, it's a [Chromium Embedded Framework](https://github.com/chromiumembedded/cef) thingie.

barthalion:
Yeah, but I think it's "traditionally" provided as part of the tarball with the app.

x80486:
I will leave this one for the pros...I don't quite know how to package this one :innocent:

PR #1806, opened 2020-09-07
Add sound-juicer

hadess:
(yes, I keep adding my build scripts because my brain is small, and I'll do it when you're not looking anyway)

### DIFF
--- a/org.gnome.SoundJuicer.json
+++ b/org.gnome.SoundJuicer.json
@@ -67,6 +67,61 @@
                 }
             ]
         },
+        /* To encode in MP3 */
+        {
+          "name": "taglib",
+          "buildsystem": "cmake-ninja",
+          "config-opts": [
+            "-DBUILD_SHARED_LIBS=ON"
+          ],
+          "cleanup": [
+            "/include",
+            "/lib/*.so"
+          ],
+          "sources": [
+            {
+              "type": "archive",
+              "url": "http://taglib.github.io/releases/taglib-1.11.1.tar.gz",
+              "sha256": "b6d1a5a610aae6ff39d93de5efd0fdc787aa9e9dc1e7026fa4c961b26563526b"
+            }
+         ]
+        },
+        {
+            "name": "gst-plugins-good",
+            "buildsystem": "meson",
+            "config-opts": [
+                "--prefix=/app",
+                "-Dauto_features=disabled",
+                "-Dtaglib=enabled"
+            ],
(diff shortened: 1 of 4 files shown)


hadess:
bot, build org.gnome.SoundJuicer

org.gnome.SoundJuicer.json
```diff
@@ -0,0 +1,157 @@
...
+        /* Migrate DConf settings from the host */
+        "--metadata=X-DConf=migrate-path=/org/gnome/sound-juicer/",
+        /* optical media detection */
+        "--talk-name=org.gtk.vfs", "--talk-name=org.gtk.vfs.*",
+        /* Force volume monitor with optical media detection support */
+        "--env=GIO_USE_VOLUME_MONITOR=udisks2",
+        /* Ensure cdda gstreamer plugin is picked found for audio CD's */
+        "--env=GST_PLUGIN_PATH=/app/lib/codecs/lib/gstreamer-1.0"
+    ],
+    "build-options" : {
```
nedrichards, line 29: these should be unnecessary in 3.36 (as in - they're already set in the SDK)

nedrichards:
Other than OARS and release tags looks initially decent. Thanks for adding this, as it provides the invaluable opportunity to troll @rossburton about https://gitlab.gnome.org/GNOME/sound-juicer/-/blob/master/MAINTAINERS

hadess:
bot, build org.gnome.SoundJuicer

hadess:
bot, build org.gnome.SoundJuicer

hadess:
Not sure whether https://github.com/flatpak/flatpak/pull/3839 should be a blocker.

I've fixed all the warnings I could see during a normal run, and all the profiles listed are supported out-of-the-box, so ready to go for me.

nedrichards:
bot, build org.gnome.SoundJuicer

nedrichards:
I can't seem to actually make this see my USB CD drive, which is unfortunate. The dconf migrate issue isn't a blocker though I think.

hadess:
> I can't seem to actually make this see my USB CD drive, which is unfortunate. The dconf migrate issue isn't a blocker though I think.

Are you running this under GNOME? What's the output of `gio mount -li` from inside the sandbox?

nedrichards:
gio sees it - further testing on other devices suggests that this is a mechanical issue with the drive as much as anything else. Sorry for the false alarm.

PR #1815, opened 2020-09-09
Add org.gnome.FileRoller

A6GibKm:


A6GibKm:
Upstream issue: https://gitlab.gnome.org/GNOME/file-roller/-/issues/87

A6GibKm:
bot, build org.gnome.FileRoller

A6GibKm:
bot, build org.gnome.FileRoller

A6GibKm:
Created a PR upstream for the appdata: https://gitlab.gnome.org/GNOME/file-roller/-/merge_requests/43

nedrichards:
blocking: there isn't an app icon in the about dialog. 
non-blocking:  the file preview/open functionality isn't the best integrated into the flatpak system

@bilelmoussaoui FYI with the above fixed I'm good to merge.

A6GibKm:
> blocking: there isn't an app icon in the about dialog.

MIght be related to the icon being named as ArchiveManager originally?

barthalion:
Bundling `unrar` and `p7zip` will make the app much more useful.

A6GibKm:
Yes, I will do `7zip` now, but `unrar` might be less easy as it requires `extra-data` due to license issues, talking about that, it might be worth noticing that there are two apps on flathub which distribute it.

 I am not sure how to build it with, say `unrar`, I tried installing a `unrar` as a module but I was not able to find a difference.

barthalion:
> `unrar` might be less easy as it requires `extra-data` due to license issues

Unrar license allows redistribution.

A6GibKm:
Added unrar and p7zip, credits to @p1u3o [p1u3o/org.gnome.FileRoller](https://github.com/p1u3o/org.gnome.FileRoller).

Ok, I don't know the details but for all I know the unrar license is messy, I will leave it as extra-data for now.

A6GibKm:
bot, build org.gnome.FileRoller

barthalion:
Extra-data just leads to bad user experience and binds the application to external server instead of Flathub CDN. Unrar license is very clear that it can be distributed as part of other applications:

> The UnRAR sources may be used in any software to handle RAR
archives without limitations free of charge, but cannot be used
to re-create the RAR compression algorithm, which is proprietary.
Distribution of modified unRAR sources in separate form or as a
part of other software is permitted, provided that it is clearly
stated in the documentation and source comments that the code may
not be used to develop a RAR (WinRAR) compatible archiver.

A6GibKm:
> Extra-data just leads to bad user experience and binds the application to external server instead of Flathub CDN. 

Agree.

As it is `7z` and `unrar` are in the path but `7za` and `7rz` are not being built. Is this Ok?

A6GibKm:
bot, build org.gnome.FileRoller

A6GibKm:
@nedrichards using `copy-icon` the issue you mention is fixed.

org.gnome.FileRoller.json
```diff
@@ -0,0 +1,91 @@
+{
+    "app-id" : "org.gnome.FileRoller",
+    "runtime" : "org.gnome.Platform",
+    "runtime-version" : "3.36",
+    "sdk" : "org.gnome.Sdk",
+    "command" : "file-roller",
+    "rename-icon": "org.gnome.ArchiveManager",
+    "copy-icon": true,
```
bilelmoussaoui, line 8: Hmm would be nice to rename those upstream. Feel free to open an issue regarding that.

A6GibKm: Will do, but I have the impression that it is more involved than that... the app uses the names FileRoller and ArchiveManager interchangeably.

A6GibKm: https://gitlab.gnome.org/GNOME/file-roller/-/issues/98

p1u3o:
> > `unrar` might be less easy as it requires `extra-data` due to license issues
> 
> Unrar license allows redistribution.

The reason I used extra-data in my original manifest was that whilst `unrar` is open source (under restrictive licensing), `rar` itself is not.

IMHO the best path forward is to add an extension point for archive specific tools, perhaps a generic one that other applications can use, that's added to the flatpaks path. File Roller will pick these up.

PR #961, opened 2019-04-16
Add GNOME Screenshot

alex285:
GNOME Screenshot is missing from Silverblue base
so it would be nice if we could get it from Flathub

the Flatpak manifest is identical to upstream
with the addition of the "home" permission,
see PR on Gitlab: https://gitlab.gnome.org/GNOME/gnome-screenshot/merge_requests/13

Also the issue on Gitlab: https://gitlab.gnome.org/GNOME/gnome-screenshot/issues/33

PR #628, opened 2018-09-18
add gnome-screenshot

bilelmoussaoui:
The latest release is 3.30. The one before is 3.26, used that until the 3.30 runtime is ready!

PR #1816, opened 2020-09-09
Add org.gnome.Screenshot

A6GibKm:
Please review the cleanups. I saw that most of them were actually removed by flatpak builder so they might be non-redundant. Using https://gitlab.gnome.org/GNOME/gnome-screenshot/-/blob/master/org.gnome.Screenshot.json as a base.

### DIFF
--- a/org.gnome.Screenshot.json
+++ b/org.gnome.Screenshot.json
@@ -21,10 +21,7 @@
         "/share/gtk-doc",
         "/include",
         "/lib/pkgconfig",
-        "/share/pkgconfig",
-        "/man",
-        "*.la",
-        "*.a"
+        "/man"
     ],
     "modules" : [
         {
@@ -34,11 +31,13 @@
                 "/lib/gnome-settings-daemon-3.0",
                 "/share/doc",
                 "/share/gdm",
-                "/share/gnome"
+                "/share/gnome",
+                "*.la"
             ],
             "config-opts" : [
                 "--disable-alsa",
                 "--disable-null",
+                "--disable-static",
                 "--disable-oss"
             ],
             "sources" : [
--- a/org.gnome.Screenshot.json
+++ b/org.gnome.Screenshot.json
@@ -76,6 +76,10 @@
                     "type": "archive",
                     "url": "https://download.gnome.org/sources/gnome-screenshot/3.36/gnome-screenshot-3.36.0.tar.xz",
                     "sha256": "33495d892707179254b743f8f70c9a82cde5c5f2c7ea3db634a2ba7ea7331266"
+                },
+                {
+                    "type" : "patch",
+                    "path" : "metainfo.patch"
                 }
             ]
         }
--- a/org.gnome.Screenshot.json
+++ b/org.gnome.Screenshot.json
@@ -1,0 +1,83 @@
+{
+    "app-id" : "org.gnome.Screenshot",
+    "runtime" : "org.gnome.Platform",
+    "runtime-version" : "3.36",
+    "sdk" : "org.gnome.Sdk",
+    "command" : "gnome-screenshot",
+    "x-run-args" : [
+        "--interactive"
+    ],
+    "finish-args" : [
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--socket=pulseaudio",
+        "--talk-name=org.gnome.Shell.Screenshot",
+        "--filesystem=home"
+    ],
+    "cleanup" : [
+        "/share/vala",
+        "/share/man",
+        "/share/gtk-doc",
+        "/include",
+        "/lib/pkgconfig",
+        "/share/pkgconfig",
+        "/man",
+        "*.la",
+        "*.a"
+    ],
+    "modules" : [
+        {
+            "name" : "libcanberra-gtk3",
+            "cleanup" : [
+                "/bin",
+                "/lib/gnome-settings-daemon-3.0",
+                "/share/doc",
+                "/share/gdm",
+                "/share/gnome"
+            ],
+            "config-opts" : [
+                "--disable-alsa",
+                "--disable-null",
+                "--disable-oss"
+            ],
+            "sources" : [
+                {
+                    "type" : "archive",
+                    "url" : "http://0pointer.de/lennart/projects/libcanberra/libcanberra-0.30.tar.xz",
+                    "sha256" : "c2b671e67e0c288a69fc33dc1b6f1b534d07882c2aceed37004bf48c601afa72"
+                }
+            ]
+        },
+        {
+            "name" : "libhandy",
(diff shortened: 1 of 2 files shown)


A6GibKm:
bot, build org.gnome.Screenshot

A6GibKm:
Upstream issue: https://gitlab.gnome.org/GNOME/gnome-screenshot/-/issues/33

Exalm:
I'm fine with maintaining it here, but I'm not sure it should be Flatpak-ed to begin with, since it won't work in e.g. KDE/Wayland, and we may in future remove the X11 backend or even move the whole UI to gnome-shell and make the app a dummy launcher for it: https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/2947

A6GibKm:
The patches for the metainfo are already upstream. It might fail due to the size of the image.

Exalm:
I don't think it makes sense to shoe 3.38 screenshot for 3.36: the UI is considerably different and the old screenshot matches better.

That said, I'll do 3.38 release, well, tomorrow.

org.gnome.Screenshot.json
```diff
@@ -0,0 +1,87 @@
...
+                    "type": "archive",
+                    "url": "https://download.gnome.org/sources/libhandy/0.91/libhandy-0.91.0.tar.xz",
+                    "sha256": "cfc1c1cc272d6932ef253fbed07b69a655b33e7c49fb66806beedb7199128c86"
+                }
+            ]
+        },
+        {
+            "name" : "gnome-screenshot",
+            "buildsystem" : "meson",
+            "builddir" : true,
```
Exalm, line 73: You shouldn't need this, it's only needed upstream for Builder.

org.gnome.Screenshot.json
```diff
@@ -0,0 +1,87 @@
...
+                "-Dexamples=false",
+                "-Dglade_catalog=disabled",
+                "-Dintrospection=disabled",
+                "-Dtests=false",
+                "-Dvapi=false"
+            ],
+            "sources" : [
+                {
+                    "type": "archive",
+                    "url": "https://download.gnome.org/sources/libhandy/0.91/libhandy-0.91.0.tar.xz",
```
Exalm, line 65: There's 1.0.

Exalm: Speaking of which, I should totally bump the version.

A6GibKm:
At the moment my priority is to be able to successfully build this, the 3.38-2 screenshot is mostly in line with the current app. Since 3.37 already has a correct metainfo the patch will be dropped anyways. 

issues:
- `appstream-util validate` fails on the new set of screenshot due to small size.
- 3.38 runtime (should) launches in a week. But to wait a week for the release is ok.

I am looking forward to the 3.38 release.

A6GibKm:
bot, build org.gnome.Screenshot

Exalm:
> appstream-util validate fails on the new set of screenshot due to small size. (tested on master's metainfo)

The old one was too small too. The app in general is too small, what can I do 🤷️

A6GibKm:
The validation fails, but the flatpak builds just fine. The app can't be enlarged so it is fine, I just didn't knew if it would build 10 minutes ago.

A6GibKm:
bot, build org.gnome.Screenshot

A6GibKm:
bot, build org.gnome.Screenshot

A6GibKm:
> but I'm not sure it should be Flatpak-ed to begin with

Silverblue users are currently use the flatpak in the fedora repos, this would allow user from other distros. I don't know if there is a better way to handle the cases when the flatpak won't work.

Exalm:
TBH it should just be preinstalled on Silverblue, I don't know why it isn't. It's a core app.

A6GibKm:
tl;dr newer versions install the flatpak version from the fedora repo during installation https://github.com/fedora-silverblue/issue-tracker/issues/34.

A6GibKm:
whoops I missed the two PRs were this is discussed in flathub previously, #961 and #628.

A6GibKm:
Needed to take screenshots in a sane way in Fedora Silverblue, there is a version on the `fedora` flatpak repository but there are many ways in which a flathub version would be superior.

I think the question of *why* is irrelevant as long as there is an interest, the more interesting question is whether it is worth it to have a flatpak given the drawback of it only being supported on GNOME+wayland, specially given that there is no clear way to communicate this to the end user.

I suspect that other desktop environments/distributions come with sane ways of taking screenshots out of the box and the only use-case for this is in GNOME.

alatiera:
> I think the question of _why_ is irrelevant as long as there is an interest

Its not, Screenshot is not an app, but part of the Operating System. Its a core utility of GNOME. If your distro doesn't include it, or provide a replacement utility that's a bug with them and shouldn't really spill here. This looks like packaging something for the shake of packaging here.

A6GibKm:
https://github.com/fedora-silverblue/issue-tracker/issues/34.

Exalm:
To be clear, yes, Screenshot's upstream repo does include a flatpak manifest, but that's a development tool really, it exists for:

1. Builder, so that Run button works
2. CI

E.g. gnome-control-center also has a flatpak for similar purposes, and so do libraries like GTK.

bilelmoussaoui:
I'm going to go ahead and close this one, see the other two merge requests for details. Silverblue should probably ship it as part of the base system and not as a Fltatpak neither.

tpopela:
> Well ... if it is a core utility ... maybe Silverblue should have Gnome Screenshot installed as an RPM inside the vanilla image and not as a Flatpak.

From our point of view, we don't see Cheese as a core utility. It's another application and thus is available as a Flatpak in Silverblue. Shipping it as a Flatpak was approved by David King (the upstream Cheese maintainer). He actually created it as well - https://src.fedoraproject.org/flatpaks/cheese/c/18e64eb8e013f048227bc3d5bce5eb3614a810b1?branch=master

> (Because it is crazy to have an enormous runtime from `fedora` Flatpak repo just for Screenshot.)

It's not just for Screenshot. If you install the Silverblue (starting from Silverblue 32) you have plenty of Flatpaks preinstalled and sharing the same Fedora runtime. If user wants to avoid that he is free to overlay a Cheese RPM.

tpopela:
I was pointed in the Silverblue ticket that I'm talking about Cheese, but the question is about Screenshot. Sorry about that. But it doesn't change anything about my answer.

Exalm:
It does though, since Screenshot doesn't do anything on its own and talks to gnome-shell.

As mentioned, it may even become a dummy launcher for https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/2947 in future.

alatiera:
There is no issue with Silverblue providing screenshots as a flatpak ootb, but we can't really ship screenshots in Flathub.

PR #1818, opened 2020-09-10
Add org.gnome.Boxes.Extension.OsinfoDb

felipeborges:
This is a GNOME Boxes extension that will update the operating system's data that Boxes uses for setting up virtual machines.

### DIFF
--- a/org.gnome.Boxes.Extension.OsinfoDb.json
+++ b/org.gnome.Boxes.Extension.OsinfoDb.json
@@ -9,6 +9,20 @@
     "appstream-compose" : false,
     "modules" : [
         {
+            "name" : "appdata",
+            "buildsystem" : "simple",
+            "build-commands" : [
+                "install -Dm644 --target-directory=${FLATPAK_DEST}/share/metainfo org.gnome.Boxes.Extension.OsinfoDb.metainfo.xml",
+                "appstream-compose --basename=org.gnome.Boxes.Extension.OsinfoDb --prefix=${FLATPAK_DEST} --origin=flatpak org.gnome.Boxes.Extension.OsinfoDb"
+            ],
+            "sources" : [
+                {
+                    "type" : "file",
+                    "path" : "org.gnome.Boxes.Extension.OsinfoDb.metainfo.xml"
+                }
+            ]
+        },
+        {
             "name" : "osinfo-db",
             "buildsystem" : "simple",
             "builddir": true,
--- a/org.gnome.Boxes.Extension.OsinfoDb.json
+++ b/org.gnome.Boxes.Extension.OsinfoDb.json
@@ -1,0 +1,27 @@
+{
+    "id" : "org.gnome.Boxes.Extension.OsinfoDb",
+    "branch" : "master",
+    "runtime" : "org.gnome.Boxes",
+    "runtime-version" : "stable",
+    "sdk" : "org.gnome.Sdk//3.36",
+    "build-extension" : true,
+    "separate-locales" : false,
+    "appstream-compose" : false,
+    "modules" : [
+        {
+            "name" : "osinfo-db",
+            "buildsystem" : "simple",
+            "builddir": true,
+            "build-commands" : [
+                "make",
+                "osinfo-db-import --dir /app/share/osinfo/ osinfo-db-*.tar.xz"
+            ],
+            "sources" : [
+                {
+                    "type" : "git",
+                    "url" : "https://gitlab.com/libosinfo/osinfo-db.git"
+                }
+            ]
+        }
+    ]
+}
--- a/org.gnome.Boxes.Extension.OsinfoDb.json
+++ b/org.gnome.Boxes.Extension.OsinfoDb.json
@@ -1,6 +1,5 @@
 {
     "id" : "org.gnome.Boxes.Extension.OsinfoDb",
-    "branch" : "master",
     "runtime" : "org.gnome.Boxes",
     "runtime-version" : "stable",
     "sdk" : "org.gnome.Sdk//3.38",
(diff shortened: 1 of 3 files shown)


nedrichards:
bot, build org.gnome.Boxes.Extension.OsinfoDb

nedrichards:
@felipeborges this looks like the upstream "master" version rather than something aimed at flathub which would need stable versions and pinned commits etc. you probably need a skip-appstream-checker flathub.json as well.

felipeborges:
@nedrichards osinfo-db is really just a database with VMs preferences, we wanted to have it as up-to-date as possible, that's why I went with "master".

Oh, I can actually add appdata.

felipeborges:
bot, build org.gnome.Boxes.Extension.OsinfoDb

felipeborges:
Now the build fails because we are missing the extension data in org.gnome.Boxes.json.

Will I break the Boxes build if I add the extension metadata to its manifest before we merge this? It works locally.

Also, I updated the extension's manifest to refer to org.gnome.Boxes//stable and use org.gnome.Sdk//3.36. Still, I would like to pull osinfo-db from its master branch.

nedrichards:
Yeah, the source of the bits that you put here can be master - but it should be a pinned commit at least. If this is going to be absolutely required then it might be better being exported out of the existing boxes repo - I'll have a quick think about what will work best to keep this "fresh" and reduce complexity.

felipeborges:
Pinning a commit should be fine. My main goal here is to decouple the OS info updates from the Boxes updates.

Osinfo updates are small but nice to have, especially URL updates when distros go EOL, when vendors move ISOs around breaking our URLs, when vendors release .1 versions, etc...

Besides, we have Boxes stable, beta, unstable and we also plan on having Boxes Flatpaks build from fedora/rhel RPMs. All these versions could consume the osinfo updates from the extension.

barthalion:
> Will I break the Boxes build if I add the extension metadata to its manifest before we merge this? It works locally.

No, there will be just nothing to satisfy that extension point, but it won't cause issues.

felipeborges:
> if this is going to be absolutely required then it might be better being exported out of the existing boxes repo

@nedrichards this extension isn't necessarily required. Boxes will still ship its own osinfo-db, but with the extension we can deliver updates to it.

felipeborges:
Any follow ups on this? I merged https://github.com/flathub/org.gnome.Boxes/pull/53 in Boxes.

Besides, we now have very strict CI gating in osinfo-db upstream.

felipeborges:
bot, build org.gnome.Boxes.Extension.OsinfoDb

barthalion:
bot, build org.gnome.Boxes.Extension.OsinfoDb

org.gnome.Boxes.Extension.OsinfoDb.json
```diff
@@ -0,0 +1,41 @@
+{
+    "id" : "org.gnome.Boxes.Extension.OsinfoDb",
+    "branch" : "master",
+    "runtime" : "org.gnome.Boxes",
+    "runtime-version" : "stable",
+    "sdk" : "org.gnome.Sdk//3.36",
```
barthalion, line 6:
```suggestion
    "sdk" : "org.gnome.Sdk//3.38",
```

org.gnome.Boxes.Extension.OsinfoDb.json
```diff
@@ -0,0 +1,41 @@
+{
+    "id" : "org.gnome.Boxes.Extension.OsinfoDb",
+    "branch" : "master",
```
barthalion, line 3:
```suggestion
    "branch" : "master",
```

org.gnome.Boxes.Extension.OsinfoDb.json
```diff
@@ -0,0 +1,41 @@
...
+            "buildsystem" : "simple",
+            "builddir": true,
+            "build-commands" : [
+                "make",
+                "osinfo-db-import --dir /app/share/osinfo/ osinfo-db-*.tar.xz"
+            ],
+            "sources" : [
+                {
+                    "type" : "git",
+                    "url" : "https://gitlab.com/libosinfo/osinfo-db.git"
```
barthalion, line 36: It should use a specific commit as we want builds to be reproducible.

barthalion (commented):
I'm not really sure if extension definition in Boxes is correct.

```
    "add-extensions" : {
        "org.gnome.Boxes.Extension" : {
            "version" : "master",
            "directory" : "share/osinfo",
            "no-autodownload" : false,
            "autodelete" : true
        }
    },
```

It should specify full extension name given it specifies `share/osinfo` for the mount point and also use `stable` for the version.

org.gnome.Boxes.Extension.OsinfoDb.json
felipeborges: sorry, should this be "stable" too? It is unclear to me whether this refer to the flatpak branching or the osinfo-db git branch name.

felipeborges:
thanks a lot for your review!

org.gnome.Boxes.Extension.OsinfoDb.json
barthalion: This is flatpak branch, Flathub automation sets it automatically based on git branch. Usually there's no need to specify it.

barthalion:
bot, build org.gnome.Boxes.Extension.OsinfoDb

barthalion:
The last missing piece is `skip-icons-check` like here: https://github.com/flathub/org.freedesktop.Sdk.Extension.golang/blob/branch/19.08/flathub.json#L2

felipeborges:
bot, build org.gnome.Boxes.Extension.OsinfoDb

barthalion:
:tada: 

/merge

felipeborges:
Thank you very much!

PR #1822, opened 2020-09-13
Publish raceintospace to flathub

pemensik:


pemensik:
This covers old game Race into Space, as presented on http://www.raceintospace.org/ or https://github.com/raceintospace/raceintospace

org.raceintospace.Raceintospace.yml
```diff
@@ -0,0 +1,89 @@
+# vim: sts=4:
+app-id: org.raceintospace.Raceintospace
+runtime: org.freedesktop.Platform
+runtime-version: '19.08'
```
LeandroStanger, line 4: Update runtime to 20.08
```suggestion
runtime-version: '20.08'
```

LeandroStanger:
bot, build org.raceintospace.Raceintospace

org.raceintospace.Raceintospace.yml
```diff
@@ -0,0 +1,89 @@
...
+        - '/include'
+        - '/lib/lib*.a'
+    # Icon should be in different place build system puts it, workaround
+    # .xpm icons are unsupported
+    - name: icon-rename
+      buildsystem: simple
+      build-commands:
+        - install -D /app/share/pixmaps/raceintospace.png /app/share/icons/hicolor/32x32/apps/org.raceintospace.Raceintospace.png
+        - sed -e 's/^Icon=raceintospace/Icon=org.raceintospace.Raceintospace/' -i /app/share/applications/*.desktop
+finish-args:
```
barthalion, line 82: `finish-args` traditionally is before `modules` section.

org.raceintospace.Raceintospace.yml
```diff
@@ -0,0 +1,89 @@
...
+      buildsystem: cmake
+      sources:
+        - type: git
+          url: https://github.com/raceintospace/raceintospace.git
+      cleanup:
+        - '/include'
+        - '/lib/lib*.a'
+    # Icon should be in different place build system puts it, workaround
+    # .xpm icons are unsupported
+    - name: icon-rename
```
barthalion, line 77: Any reason not to use `post-install` in `raceintospace` module to run these commands?

org.raceintospace.Raceintospace.yml
```diff
@@ -0,0 +1,89 @@
...
+      cleanup:
+        - '/lib/libproto*.a'
+        - '/include'
+    - name: protobuf-c
+      buildsystem: autotools
+      sources:
+        - type: git
+          url: https://github.com/protobuf-c/protobuf-c.git
+          tag: v1.3.3
+    - name: boost
```
barthalion, line 53: Boost build should be limited to specific modules the game needs.

org.raceintospace.Raceintospace.yml
```diff
@@ -0,0 +1,89 @@
...
+app-id: org.raceintospace.Raceintospace
+runtime: org.freedesktop.Platform
+runtime-version: '19.08'
+sdk: org.freedesktop.Sdk
+command: raceintospace
+#rename-icon: raceintospace
+rename-desktop-file: raceintospace.desktop
+rename-appdata-file: raceintospace.appdata.xml
+modules:
+    - name: sdl
```
barthalion, line 11: `sdl` is already there in shared-modules: https://github.com/flathub/shared-modules/tree/master/SDL

org.raceintospace.Raceintospace.yml
```diff
@@ -0,0 +1,89 @@
+# vim: sts=4:
+app-id: org.raceintospace.Raceintospace
+runtime: org.freedesktop.Platform
+runtime-version: '19.08'
+sdk: org.freedesktop.Sdk
+command: raceintospace
+#rename-icon: raceintospace
+rename-desktop-file: raceintospace.desktop
```
barthalion, line 8: These renames should ideally happen upstream.

barthalion (changes requested):

org.raceintospace.Raceintospace.yml
pemensik: It might work. I could not find any documentation related to post-install however. I was just following https://docs.flatpak.org/en/latest/manifests.html when preparing this flatpak. Can you point me to some documentation how to use it?

org.raceintospace.Raceintospace.yml
pemensik: Sounds nice. However, it is possible to include it into Yaml formatted file? If so, how? I haven't found any reference on that repo. It expect json format, which was not mentioned by tutorial I was following. Did I choose wrong format? Is there up-to-date documentation how shared modules could be used?

org.raceintospace.Raceintospace.yml
pemensik: I used --with-headers only. I don't know how to choose any more specific. Is there shared knowledge base for common modules? Any project with boost usage?

org.raceintospace.Raceintospace.yml
pemensik: Okay, found it is simple. It would be nice to have YAML example how to do it somewhere, instead just guessing.

org.raceintospace.Raceintospace.yml
pemensik: Flatpak is not yet primary source of installation. And  it was historically in current paths, I would not change it right away.

org.raceintospace.Raceintospace.yml
pemensik: Okay, it just works. Thanks! But again, I found no information about it in documentation.

org.raceintospace.Raceintospace.yml
pemensik: Oh, it seems install target needs it also. Thanks, fixed.

org.raceintospace.Raceintospace.yml
pemensik: Oh, maybe tutorial should be updated. I just followed whatever was recommended there. Not really can find current platform to choose.

org.raceintospace.Raceintospace.yml
bilelmoussaoui: IIRC both the desktop file & appstream specs suggests using a reverse domain name for the file names see https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#file-naming & https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic.
It's just Flatpak enforces that every exported file should be prefixed by the app-id.

barthalion:
bot, build org.raceintospace.Raceintospace

barthalion:
Sorry, misclicked.

org.raceintospace.Raceintospace.yml
```diff
@@ -0,0 +1,77 @@
...
+rename-appdata-file: raceintospace.appdata.xml
+finish-args:
+    - --share=ipc
+    - --socket=x11
+    - --socket=pulseaudio
+    - --persist=.raceintospace
+    # Until support for SDL2, only X11 will work
+    #- --socket=wayland
+modules:
```
barthalion, line 16: Not critical, but would be nice to follow https://github.com/flathub/flathub/wiki/YAML-Style-Guide. I forgot about it on previous review.

barthalion:
Looks like appdata went missing this time: `stat: cannot stat 'builddir/*/share/app-info/xmls/org.raceintospace.Raceintospace.xml.gz': No such file or directory`

BTW, icon size requirement can be disabled by adding `flathub.json` file with following content: `{ "skip-icons-check": true }`. It won't look good on the website/native frontends but it is what it is.

barthalion:
bot, build org.raceintospace.Raceintospace

pemensik:
Oh, I did not understand where is icon size limited. It is build check I see. Okay, ignored for now.

barthalion:
bot, build org.raceintospace.Raceintospace

org.raceintospace.Raceintospace.yml
pemensik: At the same time, application had raceintospace.desktop id in ancient history. Icon change would not affect installations into /usr/local, but desktop name change would. I am not sure we should move it on upstream side now. I doubt there would be any real benefit in that. I were unable to display my meta file in gnome-software until now. Guess there is something wrong with that, which I cannot find.

barthalion:
LGTM.

@pemensik Who from the upstream team should be invited to the new repository?

org.raceintospace.Raceintospace.yml
bilelmoussaoui: Because any store out there doesn't use the metainfo provided by the application, they are combined in an appstream collection file that's defined per repository (otherwise you will need to download the app before being able to know more details about that app). If you're running a test build or a local build, you won't be able to see any information about the locally built package.

pemensik: I pushed this application to official [Fedora package](https://src.fedoraproject.org/rpms/raceintospace), but even then I were unable to display my metadata. I am not sure what is correct process to add metadata to the distribution. I thought it is just read by gnome-software from /usr/share/metainfo. But what you say actually makes sense, it has to be somewhere to offer nice data before installation.

pemensik:
I think just @pemensik, @peyre would be enough for start. Can it be modified later by us?

pemensik:
We have also organizationhttps://github.com/orgs/raceintospace/people  here on github. Would it be possible to give access to whole group?

peyre:
Sorry guys, I've been watching the discussion, but I haven't responded because this is beyond my level of knowledge.  So don't take my silence as disapproval.

barthalion:
> We have also [organization](https://github.com/orgs/raceintospace/people) here on github. Would it be possible to give access to whole group? I am also member of it. Can it be just @raceintospace?

Not currently but it's on my list.

PR #1825, opened 2020-09-13
Add com.gitlab.ColinDuquesnoy.MellowPlayer

ColinDuquesnoy:
Add [MellowPlayer](https://gitlab.com/ColinDuquesnoy/MellowPlayer) to flathub 

*Note about permissions*

We use ``--socket=session-bus`` and ``--socket=system-bus`` in order to get mpris to work. Let me know if there is a more "secure" solution.

### DIFF
--- a/com.gitlab.ColinDuquesnoy.MellowPlayer.yml
+++ b/com.gitlab.ColinDuquesnoy.MellowPlayer.yml
@@ -7,7 +7,7 @@ sdk: org.kde.Sdk
 command: MellowPlayer
 finish-args:
   - --share=ipc
-  - --socket=x11
+  - --socket=fallback-x11
   - --socket=wayland
   - --share=network
   - --socket=pulseaudio
@@ -34,4 +34,3 @@ modules:
         url: https://gitlab.com/ColinDuquesnoy/MellowPlayer
         tag: "3.6.6"
         commit: "a55f920d5b0f73ac9408a5c26efc4851b45507b0"
-
--- a/com.gitlab.ColinDuquesnoy.MellowPlayer.yml
+++ b/com.gitlab.ColinDuquesnoy.MellowPlayer.yml
@@ -14,6 +14,10 @@ finish-args:
   - --device=dri
   - --talk-name=org.mpris.MediaPlayer2.Player
   - --own-name=org.mpris.MediaPlayer2.MellowPlayer3.*
+  - --env=TMPDIR=/var/rpk
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --own-name=org.kde.*
+  - --talk-name=org.freedesktop.Notifications
 rename-desktop-file: "mellowplayer.desktop"
 rename-icon: "mellowplayer"
 modules:
@@ -31,5 +35,5 @@ modules:
     sources:
       - type: git
         url: https://gitlab.com/ColinDuquesnoy/MellowPlayer
-        tag: "3.6.6-flathub-fixes2"
-        commit: "0a1d55110861ac4f0eb7f86b39ae4f0d89a313b9"
+        tag: "3.6.6-flathub-fixes3"
+        commit: "399cea8ba8828d57a1ea1a32458cf21d02517bb0"
--- a/com.gitlab.ColinDuquesnoy.MellowPlayer.yml
+++ b/com.gitlab.ColinDuquesnoy.MellowPlayer.yml
@@ -12,9 +12,8 @@ finish-args:
   - --share=network
   - --socket=pulseaudio
   - --device=dri
-  - --socket=session-bus
-  - --socket=system-bus
-  - --filesystem=/tmp
+  - --talk-name=org.mpris.MediaPlayer2.Player
+  - --own-name=org.mpris.MediaPlayer2.MellowPlayer3.*
 rename-desktop-file: "mellowplayer.desktop"
 rename-icon: "mellowplayer"
 modules:
@@ -32,5 +31,5 @@ modules:
     sources:
       - type: git
         url: https://gitlab.com/ColinDuquesnoy/MellowPlayer
-        tag: "3.6.6"
-        commit: "a55f920d5b0f73ac9408a5c26efc4851b45507b0"
+        tag: "3.6.6-flathub-fixes"
+        commit: "2823aa5bfa42fcc23ec9a0959f0ec43ac7bfe49f"
--- a/com.gitlab.ColinDuquesnoy.MellowPlayer.yml
+++ b/com.gitlab.ColinDuquesnoy.MellowPlayer.yml
@@ -32,4 +32,4 @@ modules:
       - type: git
         url: https://gitlab.com/ColinDuquesnoy/MellowPlayer
         tag: "3.6.6-flathub-fixes2"
-        commit: "2823aa5bfa42fcc23ec9a0959f0ec43ac7bfe49f"
+        commit: "0a1d55110861ac4f0eb7f86b39ae4f0d89a313b9"
--- a/com.gitlab.ColinDuquesnoy.MellowPlayer.yml
+++ b/com.gitlab.ColinDuquesnoy.MellowPlayer.yml
@@ -31,5 +31,5 @@ modules:
     sources:
       - type: git
         url: https://gitlab.com/ColinDuquesnoy/MellowPlayer
-        tag: "3.6.6-flathub-fixes"
+        tag: "3.6.6-flathub-fixes2"
         commit: "2823aa5bfa42fcc23ec9a0959f0ec43ac7bfe49f"
--- a/com.gitlab.ColinDuquesnoy.MellowPlayer.yml
+++ b/com.gitlab.ColinDuquesnoy.MellowPlayer.yml
@@ -1,0 +1,37 @@
+app-id: com.gitlab.ColinDuquesnoy.MellowPlayer
+runtime: org.kde.Platform
+runtime-version:  "5.15"
+base: io.qt.qtwebengine.BaseApp
+base-version: "5.15"
+sdk: org.kde.Sdk
+command: MellowPlayer
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --socket=wayland
+  - --share=network
+  - --socket=pulseaudio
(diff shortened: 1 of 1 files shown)


com.gitlab.ColinDuquesnoy.MellowPlayer.yml
```diff
@@ -0,0 +1,37 @@
...
+sdk: org.kde.Sdk
+command: MellowPlayer
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --socket=wayland
+  - --share=network
+  - --socket=pulseaudio
+  - --device=dri
+  - --socket=session-bus
```
barthalion, line 15: Something like this should work for mpris:

```
    "--talk-name=org.mpris.MediaPlayer2.Player",
    "--own-name=org.mpris.MediaPlayer2.MellowPlayer.*"
```

com.gitlab.ColinDuquesnoy.MellowPlayer.yml
```diff
@@ -0,0 +1,37 @@
+app-id: com.gitlab.ColinDuquesnoy.MellowPlayer
+runtime: org.kde.Platform
+runtime-version:  "5.15"
+base: io.qt.qtwebengine.BaseApp
+base-version: "5.15"
+sdk: org.kde.Sdk
+command: MellowPlayer
+finish-args:
+  - --share=ipc
+  - --socket=x11
```
barthalion, line 10:
```suggestion
  - --socket=fallback-x11
```

com.gitlab.ColinDuquesnoy.MellowPlayer.yml
```diff
@@ -0,0 +1,37 @@
...
+  - --socket=wayland
+  - --share=network
+  - --socket=pulseaudio
+  - --device=dri
+  - --socket=session-bus
+  - --socket=system-bus
+  - --filesystem=/tmp
```
barthalion, line 17: Does it need to share `/tmp` with the host?

barthalion:
bot, build com.gitlab.ColinDuquesnoy.MellowPlayer

barthalion:
Build failed due to missing metainfo file, if it's part of the code, it might be installed in the wrong place or under wrong name.

bilelmoussaoui:
> Build failed due to missing metainfo file, if it's part of the code, it might be installed in the wrong place or under wrong name.

Seems so
https://gitlab.com/ColinDuquesnoy/MellowPlayer/-/blob/master/src/main/share/metainfo/mellowplayer.appdata.xml

The metainfo (or any exported file) should be prefixed by the app-id. So it should be `com.gitlab.ColinDuquesnoy.MellowPlayer.metainfo.xml` 
The metainfo seems to have some missing tags that are required for Flathub. See https://github.com/flathub/flathub/wiki/AppData-Guidelines for details.

com.gitlab.ColinDuquesnoy.MellowPlayer.yml
ColinDuquesnoy: Thank you I'll try that

com.gitlab.ColinDuquesnoy.MellowPlayer.yml
ColinDuquesnoy: I'll check that

ColinDuquesnoy:
> > Build failed due to missing metainfo file, if it's part of the code, it might be installed in the wrong place or under wrong name.
> 
> Seems so
> https://gitlab.com/ColinDuquesnoy/MellowPlayer/-/blob/master/src/main/share/metainfo/mellowplayer.appdata.xml
> 
> The metainfo (or any exported file) should be prefixed by the app-id. So it should be `com.gitlab.ColinDuquesnoy.MellowPlayer.metainfo.xml`
> The metainfo seems to have some missing tags that are required for Flathub. See https://github.com/flathub/flathub/wiki/AppData-Guidelines for details.

I'll fix that upstream. Thank you

com.gitlab.ColinDuquesnoy.MellowPlayer.yml
ColinDuquesnoy: That work fine.

com.gitlab.ColinDuquesnoy.MellowPlayer.yml
ColinDuquesnoy: I've removed --filesystem=/tmp

ColinDuquesnoy:
bot, build com.gitlab.ColinDuquesnoy.MellowPlayer

ColinDuquesnoy:
bot, build com.gitlab.ColinDuquesnoy.MellowPlayer

ColinDuquesnoy:
bot, build com.gitlab.ColinDuquesnoy.MellowPlayer

barthalion:
It started only once for me, I had to delete `/var/home/bpiotrowski/.var/app/com.gitlab.ColinDuquesnoy.MellowPlayer/data/MellowPlayer/MellowPlayer3/single-instance.lock` to make it work again.

Defaulting to hiding in tray may also be misleading as GNOME by default has no tray icons at all.

ColinDuquesnoy:
> It started only once for me, I had to delete /var/home/bpiotrowski/.var/app/com.gitlab.ColinDuquesnoy.MellowPlayer/data/MellowPlayer/MellowPlayer3/single-instance.lock to make it work again.

How did you quit the application the first time? That shouldn't happen if you exit the application by using Menu -> Quit (or Ctrl+Q)...

> Defaulting to hiding in tray may also be misleading as GNOME by default has no tray icons at all.

There is a setting to disable that behaviour (F2->General -> Close application to system tray) for those who prefer the gnome behaviour. It's a sane default on all other desktops.

Does those two "problems" prevent from publishing the app? To me these looks like bugs that may/should be reported upstream and that are not related to flatpak.

barthalion:
> How did you quit the application the first time? That shouldn't happen if you exit the application by using Menu -> Quit (or Ctrl+Q)...

I closed the window via equivalent of Alt+F4.

> There is a setting to disable that behaviour (F2->General -> Close application to system tray) for those who prefer the gnome behaviour. It's a sane default on all other desktops.

Being hidden in tray – maybe, being locked out of the application completely, not really. I assume all other applications like Zoom use D-Bus to query if application is already running. (Bilal, please correct me if I'm wrong!)

> Does those two "problems" prevent from publishing the app?

Not a blockers certainly but it might be confusing to regular users.

ColinDuquesnoy:
> being locked out of the application completely, not really

I agree.

MellowPlayer is a single instance application (implemented with QLocalServer & QLocalSocket instead of dbus so that it also works on Windows). Starting a second instance of the application should bring back the first instance (even if closed with Alt+F4). 

I've checked this mechanism is still working: it works fine with native binary (except if you kill the process or use Ctrl+C but that is another issue) but it fails when using the flatpak. So in the end I think this is a bug that should be fixed before publishing.

ColinDuquesnoy:
@barthalion The bug occurs because I removed ``--filesystem=/tmp`` from the flatpak manifest.

I forgot QLocalServer & QLocalSocket use ``/tmp/APP_Name`` for the local socket file. Do you think it is problematic to add ``--filesystem=/tmp`` back?

barthalion:
Can you try with `--env=TMPDIR=/var/tmp` instead?

gombosg:
#off
MellowPlayer! Lovely! I use this daily for Deezer. :slightly_smiling_face:

ColinDuquesnoy:
> Can you try with `--env=TMPDIR=/var/tmp` instead?

I tried to add it to the flatpak yaml manifest, doesn't work. Any idea?

barthalion:
I'll try to poke around tommorow(ish) and see. You might want to take a look at https://github.com/flathub/com.nextcloud.desktopclient.nextcloud (source code specifically, not the manifest) as I imagine it faces similar problems with locking etc..

ColinDuquesnoy:
I did a bit more testing, ``--env=TMPDIR=/var/tmp`` actually works on recent distributions (ubuntu) but not on CentOS 7. Strangely, on ubuntu, the file is not created (ls /var/tmp/MellowPlayer3 fails) but communication over the local socket seems to work.

I've also found a way to create the socket file in the home directory (e.g ``.var/app/com.gitlab.ColinDuquesnoy.MellowPlayer/data/MellowPlayer/MellowPlayer3/single-instance.lock``), there the socket file is correctly created and communication also works fine. I will try it on CentOS 7 tomorrow. If that works I'll use that solution.

Another problem I've encountered is that Qt is unable to detect a stale lock file (e.g. after a crash) because the application PID is still the same (I've learned pid are namespaced in flatpak so this "normal" behavior, although a bit surprising) so I may need to find another way to detect that the application is already running (maybe using dbus)

barthalion:
> I did a bit more testing, --env=TMPDIR=/var/tmp actually works on recent distributions (ubuntu) but not on CentOS 7. Strangely, on ubuntu, the file is not created (ls /var/tmp/MellowPlayer3 fails) but communication over the local socket seems to work.

It was fixed in flatpak 1.6.2 IIRC. If bubblewrap is SUID, glibc drops some environment variables such as TMPDIR; the work around would be to have a wrapper script that would set TMPDIR and `exec` the main binary instead.

> Another problem I've encountered is that Qt is unable to detect a stale lock file (e.g. after a crash) because the application PID is still the same (I've learned pid are namespaced in flatpak so this "normal" behavior, although a bit surprising) so I may need to find another way to detect that the application is already running (maybe using dbus)

Yeah, that's correct. If you are using Qt to set up tray icon, you also want to add

```
"--talk-name=org.kde.StatusNotifierWatcher",
"--own-name=org.kde.*"
```
…to finish-args.

ColinDuquesnoy:
bot, build com.gitlab.ColinDuquesnoy.MellowPlayer

ColinDuquesnoy:
@barthalion 

I've fixed the single instance applications issues: I fixed QLocalSocket/QLocalServer socket file path and used a native lock file instead of QLockFile (looking at the source code of nextcloud desktop was helpful, thanks for that :+1:)

ColinDuquesnoy:
I also added ``--talk-name=org.freedesktop.Notifications`` for libnotify

PR #1826, opened 2020-09-13
Add Org.qgis.qgis-ltr

hristo-mavrodiev:
The Org.qgis.qgis contains the latest version of QGIS. This version is not a "production" environment for GIS projects. Purpose of this repo is to add the latest stable LTR version of QGIS. For example in Arch linux there are qgis and qgis-ltr packages - https://www.archlinux.org/packages/community/x86_64/qgis/  and https://aur.archlinux.org/packages/qgis-ltr/?O=20&PP=10

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,15 @@
+# Flatpak QGIS packaging
+
+Just flatpak packaging for QGIS.
+
+## I need more python modules!
+Often you will need extra python modules to run extra plugins. We can't possibly package them all with
+the application. To get them in your system, you can install them locally as follows:
+
+```
+flatpak run --command=pip3 org.qgis.qgis install scipy --user
+```
+
+Where `scipy` is the package you want, replace it with whatever you think it might be necessary.
+
+If you feel like it definitely should be bundled, don't hesitate to report an issue in this repository.
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
+## The LTR version of QGIS copied from https://github.com/flathub/org.qgis.qgis 
+The purpose of this repo is to maintain the LTR version of QGIS.
+
 # Flatpak QGIS packaging
 
 Just flatpak packaging for QGIS.
--- a/README.md
+++ b/README.md
@@ -1,4 +1,3 @@
-## The LTR version of QGIS copied from https://github.com/flathub/org.qgis.qgis 
 The purpose of this repo is to maintain the LTR version of QGIS.
 
 # Flatpak QGIS packaging
--- a/flathub.json
+++ b/flathub.json
@@ -1,0 +1,4 @@
+{
+    "only-arches": ["aarch64", "x86_64"],
+    "skip-appstream-check": true
+}
--- a/flathub.json
+++ b/flathub.json
@@ -1,4 +1,3 @@
 {
-    "only-arches": ["aarch64", "x86_64"],
-    "skip-appstream-check": true
+    "only-arches": ["aarch64", "x86_64"]
 }
--- a/org.qgis.qgis-ltr.json
+++ b/org.qgis.qgis-ltr.json
@@ -1,0 +1,614 @@
+{
+    "id" : "org.qgis.qgis-ltr",
+    "branch" : "stable",
+    "base": "io.qt.qtwebkit.BaseApp",
+    "base-version": "5.12",
+    "runtime" : "org.kde.Platform",
+    "runtime-version" : "5.12",
+    "sdk" : "org.kde.Sdk",
+    "command" : "qgis",
+    "rename-icon": "qgis",
+    "finish-args" : [
+        "--env=LD_LIBRARY_PATH=/app/lib:/app/grass76/lib",
+        "--share=ipc",
+        "--share=network",
+        "--socket=x11",
+        "--filesystem=host",
+        "--device=dri"
+    ],
+    "separate-locales" : false,
+    "build-options" : {
+        "env" : {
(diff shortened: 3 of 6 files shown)


README.md
```diff
@@ -0,0 +1,18 @@
+## The LTR version of QGIS copied from https://github.com/flathub/org.qgis.qgis
```
barthalion, line 1: There's no point in having it here. No user is going to ever read it.

flathub.json
```diff
@@ -0,0 +1,4 @@
+{
+    "only-arches": ["aarch64", "x86_64"],
+    "skip-appstream-check": true
```
barthalion, line 3: Appdata should be fixed instead of disabling the check, and if it doesn't already, it should mention it's an LTR version.

barthalion (changes requested):

barthalion:
How common is it to prefer LTR version over the latest one? Also for the record, Arch ships only latest QGIS. AUR is free-for-all thing.

cc @aleixpol

barthalion:
bot, build org.qgis.qgis-ltr

hristo-mavrodiev:
bot, build org.qgis.qgis-ltr

hristo-mavrodiev:
bot, build org.qgis.qgis-ltr

aleixpol:
> How common is it to prefer LTR version over the latest one? Also for the record, Arch ships only latest QGIS. AUR is free-for-all thing.
> 
> cc @aleixpol

I was planning to look into it as well. My plan was to have it in a branch for org.qgis.qgis rather than a separate repo. If someone else to maintain it, that works too.

I asked QGIS maintainers and they recommended to prefer this version to the latest. I'm guessing it's because of the extended use of the python scripting this system does.

I can commit to keep this LTR (or a branch, I actually asked in flatpak@ mailing list about this already) up to date if you want together with the stable version if someone wants.

It feels a bit odd that as is here, they will be listed as separate applications.

org.qgis.qgis-ltr.json
```diff
@@ -0,0 +1,614 @@
+{
+    "id" : "org.qgis.qgis-ltr",
+    "branch" : "stable",
+    "base": "io.qt.qtwebkit.BaseApp",
+    "base-version": "5.12",
+    "runtime" : "org.kde.Platform",
+    "runtime-version" : "5.12",
```
LeandroStanger, line 7: Update runtime to 5.15
```suggestion
    "base-version": "5.15",
    "runtime" : "org.kde.Platform",
    "runtime-version" : "5.15",
```

barthalion:
> I was planning to look into it as well. My plan was to have it in a branch for org.qgis.qgis rather than a separate repo. If someone else to maintain it, that works too.

It depends what you want to achieve. If LTR is pushed as a separate branch of existing QGIS, only one will be visible in the system application list. The other can be still launched from CLI with `flatpak run --branch=foobar` or made default via `flatpak make-current`.

This also means that LTR and the latest releases will share configuration directories. Unless they're already versioned by the app, it may (or not! I don't use it) cause some issues if they are not compatible.

In such situations separate app id is IMHO preferred but I leave the decision to you.

hristo-mavrodiev:
Sometimes the QGIS config is messed up (even when not switching versions) - QGIS fails to start, some bugs appear also.By default QGIS configs are stored in "/home/USER/.local/share/QGIS/QGIS3/"
The configuration files when using flatpak are store in   "$HOME/.var/app/org.qgis.qgis/data/QGIS/QGIS3"
Just deleting "org.qgis.qgis" folder will solve the issue and QGIS creates new configuration files. It's like a fresh new install.
The proposed build for LTR version fails, because it's now searching for qgis-ltr executable(desktop icon) instead of qgis. I have no idea how to solve this.
As a GIS user what I think that other people will love to see is to have the ability to choose a specific version for QGIS, like they have this choice on WinOS or MacOS. Also to have the same version of the important GIS dependencies (Qt, GDAL/OGR, GEOS, SQLite, PostgreSQL Client Version, QWT Version, PROJ, SpatiaLite Version, QScintilla2 Version, GRASS, SAGA,OracleDB driver, MSSQL DB driver) between the existing windows built and flatpak. So no matter you are using Linux or Windows you will have same result for this QGIS version.  It will be great if we could provide same experience as choosing an org.kde.sdk for example:
kde sdk
I will try to help you, but I don't have the knowledge to fully maintain the LTR version. Probably to stay with org.qgis.qgis will be better...

eszlari:
I would try to avoid dashes in the App ID, since only the last part of the domain can contain one. If you would add extensions later, you would have using some like `org.qgis.qgis_ltr.Extension`.

I would use `org.qgis.qgis.LTR` as an ID.

aleixpol:
I created an lts branch of it, I'd say we should go with that and close this PR

eszlari:
> The proposed build for LTR version fails, because it's now searching for qgis-ltr executable(desktop icon) instead of qgis. I have no idea how to solve this.

You need to add `rename-desktop-file` (and `rename-appdata-file`) to the manifest to adjust the ID.

eszlari:
> I created an lts branch of it, I'd say we should go with that and close this PR

Depends on if @hristo-mavrodiev wants to use both version in parallel easily.

Also: AFAIK this branch is not exposed in gnome-software, plasma-discover or flathub.org at the moment.

aleixpol:
Discover should, and I don't see why other clients wouldn't.

barthalion:
Aleix decided to use a branch so we're going to stick with it.

PR #1827, opened 2020-09-14
Add com.github.kmwallio.thiefmd

kmwallio:
This is a markdown editor designed to organize and compile multiple folders into a single html/pdf/markdown file.

### DIFF
--- a/com.github.kmwallio.thiefmd.json
+++ b/com.github.kmwallio.thiefmd.json
@@ -30,15 +30,6 @@
   ],
   "modules":[
     {
-      "name": "libgee",
-      "make-install-args": ["girdir=/app/share/gir-1.0", "typelibdir=/app/lib/girepository-1.0"],
-      "sources": [{
-        "type": "archive",
-        "url": "https://download.gnome.org/sources/libgee/0.20/libgee-0.20.1.tar.xz",
-        "sha256": "bb2802d29a518e8c6d2992884691f06ccfcc25792a5686178575c7111fea4630"
-      }]
-    },
-    {
       "name": "granite",
       "buildsystem": "meson",
       "cleanup": ["/bin", "/share/applications"],
--- a/com.github.kmwallio.thiefmd.json
+++ b/com.github.kmwallio.thiefmd.json
@@ -1,0 +1,106 @@
+{
+  "app-id":"com.github.kmwallio.thiefmd",
+  "runtime":"org.gnome.Platform",
+  "runtime-version":"3.36",
+  "sdk":"org.gnome.Sdk",
+  "command":"com.github.kmwallio.thiefmd",
+  "copy-icon": true,
+  "finish-args":[
+    "--share=ipc",
+    "--share=network",
+    "--filesystem=home",
+    "--device=dri",
+    "--socket=x11"
+  ],
+  "cleanup":[
+    "/cache",
+    "/man",
+    "/share/aclocal",
+    "/share/devhelp",
+    "/lib/systemd",
+    "/include",
+    "/lib/pkgconfig",
+    "/lib/libvala*",
+    "/share/gtk-doc",
+    "/share/man",
+    "/share/vala",
+    "/bin/vapi*",
+    "*.a",
+    "*.la"
+  ],
+  "modules":[
+    {
+      "name": "libgee",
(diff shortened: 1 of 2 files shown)


com.github.kmwallio.thiefmd.json
```diff
@@ -0,0 +1,106 @@
...
+  "modules":[
+    {
+      "name": "libgee",
+      "make-install-args": ["girdir=/app/share/gir-1.0", "typelibdir=/app/lib/girepository-1.0"],
+      "sources": [{
+        "type": "archive",
+        "url": "https://download.gnome.org/sources/libgee/0.20/libgee-0.20.1.tar.xz",
+        "sha256": "bb2802d29a518e8c6d2992884691f06ccfcc25792a5686178575c7111fea4630"
+      }]
+    },
```
bilelmoussaoui, line 40:
```suggestion
```
Part of the GNOME SDK

com.github.kmwallio.thiefmd.json
```diff
@@ -0,0 +1,106 @@
...
+    {
+      "name": "granite",
+      "buildsystem": "meson",
+      "cleanup": ["/bin", "/share/applications"],
+      "sources": [{
+        "type": "archive",
+        "url": "https://github.com/elementary/granite/archive/5.4.0.tar.gz",
+        "sha256": "8194031cac3b87d84a3fab9c30270485ce73f8b8ec23f26c9152b6859c8a18fd"
+      }]
+    },
```
bilelmoussaoui, line 50: Any reason for not using the elementary BaseApp?

com.github.kmwallio.thiefmd.json
```diff
@@ -0,0 +1,106 @@
...
+  "runtime-version":"3.36",
+  "sdk":"org.gnome.Sdk",
+  "command":"com.github.kmwallio.thiefmd",
+  "copy-icon": true,
+  "finish-args":[
+    "--share=ipc",
+    "--share=network",
+    "--filesystem=home",
+    "--device=dri",
+    "--socket=x11"
```
bilelmoussaoui, line 13:
```suggestion
    "--socket=fallback-x11",
    "--socket=wayland"
```

com.github.kmwallio.thiefmd.json
```diff
@@ -0,0 +1,106 @@
+{
+  "app-id":"com.github.kmwallio.thiefmd",
+  "runtime":"org.gnome.Platform",
+  "runtime-version":"3.36",
+  "sdk":"org.gnome.Sdk",
+  "command":"com.github.kmwallio.thiefmd",
+  "copy-icon": true,
```
bilelmoussaoui, line 7:
```suggestion
```
I don't think you need this

com.github.kmwallio.thiefmd.json
```diff
@@ -0,0 +1,106 @@
+{
+  "app-id":"com.github.kmwallio.thiefmd",
+  "runtime":"org.gnome.Platform",
```
bilelmoussaoui, line 3: You could use the freedesktop runtime/sdk along with the elementary base app instead of the gnome runtime if all you want is vala, granite & libgee.

com.github.kmwallio.thiefmd.json
kmwallio: The vala compiler is a bit older in the elementary BaseApp. There's changes between 0.40.11 and 0.48.10 that seemed to prevent compilation.

com.github.kmwallio.thiefmd.json
kmwallio: The vala compiler in the elementary base app is a bit old.

If I switch to freedesktop with elementary base app, and then compile vala, does it take up less space/less resources needed for users?

nedrichards:
bot, build com.github.kmwallio.thiefmd

com.github.kmwallio.thiefmd.json
nedrichards: There'll be less disk space used in the case that the user only has one flatpak installed (yours). It's difficult to generalise across all scenarios but if you've tested and have specific requirements for specific libraries then using the smallest runtime is probably best. It's not a blocker though.

kmwallio:
Seems to be running fine on a clean install of PopOS.

Screenshot

PR #1837, opened 2020-09-19
Add Discord Canary (for beta channel

refi64:
**This is for the beta branch.**

Based on the latest 20.08 sdk and Zypak beta with the experimental file chooser overrides. Hooked up to the external data checker so if I'm unable to update it, everyone doesn't get locked out of Discord for a crazy amount of time :sweat_smile:


refi64:
(and let's pretend I didn't forgot to push half the files)

A6GibKm:
bot, build com.discordapp.DiscordCanary

A6GibKm:
While I think it is fine as it is, it might be debatable whether is better to create a new app instead of just using the beta branch for com.discordapp.Discord.

refi64:
> While I think it is fine as it is, it might be debatable whether is better to create a new app instead of just using the beta branch for com.discordapp.Discord.

Well I think the PTB builds would be a more direct analogue for a beta, whereas Canary is more in the style of nightly builds (there are often periods where there is a build every 4 days, leading me to think that they're automated). This is sort of shown by the PTB using the same icon as stable, whereas canary uses a different icon.

refi64:
Any updates?

PR #1844, opened 2020-09-20
Add com.fightcade.Fightcade

Pobega:
This is a frontend for playing arcade games online using rollback-based GGPO netcode.

The frontend is an electron application that joins you to chat rooms and calls out to `fcade://` URLs to launch emulators, which I've wrapped in a secondary script/desktop file.

The application is proprietary, though I spoke to the devs and got permission to redistribute. ~~I'm still using extra-data due to needing to create some symlinks in `apply_extra` that would just be kind of cumbersome to do in the manifest, but if anyone here feels strongly about switching this to an archive we can certainly do that.~~

The application is meant to live in a directory where it self updates so a lot of the paths it writes to are relative paths, which is why there is a need for so many symlinks out to /var/data/ and persistent home paths.

I've tested that this works with a friend also using the Flatpak.


Pobega:
bot, build com.fightcade.Fightcade

Pobega:
I'm a bit confused what it's failing on here? Could someone point me in the right direction
```
zgrep "<id>com.fightcade.Fightcade\(\.\w\+\)*\(.desktop\)\?</id>" builddir/*/share/app-info/xmls/com.fightcade.Fightcade.xml.gz
 in dir /srv/buildbot/worker/build-x86_64-5/build (timeout 1200 secs)
 watching logfiles {}
 argv: b'zgrep "<id>com.fightcade.Fightcade\\(\\.\\w\\+\\)*\\(.desktop\\)\\?</id>" builddir/*/share/app-info/xmls/com.fightcade.Fightcade.xml.gz'
 using PTY: False
```

Pobega:
bot, build com.fightcade.Fightcade

com.fightcade.Fightcade.desktop
```diff
@@ -0,0 +1,12 @@
+[Desktop Entry]
```
Pobega, line 1: The reason this isn't upstream is because the application isn't _installed_ exactly, they provide a desktop file via a shell script that echos it into `~/.local/share/applications`. Same for the other desktop file.

I am just providing desktop files for my wrappers.

metainfo.xml
```diff
@@ -0,0 +1,59 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Pobega, line 1: Same as the desktop file, this isn't upstreamed because the application is just an archive that you extract anywhere and run from there. It's not meant to be an application that's installed to your system, so we need to provide these.

com.fightcade.Fightcade.desktop
Pobega, line 1: There is a desktop file provided by the upstream if it's preferable that I use that.

```
#!/usr/bin/env xdg-open
[Desktop Entry]
Version=1.0
Type=Application
Terminal=false
Exec=sh -e -c "exec \\"\\$(dirname \\"\\$0\\")/Fightcade2.sh\\"" %k
Name=Fightcade
Comment=Fightcade
Categories=Game;Emulator;ArcadeGame
Icon=
```

fcade-quark.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
```
Pobega, line 1: This desktop file isn't provided properly by upstream, so I've made my own. Their `fcade.sh` wrapper script actually builds it inside the script but that is not viable for us since their wrapper does some other checks (specifically for `/usr/bin/wine`) that would cause us to fail out. I've contacted upstream about respecting the `PATH` and it's on their ToDo list, but for now this works.

Pobega: Snipper from upstream's `fcade.sh`:

```
if [ -x /usr/bin/xdg-mime ]; then
        # register fcade:// url handler
        mkdir -p ~/.local/share/applications/
        echo "[Desktop Entry]
Type=Application
Encoding=UTF-8
Name=FightCade Replay
Exec=${THIS_SCRIPT_DIR}/fcade.sh %U
Terminal=false
MimeType=x-scheme-handler/fcade
" > ~/.local/share/applications/fcade-quark.desktop
        xdg-mime default fcade-quark.desktop x-scheme-handler/fcade
fi
if [ -x /usr/bin/gconftool-2 ]; then
        gconftool-2 -t string -s /desktop/gnome/url-handlers/fcade/command "${THIS_SCRIPT_DIR}/fcade.sh %s"
        gconftool-2 -s /desktop/gnome/url-handlers/fcade/needs_terminal false -t bool
        gconftool-2 -t bool -s /desktop/gnome/url-handlers/fcade/enabled true
fi
```

scripts/fcade-quark.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
```
Pobega, line 1: The reason I did this rather than symlinking to `/app/extra/Fightcade/emulator/fcade` is because I'm not sure if there will be calls I need to stub out. Originally I wrote an early exit for `fcade://autoupdate` but that proved to not be an issue since it's read-only anyway, but I'd prefer to keep this here in case future calls to `fcade` need some middle-man translation.

Pobega:
I've got a commit [in my Flatpak repo](https://github.com/Pobega/com.fightcade.Fightcade/commit/3f397d64cd17551b90a860db4d7683e73d13f0d9) that switches from `extra-data` to `archive`, I assume this is preferable here?

Pobega:
bot, build com.fightcade.Fightcade

Pobega:
bot, build com.fightcade.Fightcade

barthalion:
Is there anyone from upstream team who should be added to the new repository?

poliva:
Hi, I'm Pau Oliva (pof), maintainer of the upstream Linux build from the Fightcade team. You can add me to the repository if you want.

com.fightcade.Fightcade.desktop
```diff
@@ -0,0 +1,12 @@
...
+Type=Application
+Terminal=false
+Exec=fightcade
+Name=Fightcade
+Comment=Play arcade games online
+Categories=Game;Emulator;ArcadeGame
+Icon=com.fightcade.Fightcade
+X-Desktop-File-Install-Version=0.26
+X-Flatpak-Tags=proprietary;
+X-Flatpak=com.fightcade.Fightcade
```
bilelmoussaoui, line 12:
```suggestion
```

nedrichards:
bot, build com.fightcade.Fightcade

PR #1852, opened 2020-09-24
Add com.brave.Browser

analotia:


analotia:
bot, build com.brave.Brave

com.brave.Brave.yaml
```diff
@@ -0,0 +1,80 @@
+app-id: com.brave.Brave
+base: org.electronjs.Electron2.BaseApp
+base-version: '19.08'
```
x80486, line 3: I would give it a shot with `20.08` and see what happens :thinking:

analotia: I honestly don't know for sure if this is a good idea. I wanted to do that but https://github.com/refi64/zypak says I need to use 19.08

From https://github.com/refi64/zypak :
> This requires your Flatpak to be using:
> 
>     org.freedesktop.Platform / Sdk version 19.08.
>     org.electronjs.Electron2.BaseApp as your base app. Recent releases include Zypak built-in.
> 
> Now, instead of running your Electron binary directly, call it via zypak-wrapper PATH/TO/MY/ELECTRON/BINARY.

analotia: Can you check if it works properly with 20.08?

analotia: 20.08 is still building. I will push if it works

x80486: I think it should. Version `20.08` was updated recently and that's what I'm using in Wire. Let me try.

com.brave.Brave.yaml
```diff
@@ -0,0 +1,80 @@
...
+      # For extra-data to be able to extract the .deb
+      - 'install -Dm 755 /{usr,app}/bin/ar'
+      - 'cp /usr/lib/$(gcc --print-multiarch)/libbfd-*.so /app/lib'
+    sources:
+      - type: extra-data
+        # From https://brave-browser-apt-release.s3.brave.com/dists/stable/main/binary-amd64/Packages
+        url: https://brave-browser-apt-release.s3.brave.com/pool/main/b/brave-browser/brave-browser_1.14.84_amd64.deb
+        sha256: 3338e0723cf1186a8bd6a7cbfe5045773eedfa00f798c2b4aecd88850842832e
+        size: 75314612
+        filename: chrome.deb
```
x80486, line 53: I know it doesn't matter, but I would rename anything-chrome into `brave` — across all files.

com.brave.Brave.yaml
analotia: It worked for me. Tell me if you notice anything wrong. Also, I appreciate if you could help with the printing problem (note: `--sockets=cups` doesn't help)

com.brave.Brave.yaml
analotia: Done, tell me if something weird happened.

analotia:
bot, build com.brave.Brave

analotia:
What should the license for Brave be? I just noticed I forgot to change license

analotia:
bot, build com.brave.Brave

analotia:
bot, build com.brave.Brave

analotia:
bot, build com.brave.Brave

analotia:
bot, build com.brave.Brave

analotia:
bot, build com.brave.Brave

analotia:
Update: with the zypak sandbox I am unable to view PDFs or print.

analotia:
`/app/extra/brave --no-sandbox` could view PDFs just fine.

fmarier (changes requested):
Thanks for getting the ball rolling on this, much appreciated!

I left a few comments inline, mostly just to align with the community-maintained Snap (https://snapcraft.io/install/brave/ubuntu).

com.brave.Brave.yaml
```diff
@@ -0,0 +1,81 @@
...
+  - name: brave
+    buildsystem: simple
+    build-commands:
+      - 'install -Dm 755 brave.sh /app/bin/brave'
+      - 'install -Dm 755 apply_extra.sh /app/bin/apply_extra'
+      - 'install -Dm 755 stub_sandbox.sh /app/bin/stub_sandbox'
+      - 'install -Dm 644 -t /app/share/applications com.brave.Brave.desktop'
+      - 'install -Dm 644 -t /app/share/metainfo com.brave.Brave.metainfo.xml'
+      - |
+          for icon in 16 24 32 48 64 128 256; do
```
x80486, line 41: You can download the [PublisherMediaAssets.zip](https://www.brave.com/static-assets/files/PublisherMediaAssets.zip) and use `../Logos/Brave/SVG/logo/primary/brave_lion.svg` instead of all the PNGs: `install -D --mode=0644 /path/to/brave_lion.svg /app/share/icons/hicolor/scalable/apps/com.brave.Brave.svg`.

If you want to make sure you always have the latest, you can [use the ZIP file as a source](https://github.com/flathub/com.jetbrains.IntelliJ-IDEA-Ultimate/blob/master/com.jetbrains.IntelliJ-IDEA-Ultimate.json#L27).

com.brave.Brave.yaml
```diff
@@ -0,0 +1,81 @@
...
+            install -Dm 644 com.brave.Brave-$icon.png /app/share/icons/hicolor/${icon}x${icon}/apps/com.brave.Brave.png
+          done
+
+      # For extra-data to be able to extract the .deb
+      - 'install -Dm 755 /{usr,app}/bin/ar'
+      - 'cp /usr/lib/$(gcc --print-multiarch)/libbfd-*.so /app/lib'
+    sources:
+      - type: extra-data
```
x80486, line 49: By the way, if Brave's license supports this, I think you can get rid of the `extra-data` business and use `file`. It's way more simple...but I don't know about the license restrictions, etc.

com.brave.Brave.yaml
analotia: I don't think I can because this will be used by `.desktop`. I'm not aware of any way to use `.svg`s with `.desktop`

analotia: Feel free to unresolve this

com.brave.Brave.metainfo.xml
```diff
@@ -0,0 +1,30 @@
+<!-- Copyright 2017 the Chromium authors, with modifications -->
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id>com.brave.Brave</id>
+  <launchable type="desktop-id">com.brave.Brave.desktop</launchable>
+  <name>Brave Browser</name>
+  <developer_name>Brave Software</developer_name>
+  <summary>The web browser from Brave</summary>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>GPL-2.0-or-later or MPL-2.0</project_license>
```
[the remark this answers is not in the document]
analotia: Thank you!

com.brave.Brave.yaml
x80486: When you use the SVG, the location is different: `/app/share/icons/hicolor/scalable/apps/`. That alone is sufficient to convert and place automatically those PNGs during the build/packaging process.

analotia:
bot, build com.brave.Brave

analotia:
bot, build com.brave.Brave

analotia:
bot, build com.brave.Brave

brave.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/bash
+export TMPDIR="$XDG_RUNTIME_DIR/app/$FLATPAK_ID"
+#exec zypak-wrapper /app/extra/brave "$@"
+exec /app/extra/brave --no-sandbox "$@"
```
fmarier, line 4: Just a thought, but is it possible to disable the flatpak sandbox instead of the Chromium one?

com.brave.Brave.yaml
fmarier: @x80486 Can you talk more about what this means and what sort of license restrictions you are unsure about?

x80486: You would have to wait for Flathub's owners to explain this better, but I think it boils down to whether Brave allows or not to redistribute/repackage the browser...is that right, @barthalion?

brave.sh
analotia: Unfortunately not.  

With flatpak packaging you need to either disable the Chromium sandbox or use zypak to provide a fake sandbox to Chromium (zypak trick uses LD_PRELOAD).

analotia: I prefer to use zypak because it's a bit more secure and you wouldn't get the `--no-sandbox` warning every time.

analotia: I disabled Chromium sandbox. If I'm not mistaken it shouldn't have any impact on security because flatpak is already sandboxing everything

analotia: https://flatpak.org/faq/#Is_Flatpak_compatible_with_other_desktop_isolation_frameworks_

analotia:
bot, build com.brave.Brave

brave.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/sh
+export TMPDIR="$XDG_RUNTIME_DIR/app/$FLATPAK_ID"
+exec /app/extra/brave-browser --test-type --no-sandbox "$@"
```
fmarier, line 3: What's the purpose of the `--test-type` parameter here?

According to [this list](https://peter.sh/experiments/chromium-command-line-switches/#test-type), it looks like it's just for running the Chromium tests:

> Type of the current test harness ("browser" or "ui").

analotia: It hides the `--no-sandbox` warning.  If you don't use `--test-type` you will be given the following message every time you open up Brave browser/Chromium : 
>  You are using an unsupported command-line flag: --no-sandbox. Stability and security will suffer. 

I don't know of a better way to get rid of the message.

analotia: Try `brave-browser --no-sandbox` and `brave-browser --test-type --no-sandbox` and you'll understand what I mean.  

It would be better if there was a better/cleaner way of doing this..

analotia: https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/ui/startup/startup_browser_creator_impl.cc;l=743

This is the Chromium code that disables the warning. `--test-type` is the only choice I have.   

If I use `--enable-automation` then a warning would show. `--test-type` is the only flag that disables the warning and doesn't notify the user that it's set.

fmarier: I see. It might be best to avoid that option and to just live with the warning. `--test-type` is even less of a supported configuration than just `--no-sandbox`, and I don't know what the consequences of using it might be. There's also no guarantee that it will keep working that way since it's only meant for running tests.

analotia: I think it's best to just revert to using `zypak` even though it's broken. The developer explained that `zypak` was actually sandboxing for chromium and that it wasn't just providing a fake useless sandbox.  

Basically with `zypak` Chrome's sandboxed processes have no access to the internet, files, etc. which is the goal of the regular Chromium sandbox.  It works by using `flatpak-spawn`   https://docs.flatpak.org/en/latest/flatpak-command-reference.html#flatpak-spawn

analotia: I'm not sure about whether I should use `--no-sandbox` or use `zypak`.  

Security in something like a web browser is pretty important but having it work is just as important

brave.sh
bilelmoussaoui: Any reason for not using zypak? The commits don't explain anything in that regard

.gitmodules
```diff
@@ -0,0 +1,3 @@
+#[submodule "shared-modules"]
```
bilelmoussaoui, line 1: Not used.

com.brave.Brave.desktop
```diff
@@ -0,0 +1,221 @@
+[Desktop Entry]
```
bilelmoussaoui, line 1: should be contributed upstream.

com.brave.Brave.metainfo.xml
```diff
@@ -0,0 +1,37 @@
+<!-- Copyright 2017 the Chromium authors, with modifications -->
+<?xml version="1.0" encoding="UTF-8"?>
```
bilelmoussaoui, line 2: Same as the desktop file

com.brave.Brave.metainfo.xml
```diff
@@ -0,0 +1,37 @@
...
+<component type="desktop">
```
bilelmoussaoui, line 3:
```suggestion
<component type="desktop-application">
```

com.brave.Brave.metainfo.xml
```diff
@@ -0,0 +1,37 @@
...
+    </screenshot>
+    <screenshot>
+      <image>https://i.imgur.com/k6ozpw1.png</image>
+    </screenshot>
+  </screenshots>
+  <releases>
+    <release version="1.14.84" date="2020-08-25"/>
+  </releases>
+  <content_rating type="oars-1.1">
+  </content_rating>
```
bilelmoussaoui, line 35:
```suggestion
  <content_rating type="oars-1.1" />
```

com.brave.Brave.yaml
bilelmoussaoui: Installing the svg as well would be appreciated.

brave.sh
analotia: @bilelmoussaoui zypak breaks PDF, printing, extensions, and some other things

com.brave.Brave.desktop
analotia: it is from upstream but I had to change the `icon=` part to `com.brave.Brave`

bilelmoussaoui: Use rename-icon instead of copying the file here then.

brave.sh
bilelmoussaoui: Have you reported that to upstream zypak?

analotia:
bot, build com.brave.Brave

com.brave.Brave.desktop
analotia: is it possible for me to change, `Exec=` too? I checked the original and it seems like they used an absolute path to `/usr/bin/brave-browser`

brave.sh
analotia: yes, it's a known issue

analotia:
Is there any better way I could do https://github.com/flathub/flathub/pull/1852/commits/46b49fd32c2646cfa91fb8a97e7878be62d20bf4 ?

analotia:
Update2: zypak is working great with user namespaces. Printing, etc. all work.

analotia:
bot, build com.brave.Brave

brave.sh
analotia: Nevermind. Fixed. We're using zypak now!

analotia:
bot, build com.brave.Brave

analotia:
bot, build com.brave.Brave

analotia:
bot, build com.brave.Brave

com.brave.Brave.yaml
```diff
@@ -0,0 +1,110 @@
...
+separate-locales: false
+build-options:
+  no-debuginfo: true
+finish-args:
+  - '--device=all'
+  - '--share=ipc'
+  - '--share=network'
+  - '--socket=pulseaudio'
+  - '--socket=x11'
+  - '--socket=wayland'
```
TheEvilSkeleton, line 17: Can you remove that? Brave doesn't natively support Wayland.

analotia: done

analotia:
When I try to add a web app to desktop, I get this: 
```
LaunchProcess: failed to execvp:
xdg-desktop-menu
LaunchProcess: failed to execvp:
xdg-icon-resource
LaunchProcess: failed to execvp:
xdg-icon-resource
LaunchProcess: failed to execvp:
xdg-icon-resource
LaunchProcess: failed to execvp:
xdg-icon-resource
LaunchProcess: failed to execvp:
xdg-desktop-menu
```


Any way I could add `xdg-desktop-menu` and `xdg-icon-resource` to the flatpak? I tried to simply copy these scripts but then I need `update-desktop-database`.   

Also when that works I'm stuck with the exec being `/app/extra/brave-browser` which wouldn't work. Is it possible to automatically translate these to `flatpak run com.brave.Brave...`?

analotia:
bot, build com.brave.Brave

analotia:
I made a small hack so that web apps will show on the user's desktop. However the problem is that there is no real way to do this with flatpak mainly because there is no portal for that and that $XDG_DATA_HOME can't be exposed to the flatpak

analotia:
Therefore `XDG_DATA_HOME` will always be set to `~/.local/share`.  

If there is a better way please tell me

analotia:
Could someone please review the permissions?

analotia:
bot, build com.brave.Brave

analotia:
Battery API doesn't work for some reason: https://googlechrome.github.io/samples/battery-status/

analotia:
Is there a mailing list so I could be notified of stable updates?

com.brave.Brave.yaml
```diff
@@ -20,15 +20,11 @@ finish-args:
...
   - '--system-talk-name=org.freedesktop.Avahi'
-  - '--own-name=org.mpris.MediaPlayer2.*'
-  - '--filesystem=host-etc'
-  - '--filesystem=xdg-documents'
-  - '--filesystem=xdg-download'
-  - '--filesystem=xdg-music'
-  - '--filesystem=xdg-videos'
+  - '--system-own-name=org.mpris.MediaPlayer2.chromium.*'
+  - '--system-own-name=org.freedesktop.GeoClue2'
+  - '--filesystem=host'
```
TheEvilSkeleton, line 27: Doesn't that make the sandboxing somewhat nonexistent? You're giving the Flatpak rw access to the whole filesystem (ro if directories/files are owned by someone else)

analotia: This true but the problem is that I use chromium to load PDFs from other places like `/media` and sometimes `/mnt`. 

So I figured that I might as well give full access because you never know what a user might want to load.

analotia: I've decided to only allow /etc and xdg-download. The user decides what they want to override later, not my job

analotia:
bot, build com.brave.Brave

com.brave.Brave.yaml
TheEvilSkeleton: > This true but the problem is that I use chromium to load PDFs from other places like /media and sometimes /mnt.

Can you try

```json
- '--filesystem=/media:ro'
- '--filesystem=/mnt:ro`
```
and check if it works?

com.brave.Brave.yaml
```diff
@@ -0,0 +1,117 @@
...
+  - '--talk-name=org.freedesktop.FileManager1'
+  - '--talk-name=org.freedesktop.Notifications'
+  - '--talk-name=org.freedesktop.secrets'
+  - '--talk-name=org.gnome.SessionManager'
+  - '--own-name=org.mpris.MediaPlayer2.chromium.*'
+  - '--system-talk-name=org.freedesktop.Avahi'
+  - '--system-talk-name=org.freedesktop.GeoClue2'
+  - '--system-talk-name=org.freedesktop.UPower'
+  - '--filesystem=host-etc:ro'
+  - '--filesystem=xdg-download'
```
TheEvilSkeleton, line 28: https://github.com/flathub/flathub/pull/1852#discussion_r495574128

Can you try adding

```json
- '--filesystem=/media:ro'
- '--filesystem=/mnt:ro'
```
and check if it works?

com.brave.Brave.yaml
analotia: It would work but not for everybody. Let's be restrictive by default and only allow rw to ~/Downloads and then let the user allow whatever they want with Flatseal  

That's what Firefox is doing

com.brave.Brave.yaml
analotia: It would work but not for everybody. Let's be restrictive by default and only allow rw to ~/Downloads and then let the user allow whatever they want with Flatseal

That's what Firefox is doing

analotia: Why would we want to do something as specific as this?

TheEvilSkeleton: > It would work but not for everybody. Let's be restrictive by default and only allow rw to ~/Downloads and then let the user allow whatever they want with Flatseal
> 
> That's what Firefox is doing

Understandable.

fmarier:
> Battery API doesn't work for some reason: https://googlechrome.github.io/samples/battery-status/
> 
> Edit: That's probably intentional. Never mind

Yes, we disable that for privacy reasons: https://github.com/brave/brave-browser/wiki/Deviations-from-Chromium-(features-we-disable-or-remove)#services--features-we-disable-entirely

analotia:
bot, build com.brave.Brave

com.brave.Brave.yaml
```diff
@@ -0,0 +1,119 @@
...
+      # - type: dir
+      #   path: ../../code/zypak
+
+  - name: brave
+    buildsystem: simple
+    build-commands:
+      - 'install -Dm 644 brave_lion.svg /app/share/icons/hicolor/scalable/apps/com.brave.Brave.svg'
+      - 'install -Dm 755 brave.sh /app/bin/brave'
+      - 'install -Dm 755 apply_extra.sh /app/bin/apply_extra'
+      - 'install -Dm 755 update-desktop-database /app/bin/update-desktop-database'
```
x80486, line 81: I wonder why would you need these for?

analotia: you could add web apps that way. These "installable" web pages like Twitter, Element/Riot.im, etc.

analotia: If you have the flatpak you should be able to install web pages to desktop/application menu.

but you will need the latest build: `flatpak install --user https://dl.flathub.org/build-repo/27502/com.brave.Brave.flatpakref
`

com.brave.Brave.desktop
barthalion: You can use `desktop-file-edit` for that.

com.brave.Brave.metainfo.xml
```diff
@@ -0,0 +1,36 @@
...
+  <summary>Community-maintained unofficial Flatpak of Brave Browser</summary>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>MPL-2.0</project_license>
+  <url type="homepage">https://brave.com/</url>
+  <description>
+   <p>
+      Browse faster by blocking ads and trackers that violate your privacy and cost
+      you time and money.
+   </p>
+   <p>
```
barthalion, line 17: This needs explicit disclaimer like https://flathub.org/apps/details/com.jetbrains.PyCharm-Professional has.

com.brave.Brave.metainfo.xml
```diff
@@ -0,0 +1,36 @@
+<!-- Copyright 2017 the Chromium authors, with modifications -->
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.brave.Brave</id>
+  <launchable type="desktop-id">com.brave.Brave.desktop</launchable>
+  <name>Brave Browser</name>
+  <developer_name>Brave Software</developer_name>
+  <summary>Community-maintained unofficial Flatpak of Brave Browser</summary>
```
barthalion, line 8: I don't see why summary should say "community-maintained".

com.brave.Brave.yaml
barthalion: > you could add web apps that way. These "installable" web pages like Twitter, Element/Riot.im, etc.

No, you cannot. The application has no access to host directory with desktop launchers. Same about `xdg-*` scripts.

com.brave.Brave.yaml
```diff
@@ -0,0 +1,119 @@
...
+      - 'install -Dm 644 -t /app/share/applications com.brave.Brave.desktop'
+      - 'install -Dm 644 -t /app/share/metainfo com.brave.Brave.metainfo.xml'
+
+      # For extra-data to be able to extract the .deb
+      - 'install -Dm 755 /{usr,app}/bin/ar'
+      - 'cp /usr/lib/$(gcc --print-multiarch)/libbfd-*.so /app/lib'
+    sources:
+      - type: extra-data
+        # From https://brave-browser-apt-release.s3.brave.com/dists/stable/main/binary-amd64/Packages
+        url: https://brave-browser-apt-release.s3.brave.com/pool/main/b/brave-browser/brave-browser_1.14.84_amd64.deb
```
barthalion, line 94: This needs to be managed automatically via flatpak-external-data-checker.

com.brave.Brave.metainfo.xml
```diff
@@ -0,0 +1,36 @@
...
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>MPL-2.0</project_license>
+  <url type="homepage">https://brave.com/</url>
+  <description>
+   <p>
+      Browse faster by blocking ads and trackers that violate your privacy and cost
```
barthalion, line 14: This could be expanded upon.

com.brave.Brave.metainfo.xml
```diff
@@ -0,0 +1,36 @@
+<!-- Copyright 2017 the Chromium authors, with modifications -->
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.brave.Brave</id>
+  <launchable type="desktop-id">com.brave.Brave.desktop</launchable>
+  <name>Brave Browser</name>
+  <developer_name>Brave Software</developer_name>
+  <summary>Community-maintained unofficial Flatpak of Brave Browser</summary>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>MPL-2.0</project_license>
```
barthalion, line 10: Unless binary is under different license, you don't need extra-data.

barthalion (changes requested):
Just to be clear, unless you get upstream on board with this, I don't see it being available outside Flathub beta. Upstream involvement is a must for major browsers (or at least someone seriously committed to maintenance and established track record in the community).

com.brave.Brave.yaml
```diff
@@ -0,0 +1,119 @@
...
+  - '--talk-name=org.freedesktop.secrets'
+  - '--talk-name=org.gnome.SessionManager'
+  - '--own-name=org.mpris.MediaPlayer2.chromium.*'
+  - '--system-talk-name=org.freedesktop.Avahi'
+  - '--system-talk-name=org.freedesktop.GeoClue2'
+  - '--system-talk-name=org.freedesktop.UPower'
+  - '--filesystem=host-etc:ro'
+  - '--filesystem=xdg-desktop'
+  - '--filesystem=xdg-download'
+  - '--filesystem=~/.local/share/applications/'
```
barthalion, line 30: That's too big hole to be ever allowed on Flathub.

com.brave.Brave.yaml
```diff
@@ -0,0 +1,119 @@
...
+base: org.electronjs.Electron2.BaseApp
+base-version: '20.08'
+runtime: org.freedesktop.Platform
+runtime-version: '20.08'
+sdk: org.freedesktop.Sdk
+command: brave
+separate-locales: false
+build-options:
+  no-debuginfo: true
+finish-args:
```
barthalion, line 11: Why `--require-version=1.8.1` has been dropped?

com.brave.Brave.metainfo.xml
analotia: It was a suggestion from Brave. They said they wanted it to say that it's "community-maintained" and that it's not recommended by Brave themselves

com.brave.Brave.yaml
analotia: Removed

com.brave.Brave.metainfo.xml
analotia: Is the new version better?

com.brave.Brave.metainfo.xml
analotia: Should I resolve it now?

analotia:
> Just to be clear, unless you get upstream on board with this, I don't see it being available outside Flathub beta. Upstream involvement is a must for major browsers (or at least someone seriously committed to maintenance and established track record in the community).

So if we could get Brave Developers as owners of the repo it could possibly make it to flathub?

com.brave.Brave.metainfo.xml
analotia: I'm not sure because of widevine and other proprietary stuff bundled with Brave browser's official binary

analotia:
bot, build com.brave.Brave

analotia:
bot, build com.brave.Brave

com.brave.Brave.yaml
analotia: Fixed! Tell me if I set it up wrong.

analotia:
bot, build com.brave.Brave

com.brave.Brave.metainfo.xml
```diff
@@ -0,0 +1,34 @@
+<!-- Copyright 2017 the Chromium authors, with modifications -->
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.brave.Brave</id>
```
barthalion, line 4: Just as another data point, Android seems to use `com.brave.browser`. Would be good to settle on something with upstream to avoid renames in the future.

com.brave.Brave.metainfo.xml
barthalion: Yeah, that's a fair point.

com.brave.Brave.metainfo.xml
barthalion: cc @fmarier

com.brave.Brave.metainfo.xml
barthalion: cc @fmarier

com.brave.Brave.yaml
barthalion: analotia raised a good point about widevine so I guess extra-data needs to stay.

barthalion:
Left from the reviews:

- [ ] Improve description in metainfo file.
- [ ] Ensure the ID is right.
- [ ] Contribute desktop and metainfo files upstream.
- [ ] Clarify upstream involvement.

Most importantly, we want to make sure Brave at least "approves" the submission and helps with maintenance.

com.brave.Brave.metainfo.xml
fmarier: `com.brave.browser` sounds good to me.

analotia: Should I capitalize `browser` so it becomes `com.brave.Browser`?

analotia: @fmarier Could Brave approve and maintain this?

analotia:
@fmarier Could Brave approve and maintain this? Check barthalion's comment above

fmarier:
>￼ Clarify upstream involvement.

With respect to the Snap, what happened was that a community member put something unofficial together and maintained it for a while. After showing that it was working, stable and that there was enough interest for it, we eventually took it over and are currently automating it. I can see something similar happening here.

>  Contribute desktop and metainfo files upstream.

In general, we'd rather keep as much of the packaging code out of the main repo as possible. For example, we have split a bunch of code from the `.deb` and `.rpm` packages into https://github.com/brave/brave-release.

fmarier:
The flatpak generally looks good to me. Thanks @analotia for all of your work on this!

The one thing we'd like to see, based on our experience with the unofficial Snap (which is in the process of becoming official as I mentioned above), would be to list this as unofficial and community-maintained explicitly for now. Ideally linking to our [official installation instructions](https://brave-browser.readthedocs.io/en/latest/installing-brave.html#linux).

analotia:
Can the extra-data script copy to `/app/share/applications` or is that illegal in flatpak ?

barthalion:
@fmarier Can we assume we are granted redistribution rights or should the manifest keep using `extra-data` mechanism? (i.e. deb is downloaded and processed by and on the user's computer)

fmarier:
> Can we assume we are granted redistribution rights or should the manifest keep using `extra-data` mechanism? (i.e. deb is downloaded and processed by and on the user's computer)

The Brave source code is released under the MPL 2.0 license, so that does include the right to redistribute.

analotia:
> > Can we assume we are granted redistribution rights or should the manifest keep using `extra-data` mechanism? (i.e. deb is downloaded and processed by and on the user's computer)
> 
> The Brave source code is released under the MPL 2.0 license, so that does include the right to redistribute.

Does that include the right to redistribute the binaries? The problem for me is that the binaries contain some proprietary things like Widevine.

fmarier:
> Does that include the right to redistribute the binaries? The problem for me is that the binaries contain some proprietary things like Widevine.

Widevine is [not bundled with Brave, it's downloaded at runtime](https://github.com/brave/brave-browser/wiki/Support-widevine-on-Brave-linux#brave-browser-changes). So the binaries you compile will not include it.

barthalion:
@analotia Mind replacing `extra-data` with native modules?

analotia:
bot, build com.brave.Browser

analotia:
bot, build com.brave.Browser

barthalion:
No complaints from me. @refi64, is it good enough to go to regular Flathub or should we keep it in beta till all Chromium-related issues are ironed out?

com.brave.Browser.yaml
```diff
@@ -0,0 +1,110 @@
...
+      - 'install -Dm 644 gtk-settings.ini /app/etc/xdg/gtk-3.0/settings.ini'
+    sources:
+      - type: file
+        path: gtk-settings.ini
+
+  - name: zypak
+    sources:
+      - type: git
+        url: https://github.com/refi64/zypak
+        branch: master
```
x80486, line 68: I think you might want to _pin_ this one to a particular commit or tag rather than use `master`.

refi64:
Ditto on pinning the zypak commit, but yeah IMO it should stay in beta
while the remaining Chromium/Chrome-family issues are suitably solved.

barthalion:
bot, build com.brave.Browser

PR #1863, opened 2020-09-30
Add libdvdcss extension for HandBrake

hadess:
`libdvdcss.json` is copy/pasta from org.gnome.Totem.


nedrichards:
I'm not against adding this when considered on its own but would prefer to see how https://github.com/HandBrake/HandBrake/pull/3164 ends up before merging since HandBrake is very much maintained by the upstream developers on Flathub.

hadess:
> I'm not against adding this when considered on its own but would prefer to see how [HandBrake/HandBrake#3164](https://github.com/HandBrake/HandBrake/pull/3164) ends up before merging since HandBrake is very much maintained by the upstream developers on Flathub.

The changes there are certainly necessary to get access to the optical drives, but I don't think I'd need authorisation to get this plugin merged in Flathub, would I?

darkdragon-001:
Related upstream issue is still open.

daviewales:
@nedrichards, is there a chance this can be re-opened? The related upstream PR has since been merged, so I'm guessing it may now be possible to merge this one.

bbhtt:
Please create a new PR.

PR #1869, opened 2020-10-03
Add com.todoist.Todoist

jurf:
Tested, should fully work. Entirely based on https://github.com/flathub/com.nordpass.NordPass, big thanks to them.


jurf:
I contacted upstream, they said they’ll support Flatpak once electron-builder supports it. In the meantime I made this package. I’ll notify them once it gets merged.

jurf:
bot, build com.todoist.Todoist

jurf:
bot, build com.todoist.Todoist

barthalion:
I'm somewhat wary about accepting this as I suspect how it's going to end: Todoist legal team contacting us about "illegal redistribution". I'd prefer if it were made clear to the vendor what Flathub is and why the app is already maintanable here today, even if by re-using snap artifact.

com.todoist.Todoist.yaml
```diff
@@ -0,0 +1,70 @@
...
+        filename: todoist.snap
+        url: https://api.snapcraft.io/api/v1/snaps/download/pP800io06HKwwkUmX8gDG1pFCfsinLHY_2.snap
+        sha256: c30cf3e0d5efc8131fcaacd847c7f2ad1932591c9ca2bca643f159b74c3c9a31
+        size: 66150400
+        x-checker-data:
+          type: snapcraft
+          name: todoist
+          channel: stable
+      - type: script
+        dest-filename: apply_extra
```
barthalion, line 55: The output should be silenced.

jurf:
I have shot them an email to see if they are okay with this and tried to explain the situation fully.

com.todoist.Todoist.yaml
jurf: How? I have yet to see an example of this.

barthalion: `1>/dev/null` kind of redirection will be good enough. Otherwise it breaks installation progress in terminal.

jurf: You mean on every line of the script? Shouldn’t it be silenced by Flatpak if it breaks progress?

barthalion: That's arguable. If there's an error during apply_extra, that's fine, but you shouldn't flood the terminal with irrelevant information.

nedrichards:
bot, build com.todoist.Todoist

jurf:
Okay, an excerpt with the communication with Todoist:

 * They do not have the resources to maintain this currently but are open to in the future, notably after the first stable release of the Linux version
 * They do not have a problem with it being here unofficially, as long as it’s with `extra-data` and a disclaimer

Thus, this should be good to go.

PR #1873, opened 2020-10-05
Initial clubhouse version

danigm:
Please, merge into `beta` branch. I don't want to go directly to the `stable` repository when this app gets approved, lets move to `beta` at first if it's possible, because we want to test a bit the flathub version before we do the real release for the final user.

This application is part of the "hack experience" that we developed for the [hack computer](https://www.hack-computer.com/) project.

As part of the new [EndlessOS foundation](https://endlessos.com/) goal, we're trying to release and publish "Hack" to make it possible to use outside of Endless.

This PR includes the clubhouse application manifest and the Clippy Extension.

I'm not sure about how to publish the extensions on flathub, so I added it here too. Let me know if I should create a different PR for the extension.

The clippy extension is something that we use in different apps to provide a full integration with the clubhouse, that's why we've this like an extension instead of a dependency on each toy app.


com.hack_computer.Clippy.Extension.json
```diff
@@ -0,0 +1,36 @@
+{
+  "app-id": "com.hack_computer.Clippy.Extension",
```
bilelmoussaoui, line 2: Should be sent as a separate PR

com.hack_computer.Clubhouse.json
```diff
@@ -0,0 +1,276 @@
...
+    "app-id": "com.hack_computer.Clubhouse",
+    "add-extensions": {
+        "com.hack_computer.Clippy.Extension": {
+            "version": "master",
+            "directory": "clippy",
+            "no-autodownload": false,
+            "autodelete": true
+        }
+    },
+    "branch": "master",
```
nedrichards, line 11: this needs to either be stable or omitted

com.hack_computer.Clubhouse.json
```diff
@@ -0,0 +1,276 @@
...
+        "--own-name=com.endlessm.GameStateService",
+        "--own-name=com.hack_computer.Clubhouse",
+        "--own-name=com.hack_computer.GameStateService",
+        "--own-name=com.hack_computer.HackSoundServer",
+        "--own-name=com.hack_computer.HackToolbox",
+        "--share=ipc",
+        "--socket=pulseaudio",
+        "--socket=session-bus",
+        "--socket=wayland",
+        "--socket=x11",
```
nedrichards, line 41: we prefer fallback-x11 here if wayland is available these days as that prefers wayland where available.

com.hack_computer.Clubhouse.json
```diff
@@ -0,0 +1,276 @@
...
+        "--filesystem=~/.local/share/flatpak",
+        "--nofilesystem=~/.local/lib/python3.5",
+        "--own-name=com.endlessm.GameStateService",
+        "--own-name=com.hack_computer.Clubhouse",
+        "--own-name=com.hack_computer.GameStateService",
+        "--own-name=com.hack_computer.HackSoundServer",
+        "--own-name=com.hack_computer.HackToolbox",
+        "--share=ipc",
+        "--socket=pulseaudio",
+        "--socket=session-bus",
```
nedrichards, line 39: you're going to need specific justification for snooping everything here rather than specific address'

nedrichards:
I'm fairly familiar with the concept of Hack (and a user ;-) but we're really going to need a permission by permission breakdown of why you need each thing here because there's a *lot*.

com.hack_computer.Clubhouse.json
```diff
@@ -0,0 +1,276 @@
...
+        "--filesystem=/var/lib/AccountsService/icons/:ro",
+        "--filesystem=/var/lib/flatpak/exports/share/applications:ro",
+        "--filesystem=/var/lib/flatpak/exports/share/icons:ro",
+        "--filesystem=host",
+        "--filesystem=xdg-run/dconf",
+        "--filesystem=~/.config/dconf:ro",
+        "--filesystem=~/.local/share/flatpak",
+        "--nofilesystem=~/.local/lib/python3.5",
+        "--own-name=com.endlessm.GameStateService",
+        "--own-name=com.hack_computer.Clubhouse",
```
bilelmoussaoui, line 33:
```suggestion
```
you own your app-id by default.

com.hack_computer.Clubhouse.json
```diff
@@ -0,0 +1,276 @@
...
+            "fc-cache -f /app/share/fonts/"
+          ],
+          "sources": [
+            {
+              "type": "archive",
+              "url": "https://github.com/source-foundry/Hack/releases/download/v3.003/Hack-v3.003-ttf.tar.xz",
+              "sha256": "d9ed5d0a07525c7e7bd587b4364e4bc41021dd668658d09864453d9bb374a78d"
+            }
+          ]
+        },
```
bilelmoussaoui, line 217:
```suggestion        {
          "name": "fonts",
          "buildsystem": "simple",
          "build-commands": [
            "install -D -m644 *.otf *.ttf -t /app/share/fonts"
          ],
          "post-install": [
            "fc-cache -f /app/share/fonts/"
          ],
          "sources": [
            {
              "type": "archive",
              "url": "https://github.com/chrismsimpson/Metropolis/archive/r9.tar.gz",
              "sha256": "283d1f51d96777c648dc1c0899b986fc99f14bfb709732428c14881c075c356b"
            },
            {
                "type": "archive",
                "url": "http://www.latofonts.com/download/Lato2OFL.zip",
                "sha256": "42b54e96c07e299d967fc3227c7bd63a20d6cfb1dc8fd6dae83628091e20a5b8"
            },
            {
                "type": "archive",
                "url": "https://github.com/source-foundry/Hack/releases/download/v3.003/Hack-v3.003-ttf.tar.xz",
                "sha256": "d9ed5d0a07525c7e7bd587b4364e4bc41021dd668658d09864453d9bb374a78d"
            }
          ]
        },
        
```

You could simplify this a lot

com.hack_computer.Clubhouse.json
```diff
@@ -0,0 +1,276 @@
...
+        {
+          "name": "hack-toolbox",
+          "buildsystem": "meson",
+          "run-tests": false,
+          "sources": [
+              {"type": "git", "url": "https://github.com/endlessm/hack-toolbox-app.git", "branch": "master", "commit": "e67b603cf7d949c36eb6c7dc4af086f34fbcfdb6"}
```
bilelmoussaoui, line 223: Please split those into multiple lines.

com.hack_computer.Clubhouse.json
```diff
@@ -0,0 +1,276 @@
...
+          ]
+        },
+        {
+            "name": "game-state-service",
+            "buildsystem": "meson",
+            "config-opts" : [
+                "-Dsession-bus-services-dir=/app/share/dbus-1/services"
+            ],
+            "sources": [
+                {"type": "git", "url": "https://github.com/endlessm/hack-game-state-service.git", "branch": "master", "commit": "bd09e69d3e17671f231afc04fa40775701b96266"}
```
bilelmoussaoui, line 233: Same here

com.hack_computer.Clubhouse.json
```diff
@@ -0,0 +1,276 @@
...
+            "sources": [
+                {
+                    "type": "git",
+                    "url": "https://github.com/endlessm/glibcoro.git",
+                    "commit": "b7e5e73e5813fda0d6631fef594cf50972ec811b"
+                }
+            ],
+            "build-options": [
+
+            ],
```
bilelmoussaoui, line 248:
```suggestion
```

com.hack_computer.Clubhouse.json
```diff
@@ -0,0 +1,276 @@
...
+            ]
+        },
+        {
+            "name": "clubhouse",
+            "buildsystem": "meson",
+            "build-options": {
+              "build-args": "",
+              "test-args": [
+                  "--env=GSETTINGS_SCHEMA_DIR=/run/host/usr/share/glib-2.0/schemas/",
+                  "--filesystem=host"
```
bilelmoussaoui, line 260: not sure about this?

com.hack_computer.Clubhouse.json
```diff
@@ -0,0 +1,276 @@
...
+            "version": "master",
+            "directory": "clippy",
+            "no-autodownload": false,
+            "autodelete": true
+        }
+    },
+    "branch": "master",
+
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "3.36",
```
LeandroStanger, line 14: Updade runtime to 3.38
```suggestion
    "runtime-version": "3.38",
```

com.hack_computer.Clippy.Extension.json
```diff
@@ -0,0 +1,36 @@
...
+  "branch": "master",
+  "runtime": "com.hack_computer.Clubhouse",
+  "runtime-version": "master",
+  "sdk": "org.gnome.Sdk//3.36",
```
LeandroStanger, line 6: Update sdk to 3.38
```suggestion
  "sdk": "org.gnome.Sdk//3.38",
```

com.hack_computer.Clippy.Extension.json
```diff
@@ -0,0 +1,36 @@
...
+      "name": "appdata",
+      "buildsystem": "simple",
+      "build-commands": [
+        "install -Dm644 --target-directory=${FLATPAK_DEST}/share/appdata com.hack_computer.Clippy.Extension.appdata.xml",
+        "appstream-compose --basename=com.hack_computer.Clippy.Extension --prefix=${FLATPAK_DEST} --origin=flatpak com.hack_computer.Clippy.Extension"
+      ],
+      "sources": [
+        {
+          "type": "file",
+          "path": "data/com.hack_computer.Clippy.Extension.appdata.xml"
```
LeandroStanger, line 31: New appdata files should use metainfo instead. https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html

com.hack_computer.Clubhouse.json
danigm: We talk to "random" apps using the `org.gtk.Actions` or `org.gtk.Application` interfaces, that's why we need this. We can remove this and just add the dbus names for the applications, but we've a lot of different quests with an app associated and that list will grow in the future. And also we want to allow third party quests, linked to third party apps, so I think that we can't control the full list of apps this app will talk to. Thinking about quests defined on user home, like `./local` or other common place, to allow to distribute `quests` independently of the clubhouse app.

Here's the code that creates those dbus proxies for apps in quests: https://github.com/endlessm/clubhouse/blob/master/eosclubhouse/system.py#L526-L616

com.hack_computer.Clubhouse.json
danigm: This app doesn't work yet with 3.38.

danigm:
> I'm fairly familiar with the concept of Hack (and a user ;-) but we're really going to need a permission by permission breakdown of why you need each thing here because there's a _lot_.

Let's try to justify each permission, it's possible that we've some permissions that we don't need anymore, so we can end with a smaller list:

 * FileSystem:
   * `--filesystem=/var/lib/AccountsService/icons/:ro`: This is needed to show the user avatar on the app: https://github.com/endlessm/clubhouse/blob/master/eosclubhouse/clubhouse.py#L2378
   * `--filesystem=/var/lib/flatpak/exports/share/applications:ro`: **Can't justify, to be removed**
   * `--filesystem=/var/lib/flatpak/exports/share/icons:ro`: **Can't justify, to be removed**
   * `--filesystem=host`, related with the user icons, https://github.com/endlessm/clubhouse/blob/master/eosclubhouse/clubhouse.py#L2380.
   * `--filesystem=xdg-run/dconf`: Settings
   * `--filesystem=~/.config/dconf:ro`: Settings
   * `--filesystem=~/.local/share/flatpak`: **leftover, to be removed**
   * `--nofilesystem=~/.local/lib/python3.5`: This is here to avoid collision with custom python modules installed, as you can see, this is old, and it's not working with new python versions, so I think it's safe **to remove**, or we'll need to add the current sdk python version.

 * Permissions:
   * `--share=ipc`: Graphical app
   * `--socket=wayland`: Graphical app
   * `--socket=fallback-x11`: Graphical app
   * `--socket=pulseaudio`: Play sounds
   * `--socket=session-bus`: Talk to "random" apps, using the `org.gtk.Application` interface. Take a look to this comment: https://github.com/flathub/flathub/pull/1873#discussion_r500116002

 * System talk:
   * `--system-talk-name=org.freedesktop.Accounts`: Used to get the user name
   * `--system-talk-name=org.freedesktop.NetworkManager`: Used to detect if there's network connection
   * `--system-talk-name=org.freedesktop.hostname1`: Used to detect if the device is a desktop or laptop
   * `--system-talk-name=com.endlessm.Metrics`: For endlessos metrics integration

 * Session talk:
   * `--talk-name=ca.desrt.dconf`: Settings
   * `--talk-name=com.endlessm.Libanimation`: **leftover, to be removed**
   * `--talk-name=org.freedesktop.Flatpak`: To be able to run `flatpak-spawn`, to check if a flatpak app is installed: https://github.com/endlessm/clubhouse/blob/master/eosclubhouse/system.py#L661-L669
   * `--talk-name=org.freedesktop.DBus`: Used to check if an app is running: https://github.com/endlessm/clubhouse/blob/master/eosclubhouse/system.py#L316
   * `--talk-name=org.gtk.Notifications`: The app uses notifications
   * `--talk-name=org.gnome.Shell`: Integrates with the gnome-shell, to install the hack extension, check the shell and other desktop integrations: https://github.com/endlessm/clubhouse/blob/master/eosclubhouse/system.py#L36
   * `--talk-name=com.hack_computer.hack`: Integration with the shell extension dbus API
   * `--talk-name=org.endlessos.onboarding`: Integrations with the (not published yet) onboarding extension
   * `--talk-name=org.gnome.Software`: To install required apps for quests: https://github.com/endlessm/clubhouse/blob/master/eosclubhouse/software.py#L56
   * `--talk-name=com.endlessm.EknServices3.SearchProviderV3`: Integration with endlessos eknservice

nedrichards:
@danigm thanks that's really helpful - and so to be clear for everyone else, when you mean 'settings' you mean 'this app needs to be able to change the settings of the system and other applications, so using the sandboxed settings is unhelpful', right? ;-)

nedrichards:
For the user icons, are they really stored in a totally arbitrary path? Or can we narrow the host permissions to home+some specific holes.

danigm:
> @danigm thanks that's really helpful - and so to be clear for everyone else, when you mean 'settings' you mean 'this app needs to be able to change the settings of the system and other applications, so using the sandboxed settings is unhelpful', right? ;-)

About the settings, previous versions modified some settings on the shell, so maybe we can remove this because I can't find anything related in the current code.

> For the user icons, are they really stored in a totally arbitrary path? Or can we narrow the host permissions to home+some specific holes.

I'm not sure about this, I think that the path depends on the final distribution, for example in my current ubuntu (20.04) default icons are in `/usr/share/pixmaps/faces/`, so we can think about remove the full `host` filesystem and instead use `/usr/share/pixmaps/`.

nedrichards:
bot, build com.hack_computer.Clubhouse

danigm:
bot, build com.hack_computer.Clubhouse

danigm:
bot, build com.hack_computer.Clubhouse

barthalion:
At least in Operating Systems track, attempts to start the course end up with "this App isn't installed" notification, even though GNOME Terminal is installed. (This is Fedora Silverblue 32.)

nedrichards:
bot, build com.hack_computer.Clubhouse

com.hack_computer.Clubhouse.json
nedrichards: Yeah, that's legit (and it's what I thought but we need to make it clear in review)

danigm:
> At least in Operating Systems track, attempts to start the course end up with "this App isn't installed" notification, even though GNOME Terminal is installed. (This is Fedora Silverblue 32.)

We check for app installed as flatpak and it's a workaround. This is not working correctly right now with system apps, we need to find a way to detect if an app is installed from the flatpak sandbox, but maybe that's not possible. Another solution could be to just notify but let the user continue if we don't detect that.

I'm working right now on this. This kind of issues are the reason I want to have this on flathub-beta, because we need to test on different systems. This app has been designed to work on EndlessOS with some integrations and we need to test outside of EndlessOS

nedrichards:
@danigm who else from endless needs to be added to the repo?

danigm:
> @danigm who else from endless needs to be added to the repo?

You can add @erikos and @ramcq

PR #1875, opened 2020-10-07
Initial flameshot flatpak release

borgmanJeremy:


nedrichards:
bot, build org.flameshot.flameshot

borgmanJeremy:
bot, build org.flameshot.flameshot

borgmanJeremy:
bot, build org.flameshot.flameshot

borgmanJeremy:
bot, build org.flameshot.flameshot

COPYING
```diff
@@ -0,0 +1,502 @@
+                  GNU LESSER GENERAL PUBLIC LICENSE
```
bilelmoussaoui, line 1: Is this needed?

org.flameshot.flameshot.yml
```diff
@@ -0,0 +1,33 @@
...
+  - --share=network
+  # QtSingleApplication, allow other instances to see log files
+  - --env=TMPDIR=/var/tmp
+  # Allow loading/saving files from anywhere
+  - --filesystem=host
+  # Notification access
+  - --talk-name=org.freedesktop.Notifications
+  # System Tray Icon
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --own-name=org.kde.StatusNotifierItem-2-1
```
bilelmoussaoui, line 23: This won't work.

borgmanJeremy: Can you provide more clarity on which part won't work?

COPYING
borgmanJeremy: No I can pull that out.

borgmanJeremy:
bot, build org.flameshot.flameshot

borgmanJeremy:
bot, build org.flameshot.flameshot

borgmanJeremy:
@bilelmoussaoui Is there a better way to test this locally without waiting on the flathubot? This is my first flathub submission and I'm struggling to get everything done quite right.

org.flameshot.flameshot.yml
nedrichards: the numbers at the end are the PID. Flatpak apps have sandboxed PIDs so they can collide with PIDs from other apps or the system, not a problem if that ID stays inside the sandbox. Once you rely on poking a hole into the rest of the system and using that ID as a unique value though, things can go wrong. The more flatpak apps you use that use the KDE statusnotifier protocol, the more likely this is to happen.

bilelmoussaoui:
@borgmanJeremy you can run `flatpak-builder --install repo org.flameshot.flameshot.json --user --force-clean` which will build & install the app. Note that current build is failing due to a missing 128px icon, you can either add that or ship a scalable SVG one.

borgmanJeremy:
@bilelmoussaoui When I build it locally it correctly finds the icons: 

```
Renaming org.flameshot.flameshot.metainfo.xml to share/appdata/org.flameshot.flameshot.appdata.xml
Running appstream-compose
Processing application org.flameshot.flameshot
Saving icon /app/share/app-info/icons/flatpak/64x64/flameshot.png
Saving icon /app/share/app-info/icons/flatpak/128x128/flameshot.png
Saving AppStream /app/share/app-info/xmls/org.flameshot.flameshot.xml.gz
Done!
Committing stage cleanup to cache
Finishing app
Not exporting share/applications/flameshot.desktop, non-allowed export filename
Exporting share/icons/hicolor/48x48/apps/org.flameshot.flameshot.png
Exporting share/icons/hicolor/128x128/apps/org.flameshot.flameshot.png
Exporting share/icons/hicolor/scalable/apps/org.flameshot.flameshot.svg
Not exporting share/dbus-1/services/org.flameshot.Flameshot.service, non-allowed export filename
Please review the exported files and the metadata
Committing stage finish to cache
```

bilelmoussaoui:
Well, Flathub expects that you install the appdata, desktop file & icons under the same name (the app-id). Which is currently incorrect see https://github.com/flameshot-org/flameshot/blob/flathub/src/CMakeLists.txt#L146-L147
you also have to update the name here https://github.com/flameshot-org/flameshot/blob/flathub/docs/appdata/org.flameshot.flameshot.metainfo.xml#L31.
You could use rename-desktop-file if you don't want to do the renames & just rename during install (not very recommended).

The flathub builder has more restrictions as it requires a proper icon and so on that are not part of flatpak-builder.

borgmanJeremy:
bot, build org.flameshot.flameshot

org.flameshot.flameshot.yml
borgmanJeremy: Okay thanks for clarifying, ill figure out a way to fix it.

bilelmoussaoui, line 23:
```suggestion
  - --own-name=org.kde.*
```
That's the only way to "fix it"

org.flameshot.flameshot.yml
```diff
@@ -0,0 +1,33 @@
...
+modules:
+  - name: flameshot
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+    sources:
+      - type: git
+        url: https://github.com/flameshot-org/flameshot.git
+        branch: "flathub"
+        commit: "700ad3cf7361aa9077d94c978e7aad6a97faca84"
```
bilelmoussaoui, line 33: This should be a stable release unless this one is

org.flameshot.flameshot.yml
borgmanJeremy: Thanks, I will make this change and update.

org.flameshot.flameshot.yml
borgmanJeremy: Yeah I will move to our master branch and use tag / commit rather than branch / commit. I was doing this work on a side branch to get everything working before merging.

bilelmoussaoui: Sounds good, let us know once it's ready

tinywrkb:
app-id is not correct, it should be capitalized to `org.flameshot.Flameshot`, also add `--own-name=org.flameshot.Flameshot` and you can see (e.g. D-Feet) that the app registers itself with this name.  
I can't remember how I noticed this and my D-Bus know-how is limited but the app registers itself and then can be started via D-Bus.

It's probably not reported but there's is an issue with the StatusNotifierItem where the IconName is not using the app-id as expected and StatusNotifierHost / system tray only has access to exported icons. Again, D-Feet is your friend look at `org.kde.StatusNotifierItem-[i]-[j] -> StatusNotifierItem -> Interfaces -> org.kde.StatusNotifierItem-[i]-[j] -> Properties -> IconName`.  
You're going to need to patch the app or report this upstream. I haven't bother reporting it because I didn't find the app useful as it still can't capture under Wayland.

borgmanJeremy:
@tinywrkb Thanks for the feedback I'll get that resolved.

FYI I am the maintainer of the upstream now and always appreciate well written bug reports so please submit your issues :). Other users are having success with wayland and we are trying to hammer out any more bugs with wayland support.

tinywrkb:
@borgmanJeremy Thanks for your reply.   
I will add that haven't built this Flatpak manifest but based on my own manifest (see [here](https://github.com/tinywrkb/flatpaks/blob/master/org.flameshot.Flameshot/org.flameshot.Flameshot.yaml)) I suggest to remove the shell completion scripts, and to move the metadata file from the metainfo folder into the appdata folder due to a flatpak-builder annoyance (see my manifest).

borgmanJeremy:
bot, build org.flameshot.flameshot

borgmanJeremy:
Closing in favor of PR with proper name

PR #1884, opened 2020-10-13
openMSX flatpak support

rderooy:
This adds the openMSX emulator to flathub.
https://github.com/openMSX/openMSX

It provides two desktop icons, one for directly launching openMSX and one for the Catapult launcher.

By default emulation will be limited to the open-source C-BIOS implementation that is included, which only supports cartridge MSX games. For a full emulation experience the user needs to provide his own MSX and/or SVI ROMs.

Please add the openMSX maintainers to the project:
@mthuurne
@MBilderbeek
@andete


rderooy:
bot, build org.openmsx.openMSX

rderooy:
Forgot to mention. There is no wayland permission being asked for, as this is an SDL2 program, and SDL right now does not have client-side-decorations, so if you where to force it to run in wayland mode on Gnome (Mutter), you would not get window decorations. Until that is fixed in SDL2 (they are looking into adding libdecoration for this), I don't want to enable wayland support on SDL programs.

nedrichards:
Is there a reason why the desktop files and appdata aren't part of the upstream release? They are generally useful for all Linux users and are best maintained with the rest of the application. Obviously they can be a patch for now whilst waiting for an upcoming release but we like to see an issue/pull request or similar at least.

rderooy:
Upstream does not regularly release new updates. So for now these are effectively patches.

But it was my intention to try to get them upstream, if there are no objections.

rderooy:
I have created a PR against upstream for the desktop files https://github.com/openMSX/openMSX/pull/1292

Desktop files already exist in the tree, but I modified them so much, that I did not think editing them in the yaml file post-install, or as a patch file made much sense.

The appdata requires more thought. Ideally things would get setup properly during the "make install" phase, but right now it does not work that way.

org.openmsx.openMSX.metainfo.xml
```diff
@@ -0,0 +1,61 @@
...
+	</screenshots>
+	<developer_name>openMSX team</developer_name>
+	<content_rating type="oars-1.1"/>
+	<update_contact>noreply@openmsx.org</update_contact>
+	<recommends>
+		<control>keyboard</control>
+		<control>pointing</control>
+		<control>gameplad</control>
+	</recommends>
+	<mimetypes>
```
bilelmoussaoui, line 55: Set those on the desktop file instead

rderooy: Not quite sure what your referring to, the mimetypes?

They are set in the .desktop file also. I thought the purpose of setting them in the metainfo was such that software center can suggest appropriate apps based on mime type?

barthalion: I think appstream-glib copies the metadata from the `.desktop` file to the resulting appinfo, but I haven't checked.

README.md
```diff
@@ -0,0 +1,77 @@
+# Flatpak for openMSX
```
barthalion, line 1: My usual comment about README's in Flathub repos is that any user-facing information is likely going to be ignored, and things related to development is duplication of the documentation, but you feel strongly about keeping it here, I won't object.

org.openmsx.openMSX.metainfo.xml
rderooy: Hmm.. I have not seen anything like that. And since I would like to keep the same metainfo.xml as I'm working to get upstreamed, I would rather not change it if does not cause problems, or if it is not deprecated.

README.md
rderooy: Yeah, I agree that there is little change that users will find it this way, but the description is just a copy/paste from the XML, and the rest is just a template that I modified a bit, so it was pretty quick to put it together. And I have seen various other flathub repos that have it.

org.openmsx.openMSX.metainfo.xml
```diff
@@ -0,0 +1,61 @@
...
+		<screenshot>https://openmsx.org/images/screenshots/ide.png</screenshot>
+		<screenshot>https://openmsx.org/images/screenshots/tb-underwater.png</screenshot>
+	</screenshots>
+	<developer_name>openMSX team</developer_name>
+	<content_rating type="oars-1.1"/>
+	<update_contact>noreply@openmsx.org</update_contact>
+	<recommends>
+		<control>keyboard</control>
+		<control>pointing</control>
+		<control>gameplad</control>
```
hadess, line 53: Typo here. Not sure whether it's verified by the appstream-glib tool.

rderooy: Thanks, updated. I had checked with the appstream tool, but it did not complain about this.

org.openmsx.openMSX.metainfo.xml
bilelmoussaoui: appstream compose, merges both the desktop file & the appdata file since forever. So there's no need to duplicate that information twice.

rderooy: mimetypes removed

barthalion:
bot, build org.openmsx.openMSX

PR #1182, opened 2019-09-27
Add com.etlegacy.ETL

Eonfge:
This one took the longest, but it likely also the most anticipated. Enemy Territory: Legacy is an actively maintained game, with a vibrant community and a lot of cultural significance. This game had once as many XFire users as World of Warcraft.

The biggest hurdle to presenting this manifest, was the game's reluctance to work with 64bit. There are known issues upstream, and worse, there are compatibility reasons which make a 64bit release undesirable to the majority of users.  

As such, I've made the controversial choice to build this application with the older 18.08/i386 package. While the 19.08/multiarch is already available, getting all dependencies (GLEW, GLU) compiled to 32bit is proving to be a very daunting task, and it's not certain that it will even be compatible with current clients and servers. I have tested this 18.08/i386 version though and found it to be binary compatible.

PR #1900, opened 2020-10-22
Add com.github.iortcw

Eonfge:
This one was on the shelf for a while, but I surprised myself. It was almost done and only minor tweaks were required.

Not to be mistaken for #1182, which I put on hold in the faint hope that upstream can release a new version first.


Eonfge:
bot, build com.github.iortcw

Eonfge:
bot, build com.github.iortcw

Eonfge:
bot, build com.github.iortcw

com.github.iortcw.yaml
```diff
@@ -0,0 +1,85 @@
...
+    dest-filename: iortcw_mp.sh
+    commands:
+    - ls ~/.wolf/main/pak0.pk3 || zenity --error --text "Game data (pak0.pk3 etc.) must be added to:\n~/.var/app/com.github.iortcw/.wolf/main/" --ok-label "Close" --width=400
+    - /app/bin/iowolfmp.x86_64
+  - type: file
+    path: com.github.iortcw.sp.desktop
+  - type: file
+    path: com.github.iortcw.mp.desktop
+  - type: file
+    path: com.github.iortcw.appdata.xml
```
LeandroStanger, line 68: New appdata files should use metainfo instead. https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html

Eonfge: Totally forgot about that

Eonfge:
bot, build com.github.iortcw

Eonfge:
@LeandroStanger Can you tell me why it fails? I added an icon of 128x128 specifically for Flathub, but it doesn't seem to parse it.

com.github.iortcw.yaml
```diff
@@ -0,0 +1,82 @@
...
+runtime: org.freedesktop.Platform
+sdk: org.freedesktop.Sdk
+runtime-version: "20.08"
+command: iortcw_sp.sh
+
+finish-args:
+- --device=dri
+- --socket=wayland
+- --socket=fallback-x11
+- --socket=x11
```
barthalion, line 11:
```suggestion
```

barthalion: Only if it actually works natively on Wayland, that is.

com.github.iortcw.metainfo.xml
```diff
@@ -0,0 +1,79 @@
...
+  <summary>Game engine for Return to Castle Wolfenstein</summary>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>GPL-3.0</project_license>
+
+  <developer_name>iortcw team</developer_name>
+  <update_contact>eonfge@kevindegeling.nl</update_contact>
+
+  <url type="homepage">https://github.com/iortcw/iortcw</url>
+
+  <launchable type="desktop-id">com.github.iortcw.desktop</launchable>
```
barthalion, line 16: The catch here is that it either should point to the launcher you consider main or at least one launcher should match the full ID of the app, otherwise `appstream-compose` does not know where to look for icon information. This file as it is simply does not exist.

barthalion (commented):
Have you reached out to upstream devs about this submission?

com.github.iortcw.yaml
Eonfge: Removed it

com.github.iortcw.metainfo.xml
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2019 Kevin Degeling -->
+
+<component type="desktop">
+  <id>com.github.iortcw</id>
```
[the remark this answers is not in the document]
Eonfge: Changed it. Unless upstream wants something else, but I proposed this slug to them.

com.github.iortcw.metainfo.xml
Eonfge: I've settled on making the singleplayer mode, the primary mode. I already had to compromise a bit, so this works. Just so you know, the original Windows Desktop release from 2001 also had to launchers ;)

Eonfge:
bot, build com.github.iortcw.iortcw

com.github.iortcw.iortcw.yaml
```diff
@@ -0,0 +1,85 @@
+app-id: com.github.iortcw.iortcw
+runtime: org.freedesktop.Platform
+sdk: org.freedesktop.Sdk
+runtime-version: "20.08"
+command: iortcw.sh
+
+finish-args:
+- --device=dri
+- --socket=wayland
+- --socket=x11
```
barthalion, line 10:
```suggestion
- --socket=fallback-x11
```

Eonfge:
bot, build com.github.iortcw.iortcw

fishxz:
the game itself is reporting wrong paths for missing game files.

Eonfge:
bot, build com.github.iortcw.iortcw

---
@fishxz Good catch, I kind of missed the second message for a reason. I restructured the folder-init so that it directly uses the right /flatpak/ path, no workarounds or persists required. Do make sure to move your config a level up since /.wolf/ is no longer required

Eonfge:
bot, build com.github.iortcw.iortcw

fishxz:
> bot, build com.github.iortcw.iortcw
> 
> @fishxz Good catch, I kind of missed the second message for a reason. I restructured the folder-init so that it directly uses the right /flatpak/ path, no workarounds or persists required. Do make sure to move your config a level up since /.wolf/ is no longer required

shouldnt this go to the `data` folder? it otherwise clutters the main app dir.

Eonfge:
> shouldnt this go to the `data` folder? it otherwise clutters the main app dir.

I'm not sure if that adds anything. See, the whole idea of having a `data` and a `config` folder is so that applications can split their resources. same in ~/, where you have the traditional .config and .local

IORTCW doesn't do that though. They still use a `~/.wolf` by default, which is a more old-fashioned way of working. As such, I'm not sure about how to compromise. 

Should we put all .wolf stuff in 
- `~/.var/app/com.github.iortcw.iortcw/` or 
- `~/.var/app/com.github.iortcw.iortcw/wolf/` or
- `~/.var/app/com.github.iortcw.iortcw/.wolf/` ?

I lean towards the first because adding extra folders does nothing for the user

Eonfge:
@fishxz Any comments, else I would just stay with the first option.

fishxz:
Whatever you want. I decided to put everything in data on the q2pro build.

Eonfge:
It's a bit late to try to keep it all compatible..., Yamagi Q2 uses /.yq2 :rofl: 

@barthalion Any recommendations? We're to late to standardize this now, but we might still do better in the future?

Also, upstream is still uncertain about maintaining a shared metainfo. I've asked another person if he's involved with the Arch Package so we might be able to team up with them.

_Off Topic: I saw your presentation. Was fun to listen to, so thanks for sharing. A a contributor/maintainer, I enjoy the work I can do with Flathub and I think that the tools are actually quite good. Certainly less challenging then rpm if you ask me.._

barthalion:
`~/.var/app/*/` dirs usually consist of following subdirectories: `cache  config  dat`

So I would say it would be nice to follow this scheme, but I don't think it ultimately matters, I'm fine with the first option.

(Thanks for your kind words about my presentation!)

Eonfge:
bot, build com.github.iortcw.iortcw

In that case, lets keep what we have. 10 years from now, we'll know what the right answer is.

As far as upstream involvement is concerned. They're on the fence so perhaps it's best to move on. Hopefully, in the future, we (Flathub, Debian and Arch) can team up to maintain the metainfo together.

PR #1910, opened 2020-10-30
add notepad++

thejavascriptman:


org.notepad_plus_plus.NotepadPlusPlus.yml
```diff
@@ -0,0 +1,132 @@
...
+        url: https://savannah.nongnu.org/download/icoutils/icoutils-0.32.3.tar.bz2
+        sha256: 17abe02d043a253b68b47e3af69c9fc755b895db68fdc8811786125df564c6e0
+    post-install:
+      - install -Dm 644 COPYING /${FLATPAK_DEST}/share/licenses/icoutils/LICENSE
+
+  - name: npp
+    buildsystem: simple
+    sources:
+      - type: file
+        path: org.notepad_plus_plus.NotepadPlusPlus.appdata.xml
```
LeandroStanger, line 82: New appdata files should use metainfo instead. https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html

LeandroStanger:
bot, build org.notepad_plus_plus.NotepadPlusPlus

thejavascriptman:
bot, build org.notepad_plus_plus.NotepadPlusPlus

thejavascriptman:
bot, build org.notepad_plus_plus.NotepadPlusPlus

flathub.json
```diff
@@ -0,0 +1,5 @@
+{
+  "only-arches": [
+    "x86_64"
+  ]
+}
```
LeandroStanger, line 5:
```suggestion
{
  "only-arches": ["x86_64"]
}
```

thejavascriptman:
bot, build org.notepad_plus_plus.NotepadPlusPlus

barthalion:
In general, I don't think we want to publish Wine apps directly to Flathub.

thejavascriptman:
@barthalion wouldn't you be willing to at least allow some popular applications? You could always pull it from Flathub if it proves not to be popular.

p1u3o:
Wouldn't it be better to have common Wine extensions or runtimes rather than every Flatpak bundling its own copy of Wine, similar to Java?

FastOSlinux:
notepad ++ is already on the flathub with Phoenicis
Captura de tela de 2020-11-05 12-15-38
Captura de tela de 2020-11-05 12-13-14

npp.sh
```diff
@@ -0,0 +1,127 @@
...
+  if [ ! -z "$SELECTED_UI_FONT" ] && [ "$SELECTED_UI_FONT" != "Default" ]; then
+    set_ui_font "$SELECTED_UI_FONT"
+  fi
+}
+
+# initialize wine and install npp if prefix directory doesn't exist or is empty
+if [ ! -d "$WINEPREFIX/drive_c" ]; then
+  initialize_wine
+  mkdir -p "$NPP_ROOT"
+  tar xf /app/data/npp.tar.xz -C "$NPP_ROOT"
```
tinywrkb, line 124: This feels very wrong. Will every wine app gonna ship like this? Who gonna manage bloat pilling in the user's home? How the app manages updates?  
You can inject a version file to compare against, I've done this [here](https://github.com/tinywrkb/flatpaks/blob/master/com.ti.ccstudio/com.ti.ccstudio.yaml#L55), see also [init script](https://github.com/tinywrkb/flatpaks/blob/master/com.ti.ccstudio/resources/ccstudio#L58), but not in a Flathub distributed package, and it still feels very wrong.  
Wine should have proper Flatpak support instead of making the user's home a mismanaged installation target, especially when the user does not create the Wine prefix, unlike when using Lutris.

PR #1914, opened 2020-11-01
Package for "com.github.marinm.songrec"

marin-m:
Hello,

This is the pull request for the Flatpak package of [SongRec](https://github.com/marin-m/SongRec), a graphical Shazam client I wrote.

It was recently featured on a few Linux-related news websites:

* https://www.linuxuprising.com/2020/10/identify-songs-on-your-linux-desktop.html
* http://www.tuxmachines.org/node/143523
* https://gnulinux.ch/songrec-ein-freier-shazam-client

The tarball downloaded by the Flatpak manifest was generated using this script: [`generate_dist_tarball_for_flathub.sh`](https://github.com/marin-m/SongRec/blob/872553e/packaging/flatpak/generate_dist_tarball_for_flathub.sh)

### DIFF
--- a/com.github.marinm.songrec.json
+++ b/com.github.marinm.songrec.json
@@ -15,8 +15,7 @@
         "--device=dri",
         "--filesystem=host:ro",
         "--filesystem=/tmp:ro",
-        "--filesystem=xdg-config",
-        "--filesystem=xdg-cache",
+        "--filesystem=xdg-data",
         "--socket=pulseaudio"
     ],
     "build-options": {
--- a/com.github.marinm.songrec.json
+++ b/com.github.marinm.songrec.json
@@ -15,7 +15,6 @@
         "--device=dri",
         "--filesystem=host:ro",
         "--filesystem=/tmp:ro",
-        "--filesystem=xdg-data",
         "--socket=pulseaudio"
     ],
     "build-options": {
--- a/com.github.marinm.songrec.json
+++ b/com.github.marinm.songrec.json
@@ -1,0 +1,47 @@
+{
+    "app-id": "com.github.marinm.songrec",
+    "runtime": "org.freedesktop.Platform",
+    "runtime-version": "20.08",
+    "sdk": "org.freedesktop.Sdk",
+    "sdk-extensions" : [
+        "org.freedesktop.Sdk.Extension.rust-stable"
+    ],
+    "command": "songrec",
+    "finish-args": [
+        "--share=network",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--device=dri",
+        "--filesystem=host:ro",
+        "--filesystem=/tmp:ro",
+        "--filesystem=xdg-config",
+        "--filesystem=xdg-cache",
+        "--socket=pulseaudio"
+    ],
+    "build-options": {
+        "append-path" : "/usr/lib/sdk/rust-stable/bin",
+        "env" : {
+            "CARGO_HOME" : "/run/build/songrec/.cargo"
+        }
+    },
+    "modules": [
+        {
+            "name": "rust-flatpak",
+            "buildsystem": "simple",
+            "build-commands": [
+                "cargo --offline fetch --verbose",
+                "cargo --offline build --release --verbose",
+                "install -Dm755 ./target/release/songrec -t /app/bin/",
+                "cp -ra ./packaging/rootfs/usr/share /app/"
+            ],
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://github.com/marin-m/SongRec/releases/download/0.1.1/songrec_tarball_0.1.1_for_flathub_build.tar.gz",
+                    "sha256": "c57c4c9fde94e54a7791a6e8a9964399f5b7ca2cd0552b1ad2f1e485202dfb50"
+                }
+            ]
+        }
+    ]
+}
--- a/com.github.marinm.songrec.json
+++ b/com.github.marinm.songrec.json
@@ -13,8 +13,7 @@
         "--socket=fallback-x11",
         "--socket=wayland",
         "--device=dri",
-        "--filesystem=host:ro",
-        "--filesystem=/tmp:ro",
+        "--filesystem=home:ro",
         "--socket=pulseaudio"
     ],
     "build-options": {
@@ -36,8 +35,8 @@
             "sources": [
                 {
                     "type": "archive",
-                    "url": "https://github.com/marin-m/SongRec/releases/download/0.1.1/songrec_tarball_0.1.1_for_flathub_build.tar.gz",
-                    "sha256": "c57c4c9fde94e54a7791a6e8a9964399f5b7ca2cd0552b1ad2f1e485202dfb50"
+                    "url": "https://github.com/marin-m/SongRec/releases/download/0.1.2/songrec_tarball_0.1.2_for_flathub_build.tar.gz",
+                    "sha256": "ff1f20b223ba49182a95e19b1e4b6cbc43e93a4b9004eb4e22f2cf4314ea52fb"
                 }
             ]
         }
--- a/com.github.marinm.songrec.json
+++ b/com.github.marinm.songrec.json
@@ -27,7 +27,7 @@
     },
     "modules": [
         {
-            "name": "rust-flatpak",
+            "name": "songrec",
             "buildsystem": "simple",
             "build-commands": [
                 "cargo --offline fetch --verbose",


LeandroStanger:
bot, build com.github.marinm.songrec

com.github.marinm.songrec.json
```diff
@@ -0,0 +1,47 @@
...
+    "finish-args": [
+        "--share=network",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--device=dri",
+        "--filesystem=host:ro",
+        "--filesystem=/tmp:ro",
+        "--filesystem=xdg-config",
+        "--filesystem=xdg-cache",
```
bilelmoussaoui, line 19:
```suggestion
```
I don't think you need access to those ?

com.github.marinm.songrec.json
```diff
@@ -0,0 +1,47 @@
...
+    ],
+    "command": "songrec",
+    "finish-args": [
+        "--share=network",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--device=dri",
+        "--filesystem=host:ro",
+        "--filesystem=/tmp:ro",
```
bilelmoussaoui, line 17: Why read access to those?

com.github.marinm.songrec.json
```diff
@@ -0,0 +1,47 @@
...
+    ],
+    "build-options": {
+        "append-path" : "/usr/lib/sdk/rust-stable/bin",
+        "env" : {
+            "CARGO_HOME" : "/run/build/songrec/.cargo"
+        }
+    },
+    "modules": [
+        {
+            "name": "rust-flatpak",
```
bilelmoussaoui, line 30:
```suggestion
            "name": "songrec",
```
(otherwise your CARGO_HOME wouldn't make sense as the build dirs are `/run/build/$moduleName`

com.github.marinm.songrec.json
marin-m: It's for the file picker of the app to work. I personally often use `/tmp` for storing things (such as video soundtracks downloaded using `youtube-dl`...).

com.github.marinm.songrec.json
marin-m: Sorry, I actually store things in `~/.local/share/SongRec`, fixed in commit 74c85ee

nedrichards:
bot, build com.github.marinm.songrec

@marin-m you can use that command as well to generate a new test build when you update the PR

com.github.marinm.songrec.json
barthalion: In general, we vastly prefer permissions to be locked down by default, so here I'd say `xdg-music` and `xdg-download` would be sufficient. It's easy to change via CLI or [GUI](https://flathub.org/apps/details/com.github.tchx84.Flatseal).

barthalion: That being said, any reason you are not using `GtkFileChooserNative` here? Looks like it operates only on single files (for both sound files and CSV export). This way you could drop all filesystem permissions and let the portal handle it.

com.github.marinm.songrec.json
```diff
@@ -0,0 +1,46 @@
...
+    "command": "songrec",
+    "finish-args": [
+        "--share=network",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--device=dri",
+        "--filesystem=host:ro",
+        "--filesystem=/tmp:ro",
+        "--filesystem=xdg-data",
```
barthalion, line 18: This shouldn't be technically needed. Your application should follow `$XDG_DATA_HOME` which flatpak automatically grants access to.

com.github.marinm.songrec.json
marin-m: I didn't know about it. I could, but won't the command-line interface that directly accepts a sound file path from the user be broken?

barthalion: Yeah, but while I don't have any data for this, I don't think many Flathub users run apps from terminal.

com.github.marinm.songrec.json
marin-m: Thanks, fixed in commit c879611

marin-m:
bot, build com.github.marinm.songrec

com.github.marinm.songrec.json
marin-m: I think however that the command line input is an important feature of the app (for example, to enable batch recognition, to the extent accepted by the law) and that users wouldn't be pleased to change distribution channel to use the full version of the app.

barthalion: Something like `flatpak run --file-forwarding appid @@ %f @@` also works from CLI via portals. Ultimately, it's of course your choice as the developer, however `host` and `/tmp` are still too wide. When these are limited, at least two `home:ro`, LGTM.

marin-m: I'm not sure how I could limit `/tmp` to `home`? Should users explicitly be warned not to store things in `/tmp` (which is a common practice especially when using tools such as `youtube-dl` which don't have commonly used GUI frontends) when installing the app through the Flatpak distribution channel?

barthalion: I mean, it is what it is. We as Flathub decided it's better to ship software with limited but not unexpected permissions instead of just adding `--filesystem=host` by default. I'm not denying it's not inconvenient, but we need to draw the line somewhere. I still think that outside of blogs or project website, majority of users will come from https://flathub.org and GNOME/KDE frontends, and so they will never know it's usable from CLI.

marin-m: I guess then that the way to incur the less UX degradation patterns would be implement using `GtkFileChooserNative` for the graphical file picker, and keep a read access to `home` for the command line.

However, when using the `show()` method on `GtkFileChooserNative` (from [`gtk::NativeDialogExt::show`](http://gtk-rs.org/docs/gtk/trait.NativeDialogExt.html#tymethod.show)) the dialog shows a short instant and almost immediately hides, while this is not what the docs says ("_When the user accepts the state of the dialog_ the dialog will be automatically hidden and the “response” signal will be emitted.").

Should this be considered as abnormal behavior and first reported to `gtk-rs`?

barthalion: Yeah, that sounds wrong. cc @bilelmoussaoui as I don't know Rust.

barthalion: For the time being it's fine to go with `--filesystem=home` so the submission is not blocked.

bilelmoussaoui: > I guess then that the way to incur the least UX degradation patterns would be to implement using `GtkFileChooserNative` for the graphical file picker, and keep a read access to `home` for the command line.
> 
> However, when using the `show()` method on `GtkFileChooserNative` (from [`gtk::NativeDialogExt::show`](http://gtk-rs.org/docs/gtk/trait.NativeDialogExt.html#tymethod.show)) the dialog shows a short instant and almost immediately hides, while this is not what the docs says ("_When the user accepts the state of the dialog_ the dialog will be automatically hidden and the “response” signal will be emitted.").
> 
> Should this be considered as abnormal behavior and first reported to `gtk-rs`?

There's probably something you have done wrong. It's very easy to break things with dialogs in general with gtk3. Do you have your code that uses where I can have a look?

Also, gtk application supports opening a specific file, from cli without any file system access. (that's what's run when you are asked to open a file with a specific app)

marin-m: The full content of my input dialog event handler after the modifications is here:

```rust
            let file_chooser = gtk::FileChooserNative::new(
                Some("Select a file to recognize"),
                Some(&window),
                gtk::FileChooserAction::Open,
                Some("_Open"),
                Some("_Cancel")
            );
            
            let file_filter = gtk::FileFilter::new();
            
            file_filter.add_pattern("*.mp3");
            file_filter.add_pattern("*.wav");
            file_filter.add_pattern("*.flac");
            file_filter.add_pattern("*.ogg");
            
            file_chooser.set_filter(&file_filter);
            
            let processing_tx_3 = processing_tx.clone();

            file_chooser.connect_response(move |file_chooser, response| {
                if response == ResponseType::Ok {
                    recognize_file_button.hide();
                    
                    spinner.show();
                    
                    let input_file_path = file_chooser.get_filename().expect("Couldn't get filename");
                    let input_file_string = input_file_path.to_str().unwrap().to_string();
                    
                    processing_tx_3.send(ProcessingMessage::ProcessAudioFile(input_file_string)).unwrap();
                }
            });
            file_chooser.show();
```

Put back in context it's from [here](https://github.com/marin-m/SongRec/blob/0.1.1/src/gui/main_window.rs#L189)

marin-m: I replaced it with the blocking method `run()` instead of the non-blocking method `show()` (which shouldn't be much a problem as the app is multi-threaded) and now this works better. I also restricted the access in the manifest as requested.

marin-m:
bot, build com.github.marinm.songrec

com.github.marinm.songrec.json
bilelmoussaoui: @marin-m your code is a bit wrong, the file path isn't something you should use inside the sandbox as the files you have access to are mounted inside the container as a completely different path. You should use the file uri instead. By using dialog.get_file, you should have a gio::File that would allow you to read/write into the file if needed.

marin-m: I tried to update my code to use `gio::File::new_for_uri` an `gio::File::new_for_path` and `gio::FileExt::load_contents` where relevant, but it gets tricky when it comes to the fact that the audio format decoding library that I use (whose entry point [`rodio::decoder::Decoder::new`](https://docs.rs/rodio/0.12.0/rodio/decoder/struct.Decoder.html)) requires an object implementing both the `Read` trait and the `'static` lifetime, which a native Rust `std::fs::File` seems to implement but a dynamically obtained buffer obtained from GIO does not.

I'm not sure if I can adapt my code without changing the audio decoding library then.

bilelmoussaoui: You can get a stream from your gio file instead of using a load_contents, the input stream implements Read I guess.

marin-m: Not according to the doc: http://gtk-rs.org/docs/gio/struct.FileInputStream.html

The `read()` method on it is implemented by `InputStreamExtManual` which is a distinct trait. http://gtk-rs.org/docs/gio/prelude/trait.InputStreamExtManual.html

marin-m: I think that I could hack this out through reading the file with GIO, then writing to a temporary file, and passing it back to the library, but I don't know whether this is needed as trying the Flatpak build by the buildbot with the current code (https://github.com/flathub/flathub/pull/1914#issuecomment-720511946) and using it to loa a file from `/tmp` or the home directory through the GUI works.

So adapting the code for using URIs is not needed I guess?

bilelmoussaoui: @marin-m, fileInputStream does implement the ExtManual trait (it's just a manual trait that contains code that wasn't generated automatically from the gir annotations), you just need to include it with `use gio::InputStreamExtManual`

marin-m: Yes, but this shouldn't make Rust think that the trait implements `Read`, except if I make a `Read` wrapper myself. Anyway, reading the file-picker selected files inside Flatpak can be observed to work without URIs or GIO, so I guess that I can avoid using these with the sale results and simplify the code.

nedrichards:
Other than the discussion in https://github.com/flathub/flathub/pull/1914#pullrequestreview-521266358 which should be closed out satisfactorily this lgtm.

(non compulsory note: this is a really nicely polished app, other than the horizontal scroll that I got in the results table. That may just be inevitable with the differing lengths of songs but it's a quick bit of user feedback)

com.github.marinm.songrec.json
bilelmoussaoui: Well I don't follow here anymore. If your library takes something that implement Read, InputStream should work.

marin-m: Again, it seems that it clearly doesn't implement `Read` (whichever `Ext` trait I include or not, as the `read()` function implemented by `InputStreamExtManual` doesn't have the same prototype as the one implemented by `std::io::Read` anyway):

```rust
error[E0277]: the trait bound `gio::FileInputStream: std::io::Read` is not satisfied
  --> src/fingerprinting/algorithm.rs:48:72
   |
48 |         let converted_file = rodio::source::UniformSourceIterator::new(decoder, 1, 16000);
   |                                                                        ^^^^^^^ the trait `std::io::Read` is not implemented for `gio::FileInputStream`
   |
   = note: required because of the requirements on the impl of `std::iter::Iterator` for `rodio::Decoder<gio::FileInputStream>`

error[E0277]: the trait bound `gio::FileInputStream: std::io::Seek` is not satisfied
  --> src/fingerprinting/algorithm.rs:48:72
   |
48 |         let converted_file = rodio::source::UniformSourceIterator::new(decoder, 1, 16000);
   |                                                                        ^^^^^^^ the trait `std::io::Seek` is not implemented for `gio::FileInputStream`
   |
   = note: required because of the requirements on the impl of `std::iter::Iterator` for `rodio::Decoder<gio::FileInputStream>`
```

But that using GIO isn't needed to have the application working at all, as the currently produced Flatpak snapshot fully works for the GUI part, including file picking from various parts of the filesystem. So I'm not sure why the current code would be "wrong" or would need to use GIO or URIs.

bilelmoussaoui: https://gtk-rs.org/docs/gio/prelude/trait.InputStreamExtManual.html#method.into_read this gives you something that implements std::io::Read.

bilelmoussaoui: because you're still assuming that you have permission to every file out there. while the home permission works fine for those files in home, it won't work for any file outside of that directory.

marin-m: >https://gtk-rs.org/docs/gio/prelude/trait.InputStreamExtManual.html#method.into_read this gives you something that implements std::io::Read.

Ok, but the audio file format library also requires the concerned object to the thread-safe (to implement the `Send` trait:)

```rust
error[E0277]: `std::ptr::NonNull<glib::object::GObject>` cannot be sent between threads safely
  --> src/fingerprinting/algorithm.rs:45:23
   |
45 |         let decoder = rodio::Decoder::new(gio_file.read::<gio::Cancellable>(None)?.into_read())?;
   |                       ^^^^^^^^^^^^^^^^^^^ `std::ptr::NonNull<glib::object::GObject>` cannot be sent between threads safely
   |
   = help: within `gio::InputStreamRead<gio::FileInputStream>`, the trait `std::marker::Send` is not implemented for `std::ptr::NonNull<glib::object::GObject>`
   = note: required because it appears within the type `glib::object::ObjectRef`
   = note: required because it appears within the type `gio::FileInputStream`
   = note: required because it appears within the type `gio::InputStreamRead<gio::FileInputStream>`
   = note: required by `rodio::Decoder::<R>::new`
```

>because you're still assuming that you have permission to every file out there. while the home permission works fine for those files in home, it won't work for any file outside of that directory.

It seems that it will, at least on `/tmp` which is located outside of the home directory, as it can be observed by going through the following testing procedure:

1. Uninstall any former installation: `flatpak uninstall com.github.marinm.songrec -y`
2. Locally install the latest build: `flatpak install --user https://dl.flathub.org/build-repo/30447/com.github.marinm.songrec.flatpakref -y`
3. Put any decodable sound file (.WAV, .FLAC, .MP3...) in your `/tmp` directory
4. Run the application: `flatpak run com.github.marinm.songrec`
5. Click "Recognize from file"
6. Use the file picker to select the sound file that you have put in `/tmp` (outside of the home directory)

The current Flatpak snapshot (which is limited to `home:ro` by its permissions for reading, as it can be verified when trying to recognize a file from the command line) will then correctly work.

barthalion:
On that note, playing Shazam lures results with bunch of beeps and bops here.

marin-m:
>On that note, playing Shazam lures results with bunch of beeps and bops here.

It's the intended behavior. I tried to make it sound less though by adding harmonics but the lures seemed to be sometimes better recognized without those. (Shazam works by finding frequency peaks in the audio spectrum and sending the corresponding (time, frequency, amplitude) tuples to the remote servers, the beeps map these peaks.)

barthalion:
Ah, I see. I guess I misunderstood what a lure is.

Issue #1023, opened 2019-06-02
Add Adwaita Dark theme


memeplex:
This seems like a gross omission to me, there are now a lot of themes as flatpaks but the official Adwaita-dark is missing. Flatpak apps use Adwaita when the system theme is Adwaita-dark.

memeplex:
See discussions supporting adding Adwaita-dark here:

* In flatpak https://github.com/flatpak/flatpak/issues/2937 they suggested adding support in the freedesktop runtime.

* In freedesktop runtime https://gitlab.com/freedesktop-sdk/freedesktop-sdk/issues/801 they said: it's already there but since there is no simple way to set the variant (let alone pick it up from the sandbox) a workaround would be to add Adwaita-dark to flathub the same than other alternative themes.

* In https://gitlab.gnome.org/GNOME/gnome-tweaks/issues/220 I requested an improvement to this confusing situation.

TingPing:
Added https://github.com/flathub/org.gtk.Gtk3theme.Adwaita-dark

PR #1918, opened 2020-11-03
Add  org.desmume.DeSmuME

linkmauve:
DeSmuME is a [Nintendo DS](https://en.wikipedia.org/wiki/Nintendo_DS) [emulator](https://en.wikipedia.org/wiki/Emulator) written in C++ using GTK.

DeSmuME hasn’t seen a release [since 2015-04-15](http://desmume.org/2015/04/15/desmume-0-9-11-released/) but has been progressing a lot in master since then, hence the `git` source. The maintainer would prefer not to do a release ever again.

### DIFF
--- a/org.desmume.DeSmuME.json
+++ b/org.desmume.DeSmuME.json
@@ -1,0 +1,53 @@
+{
+    "app-id": "org.desmume.DeSmuME",
+    "runtime": "org.freedesktop.Platform",
+    "runtime-version": "20.08",
+    "sdk": "org.freedesktop.Sdk",
+    "command": "desmume",
+    "finish-args": [
+        "--socket=wayland",
+        "--socket=x11",
+        "--socket=pulseaudio",
+        "--device=dri"
+    ],
+    "cleanup": [
+        "/bin/pcap-config",
+        "/include",
+        "/lib/libpcap.a",
+        "/lib/pkgconfig",
+        "/share/man"
+    ],
+    "modules": [
+        "shared-modules/SDL/SDL-1.2.15.json",
+        {
+            "name": "pcap",
+            "buildsystem": "autotools",
+            "config-opts": [
+                "--disable-shared",
+                "--disable-dbus"
+            ],
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://www.tcpdump.org/release/libpcap-1.9.1.tar.gz",
+                    "sha512": "ae0d6b0ad8253e7e059336c0f4ed3850d20d7d2f4dc1d942c2951f99a5443a690f0cc42c6f8fdc4a0ccb19e9e985192ba6f399c4bde2c7076e420f547fddfb08"
+                }
+            ]
+        },
+        {
+            "name": "desmume",
+            "buildsystem": "meson",
+            "subdir": "desmume/src/frontend/posix/",
+            "config-opts": [
+                "-Dfrontend-cli=false"
+            ],
+            "sources": [
+                {
+                    "type": "git",
+                    "url": "https://github.com/TASVideos/desmume",
+                    "commit": "25d71947796144a0a72bbd25818e447102dd1a3c"
+                }
+            ]
+        }
+    ]
+}
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "shared-modules"]
+	path = shared-modules
+	url = https://github.com/flathub/shared-modules


org.desmume.DeSmuME.appdata.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
linkmauve, line 1: This file is temporary, once it is reviewed here I’ll push it upstream.

org.desmume.DeSmuME.appdata.xml
```diff
@@ -0,0 +1,50 @@
...
+
+  <description>
+    <p>
+      DeSmuME is a Nintendo DS emulator.
+    </p>
+  </description>
+
+  <screenshots>
+    <screenshot type="default">
+      <image>https://linkmauve.fr/files/desmume-pokémon.png</image>
```
linkmauve, line 22: This screenshot of DeSmuME at its native resolution is smaller than the recommended size, but I’d like to keep it that way instead of padding it with useless blank pixels, will that be an issue?

org.desmume.DeSmuME.appdata.xml
```diff
@@ -0,0 +1,50 @@
...
+    <release version="0.9.2" date="2009-04-12"/>
+    <release version="0.9.1" date="2009-02-13"/>
+    <release version="0.9" date="2009-01-04"/>
+    <release version="0.8" date="2008-04-22"/>
+    <release version="0.7.3" date="2007-08-12"/>
+    <release version="0.7.2" date="2007-07-10"/>
+    <release version="0.7.1" date="2007-06-13"/>
+    <release version="0.7.0" date="2007-05-11"/>
+    <release version="0.6.0" date="2007-02-04"/>
+    <release version="0.5.0" date="2006-12-23"/>
```
bilelmoussaoui, line 46:
```suggestion
```
There's no need to keep all the versions, especially if they don't a description

org.desmume.DeSmuME.json
```diff
@@ -0,0 +1,68 @@
+{
+    "app-id": "org.desmume.DeSmuME",
+    "runtime": "org.freedesktop.Platform",
+    "runtime-version": "20.08",
+    "sdk": "org.freedesktop.Sdk",
+    "command": "desmume",
+    "rename-desktop-file": "desmume.desktop",
+    "finish-args": [
+        "--filesystem=xdg-config/desmume:create",
```
bilelmoussaoui, line 9: Just use $XDG_DATA_DIRS upstream and remove this permission

org.desmume.DeSmuME.json
```diff
@@ -0,0 +1,68 @@
+{
+    "app-id": "org.desmume.DeSmuME",
+    "runtime": "org.freedesktop.Platform",
+    "runtime-version": "20.08",
+    "sdk": "org.freedesktop.Sdk",
+    "command": "desmume",
+    "rename-desktop-file": "desmume.desktop",
```
bilelmoussaoui, line 7: Ideally, this should be renamed upstream

org.desmume.DeSmuME.json
```diff
@@ -0,0 +1,68 @@
...
+            ],
+            "sources": [
+                {
+                    "type": "file",
+                    "path": "org.desmume.DeSmuME.appdata.xml"
+                }
+            ]
+        }
+    ],
+    "cleanup": [
```
bilelmoussaoui, line 61: Per conventions, cleanup should be between finish-args & modules

org.desmume.DeSmuME.json
```diff
@@ -0,0 +1,68 @@
...
+            "build-commands": [
+                "install -Dm644 -t /app/share/appdata/ org.desmume.DeSmuME.appdata.xml"
+            ],
+            "sources": [
+                {
+                    "type": "file",
+                    "path": "org.desmume.DeSmuME.appdata.xml"
+                }
+            ]
+        }
```
bilelmoussaoui, line 59: I guess this will moved upstream?

org.desmume.DeSmuME.appdata.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>org.desmume.DeSmuME</id>
+  <name>DeSmuME</name>
+  <summary>Nintendo DS emulator</summary>
+  <project_license>GPL-2.0-or-later</project_license>
+  <metadata_license>CC0-1.0</metadata_license>
+  <developer_name>DeSmuME Team</developer_name>
+  <icon type="remote">https://linkmauve.fr/files/desmume-logo.svg</icon>
```
bilelmoussaoui, line 9: if you're already shipping an icon with the app, there's no need to add one unless you want it as a fallback?

org.desmume.DeSmuME.json
linkmauve: Indeed, this is just to make it work for now.

org.desmume.DeSmuME.appdata.xml
```diff
@@ -0,0 +1,50 @@
...
+
```
bilelmoussaoui, line 10: Could use a `<launchable type="desktop-id">org.desmume.DeSmuME.desktop</launchable>` (hence the suggestion to rename the desktop file upstream)

bilelmoussaoui:
bot, build org.desmume.DeSmuME

org.desmume.DeSmuME.json
linkmauve: Why?  Doesn’t this bind-mount `~/.config/desmume` into the sandbox, thus letting the user configure it from the documented directory in all cases?

org.desmume.DeSmuME.json
```diff
@@ -0,0 +1,68 @@
...
+        {
+            "name": "appdata",
+            "buildsystem": "simple",
+            "build-commands": [
+                "install -Dm644 -t /app/share/appdata/ org.desmume.DeSmuME.appdata.xml"
+            ],
+            "sources": [
+                {
+                    "type": "file",
+                    "path": "org.desmume.DeSmuME.appdata.xml"
```
LeandroStanger, line 56: New appdata files should use metainfo instead. https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html

linkmauve:
There, I’ve fixed all of your comments (except for the `~/.config` one) and pushed the metainfo file upstream, as well as fixed the ARM build failure!

org.desmume.DeSmuME.json
bilelmoussaoui: Users can still configure it from `.var/app/$app-id/config`.

bilelmoussaoui:
bot, build org.desmume.DeSmuME

org.desmume.DeSmuME.json
linkmauve: This is much less user-friendly though. :/

If possible I’d like to keep `~/.config/desmume` as the user configuration directory, to avoid having to modify all of the documentation everywhere.

That’s also what [I did in poezio](https://github.com/flathub/io.poez.Poezio/blob/master/io.poez.Poezio.json#L7) (where it is arguably even more important, since users are *expected* to modify these files manually).

linkmauve:
I’ve replaced `--socket=fallback-x11` with `--socket=x11` because SDL 1.2 requires X11 access atm. I’ll work upstream on moving off from this library.

After testing, every feature seems to work properly, the `GtkFileChooserNative` correctly uses the portals to open files, etc. :)

The only issues so far were with X11, and also that my dark variant theme isn’t respected. Do I need to do something special for that? I’ve configured it in `~/.config/gtk-3.0/settings.ini` on the host system.

org.desmume.DeSmuME.json
barthalion: We made an exception for poezio as it's terminal app. Graphical apps should respect XDG_CONFIG_HOME.

barthalion:
Your theme needs to be available on Flathub – then flatpak will install it automatically.

barthalion:
Also it looks like there's something wrong with desktop file as it uses stock icon here. Probably `Icon=` should set to app id.

org.desmume.DeSmuME.json
linkmauve: Alright, I’ve removed this line.

linkmauve:
There, I’ve removed the `XDG_CONFIG_HOME` override and renamed the icon upstream!

> Your theme needs to be available on Flathub – then flatpak will install it automatically.

It is Adwaita’s dark variant, do I need to do anything special for it?

barthalion:
bot, build org.desmume.DeSmuME

barthalion:
> There, I’ve removed the `XDG_CONFIG_HOME` override and renamed the icon upstream!
> 
> > Your theme needs to be available on Flathub – then flatpak will install it automatically.
> 
> It is Adwaita’s dark variant, do I need to do anything special for it?

Hm, I don't remember the details. I thought it should work just fine; https://github.com/flathub/flathub/issues/1023 is one of the issues reported in the past. Maybe you'll find some hints there.

barthalion:
Somehow it's still not exported on my system but I'm not sure why; appstream-compose finds SVG file and I see PNG files in app-info dir. Does it work for you?

linkmauve:
I’ve installed GNOME and indeed it doesn’t display the icon. I’ve then compared with the build of poezio, and the main difference is that I’m using `/app/share/pixmaps/` in DeSmuME and `/app/share/icons/hicolor/scalable/apps/` in poezio. I’ll try doing that shortly.

linkmauve:
There, it works!

barthalion:
bot, build org.desmume.DeSmuME

barthalion:
LGTM. @linkmauve Should we give someone beside you write access to Flathub repo?

linkmauve:
I’m the only one atm caring about Linux support in this emulator, but if someone else from the project comes and ask for it please just grant them access. :)

PR #1931, opened 2020-11-11
add com.github.emmanueltouzery.projectpad

emmanueltouzery:
i'm submitting a new application for flathub.

I am the author, it's hosted at https://github.com/emmanueltouzery/projectpad2

Thank you!

### DIFF
--- a/com.github.emmanueltouzery.projectpad.json
+++ b/com.github.emmanueltouzery.projectpad.json
@@ -46,11 +46,10 @@
     },
     {
       "name": "sqlcipher-4.4.1",
-      "only-arches": [],
-      "buildsystem": "simple",
-      "build-commands": [
-        "./configure --enable-tempstore=yes CFLAGS=\"-DSQLITE_HAS_CODEC\" LDFLAGS=\"-lcrypto\" --prefix=/app",
-        "make install"
+      "config-opts": [
+        "--enable-tempstore=yes",
+        "CFLAGS=-DSQLITE_HAS_CODEC",
+        "LDFLAGS=-lcrypto"
       ],
       "sources": [
         {
--- a/com.github.emmanueltouzery.projectpad.json
+++ b/com.github.emmanueltouzery.projectpad.json
@@ -33,11 +33,9 @@
     },
     {
       "name": "tcl-8.6.10",
-      "only-arches": [],
-      "buildsystem": "simple",
-      "build-commands": [
-        "cd unix && ./configure --prefix=/app --enable-shared=no",
-        "cd unix && make install"
+      "subdir": "unix",
+      "config-opts": [
+        "--enable-shared=no"
       ],
       "sources": [
         {
--- a/com.github.emmanueltouzery.projectpad.json
+++ b/com.github.emmanueltouzery.projectpad.json
@@ -77,7 +77,7 @@
         {
           "type": "git",
           "url": "https://github.com/emmanueltouzery/projectpad2.git",
-          "commit": "290400f5d7784fafff5b0dc4b25674bb7d85ca94"
+          "commit": "e64e2b365999a6b0db7ba78ea6e704843e58436e"
         },
 	"cargo-sources.json",
 	{
--- a/com.github.emmanueltouzery.projectpad.json
+++ b/com.github.emmanueltouzery.projectpad.json
@@ -1,0 +1,99 @@
+{
+  "app-id": "com.github.emmanueltouzery.projectpad",
+  "runtime": "org.freedesktop.Platform",
+  "runtime-version": "19.08",
+  "sdk": "org.freedesktop.Sdk",
+  "sdk-extensions": ["org.freedesktop.Sdk.Extension.rust-stable"],
+  "command": "projectpad",
+  "finish-args": [
+    "--share=ipc",
+    "--socket=fallback-x11",
+    "--socket=wayland",
+    "--device=dri",
+    "--filesystem=home",
+    "--share=network",
+    "--talk-name=org.freedesktop.secrets"
+  ],
+  "build-options": {
+    "append-path": "/usr/lib/sdk/rust-stable/bin",
+    "env": {
+      "CARGO_HOME": "/run/build/projectpad/cargo"
+    }
+  },
+  "modules": [
+    {
+      "name": "gtksourceview-3.24.11",
+      "only-arches": [],
+      "buildsystem": "simple",
+      "build-commands": [
+        "./configure --prefix=/app",
+        "make install"
+      ],
+      "sources": [
+        {
+          "type": "archive",
+          "sha256": "691b074a37b2a307f7f48edc5b8c7afa7301709be56378ccf9cc9735909077fd",
+          "url": "http://ftp.gnome.org/pub/gnome/sources/gtksourceview/3.24/gtksourceview-3.24.11.tar.xz"
+        }
+      ]
+    },
+    {
+      "name": "tcl-8.6.10",
+      "only-arches": [],
+      "buildsystem": "simple",
+      "build-commands": [
+        "cd unix && ./configure --prefix=/app --enable-shared=no",
+        "cd unix && make install"
+      ],
+      "sources": [
+        {
+          "type": "archive",
+          "sha256": "5196dbf6638e3df8d5c87b5815c8c2b758496eb6f0e41446596c9a4e638d87ed",
+          "url": "https://prdownloads.sourceforge.net/tcl/tcl8.6.10-src.tar.gz"
+        }
+      ]
+    },
+    {
+      "name": "sqlcipher-3.4.2",
+      "only-arches": [],
+      "buildsystem": "simple",
+      "build-commands": [
+        "./configure --enable-tempstore=yes CFLAGS=\"-DSQLITE_HAS_CODEC\" LDFLAGS=\"-lcrypto\" --prefix=/app",
+        "make install"
+      ],
+      "sources": [
+        {
+          "type": "archive",
+          "sha256": "69897a5167f34e8a84c7069f1b283aba88cdfa8ec183165c4a5da2c816cfaadb",
+          "url": "https://github.com/sqlcipher/sqlcipher/archive/v3.4.2.tar.gz"
+        }
+      ]
+    },
+    {
+      "name": "projectpad",
+      "buildsystem": "simple",
+      "build-commands": [
+	"mkdir projectpad/fontawesome-5.13.0 && mv svgs projectpad/fontawesome-5.13.0",
+        "cargo --offline fetch --manifest-path Cargo.toml --verbose",
+        "cargo --offline build --release --verbose",
+        "install -Dm755 ./target/release/projectpad -t /app/bin/",
+        "install -Dm644 ./flatpak/${FLATPAK_ID}.metainfo.xml -t /app/share/metainfo/",
+        "install -Dm644 ./projectpad/${FLATPAK_ID}.desktop -t /app/share/applications/",
+        "install -Dm644 ./projectpad/${FLATPAK_ID}.svg -t /app/share/icons/hicolor/scalable/apps/"
+      ],
+      "sources": [
+        {
+          "type": "git",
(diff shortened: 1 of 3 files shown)


com.github.emmanueltouzery.projectpad.json
```diff
@@ -0,0 +1,99 @@
+{
+  "app-id": "com.github.emmanueltouzery.projectpad",
+  "runtime": "org.freedesktop.Platform",
+  "runtime-version": "19.08",
```
barthalion, line 4:
```suggestion
  "runtime-version": "20.08",
```

com.github.emmanueltouzery.projectpad.json
```diff
@@ -0,0 +1,99 @@
...
+  "sdk": "org.freedesktop.Sdk",
+  "sdk-extensions": ["org.freedesktop.Sdk.Extension.rust-stable"],
+  "command": "projectpad",
+  "finish-args": [
+    "--share=ipc",
+    "--socket=fallback-x11",
+    "--socket=wayland",
+    "--device=dri",
+    "--filesystem=home",
```
barthalion, line 13: Why does need access to home?

barthalion:
bot, build com.github.emmanueltouzery.projectpad

com.github.emmanueltouzery.projectpad.json
emmanueltouzery: * it stores a database in the home folder which can then be used by other applications (concretely, the sister ppcli tool)
* you can add files (for instance keys) in the application's database
* you can associate pictures for projects, meaning pick files on the local hard drive

barthalion:
sqlcipher failed to build for aarch64, maintainers usually do something like this to fix it: https://github.com/flathub/org.gimp.GIMP/blob/b0aa516ffb389285df99ebe55ca090e083168d33/org.gimp.GIMP.json#L245

com.github.emmanueltouzery.projectpad.json
emmanueltouzery: thank you! i otherwise already have another flatpak and wanted to keep that runtime up-to-date but couldn't find any info about the latest version. In https://docs.flatpak.org/en/latest/available-runtimes.html they talk about 19.08 and I couldn't find a way to list available runtimes in the flatpak CLI? Otherwise I tested locally and things are OK also with that newer runtime.

emmanueltouzery:
hmm i have no idea about that aarch64 issue. i'll check tonight.

com.github.emmanueltouzery.projectpad.json
barthalion: I think for 2 and 3 it would be better to use GtkFileChooserNative which uses portal for all files operations. This way there's no need to specify filesystem permissions.

As for 1, flatpak grants write access to `$XDG_DATA_HOME` by default.

com.github.emmanueltouzery.projectpad.json
```diff
@@ -0,0 +1,99 @@
...
+      "buildsystem": "simple",
+      "build-commands": [
+        "./configure --prefix=/app",
+        "make install"
+      ],
+      "sources": [
+        {
+          "type": "archive",
+          "sha256": "691b074a37b2a307f7f48edc5b8c7afa7301709be56378ccf9cc9735909077fd",
+          "url": "http://ftp.gnome.org/pub/gnome/sources/gtksourceview/3.24/gtksourceview-3.24.11.tar.xz"
```
bilelmoussaoui, line 36: use download.gnome.org instead

com.github.emmanueltouzery.projectpad.json
```diff
@@ -0,0 +1,99 @@
...
+  },
+  "modules": [
+    {
+      "name": "gtksourceview-3.24.11",
+      "only-arches": [],
+      "buildsystem": "simple",
+      "build-commands": [
+        "./configure --prefix=/app",
+        "make install"
+      ],
```
bilelmoussaoui, line 31:
```suggestion
```
the default build system should be fine here.

com.github.emmanueltouzery.projectpad.json
```diff
@@ -0,0 +1,99 @@
...
+      ]
+    },
+    {
+      "name": "tcl-8.6.10",
+      "only-arches": [],
+      "buildsystem": "simple",
+      "build-commands": [
+        "cd unix && ./configure --prefix=/app --enable-shared=no",
+        "cd unix && make install"
+      ],
```
bilelmoussaoui, line 47: Same here, just use subdir: unix.

com.github.emmanueltouzery.projectpad.json
```diff
@@ -0,0 +1,99 @@
...
+      ]
+    },
+    {
+      "name": "sqlcipher-3.4.2",
+      "only-arches": [],
+      "buildsystem": "simple",
+      "build-commands": [
+        "./configure --enable-tempstore=yes CFLAGS=\"-DSQLITE_HAS_CODEC\" LDFLAGS=\"-lcrypto\" --prefix=/app",
+        "make install"
+      ],
```
bilelmoussaoui, line 63: Same here & pass config-options.

com.github.emmanueltouzery.projectpad.json
barthalion: (FTR ftp.gnome.org redirects to download.gnome.org since a week or so.)

bilelmoussaoui: Nice! it would be nice to avoid using it on the newly written manifests though

com.github.emmanueltouzery.projectpad.json
emmanueltouzery: I tried this:

```json
      "config-opts": [
        "--enable-tempstore=yes",
        "CFLAGS=\"-DSQLITE_HAS_CODEC\"",
        "LDFLAGS=\"-lcrypto\""
]
```
but it failed, saying that the C compiler cannot create executables. I think the CFLAGS and LDFLAGS must be passed in another way, but I'm not sure how?

emmanueltouzery:
for the aarch64 issue, i upgrade the sqlcipher version -- i was using 3 since i'm using the sqlcipher3 format anyway for now, but if it's too old to build correctly on all platforms, better to upgrade. now i'll look at the permissions and the native file picker thing.

com.github.emmanueltouzery.projectpad.json
emmanueltouzery: I've looked at $XDG_DATA_HOME: it's `~/.var/app/com.github.emmanueltouzery.projectpad/data` when running in flatpak, but `~/.local/share/` when running native. I have tried many combinations (mostly huge hacks) so that I could keep supporting running the app as native as well as in flatpak, while having my external app (which is always native) always finding that database, and I failed. The problem is that the app which is not running in flatpak must guess the location used by flatpak, keeping in mind the other may not actually be using flatpak. It just complicated everything for me, I think my use-case doesn't fit in this case. Having a consistent and standard folder (and easily findable also from outside of flatpak) where data is stored is important in this case for me. So I'd just leave the home permission.

emmanueltouzery:
bot, build com.github.emmanueltouzery.projectpad

emmanueltouzery:
I triggered a build to see whether the aarch64 issues are fixed

com.github.emmanueltouzery.projectpad.json
barthalion: I understand that it's pretty much `if $XDG_DATA_HOME then $XDG_DATA_HOME else ~/.local/share`. @bilelmoussaoui is it how GNOME apps usually handle it?

emmanueltouzery: My problem is that I split my app in two in part due to flatpak. In reality, it's impossible to do all that I want to do with this app within flatpak. So I made two apps: this one, which has a GUI and doesn't need to run on the host, and a CLI statically linked companion app that _must_ run on the host. And they share the database. And I'd like to support the case of the flatpak app being run on the metal too. That means that the CLI app must look in ~/.var/com.github... where the flatpak app saves, as well as ~/. local/share. And if a user wants to contribute to the app and run it on the host, I'd also like for the db to be found. It gets really messy. It is possible, but I got far enough that I thought that this system was not designed for that.

emmanueltouzery: Maybe if there was a not-too-hacky way for the app on the host to check whether flatpak is installed, and the com.github... app is installed, and what is the home sir of that app on the host.. but I doubt there is such an API or that you're supposed to do that

emmanueltouzery: ok, the current version right now doesn't work well. I used to store in $HOME/.projectpad and that worked well from outside flatpak and from the CLI app to. now I changed it to $XDG_DATA_HOME and that broke it, because it's different in the flatpak and out of the flatpak -- even if the flatpak has home permissions. I'll think about it a little more. The PR as it is is not OK.

com.github.emmanueltouzery.projectpad.json
emmanueltouzery: @bilelmoussaoui any suggestion?

bilelmoussaoui: Those should be set as build-options. Check how other apps do that

emmanueltouzery: I managed to do that in the other cases for this app, but this one eludes me. but ok, i'll try more.

com.github.emmanueltouzery.projectpad.json
emmanueltouzery: ok i think i have a work, if hackish, solution, without home privileges. i think it'll work out, i'm currently researching how to specify file extension filters in file dialogs through the portals. i'll write here when this is done (and presumably push a version of the json that does not request home privileges)

com.github.emmanueltouzery.projectpad.json
emmanueltouzery: ok, I finally made it work. I couldn't make file extension filtering, but I added a check for wrong file extensions in the app. The latest manifest doesn't ask for the home permissions anymore.

com.github.emmanueltouzery.projectpad.json
emmanueltouzery: ok, I finally made it work. I couldn't make file extension filtering, but I added a check for wrong file extensions in the app. The latest manifest doesn't ask for the home permissions anymore.

com.github.emmanueltouzery.projectpad.json
emmanueltouzery: managed to make also that last case work. the problem were the quotes (even escaped), removing them fixed it.

emmanueltouzery:
I have fixed all the review comments

emmanueltouzery:
i'll trigger a final build now, as far as i'm concerned the app is good to go

emmanueltouzery:
bot, build com.github.emmanueltouzery.projectpad

emmanueltouzery:
Ok I found another issue with giving up the "home" rights. The application works fine, but then you can't sync the database through dropbox, which is an important feature. I checked an it seems that dropbox doesn't allow syncing outside of their folder, which is understandable:
https://www.dropboxforum.com/t5/Dropbox-files-folders/Can-I-sync-folders-outside-the-Dropbox-Folder/td-p/202133

I guess not all users of projectpad would need that sync, but those who do need it would be stuck without home access. So I think I'll put back the home access due to that. We can discuss options if you have other ideas, but this syncing between computers seems an important feature to me. Of course there are also other options besides dropbox, but I expect all the popular ones would require you to put the database in a specific locations. Alternatives to that would be.. a cron job copying the database file regularly? but that's considerably less useful than real-time sync...

emmanueltouzery:
ok, false alert, everything is fine, this works:

    flatpak override com.github.emmanueltouzery.projectpad --filesystem=/home/emmanuel/Dropbox

I'll now revert my previous commit, restricting again the access to forbid home.

nedrichards:
bot, build com.github.emmanueltouzery.projectpad

barthalion:
bot, build com.github.emmanueltouzery.projectpad

PR #1933, opened 2020-11-14
Add nz.mega.MEGAsync

tim77:
Requested for long time ago [here](https://github.com/meganz/MEGAsync/issues/479), [here](https://github.com/meganz/MEGAsync/issues/219) and [here](https://github.com/meganz/MEGAsync/issues/88). Expect upstream involvement.

TODO:

- [x] Add AppData manifest
- [x] More cleanup (`/man` probably)


LeandroStanger:
bot, build nz.mega.MEGAsync

tim77:
Please wait a little bit. :) I'll write quick and dirty appdata manifest first and try to send upstream it first.

tim77:
bot, build nz.mega.MEGAsync

tim77:
bot, build nz.mega.MEGAsync

nz.mega.MEGAsync.yml
```diff
@@ -0,0 +1,169 @@
...
+    post-install:
+      # Automatically add build version and date into AppData
+      - |
+        set -e
+        VERSION=$(git describe --tags | sed -e 's/^[vV]//' -e 's/_Linux$//')
+        DATE=$(date -d @$(git show -s --format=%ct) +'%Y-%m-%d')
+        sed \
+          -e "s/@VERSION@/${VERSION}/g" \
+          -e "s/@DATE@/${DATE}/g" \
+          -i /app/share/metainfo/${FLATPAK_ID}.metainfo.xml
```
eszlari, line 63: I think renaming  the file to `*.metainfo.xml.in` and generating a new file based on it, would be cleaner.

nz.mega.MEGAsync.metainfo.xml
```diff
@@ -0,0 +1,58 @@
...
+
+  <content_rating type="oars-1.0">
+  </content_rating>
+
+  <summary>
+    Easy automated syncing between your computers and your MEGA Cloud Drive.
+  </summary>
+
+  <description>
+    <p>Easy automated syncing between your computers and your MEGA cloud drive.</p>
```
barthalion, line 23: This needs a better description.

nz.mega.MEGAsync.metainfo.xml
```diff
@@ -0,0 +1,58 @@
...
+<!-- Copyright 2020 Artem Polishchuk <ego.cordatus@gmail.com> -->
+<component type="desktop-application">
+  <id>nz.mega.MEGAsync</id>
+  <metadata_license>CC0-1.0</metadata_license>
+
+  <!-- MEGAsync is under a proprietary license, except the SDK which is BSD -->
+  <project_license>LicenseRef-proprietary</project_license>
+
+  <name>MEGAsync</name>
+  <provides>
```
barthalion, line 11: I don't think it's needed. Adding `launchable` is, though.

nz.mega.MEGAsync.metainfo.xml
tim77: Fixed (i hope). BTW if to much paragraphs validate tool also complain so i hope this would enough.
`• style-invalid         : Too many <p> tags for a good description [16/10]
Validation of files failed`

nz.mega.MEGAsync.metainfo.xml
tim77: Replaced with `launchable`. Should i also rename .desktop file with proper UUID.desktop and try to send upstream? Flatpak this could fix automatically but this could ruin other non-flatpak build i suppose. Not so big deal though.

nz.mega.MEGAsync.yml
tim77: This is definitely what i want to do as well and what should upstreamed.

nz.mega.MEGAsync.metainfo.xml
barthalion: `rename-desktop-file` & co exists because people usually don't follow the specification. Ideally it should be fixed upstream.

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "skip-arches": ["aarch64"]
```
hfiguiere, line 2: what's the rationale for skipping aarch64?

tim77: [Something](https://github.com/flathub/flathub/pull/1933#issuecomment-727269231) with freeimage on this arch. I've noticed some other distros skip building on aarch64 as well. Would be nice to investigate and fix though.

barthalion:
bot, build nz.mega.MEGAsync

nz.mega.MEGAsync.metainfo.xml
tim77: `• tag-missing           : <name> is not present`
https://flathub.org/builds/#/builders/48/builds/3873

Probably we should back name here?

tim77:
bot, build nz.mega.MEGAsync

tim77:
Queued test build for nz.mega.MEGAsync.

tim77:
bot, build nz.mega.MEGAsync

barthalion:
Any news on upstream involvement?

tim77:
Unfortunately not yet. In early days devs was sceptical because lack of Qt support but that's in the past. Now KDE runtime polished and hundreds of functional Qt apps on Flathub proves it. How awesome is to have one fully tested and polished build instead of all of this:
image

nz.mega.MEGAsync.yml
```diff
@@ -0,0 +1,166 @@
+id: nz.mega.MEGAsync
+sdk: org.kde.Sdk
+runtime: org.kde.Platform
+runtime-version: '5.15'
+finish-args:
+  - --filesystem=~/MEGA:create
```
tinywrkb, line 6: Would be nice if this could be avoided. creating files & folders directly under ~ is something that some users do not like.  In my system ~ is read-only so this gonna fail.  
I believe also changing that location via the app's settings will fail due to the fact `DontUseNativeDialog` is set as false (`src/MEGASync/{gui/MultiQFileDialog,MegaApplication}.cpp`) and selecting folders via XDG File portal does not work correctly (or at least with this app).
Can this location be patched in the app to default to another location? like `xdg-documents/MEGA` or `xdg-public-share/MEGA`

tim77: > creating files & folders directly under ~ is something that some users do not like.

Not working app out of box is that something that most users do not like even more.

> Can this location be patched in the app to default to another location? like xdg-documents/MEGA or xdg-public-share/MEGA

Both locations looks like not proper place for this app and also insecure.

I would stick with upstream defaults.

tim77: But for `DontUseNativeDialog` - i would try to experiment locally with this as soon as i can. Would be nice to make portals works for sure. Thanks for tip.

nz.mega.MEGAsync.yml
```diff
@@ -98,15 +98,16 @@ modules:
...
           - DESTDIR=/app
           - INCDIR=/app/include
           - INSTALLDIR=/app/lib
         sources:
-          - type: archive
-            url: https://sourceforge.net/code-snapshots/svn/f/fr/freeimage/svn/freeimage-svn-r1859-FreeImage-trunk.zip
-            sha256: 29f01d45ac81246edfcb3bc0c66f996828f7b18624787e2f5e0558effe79d5f3
+          - type: svn
+            url: https://svn.code.sf.net/p/freeimage/svn/
+            revision: r1859
```
hfiguiere, line 110: If this is to build on `aarch64`, there is an easier fix using the release.

All you need is to add
```yaml
  build-options:
      cflags: -fPIC
      cxxflags: -fPIC
```

(maybe conditional for aarch64)

tim77:
bot, build nz.mega.MEGAsync

hfiguiere:
Google Breakpad is apparently not supported on aarch64. MuseScore did disable it.

nz.mega.MEGAsync.yml
tim77: @tinywrkb you right, platform-native file dialog doesn't work correctly and would be really nice if @grulja help us at least to point where the trouble - in portals/runtime or this is 'MEGA' bug?

I've also found your treasure [here](https://github.com/tinywrkb/flatpaks/blob/master/nz.mega.megasync/nz.mega.megasync.yaml) so i've enabled for debugging purposes platform-native file dialog in `2d9dcdf` for now.

tinywrkb: @tim77 I should have limited my changes to affect only the folder selection setting as we still want to be able to use the XDG File portal file selection for upload/download.
Also, thanks for the attribution

tim77: We will able to use **native** XDG File portal if we will fix it entirely (in app itself, probaly). :) This will be win-win solution. You able to use file selection now only because of `--filesystem=home` in your build which if far from ideal.

barthalion:
bot, build nz.mega.MEGAsync

barthalion:
/merge

@tim77 Please poke us somewhere if upstream expresses interest in maintaining the app.

tim77:
I realized that devs might not have received any notification because it works like that for collaborators on GH. @polmr @javiserrano @lpetrov @jnavarrom @khaleddaifallah-mega - please join. I believe Flatpak is the proper and right tool to make best MEGA Linux build. The only one [issue](https://github.com/flathub/nz.mega.MEGAsync/commit/2d9dcdf79c60cf7ad03d74f877b4d7fb71cc9809) at this moment.

PR #1946, opened 2020-11-22
Add org.soundconverter.SoundConverter

jmaibaum:


org.soundconverter.SoundConverter.yml
```diff
@@ -0,0 +1,47 @@
...
+runtime: org.gnome.Platform
+runtime-version: '3.38'
+sdk: org.gnome.Sdk
+rename-appdata-file: soundconverter.appdata.xml
+rename-desktop-file: soundconverter.desktop
+rename-icon: soundconverter
+command: soundconverter
+finish-args:
+  # X11
+  - --socket=x11
```
barthalion, line 11:
```suggestion
  - --socket=fallback-x11
```

org.soundconverter.SoundConverter.yml
```diff
@@ -0,0 +1,47 @@
...
+  # Wayland
+  - --socket=wayland
+  # Users should be able to edit files anywhere in their home
+  - --filesystem=home
```
barthalion, line 15: Unless it's something insists on, I'd prefer `xdg-music`, `xdg-download`, `/media` and `/run/media`.

org.soundconverter.SoundConverter.yml
```diff
@@ -0,0 +1,47 @@
...
+modules:
+  - name: distutils-extra
+    buildsystem: simple
+    build-commands:
+      - python3 setup.py install --prefix=/app
+    sources:
+      - type: git
+        url: https://salsa.debian.org/python-team/packages/python-distutils-extra.git
+        tag: '2.45'
+  - name: intltool
```
barthalion, line 30: Available in shared-modules: https://github.com/flathub/shared-modules/

soundconverter.appdata.xml.patch
```diff
@@ -0,0 +1,27 @@
+diff --git a/data/soundconverter.appdata.xml b/data/soundconverter.appdata.xml
```
barthalion, line 1: Should be upstreamed.

org.soundconverter.SoundConverter.yml
jmaibaum: For now, I have changed it according to your suggestion, since the app does not require any specific location to work correctly.

Though, from my own perspective as user of flatpak apps, I often struggle with flatpaks providing some generic functionality, which then only allow you to open files stored in specific directories. I understand the idea of limiting access only to the bare minimum, but for generic apps like SoundConverter, this often disrupts user workflows: They have to move/copy files into some "allowed path", do the thing inside the application, and then move/copy them back to their original location again.

Personally, I would prefer `home` as it gives users a maximum of flexibility which of their sound files they are allowed to convert. I am often converting sound files which are not necessarily music, nor downloaded from the internet, nor coming from external media.

soundconverter.appdata.xml.patch
jmaibaum: See https://github.com/kassoulet/soundconverter/pull/46

barthalion:
bot, build org.soundconverter.SoundConverter

(You can trigger new builds on your own!)

barthalion:
LGTM. Have you asked upstream developer if they would like to maintain their app on Flathub?

jmaibaum:
>Have you asked upstream developer if they would like to maintain their app on Flathub?

Not yet, since there are still a few issues with this flatpak which I do not yet understand. For example, I don't get a translated user interface, even though the locale extension looks OK on first sight.

I will keep this in draft mode for a few more days at least, until I will have more time to look into this.

Thanks for the hint how to build test versions!

nedrichards:
bot, build org.soundconverter.SoundConverter

jmaibaum:
bot, build org.soundconverter.SoundConverter

jmaibaum:
bot, build org.soundconverter.SoundConverter

jmaibaum:
The latest commit includes a patch to fix the l10n issue I was having before, so I now consider org.soundconverter.SoundConverter ready for flathub.

Patch is pending upstream here: https://github.com/kassoulet/soundconverter/pull/49

jmaibaum:
Force-pushed to remove the patch again, which was just included upstream (see PR linked above).

jmaibaum:
bot, build org.soundconverter.SoundConverter

PR #1949, opened 2020-11-25
Add org.gnome.BreakTimer

dylanmccall:
This adds GNOME Break Timer to Flathub:

https://gitlab.gnome.org/GNOME/gnome-break-timer/

It needs to talk with `org.gnome.Shell`, `org.gnome.Mutter.IdleMonitor`, and `org.gnome.ScreenSaver` in order to keep track of how much someone is currently using their computer. It needs to talk with `org.gnome.ControlCenter` to provide a shortcut to the Applications settings panel in the event that the application hasn't been given permission to run in the background.

I opted to not use intltool from shared-modules in order to keep this manifest file consistent with the one upstream (https://gitlab.gnome.org/GNOME/gnome-break-timer/-/blob/master/build-aux/flatpak/org.gnome.BreakTimer.Devel.json). Please let me know if you would prefer I use shared-modules there :)

### DIFF
--- a/org.gnome.BreakTimer.json
+++ b/org.gnome.BreakTimer.json
@@ -72,7 +72,7 @@
                 {
                     "type" : "git",
                     "url" : "https://gitlab.gnome.org/GNOME/gnome-break-timer.git",
-                    "commit": "8625f78251cf51eb7e3d930a1d49f1c7dd4c757d"
+                    "tag": "2.0.2"
                 }
             ]
         }
--- a/org.gnome.BreakTimer.json
+++ b/org.gnome.BreakTimer.json
@@ -1,0 +1,87 @@
+{
+    "id" : "org.gnome.BreakTimer",
+    "runtime" : "org.gnome.Platform",
+    "runtime-version" : "3.38",
+    "sdk" : "org.gnome.Sdk",
+    "command" : "gnome-break-timer-settings",
+    "finish-args" : [
+        "--share=ipc",
+        "--socket=x11",
+        "--socket=wayland",
+        "--socket=pulseaudio",
+        "--talk-name=org.gnome.ControlCenter",
+        "--talk-name=org.gnome.Shell",
+        "--talk-name=org.gnome.Mutter.IdleMonitor",
+        "--talk-name=org.gnome.ScreenSaver",
+        "--talk-name=org.freedesktop.Notifications"
+    ],
+    "build-options" : {
+        "cflags" : "-O2 -g",
+        "cxxflags" : "-O2 -g",
+        "env" : {
+            "V" : "1"
+        }
+    },
+    "modules" : [
+        {
+            "name" : "libcanberra",
+            "sources" : [
+                {
+                    "type" : "archive",
+                    "url" : "http://0pointer.de/lennart/projects/libcanberra/libcanberra-0.30.tar.xz",
+                    "sha256" : "c2b671e67e0c288a69fc33dc1b6f1b534d07882c2aceed37004bf48c601afa72"
+                }
+            ],
+            "config-opts" : [
+                "--disable-alsa",
+                "--disable-null",
+                "--disable-oss"
+            ]
+        },
+        {
+            "name" : "gsound",
+            "sources" : [
+                {
+                    "type" : "git",
+                    "url" : "https://gitlab.gnome.org/GNOME/gsound.git",
+                    "branch" : "master"
+                }
+            ]
+        },
+        {
+            "name" : "intltool",
+            "sources" : [
+                {
+                    "type" : "archive",
+                    "url" : "https://launchpad.net/intltool/trunk/0.51.0/+download/intltool-0.51.0.tar.gz",
+                    "sha256" : "67c74d94196b153b774ab9f89b2fa6c6ba79352407037c8c14d5aeb334e959cd"
+                }
+            ],
+            "cleanup" : [
+                "*"
+            ]
+        },
+        {
+            "name" : "sound-theme-freedesktop",
+            "sources" : [
+                {
+                    "type" : "archive",
+                    "url" : "http://people.freedesktop.org/~mccann/dist/sound-theme-freedesktop-0.8.tar.bz2",
+                    "sha256" : "cb518b20eef05ec2e82dda1fa89a292c1760dc023aba91b8aa69bafac85e8a14"
+                }
+            ]
+        },
+        {
+            "name" : "gnome-break-timer",
+            "buildsystem" : "meson",
+            "builddir" : true,
+            "sources" : [
+                {
+                    "type" : "git",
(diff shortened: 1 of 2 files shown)


org.gnome.BreakTimer.json
```diff
@@ -0,0 +1,87 @@
...
+        "--talk-name=org.gnome.ScreenSaver",
+        "--talk-name=org.freedesktop.Notifications"
+    ],
+    "build-options" : {
+        "cflags" : "-O2 -g",
+        "cxxflags" : "-O2 -g",
+        "env" : {
+            "V" : "1"
+        }
+    },
```
bilelmoussaoui, line 24:
```suggestion
```
Not needed, they are set by default by the sdk

org.gnome.BreakTimer.json
```diff
@@ -0,0 +1,87 @@
+{
+    "id" : "org.gnome.BreakTimer",
```
bilelmoussaoui, line 2: Is it a GNOME app that can use org.gnome prefix?

org.gnome.BreakTimer.json
```diff
@@ -0,0 +1,87 @@
...
+    "runtime" : "org.gnome.Platform",
+    "runtime-version" : "3.38",
+    "sdk" : "org.gnome.Sdk",
+    "command" : "gnome-break-timer-settings",
+    "finish-args" : [
+        "--share=ipc",
+        "--socket=x11",
```
bilelmoussaoui, line 9:
```suggestion
        "--socket=fallback-x11",
```

bilelmoussaoui (changes requested):

org.gnome.BreakTimer.json
dylanmccall: It should be (historically it has been, and it still lives in https://gitlab.gnome.org/GNOME), but you've reminded me now is the best time to make sure that is in fact the right thing to do. I'll check on that and let you know in a bit, and indeed should avoid publishing this until then. Thanks for the quick review! :)

bilelmoussaoui:
bot, build org.gnome.BreakTimer

bilelmoussaoui:
bot, build org.gnome.BreakTimer

dylanmccall:
Oh, that build failure (in _Check for 128x128 icon_) is a funny one. So, this app comes with two desktop files: one, `org.gnome.BreakTimer.desktop`, for a daemon that runs in the background. It uses the base name because that's what the Background portal (and GNOME's Notifications settings panel) expects, but it is hidden from menus because it doesn't make sense for users to launch it directly. The _real_ entry point is `org.gnome.BreakTimer.Settings.desktop`, and our metainfo.xml has `<launchable type="desktop-id">org.gnome.BreakTimer.Settings.desktop</launchable>`.

However, appstream-compose does something that isn't obvious here: it takes the icon for `org.gnome.BreakTimer.Settings.desktop` (which is named `org.gnome.BreakTimer`) and then renders 64x64 and 128x128 icons to files with names like `/app/share/app-info/icons/128x128/org.gnome.BreakTimer.Settings.png`. That is, named after the _launchable_, not the component's id.

Trying this locally, it looks like the generated xml file in `/app/share/app-info/xmls` has the right stuff in it (`<icon type="cached" height="128" width="128">org.gnome.BreakTimer.Settings.png</icon>`), but perhaps that step in Flathub's buildbot is guessing the name of the icon rather than parsing the xml. So, I guess I can add `"skip-icons-check": true` to flathub.json here, as long as I promise we are rendering fixed size icons? (I'm hoping this oddity won't break anything other than the icons check script).

bilelmoussaoui:
> Oh, that build failure (in _Check for 128x128 icon_) is a curious one. So, this app comes with two desktop files: one, `org.gnome.BreakTimer.desktop`, for a daemon that runs in the background. It uses the base name because that's what the Background portal (and GNOME's Notifications settings panel) expects, but it is hidden from menus because it doesn't make sense for users to launch it directly. The _real_ entry point is `org.gnome.BreakTimer.Settings.desktop`, and our metainfo.xml has `<launchable type="desktop-id">org.gnome.BreakTimer.Settings.desktop</launchable>`.
> 
> However, appstream-compose does something that isn't obvious here: it takes the icon for `org.gnome.BreakTimer.Settings.desktop` (which is named `org.gnome.BreakTimer`) and then renders 64x64 and 128x128 icons to files with names like `/app/share/app-info/icons/128x128/org.gnome.BreakTimer.Settings.png`. That is, named after the _launchable_, not the component's id or the icon referred to by the desktop file.
> 
> Trying this locally, it looks like the generated xml file in `/app/share/app-info/xmls` has the right stuff in it (`<icon type="cached" height="128" width="128">org.gnome.BreakTimer.Settings.png</icon>`), but perhaps that step in Flathub's buildbot is guessing the name of the icon rather than parsing the xml. So, I guess I can add `"skip-icons-check": true` to flathub.json here, as long as I promise we are rendering fixed size icons? (I'm hoping this oddity won't break anything other than the icons check script).

What you should probably do is rename the main desktop file/metainfo to just the app-id instead of appending a `.Settings` suffix.

dylanmccall:
> What you should probably do is rename the main desktop file/metainfo to just the app-id instead of appending a `.Settings` suffix.   

Thank you for the suggestion. Unfortunately, while I did try that, it runs into trouble due to several assumptions made in GNOME and elsewhere. (While it solves some issues, it causes some funny stuff with the Background portal and dbus activation, as well as the notification settings panel).

There is a ticket upstream about refactoring the application so it's less weird (https://gitlab.gnome.org/GNOME/gnome-break-timer/-/issues/3). It isn't _too_ difficult to fix, but I think as long as we aren't causing trouble I'd be happy living with the current structure for an initial release. If this is actually a problem that breaks stuff, though, I'm happy to prioritize fixing that first.

Incidentally, would you be able to point me where these tests are implemented? I'd love to check on whether my guess is correct about what it's doing :)

barthalion:
The CI indeed just checks if the file at `builddir/*/share/app-info/icons/flatpak/128x128/$APP_ID` as it's the correct thing to do in 99% of cases. The app-info file looks fine, but similar assumption is made by the backend. Feel free to disable the icons check but let's wait with the merge until I fix it in the backend.

nedrichards:
bot, build org.gnome.BreakTimer

barthalion:
Welp, backend does the right thing now but the icon is not there on the server somehow? I will add it to my todo.

dylanmccall:
Thank you! And thanks for fixing that in the backend :) I'll check on that icon tomorrow as well. I think it should be reasonably fine to change the appdata to use a different launchable for now as well, so I'll probably make a quick release to do that. I made that ticket in the GNOME GitLab project about fixing it all so we don't need multiple launchers, and the appdata being a little nonsensical until then isn't the end of the world.

PR #1953, opened 2020-11-27
Add Proton GE

Lctrs:
Tested with Football Manager 2021, seems to work.

Should be merged in a `beta` branch.

Friendly ping @nanonyme and @gasinvein. You might be interested into reviewing this. :)


gasinvein:
I'd prefer to keep common git history with original proton flatpak, if possible. It would be useful for porting changes back and forth.

Lctrs:
@gasinvein Done.

gasinvein:
@Lctrs Did you do a 3-way-merge or something? Github doesn't seem to recognize changes against 589f1fb

Lctrs:
@gasinvein Fixed.

gasinvein:
@Lctrs Thanks. You can also use symlinks to avoid renaming the manifest and appdata, yet I'm not sure if flathub infra will handle it fine.

patches/proton/keep-system-gstreamer-plugins.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/proton b/proton
+index ea6e9af..17ce908 100755
+--- a/proton
++++ b/proton
+@@ -607,7 +607,7 @@ class Session:
+ 
+         self.env["WINEDLLPATH"] = g_proton.lib64_dir + "/wine:" + g_proton.lib_dir + "/wine"
+ 
+-        self.env["GST_PLUGIN_SYSTEM_PATH_1_0"] = g_proton.lib64_dir + "gstreamer-1.0" + ":" + g_proton.lib_dir + "gstreamer-1.0"
++        self.env["GST_PLUGIN_PATH"] = g_proton.lib64_dir + "gstreamer-1.0" + ":" + g_proton.lib_dir + "gstreamer-1.0"
```
gasinvein, line 10: This patch isn't needed since you're bundling gstreamer.

Lctrs:
@gasinvein Let's find out.

patches/proton/keep-system-gstreamer-plugins.patch
Lctrs: Adressed.

gasinvein:
I guess bundling gstreamer may be more tricky than it seems. We'll need to make sure that gst-plugins pick up proper (bundled) gstreamer via pkgconfig, and so do FAudio and Wine. Also RPATH probably need to be set so that proper gstreamer libs are loaded at run time.

Lctrs:
How can I check this ?

gasinvein:
Good question. At run-time, you can use `ldd` on faudio/wine libs. At build time, I guess the easiest way is to run a shell in flatpak-builder sandbox with `--build-shell=gst-plugins-base` and manually check which pkgconfig file was selected with `pkg-config --path gstreamer-1.0`

Lctrs:
OK they effectively pick system gstreamer (and not the bundled one).

gasinvein:
@Lctrs Submitted flathub/com.valvesoftware.Steam.CompatibilityTool.Proton#22. It may help with bundling gstreamer.

Lctrs:
@gasinvein You're a beast ! Everything is good now.

Lctrs:
bot, build com.valvesoftware.Steam.CompatibilityTool.ProtonGECustom

Lctrs:
@gasinvein looks like flathub infra supports symlinks :)

gasinvein:
Ok, so, to keep git difference at minimum, before it's merged, I'd suggest to do the following:
- Rebase on flathub/com.valvesoftware.Steam.CompatibilityTool.Proton@c430b5b and drop commits 6c08a12 and 387212b
- Squash commits 0b9d1e6 and 33aa754

Lctrs:
@gasinvein Done.

Lctrs:
bot, build com.valvesoftware.Steam.CompatibilityTool.ProtonGECustom

com.valvesoftware.Steam.CompatibilityTool.Proton.yml
```diff
@@ -13,43 +13,62 @@ sdk-extensions:
...
+  type: file
+  url: https://raw.githubusercontent.com/KhronosGroup/Vulkan-Docs/v1.2.160/xml/vk.xml
+  sha256: 73b449328899fb67703e9a229b6c8858f31b2933485b08710b4f69f842add94b
+
+x-proton-patches: &proton_patches
+  type: shell
+  commands:
+    - |
+      sed \
+        -e "s#\./dlls/winevulkan/make_vulkan#./dlls/winevulkan/make_vulkan --verbose --xml $(pwd)/vk.xml#g" \
```
gasinvein, line 64: Maybe just put `vk.xml` to `wine/dlls/winevulkan/vk-1.2.160.xml` (or wherever `make_vulkan` expects it to be) instead editing the script?

Lctrs: I would like to have the verbose flag as well.

compatibilitytool.vdf
```diff
@@ -0,0 +1,13 @@
+"compatibilitytools"
+{
+  "compat_tools"
+  {
+    "proton-ge-custom_flatpak"
+    {
+      "display_name" "Proton GE Custom @VERSION@ (community build @BUILD@)"
```
gasinvein, line 7: I think "community" doesn't make much sense here, since Proton-GE itself isn't an "official" thing. Ping @nanonyme

gasinvein:
This takes ridiculous amount of time to build, more than Chromium, it's even worse than normal Proton (which also builds much slower than it should). I believe this is due to enormous amount of git fetches/checkouts. We should do something about it.
My initial idea was to pack upstream sources to a tarball externally, then use that tarball as archive type source, yet I didn't go into details, namely where the tarball should be created and stored. Any suggestions? Maybe ask upstream to provide source tarball (with patches applied)?

Lctrs:
Will contact upstream about this.

Lctrs:
I don't know if that's feasible or not but it could also be an idea of improvement for `flatpak-builder` to mutualise sources operations across modules when they share same sources definitions.

gasinvein:
As a workaround, we can also use single git repo per flatpak-builder module instead of checking out the whole Proton-GE repo each time, but this would require much more handwork.

Lctrs:
No response from upstream yet. Anyway, I created https://github.com/Lctrs/proton-ge-custom-tarball-maker, a simple Github Actions to create source tarball from Proton GE patched sources. Let's see how it reduces build time.

bot, build com.valvesoftware.Steam.CompatibilityTool.ProtonGECustom

compatibilitytool.vdf
Lctrs: Well, it's a community build of a community Proton distribution :smile:

gasinvein:
@Lctrs vkd3d won't build outside of a git repo. I've previously made a fix for this gasinvein/proton-flatpak@9b453c4b808c11ea090e1e87444dc7ee508a6f9f

Lctrs:
bot, build com.valvesoftware.Steam.CompatibilityTool.ProtonGECustom

barthalion:
No complaints from me, but as it's not my area, I'll wait for approvals from @gasinvein and @nanonyme.

gasinvein:
@barthalion Do you have an opinion on using custom source tarball? IMO it adds complexity and isn't not good for transparency, but it reduces build time from ~3 hours to ~40 minutes.

compatibilitytool.vdf
gasinvein: Also, won't the name be too long? Is the version still visible in Steam UI?

Lctrs: Yep.

Capture d’écran de 2020-12-01 15-02-53

Lctrs:
@gasinvein I have inlined flatpak's id and display name with what you did in https://github.com/gasinvein/proton-flatpak.

gasinvein:
@Lctrs That wasn't necessary, I've did some PoC there that shouldn't be considered entirely correct. But well, I guess it'll be easier to sync changes now.

barthalion:
> @barthalion Do you have an opinion on using custom source tarball? IMO it isn't good for transparency, but it reduces build time from ~3 hours to ~40 minutes.

Yeah, it's not ideal but not completely uncommon in the packaging world either so let's keep it.

Lctrs:
Looks like the merge action doesn't like symlinks : https://github.com/flathub/flathub/runs/1491179128

PR #1956, opened 2020-11-28
Add org.freedesktop.fldigi

lu9dce:
my first flatpak

Fldigi is a free and open-source program which allows an ordinary computer's sound card to be used as a simple two-way data modem. The software is mostly used by amateur radio operators who connect the microphone and headphone connections of an amateur radio SSB transceiver or an FM two way radio to the computer's headphone and microphone connections, respectively. This interconnection creates a "sound card defined radio" whose available bandwidth is limited by the sound card's sample rate and the external radio's bandwidth. Such communications are normally done on the shortwave amateur radio bands in modes such as PSK31, MFSK, RTTY, Olivia, and CW. Increasingly, the software is also being used for data on VHF and UHF frequencies using faster modes such as 8-PSK. Using this software, it is possible for amateur radio operators to communicate worldwide while using only a few watts of RF power


org.freedesktop.fldigi.yml
```diff
@@ -0,0 +1,38 @@
+app-id: org.freedesktop.fldigi
+runtime: org.freedesktop.Platform
+runtime-version: '20.08'
+sdk: org.freedesktop.Sdk
+command: fldigi
+finish-args:
+  - --share=ipc
+  - --fallback-x11
```
bilelmoussaoui, line 8:
```suggestion
  - --socket=fallback-x11
```

LeandroStanger:
bot, build org.freedesktop.fldigi

lu9dce:
what is this how i fix it?

stat: cannot stat 'builddir/*/share/app-info/xmls/org.freedesktop.fldigi.xml.gz': No such file or directory

org.freedesktop.fldigi.metainfo.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2020 hamradio lu9dce -->
+<component type="desktop-application">
+ <id>org.freedesktop.fldigi</id>
```
nedrichards, line 4: is this really the app-id? e.g. is this actually part of the freedesktop.org project?

org.freedesktop.fldigi.metainfo.xml
```diff
@@ -0,0 +1,29 @@
...
+ <metadata_license>CC0-1.0</metadata_license>
+ <project_license>GPL-3.0</project_license>
+ <name>Fldigi</name>
+ <summary>Amateur Radio Sound Card Communications</summary>
+ <developer_name>dave-w1hkj, mdblack98, sb42</developer_name>
+ <url type="homepage">https://sourceforge.net/projects/fldigi/
```
nedrichards, line 10: this tag isn't closed so the xml is invalid and it fails to compile properly.

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64"]
```
nedrichards, line 2: what about this doesn't work on aarch64?

lu9dce: I have no way to ensure that

org.freedesktop.fldigi.metainfo.xml
lu9dce: ok

org.freedesktop.fldigi.metainfo.xml
lu9dce: I really can't answer this to you .. where did I get that information from? or what corresponds to fence?

lu9dce:
How often does the bot control?

PR #1957, opened 2020-11-28
Update Video Downloader 

GustavoPeredo:
Hello! I really like the app Video Downloader, but recently the current maintainer, Unrud, archived the main repository and removed youtube-dl as a dependency (meaning the app is currently broken on flathub). I want to keep maintaining the app. I also added a dark mode :)

### DIFF
--- a/com.github.unrud.VideoDownloader.json
+++ b/com.github.unrud.VideoDownloader.json
@@ -95,7 +95,7 @@
                 {
                     "type" : "archive",
                     "url" : "https://yt-dl.org/downloads/2020.11.26/youtube-dl-2020.11.26.tar.gz",
-                    "sha256" : "2ec41fc129f8c42745899a8c9f0b5d96c30a2930"
+                    "sha256" : "fb106cd60f94fe00afee46d4e028cdca8ebff171ddfa9d4ff2a2fd5956586f7f"
                 }
             ]
         },
@@ -108,7 +108,7 @@
                     "type" : "git",
                     "url" : "https://github.com/GustavoPeredo/video-downloader.git",
                     "branch" : "release",
-                    "commit" : "b9f9f4e7fb2d0a8bd572ea61d1940f3315cee82a"
+                    "commit" : "2ec41fc129f8c42745899a8c9f0b5d96c30a2930"
                 }
             ]
         }
--- a/com.github.unrud.VideoDownloader.json
+++ b/com.github.unrud.VideoDownloader.json
@@ -1,0 +1,116 @@
+{
+    "app-id" : "com.github.unrud.VideoDownloader",
+    "runtime" : "org.gnome.Platform",
+    "runtime-version" : "3.38",
+    "sdk" : "org.gnome.Sdk",
+    "command" : "video-downloader",
+    "finish-args" : [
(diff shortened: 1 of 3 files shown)


GustavoPeredo:
bot, build com.github.unrud.VideoDownloader

GustavoPeredo:
bot, build com.github.unrud.VideoDownloader

com.github.unrud.VideoDownloader.json
```diff
@@ -0,0 +1,116 @@
+{
+    "app-id" : "com.github.unrud.VideoDownloader",
```
nedrichards, line 2: if this is a new app that you plan to maintain (albeit forked from videodownloader) then you need to change the app-id to match.

GustavoPeredo: My original idea was to keep updating video-downloader that is already on Flathub, would that be possible?

bilelmoussaoui: Except that's not the same app at all but a fork of it. So please use a different app-id in such case.

GustavoPeredo: Understandable, will do so

GustavoPeredo:
bot, build org.gustavoperedo.VideoDownloader

GustavoPeredo:
I changed the app-id

org.gustavoperedo.VideoDownloader.json
```diff
@@ -0,0 +1,116 @@
...
+    "sdk" : "org.gnome.Sdk",
+    "command" : "video-downloader",
+    "finish-args" : [
+        "--share=network",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--talk-name=org.freedesktop.FileManager1",
+        "--filesystem=xdg-download/VideoDownloader:create",
+        "--metadata=X-DConf=migrate-path=/org/gustavoperedo/VideoDownloader/"
```
barthalion, line 14: I don't think it's needed. It's not as if you were using this name previously so there's nothing to migrate.

org.gustavoperedo.VideoDownloader.json
```diff
@@ -0,0 +1,116 @@
+{
+    "app-id" : "org.gustavoperedo.VideoDownloader",
```
barthalion, line 2: Do you own this domain? If not, it should be `com.github.GustavoPeredo.VideoDownloader`.

GustavoPeredo: I changed it to io.github.gustavoperedo.VideoDownloader (following the [documentation](https://docs.flatpak.org/en/latest/conventions.html#application-ids).)

I have another app that I used the org.gustavoperedo prefix, for me to update it, do I just have to change the manifest's app-id? (The app is already on flathub)

GustavoPeredo:
bot, build io.github.gustavoperedo.VideoDownloader

org.gustavoperedo.VideoDownloader.json
barthalion: No, it's fine for existing app to stay that way. We try to limit friction caused by ID changes to minimum.

PR #1958, opened 2020-11-29
Add io.github.cmus.cmus

X-m7:
The desktop file, icon and AppStream metadata are not in upstream unfortunately, a couple of people uploaded theirs in the respective issues (cmus/cmus#978 and cmus/cmus#965) so I used those here, and I wrote the AppStream xml myself with bits taken from the [website](https://cmus.github.io/), which are the description and screenshot. I have also opened an [issue](https://github.com/cmus/cmus/issues/1007) upstream regarding Flatpak/Flathub distribution, and I also mentioned the AppStream thing there too.

Also, `configure.patch` is there because the prefix is apparently passed differently than what is normally expected by the upstream's configure script (specifically the problem is the `--`).


X-m7:
bot, build io.github.cmus.cmus

X-m7:
bot, build io.github.cmus.cmus

X-m7:
bot, build io.github.cmus.cmus

X-m7:
bot, build io.github.cmus.cmus

X-m7:
bot, build io.github.cmus.cmus

X-m7:
bot, build io.github.cmus.cmus

io.github.cmus.cmus.yml
```diff
@@ -0,0 +1,134 @@
...
+        url: https://github.com/cmus/cmus/archive/v2.8.0.tar.gz
+        sha512: cf359dfcefa833a5b10a2d16ac405672bea762b62b7177c115560127035682fba65c15b9a8710179a343d1f99212a0260b5c095542982202e2cd1bef5b0c17fc
+      - type: patch
+        path: configure.patch
+      - type: file
+        path: cmus.desktop
+      - type: file
+        path: cmus.svg
+      - type: file
+        path: appdata.xml
```
LeandroStanger, line 134:
```suggestion
        path: io.github.cmus.cmus.metainfo.xml
```

X-m7:
bot, build io.github.cmus.cmus

X-m7:
bot, build io.github.cmus.cmus

README.md
```diff
@@ -0,0 +1,6 @@
+# cmus Flatpak
+
+[Icon](https://github.com/cmus/cmus/issues/965) and [desktop file](https://github.com/cmus/cmus/issues/978) not from upstream.
+
+## Current issues
+- MPRIS fails with "connection reset by peer" without full session bus permissions
```
barthalion, line 6: At first glance, other apps do this: `--own-name="org.mpris.MediaPlayer2.firefox.*"`. But I assume there's a different problem with the name cmus uses.

cmus.svg
```diff
@@ -0,0 +1,236 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
barthalion, line 1: Ideally this should be part of upstream repo too.

io.github.cmus.cmus.yml
```diff
@@ -0,0 +1,134 @@
...
+
+  - name: libcddb
+    cleanup:
+      - /include
+      - /bin
+    sources:
+      - type: archive
+        url: https://downloads.sourceforge.net/libcddb/libcddb-1.3.2.tar.bz2
+        sha512: 0e07e7d37f2013164d75b530a5528f54847271fd20f3b7bedb192a05d1f61dcf006d10dc2927efe155a01eddcc57b940bc31d8ac88d5dfc4f1a09771caa84e0a
+      - type: file
```
barthalion, line 93: You can just copy over these files like here: https://github.com/flathub/org.kde.okular/blob/909e7691e274b8d1f991c4b2e045d23bc33a416b/org.kde.okular.json#L191

README.md
X-m7: I tried adding the * in, as well as the quotes, neither fixed it. There's more details on the problem in flatpak/flatpak#3978, looks like it's a disagreement between sd-bus (the systemd dbus implementation, used by cmus) and Flatpak's xdg-dbus-proxy.

cmus.svg
X-m7: Yes, as I said there is an issue upstream about it (the icon is from that issue's reporter).

X-m7:
bot, build io.github.cmus.cmus

io.github.cmus.cmus.metainfo.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
```
bilelmoussaoui, line 2:
```suggestion
<component type="console-application">
```

io.github.cmus.cmus.yml
X-m7: Fixed by 271adf3.

X-m7:
bot, build io.github.cmus.cmus

io.github.cmus.cmus.metainfo.xml
barthalion: I thought we were fine with calling it `desktop-application` if it's more TUI than CLI? We have done it already for Poezio and something Bitcoin related I think.

barthalion: Although, interestingly, it seems to retain icons and screenshots in app-info, I guess thanks to the desktop file. It won't be displayed on the website though.

X-m7: This is definitely more TUI than CLI. Neovim for one uses `desktop-application` as its type as well (although that's from its upstream instead of being Flathub specific).

I'll switch it back to desktop application then, also the launch button in GNOME Software doesn't work (I guess since a .desktop file wouldn't apply to console applications normally).

barthalion:
/merge

Please poke us if anyone from upstream is interested in maintenance.

PR #1961, opened 2020-11-29
Add io.github.mightycreak.Diffuse

MightyCreak:


com.github.mightycreak.Diffuse.yml
```diff
@@ -0,0 +1,19 @@
...
+  - --filesystem=home
+modules:
+  - name: diffuse
+    buildsystem: meson
+    sources:
+      - type: git
+        url: https://github.com/MightyCreak/diffuse
+        branch: v0.6.0
+rename-desktop-file: diffuse.desktop
+rename-icon: diffuse
```
bilelmoussaoui, line 19: You probably want to rename those upstream

bilelmoussaoui:
bot, build com.github.mightycreak.Diffuse

com.github.mightycreak.Diffuse.yml
MightyCreak: I'm sorry, I'm new to flatpak packaging, I created the manifest by looking at manifest on other apps. How should I rename these?

X-m7: What he's saying is that you should change the names of those files in the application's source directly as opposed to making the Flatpak builder do it, since it appears to be your code (or at least your repository).

bilelmoussaoui: Thanks @x-m7

barthalion:
Your app needs to install appdata and desktop files. You can use https://bilelmoussaoui.github.io/metainfo-generator/ or https://www.freedesktop.org/software/appstream/metainfocreator/#/ for the first.

MightyCreak:
FYI I haven't forgotten, I just wait for the weekend to start on it again :wink:

MightyCreak:
I got a question about the app ID, most of the application I see have upper case letters in their ID (mostly for the app name, e.g. `com.github.bilelmoussaoui.Authenticator`), but the `appstreamcli validate` warns about using upper case letters.

What should I do? follow the validation warning or what appears to be the de facto standard?

Thanks!

com.github.mightycreak.Diffuse.yml
MightyCreak: Ok, so I've looked a bit more at what all that is doing. The thing is, while it is my repository, it is a fork of a no longer maintained repo (on sourceforge).

I'm still learning a lot about quite all the concepts (Python, GTK, Meson, Flatpak, Flathub) on top of the application code and structure themselves, so I'd rather keep these "rename" properties for now as I don't have to do much changes in the existing repo files. (What I'm afraid is if my application will still work when installed without Flatpak if I rename the files upstream).

I'd like to be a bit more comfortable with all that before renaming the files, if it's ok with you?

MightyCreak:
In the meantime, I've added a metainfo file in the repo and pushed the manifest with the latest commit. Hope it'll work this time :wink:

MightyCreak:
bot, build io.github.mightycreak.Diffuse

MightyCreak:
:tada:

barthalion:
> I got a question about the app ID, most of the application I see have upper case letters in their ID (mostly for the app name, e.g. `com.github.bilelmoussaoui.Authenticator`), but the `appstreamcli validate` warns about using upper case letters.

It's totally fine; also it doesn't complain on our infrastructure. Maybe some old version?

io.github.mightycreak.Diffuse.yml
```diff
@@ -0,0 +1,19 @@
+app-id: io.github.mightycreak.Diffuse
+runtime: org.gnome.Platform
+runtime-version: '3.38'
+sdk: org.gnome.Sdk
+command: diffuse
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc
+  - --filesystem=home
```
barthalion, line 10: Ideally, you should switch to https://developer.gnome.org/gtk3/stable/gtk3-GtkFileChooserNative.html in the future – that will allow you to drop this line as it will be handled by the portal.

MightyCreak: Thank you! I created an issue in my repo: https://github.com/MightyCreak/diffuse/issues/78

PR #1968, opened 2020-12-01
Add com.locomalito.MalditaCastilla

bertogg:
I'm submitting a new game called "Maldita Castilla".

The game is not open source but it is distributed under the Creative Commons BY-NC-ND-3.0 license.

The package is quite simple because I'm simply shipping the upstream binaries and it doesn't need any additional dependency, but since there is only a 32-bit version available I had to use the most recent i386 runtime, which seems to be 18.08. Please tell me if there's a better way to ship 32-bit apps.


bertogg:
Also: I didn't use an extra-data source for the game itself because of its permissive license and its relatively small size, but I can change that if you think it's more appropriate.

lu9dce:
hola alberto a ver si te puedo ayudar
Hi Alberto, see if I can help you

lu9dce:
it's easy to do .. start making the manifest
let's see how it looks
I only have 64bit and I don't know how much flatpak will go in 32

bertogg:
Hi, the manifest and all required files are already in the pull request.

lu9dce:
bot, build com.locomalito.MalditaCastilla

bertogg:
I see that it complains about the lack of 128x128 icon but the app only provides 64x64, I would need to scale it.

The other arches are not supposed to build, is there a way to specify that only the i386 version should be built?

bertogg:
Ok I added the 128x128 icon (scaled as I said earlier) and a flathub.json file to indicate that only the i386 version should be built.

lu9dce:
alberto, en un comentario nuevo pone bot, build com.locomalito.MalditaCastilla
para que el bot te revise a ver si esta todo bien

Alberto, in a new comment put 
bot, build com.locomalito.MalditaCastilla
so that the bot will check you to see if everything is ok

bertogg:
Ok the app works fine, the only thing is that I have to specify --arch=i386 in order to install it, which brings me again to the question of whether this is the preferred way to ship 32-bit apps.

Otherwise everything seems to work fine.

lu9dce:
always try to use the source code
Okay, this is the case .. we would have to see if they don't give the code
since based on it it can be compiled for any architecture
few 32bit pc left
and when a pr lowers it and practically does not have installation
as in the case of this game, I don't know if it's worth passing it to flatpak
for the user it is easier to download the game and run it than to put
quite a flatpak to run it
at least it seems to me

bertogg:
There is no source code to this game.

It would normally be enough to tell people to download the binary from upstream, but since it's a 32-bit binary it won't work on the average user's system.

Flatpak hides all that complexity by providing a 32-bit runtime that just works in a modern x86_64 computer.

Plus, the game is run inside a container, which is nicer for running software that doesn't have source code available.

com.locomalito.MalditaCastilla.yaml
```diff
@@ -0,0 +1,44 @@
...
+        path: maldita-castilla.desktop
+      - type: file
+        path: icon.png
+      - type: script
+        dest-filename: maldita-castilla
+        commands:
+          - ln -sfT "$XDG_CONFIG_HOME" ~/.config
+          - exec /app/lib/maldita-castilla/runner "$@"
+      - type: file
+        path: com.locomalito.MalditaCastilla.appdata.xml
```
LeandroStanger, line 44: New appdata files should use metainfo instead. https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html

com.locomalito.MalditaCastilla.yaml
```diff
@@ -0,0 +1,44 @@
+app-id: com.locomalito.MalditaCastilla
+branch: stable
+runtime: org.freedesktop.Platform
+runtime-version: '18.08'
```
LeandroStanger, line 4: Update runtime 20.08
```suggestion
runtime-version: '20.08'
```

bertogg: > Update runtime 20.08

Hi, but the binary is for i386 only, how do I make it work with an x86_64 runtime?

barthalion:
The ideal way would be to use org.freedesktop.Platform.Compat.i386 extension to run it on the latest runtime like this (I think!):

```diff
diff --git a/com.locomalito.MalditaCastilla.yaml b/com.locomalito.MalditaCastilla.yaml
index 541a9ab..90c9a2d 100644
--- a/com.locomalito.MalditaCastilla.yaml
+++ b/com.locomalito.MalditaCastilla.yaml
@@ -1,7 +1,7 @@
 app-id: com.locomalito.MalditaCastilla
-branch: stable
+default-branch: stable
 runtime: org.freedesktop.Platform
-runtime-version: '18.08'
+runtime-version: '20.08'
 sdk: org.freedesktop.Platform
 command: maldita-castilla
 rename-desktop-file: maldita-castilla.desktop
@@ -14,12 +14,23 @@ finish-args:
   - --socket=pulseaudio
   # Direct Rendering access
   - --device=dri
+
+add-extensions:
+  # Support 32-bit python
+  'org.freedesktop.Platform.Compat.i386':
+    directory: 'lib/i386-linux-gnu'
+    version: '20.08'
+
+  'org.freedesktop.Platform.Compat.i386.Debug':
+    directory: 'lib/debug/lib/i386-linux-gnu'
+    version: '20.08'
+    no-autodownload: true
+
 modules:
   - name: maldita-castilla
-    only-arches:
-      - i386
     buildsystem: simple
     build-commands:
+      - install -d /app/lib/i386-linux-gnu
       - install -m755 -D maldita-castilla /app/bin/maldita-castilla
       - install -m644 -D maldita-castilla.desktop /app/share/applications/maldita-castilla.desktop
       - install -m644 -D assets/icon.png /app/share/icons/hicolor/64x64/apps/icon.png
```

This unfortunately gives me an unhelpful error though:

```
[📦 com.locomalito.MalditaCastilla ~]$ /app/lib/maldita-castilla/runner
Bad system call (core dumped)
```

bertogg:
> [📦 com.locomalito.MalditaCastilla ~]$ /app/lib/maldita-castilla/runner
> Bad system call (core dumped)

Ok I have been taking a look at this, it seems that you were missing `--allow=multiarch`.

The problem with using `org.freedesktop.Platform.Compat.i386` version 20.08 is that the version of openssl is too recent, and even if I build an older one myself the app crashes during `glXCreateContext()`. I can try to dig further into this, perhaps I also need to build older versions of some of the libraries.

Alternatively I could use `org.freedesktop.Platform.Compat.i386` version 18.08, that one works out of the box and the game seems to work fine, but I don't know if it's ok to mix version 20.08 of the runtime with version 18.08 of the i386 extension (the manifest file does allow that).

barthalion:
Ah, good catch. As for mixing extension version, this can technically lead to random breakage. …but I guess that if it works, it works.

bertogg:
> Ah, good catch. As for mixing extension version, this can technically lead to random breakage. …but I guess that if it works, it works.

If we want to avoid mixing the runtime and extension versions I only see two alternatives:

* Use the 18.08 x86_64 runtime with the 18.80 Compat.i386 extension. This should work just fine.

* Use the 20.08 version in both cases. I would need to build openssl for i386 and try to figure out why the app is crashing. This may involve building additional libraries and revisiting this every time the runtime is updated.

I suspect that in either case the game is only going to use the Compat.i386 libraries and won't use any files from the runtime itself.

This is just a game so it's maybe not the best example, but I guess that the general problem is what to do with an app that provides some useful functionality but hasn't been updated in a while, depends on an old set of libraries, and does not provide source code so it cannot be recompiled.

As I said in a previous comment Flatpak does a very nice job of hiding all this complexity and giving the user a containerized version of the app that just works, is integrated with the desktop, and won't break when you upgrade your operating system.

But I don't know if apps in Flathub are always expected to use the most recent runtimes or if it's ok to stick to older versions when there are good reasons to do so. Or, in other words, if packages like this game are out of scope for what Flathub seeks to provide.

bertogg:
Ok I'll update the packaging and push the new changes, thanks!

bertogg:
bot, build com.locomalito.MalditaCastilla

bertogg:
Ok this works but the Compat.i386 extension is not downloaded automatically, at least not from the command-line :-?

barthalion:
Yeah, it's flatpak bug/feature not to download extensions automatically from foreign remotes. It will work fine after it hits Flathub.

Unfortunately the game does not work for me; for a second there's a full-screen black window which immediately disappears. There's no errors in terminal.

bertogg:
> Unfortunately the game does not work for me; for a second there's a full-screen black window which immediately disappears. There's no errors in terminal.

Hmmm... I think this game needs X11, but you at least get to see a black window so I assume that the display is being detected.

What are that last messages that you see in the terminal?

barthalion:
Nothing really interesting:

```
***************************************
*     YoYo Games Linux Runner V0.1    *
***************************************	 
CommandLine: -game game.unx
ExeName= /app/lib/maldita-castilla/runner
MemoryManager allocated: 16827
RunnerLoadGame: game.unx
RunnerLoadGame() - assets/game.unx
Checking if INIFile assets/options.ini Exists
assets/options.ini file contents: [Linux]
DisplayName="Maldita Castilla"

INI DisplayName=Maldita Castilla
Reading File assets/game.unx
Loaded File assets/game.unx
IFF wad found
Get Resolution
Get Header Information
InitGMLFunctions
HighScore..SavePrePend /var/home/bpiotrowski/.config/maldita_castilla_arcade_cabinet/ 
Game..Math..Graphic..Action..File..Resource..Interaction..3D..Particle..Misc..DS..Sound..DPlay..Physics..Gamepad..GAMEPAD: Initialising Ubuntu support
GAMEPAD: 4 axis values (last) at 0x95aced0
GAMEPAD: 4 axis values (last) at 0x95acff0
GAMEPAD: 4 axis values (last) at 0x95ac000
GAMEPAD: 4 axis values (last) at 0x95ac350
Buffers..Networking..Shaders..YoYo..Fini
Code_Load()
VARI_Load()
ID_STRG
Create Window
Display Size(Pixels): 2560,1440
GameDisplayName=Maldita Castilla
XF86VidModeExtension-Version 2.2
Got Doublebuffered Visual!
glX-Version 1.4
Icon: w=64 h=64
WindowCentre: 1120,608
Depth 24
Congrats, you have Direct Rendering!
Set Fullscreen 1
sync = 1
**** GLX Extensions ***
GLX_ARB_create_context GLX_ARB_create_context_no_error GLX_ARB_create_context_profile GLX_ARB_fbconfig_float GLX_ARB_framebuffer_sRGB GLX_ARB_get_proc_address GLX_ARB_multisample GLX_EXT_buffer_age GLX_EXT_create_context_es2_profile GLX_EXT_create_context_es_profile GLX_EXT_fbconfig_packed_float GLX_EXT_framebuffer_sRGB GLX_EXT_import_context GLX_EXT_texture_from_pixmap GLX_EXT_visual_info GLX_EXT_visual_rating GLX_MESA_copy_sub_buffer GLX_MESA_query_renderer GLX_MESA_swap_control GLX_OML_swap_method GLX_OML_sync_control GLX_SGIS_multisample GLX_SGIX_fbconfig GLX_SGIX_pbuffer GLX_SGIX_visual_select_group GLX_SGI_make_current_read GLX_SGI_video_sync 
Checking for GLX_EXT_swap_control
Checking for GLX_SGI_swap_control
Checking for GLX_MESA_swap_control
Vsync: GLX_MESA
DOUBLE BUFFERED
```

bertogg:
Yeah, that is not very informative. It should continue like this:

```
[...]
Congrats, you have Direct Rendering!
Set Fullscreen 1
sync = 1
**** GLX Extensions ***
GLX_ARB_create_context GLX_ARB_create_context_no_error GLX_ARB_create_context_profile GLX_ARB_create_context_robustness GLX_ARB_fbconfig_float GLX_ARB_framebuffer_sRGB GLX_ARB_get_proc_address GLX_ARB_multisample GLX_EXT_buffer_age GLX_EXT_create_context_es2_profile GLX_EXT_create_context_es_profile GLX_EXT_fbconfig_packed_float GLX_EXT_framebuffer_sRGB GLX_EXT_import_context GLX_EXT_texture_from_pixmap GLX_EXT_visual_info GLX_EXT_visual_rating GLX_INTEL_swap_event GLX_MESA_copy_sub_buffer GLX_MESA_query_renderer GLX_MESA_swap_control GLX_OML_swap_method GLX_OML_sync_control GLX_SGIS_multisample GLX_SGIX_fbconfig GLX_SGIX_pbuffer GLX_SGIX_visual_select_group GLX_SGI_make_current_read GLX_SGI_swap_control GLX_SGI_video_sync 
Checking for GLX_EXT_swap_control
Checking for GLX_SGI_swap_control
Vsync: GLX_SGI
DOUBLE BUFFERED
Init Graphics
GR_D3D_Init()
OpenGL: version string 3.0 Mesa 19.1.7
OpenGL GLSL: version string 1.30
Texture is a render target - 320,224
Texture is new texture - 16,16
Check TextureID
finished!!
[...]
```

One thing that I noticed is that your display size (2560,1440) is larger than mine (1920,1200). I wonder if that has anything to do... I'll try to talk to upstream.

bertogg:
Ok I talked to upstream and I got to the conclusion that there isn't a lot that we can do to debug or fix this, so I guess I'll just withdraw the request.

Thanks for your time anyway!

PR #1973, opened 2020-12-03
Add be.telenet.user.QSSTV

lu9dce:
QSSTV is a program for receiving and transmitting SSTV and HAMDRM
(sometimes called DSSTV). It is compatible with most of MMSSTV and EasyPal

transmitdrmb

PR #1977, opened 2020-12-04
Viking v1.8 Flatpak

rnorris:
Built without deploying GPSBabel application.

https://sourceforge.net/projects/viking/


nedrichards:
bot, build org.viking.Viking

rnorris:
Is it possible to disable the aarch64 build? It fails in building the dependency exiv2, I don't have the skills to fix it. I basically copied the methods for building this part from the GIMP flatpak build.

In the x86_64 buid it fails at the step  "Check for AppStream xml", I've created an org.viking.Viking.appdata.xml file, which seemed to be similar to other flatpaks.
Can anyone point to documentation that explains what I should be doing?

lu9dce:
if this the source code it should compile for aarch64
leave this to last to see if the bot compiles it

AppStream xml see https://github.com/flathub/flathub/pull/1973/files

rename and install it in /app/share/metainfo/

lu9dce:
bot, build org.viking.Viking

lu9dce:
Check for 128x128 icon

rnorris:
bot, build org.viking.Viking

org.viking.Viking.appdata.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2020 Rob Norris -->
+<component>
+  <!-- About this metadata file -->
+  <!-- https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html -->
+  <id type="desktop">org.viking.Viking</id>
+  <launchable type="desktop-id">viking.desktop</launchable>
```
barthalion, line 7:
```suggestion
  <launchable type="desktop-id">org.viking.Viking.desktop</launchable>
```

org.viking.Viking.appdata.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
barthalion, line 1: FYI, please consider maintaining this code in the upstream repo – it's not specific to Flathub and other distributions/stores could benefit from it too.

org.viking.Viking.yml
```diff
@@ -0,0 +1,183 @@
+app-id: org.viking.Viking
+# https://docs.flatpak.org/en/latest/available-runtimes.html
+# NB Viking doesn't use any Gnome specifics so freedesktop should suffice for the build
+# Albeit Yelp for help is Gnome only
+runtime: org.gnome.Platform
+runtime-version: '3.38'
+sdk: org.gnome.Sdk
+command: viking
+rename-icon: viking
+rename-desktop-file: viking.desktop
```
barthalion, line 10: Ideally these renames should happen upstream.

org.viking.Viking.yml
```diff
@@ -0,0 +1,183 @@
...
+finish-args:
+  - --share=network
+  - --system-talk-name=org.freedesktop.GeoClue2
+  - --share=ipc
+  - --socket=x11
+# Allow to read/save anywhere (subject to normal permissions)
+  - --filesystem=host
```
barthalion, line 17: How common is it to save/load files from outside of /home? It's not much but if it's not very popular, `home` would be slightly better here.

org.viking.Viking.appdata.xml
rnorris: I will be including all related code in the upstream repo, once the refinements here are completed and suitable for publication.

org.viking.Viking.yml
rnorris: ?
I presumed the intention of these renames were such that upstream didn't need any changes (or create otherwise duplicate files) so the flatpak build would rename the existing icon / desktop files that exist from 'program.extension' --> 'org.program.Program.extension'

org.viking.Viking.yml
rnorris: Very common.
One of the primary uses of Viking is to read and write GPX files from GPS devices.
Many of the popular GPS devices (e.g. Garmin Etrex) use a USB filesystem for data transfer of Tracks, Routes and Waypoints.
Read is mostly used for accessing recordings of tracks of where one has been (and Waypoints created on the GPS device); and write is for putting a planned route/track on to the device that you subsequently may follow when one is out and about (and maybe Waypoints of Points Of Interest (POIs)to visit - e.g. GeoCaches).
Under Linux the GPS Device is mounted just like any external USB drive and this is often located in /media/<user>/<DeviceName>/ by default.
So IFAIK this sits outside any $XDG_* scheme, so I'm afraid $HOME would be too limiting.

PS Thanks for the review.

nedrichards:
bot, build org.viking.Viking

PR #1989, opened 2020-12-10
Add so.notion.Notion

e-user:
Notion is a proprietary all-in-one productivity software that doesn't officially support Linux at the moment. This build extracts and patches their Windows Electron application to create a Linux Electron app from it.

My repository at https://github.com/e-user/so.notion.Notion has additional tools to generate new sources for newer versions of Notion but I wasn't sure whether they can or should be added, here.


so.notion.Notion.yaml
```diff
@@ -0,0 +1,142 @@
...
+        sha256: 4528bb49ab677eb73b84da3a60840f6d1081ff0afd5c2d64ef6247058a38c19d
+        dest-filename: notion.exe
+      - type: file
+        path: package-lock.json
+      - type: file
+        path: package.json
+      - type: file
+        path: so.notion.Notion.desktop
+      - type: file
+        path: so.notion.Notion.appdata.xml
```
LeandroStanger, line 97: New appdata files should use metainfo instead. https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html

LeandroStanger:
bot, build  so.notion.Notion

barthalion:
bot, build so.notion.Notion

so.notion.Notion.yaml
```diff
@@ -0,0 +1,142 @@
...
+  cxxflags: -O2 -g
+  env:
+    NPM_CONFIG_LOGLEVEL: info
+
+modules:
+  - name: python2.7
+    sources:
+      - type: archive
+        url: https://www.python.org/ftp/python/2.7.18/Python-2.7.18.tar.xz
+        sha256: b62c0e7937551d0cc02b8fd5cb0f544f9405bafc9a54d3808ed4594812edef43
```
barthalion, line 33: Is it needed for anything on runtime? If not, should be removed after build via `cleanup: [*]`

so.notion.Notion.yaml
```diff
@@ -0,0 +1,142 @@
...
+  - --share=network
+  - --filesystem=xdg-documents:ro
+  - --filesystem=xdg-pictures:ro
+  - --filesystem=xdg-music:ro
+  - --filesystem=xdg-videos:ro
+  - --filesystem=xdg-download
+
+build-options:
+  cflags: -O2 -g
+  cxxflags: -O2 -g
```
barthalion, line 24: CFLAGS/CXXFLAGS are provided by the runtime.

so.notion.Notion.yaml
```diff
@@ -0,0 +1,142 @@
...
+          --platform linux \
+          --arch x64 \
+          --out $FLATPAK_DEST \
+          --electron-version 11.0.4 \
+          --executable-name notion \
+          --protocol notion \
+          --electron-zip-dir=flatpak-node/electron-cache
+      - install notion ${FLATPAK_DEST}/bin/
+      - convert app/icon.ico icon.png
+      - install -D icon-0.png ${FLATPAK_DEST}/share/icons/hicolor/256x256/apps/${FLATPAK_ID}.png
```
barthalion, line 135: 256x256 is enough, flatpak will downscale it. SVG would be even better.

so.notion.Notion.appdata.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="utf-8"?>
+<component type="desktop-application">
+  <id>so.notion.Notion</id>
+  <metadata_license>CC-BY-SA-3.0</metadata_license>
+  <project_license>LicenseRef-proprietary</project_license>
```
barthalion, line 5: The problem I see here is that we don't have rights to redistribution (unless Notion's EULA allows it or you have contacted them and they said it's ok).

It would need to be reworked to use extra-data and run the steps at https://github.com/flathub/flathub/pull/1989/files#diff-8916716d83ed8a4e8d742b7f20df159b50fd930948e3e2877aeb74f3c579147eR116 in apply_extra.

barthalion:
> My repository at https://github.com/e-user/so.notion.Notion has additional tools to generate new sources for newer versions of Notion but I wasn't sure whether they can or should be added, here.

Yes, it's fine to keep them here,

so.notion.Notion.appdata.xml
```diff
@@ -0,0 +1,40 @@
...
+  </content_rating>
+  <developer_name>Notion Labs, Inc.</developer_name>
+  <name>Notion</name>
+  <summary>All-in-one workspace</summary>
+  <description>
+    <p>Notion is an application that provides components such as databases, kanban
+    boards, wikis, calendars and reminders.</p>
+    <p>Users can connect these components to
+    create their own systems for knowledge management, note taking, data management,
+    project management, among others.</p>
```
barthalion, line 19: It needs a disclaimer like https://github.com/flathub/com.microsoft.Teams/blob/master/com.microsoft.Teams.metainfo.xml#L16.

so.notion.Notion.yaml
e-user: Cool, I didn't even know about this 😄

so.notion.Notion.yaml
e-user: I found an official SVG on their branding guidelines page and will include it in the next iteration.

e-user:
Is there a way of applying patches conditionally for different architectures? Otherwise, I'll limit this Flatpak to x86_64 for the time being because supporting aarch64 at least is out of my reach.

so.notion.Notion.appdata.xml
e-user: I've been trying to make it work with extra-data and apply_extra for a while now, but it's a hot steaming mess - mostly because some of the npm stuff requires `g++` to be available, the patch requires the `patch` tool, the npm cache needs to be preserved inside the image.. just preparing node_modules during the build however also doesn't work because there's an internal repository referenced inside Notion's package.json.

Hence I've reached out to the company and asked for permission. If they decline, I'll try to make it work somehow.

e-user: Good news, they replied positively, saying that while they won't endorse Linux versions of Notion in the meantime, _the current terms don't disallow this_ and they are _fine ... to try it out_. Would you like to continue the review based on my last pushed commits?

nedrichards:
You can put modules into specific archictures (e.g. have one for x86 and one for aarch64 with different patches). there's a few examples around flathub with this pattern.

so.notion.Notion.appdata.xml
e-user: OK I got another email, they revoked their initial statement and are not _able to provide written permission to redistribute the program, as this creates potential issues with how we maintain, support, and troubleshoot the app_. I'm not sure I still want to continue at this point after that kind of unsupportive exchange. I'll try a bit more with the extra-data approach and if I can't make it work, close the request.

e-user:
I'm closing this PR because Notion doesn't want unlicensed clients ATM and the extra-data approach is a time sink I can't afford to waste more into right now.

PR #2003, opened 2020-12-20
Add io.github.Soundux

D3SOX:
This PR adds [Soundux](https://github.com/Soundux/Soundux)
I'm an upstream contributor and this is my first Flatpak, so criticism of the manifest etc is welcome


D3SOX:
bot, build io.github.Soundux

io.github.Soundux.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Type=Application
+Version=1.0
+Name=Soundux
+GenericName=Soundboard
+Exec=/app/bin/soundux %U
```
bilelmoussaoui, line 6:
```suggestion
Exec=soundux %U
```

io.github.Soundux.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
```
bilelmoussaoui, line 1: The desktop file, icons & metainfo/appdata files should be upstreamed if possible

io.github.Soundux.metainfo.xml
```diff
@@ -0,0 +1,27 @@
...
+  <screenshots>
+    <screenshot type="default">
+      <image>https://raw.githubusercontent.com/Soundux/soundux.github.io/master/src/assets/screenshots/1.png</image>
+    </screenshot>
+  </screenshots>
+  <releases>
+    <release version="0.1" date="2020-12-18"/>
+  </releases>
+  <content_rating type="oars-1.1">
+  </content_rating>
```
bilelmoussaoui, line 26:
```suggestion
  <content_rating type="oars-1.1" />
```

bilelmoussaoui:
bot, build  io.github.Soundux

bilelmoussaoui:
bot, build io.github.Soundux

io.github.Soundux.desktop
D3SOX: Okay, I'll commit them there and update the tag

D3SOX:
bot, build io.github.Soundux

D3SOX:
Oops, forgot to update the paths to the meta, icons and desktop files

D3SOX:
bot, build io.github.Soundux

io.github.Soundux.yml
```diff
@@ -0,0 +1,26 @@
+app-id: io.github.Soundux
+runtime: org.kde.Platform
+runtime-version: '5.15'
+sdk: org.kde.Sdk
+command: soundux
+finish-args:
+  - --socket=pulseaudio
+  - --socket=x11
```
nedrichards, line 8: you also need to --share=ipc if you use --socket=x11 (it allows x shared memory which is important for performance)

D3SOX: Thank you for mentioning. I didn't know that.

D3SOX:
bot, build io.github.Soundux

io.github.Soundux.yml
```diff
@@ -0,0 +1,26 @@
...
+  - --filesystem=home
+  - --filesystem=host:ro
+  - --talk-name=org.freedesktop.Flatpak
+modules:
+  - name: soundux
+    buildsystem: cmake
```
nedrichards, line 15: If it works then switching this to cmake-ninja saves a bunch of build time but it's just a nitpick and entirely your choice.

D3SOX: Thanks, it seems to work so I updated it

D3SOX:
bot, build io.github.Soundux

io.github.Soundux.yml
```diff
@@ -0,0 +1,27 @@
...
+runtime: org.kde.Platform
+runtime-version: '5.15'
+sdk: org.kde.Sdk
+command: soundux
+finish-args:
+  - --socket=pulseaudio
+  - --socket=x11
+  - --share=ipc
+  - --filesystem=home
+  - --filesystem=host:ro
```
barthalion, line 11: Just to double check, I understand there's no way to make it use portals?

io.github.Soundux.yml
```diff
@@ -0,0 +1,27 @@
...
+  - --talk-name=org.freedesktop.Flatpak
```
barthalion, line 12: What is it needed for?

D3SOX: I think this can be improved.
I added write access to home because it writes a config file. I also added read only access to root because maybe the user wants to add a folder that's under /run/media

barthalion: I mean `--talk-name=org.freedesktop.Flatpak` here.

io.github.Soundux.yml
barthalion: I'd limit it to xdg-music, xdg-download, /media and /run/media then, if it makes sense from your perspective.

io.github.Soundux.yml
D3SOX: We use the output of `pacmd` to get information from PulseAudio and move source outputs

io.github.Soundux.yml
D3SOX: Okay, sounds good. And for the write permissions to the `.config/Soundux` folder?

D3SOX:
I added `--device=dri` because without it I got the following console warnings:
```
libGL error: MESA-LOADER: failed to retrieve device information
libGL error: MESA-LOADER: failed to open amdgpu (search paths /usr/lib/x86_64-linux-gnu/GL/default/lib/dri)
libGL error: failed to load driver: amdgpu
libGL error: failed to open /dev/dri/card0: No such file or directory
libGL error: failed to load driver: radeonsi
```

io.github.Soundux.yml
D3SOX: I updated the filesystem permissions. I have to check if I can use portals in QtQuick.

D3SOX:
When testing I found a bug that works when using it without flatpak. The folder chooser seems to be a Qt file picker and you can't choose a folder with it. I'll have to investigate this before publishing this on Flathub.

io.github.Soundux.yml
barthalion: Wouldn't be better to bundle `pacdm` into the app? Running commands on host is major sandbox hole.

D3SOX: How can we achieve this? It is a command line utility from PulseAudio

barthalion:
```diff
diff --git a/io.github.Soundux.yml b/io.github.Soundux.yml
index 6795f7d..bd36fb2 100644
--- a/io.github.Soundux.yml
+++ b/io.github.Soundux.yml
@@ -13,8 +13,19 @@ finish-args:
   - --filesystem=/media:ro
   - --filesystem=/run/media:ro
   - --filesystem=~/.config/Soundux
-  - --talk-name=org.freedesktop.Flatpak
 modules:
+  - name: pacmd
+    buildsystem: simple
+    build-commands:
+      - ./autogen.sh
+      - make -C src pacmd
+      - install -Dm644 src/.libs/libpulsecommon-14.0.so /app/lib/libpulsecommon-14.0.so
+      - install -Dm755 src/.libs/pacmd /app/bin/pacmd
+    sources:
+      - type: git
+        url: https://github.com/pulseaudio/pulseaudio
+        tag: v14.0
+
   - name: soundux
     buildsystem: cmake-ninja
     config-opts:
```

Attempts to run pacmd give me this though:

```
No PulseAudio daemon running, or not running as session daemon.
```

So I guess that's not the way. Maybe something in pavucontrol source code would give you inspiration? It's available on Flathub: https://github.com/flathub/org.pulseaudio.pavucontrol

D3SOX: I think it's somehow possible with the `PULSE_SERVER` environment variable but I don't know what to set it to.
What I find weird is that `pactl` works just fine. (f.e. `pactl list source-outputs`). Maybe we should just switch to that

barthalion: `pactl list source-outputs` does not return anything for me (neither in flatpak or on the host) but if you tested it and it works properly, then yeah, makes sense to me.

D3SOX: It only returns something when you have an application open which listens to your microphone, e.g. Discord or TeamSpeak
`pactl list sources` should always return something

io.github.Soundux.yml
D3SOX: Are these okay now?

io.github.Soundux.yml
D3SOX: We adjusted it to use `pactl` instead of `pacmd` and it works now. I have removed the permission.

D3SOX:
bot, build io.github.Soundux

D3SOX:
I added a permission for `xdg-documents` because you might have your sounds in your documents folder (like me, because I have them in my Nextcloud)

D3SOX:
bot, build io.github.Soundux

D3SOX:
bot, build io.github.Soundux

io.github.Soundux.yml
```diff
@@ -0,0 +1,31 @@
...
+command: soundux
+finish-args:
+  - --device=dri
+  - --socket=pulseaudio
+  - --socket=x11
+  - --share=ipc
+  - --filesystem=xdg-music:ro
+  - --filesystem=xdg-download:ro
+  - --filesystem=xdg-documents:ro
+  - --filesystem=/media:ro
```
barthalion, line 14:
```suggestion
```

Duplicate, it's already there below.

io.github.Soundux.yml
```diff
@@ -0,0 +1,31 @@
...
+  - --filesystem=/run/media:ro
+  - --filesystem=/media:ro
+  - --filesystem=~/.config/Soundux
```
barthalion, line 17:
```suggestion
```

This is something that should generally work out of the box. Flatpak sets `XDG_CONFIG_HOME` to a directory writable by default.

D3SOX: I tried it but it doesn't work without it

D3SOX: I think this is because we save the config under `$HOME/.config/Soundux`

D3SOX: I adjust it to use `$XDG_CONFIG_HOME` if set

D3SOX:
bot, build io.github.Soundux

D3SOX:
I removed the permission. Now it's working without it. I hope now everything is okay and it's ready to publish

barthalion:
LGTM, thanks a lot for your patience!

/merge @Curve

PR #2004, opened 2020-12-20
org.gramps_project.Gramps

OzarkShepherd:
## Please confirm your submission meets all the criteria ##

I have read the guidelines at 
https://github.com/flathub/flathub/wiki/App-Requirements and https://github.com/flathub/flathub/wiki/App-Maintenance and this PR follows the instructions at https://github.com/flathub/flathub/wiki/App-Submission

I am using only the minimal set of permissions. If not, the reasons the application needs each permission are:

All content stored in the flathub is redistributible by any party. If not, the unredistributable parts are using an extra-data source type.

I am an upstream contributor to the project. If not I link to a request for upstream to submit their software to Flathub directly.

I own the domain used in the application id or the domain has a policy for delegating subdomains (e.g. GitHub, SourceForge).

Any additional patches or files have been submitted to the upstream projects concerned. If not, explain why.


org.gramps_project.Gramps.yml
```diff
@@ -0,0 +1,115 @@
+app-id: org.gramps_project.Gramps
+# flatpak-builder will not take com.gramps-project.Gramps
+runtime: org.gnome.Platform
+runtime-version: '3.38'
+sdk: org.gnome.Sdk
+command: /app/bin/gramps
+rename-icon: gramps
+finish-args:
+# needs home access to access documents and pictures
+  - --filesystem=home
```
tinywrkb, line 10: I'm not a reviewer but as a user, I would greatly appreciate a less permissive filesystem access.
```
  - --filesystem=xdg-documents
  - --filesystem=xdg-download
  - --filesystem=xdg-pictures
  - --persist=.gramps
```
Maybe you would be able to drop the persistent folder with wrapper script using setting `export GRAMPSHOME=$XDG_CONFIG_HOME/gramps`.

LeandroStanger:
bot, build org.gramps_project.Gramps

org.gramps_project.Gramps.yml
OzarkShepherd: Hi, I noticed that if restoring a backup in Gramps that contains media files, Gramps will make a new directory in the home directory for the media files, and the name varies depending on what the name of the backup is.  So it needs write access to the home directory for that.

nedrichards: Please note: there is a template in the PR that asked this exact question. It would have been helpful to look through the questions there and proactively fill them out to reduce back and forth during review.

tinywrkb: @OzarkShepherd I just tested this and with `GRAMPSHOME` set the default location in actually under `GRAMPSHOME`.
Swappshot Sun Dec 20 23:49:29 2020

OzarkShepherd: @tinywrkb If I understand correctly, you want me to include another script and have the Gramps manifest call the init script during installation to change the default home location?  I don't know how to do this and I don't want to be unresponsive, but I am concerned about further complicating things in a way that will reduce functionality for some users, especially less advanced users.  For example, someone might make a new directory in home just for genealogy (which I have done in the past) and then get confused why the Gramps flatpak won't access it while the distro-installed or python-installed versions could access it.  Also some people with poor memory and concentration like me will see the altered base path and have trouble finding the media directory again once the close button is clicked.  I really do appreciate the help here and from reddit in the past, though.

tinywrkb: What about setting  `--env=GRAMPSHOME=~/Gramps --filesystem=~/Gramps:create`?
This way you have a dedicated folder for Gramps, in a visible location, not in a hidden folder, you don't give permissive access to `~`, and you don't even need an init script (and patching the desktop file).  
It still changes the default behavior but considering upstream didn't step in and don't follow the XDG dirs spec then it's not a bad workaround, just needs to be properly tested. I think it's a packaging decision left for the maintainer.    
You can add a comment to the application page on Flathub about the default Gramps folder location.  

I will add that this doesn't affect me, I have a global `--nofilesystem=home` override because I find it unacceptable permission.  
Anyway, this is all I have to say about this issue.

org.gramps_project.Gramps.yml
```diff
@@ -0,0 +1,115 @@
...
+      - --enable-shared
+      - --enable-cxx
+      - --enable-dbm
+      - --enable-stl
+    sources:
+      - type: archive
+        url: https://download.oracle.com/berkeley-db/db-5.3.28.tar.gz
+        sha256: e0a992d740709892e81f9d93f06daf305cf73fb81b545afe72478043172c3628
+      - type: patch
+        path: atomic.patch
```
hfiguiere, line 34: You should add here in the sources this entry:
```yaml
- type: shell
  commands:
    - cp -p /usr/share/automake-*/config.{sub,guess} .
```

This should fix the build on `aarch64`

hfiguiere:
bot, build org.gramps_project.Gramps

org.gramps_project.Gramps.yml
OzarkShepherd: Thank you for the help.

org.gramps_project.Gramps.yml
```diff
@@ -0,0 +1,118 @@
...
+       - type: archive
+         url:  https://github.com/nzjrs/osm-gps-map/releases/download/1.1.0/osm-gps-map-1.1.0.tar.gz
+         sha256:  8f2ff865ed9ed9786cc5373c37b341b876958416139d0065ebb785cf88d33586
+
+  - name: exiv2Dependency
+    buildsystem: cmake
+    sources:
+      - type: archive
+        url:  https://www.exiv2.org/builds/exiv2-0.27.3-Source.tar.gz
+        sha256:  a79f5613812aa21755d578a297874fb59a85101e793edc64ec2c6bd994e3e778
```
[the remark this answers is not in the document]
OzarkShepherd: Thank you.  I copied and pasted the patches into my github as a test and added them to the manifest.  The url of exiv2 we use looks the same to me.  However, using the patches breaks the build, and I don't understand patches well enough to modify them myself.  How important is it for arch64 to be supported?  

[https://github.com/OzarkShepherd/flatpak/runs/1614006445?check_suite_focus=true]
> Building module exiv2Dependency in /home/runner/work/flatpak/flatpak/.flatpak-builder/build/exiv2Dependency-1
> ========================================================================
> Applying patch exiv2-arch-flags.patch
> patching file cmake/compilerFlags.cmake
> Hunk #1 FAILED at 26 (different line endings).
> 1 out of 1 hunk FAILED -- saving rejects to file cmake/compilerFlags.cmake.rej
> Error: module exiv2Dependency: Child process exited with code 1
> Error: Process completed with exit code 1.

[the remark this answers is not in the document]
hfiguiere: I tried it locally here and it did work.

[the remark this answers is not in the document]
hfiguiere: The patches you added don't apply because they are not the same as the one I linked to. Probably because the end of lines got changed.

And yes ARM 64 is very important.

org.gramps_project.Gramps.yml
```diff
@@ -0,0 +1,118 @@
...
+      - --enable-stl
+    sources:
+      - type: archive
+        url: https://download.oracle.com/berkeley-db/db-5.3.28.tar.gz
+        sha256: e0a992d740709892e81f9d93f06daf305cf73fb81b545afe72478043172c3628
+      - type: patch
+        path: atomic.patch
+      - type: shell
+        commands:
+          - cp -p /usr/share/automake-*/config.{sub,guess} .
```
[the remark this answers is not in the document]
OzarkShepherd: Ok, I changed that part, thank you.

org.gramps_project.Gramps.yml
```diff
@@ -0,0 +1,118 @@
+app-id: org.gramps_project.Gramps
+# flatpak-builder will not take com.gramps-project.Gramps
+runtime: org.gnome.Platform
+runtime-version: '3.38'
+sdk: org.gnome.Sdk
+command: /app/bin/gramps
```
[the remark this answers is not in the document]
OzarkShepherd: ok, I changed it back to just the name and tested it, thank you.

OzarkShepherd:
bot, build org.gramps_project.Gramps

org.gramps_project.Gramps.yml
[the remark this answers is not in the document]
OzarkShepherd: I removed the exiv2 and gexiv dependencies since they are optional and I couldn't figure out how to get them aarch64 compliant.

OzarkShepherd:
bot, build org.gramps_project.Gramps

org.gramps_project.Gramps.yml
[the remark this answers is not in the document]
hfiguiere: Making sure the patches are not changed by the process to add the patches to the repository. I had no problem.

They where failing to apply because the end of lines changed between the original an the one in the repository.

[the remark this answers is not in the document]
OzarkShepherd: How did you make sure that the end of lines did not change between the original and the new file in the repository?  I'm using OpenSUSE with Gnome, and copied the contents of your original for Darktable into a new file at github in my repository.  It looked identical and had the same number of lines, but the file size was less than 10 bytes off and it got an error.  Then I tried copying from RAW format, but the paste still changed the line endings and file size was off slightly and it got an error.  Then I manually changed the line endings to carriage returns at the end of lines to see if that was the difference, but still got an error.  I don't really know what the difference in environment is that is causing the line endings to change in the patch, or what the original line endings are supposed to be.  Perhaps are you using a different OS than me?  Microsoft apparently uses different line endings than Linux.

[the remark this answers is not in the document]
hfiguiere: I usually use wget or curl directly to get the file. No text editor involved.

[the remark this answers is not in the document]
hfiguiere: Downloading the raw file in the web browser should work too (albeit I didn't check)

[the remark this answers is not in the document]
OzarkShepherd: Downloading the RAW file in the browser seems to have worked, thank you

OzarkShepherd:
bot, build org.gramps_project.Gramps

nedrichards:
@OzarkShepherd asking the question in the PR template again since I don't see it linked in the thread (apologies if I missed this)

> I am an upstream contributor to the project. If not I link to a request for upstream to submit their software to Flathub directly.

Basically, if you're not upstream yourself have you contacted them about this? Flathub is intended to be a service for application developers to get their code directly to users, whilst there are also a fair few apps that are maintained by external people, like a Linux distribution that's not the preferred way - which is why we have this step.

OzarkShepherd:
I didn't realize that I needed to explicitly answer each question, so here they are:
1. I believe that the permissions are minimal for this app.
2. I don't understand if using Oracle BSDDB3 in the flatpak is stored on flathub, nor do I understand the "extra-data source type" definition, but we got the code for including it from another flatpak.  BSDDB3 is now owned by Oracle, and an older version is downloaded as part of building the Gramps flatpak.  Users have been encouraged to switch their databases to SQLite, but BSDDB3 is included for legacy users who haven't switched yet.  If BSDDB3 is no longer allowed at flathub, I could take it out.
3. I'm an upstream contributor to the gramps-project flatpak.  https://github.com/gramps-project/flatpak
4. The domain in the application ID is not mine, but I do contribute to the flatpak and am submitting this flatpak with the knowledge of the administrator.
5. All additional patches and files are included in this PR.

org.gramps_project.Gramps.yml
barthalion: @OzarkShepherd

The general idea is to keep application data as self-contained as possible, so `--persist=.gramps` is the bare minimum. By default, applications have write access to `$XDG_DATA_HOME` and it could be used as `$GRAMPSHOME` (`~/.var/app/org.gramps_project.Gramps/data`) instead; users can always modify these permissions with `flatpak override` or [Flatseal](https://flathub.org/apps/details/com.github.tchx84.Flatseal).

I will make these changes for you if you don't mind.

org.gramps_project.Gramps.desktop
```diff
@@ -0,0 +1,146 @@
+[Desktop Entry]
```
barthalion, line 1: This ideally should be maintained in Gramps Project repository as it can be used by other distributions/stores.

org.gramps_project.Gramps.metainfo.xml
```diff
@@ -0,0 +1,207 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
barthalion, line 1: This ideally should be maintained in Gramps Project repository as it can be used by other distributions/stores.

org.gramps_project.Gramps.yml
barthalion: Ah well, I don't have write access to your repo.

You can run `curl https://paste.xinu.at/Xpbx4stc | git am` locally to apply my changes.

org.gramps_project.Gramps.desktop
barthalion: I just noticed it already is; feel free to handle it later on.

org.gramps_project.Gramps.metainfo.xml
barthalion: I just noticed it already is; feel free to handle it later on.

org.gramps_project.Gramps.yml
OzarkShepherd: > The general idea is to keep application data as self-contained as possible, so --persist=.gramps is the bare minimum. By default, applications have write access to $XDG_DATA_HOME and it could be used as $GRAMPSHOME (~/.var/app/org.gramps_project.Gramps/data) instead; users can always modify these permissions with flatpak override or Flatseal.

In a genealogy program like Gramps, users tend to download documents like pdf's and images of old documents like birth and death records to use as evidence.  Without this evidence to back up their claims, some people consider the genealogy claims to be spurious and useless at best, and misinformation at worst.  The program also organizes this evidence into a database by associating them with people and events so that they can be easily checked.  That is why the genealogy program Gramps must have access to the Documents, Downloads, and Pictures directories as a default, so that the evidence can be accessed.  I personally have my evidence in the Documents folder, others use Pictures or Downloads.  Gramps uses .gramps for the database, but not the pictures and other evidence.  The evidence is not duplicated into the database, only referenced; duplication only happens when the database is backed up with media files included.  My backups tend to be about 500 MB, and my wife's are much bigger, so we aren't just discussing a few pictures.  Furthermore, the Gramps flatpak has metadata capability so that notes can be made about the pictures, like what they are from and how they were obtained, so write permission is needed.

While advanced users can use flatseal or terminal commands to alter permissions, not everyone who uses flatpaks is an advanced user.  So that is not an option for many users.  I didn't even know about flatseal until I started to research how to make a flatpak.  They will just be stuck with an flatpak with reduced functionality from the older version that they could obtain from their distro.  The reason I got interested in making a flatpak for Gramps is for users like me who still want a fully functional and up to date genealogy app, but want to use a stable distribution so they don't have to deal with instability in their OS.  There is the option of installing dependencies and building Gramps from source, but that is too advanced for some users and since it is a python installation, removing the old python files when upgrading can get messy.  I still have Gramps 5.1.2 installed through python on my OpenSUSE system, even though the current version is 5.1.3.  And another popular stable distribution, Debian 10, through normal channels is Gramps 5.0.1 according to the gramps-project wiki, even though backports has 5.1.2 which is still outdated.

org.gramps_project.Gramps.metainfo.xml
OzarkShepherd: I had to modify the metainfo.xml to go with flathub, and as I understand it needs to be in a specific place for flathub.  So that is why there is a separate metainfo file being moved in the flatpak manifest.  I am confused about what you want me to do.

org.gramps_project.Gramps.desktop
OzarkShepherd: I had to modify the desktop file to work with the flatpak, and as I understand it needs to be in a specific place for flathub.  So that is why there is a separate desktop file being moved in the flatpak manifest.  I am confused about what you want me to do.

org.gramps_project.Gramps.metainfo.xml
barthalion: These files are not specific to Flathub by any means, so any changes done there should be applicable to the one shipped within the project.

org.gramps_project.Gramps.yml
barthalion: All we can do is advise; as it's "yours" application, you get to decide what feels right.

barthalion:
Is there anyone besides you from Gramps Project who should be given write access to the Flathub repo?

org.gramps_project.Gramps.metainfo.xml
OzarkShepherd: The current version of Gramps is 5.1.3, but 5.2.0 is due out in a couple of months.  At that time the desktop and appdata file shipped with the project might be changed to work with the flatpak, but there are not currently any plans to make changes to the current release since it was released 4 months ago.  

Changes in the desktop file for this flatpak from the desktop file shipped with the release are:
1. different name
2. different icon name
3. modified executable line
4. different location

The appdata/metainfo file changes are:
1. renaming to fit flathub conventions
2. different location
3. id in line 3 is different
4. launchable in line 192 is different

OzarkShepherd:
> Is there anyone besides you from Gramps Project who should be given write access to the Flathub repo?

The administrator of gramps-project, Nick Hall, cloned the flathub repo for gramps-project, so I am not sure who else has write access to it besides probably him and me.

org.gramps_project.Gramps.yml
OzarkShepherd: I'm not trying to be difficult, I just want Gramps to be functional and can't find any good explanation of xdg online through google, that will ensure that the users will have r/w access to the necessary Documents, Downloads, and Pictures.  That is why I resist making the change.  I have read the flatpak manual and saw that r/w is default for xdg but I don't understand the point.  I understand that flathub wants minimum permissions for the sandbox, but I suspect a lot of users might run into problems if they don't have access to their documents.  I did advise the gramps-project administrator about a week ago for the changes proposed by tinywrkb, but am waiting for a response.

PR #2005, opened 2020-12-21
Add dev.tchx84.Portfolio

tchx84:
Notes:

Re. Permissions: this is a file manager for Linux mobile phones, therefore, it requires permission to the HOME directory and other common volumes mount points.


tchx84:
bot, build dev.tchx84.Portfolio

PR #1830, opened 2020-09-14
Add baseapp for building Wine on top of it

gasinvein:
This app contains dependencies used by Wine (as of 5.10+). It's intended to be reused as a baseapp for building apps that make use of Wine, like Lutris, Minigalaxy and such. Different apps may want different Wine flavors, but they have more or less common set of dependencies - thus, a common baseapp. Building a standalone Wine flatpak on top of it is also a possible use-case.

`generated/*.json` modules are procedural-generated from `modules/*.yml`.

PR #2022, opened 2021-01-03
Add com.usebottles.bottles

francescomasala:


LeandroStanger:
bot, build com.usebottles.bottles

francescomasala:
Updated the repo

barthalion:
bot, build com.usebottles.bottles

mirkobrombin:
Our application cannot be built on aarch64 due to compat32 dependencies. I added the **flathub.json** file instructing the `x86_64` architecture only, I don't know if this is the correct method.

nedrichards:
bot, build com.usebottles.bottles

barthalion:
One thing I'd change is not to hang/fail if a runner(?) hasn't been downloaded yet, as otherwise creating a bottle fails:

```
2021-01-04 12:39:53 ERROR: Error while running async job: <bound method BottlesRunner.async_create_bottle of <bottles.runner.BottlesRunner object at 0x7f5c6a42a280>>
Exception: list index out of range
  File "/app/share/bottles/bottles/utils.py", line 152, in target
    result = self.task_func(*args, **kwargs)
  File "/app/share/bottles/bottles/runner.py", line 807, in async_create_bottle
    if not runner: runner = self.runners_available[0]
```

com.usebottles.bottles.yml
```diff
@@ -0,0 +1,69 @@
...
+- --share=network
+- --device=dri
+- --share=ipc
+- --socket=fallback-x11
+- --socket=wayland
+- --socket=pulseaudio
+- --socket=session-bus
+- --filesystem=~/Games
+- --filesystem=~/wineprefixes
+- --filesystem=~/.Bottles
```
barthalion, line 22: It should use `--persist`.

com.usebottles.bottles.yml
```diff
@@ -0,0 +1,69 @@
...
+- --filesystem=~/.local/share/bottles:create
```
barthalion, line 23: Any reason to use this directory and not XDG_DATA_HOME?

com.usebottles.bottles.yml
```diff
@@ -0,0 +1,69 @@
...
+- --allow=multiarch
+- --share=network
+- --device=dri
+- --share=ipc
+- --socket=fallback-x11
+- --socket=wayland
+- --socket=pulseaudio
+- --socket=session-bus
+- --filesystem=~/Games
+- --filesystem=~/wineprefixes
```
barthalion, line 21: Similar question as with XDG_DATA_HOME.

com.usebottles.bottles.yml
```diff
@@ -0,0 +1,69 @@
...
+    path: ld.so.conf
+- name: bottles
+  builddir: true
+  buildsystem: meson
+  sources:
+  - type: git
+    url: https://github.com/bottlesdevs/Bottles
+    tag: 2.0.9.1
+build-options:
+  env: {}
```
barthalion, line 69:
```suggestion
```

com.usebottles.bottles.yml
```diff
@@ -0,0 +1,69 @@
...
+    directory: lib/i386-linux-gnu
+    version: '3.38'
+command: bottles
+finish-args:
+- --allow=multiarch
+- --share=network
+- --device=dri
+- --share=ipc
+- --socket=fallback-x11
+- --socket=wayland
```
barthalion, line 17: Do Wine applications work natively on Wayland these days? If not, I'd stick to `x11`.

mirkobrombin: No wine doesn't work with wayland at this time, my fault

com.usebottles.bottles.yml
mirkobrombin: This one is used to obtain wineprefixes from old Bottles version. We used a different path (.Bottles) in v1, now we use `.local/share/bottles` we don't really need to map it.

com.usebottles.bottles.yml
mirkobrombin: These are paths used by other wineprefix managers. We must be able to read them in order to import wineprefix from a manager external to ours. Maybe there is a better way to declare these paths?

mirkobrombin:
Now I think everything is correct, should it be objectionable to configure the paths with XDG_DATA_HOME, if an example is possible, thanks.

I also fixed the python code on bottle creation.

com.usebottles.bottles.yml
nedrichards: Is it just read only? If so, add :ro afterwards.

hadess:
If I understood correctly, the actual versions of Wine to run the software are all downloaded. Is there a way that at least one version of Wine could be "shipped with the app", for example, by basing it on top of:
https://github.com/flathub/flathub/pull/1830

(Let me know if I misunderstood, I tried to read the Wiki, but all the links were dead: https://github.com/bottlesdevs/Bottles/wiki)

mirkobrombin:
> If I understood correctly, the actual versions of Wine to run the software are all downloaded. Is there a way that at least one version of Wine could be "shipped with the app", for example, by basing it on top of:
> #1830
> 
> (Let me know if I misunderstood, I tried to read the Wiki, but all the links were dead: https://github.com/bottlesdevs/Bottles/wiki)

This is  the documentation for the old (v1) release. The new one download wine at first launch or from the Preferences tab. We are creating the new documentation because that is broken.

hadess:
> This is the documentation for the old (v1) release. The new one download wine at first launch or from the Preferences tab. We are creating the new documentation because that is broken.

Would there be a way to make it so at least one version of wine is already pre-packaged/installed in the Flatpak? It would make the initial use much better, and avoid downloading something in 2 parts, and would make sure that the default wine install is trusted (as it would be from the same source as the application itself).

mirkobrombin:
> > This is the documentation for the old (v1) release. The new one download wine at first launch or from the Preferences tab. We are creating the new documentation because that is broken.
> 
> Would there be a way to make it so at least one version of wine is already pre-packaged/installed in the Flatpak? It would make the initial use much better, and avoid downloading something in 2 parts, and would make sure that the default wine install is trusted (as it would be from the same source as the application itself).

Bottles' purpose is to offer the latest version of wine from the Lutris and Proton-GE repositories. We currently do not recognize and manage versions of Wine installed on the system or present in other locations outside of `.local/share/bottles/runners`.

The installation takes place via our download manager, the operation takes a maximum of one minute. No user operation is required other than a click.

hadess:
> The installation takes place via our download manager, the operation takes a maximum of one minute. No user operation is required other than a click.

It takes one minute on your presumably quite fast always on Internet, that's not the case for every user of Flathub, and we usually try to make it so that applications don't need to download large amounts of data on first start unless they're "internet" applications.

mirkobrombin:
> > The installation takes place via our download manager, the operation takes a maximum of one minute. No user operation is required other than a click.
> 
> It takes one minute on your presumably quite fast always on Internet, that's not the case for every user of Flathub, and we usually try to make it so that applications don't need to download large amounts of data on first start unless they're "internet" applications.

This could be done in the future, the code needs to be adapted to support wine from other paths. We are already working on our own repository from which to download wine but this will take some time and is (unfortunately) not the priority at the moment.

mirkobrombin:
Is there more we need to do to complete the merge?

Erick555:
@francescomasala @mirkobrombin sorry for late question but why does your app need to use Sdk as runtime? Do you compile something during app usage?

alvarlagerlof:
I'm not seeing logos or descriptions in Souk or GNOME Software

mirkobrombin:
> @francescomasala @mirkobrombin sorry for late question but why does your app need to use Sdk as runtime? Do you compile something during app usage?

This could be a mistake on our part, I check

alvarlagerlof:
> I'm not seeing logos or descriptions in Souk or GNOME Software

Nevermind it started popping up

PR #2034, opened 2021-01-08
Add Pixelitor image editor app

nbenitez:

### DIFF
--- a/io.sourceforge.Pixelitor.desktop
+++ b/io.sourceforge.Pixelitor.desktop
@@ -1,0 +1,12 @@
+[Desktop Entry]
+Name=Pixelitor
+GenericName=Image Editor
+Comment=Pixelitor is an open source image editor that supports layers, layer masks, text layers, filters, multiple undo etc.
+Exec=pixelitor
+Icon=io.sourceforge.Pixelitor
+Terminal=false
+Type=Application
+Keywords=pixelitor;graphic;design;illustration;painting;
+Categories=Graphics;2DGraphics;RasterGraphics;
+MimeType=application/x-pixelitor;image/openraster;image/jpeg;image/png;image/x-icon;image/gif;image/bmp;
+StartupNotify=false
--- a/io.sourceforge.Pixelitor.json
+++ b/io.sourceforge.Pixelitor.json
@@ -1,0 +1,139 @@
+{
+    "id": "io.sourceforge.Pixelitor",
+    "runtime": "org.freedesktop.Platform",
+    "runtime-version": "20.08",
+    "command": "pixelitor",
+    "sdk": "org.freedesktop.Sdk",
+    "sdk-extensions": ["org.freedesktop.Sdk.Extension.openjdk"],
+    "finish-args": [
+        "--env=PATH=/app/jre/bin:/app/bin:/usr/bin",
+        "--env=JAVA_HOME=/app/jre",
+        "--env=JAVA=/app/jre/bin/java",
+        "--share=ipc",
+        "--socket=x11",
+        "--socket=pulseaudio",
+        "--share=network",
+        "--device=dri",
+        "--filesystem=home",
+        "--persist=.java"
+    ],
+    "modules": [
+        {
+            "name" : "openjdk",
+            "buildsystem" : "simple",
+            "build-commands" : [ "/usr/lib/sdk/openjdk/install.sh" ]
+        },
+        {
+            "name": "xgetres",
+            "buildsystem": "simple",
+            "build-commands": [
+                "make",
+                "make install PREFIX=/app"
+            ],
+            "sources": [
+              {
+                "type": "git",
+                "url": "https://github.com/tamirzb/xgetres",
+                "tag": "1.0"
+              }
+            ]
+        },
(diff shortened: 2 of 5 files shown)


nedrichards:
bot, build io.sourceforge.Pixelitor

lbalazscs:
Hi, I am the author and maintainer of Pixelitor, and I am here as a reaction to https://github.com/lbalazscs/Pixelitor/pull/146
I'm OK with adding Pixelitor to flathub, but I am not familiar with flathub. Anyway, here are some comments:

- About permissions: obviously Pixelitor, being an image editor needs to read and write files. It uses the internet only for connecting to https://pixelitor.sourceforge.io/ in the "Check for updates" menu. This functionality is not necessary if flathub takes care of upgrading.
- About Java: the last released version needs Java 8+, but the current development version needs Java 15+. This could be a problem as the Pixelitor.json in this pull request mentions openjdk11. I plan to release a  new version in less than a month.
- There are some minor typos in this PR: it's pixelitor and not pixelator :)

lbalazscs:
I wouldn't mind becoming a maintainer for Pixelitor here, but I don't understand the rules yet. About Java I found http://blog.matbooth.co.uk/08-openjdk-in-flathub.html, based on that it seems that the JSON file should reference org.freedesktop.Sdk.Extension.openjdk and not org.freedesktop.Sdk.Extension.openjdk11.

barthalion:
About Java version, there are multiple versions available: https://github.com/flathub?q=openjdk&type=&language= with `org.freedesktop.Sdk.Extension.openjdk` being the one targeting 15 at the moment.

lbalazscs:
Nelson, I wonder why did you write that the flatpak directory should be included in the jar. As far as I understand, flatpak doesn't care about this, and you said this because your plan was that I configure these files for each release, and then you create the actual package by extracting them from the jar. I appreciate the help, but this could become quite boring for you, especially considering that one of my plans is to set up GitHub Actions to do regular  automatic beta builds. So probably the smarter solution would be if we fully automated the whole process. I think this should also be possible with GitHub Actions (I'm not familiar with that either, but I'm willing to learn it...).

I was also thinking about security, and I realized that Pixelitor is also writing files in the user's home directory (filter presets), and the user preferences as well. It can also start the default browser to open either the Pixelitor home page or Wikipedia pages. It can open the file manager to show the directory of a file (if you right-click on a tab). And a feature that is almost ready (I will commit it in a few days) is that it can use ImageMagick 7, if it is installed. Currently this works not by directly linking to the ImageMagick library, but by starting a command-line process, and piping files to its standard input, and reading its standard output. This is much simpler, but I read at https://docs.flatpak.org/en/latest/basic-concepts.html that "Access to other things, such as other processes, is deliberately not possible.". What if those other processes are started by the app, is it then possible? Anyway, this ImageMagick thing is not essential, even with normal installation it works only if the user also installs ImageMagick, but it helps advanced users to use all the file formats supported by ImageMagick.

superdreamkilla:
It is important 4.24 is made a flatpak. As it has a HUD search and preset features that are absent in 4.23.

barthalion:
Yeah, if extracting files from JARs is inconvenient I guess it may be better to just keep the metadata maintained here.

The manifest already grants access to entire /home. If opening browser/directory is handled via `xdg-open` then it's going to work fine. ImageMagick can be bundled inside the app and then it's going to be available for Pixelitor to use; there's many apps doing that already.

nbenitez:
> Nelson, I wonder why did you write that the flatpak directory should be included in the jar. As far as I understand, flatpak doesn't care about this, and you said this because your plan was that I configure these files for each release, and then you create the actual package by extracting them from the jar. I appreciate the help, but this could become quite boring for you, especially considering that one of my plans is to set up GitHub Actions to do regular automatic beta builds. So probably the smarter solution would be if we fully automated the whole process. I think this should also be possible with GitHub Actions (I'm not familiar with that either, but I'm willing to learn it...).
> 
> I was also thinking about security, and I realized that Pixelitor is also writing files in the user's home directory (filter presets), and the user preferences as well. It can also start the default browser to open either the Pixelitor home page or Wikipedia pages. It can open the file manager to show the directory of a file (if you right-click on a tab). And a feature that is almost ready (I will commit it in a few days) is that it can use ImageMagick 7, if it is installed. Currently this works not by directly linking to the ImageMagick library, but by starting a command-line process, and piping files to its standard input, and reading its standard output. This is much simpler, but I read at https://docs.flatpak.org/en/latest/basic-concepts.html that "Access to other things, such as other processes, is deliberately not possible.". What if those other processes are started by the app, is it then possible? Anyway, this ImageMagick thing is not essential, even with normal installation it works only if the user also installs ImageMagick, but it helps advanced users to use all the file formats supported by ImageMagick.

Because flathub maintainers encourage manifest file and .desktop file to be upstreamed, it's not my preference.

I encourage you and support you in being maintainer of your own software in flathub, that's the ideal situation. You're free to push-force to this PR, while you learn flathub packaging I can help you with it, I will now push-force to this PR to add ImageMagick, rename pixelator to pixelitor where appropiate, use jdk 15..

PR #2059, opened 2021-01-16
Add org.kicad.KiCad.Library.Templates

jmaibaum:

- [x] I am an upstream contributor to the project. Please also add @sethhillbrand and @imciner2 to the list of maintainers.
- [x] The KiCad project owns the domain used in the application ID.

### DIFF
--- a/org.kicad.KiCad.Library.Templates.yml
+++ b/org.kicad.KiCad.Library.Templates.yml
@@ -1,0 +1,25 @@
+id: org.kicad.KiCad.Library.Templates
+branch: stable
+runtime: org.kicad.KiCad
+runtime-version: stable
+sdk: org.freedesktop.Sdk//20.08
+build-extension: true
+separate-locales: false
+appstream-compose: false
+build-options:
+  prefix: /app/extensions/Library/Templates
+modules:
+- name: kicad-templates
+  buildsystem: cmake-ninja
+  post-install:
+  # Link kicad.pro template from main app into extension mountpoint
+  - ln -s /app/share/kicad/template/kicad.pro ${FLATPAK_DEST}/share/kicad/template/kicad.pro
+  - install -Dm644 --target-directory=${FLATPAK_DEST}/share/metainfo org.kicad.KiCad.Library.Templates.metainfo.xml
+  - appstream-compose --basename=org.kicad.KiCad.Library.Templates --prefix=${FLATPAK_DEST}
+    --origin=flatpak org.kicad.KiCad.Library.Templates
+  sources:
+  - type: archive
+    url: https://gitlab.com/kicad/libraries/kicad-templates/-/archive/5.1.9/kicad-templates-5.1.9.tar.gz
+    sha256: 0c1bf3d2e6d8d1056a5da6c1f7a173551c154b4bdaddb86b6a34155b18e65da6
+  - type: file
+    path: org.kicad.KiCad.Library.Templates.metainfo.xml
--- a/flathub.json
+++ b/flathub.json
@@ -1,0 +1,3 @@
+{
+    "skip-icons-check": true
+}
--- a/org.kicad.KiCad.Library.Templates.metainfo.xml
+++ b/org.kicad.KiCad.Library.Templates.metainfo.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="addon">
+  <id>org.kicad.KiCad.Library.Templates</id>
+  <extends>org.kicad.KiCad</extends>
+  <name>KiCad Templates</name>
+  <summary>Official KiCad project and worksheet templates</summary>
+  <url type="homepage">https://gitlab.com/kicad/libraries/kicad-templates</url>
+  <project_license>CC-BY-SA-4.0</project_license>
+  <metadata_license>CC-BY-SA-4.0</metadata_license>
+  <update_contact>jmaibaum_AT_gmail.com</update_contact>
+</component>


jmaibaum:
bot, build org.kicad.KiCad.Library.Templates

org.kicad.KiCad.Library.Templates.yml
```diff
@@ -0,0 +1,40 @@
...
+    - mv kicad.pro template/
+    - rm -rf kicad-templates.tar.gz Projects/ Worksheets/
+    dest-filename: apply_extra
+  - type: extra-data
+    filename: kicad-templates.tar.gz
+    url: https://gitlab.com/kicad/libraries/kicad-templates/-/archive/5.1.9/kicad-templates-5.1.9.tar.gz
+    sha256: 0c1bf3d2e6d8d1056a5da6c1f7a173551c154b4bdaddb86b6a34155b18e65da6
+    size: 1643954
+    installed-size: 5877944
+  - type: extra-data
```
barthalion, line 34: Why does it use extra-data instead of embedding it into the extension?

org.kicad.KiCad.Library.Templates.yml
```diff
@@ -0,0 +1,40 @@
+id: org.kicad.KiCad.Library.Templates
+branch: '5.1'
```
barthalion, line 2: It should target `stable`; we don't "support" multiple branches for apps and apps extensions.

barthalion:
The same comments apply to the other 2 PRs so let's keep the discussion here.

org.kicad.KiCad.Library.Templates.yml
jmaibaum: Ok, didn't know that before, and it has worked OK so far for org.kicad.KiCad.Packages3D. Will fix it later.

org.kicad.KiCad.Library.Templates.yml
jmaibaum: I used extra-data, because for the huge existing .Packages3D extension (~900MB), the aarch64 builds timed out in about 9 out of 10 tries. When I switched this to extra-data, it worked 100% of the time.

Coming from this, I thought I had to use extra-data as well for the new extensions in order to make merge-dirs and the directory layout work.

I will happily switch all of them to store the data inside the extensions themselves. This would be so much easier, really. But I fear that this would break aarch64 builds again.

Do you know how to avoid timing out builds with larger builds?

jmaibaum:
Hi @barthalion,

I commented above.

Incidentally, the aarch64 build timeout issues seem to have been reappearing when I fixed the .Debug extension for org.kicad.KiCad (see #1 over there), which seem to have increased that package build to become larger again (around ~800MB) again. I had to rebuild both the PR and the build on the master branch, because the aarch64 build timed out each time. So, to me this looks like a general issue with large builds on the aarch64 buildbot runners, and keeping the extensions small by using extra-data seems to be a workaround...

(I'm on mobile right now, sorry for not posting proper links.)

barthalion:
@jmaibaum Mind changing one of the PRs to non-`extra-data` version? I'll investigate.

jmaibaum:
@barthalion Since I think the issue is related to the final size of the flatpak, I have added this draft PR for org.kicad.KiCad.Library.Packages3D: https://github.com/flathub/org.kicad.KiCad.Library.Packages3D/pull/1

Let's see how this goes.

jmaibaum:
The proposed 3 new extensions are all "tiny" (sub 50M) in size compared to the 3D packages.

jmaibaum:
@barthalion The test build for the PR linked above shows the known issue again: aarch64 timed out during the generate-deltas step.

barthalion:
Running the delta step manually now on the aarch64 server just to check how long it takes, but I guess that Packages3D is a corner case: `Content Bytes Written: 5533414486 (5.5 GB)`. Things below 2GB should be generally fine and benefit from static and in-between deltas for smaller downloads; extra-data means that given module is always downloaded from scratch.

barthalion:
And on that note, even if `--generate-static-deltas` was optional, the initial installation would be rather terrible as every file would need to be downloaded separately. So extra-data really might be the correct choice here.

(Also yes, still waiting for the command to finish.)

barthalion:
```
real	29m59.752s
user	29m32.059s
sys	0m27.533s
```

I guess I can just extend it to 40 minutes.

jmaibaum:
Thanks for checking, @barthalion . Though I am not quite sure if I got everything right now.

You say that extra-data is OK for Packages3D, because of 5.5G via normal module download vs. ~900M download via extra-data, but we should be using a non-extra-data approach for the remaining Library extensions (those currently in these PRs)? Or does your increasing the timeout for Packages3D mean that we should be changing back everything, including Packages3D, to non-extra-data?

If your answer to the former question is yes, I will have to double check if I can manage to merge all the necessary dirs for both extra-data and non-extra-data extensions so that the main app finds all the files in the right places. Not sure right now if this is easily possible or not without changes to KiCad itself.

Is it possible to get rid of the `<extension-mount-point>/extra/` subpath for extra-data extensions on the flatpak manifest side? Or is this mandatory?

jmaibaum:
A few more datapoints for you to consider:

1. I guess that even while the initial download of 5.5G is nasty for users, static delta download for updates between versions should be good, as I don't think that lots of data will change between say KiCad 5.1.9 and a possible upcoming 5.1.10 release.

2. The Packages3D files will be stored in compressed form on disk for the upcoming KiCad 6.0 release, so that in the future, Packages3D should become significantly smaller than 5.5G again.

barthalion:
@jmaibaum In my opinion it's fine to switch everything including Packages3D to "native" (not extra-data) module types.

jmaibaum:
@barthalion I am going to change all the extensions then, and update the PRs.

org.kicad.KiCad.Library.Templates.yml
jmaibaum: @barthalion Could you please elaborate why flathub doesn't "support" branches? The extension endpoint definition in  `org.kicad.KiCad` was using branch `5.1` before, and this was not considered an issue, when I submitted `org.kicad.KiCad.Library.Packages3D` with this setting.

Given this, changing everything to move to `stable` would now mean to alter both the extension(s) and the main app.

For reference, here is the current definition of the extension mountpoint from `org.kicad.KiCad`:
```
add-extensions:
  org.kicad.KiCad.Library:
    version: '5.1'
    directory: extensions/Library
    autodelete: true
    no-autodownload: true
    subdirectories: true
    merge-dirs: share/kicad/template
```

barthalion: Applications target only the `stable` branch, so there is no reason to version an extension if it's not going to be used by any other application. This usually passes reviews because there's no linter yet so the process is prone to oversights.

jmaibaum: Ok, the I will move all of the proposed extensions, and the mount point to `stable` instead.

jmaibaum:
PRs are all "native" modules again, and have been moved to `stable`, so they need https://github.com/flathub/org.kicad.KiCad/pull/4 to be loaded (test build pending).

jmaibaum:
bot, build org.kicad.KiCad.Library.Templates

jmaibaum:
@barthalion I have now completed all your proposed changes for all 3 PRs, and changed the main app's extension point (merged to master, build/deploy still pending). So I now consider all 3 extension PRs ready for merging/adding.

PR #2075, opened 2021-01-22
Initial servo

BryanQuigley:
Just pulling the nightly build
no automation

WIP (mostly works)-  just for more testing. May switch to using a build process instead.

- [] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://github.com/servo/servo/issues/27942

### DIFF
--- a/org.servo.Servo.yaml
+++ b/org.servo.Servo.yaml
@@ -1,0 +1,31 @@
+app-id: org.servo.Servo
+runtime: org.freedesktop.Platform
+runtime-version: '20.08'
+sdk: org.freedesktop.Sdk
+command: servo
+finish-args:
+  - --socket=x11
+  - --device=dri
+  - --socket=pulseaudio
+  - --share=ipc
+  - --share=network
+modules:
+  - name: servo
+    buildsystem: simple
+    build-commands:
+       - mkdir -p /app/bin/resources/
+       - cp -R resources/* /app/bin/resources/
+       - install -D resources/servo.svg /app/share/icons/hicolor/scalable/apps/${FLATPAK_ID}.svg
+      # servo_1024.png, servo_64.png, servo.ico
+       - install -D servo /app/bin/servo
+       - install -D org.servo.Servo.desktop /app/share/applications/${FLATPAK_ID}.desktop
+       - install -D org.servo.Servo.metainfo.xml /app/share/metainfo/${FLATPAK_ID}.metainfo.xml
+    sources:
+      - type: archive
+        url: https://download.servo.org/nightly/linux/servo-latest.tar.gz
+        sha256: 2ff0a2558cb1f4cee1b3382f1877a5e0c59f11758433ea4a5553f425784c50ef
+        #https://download.servo.org/nightly/linux/servo-latest.tar.gz.sha256
+      - type: file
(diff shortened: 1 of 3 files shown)


BryanQuigley:
bot, build org.servo.Servo

org.servo.Servo.yaml
```diff
@@ -0,0 +1,31 @@
...
+       - mkdir -p /app/bin/resources/
+       - cp -R resources/* /app/bin/resources/
+       - install -D resources/servo.svg /app/share/icons/hicolor/scalable/apps/${FLATPAK_ID}.svg
+      # servo_1024.png, servo_64.png, servo.ico
+       - install -D servo /app/bin/servo
+       - install -D org.servo.Servo.desktop /app/share/applications/${FLATPAK_ID}.desktop
+       - install -D org.servo.Servo.metainfo.xml /app/share/metainfo/${FLATPAK_ID}.metainfo.xml
+    sources:
+      - type: archive
+        url: https://download.servo.org/nightly/linux/servo-latest.tar.gz
```
bilelmoussaoui, line 25: Can't we build it here instead of using the binary blob directly?

BryanQuigley: Being attempted here https://github.com/servo/servo/issues/27942#issuecomment-765905405

barthalion:
Is it expected to be useful for regular users? My understanding always was that Servo is not supposed to become a full-fledged browser.

BryanQuigley:
> Is it expected to be useful for regular users? My understanding always was that Servo is not supposed to become a full-fledged browser.

It is very much not ready as anything but a demo right now - and it's focus is on embedded.  But it could develop more in the future - see https://github.com/servo/servo/wiki/Roadmap.

nedrichards:
Given that it's likely suitable for flathub beta right now rather than the main "stable" branch.

BryanQuigley:
That was a general question I had - there is no "Beta" version of Servo - it is what it is today.   Should it just be made available in the stable repo as the "Stable" Servo is a preview release?

org.servo.Servo.yaml
hfiguiere: It builds fine from the confines of Thunderbird, except when you only have 4GB of RAM (not a problem on the building infra)

This would also allow `aarch64` support, which Mozilla infra doesn't do:
```
% file servo/servo
servo/servo: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=5161d2c948c592b019a98261171c40c140254830, with debug_info, not stripped
```

Do the Flathub builders really let the build succeed on aarch64? It seems that it doesn't care but the Flatpak surely will never run.

Also by using this URL you make in non-reproducible. If I try to build the Flatpak right now it will tell me it is the wrong sha.

BryanQuigley: Sorry, meant to close this PR in favor of the https://github.com/servo/servo/pull/28088

BryanQuigley:
Closing in favor of https://github.com/servo/servo/pull/28088

PR #2090, opened 2021-01-29
Added app manifest for fi.skyjake.Lagrange

skyjake:

### DIFF
--- a/fi.skyjake.Lagrange.json
+++ b/fi.skyjake.Lagrange.json
@@ -21,14 +21,15 @@
             "buildsystem": "cmake-ninja",
             "config-opts": [
                 "-DCMAKE_BUILD_TYPE=RelWithDebInfo",
-                "-DCMAKE_INSTALL_PREFIX=/app"
+                "-DCMAKE_INSTALL_PREFIX=/app",
+                "-DENABLE_DOWNLOAD_EDIT=NO"
             ],
             "sources": [
                 {
                     "type": "git",
                     "url": "https://github.com/skyjake/lagrange",
-                    "tag": "v1.1.0",
-                    "commit": "abcd2daa61c0ef5abf542c3e681657672858885b"
+                    "tag": "v1.1.1",
+                    "commit": "f7f335d0d5b4f488e9707c18a2128190a4549738"
                 }
             ]
         }
--- a/fi.skyjake.Lagrange.json
+++ b/fi.skyjake.Lagrange.json
@@ -1,0 +1,35 @@
+{
+    "app-id": "fi.skyjake.Lagrange",
+    "default-branch": "release",
+    "runtime": "org.freedesktop.Platform",
+    "runtime-version": "20.08",
+    "sdk": "org.freedesktop.Sdk",
+    "command": "lagrange",
+    "rename-desktop-file": "fi.skyjake.lagrange.desktop",
+    "rename-icon": "fi.skyjake.lagrange",
+    "finish-args": [
+        "--socket=x11",
+        "--socket=pulseaudio",
+        "--share=network",
+        "--device=dri",
+        "--filesystem=home"
+    ],
+    "modules": [
+        {
+            "name": "lagrange",
+            "buildsystem": "cmake",
+            "config-opts": [
+                "-DCMAKE_BUILD_TYPE=RelWithDebInfo",
+                "-DCMAKE_INSTALL_PREFIX=/app"
+            ],
+            "sources": [
+                {
+                    "type": "git",
+                    "url": "https://github.com/skyjake/lagrange",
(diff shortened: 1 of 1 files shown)


fi.skyjake.Lagrange.json
```diff
@@ -0,0 +1,35 @@
...
+    "sdk": "org.freedesktop.Sdk",
+    "command": "lagrange",
+    "rename-desktop-file": "fi.skyjake.lagrange.desktop",
+    "rename-icon": "fi.skyjake.lagrange",
+    "finish-args": [
+        "--socket=x11",
+        "--socket=pulseaudio",
+        "--share=network",
+        "--device=dri",
+        "--filesystem=home"
```
barthalion, line 15: What does it need home access for?

fi.skyjake.Lagrange.json
```diff
@@ -0,0 +1,35 @@
...
+    ],
+    "modules": [
+        {
+            "name": "lagrange",
+            "buildsystem": "cmake",
```
barthalion, line 20:
```suggestion
            "buildsystem": "cmake-ninja",
```

barthalion:
bot, build fi.skyjake.Lagrange

fi.skyjake.Lagrange.json
barthalion: If this is about downloads directory, I'd rather change it to `xdg-download` (and make sure the app follows that environment variable).

skyjake: Yes, the app needs write access to a downloads directory, so `--filesystem=xdg-download` would work for that. Do you suggest checking a `XDG_DOWNLOAD_DIR` environment variable?

The app also needs write access to its XDG config directory. Does that need a separate permission flag or is that allowed by default? Can I rely on `XDG_CONFIG_HOME` being set?

The app needs read-only access to any drag-and-dropped file based on its path. This could be anywhere under the user's home directory, so could `--filesystem=home:ro` be acceptable for that?

nedrichards:
bot, build fi.skyjake.Lagrange

fi.skyjake.Lagrange.json
barthalion: Yeah, I think following XDG spec is worth regardless of flatpak. `XDG_CONFIG_HOME` (and `XDG_DATA_HOME`) are places your app has write access by default. As for drag&drop, I guess `home:ro` is fine as otherwise it will break.

fi.skyjake.Lagrange.json
```diff
@@ -0,0 +1,35 @@
+{
+    "app-id": "fi.skyjake.Lagrange",
+    "default-branch": "release",
```
nedrichards, line 3: this shouldn't be in here for a flathub build - everything in flathub is on the "stable" branch.

fi.skyjake.Lagrange.json
```diff
@@ -0,0 +1,35 @@
...
+    "runtime": "org.freedesktop.Platform",
+    "runtime-version": "20.08",
+    "sdk": "org.freedesktop.Sdk",
+    "command": "lagrange",
+    "rename-desktop-file": "fi.skyjake.lagrange.desktop",
+    "rename-icon": "fi.skyjake.lagrange",
+    "finish-args": [
+        "--socket=x11",
```
nedrichards, line 11: you also need to pass --share=ipc if you use --socket=x11 to allow x acceleration to work properly

nedrichards:
App looks really nice - would be even more awesome if it was hidpi aware but that's not a blocker! Other than the tweaks mentioned above lgtm.

fi.skyjake.Lagrange.json
barthalion: AFAIK `default-branch` is the correct thing if someone wants a sane default for local builds – this gets overwritten as opposed to just `branch` which is not.

skyjake:
> App looks really nice - would be even more awesome if it was hidpi aware but that's not a blocker! Other than the tweaks mentioned above lgtm.

Thanks! Any suggestions how to correctly detect HiDPI when running under Flatpak? I'm not using Qt or GNOME, but I could look at environment variables and the X DPI factor I suppose.

fi.skyjake.Lagrange.json
nedrichards: Hrm, yeah - that's a point. I've always thought that the ergonomics of switching branches are slightly sub-optimal and observe that most local testing seems to happen with different repos (e.g. flathub in system and changes in --user) but yeah, you've got a point.

nedrichards:
bot, build fi.skyjake.Lagrange

skyjake:
bot, build fi.skyjake.Lagrange

PR #2121, opened 2021-02-18
Add org.nmap.Zenmap

Eonfge:

---

Hey all, 

I thought that adding Zenmap to Flathub would be a good way of introducing new users to the power of nmap. Nothing beats the power of the CLI version, but this graphical version allows people to quickly get started with penetration testing. Now... as for the submission rules.

## Upstream involvement

First, I'm not actually the first person to look into this, and upstream has already stated that they're not directly interested, although they seem to be open for future suggestions related to the FreeDesktop AppStream standard.

**Issue on the topic:**
https://github.com/nmap/nmap/issues/1295

## Permissions

It's a security penetration tool written in Python 2 and Gtk 2. I can't do any better then just giving it home access. File browsers are not an option for example. The application also warns that without root permissions, it's not at its full potential. Really, this package is mostly just there to keep Python 2 and Gtk 2 out of your main system.

## Patches

There is a file py2gtk2.json which might actually be a good candidate for shared-modules. It combines all elements that you need to link python 2 and gtk 2, who are themselves already shared-modules.


Eonfge:
bot, build org.nmap.zenmap

Eonfge:
@LeandroStanger Feel free to chip in ;)

Eonfge:
bot, build org.nmap.Zenmap

Eonfge:
bot, build org.nmap.Zenmap

Eonfge:
bot, build org.nmap.Zenmap

barthalion:
I'd change the default "unprivileged is worse" message to something about limitations related to Flatpak so people don't try to run it with sudo.

nedrichards:
Yeah, other than @barthalion's comment - which I think we should do, this lgtm

org.nmap.Zenmap.desktop
```diff
@@ -0,0 +1,11 @@
+[Desktop Entry]
+Name=Zenmap
+GenericName=GUI Port Scanner
+TryExec=zenmap
+Exec=zenmap %F
+Terminal=false
+Icon=org.nmap.Zenmap
+Type=Application
+Categories=Application;Network;Security;
+Comment=A cross-platform GUI for the Nmap Security Scanner.
```
AsciiWolf, line 10: Comments should ideally not end with a dot.

org.nmap.Zenmap.metainfo.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2019 Kevin Degeling -->
+
+<component type="desktop-application">
+  <id>org.nmap.Zenmap</id>
+  <name>Zenmap</name>
+  <summary>Zenmap is the official Nmap Security Scanner GUI.</summary>
```
AsciiWolf, line 7: Ditto.

AsciiWolf: Also, a shorter summary like "The official Nmap Security Scanner GUI", or the same text as in the desktop file Comment ("A cross-platform GUI for the Nmap Security Scanner") would be better. ;-)

Eonfge:
This is the official message that Zenmap shows when starting as non-root:
Screenshot from 2021-02-19 16-26-17

Should I suppress the message completely? That would require a minor code-patch removing 4 lines of Python. I honestly think that it's fine to just keep the message as-is: Zenmap is not a general purpose tool after all.

Eonfge:
bot, build org.nmap.Zenmap

AsciiWolf:
> Should I suppress the message completely? That would require a minor code-patch removing 4 lines of Python. I honestly think that it's fine to just keep the message as-is: Zenmap is not a general purpose tool after all.

I would remove that message or change its text to an information about the Flatpak version.

Eonfge:
bot, build org.nmap.Zenmap

@AsciiWolf Was able to pull it off in a one-liner

nedrichards:
With this latest test build I still get the warning popup. That's Flatpak 1.8.2 running on Endless OS 3.9.2 @AsciiWolf FYI

Eonfge:
bot, build org.nmap.Zenmap

---

That's likely because they're still on 1.8.x, while I'm on Fedora with 1.10.x

I made a second patch who fixes this on all versions, but I don't think we should upstream that one. It's only here to support older version of Flatpak. It should work with all Flatpak versions, even insecure ones like 1.6.x

nedrichards:
Agreed that we shouldn't upstream that version, but for us - latest stats show that approx 80% of active users are currently using 1.8 or older https://ahayzen.com/direct/flathub.html so it's worth carrying here for a little while.

Eonfge:
Those are some depressing numbers. So 60% of the people runs 1.6 or lower? Those versions are not even officially supported any more and don't get security fixes... and they sandbox for shit since most portal functionality doesn't work.

https://repology.org/project/flatpak/versions :disappointed:

AsciiWolf:
Yeah, for example Linux Mint still use Flatpak 1.6.x. :-/

nedrichards:
@Eonfge Yeah "stable distros exist" ;-) We're at a late part in the cycle as you see two waves of upgrades a year (around April and around October) and also Debian Bullseye is due out soonish. Endless will probably do their 4.0 release at some point too (I suspect that they're most of the 1.8 versions)

You can absolutely use --require-version to ensure you get the version of Flatpak you want (and I'd encourage you to do so!) but there's always a tradeoff with reach.

nedrichards:
/merge

I'm delighted to see this merged since I can now delete the half done flatpak of zenmap I've had hanging around for the past three years. Thanks!

PR #2140, opened 2021-03-01
Add com.github.sqlitebrowser.sqlitebrowser

Qwertie-:

  - `<content_rating>` and `<release>` are missing from appdata in the tagged release. Its in master but not tagged yet. Is this a problem?
  - `--filesystem=home` The app does sort of support portals but there are errors after opening files. FS access makes sense until that can be resolved
  - `--share=network` The only place I can see this used is for printing. Removing network access only allows the "Print to file" option. I guess if people want to lock this down more they can use flatseal but better to have the default config with everything working.
- [ ] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link: https://github.com/sqlitebrowser/sqlitebrowser/issues/2103**
  - Its a github repo I do not control
  - No patches used

### DIFF
--- a/com.github.sqlitebrowser.sqlitebrowser.desktop.appdata.xml
+++ b/com.github.sqlitebrowser.sqlitebrowser.desktop.appdata.xml
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id>sqlitebrowser.desktop</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>MPL-2.0 and GPL-3.0+</project_license>
+  <name>DB Browser for SQLite</name>
+  <summary>DB Browser for SQLite is a light GUI editor for SQLite databases</summary>
+  <description>
+    <p>DB Browser for SQLite is a high quality, visual, open source tool to create, design, and edit database files compatible with SQLite.</p>
+    <p>It is for users and developers wanting to create databases, search, and edit data. It uses a familiar spreadsheet-like interface, and you don't need to learn complicated SQL commands.</p>
+    <p>Controls and wizards are available for users to:</p>
+    <ul>
+      <li>Create and compact database files</li>
+      <li>Create, define, modify and delete tables</li>
+      <li>Create, define and delete indexes</li>
+      <li>Browse, edit, add and delete records</li>
+      <li>Search records</li>
+      <li>Import and export records as text</li>
+      <li>Import and export tables from/to CSV files</li>
+      <li>Import and export databases from/to SQL dump files</li>
+      <li>Issue SQL queries and inspect the results</li>
+      <li>Examine a log of all SQL commands issued by the application</li>
+    </ul>
+  </description>
+  <content_rating type="oars-1.1" />
+  <screenshots>
+    <screenshot type="default">
+      <image>https://raw.githubusercontent.com/sqlitebrowser/db4s-screenshots/master/v3.3/gnome3_2-execute.png</image>
+      <caption>DB Browser for SQLite, executing query</caption>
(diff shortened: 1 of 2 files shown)


nedrichards:
@Qwertie- you'll need to take the appdata changes as a patch otherwise I suspect it'll fail the validator but we can find out.

nedrichards:
bot, build com.github.sqlitebrowser.sqlitebrowser

Qwertie-:
Looks like the spec couldn't find the appdata file at all. So its possible to copy the app data in to this branch to be used instead of the one in the main repo?

com.github.sqlitebrowser.sqlitebrowser.yaml
```diff
@@ -0,0 +1,22 @@
+app-id: com.github.sqlitebrowser.sqlitebrowser
+runtime: org.kde.Platform
+runtime-version: '5.15'
+sdk: org.kde.Sdk
+command: sqlitebrowser
+rename-icon: sqlitebrowser
+rename-appdata-file: sqlitebrowser.appdata.xml
```
bilelmoussaoui, line 7:
```suggestion
rename-appdata-file: sqlitebrowser.desktop.appdata.xml
```
That's the correct file name upstream, you probably want to rename those upstream

barthalion:
bot, build com.github.sqlitebrowser.sqlitebrowser

barthalion:
Appdata needs some fixes:

```
builddir/files/share/appdata/com.github.sqlitebrowser.sqlitebrowser.appdata.xml: FAILED:
```

Qwertie-:
@barthalion Those fixes are in master but not in the tagged release. I don't control the upstream repo so I can't fix this. Is it possible we can include the latest appdata file in this PR

nedrichards:
@Qwertie- in that case you should ideally apply the upstream change as a patch and then drop it once upstream makes a new release.

barthalion:
@Qwertie- Yeah, ideally just apply files here "downstream" for now. Also DRI permission is fine.

hfiguiere:
A patch for the appdata would make this app closer to submission.

Qwertie-:
I'm not certain how to patch the file but I'll give it a shot looking at some of the other pull requests as a guide.

com.github.sqlitebrowser.sqlitebrowser.desktop.appdata.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id>sqlitebrowser.desktop</id>
```
hfiguiere, line 3: the id is wrong. (it was handled previously by the `rename-appdata-file`)

hfiguiere:
bot, build com.github.sqlitebrowser.sqlitebrowser

hfiguiere:
The appdata isn't referenced in sources, nor installed. Patching it would be much better. Since you already have a Pull Request with upstream, the hardest is already done with a creating a patch. It's just a matter of adding the source patch entry.

Qwertie-:
@hfiguiere I'm not quite sure what you are suggesting here? Do you mean leaving the app data filename as it is and keeping the yml config to rename it?

hfiguiere:
As said previously by @nedrichards: 
> @Qwertie- in that case you should ideally apply the upstream change as a patch and then drop it once upstream makes a new release.

Appstream file is fixed in upstream. So the proper approach is to take that patch, add it to the manifest to patch the sources.

Then when a new release is out with the changes, just drop the patch.

frhun:
Your pr is based on the wrong branch!
It should be based on new-pr.

Mailaender:
Superseded by #3110.

PR #2142, opened 2021-03-01
Add com.valvesoftware.SteamLink — standalone Steam Link client

smcv:
This is a standalone client for the Steam Link (Steam Remote Play) protocol, similar to the versions available for Android, iOS and Raspberry Pi. Unlike the [full Steam client](https://github.com/flathub/com.valvesoftware.Steam), it doesn't launch games locally. Instead, it connects to a computer running the full Steam client, either remote-controlling it like a game-optimized equivalent of VNC/RDP (Steam Remote Play) or joining a multiplayer game as a guest (Steam Remote Play Together).

---

    - `--device=all` is for game controller support. Steam Link prefers to use raw HID, as allowed by the [steam-devices](https://github.com/ValveSoftware/steam-devices/) udev rules, where possible, to provide the most faithful possible controller passthrough. It falls back to `evdev` if that is not available.
    - All other permissions used are on Flathub's "always OK" list.
- [ ] All assets referenced in the manifest are redistributable by any party.
    - Steam Link is proprietary and not redistributable without permission. Valve representatives have already contacted the Flathub admins to arrange Flathub-specific permission to redistribute this.
    - Sort of. I'm a consultant submitting this on behalf of Valve (cc @TTimo).
- [x] We own the domain used in the application ID
    - I'm submitting this on behalf of Valve, who own the domain
- [ ] Any additional patches or files have been submitted to the upstream projects concerned
    - Steam Link needs a patched Qt with expanded controller support, matching the one used on non-Linux OSs (built from patched source code here). I'm not sure what the upstream status of this is.
    - Steam Link needs a patched SDL with expanded controller support, matching the one used on non-Linux OSs (built by Valve and included here as binaries). My understanding is that some of the changes are non-upstreamable and are effectively a "friendly fork".
    - Steam Link needs a patched FFmpeg matching the one used on non-Linux OSs and in the [full Steam client](https://github.com/flathub/com.valvesoftware.Steam) (built by Valve and included here as binaries, source code [available elsewhere](https://media.steampowered.com/steamlink/src/ffmpeg.tar.gz)).

## Other notes

@TTimo should be given full access to the new repository (unless I can do that myself).

This is a remote-access app, analogous to a VNC/RDP client, so it provides access to whatever Steam games you happen to have in your library, and their suitability for children will vary widely. The current metadata refers only to what this app itself does, and does not attempt to take into account the games that might be accessed through it. This seems to be consistent with the [equivalent Android app](https://play.google.com/store/apps/details?id=com.valvesoftware.steamlink), which is rated PEGI 3.

This is currently x86_64-only, and I've tried to set up `flathub.json` accordingly. ARM builds of essentially the same software exist, for [Raspberry Pi](https://support.steampowered.com/kb_article.php?ref=6153-IFGH-6589) and the discontinued Steam Link hardware device, so it might be possible to enable ARM later - but my understanding is that the current Raspberry Pi version expects to run directly on a VT, whereas this x86_64 build expects to run as an ordinary full-screen app under X11 or Xwayland, so an ARM Flatpak build for X11/Wayland would require Valve to produce new binaries in parallel with the Raspberry Pi build.


smcv:
bot, build com.valvesoftware.SteamLink

com.valvesoftware.SteamLink.appdata.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.valvesoftware.SteamLink</id>
+  <launchable type="desktop-id">com.valvesoftware.SteamLink.desktop</launchable>
+  <metadata_license>MIT</metadata_license>
+  <project_license>LicenseRef-Proprietary</project_license>
+  <name>Steam Link</name>
+  <summary>Stream games from another computer with Steam</summary>
```
bilelmoussaoui, line 8: The metainfo needs at least one screenshot

smcv: Fixed. It would be great if the application submission info explicitly said either this, or which validator(s) the metainfo must pass (or perhaps it does and I missed it).

smcv:
bot, build com.valvesoftware.SteamLink

com.valvesoftware.SteamLink.yml
```diff
@@ -0,0 +1,208 @@
...
+  - "--env=SDL_GAMECONTROLLERCONFIG_FILE=/var/data/Valve Corporation/SteamLink/controller_map.txt"
+  - --share=ipc
+  - --share=network
+  - --socket=pulseaudio
+  - --socket=wayland
+  - --socket=x11
+
+tags:
+  - proprietary
+  - beta
```
barthalion, line 19: Should it be merged to the beta branch or it's a leftover?

com.valvesoftware.SteamLink.yml
```diff
@@ -0,0 +1,208 @@
...
+      arch:
+        x86_64:
+          config-opts:
+            - "-reduce-relocations"
+            - "-force-debug-info"
+        arm:
+          config-opts:
+            - "-no-reduce-relocations"
+            - "-optimize-size"
+            - "-opengl es2"
```
barthalion, line 66:
```suggestion
```

`armhf` is no longer supported.

com.valvesoftware.SteamLink.yml
```diff
@@ -0,0 +1,208 @@
...
+      - type: dir
+        path: icons
+    buildsystem: simple
+    build-commands:
+      - install -d /app/share/applications/
+      - install -d /app/share/metainfo/
+      - install -D -m644 com.valvesoftware.SteamLink.appdata.xml /app/share/metainfo/
+      - install -D -m644 com.valvesoftware.SteamLink.desktop /app/share/applications/
+      - |
+        set -eu; for size in 16 24 32 48 256; do
```
barthalion, line 204: I think it's fine to stick to 128x128 and 256x256.

com.valvesoftware.SteamLink.yml
barthalion: (I know it's doesn't matter as per PR description but just so it doesn't give a wrong idea.)

com.valvesoftware.SteamLink.yml
```diff
@@ -0,0 +1,208 @@
...
+    buildsystem: simple
+    build-commands:
+      - "install bin/steamlink /app/bin"
+      - "cp -a lib/* /app/lib"
+      - "cp -a *.txt ThirdPartyLegalNotices.* /app"
+
+  - name: packaging
+    sources:
+      - type: file
+        path: com.valvesoftware.SteamLink.appdata.xml
```
LeandroStanger, line 192:
```suggestion
        path: com.valvesoftware.SteamLink.metainfo.xml
```
New appdata files should use metainfo instead. https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html

com.valvesoftware.SteamLink.yml
smcv: These are the icons I got from Valve. I can drop the smaller versions if you want, but I'll have to ask for a 128px version (if the build doesn't automatically scale it down from 256).

barthalion: It should scale down the biggest one so it's fine.

com.valvesoftware.SteamLink.appdata.xml
bilelmoussaoui: Our documentation isn't great for now, but we have the appdata guidlines as part of the wiki https://github.com/flathub/flathub/wiki/AppData-Guidelines along with how to validate it locally. The build bot does validate it as well along with the desktop file & the icons.

com.valvesoftware.SteamLink.yml
smcv: This is the first ever public release of Steam Link for desktop Linux, so if it's possible to create the repository and release it to a beta branch, in such a way that we can re-release as stable later without needing your intervention, then that would probably be best.

Or if that's not possible, I can remove the `beta` tag and we can go directly to calling it a stable release?

com.valvesoftware.SteamLink.yml
smcv: I'd slightly prefer to keep this in sync with the closest equivalent KDE runtime (which we're not using because we need to patch Qt anyway, and because it pulls in a lot of higher-level KDE stuff that we don't need). But if you feel strongly about this, I can drop it.

com.valvesoftware.SteamLink.yml
smcv: For some context on this, Valve has historically released everything on desktop Linux as generic, hopefully-portable binaries built on some suitably old operating system, which we then have to try to support on everyone's favourite distributions, each of which is different. Obviously this doesn't scale very well.

For the Steam Link app, we're trying a different approach where the first public release is already a Flatpak app, and we get the cross-distro portability from Flatpak/Flathub/the fd.o runtime.

smcv:
bot, build

smcv:
bot, build com.valvesoftware.SteamLink

com.valvesoftware.SteamLink.yml
barthalion: Yes, we can merge it to the beta branch, which will make it available in `flathub-beta`. When you feel confident to publish it to "stable" Flathub, you just need to create the `master` branch in the git repository.

smcv: That would be perfect.

Are there any other blockers that I need to fix before this can land as a beta?

com.valvesoftware.SteamLink.yml
smcv: The build doesn't seem to do any scaling automatically, so the desktop environment will be responsible for scaling the 256px icon down to its preferred size. I assume that's OK? This seems to be the same thing that will happen for `com.valvesoftware.Steam`.

If that's a problem, I can either mechanically scale it down to 128px during build, or go back to Valve UI people and ask for a "perfect" 128px icon (or a SVG, but I don't know whether the source material for the icon exists as a SVG).

smcv:
@barthalion Please could you also invite @slouken to https://github.com/flathub/com.valvesoftware.SteamLink, or point me towards how we can do that ourselves if we can?

barthalion:
@smcv I have changed your permissions so you should be able to invite people on your own, I guess that will make things easier. Also invited @slouken.

com.valvesoftware.SteamLink.yml
barthalion: It's nice to have once it moves to stable, but I guess at least GNOME just picks up the biggest available icon or SVG, so I don't consider it a blocker.

PR #2154, opened 2021-03-06
Add com.sireliah.Dragit

sireliah:

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. [Dragit repository](https://github.com/sireliah/dragit)


sireliah:
bot, build com.sireliah.Dragit

bilelmoussaoui:
Your appdata/metainfo file lacks the following required tags
```
```

sireliah:
Thanks, I'm already fixing that.

sireliah:
bot, build com.sireliah.Dragit

sireliah:
Cool, I provided only scalable icons, but it seems 128x128 and 256x256 are required too. I'll add them.

bilelmoussaoui:
> Cool, I provided only scalable icons, but it seems 128x128 and 256x256 are required too. I'll add them.

The issue you have is probably due to https://github.com/sireliah/dragit/blob/master/com.sireliah.Dragit.metainfo.xml#L22, the desktop file name you're mentioning there is wrong.

sireliah:
bot, build com.sireliah.Dragit

nedrichards:
Is there a particular reason this is x86-64 only? My other testing device is aarch64 ;-)

com.sireliah.Dragit.yml
```diff
@@ -0,0 +1,33 @@
+app-id: com.sireliah.Dragit
+runtime: org.freedesktop.Platform
+runtime-version: '20.08'
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=x11
```
bilelmoussaoui, line 10:
```suggestion
  - --socket=fallback-x11
```

com.sireliah.Dragit.yml
```diff
@@ -0,0 +1,33 @@
...
+modules:
+  - name: dragit
+    buildsystem: simple
+    cleanup:
+      - "/include"
+      - "*.a"
+    build-commands:
+      - install -Dm644 static/logo_icon_t.svg /app/share/icons/hicolor/scalable/apps/com.sireliah.Dragit.svg
+      - install -Dm644 static/logo_icon_t_128x128.png /app/share/icons/hicolor/128x128/apps/com.sireliah.Dragit.png
+      - install -Dm644 static/logo_icon_t_256x256.png /app/share/icons/hicolor/256x256/apps/com.sireliah.Dragit.png
```
bilelmoussaoui, line 25: a scalable icon should be enough

sireliah:
> Is there a particular reason this is x86-64 only? My other testing device is aarch64 ;-)

No particular reason, except that I've built the app only on `x86_64` and I don't know if other architectures will work. But I'll remove the restriction, that's good idea.

sireliah:
bot, build com.sireliah.Dragit

nedrichards:
OK! I've got around to testing this. Due to the way this app works with drag and drop you're going to want to have --filesystem=home permissions. Without that you get a lot of:
> [2021-03-08T20:32:53Z ERROR dragit::dnd::components] Could not extract dragged content: Os { code: 2, kind: NotFound, message: "No such file or directory" }

And the app will basically not work.

sireliah:
@nedrichards I've updated the permissions, however I'm curious how I'm my local setup is handling the permissions:
```bash
$ flatpak permissions
(...)
documents     86d83d8d     com.sireliah.Dragit          read,write,grant-permissions (b'/home/sireliah/Files', 64770, 8610945, 4)
```
This directory is set using Gtk `FileChooser`, so I though that it's the portal's doing: `org.freedesktop.portal.FileChooser`. When I change directory (menu in the top left), the new directory is appended to permissions and app works without `--filesystem=home`.

I'm asking, because I would like to avoid granting home permissions if possible.

sireliah:
bot, build com.sireliah.Dragit

nedrichards:
Yes, when you change directory that should work if you have xdg-desktop-portal 1.7.1 or later which is when directory access gets added. At least one of my devices has an older version than that though. I *think* that If you don't get the user to explicitly select the directory though you don't get access, so out of the box it's broken. If you want to avoid the permission (which would be a good thing to avoid!) then you'd probably have to put a "choose where you want files to be accessed" step in the first run but this is a good point, I will investigate a little further.

sireliah:
Ok, I see. Well, this app is a file sharing program, so it needs to both read from home (when sending file) and write to it (when saving downloaded). I think it's fine to keep filesystem=home then.

PR #2236, opened 2021-04-05
Org.tribler.tribler

overflw:

_(this is my first flatpak, apologies for stupid questions/mistakes)_

  - I tried my best to use minimal permissions, but maybe one could reduce them even more.
- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. [Link](https://github.com/Tribler/tribler/issues/4938)
  - Would it be better to use the GitHub domain of tribler instead of its .org? I own neither of them.
  - Sorry, I don't understand this point.


LeandroStanger:
bot, build Org.tribler.trible

overflw:
> bot, build Org.tribler.trible

I think there is a "r" missing. I guess it should be `org.tribler.Tribler`.

overflw:
Can I start the bot too? 
bot, build Org.tribler.tribler

overflw:
bot, build org.tribler.Tribler

overflw:
bot, build org.tribler.Tribler

overflw:
bot, build org.tribler.Tribler

overflw:
bot, build org.tribler.Tribler

overflw:
bot, build org.tribler.Tribler

overflw:
bot, build org.tribler.Tribler

overflw:
bot, build org.tribler.Tribler

overflw:
bot, build org.tribler.Tribler

barthalion:
Have you contacted upstream team if they're interested to co-maintain their app on Flathub?

overflw:
I asked in the [tribler issue](https://github.com/Tribler/tribler/issues/4938).

hfiguiere:
Any reason to make it x86_64 only? The deb is `all` arch.

overflw:
> Any reason to make it x86_64 only? The deb is `all` arch.

I just didn't have the hardware to test other arches - if you think it should just work, we can totally take that limitation out.

overflw:
Btw, is it normal that the screenshots specified in metainfo.xml are not showing up in my gnome-software when i install via
`flatpak install --user https://dl.flathub.org/build-repo/42384/org.tribler.Tribler.flatpakref` ?

overflw:
@xoriole from tribler is joining as co-maintainer.

overflw:
bot, build org.tribler.Tribler

hfiguiere:
The last commit didn't do what it said it did. Still restrict to x86_64.

hfiguiere:
it's not enough since you made the source file x86_64 only in the manifest.

overflw:
So tribler usually interacts with vlc to provide streaming - how could we realise this with flatpaks? I'm still a bit lost on how to do this, as I understand it, we would need to install vlc in the same flatpak because flatpak.spawn is not an option while using the .deb as base.

hfiguiere:
FreeCAD use flatpak-spawn to call openSCAD as a flatpak for conversion purpose. And it does that through a wrapper script.

So you could have a 'vlc' script to do that.

Now maybe you just gave an argument to build from source instead of using the .deb

overflw:
Will give the script approach a try.

barthalion:
bot, build org.tribler.Tribler

org.tribler.Tribler.yaml
```diff
@@ -0,0 +1,64 @@
...
+runtime-version: '20.08'
+sdk: org.freedesktop.Sdk
+separate-locales: false
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --device=dri
+  - --socket=pulseaudio
+  - --share=network
+  - --filesystem=xdg-download
```
barthalion, line 13:
```suggestion
  - --filesystem=xdg-download
```

The default download path Tribler suggests me is `/var/home/bpiotrowski/TriblerDownloads`.

org.tribler.Tribler.metainfo.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
barthalion, line 1: Ideally this should be submitted and maintained upstream.

org.tribler.Tribler.yaml
```diff
@@ -0,0 +1,64 @@
...
+      - install -Dm644 org.tribler.Tribler.metainfo.xml -t /app/share/metainfo
+    sources:
+      - type: file
+        url: https://github.com/Tribler/tribler/releases/download/v7.8.0/tribler_7.8.0_all.deb
+        sha256: 22ac6be8a06024dd58f1759c89816100461543f593056f697b2c40d851df85a4
+        dest-filename: tribler.deb
+      - type: file
+        path: README.md
+      - type: dir
+        path: icons
```
barthalion, line 41: Are these files unavailable in the deb package?

README.md
```diff
@@ -0,0 +1,17 @@
+Tribler
```
barthalion, line 1: Just so you are aware, this file is very unlikely to be ever seen by a user. This description belongs to appdata file and not here.

org.tribler.Tribler.yaml
```diff
@@ -0,0 +1,64 @@
...
+      - |
+        for size in 64 128 256 512; do
+          install -Dm644 tribler_${size}.png /app/share/icons/hicolor/${size}x${size}/apps/org.tribler.Tribler.png || exit 1
+        done
+      - mv usr/share/tribler /app/share
+      - install -Dm644 README.md -t /app/share/tribler
+      - install -Dm644 org.tribler.Tribler.metainfo.xml -t /app/share/metainfo
+    sources:
+      - type: file
+        url: https://github.com/Tribler/tribler/releases/download/v7.8.0/tribler_7.8.0_all.deb
```
barthalion, line 35:
```
[📦 org.tribler.Tribler ~]$ file /app/share/tribler/tribler
/app/share/tribler/tribler: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=1da3a1d77c7109ce6444919f4a15e7e6c63d02fa, stripped
```

This is definitely x86_64-only, just the deb is built in a wrong way. If https://github.com/Tribler/tribler/blob/main/doc/development/development_on_linux.rst is up to date, it should be easy to build it from source with https://github.com/flatpak/flatpak-builder-tools/blob/master/pip/flatpak-pip-generator

org.tribler.Tribler.yaml
```diff
@@ -0,0 +1,64 @@
+id: org.tribler.Tribler
+command: tribler
+runtime: org.freedesktop.Platform
+runtime-version: '20.08'
+sdk: org.freedesktop.Sdk
+separate-locales: false
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
```
barthalion, line 9:
```suggestion
  - --socket=x11
```

`fallback-x11` should be used only if Wayland is also supported.

org.tribler.Tribler.yaml
overflw: Interesting, on Fedora and Ubuntu /home/<user>/Downloads/TriblerDownloads was selected. What could be an elegant solution? Possibly avoiding `--filesystem=home`?

org.tribler.Tribler.metainfo.xml
overflw: Agreed! I'll propose it, but I'm not part of tribler.

org.tribler.Tribler.yaml
overflw: Unfortunately yes, there is only a very low resolution png in the deb.

README.md
overflw: Sure! I guess we should just delete it.

org.tribler.Tribler.yaml
overflw: Thought so, I'll limit it to x86 again then. 
I tried building it from source, but this being my first flatpak I had to give up after some hours of trial and error. Maybe I'll give it another try later but for now I don't have the resources unfortunately.

org.tribler.Tribler.yaml
overflw: We probably would need to install the QtWayland module to support Wayland, lets remove the fallback for now.

hfiguiere: If that app uses Qt, then you should build against the KDE platform instead.

org.tribler.Tribler.yaml
hfiguiere: it's really better to build from source.

overflw: Can you elaborate a bit further? Why is it so much better? Especially keeping all dependencies up to date seems to me an unnecessary source of complication, when I can use a deb which is maintained by the tribler team.

overflw:
> So tribler usually interacts with vlc to provide streaming - how could we realise this with flatpaks? I'm still a bit lost on how to do this, as I understand it, we would need to install vlc in the same flatpak because `flatpak.spawn` is not an option while using the .deb as base
Tribler stopped offering streaming and doesn't need vlc anymore. So this problem is solved.

org.tribler.Tribler.yaml
hfiguiere: To be able to use in on `aarch64`.

overflw: Is there an easy way to generate a manifest from apt packages? Like it works for python packages via the flatpak-pip-generator?

hfiguiere: short: no

longer: I don't think .deb have necessary metadata if someone was willing to write such a tool. Also python is much more contained and simpler to deal deal with than arbitrary dependencies.

overflw: I was thinking about generating a manifest from a list of package names e.g. from the ubuntu or debian repositories, not from a .deb directly. This really shouldn't be any more difficult than the pip based case.. Or am I missing something?

hfiguiere: it wouldn't tell how to build them.

barthalion:
bot, build org.tribler.Tribler

org.tribler.Tribler.yaml
```diff
@@ -0,0 +1,65 @@
...
+      - install -Dm644 org.tribler.Tribler.metainfo.xml -t /app/share/metainfo
+    sources:
+      - type: file
+        url: https://github.com/Tribler/tribler/releases/download/v7.8.0/tribler_7.8.0_all.deb
+        sha256: 22ac6be8a06024dd58f1759c89816100461543f593056f697b2c40d851df85a4
+        dest-filename: tribler.deb
+        only-arches:
+          - x86_64
+      - type: file
+        path: README.md
```
barthalion, line 40:
```suggestion
```

overflw:
bot, build org.tribler.Tribler

barthalion:
/merge @xoriole

Please let us know if someone more from Tribler team should be invited to the Flathub repo.

PR #2244, opened 2021-04-08
Add org.getmonero.monero.desktop

roundtheworldman:

* [Y]  I have read the [App Requirements](https://github.com/flathub/flathub/wiki/App-Requirements) and [App Maintenance](https://github.com/flathub/flathub/wiki/App-Maintenance) pages.

* [Y]  My pull request follows the instructions at [App Submission](https://github.com/flathub/flathub/wiki/App-Submission).

* [Y]  I am using only the minimal set of permissions. _(If not, please explain each non-standard permission.)_

* [Y]  All assets referenced in the manifest are redistributable by any party.  If not, the unredistributable parts are using an extra-data source type.

* [Cant contact developers ] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub.

* [N]  I own the domain used in the application ID or the domain has a policy for delegating subdomains

* [No Patches]  Any additional patches or files have been submitted to the upstream projects concerned. _(If not, explain why.)_


roundtheworldman:
bot, build org.getmonero.monero.desktop

roundtheworldman:
bot build org.getmonero.monero.desktop/org.getmonero.Monero.yaml

roundtheworldman:
bot build org.getmonero.monero.desktop org.getmonero.Monero.yaml

org.getmonero.Monero.yaml
```diff
@@ -0,0 +1,31 @@
...
+runtime: org.kde.Platform
+runtime-version: '5.15'
+sdk: org.kde.Sdk
+command: monero-wallet-gui
+modules:
+  - name: monero-bin
+    buildsystem: simple
+    sources:
+    - type: archive
+      url: https://downloads.getmonero.org/gui/monero-gui-linux-x64-v0.17.1.9.tar.bz2
```
barthalion, line 16: Stating the obvious but it's still not built from source.

barthalion:
bot, build org.getmonero.Monero

org.getmonero.monero.desktop.xml
```diff
@@ -0,0 +1,64 @@
+
```
barthalion, line 1: You're not installing this file. It needs to be defined in `sources` array and installed in `build-commands`.

roundtheworldman:
Hi @barthalion, thanks for the info. I am brand new to flatpak packaging. I do realize the URL above is a download to a pre-compiled release for Linux x86_84.

However, I just wanted to see what the build bot said when I added the appstream xml file to the PR.
My main question at the moment is how to tell the build bot that the App description information is in a yaml file, not a json file.

For example, bitwarden app uses yaml file, and Spotify is using a json file.

https://github.com/flathub/com.bitwarden.desktop/blob/master/com.bitwarden.desktop.yaml
vs.
https://github.com/flathub/com.spotify.Client/blob/master/com.spotify.Client.json

roundtheworldman:
bot, build org.getmonero.monero.desktop

roundtheworldman:
The build bot is still looking for a json file, even though a yaml file is present:
```
Can't load 'org.getmonero.monero.desktop.json': Failed to open file “/var/lib/private/buildbot-master/master/workers/MasterWorkerUntrusted/Builds/build/org.getmonero.monero.desktop.json”: No such file or directory
program finished with exit code 1
```

roundtheworldman:
bot, build org.getmonero.Monero

roundtheworldman:
bot, build org.getmonero.Monero

org.getmonero.monero.desktop.xml
hfiguiere: also its name should be `org.getmonero.Monero.metainfo.xml` (note the case)

org.getmonero.monero.desktop.xml
```diff
@@ -0,0 +1,64 @@
...
+<?xml version="1.0" encoding="UTF-8" ?>
+<!-- Copyright 2018-2021 Flathub maintainers -->
+<component type="desktop">
+  <id>org.getmonero.monero.desktop</id>
```
hfiguiere, line 5: id should be `org.getmonero.Monero` (no `.destkop` and case sensitive)

roundtheworldman: I was under the impression that if an application has more than 1 variant, that using a keyword like Desktop or GUI is preferred?
Monero also has a CLI app. And a daemon.

e.g. https://monerodocs.org/interacting/monero-wallet-cli-reference/

I thought this is why apps on Flathub like Bitwarden use the name desktop in the identifier?

e.g. com.bitwarden.desktop
Bitwarden has a Desktop App for 3 different OS. Also, a CLI app, browser extensions, and mobile OS applications.
https://raw.githubusercontent.com/flathub/com.bitwarden.desktop/master/com.bitwarden.desktop.yaml

org.getmonero.Monero.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - name: monero-bin
+    buildsystem: simple
+    sources:
+    - type: archive
+      url: https://github.com/monero-project/monero-gui/archive/refs/tags/v0.17.1.9.tar.gz
+      sha256: 3a924b4e08b557d93337a0c05826a9fa95e5cf8c503aff0ecb3676be45db1d14
+    build-commands:
+    - tar xzf v0.17.1.9.tar.gz
+    - cd monero-gui-0.17.1.9
+    - docker build --tag monero:build-env-linux --build-arg THREADS=4 --file Dockerfile.linux .
```
hfiguiere, line 21: I doubt this will work. Also there is no network access in the build sandbox

org.getmonero.Monero.yaml
```diff
@@ -0,0 +1,34 @@
+app-id: org.getmonero.monero.desktop
```
hfiguiere, line 1: Same comment as elsewhere. org.getmonero.Monero is the id, given the name the manifest.

org.getmonero.monero.desktop.xml
hfiguiere: It's not according to the name of the manifest (which has the wrong id anyway). Also it's confusing, but that's just my opinion.

org.getmonero.Monero.yaml
roundtheworldman: I added the `docker build` line and `docker run` line because that is what the official Monero Desktop GUI build instructions say right now.

```
### Building Reproducible Linux static binaries with Docker (any OS)

1. Install Docker [https://docs.docker.com/engine/install/](https://docs.docker.com/engine/install/)
2. Clone the repository
   ```
   git clone --branch master --recursive https://github.com/monero-project/monero-gui.git
   ```
   \* `master` - replace with the desired version tag (e.g. `v0.17.1.9`) to build the release binaries.
3. Prepare build environment
   ```
   cd monero-gui
   docker build --tag monero:build-env-linux --build-arg THREADS=4 --file Dockerfile.linux .
   ```
   \* `4` - number of CPU threads to use

4. Build
   ```
   docker run --rm -it -v <MONERO_GUI_DIR_FULL_PATH>:/monero-gui -w /monero-gui monero:build-env-linux sh -c 'make release-static -j4'
   ```
   \* `<MONERO_GUI_DIR_FULL_PATH>` - absolute path to `monero-gui` directory  
   \* `4` - number of CPU threads to use
5. Monero GUI Linux static binaries will be placed in  `monero-gui/build/release/bin` directory
```

Ref: https://raw.githubusercontent.com/monero-project/monero-gui/master/README.md

It seems like it is better to use their pre-compiled builds for Linux x86_64 if there is not proper network access in the flathub build sandbox.

hfiguiere: Nope. it's better to do without Docker.

hfiguiere:
The flatpak documentation also say that "desktop" as a bad ID:

https://docs.flatpak.org/en/latest/conventions.html#application-ids

> org.example.desktop
> This is a bad ID because the Appstream standard for legacy reasons treats IDs ending with .desktop as a special case causing inconsistency. For this same reason, .Desktop suffixes should not be used for newly named applications. Don’t hesitate to repeat the application name even if it already is part of the domain name section of the identifier (eg. org.example.Example).

Also this one as one too many components.

The original ID from the original submission was fine.

roundtheworldman:
Regarding building from source...
If the build sandbox really does not have any network access, it seems that this might be near impossible to build with flathub build bot.
This application requires some build dependencies that the app developers outline in their README.md file.
They also instruct to do a git clone, which will also use https in this case.

hfiguiere:
> Regarding building from source...
> If the build sandbox really does not have any network access, it seems that this might be near impossible to build with flathub build bot.
> This application requires some build dependencies that the app developers outline in their README.md file.

They shall be added to the manifest. That's what the download phase is for.

> They also instruct to do a git clone, which will also use https in this case.

Not a problem either. Source are downloaded in the download phase. Many flatpak build out of the git source.

roundtheworldman:
bot, build org.getmonero.Monero

org.getmonero.Monero.yaml
```diff
@@ -18,10 +18,11 @@ modules:
...
     build-commands:
-    - tar xzf v0.17.1.9.tar.gz
-    - cd monero-gui-0.17.1.9
-    - docker build --tag monero:build-env-linux --build-arg THREADS=4 --file Dockerfile.linux .
-    - docker run --rm -v $PWD:/monero-gui -w /monero-gui monero:build-env-linux sh -c 'make release-static -j4'
+    - sudo apt install build-essential cmake libboost-all-dev miniupnpc libunbound-dev graphviz doxygen libunwind8-dev pkg-config libssl-dev libzmq3-dev libsodium-dev libhidapi-dev libnorm-dev libusb-1.0-0-dev libpgm-dev libprotobuf-dev protobuf-compiler libgcrypt20-dev
+    - sudo apt install qtbase5-dev qt5-default qtdeclarative5-dev qml-module-qtqml-models2 qml-module-qtquick-controls qml-module-qtquick-controls2 qml-module-qtquick-dialogs qml-module-qtquick-xmllistmodel qml-module-qt-labs-settings qml-module-qt-labs-folderlistmodel qttools5-dev-tools qml-module-qtquick-templates2 libqt5svg5-dev
+    - git clone --recursive https://github.com/monero-project/monero-gui.git
+    - cd monero-gui
+    - make release -j $FLATPAK_BUILDER_N_JOBS
```
hfiguiere, line 25: This is not how it is done. It won't work either. There is no apt-get or debian package in the platform.

Dependencies are listed as modules in the manifest, one by one.

hfiguiere: Also since this require Qt, you should build against the KDE platform.

org.getmonero.Monero.xml
```diff
@@ -2,12 +2,12 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!-- Copyright 2018-2021 Flathub maintainers -->
 <component type="desktop">
-  <id>org.getmonero.monero.desktop</id>
-  <launchable type="desktop-id">org.getmonero.monero-wallet.desktop</launchable>
```
hfiguiere: org.getmoneor.MoneroWallet would also be acceptable.

This file, the appstream metadata, should be named `org.getmonero.Monero.metainfo.xml`

roundtheworldman:
Hi @hfiguiere Thanks again for all the help and support here.
I will close this PR since it is getting rather messy here and my complete lack of experience with build bot))

I will create a new PR here shortly from a different branch name of my flathub git clone.

org.getmonero.Monero.xml
roundtheworldman: Hi, yes I made that change in my other branch, which should be visible in the new PR I recently created.

PR #2250, opened 2021-04-10
WIP: Add net.sonobus.SonoBus

kevinsmia1939:

- [X] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://github.com/sonosaurus/sonobus/issues/79


kevinsmia1939:
I could not get icon to be found.

kevinsmia1939:
bot, build net.sonobus.SonoBus

kevinsmia1939:
bot, build net.sonobus.SonoBus

net.sonobus.SonoBus.yaml
```diff
@@ -0,0 +1,50 @@
...
+    sources:
+      - type: archive
+        url: https://github.com/jackaudio/jack2/archive/refs/tags/v1.9.17.tar.gz
+        sha256: 38f674bbc57852a8eb3d9faa1f96a0912d26f7d5df14c11005ad499c8ae352f2
+
+  - name: SonoBus
+    buildsystem: simple
+    build-commands:
+      - CONFIG=Release make -j$(nproc) -C Builds/LinuxMakefile Standalone VST3
+      - install -d ${FLATPAK_DEST}/lib/vst3/
```
hfiguiere, line 37: these install commands should be in `post-install` IMHO

net.sonobus.SonoBus.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - '*'
+    sources:
+      - type: archive
+        url: https://github.com/jackaudio/jack2/archive/refs/tags/v1.9.17.tar.gz
+        sha256: 38f674bbc57852a8eb3d9faa1f96a0912d26f7d5df14c11005ad499c8ae352f2
+
+  - name: SonoBus
+    buildsystem: simple
+    build-commands:
+      - CONFIG=Release make -j$(nproc) -C Builds/LinuxMakefile Standalone VST3
```
hfiguiere, line 36: I am not sure how useful it is to build the VST3 here. Nobody will see it.

A VST3 specific package could be derivated from this.

net.sonobus.SonoBus.yaml
```diff
@@ -0,0 +1,50 @@
+app-id: net.sonobus.SonoBus
+runtime: org.freedesktop.Platform
+sdk: org.freedesktop.Sdk
+runtime-version: '20.08'
+command: sonobus
+#rename-desktop-file: sonobus.desktop
+rename-icon:
```
hfiguiere, line 7: rename-icon need an icon name. But since you install it manually, you should remove this here.

hfiguiere: This is why your build fails

hfiguiere (changes requested):

kevinsmia1939:
bot, build net.sonobus.SonoBus

net.sonobus.SonoBus.yaml
kevinsmia1939: Thanks, I miss that.

net.sonobus.SonoBus.yaml
kevinsmia1939: I build VST3 from the build script from SonoBus. Are you saying I should separate it to another build module?
https://github.com/sonosaurus/sonobus/blob/main/Builds/LinuxMakefile/install.sh

kevinsmia1939: I got it to run on my system now.
I got these warnings.

```
[W][000001617.490660][context.c:224 pw_context_new()] context 0x55cab1e30fa0: can't load config jack.conf: No such file or directory. Using client.conf fallback
[E][000001617.490720][context.c:232 pw_context_new()] context 0x55cab1e30fa0: can't load client.conf config: No such file or directory
```

kevinsmia1939:
bot, build net.sonobus.SonoBus

net.sonobus.SonoBus.metainfo.xml
```diff
@@ -0,0 +1,78 @@
...
+            
+            All settings are dynamic, network statistics are clearly visible. 
+            Works as a standalone application on macOS, Windows, iOS, and GNU/Linux, and as
+            an audio plugin (AU, VST, and AAX) on macOS and Windows. Use it on your desktop
+            or in your DAW, or on your mobile device.
+            
+            Easy to setup and use, yet still provides all the details that audio nerds want to see.
+            Audio quality can be instantly adjusted from full uncompressed PCM (16, 24, or 32 bit)
+            or with various compressed bitrates (16-256 kbps per channel) using 
+            the low-latency Opus codec.
```
Yajo, line 29:
```suggestion
            Features:
        </p>
        <p>
            Connects multiple users together to send and receive audio among all in a group,
            with fine-grained control over latency, quality and overall mix. Includes optional
            input compression, noise gate, and EQ effects, along with a master reverb. 
        </p>
        <p>
            All settings are dynamic, network statistics are clearly visible. 
            Works as a standalone application on macOS, Windows, iOS, and GNU/Linux, and as
            an audio plugin (AU, VST, and AAX) on macOS and Windows. Use it on your desktop
            or in your DAW, or on your mobile device.
        </p>
        <p>
            Easy to setup and use, yet still provides all the details that audio nerds want to see.
            Audio quality can be instantly adjusted from full uncompressed PCM (16, 24, or 32 bit)
            or with various compressed bitrates (16-256 kbps per channel) using 
            the low-latency Opus codec.
```

Yajo:
It would be nice to declare the VST plugin somewhere, so it could be usable with DAWs such as Ardour, which is already in flathub. However, I wouldn't consider that a requirement to begin.

net.sonobus.SonoBus.yaml
hfiguiere: what I mean about the VST3 is that building them in this flatpak is useless as no application will be able to see them.

it's somewhat a flatpak limitation, the solution will be a different flatpak, but first thing first.

The error below are actually a bug in freedestkop-sdk and nothing we can do for now. It broke Guitarix. See https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1223

hfiguiere:
> It would be nice to declare the VST plugin somewhere, so it could be usable with DAWs such as Ardour, which is already in flathub. However, I wouldn't consider that a requirement to begin.

As I tried to explain elsewhere, a new package would have to be made for that.

kevinsmia1939:
bot, build net.sonobus.SonoBus

Yajo:
I tested https://github.com/flathub/flathub/pull/2250#issuecomment-817265744 and JACK doesn't detect any I/O device. Also, qJackCtl doesn't display Sonobus as a possible JACK client.

Notice that I use pipewire-jack, and use it with other flatpaks.

hfiguiere:
> I tested [#2250 (comment)](https://github.com/flathub/flathub/pull/2250#issuecomment-817265744) and JACK doesn't detect any I/O device. Also, qJackCtl doesn't display Sonobus as a possible JACK client.
> 
> Notice that I use pipewire-jack, and use it with other flatpaks.

See https://github.com/flathub/flathub/pull/2250#discussion_r611198148

tl;dr, the freedesktop-sdk broke pipewire.

Yajo:
> tl;dr, the freedesktop-sdk broke pipewire.

Yikes... Can't we use a different SDK version then, while that gets fixed? 🤔

Yajo:
I found a bug, but I don't know if it's from the flatpak or sonobus itself.

The UI language appears by default as "System Default Language". In my case, that's Spanish. However, it displays in English if I don't select Spanish specifically:

imagen

net.sonobus.SonoBus.yaml
```diff
@@ -0,0 +1,46 @@
...
+sdk: org.freedesktop.Sdk
+runtime-version: '20.08'
+command: sonobus
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --device=dri
+  - --share=network
+  - --socket=pulseaudio
+  - --filesystem=home
```
Yajo, line 12: Shouldn't this be just the "Music" folder by default, and maybe "Downloads"?

net.sonobus.SonoBus.yaml
```diff
@@ -0,0 +1,46 @@
...
+    build-commands:
+      - ./waf configure --prefix=/app --htmldir=/app/share/doc/jack/ --classic
+      - ./waf build -j $FLATPAK_BUILDER_N_JOBS
+      - ./waf install
+    cleanup:
+      - '*'
+    sources:
+      - type: archive
+        url: https://github.com/jackaudio/jack2/archive/refs/tags/v1.9.17.tar.gz
+        sha256: 38f674bbc57852a8eb3d9faa1f96a0912d26f7d5df14c11005ad499c8ae352f2
```
Yajo, line 29: You can include [JACK from shared-modules](https://github.com/flathub/shared-modules/blob/b0d4c06751d0451c0a6cd9672b3132d6152f332c/linux-audio/jack2.json#L1-L21) if you want to reduce some code here.

Instructions: https://github.com/flathub/flathub/wiki/App-Requirements#shared-modules

kevinsmia1939: Thanks, will do that.

net.sonobus.SonoBus.yaml
kevinsmia1939: good point, will do that instead

kevinsmia1939:
bot, build net.sonobus.SonoBus

kevinsmia1939:
@Yajo 
For the system language, I guess it try to read some language config but can't because it cannot access it so it default to english.
Have to figure out what directory it need for this language config.

net.sonobus.SonoBus.yaml
hfiguiere: they shall be specified with `xdg-music` and `xdg-downloads` instead of directories.

See https://docs.flatpak.org/en/latest/sandbox-permissions-reference.html?highlight=xdg-music#filesystem-permissions

net.sonobus.SonoBus.yaml
```diff
@@ -0,0 +1,36 @@
...
+  - --share=network
+  - --socket=pulseaudio
+  - --filesystem=~/Downloads
+  - --filesystem=~/Music
+# Jack
+  - --filesystem=xdg-run/pipewire-0
+  - --system-talk-name=org.freedesktop.RealtimeKit1
+
+modules:
+  - shared-modules/linux-audio/jack2.json
```
hfiguiere, line 19: you must add a `cleanup` section to the whole package to remove `/lib/libjack*` and `/lib/jack`

Otherwise JACK won't work.

net.sonobus.SonoBus.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+  - name: SonoBus
+    buildsystem: simple
+    build-commands:
+      - CONFIG=Release make -j$(nproc) -C Builds/LinuxMakefile Standalone
+      - install -Dm755 -t ${FLATPAK_DEST}/bin/ Builds/LinuxMakefile/build/sonobus
```
hfiguiere, line 25: As I mentionned, it's better if this and the lined after be moved to `post-install` instead of `build-commands`

hfiguiere (changes requested):

net.sonobus.SonoBus.yaml
hfiguiere: The runtime has been updated and this problem should be resolved now.

kevinsmia1939:
bot, build net.sonobus.SonoBus

net.sonobus.SonoBus.yaml
```diff
@@ -0,0 +1,40 @@
...
+sdk: org.freedesktop.Sdk
+runtime-version: '20.08'
+command: sonobus
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --device=dri
+  - --share=network
+  - --socket=pulseaudio
+  - --filesystem=~/Downloads
```
barthalion, line 12:
```suggestion
  - --filesystem=xdg-download
```

net.sonobus.SonoBus.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - --filesystem=~/Music
```
barthalion, line 13:
```suggestion
  - --filesystem=xdg-music
```

net.sonobus.SonoBus.yaml
barthalion: Yeah, I don't think separating `post-install` makes sense here.

net.sonobus.SonoBus.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - /lib/libjack*
+  - /lib/jack
+
+modules:
+  - shared-modules/linux-audio/jack2.json
+
+  - name: SonoBus
+    buildsystem: simple
+    build-commands:
+      - CONFIG=Release make -j$(nproc) -C Builds/LinuxMakefile Standalone
```
barthalion, line 27:
```suggestion
      - CONFIG=Release make -j$FLATPAK_BUILDER_N_JOBS -C Builds/LinuxMakefile Standalone
```

kevinsmia1939:
bot, build net.sonobus.SonoBus

kevinsmia1939:
@barthalion Should I separate it to post-install?

net.sonobus.SonoBus.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - install -D "${FLATPAK_BUILDER_BUILDDIR}/images/SonoBus-Icon.svg" "${FLATPAK_DEST}/share/icons/hicolor/scalable/apps/$FLATPAK_ID.svg"
+      - install -Dm644 "${FLATPAK_BUILDER_BUILDDIR}/$FLATPAK_ID.metainfo.xml" "${FLATPAK_DEST}/share/metainfo/$FLATPAK_ID.metainfo.xml"
+    post-install:
+      - install -Dm755 -t ${FLATPAK_DEST}/bin/ Builds/LinuxMakefile/build/sonobus
+      - install -Dm644 "Builds/LinuxMakefile/sonobus.desktop" "${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop"
+      - desktop-file-edit --set-icon=${FLATPAK_ID} /app/share/applications/${FLATPAK_ID}.desktop
+    sources:
+      - type: git
+        url: https://github.com/sonosaurus/sonobus/
+        tag: '1.4.3'
```
hfiguiere, line 37: FYI 1.4.5 is out.

barthalion:
bot, build net.sonobus.SonoBus

barthalion:
Looks like upstream switched to cmake between 1.4.3 and 1.4.5. @kevinsmia1939 mind having a look? Other than that, LGTM.

barthalion:
Hi @essej and @matthewevans, would you like to be invited to the Flathub repo once it's created? The general idea behind Flathub is that packages are at least co-maintained by someone from the upstream team, but there's of course no obligation, I'd like you to have write access "just in case".

essej:
> Hi @essej and @matthewevans, would you like to be invited to the Flathub repo once it's created? The general idea behind Flathub is that packages are at least co-maintained by someone from the upstream team, but there's of course no obligation, I'd like you to have write access "just in case".

Yes, that would be great, thanks!

matthewevans:
Yes, you can add me. Thank you!

kevinsmia1939:
bot, build net.sonobus.SonoBus

essej:
Just a heads up, the cmake install targets do NOT do anything in the current sonobus build setup, so you will need to still have installation steps after the cmake build.

net.sonobus.SonoBus.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - --system-talk-name=org.freedesktop.RealtimeKit1
+cleanup:
+  - /lib/libjack*
+  - /lib/jack
+
+modules:
+  - shared-modules/linux-audio/jack2.json
+
+  - name: SonoBus
+    buildsystem: cmake
```
hfiguiere, line 25: Please use `cmake-ninja` here unless there is a good reason not to. It's always faster.

net.sonobus.SonoBus.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - --share=network
+  - --socket=pulseaudio
+  - --filesystem=xdg-download
+  - --filesystem=xdg-music
+# Jack
+  - --filesystem=xdg-run/pipewire-0
+  - --system-talk-name=org.freedesktop.RealtimeKit1
+cleanup:
+  - /lib/libjack*
+  - /lib/jack
```
hfiguiere, line 19: If you update shared-modules to the latest version, you can drop this cleanup. It's a recent change (more recent than when I said it should be here)

essej:
That last change is not going to work either... there is no Builds/LinuxMakefile anymore. The built executable when using cmake will be in the cmake build directory under SonoBus_artefacts/Release/Standalone/   and the desktop file is at the top-level of the source under the linux/ subdirectory.

kevinsmia1939:
> That last change is not going to work either... there is no Builds/LinuxMakefile anymore. The built executable when using cmake will be in the cmake build directory under SonoBus_artefacts/Release/Standalone/ and the desktop file is at the top-level of the source under the linux/ subdirectory.

I don't really understand how to build it.
Looking at the build script, it should just be this?
```
      - cmake -DCMAKE_BUILD_TYPE=Release -B build
      - cmake --build build
```

essej:
The build step is fine... it was the installation steps that follow it that are referring to the wrong file locations

barthalion:
bot, build net.sonobus.SonoBus

PR #2302, opened 2021-05-01
io.github.trevorsandy.LPub3D

trevorsandy:


hfiguiere:
Awesome ! I had that one on my list for a while.

(I have one not submitted yet for LDview though)

hfiguiere:
bot, build io.github.lpub3d.LPub3D

io.github.lpub3d.LPub3D.json
```diff
@@ -0,0 +1,178 @@
...
+        {
+          "type": "archive",
+          "dest": "/app/lpub3d_linux_3rdparty",
+          "url": "https://github.com/trevorsandy/lpub3d_libs/releases/download/v1.0.1/tenteparts.zip",
+          "sha256": "7e1a68490b1be11bddcfb3180151b42ffbb85ce3aa8ec3f7431bee5c04cc1ac5"
+        },
+        {
+          "type": "archive",
+          "dest": "/app/lpub3d_linux_3rdparty",
+          "url": "https://github.com/trevorsandy/lpub3d_libs/releases/download/v1.0.1/vexiqparts.zip.sha256",
```
hfiguiere, line 63: Did you mean the `.zip` instead?

This is why the build fail.

hfiguiere (changes requested):

io.github.lpub3d.LPub3D.json
trevorsandy: Yes, I saw that. And yes, I intended the .zip.

Cheers,

hfiguiere:
bot, build io.github.lpub3d.LPub3D

hfiguiere:
bot, build io.github.lpub3d.LPub3D

io.github.lpub3d.LPub3D.json
```diff
@@ -0,0 +1,176 @@
...
+      "sources": [
+        {
+          "type": "git",
+          "url": "https://github.com/trevorsandy/povray.git",
+          "branch": "lpub3d/raytracer-cui",
+          "commit": "54b65529090b4161193df24f0d846046b1495f0b"
+        },
+        {
+          "type": "patch",
+          "path": "patches/povray/0002-Autotool-massaging.patch"
```
hfiguiere, line 124: seems that the path doesn't match the actual location :-/

hfiguiere:
bot, build io.github.lpub3d.LPub3D

io.github.lpub3d.LPub3D.json
```diff
@@ -0,0 +1,172 @@
...
+        "CONFIG+=build_check",
+        "CONFIG-=debug_and_release",
+        "INSTALL_PREFIX=/app"
+      ],
+      "sources": [
+        {
+          "type": "git",
+          "url": "https://github.com/trevorsandy/lpub3d-ci.git",
+          "tag": "v2.4.3",
+          "commit": "76d9418a484497f822a7044670101b7749ce7bb8"
```
hfiguiere, line 163: this is not the commit I see for tag `v2.4.3`

trevorsandy: Indeed. Not sure what happene here ?

hfiguiere:
bot, build io.github.lpub3d.LPub3D

io.github.lpub3d.LPub3D.json
```diff
@@ -0,0 +1,172 @@
...
+      "buildsystem": "simple",
+      "build-commands": [
+        "unzip -d ${LP3D_3RD_PARTY_DIR} -q /app/lpub3d_linux_3rdparty/complete.zip",
+        "if test -d ${LP3D_3RD_PARTY_DIR}/ldraw; then echo 'LDraw library extracted.'; fi"
+      ],
+      "sources": [
+        {
+          "type": "archive",
+          "dest-filename": "complete.zip",
+          "dest": "/app/lpub3d_linux_3rdparty",
```
hfiguiere, line 43: I was surprised this work. `dest` shouldn't be the install directory. You should do that in the `build-commands` or `post-install`.

This apply on the others too.
You probably want to make this `"type": "file"` and unzip them in the `build-comands`

hfiguiere (changes requested):

io.github.lpub3d.LPub3D.json
trevorsandy: Is the problem the shared location with the renderers install command ? Can the archives be placed at a location in, but not at the root of `lpub3d_linux_3rdparty` ? The approach I used is also how I build on Travis, AppVeyor and my build from source scripts. The 3rdParty dir is used to 'stage' the renderers and hold the libraries. At LPub3D install, the renderers, libraries and their related content are sourced from there so `post-install` will not be possible without rewriting the LPub3D install routines.

The documentation states all type can use `dest`; moreover, here is an example https://github.com/flathub/io.github.maurycyliebner.enve/blob/master/io.github.maurycyliebner.enve.yaml 

Cheers,

trevorsandy: I've set the script to install each renderer in it's own folder and I move the installed content to the proper 'staging' location at `post-install`.

Cheers,

hfiguiere: /app is not for downloading "source" file.
`dest` is the destination directory.

Note that `type: archive` mean the archive will be extracted in that directory. But this is not the installation of assets.
For that the best is just `type: file` and then in the `post-install` or `build-commands` to unzip in the destination location.

In the example you link `dest` as actually used to checkout a git tree in a subdir of the main source tree.

trevorsandy:
> Note that type: archive mean the archive will be extracted in that directory...

Very well. I've applied the changes as you suggested.

Cheers,

io.github.lpub3d.LPub3D.json
```diff
@@ -0,0 +1,183 @@
...
+  "sdk": "org.kde.Sdk",
+  "command": "lpub3d24",
+  "rename-icon": "lpub3d",
+  "rename-desktop-file": "lpub3d.desktop",
+  "rename-appdata-file": "io.github.lpub3d.appdata.xml",
+  "build-options": {
+    "env": {
+      "LP3D_3RD_PARTY_DIR": "/stage/lpub3d_linux_3rdparty",
+      "LP3D_DIST_DIR_PATH": "/stage/lpub3d_linux_3rdparty",
+      "LDRAWDIR": "/stage/lpub3d_linux_3rdparty/ldraw",
```
hfiguiere, line 14: None of these will work. `/` isn't writable.

io.github.lpub3d.LPub3D.json
```diff
@@ -0,0 +1,183 @@
...
+    "--filesystem=home",
+    "--env=TMPDIR=/var/tmp",
+    "--env=QT_ENABLE_HIGHDPI_SCALING=1"
+  ],
+  "modules": [
+    {
+      "name": "part-archives",
+      "buildsystem": "simple",
+      "build-commands": [
+        "unzip -d ${LP3D_3RD_PARTY_DIR} -q /stage/lpub3d_linux_3rdparty/complete.zip",
```
hfiguiere, line 36: As I indicated earlier, `/` is read only. 

I'm not sure I understand what this staging is for.

io.github.lpub3d.LPub3D.json
```diff
@@ -0,0 +1,183 @@
...
+        },
+        {
+          "type": "file",
+          "url": "https://github.com/trevorsandy/lpub3d_libs/releases/download/v1.0.1/vexiqparts.zip",
+          "sha256": "e2f4114c8ff6bfe7e5a43b6c88fb47f32328ddb147578456ec0825aa76803b20"
+        }
+      ],
+      "post-install": [
+        "mv -f complete-20-03.zip ${LP3D_3RD_PARTY_DIR}/complete.zip",      
+        "mv -f lpub3dldrawunf-01-04-21.zip ${LP3D_3RD_PARTY_DIR}/lpub3dldrawunf.zip",
```
hfiguiere, line 65: even if `$LP3D_3RD_PARTY_DIR` is valid, you rename the files with `dest-filename` (good) so you should use the renamed filename here.

hfiguiere (changes requested):

hfiguiere:
bot, build io.github.lpub3d.LPub3D

hfiguiere:
I predict it will fail.

io.github.lpub3d.LPub3D.json
```diff
@@ -0,0 +1,183 @@
...
+        "./configure COMPILED_BY=\"Trevor SANDY <trevor.sandy@gmail.com> using FlatHub for LPub3D.\" --prefix=/stage/lpub3d_linux_3rdparty/povray_renderer LPUB3D_3RD_PARTY=yes --with-libsdl2 --with-boost-libdir=${FLATPAK_DEST}/lib --enable-watch-cursor",
+        "make",
+        "make install"
+      ],
+      "sources": [
+        {
+          "type": "git",
+          "url": "https://github.com/trevorsandy/povray.git",
+          "branch": "lpub3d/raytracer-cui",
+          "commit": "54b65529090b4161193df24f0d846046b1495f0b"
```
hfiguiere, line 128: usually branch + commit is a bad idea because branch are moving targets.

Usually what I do to indicate the branch I put it in a comment `/* */` (that works in the manifest) purely for information.

io.github.lpub3d.LPub3D.json
```diff
@@ -0,0 +1,183 @@
...
+      "post-install": [
+        "mv -f ${LP3D_3RD_PARTY_DIR}/ldview_renderer/* ${LP3D_3RD_PARTY_DIR}/"
+      ]
+    },
+    {
+      "name": "povray",
+      "buildsystem": "simple",
+      "build-commands": [
+        "cd unix && ./prebuild3rdparty.sh",
+        "./configure COMPILED_BY=\"Trevor SANDY <trevor.sandy@gmail.com> using FlatHub for LPub3D.\" --prefix=/stage/lpub3d_linux_3rdparty/povray_renderer LPUB3D_3RD_PARTY=yes --with-libsdl2 --with-boost-libdir=${FLATPAK_DEST}/lib --enable-watch-cursor",
```
hfiguiere, line 119: I think you should directly build in in the prefix (`${FLATPAK_DEST}`).

hfiguiere (commented):
I think I understand what you are trying to do. You probably want to build and install everything as they go into the prefix, which is  `$FLATPAK_DEST` (currently set to `/app`).

hfiguiere:
To clarify more: `/app` will start empty, and everything you put in there will be used:

1. to build the next module
2. be part of the final installation.
3. `cleanup` will be used to remove what not needed in the final package like header, static libraries, etc.

io.github.lpub3d.LPub3D.json
trevorsandy: > even if $LP3D_3RD_PARTY_DIR is valid, you rename the files with dest-filename (good)

Oops, forgot about that one.

> I'm not sure I understand what this staging is for

The staging folder is internal to how LPub3D captures the renderer and parts library artefacts. What I need from the flatpack environment is a writable location - that's it, but I'm struggling to determine where I can put my staging folder and be able to pass a full path to the renderers and LPub3D?

Do you see any restriction with using `"LP3D_3RD_PARTY_DIR": "$(realpath ./lpub3d_linux_3rdparty)"` ?

Cheers,

trevorsandy:
> To clarify more: /app will start empty, and everything you put in there will be used:

This is what I was doing in my initial submission:

 - Create my staging folder
 - Dowload my libraries, extract complete.zip so I can run my checks
 - Build and install (into the staging folder) my renderers.
 - Build and install LPub3D, grab my libraries and renderers and install, as I always do, to the specified flatpack location /app I believe
 - After LPub3D install, remove the staging folder.
 - Let flatpack do its thing and viola!

> I think you should directly build in in the prefix (${FLATPAK_DEST}).

You made this comment in reference to POV-Ray. The renderers (LDView, LDGLite and POV-Ray) are not independently installed. They are bundled into the LPub3D package with the executables installed under opts/lpub3d/3rdParty/... My renderers are all customized to follow my build and install approach, they are not off-the-shelf distributions.

Anyway, wat I need from the flatpack environment is a writable location for my staging folder - that's it, but I'm struggling to determine where I can put it be able to pass a full path to the renderers and LPub3D ?

Do you see any restriction with using `"LP3D_3RD_PARTY_DIR": "$(realpath ./lpub3d_linux_3rdparty)"` ?

Cheers,

io.github.lpub3d.LPub3D.json
trevorsandy: > usually branch + commit is a bad idea because branch are moving targets.

I don't understand you point, are you saying a commit is less of a moving target ? 

I'll use commit if there's going to be problems referencing both.

Cheers,

io.github.lpub3d.LPub3D.json
```diff
@@ -0,0 +1,183 @@
+{
+  "id": "io.github.lpub3d.LPub3D",
```
barthalion, line 2: Shouldn't it be `io.github.trevorsandy.LPub3D`?

trevorsandy: I followed the flatpack guidance and recommendation (to my understanding) and chose to go with the application name versus the GitHub username. 

If the current configuration is contributing to the packaging failure (or will contribute to runtime failure), I will be pleased to change it.

Cheers,

io.github.lpub3d.LPub3D.json
trevorsandy: > usually branch + commit is a bad idea because branch are moving targets.

Additionally, I intuitively thought it would be necessary to indicate the branch, if not master, to properly resolve the commit. 

Cheers,

trevorsandy:
Let me know if it would be better to change my submission to `io.github.trevorsandy.LPub3D` ?

Cheers,

io.github.lpub3d.LPub3D.json
trevorsandy: > usually branch + commit is a bad idea because branch are moving targets.

Additionally, I intuitively thought it would be necessary to indicate the branch, if not master, to properly resolve the commit. 

Cheers,

hfiguiere: No. a commit is the only thing you need. Branches are just named references to a commit that gets updated when you add a new commit on top.

io.github.lpub3d.LPub3D.json
hfiguiere: You don't need to redo the submission. Update the PR and changing the title works fine.

the guidance it to use a domain you own (or has been delegated). This would work if you owned `https://github.com/lpub3d` as an organization or username.

io.github.lpub3d.LPub3D.json
hfiguiere: it won't work. I'm not even sure shell substitution is even done there anyway.

`${FLATPAK_DEST}/lpub3d_linux_3rdparty` should work, but you'll need to create it early and then add `/lpub3d_linux_3rdparty` in the `cleanup` section.

io.github.lpub3d.LPub3D.json
```diff
@@ -0,0 +1,183 @@
...
+    },
+    {
+      "name": "ldglite",
+      "buildsystem": "qmake",
+      "config-opts": [
+        "CONFIG+=3RD_PARTY_INSTALL=/stage/lpub3d_linux_3rdparty/ldglite_renderer",
+        "CONFIG+=release",
+        "CONFIG+=BUILD_CHECK"
+      ],
+      "sources": [
```
hfiguiere, line 78: ldglite require `freeglut` which require `glu`.

hfiguiere (changes requested):

io.github.lpub3d.LPub3D.json
```diff
@@ -0,0 +1,183 @@
...
+      "post-install": [
+        "mv -f ${LP3D_3RD_PARTY_DIR}/ldview_renderer/* ${LP3D_3RD_PARTY_DIR}/"
+      ]
+    },
+    {
+      "name": "povray",
+      "buildsystem": "simple",
+      "build-commands": [
+        "cd unix && ./prebuild3rdparty.sh",
+        "./configure COMPILED_BY=\"Trevor SANDY <trevor.sandy@gmail.com> using FlatHub for LPub3D.\" --prefix=${LP3D_3RD_PARTY_DIR}/povray_renderer LPUB3D_3RD_PARTY=yes --with-libsdl2 --with-boost-libdir=${FLATPAK_DEST}/lib --enable-watch-cursor",
```
hfiguiere, line 119: doesn't this configure line indicate you need SDL2? If yes then it'll need to be built as well.

trevorsandy: Dependencies. Yep. This was the area I wanted to get to so I could determine the next step. 

There was no intuitive way to determine what libraries were available (or unavailable) in the runtime and; moreover, no intuitive way to simply list the off-the-shelf dependencies. If, i well understand, I’ll have to go find each needed dependency download, find or generate its sha256 and manually add it to the build ?

Cheers,

trevorsandy: The entirety of my [AppImage yml](https://github.com/trevorsandy/lpub3d-ci/blob/master/builds/linux/obs/alldeps/appimage.yml) is 60 lines - listing the required dependencies is very intuitive. 

Does flatpack have anything similar?

Cheers,

trevorsandy:
Here is a list of dependencies, for my RedHat flavoured build:

How can I determine which of these must be setup to be built, can be loaded (using pkgconfig for example) or is already pre-loaded with the runtime.

  ```
    - linuxdeployqt
    - gcc-c++
    - unzip
    - autoconf
    - automake
    - libpng-devel
    - libtiff-devel
    - libjpeg8-devel
    - tinyxml-devel
    - gl2ps-devel
    - pkgconfig(zlib)
    - pkgconfig(Qt5Gui)
    - pkgconfig(Qt5Widgets)
    - pkgconfig(Qt5OpenGL)
    - pkgconfig(Qt5PrintSupport)
    - pkgconfig(Qt5Concurrent)
    - pkgconfig(Qt5Xml)
    - libQt5Core-devel
    - libQt5Network-devel
    - libqt5-qttools-devel
    - libqt5-linguist
    - pkgconfig(OpenEXR)
    - pkgconfig(sdl2)
    - pkgconfig(fcitx)
    - xorg-x11-libX11-devel
    - xorg-x11-libXpm-devel
    - libSM-devel
    - freeglut-devel
    - glu-devel
    - libOSMesa-devel
    - Mesa-devel
    - boost-devel
```

Cheers,

**Comment by hfiguiere** [comment]
And then there is the problem of 'osmesa' that is not in the runtime.

**Comment by hfiguiere** [comment]
bot, build io.github.lpub3d.LPub3D

File: io.github.lpub3d.LPub3D.json [hunk 13]
```diff
@@ -0,0 +1,183 @@
...
+  "runtime": "org.kde.Platform",
+  "runtime-version": "5.15",
+  "sdk": "org.kde.Sdk",
+  "command": "lpub3d24",
+  "rename-icon": "lpub3d",
+  "rename-desktop-file": "lpub3d.desktop",
+  "rename-appdata-file": "io.github.lpub3d.appdata.xml",
+  "build-options": {
+    "env": {
+      "LP3D_3RD_PARTY_DIR": "$(realpath ./lpub3d_linux_3rdparty)",
```
hfiguiere, line 12: This definitely doesn't work:
1. shell substitution isn't performed here. Just variable.
2. I expect to resolve a path relative (inside) the build dir, but after building the module the build dir is deleted, which mean that using this for staging might not be a good idea.

io.github.lpub3d.LPub3D.json
```diff
@@ -0,0 +1,183 @@
...
+      "LP3D_DIST_DIR_PATH": "$(realpath ./lpub3d_linux_3rdparty)",
+      "LDRAWDIR": "$(realpath ./lpub3d_linux_3rdparty)/ldraw",
+      "PLATFORM_CODE": "osl",
+      "TARGET_CPU": "$(uname -m)",
+      "PLATFORM_VER": "$(. /etc/os-release 2>/dev/null; [ -n \"$VERSION_ID\" ] && echo $VERSION_ID || true)"
```
hfiguiere, line 17: This probably won't work as intended either.
```
$ echo $PLATFORM_VER
$(. /etc/os-release 2>/dev/null; [ -n "$VERSION_ID" ] && echo $VERSION_ID || true)
```

io.github.lpub3d.LPub3D.json
hfiguiere: no.

hfiguiere (changes requested):

io.github.lpub3d.LPub3D.json
trevorsandy: It’s hard to understand the logic of this platform. I only need the build dir through install, is it deleted before install is completed ? All my routines, including check, occur before install completion. I don’t understand why using this location, during the time that I will is not prudent.

Cheers,

io.github.lpub3d.LPub3D.json
trevorsandy: I will have to find another way to populate this var. I’ll pass it directly to `qmake` in `config-ops`

Cheers,

io.github.lpub3d.LPub3D.json
trevorsandy: So is “bump your nose” the only approach to determine missing dependencies ? I guess I could scour the other repos looking for packages that use these libs.

Cheers,

hfiguiere: One I thing I do before submitting a flatpak is building locally using `flatpak-builder`. I admit figuring out what needs to be build is sometime a guess job. Just trying to help mentioning what is needed here.

Currently (with my local changes) I'm stopped at the requirement to have `osmesa`. Mesa is in the runtime, but it's build without OSMesa. And the only other flatpak I could find, is Octave which removed the need for OSMesa a while back.

trevorsandy: I did look briefly at setting up a local flatpack-builder. But I learned I needed to set up flatpack for that so I stopped there - it difficult enough setting up a distro much less the whole platform. Coming off my experience of setting up AppImage and Open Build Service, I guess my expectations were off with regards to the difference in approach between flatpack and those platforms. 

I also managed to collect several example yml files to round out the dependency requirements. I haven’t found one for OSMesa yet either. I’ll keep looking though.

Cheers,

hfiguiere: There is not much setup to insall flatpak-builder and have it run IMHO. The biggest thing is probably intaling the SDK which can take some time depending on the speed of your internet connection. And the few GB of disk space.

For OSmesa I have this, but it probably need refining:
```json
 {
          "name": "mesa",
          "buildsystem": "meson",
          "config-opts": [
            "-Dosmesa=classic",
            "-Ddri-drivers=[]",
            "-Dplatforms=['x11']",
            "-Dgallium-drivers=[]",
            "-Dvulkan-drivers=[]"
          ],
          "sources": [
            {
              "type": "archive",
              "url": "https://archive.mesa3d.org//mesa-20.2.6.tar.xz",
              "sha256": "f12ca3c6c622f11cd79ad66b4220f04514fa96f795062fe92a37339ab19885db"
            }
          ]
        }
```
And you need to set the following for ldglite in the `config-opts`
```json
        "QMAKE_LIBDIR=/app/lib",
```

hfiguiere: By refining, I mean cleaning up unused stuff.

io.github.lpub3d.LPub3D.json
```diff
@@ -0,0 +1,183 @@
...
+        "mv -f lpub3dldrawunf.zip ${LP3D_3RD_PARTY_DIR}/lpub3dldrawunf.zip",
+        "mv -f tenteparts.zip ${LP3D_3RD_PARTY_DIR}/",
+        "mv -f vexiqparts.zip ${LP3D_3RD_PARTY_DIR}/"
+      ]
+    },
+    {
+      "name": "ldglite",
+      "buildsystem": "qmake",
+      "config-opts": [
+        "CONFIG+=3RD_PARTY_INSTALL=${LP3D_3RD_PARTY_DIR}/ldglite_renderer",
```
hfiguiere, line 74: Also this is broken when you pass an absolute path because the underlying script don't check that and just concatenate.

io.github.lpub3d.LPub3D.json
hfiguiere: I have published by WiP based on your PR https://github.com/hfiguiere/flathub/tree/io.github.lpub3d.LPub3D 

It still fail to build.

trevorsandy: Excellent. The OSMesa bit is a good starting point, I think only the build flags might need to be updated. I’ll take a look at what I use between LDView and LDGLite.

> There is not much setup to insall flatpak-builder and have it run IMHO. 

Do you know if it will work on Microsoft’s WLS (Ubuntu 16.04) ? Will configure without arguments be ok because none are suggested ?

Cheers,

io.github.lpub3d.LPub3D.json
trevorsandy: I must have changed it to use a relative path. I’ll take a look. 

Cheers,

trevorsandy:
I'm trying to seupt flatpak-builder on WSL2 (Ubuntu 16.04) and, as expected, running into some tangential issues. I'm stuck on Bubblewrap trying to create a namespace. The return is:  

```
$ sudo flatpak-builder build-dir org.flatpak.Hello.yml
Downloading sources
Starting build of org.flatpak.Hello
Cache miss, checking out last cache hit
fuse: mountpoint is not empty
fuse: if you are sure this is safe, use the 'nonempty' mount option
========================================================================
Building module hello in /home/trevorsandy/flatpak_demo/.flatpak-builder/build/hello-4
========================================================================
Running: install -D hello.sh /app/bin/hello.sh
bwrap: Creating new namespace failed: Permission denied
Error: module hello: Child process exited with code 1
```

I'm not sure about this but I've seen some posts where `chown` and `chmod` setting resolved the issue. The problem is I can't seem to locate the `bwrap` instance. It is not found with `which bwrap`. Any ideas ?

Cheers,

io.github.lpub3d.LPub3D.json
hfiguiere: Five year old Ubuntu, that's pushing the difficulty level to the max...

io.github.trevorsandy.LPub3D.json
```diff
@@ -0,0 +1,274 @@
...
+      "sources": [
+        {
+          "type": "archive",
+          "url": "https://downloads.sourceforge.net/project/freeglut/freeglut/3.2.1/freeglut-3.2.1.tar.gz",
+          "sha256": "d4000e02102acaf259998c870e25214739d1f16f67f99cb35e4f46841399da68"
+        }
+      ]
+    },
+    {
+      "name": "mesa",
```
barthalion, line 48: Why does it need its own mesa?

barthalion:
re WSL issues – I don't think it's going to be easily doable. After all WSL is not a full Linux environment. A good first step would be moving to 18.04 to 20.04 and installing flatpak as described at https://flatpak.org.

barthalion:
bot, build io.github.lpub3d.LPub3D

barthalion:
bot, build io.github.trevorsandy.LPub3D

barthalion:
bot, build io.github.trevorsandy.LPub3D

io.github.trevorsandy.LPub3D.json
trevorsandy: @hfiguiere stated:

> And then there is the problem of 'osmesa' that is not in the runtime.

So I understood OSMesa is not available in the runtime Mesa installation.

Cheers,

trevorsandy:
> re WSL issues – I don't think it's going to be easily doable. After all WSL is not a full Linux environment. A good first step would be moving to 18.04 to 20.04 and installing flatpak as described at https://flatpak.org.

I added 20.04 to my WSL but couldn’t get `bwrap` to work so I installed the DE on a Fedora 34 Docker container instead. 

All went well for the KDE 5.15 SDK except the consistent failure to install `org.freedesktop.platform.openh264`. 

I was able to successfully build `io.github.trevorsandy.LPub3D` but I think there are some tweaks remaining - e.g cleanup tiff libs from /bin.

I normally install the renderer executables to /opt/lpub3d/3rdparty but I’m not sure if this is ok for the Flatpak build. If not, I’ll set LPub3D to install the them to /bin/lpub3d/

Cheers,

io.github.trevorsandy.LPub3D.json
hfiguiere: I saw. I can have a look like that later.

hfiguiere:
It's missing an appstream file.

io.github.trevorsandy.LPub3D.json
```diff
@@ -0,0 +1,274 @@
...
+              "sha256": "8f9a5af6131583404261931d9a5c83de0a425cb4b8b25ddab2b169fbf113aecd"
+            },
+            {
+              "type": "shell",
+              "commands": [ "cp -p /usr/share/automake-*/config.{sub,guess} ." ]
+            }
+          ]
+        },
+        {
+          "name": "libtiff",
```
hfiguiere, line 232: libtiff 4.1 is in the runtime.

I removed it in my test build and it works.

Is this really necessary?

io.github.trevorsandy.LPub3D.json
```diff
@@ -0,0 +1,274 @@
...
+    {
+      "name": "ldraw-libraries",
+      "buildsystem": "simple",
+      "build-commands": [
+        "install -d ${FLATPAK_DEST}/lpub3d_linux_3rdparty/",
+        "mv -f complete.zip ${FLATPAK_DEST}/lpub3d_linux_3rdparty/",        
+        "mv -f lpub3dldrawunf.zip ${FLATPAK_DEST}/lpub3d_linux_3rdparty/",
+        "mv -f tenteparts.zip ${FLATPAK_DEST}/lpub3d_linux_3rdparty/",
+        "mv -f vexiqparts.zip ${FLATPAK_DEST}/lpub3d_linux_3rdparty/",
+        "unzip -d ${FLATPAK_DEST}/lpub3d_linux_3rdparty -q ${FLATPAK_DEST}/lpub3d_linux_3rdparty/complete.zip"
```
hfiguiere, line 79: I do have question about this section:
1. the libraries are copied to the user config directory. Which mean that now `~/.var/app/io.github.trevorsandy.LPub3D/data` is 506MB. It contains `complete.zip` and `lpub3dldrawunf.zip`, and the unzip `ldraw`
2. we unzip ldraw here and it end up in the flatpak. And it seems to not be in the search path for the library.

So isn't there a way that would avoid not having this duplicated.

hfiguiere (changes requested):

io.github.trevorsandy.LPub3D.json
```diff
@@ -0,0 +1,274 @@
...
+        }
+      ]
+    },
+    {
+      "name": "mesa",
+      "buildsystem": "meson",
+      "config-opts": [
+        "-Dosmesa=classic",
+        "-Ddri-drivers=[]",
+        "-Dplatforms=['x11']",
```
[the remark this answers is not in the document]
trevorsandy: > I dug deeper and we can get rid of this to build only OSMesa

Excellent.

Cheers,

io.github.trevorsandy.LPub3D.json
trevorsandy: > I do have question about this section:

The LDraw folder should be removed at cleanup.

There is...

```
"cleanup": [
     "/app/lpub3d_linux_3rdparty"
   ],
```

... is in the manifest but I see that the staging folder (containing LDraw, and a lot of other unneeded content. The bulk of which make up the 500 MB data size) is not removed.

Cheers,

io.github.trevorsandy.LPub3D.json
trevorsandy: > Is this really necessary?

No. I did not see any indication that libtiff was in the runtime, so I added it. Like you said, “figuring out what needs to be build is sometime a guess job”.

Anyway, I will be happy to remove it as it requires a good bit of cleanup.

Cheers,

io.github.trevorsandy.LPub3D.json
hfiguiere: Cleanup should be 
```
"cleanup": [
     "/lpub3d_linux_3rdparty"
   ],
```
It's all based on the prefix (which is `/app`)

trevorsandy: > Cleanup should be

Very well. I’ll update the manifest when I add the appdata file. 

> It's missing an appstream file.

I noticed some manifests do not include any appdata file placement reference while others use either `/app/share/appdata` or `/app/share/metainfo`. I saw the documentation say`/app/share/metainfo` should be used however. What is the difference and moreover, how come some manifesta do not have a placement reference and the build is not affected ?

Cheers,

hfiguiere: `/app/share/appdata` is deprecated. It still work, but whenever the new one should be used.
And some package install it by default so the manifest doesn't need to handle it.

hfiguiere: `${FLATPAK_DEST}/opt` is fine, but you could as well install everything into `${FLATPAK_DEST}` since it all yours anyway.

io.github.trevorsandy.LPub3D.json
```diff
@@ -0,0 +1,274 @@
...
+    {
+      "name": "mesa",
+      "buildsystem": "meson",
+      "config-opts": [
+        "-Dosmesa=classic",
+        "-Ddri-drivers=[]",
+        "-Dplatforms=[]",
+        "-Dgallium-drivers=[]",
+        "-Dvulkan-drivers=[]"
+      ],
```
hfiguiere, line 56: Here, to build for `aarch64`, you need to insert this:
```json
      "build-options": {
        "arch": {
          "aarch64": {
            "config-opts": [
              "-Dlibunwind=disabled"
            ]
          }
        }
      },
```

this disable libunwind only on aarch64.

I could finish the build (albeit not yet tested on Raspberry Pi)

hfiguiere:
bot, build io.github.trevorsandy.LPub3D

trevorsandy:
Good to see the checks have passed. 

Question: Is XVFB supported in the runtime for build activity ?

I'm running this shell command to check for xvfb and I'm getting the ERROR message:

```
if [ -z "$(command -v xvfb-run)" ]; then
    show_settings_and_exit "ERROR - XVFB (xvfb-run) not found. Build check cannot be executed."
else
    USE_XVFB="true"
fi
```

Cheers,

hfiguiere:
There is no Xvfb.

trevorsandy:
> There is no Xvfb.

Ok. Is there anything else in place to allow desktop applications to run headless ? 

Cheers,

nedrichards:
I get the following error on the pinebook pro:
> QEGLPlatformContext: eglMakeCurrent failed: 3009
Visual Editor error creating OpenGL context.
ERROR 2021-05-12T21:21:32.301 Unable to initialize Visual Editor. 
Unable to initialize Visual Editor.

I should have OpenGL 3.3 and GLES 3 enabled but, y'know.

hfiguiere:
I just tested on the Raspberry Pi400 and it fails too the same way.

trevorsandy:
> I just tested on the Raspberry Pi400 and it fails too the same way.

```
#if !defined(EGL_VERSION_1_0) && !defined(GL_ES_VERSION_2_0) && !defined(GL_ES_VERSION_3_0) && !defined(QT_OPENGL_ES)
#else
#define LC_OPENGLES 1
#endif

```

Here are the defines that trigger (or not) GLES.

Cheers,

trevorsandy:
I can check in a 'test build' with the Q_PROCESSOR_ARM directive added. Let me know if this would help and I'll post the SHA.

#if !defined(EGL_VERSION_1_0) && !defined(GL_ES_VERSION_2_0) && !defined(GL_ES_VERSION_3_0) && !defined(QT_OPENGL_ES && !defined(Q_PROCESSOR_ARM))
#else
#define LC_OPENGLES 1
#endif

Cheers,

nedrichards:
Yeah, I think that would - and we can make sure aarch64 uses that release whilst keeping x86 on big GL.

nedrichards:
Can confirm that everything runs fine on x86 btw. so aside from the arm stuff it all seems good to go.

trevorsandy:
> Let me know if this would help and I'll post the SHA.

Here you go: `2ab166c784c451a4fe86c6f5ad2a61179e972557` Flatpak test for ARM build eglMakeCurrent failed: 3009 error.

> Yeah, I think that would - and we can make sure aarch64 uses that release whilst keeping x86 on big GL.

`!defined(Q_PROCESSOR_ARM)`  is conditional on the runtime arch being ARM so being in the same release as x86 will not have any adverse impact. 

Also note, I'm validating these builds (as well as my Snap ones) on my DevOps repo - [lpub3d-ci](https://github.com/trevorsandy/lpub3d-ci) vs. Production - [lpub3d](https://github.com/trevorsandy/lpub3d). Once all is ok, I'll coutover the base to prod and release it. I'll then update the FP manifest and PR accordingly.

Cheers,

hfiguiere:
bot, build io.github.trevorsandy.LPub3D

nedrichards:
bot, build io.github.trevorsandy.LPub3D

hfiguiere:
bot, build io.github.trevorsandy.LPub3D

hfiguiere:
Is it really  necessary to ad Xvfb to run test? More like is it necessary to run these test. Flathub is not meant to be CI.

tinywrkb:
@trevorsandy I have here a working packaged [xvfb module](https://github.com/flathub/org.qutebrowser.qutebrowser/blob/master/tests-dependencies.json).

trevorsandy:
> Is it really necessary to ad Xvfb to run test? More like is it necessary to run these test. 

For me, yes - this is why I've included them in the build actions. My build checks ensure the LPub3D renderers built from source are properly built and that they, along with LPub3D application paths, are properly configured.

> Flathub is not meant to be CI.

I don't understand the point of this remark. Flathub builds not just LPub3D but also it's dependent packages from source. Ensureing LPub3D dependencies are successfully built and configured is a prudent and, arguably, necessary part of validating the application's build process. You can see this architecture pattern in moden build systems like meson.

I'm using Falthub iteratively at the moment to validate the LPub3D build actions. Certainly, I've had to perform more iterations than expected but this is due to the nature of the Flathub environment. To compare, building a recipie to generate an AppImage took me 2hrs, and half-day for a Snapcraft receipe - both include Xvfb out-of-the-box. So far, i've burned through almost 2 weeks to get my Flatpak recipie to perform as the others. Anyway, once the LPub3D build is validated, I expect to only submit tagged releases. 

Cheers,

trevorsandy:
> I have here a working packaged xvfb module.

@tinywrkb Thank you. It's a bit of a challenge to get the dependencies right. My module failed because, appearently, the `https://www.x.org/releases/individual/...` url is forbidden for Flathub. However, this url works just fine from my browser and my local flathub-builder build.  I'll use your url `https://xorg.freedesktop.org/archive/individual`. I've also removed the dependencies not present in your recipie except `libXmu` needed for `xauth`. 

Cheers,

hfiguiere:
bot, build io.github.trevorsandy.LPub3D

nedrichards:
> > I can check in a 'test build' with the `Q_PROCESSOR_ARM` directive added. Let me know if this would help and I'll post the SHA.
> 
> Here you go: `2ab166c784c451a4fe86c6f5ad2a61179e972557` _Flatpak test for ARM build eglMakeCurrent failed: 3009 error_.
> 

I build this locally on the Pinebook Pro and got the same error :-( which is sad news. I don't want to hold up progress of the overall any further so I'd recommend checking in a flathub.json that restricts building to x86-64 for now (e.g. https://github.com/flathub/com.spotify.Client/blob/master/flathub.json ) and we can fix this in a PR for a future release.

trevorsandy:
> I don't want to hold up progress of the overall any further...

Ok, I'll add the x86_64 build restriction.

The `aarch64` Flathub build log shows the same error - thanks to the now working build_checks . I know where, in the code, the problem is being source from but I don't have an Android/Arm device to enable an enulator for my IDE. 

The underlying cause may be linked to this bug [[QEGLPlatformContext::makeCurrent: eglError: 3009](https://bugreports.qt.io/browse/QTBUG-53023)] but I'm not sure as this was reported against Qt 5.2 quite some time ago. 

The QEGLPlatformContext: eglMakeCurrent return is generated from this [line](https://github.com/leozide/leocad/blob/315a693b6a3e1491797b3b4f8356e2cfe0d22f34/common/lc_context.cpp#L138)
```
std::unique_ptr<QOpenGLContext> OffscreenContext(new QOpenGLContext());
```
in the Visual Editor which is based on LeoCAD so I imagine LeoCAD will have to exact same behaviour for its Flathub [aarch64](https://flathub.org/builds/#/builders/7/builds/3633) build.

It would be interesting to see if you get the same behaviour running the [LeoCAD aarch64](https://flathub.org/apps/details/org.leocad.LeoCAD) build.

Cheers,

trevorsandy:
I found this related [Qt bug report](https://bugreports.qt.io/browse/QTBUG-49173) and guidance which looks promising. If this attempted fix is unsuccessful, I'll add the x86_64 restriction so the submission can move forward.

Cheers,

hfiguiere:
>  I know where, in the code, the problem is being source from but I don't have an Android/Arm device to enable an enulator for my IDE.

Android is the wrong kind. I have a RaspberryPi for that. Runs plain Linux.

[...]

> in the Visual Editor which is based on LeoCAD so I imagine LeoCAD will have to exact same behaviour for its Flathub [aarch64](https://flathub.org/builds/#/builders/7/builds/3633) build.

Leocad works on my RaspberryPi. (or I never hit the code path)

nedrichards:
bot, build io.github.trevorsandy.LPub3D

nedrichards:
> DEBUG Visual Editor - [OK] Construct OpenGLContext.
DEBUG Visual Editor - [OK] Create OpenGLContext.
DEBUG Visual Editor - [OK] Construct OffscreenSurface.
DEBUG Visual Editor - [OK] Create OffscreenSurface.
QEGLPlatformContext: eglMakeCurrent failed: 3009
DEBUG Visual Editor - [FAIL] Make current OffscreenContext.
Visual Editor error creating OpenGL context.
ERROR 2021-05-26T22:15:45.287 Unable to initialize Visual Editor. 
Unable to initialize Visual Editor.
Could not initialize the application.

Still no joy on the Pinebook Pro. A slightly different error though!

trevorsandy:
> Leocad works on my RaspberryPi. (or I never hit the code path)

This is good. Actually, I did not feel confident about my last attempted fix and I've since done more debugging that leads me to believe the failed eglMakeCurrent behaviour is likely a result of how I'm integrating my LeoCAD library. Of course, it is a but strange that the fault is not present on x86_64 builds. 

Anyway, I'll set up a Travis job to build and check an aarch64 AppImage so I can debug this behaviour. I already know what I want to refactor for my first test. If I'm not successful to quickly resolve this fault, I'll reset this PR to v2.4..3 with the x86_64 restriction.

Cheers,

trevorsandy:
> Still no joy on the Pinebook Pro. A slightly different error though!

Indeed. The updated messages are some debug statements I've inserted to pinpoint where the abnormal return is occurring.

Cheers,

barthalion:
bot, build io.github.trevorsandy.LPub3D

trevorsandy:
@barthalion - Be careful. The current LPub3D commit in the manifest points to my DevOps repo. It is not a tagged release.

Cheers,

barthalion:
@trevorsandy Yup, that's fine, just want to finally test it. As Nick said, aarch64 can be re-enabled later on when issues are sorted out.

barthalion:
Looks good to me at first glance, so I think we're mostly waiting for your green light to merge and publish it.

trevorsandy:
> Looks good to me at first glance.

My last TODO, before I hit the arm runtime break, was to generate some Linux AppData images. The current images are all for Windows build.

Cheers,

trevorsandy:
I've fixed the aarch64 distro. Could you trigger a build to confirm ?

My checks are successful on an `aarch64` `QEMU` Docker container. The aarch64 build log should contain a trace like that shown below. Of course paths will differ.

```
Start build_checks.sh execution at /tmp/build-M1kxu7...

--Buld Check Settings:
--LPUB3D_EXE........./tmp/build-M1kxu7/AppDir/usr/bin/lpub3d24
--SOURCE_DIR........./tmp/build-M1kxu7
--XDG_RUNTIME_DIR..../tmp/runtime-user-1000
--USE_XVFB...........YES
--LP3D_OS_NAME.......Linux
--LP3D_TARGET_ARCH...aarch64
--LP3D_PLATFORM......ubuntu
--SCRIPT_NAME......../builds/check/build_checks.sh

DEBUG Shared Context - [OK] Construct OpenGLContext.
DEBUG Shared Context - [OK] Create OpenGLContext.
DEBUG Shared Context - [OK] Construct OffscreenSurface.
DEBUG Shared Context - [OK] Create OffscreenSurface.
DEBUG Shared Context - [OK] Make current OffscreenContext.

LPub3D v2.4.3 r15 (Dev-release) for Linux 64bit
==========================
Arguments: --version --app-paths
LPub3D, Dev-release 2.4.3, Revision 15, Commit 2802, SHA 194e855c
Compiled on Jun  3 2021

--------------------------
LPub3D Application Folder....(lpub3d)
LPub3D App Data Path.........(/trevorsandy/.local/share/LPub3D Software/lpub3d24)
LPub3D Executable Path.......(/tmp/build-M1kxu7/AppDir/usr/bin)
LPub3D Log Path..............(/trevorsandy/.local/share/lpub3d software/lpub3d24/logs/lpub3dlog.txt)
LPub3D Extras Resource Path..(/tmp/build-M1kxu7/AppDir/usr/share/lpub3d)
LPub3D Renderers Exe Path....(/tmp/build-M1kxu7/AppDir/opt/lpub3d/3rdParty)
LPub3D Config File Path......(/trevorsandy/.config)
LPub3D 3D Editor Cache Path..(/trevorsandy/.cache/LPub3D Software/lpub3d24)
LPub3D Loaded LDraw Library..(LEGO Parts)
--------------------------

------------Build Checks Start--------------

- Check 1 of 7: Native File Process Check... PASSED, ELAPSED TIME [2mins 41sec]
Native File Process Command: --no-stdout-log --process-file --liblego --preferred-renderer native /tmp/build-M1kxu7/builds/check/build_checks.mpd

- Check 2 of 7: LDView File Process Check... PASSED, ELAPSED TIME [1mins 22sec]
LDView File Process Command: --no-stdout-log --process-file --clear-cache --liblego --preferred-renderer ldview /tmp/build-M1kxu7/builds/check/build_checks.mpd

- Check 3 of 7: LDView (Single Call) File Process Check... PASSED, ELAPSED TIME [52sec]
LDView (Single Call) File Process Command: --no-stdout-log --process-file --clear-cache --liblego --preferred-renderer ldview-sc /tmp/build-M1kxu7/builds/check/build_checks.mpd

- Check 4 of 7: LDGLite Export Range Check... PASSED, ELAPSED TIME [1mins 30sec]
LDGLite Export Range Command: --no-stdout-log --process-export --range 1-3 --clear-cache --liblego --preferred-renderer ldglite /tmp/build-M1kxu7/builds/check/build_checks.mpd

- Check 5 of 7: Native POV Generation Check... PASSED, ELAPSED TIME [11mins 34sec]
Native POV Generation Command: --no-stdout-log --process-file --clear-cache --liblego --preferred-renderer povray /tmp/build-M1kxu7/builds/check/build_checks.mpd

- Check 6 of 7: LDView TENTE Model Check... PASSED, ELAPSED TIME [5mins 11sec]
LDView TENTE Model Command: --no-stdout-log --process-file --clear-cache --libtente --preferred-renderer ldview /tmp/build-M1kxu7/builds/check/TENTE/astromovil.ldr

- Check 7 of 7: LDView (Snapshot List) VEXIQ Model Check... PASSED, ELAPSED TIME [50mins 56sec]
LDView (Snapshot List) VEXIQ Model Command: --no-stdout-log --process-file --clear-cache --libvexiq --preferred-renderer ldview-scsl /tmp/build-M1kxu7/builds/check/VEXIQ/spider.mpd

----Build Check Completed: PASS (7)[1,2,3,4,5,6,7], FAIL (0)[], ELAPSED TIME [1hrs 14mins 27sec]----
```

Don't mind the elapsed time, the QEMU environment really slowed things down. the normal test runs in less than five minutes.

Cheers,

hfiguiere:
bot, build io.github.trevorsandy.LPub3D

hfiguiere:
note: you are allowed use the command yourself: https://github.com/flathub/flathub/pull/2302#issuecomment-854781752

trevorsandy:
The aarch64 build was NOT successful. The return was again `Error creating OpenGL context.`.

The log trace was a bit strange however as there were no debug `DEBUG Shared Context` statements but the commit SHA was the last one checked in. The results seem to indicate the latest update was not used but the log trace shows all modules being built.

Cheers,

hfiguiere:
It fail on RaspberryPi too.

trevorsandy:
> It fail on RaspberryPi too.

I don't think the latest updates were picked up. As stated earlier, my checks were successful on my QEMU (Ubuntu/Focal/aarch64) build. I'm just about to check in an update to run the checks on Travis-CI (Ubuntu/Xenial/aarch64). If those checks succeed, I'll create a tag and re-run the FlatHub build with some more debug statements to be sure the latest updates are indeed being built.

Cheers,

trevorsandy:
My [Travis-CI aarch64 build checks](https://travis-ci.com/github/trevorsandy/lpub3d-ci) are successful.

```
Start build_checks.sh execution at /home/travis/build/trevorsandy/lpub3d-ci...

--Buld Check Settings:
--LPUB3D_EXE........./home/travis/build/trevorsandy/AppDir/usr/bin/lpub3d24
--SOURCE_DIR........./home/travis/build/trevorsandy/lpub3d-ci
--XDG_RUNTIME_DIR..../tmp/runtime-user-1000
--USE_XVFB...........YES
--LP3D_OS_NAME.......Linux
--LP3D_TARGET_ARCH...aarch64
--LP3D_PLATFORM......ubuntu
--SCRIPT_NAME......../builds/check/build_checks.sh

DEBUG Shared Context - CreateOffscreenContext
DEBUG Shared Context - [OK] Construct OpenGLContext.
DEBUG Shared Context - [OK] Create OpenGLContext.
DEBUG Shared Context - [OK] Construct OffscreenSurface.
DEBUG Shared Context - [OK] Create OffscreenSurface.
DEBUG Shared Context - [OK] Make current OffscreenContext.

LPub3D v2.4.3 r22 (Dev-release) for Linux 64bit
==========================
Arguments: --version --app-paths
LPub3D, Dev-release 2.4.3, Revision 22, Commit 2809, SHA 91a519b
Compiled on Jun  4 2021

--------------------------
LPub3D Application Folder....(lpub3d)
LPub3D App Data Path.........(/home/travis/.local/share/LPub3D Software/lpub3d24)
LPub3D Executable Path.......(/home/travis/build/trevorsandy/AppDir/usr/bin)
LPub3D Log Path..............(/home/travis/.local/share/lpub3d software/lpub3d24/logs/lpub3dlog.txt)
LPub3D Extras Resource Path..(/home/travis/build/trevorsandy/AppDir/usr/share/lpub3d)
LPub3D Renderers Exe Path....(/home/travis/build/trevorsandy/AppDir/opt/lpub3d/3rdParty)
LPub3D Config File Path......(/home/travis/.config)
LPub3D 3D Editor Cache Path..(/home/travis/.cache/LPub3D Software/lpub3d24)
LPub3D Loaded LDraw Library..(LEGO Parts)
--------------------------
...

----Build Check Completed: PASS (7)[1,2,3,4,5,6,7], FAIL (0)[], ELAPSED TIME [4mins 17sec]----
```

trevorsandy:
> note: you are allowed use the command yourself: #2302 (comment)

Is there instructions somewhere I can follow ? 

I'm not sure what values to enter into the dialogue and I'm receiving the following blocking message:

Screenshot - 05_06_2021 , 00_48_59

Cheers,

trevorsandy:
bot, build io.github.trevorsandy.LPub3D

trevorsandy:
bot, build io.github.trevorsandy.LPub3D

trevorsandy:
bot, build io.github.trevorsandy.LPub3D

trevorsandy:
bot, build io.github.trevorsandy.LPub3D

trevorsandy:
bot, build io.github.trevorsandy.LPub3D

trevorsandy:
bot, build io.github.trevorsandy.LPub3D

trevorsandy:
bot, build io.github.trevorsandy.LPub3D

trevorsandy:
bot, build io.github.trevorsandy.LPub3D

trevorsandy:
bot, build io.github.trevorsandy.LPub3D

trevorsandy:
bot, build io.github.trevorsandy.LPub3D

trevorsandy:
bot, build io.github.trevorsandy.LPub3D

nedrichards:
I'm still getting the eglMakeCurrent failed: 3009 on the latest with pinebook pro. Otherwise, where are we at with your general release readiness? Don't want to be holding you up.

trevorsandy:
> I'm still getting the eglMakeCurrent failed: 3009

It's a challenge. This behaviour appears unique to the flathub build. My builds on QEMU (Ubuntu Focal) and Travis (Ubuntu Xenial) are both passing their checks. Furthermore, I ran a flathub build and check for just the LPub3D Visual Editor module (LeoCAD), the source of the failure in LPub3D, which was also successful. It is also confounding that the failure is only on the aarch build. Taken all together, I'm not confident about releasing any build until I figure out why the context is failing. Without knowing the cause, this behaviour may ultimately have an impact on the amd release also.

Cheers,

trevorsandy:
bot, build io.github.trevorsandy.LPub3D

hfiguiere:
> It's a challenge. This behaviour appears unique to the flathub build.

The more it goes the more I believe that the KDE runtime is buggy on aarch64 when it comes to GL support. There are applications like FreeCAD that works fine on RaspberryPi from Debian, but that crash when build for flathub.

Unfortunately I don't know enough to pinpoint it.

trevorsandy:
bot, build io.github.trevorsandy.LPub3D

nedrichards:
@flathub/kde contains the KDE runtime maintainers who may have more context.

tsdgeos:
> @flathub/kde contains the KDE runtime maintainers who may have more context.

Is this about the aarch64 + GL support?

trevorsandy:
> Is this about the aarch64 + GL support?

@flathub/kde, Yes. 

As you can see in the comments, I'm simply not able to get past `eglMakeCurrent failed: 3009` error on my aarch64 build. 

> This behaviour appears unique to the flathub build. My builds on QEMU (Ubuntu Focal) and Travis (Ubuntu Xenial) are both passing their checks. Furthermore, I ran a flathub build and check for just the LPub3D Visual Editor module (LeoCAD), the source of the failure in LPub3D, which was also successful. It is also confounding that the failure is only on the aarch build.

Any insights you can share is appreciated.

You can search any successful build trace log for `Start Build_Checks.sh execution` to see more details on the failure.

Cheers,

hfiguiere:
My test system is a Raspberry Pi 400, with the official 64-bits RaspberryPi OS (pre-release, etc).

trevorsandy:
I discovered flatpak builder [Github Actions](https://github.com/marketplace/actions/flatpak-builder) so I'm going to try building on GitHub. The aim is to determine if the source of the eglMakeCurrent failed: 3009 failure could be the flathub environment (i.e. perhaps something with the mesa/KDE configuration).

Cheers,

barthalion:
@trevorsandy It's unlikely to change anything as it's technically follows what Flathub does.

nedrichards:
@tsdgeos I have a Pinebook Pro and hub has a modern raspberry pi both of which should "just work", I understand that doesn't particularly help you though ;-)

hfiguiere:
> > @flathub/kde contains the KDE runtime maintainers who may have more context.
> 
> Is this about the aarch64 + GL support?

So I filed https://invent.kde.org/packaging/flatpak-kde-runtime/-/issues/19 and we have a potential patch for the KDE runtime. tl;dr the runtime was not build with "desktop GL" on aarch64.

And I could run (a flatpak build of) LPub3D on my RaspberryPi 400. I'll check the very latest, but it's on the right track here.

trevorsandy:
> I could run (a flatpak build of) LPub3D on my RaspberryPi 400...

That’s good news.

Cheers,

hfiguiere:
Not so much as there is currently no easy fix.

nedrichards:
Right, at this stage the only sensible solution here is to disable aarch64 builds for this app (or port it to GLES from big GL!). Which is very frustrating but this seems to be an underlying Qt restriction as much as anything else.

hfiguiere:
> Right, at this stage the only sensible solution here is to disable aarch64 builds for this app (or port it to GLES from big GL!). Which is very frustrating but this seems to be an underlying Qt restriction as much as anything else.

agreed.

hfiguiere:
sorry dropped the ball. We should try to get this in after disabling aarch64.

@trevorsandy if you don't have time I don't mind taking over.

barthalion:
bot, build io.github.trevorsandy.LPub3D

hfiguiere:
Build failed because of ongoing freedesktop.org infra issues.

barthalion:
bot, build io.github.trevorsandy.LPub3D

barthalion:
No dice.

tsdgeos:
bot, build io.github.trevorsandy.LPub3D

tsdgeos:
fdo.org is behaving better now, trying again

io.github.trevorsandy.LPub3D.json
```diff
@@ -0,0 +1,320 @@
...
+              "url": "https://downloads.sourceforge.net/project/boost/boost/1.60.0/boost_1_60_0.tar.bz2",
+              "sha256": "686affff989ac2488f79a97b9479efb9f2abae035b5ed4d8226de6857933fd3b"
+            }
+          ]
+        },
+        {
+          "name": "openexr",
+          "modules": [
+            {
+              "name": "ilmbase",
```
hfiguiere, line 213:
```suggestion
              "name": "ilmbase",
              "build-options": {
                  "cxxflags": "-std=c++14"
              },
```

barthalion:
bot, build io.github.trevorsandy.LPub3D

io.github.trevorsandy.LPub3D.json
```diff
@@ -0,0 +1,320 @@
...
+                },
+                {
+                  "type": "shell",
+                  "commands": [
+                    "cp -p /usr/share/automake-*/config.{sub,guess} ."
+                  ]
+                }
+              ]
+            }
+          ],
```
hfiguiere, line 231:
```suggestion
          ],
          "build-options": {
              "cxxflags": "-std=c++14"
          },
```

hfiguiere (changes requested):

hfiguiere:
the suggestion above are what is needed to fix the runtime update. I can't push any of these directly.

barthalion:
bot, build io.github.trevorsandy.LPub3D

hfiguiere:
@trevorsandy is seems that your fork of povray has been force-pushed and therefor the revision `54b65529090b4161193df24f0d846046b1495f0b` isn't found

io.github.trevorsandy.LPub3D.json
```diff
@@ -0,0 +1,326 @@
...
+        "./configure COMPILED_BY=\"Trevor SANDY <trevor.sandy@gmail.com> using FlatHub for LPub3D.\" --prefix=${LP3D_DIST_DIR_PATH} LPUB3D_3RD_PARTY=yes --with-libsdl2 --with-boost-libdir=${FLATPAK_DEST}/lib --enable-watch-cursor",
+        "make",
+        "make check",
+        "make install"
+      ],
+      "sources": [
+        {
+          "type": "git",
+          "url": "https://github.com/trevorsandy/povray.git",
+          "commit": "54b65529090b4161193df24f0d846046b1495f0b"
```
hfiguiere, line 186:
```suggestion
          "commit": "59f04740497a259d94922eeebc4ccff34e7510f2"
```

This revision seems to work.

io.github.trevorsandy.LPub3D.json
```diff
@@ -0,0 +1,326 @@
...
+    {
+      "name": "povray",
+      "buildsystem": "simple",
+      "build-commands": [
+        "cd unix && chmod +x prebuild3rdparty.sh && ./prebuild3rdparty.sh",
+        "./configure COMPILED_BY=\"Trevor SANDY <trevor.sandy@gmail.com> using FlatHub for LPub3D.\" --prefix=${LP3D_DIST_DIR_PATH} LPUB3D_3RD_PARTY=yes --with-libsdl2 --with-boost-libdir=${FLATPAK_DEST}/lib --enable-watch-cursor",
+        "make",
+        "make check",
+        "make install"
+      ],
```
hfiguiere, line 181:
```suggestion
      ],
      "build-options": {
        "cxxflags": "-std=c++14"
      },
```

It also need this for the openexr headers

hfiguiere:
with these last changes, I do have a working builld here (x86_64)

barthalion:
bot, build io.github.trevorsandy.LPub3D

PR #2310, opened 2021-05-04
Waveform11

GeorgesStavracas:
Waveform11 is a DAW produced by Tracktion.

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** (no link, all was done via email)

### 🔥 Blockers 🔥

 * [x] The download manager currently doesn't work
 * [x] Should it be called `com.tracktion.Waveform11`, or `com.tracktion.Waveform`?

### 🔴 Important 🔴

 * [x] The AppData file is missing many important fields, like screenshots. I hope Tracktion is able provide marketing material (screenshots, descriptions, etc) to be used here.

### 🟡 Smaller Issues 🟡

 * [x]  The only icon is 1024x1024. Ideally it'd ship an SVG icon, but in the absence of it, it should have at least a 512x512, a 64x64, and 16x16 app icon.
 * [x] The desktop files contain invalid fields; currently I'm editing them while building, but ideally upstream would provide valid desktop files.


com.tracktion.Waveform11.yml
```diff
@@ -0,0 +1,128 @@
...
+  - --share=ipc
+  - --share=network
+  - --socket=pulseaudio
+  - --socket=x11
+
+cleanup:
+  - '*.a'
+  - '*.la'
+  - '/lib/libjack*'
+  - /lib/jack
```
hfiguiere, line 32: deleting jack is now par of the shared-module.

com.tracktion.Waveform11.yml
```diff
@@ -0,0 +1,128 @@
...
+  - --env=LXVST_PATH=/app/extensions/Plugins/lxvst
+  - --env=VST3_PATH=/app/extensions/Plugins/vst3
+  - --env=LADSPA_PATH=/app/extensions/Plugins/ladspa
+  - --filesystem=xdg-config
+  - --filesystem=xdg-documents
+  - --filesystem=xdg-download
+  - --filesystem=xdg-music
+  - --filesystem=xdg-pictures
+  - --filesystem=xdg-run/pipewire-0
+  - --filesystem=xdg-videos
```
hfiguiere, line 20: sadly limit the filesystem can possibly cause some plugins to not work.
It will also prohibit installing VST3 in .vst3

com.tracktion.Waveform11.yml
```diff
@@ -0,0 +1,128 @@
...
+sdk: org.freedesktop.Sdk
+tags:
+  - proprietary
+command: waveform11
+finish-args:
+  - --device=all
+  - --env=LV2_PATH=$HOME/.lv2:/app/extensions/Plugins/lv2
+  - --env=LXVST_PATH=/app/extensions/Plugins/lxvst
+  - --env=VST3_PATH=/app/extensions/Plugins/vst3
+  - --env=LADSPA_PATH=/app/extensions/Plugins/ladspa
```
hfiguiere, line 13: did you check that the app support LADSPA and LV2?
Also did you check it uses LXVST_PATH for VST. Sometime it's VST_PATH.

com.tracktion.Waveform11.yml
```diff
@@ -0,0 +1,128 @@
...
+      - install -Dm755 apply_extra.sh ${FLATPAK_DEST}/bin/apply_extra
+      - install -Dm755 waveform11.sh ${FLATPAK_DEST}/bin/waveform11
+    post-install:
+      - install -d /app/extensions/Plugins
+    sources:
+      - type: extra-data
+        filename: waveform11.deb
+        only-arches:
+          - x86_64
+        url: https://cdn.tracktion.com/file/tracktiondownload/waveform/1159/waveform_64bit_v11.5.9.deb
```
hfiguiere, line 115: you need to add the exclusion in the flathub.json.

hfiguiere (changes requested):

com.tracktion.Waveform11.yml
GeorgesStavracas: Good point

com.tracktion.Waveform11.yml
GeorgesStavracas: I didn't really test it beyond noticing it loaded a bunch of plugins with these flags.

com.tracktion.Waveform11.yml
GeorgesStavracas: What would you suggest here? `home`?

hfiguiere: that's usually the case. Not ideal though, even more since no one can see the source code. I'm just warning of the pitfalls here, and what you set could be a starting point though.

GeorgesStavracas:
I've `strace`d it and managed to kinda sorta make the download manager work. I'm not sure how this download manager app is supposed to operate, but at least Waveform11 detects it and launches it:

Captura de tela de 2021-05-06 18-58-39
Captura de tela de 2021-05-06 18-59-05

Evidently, the download manager app can't install any new apps, but it's able to download non-executable content:

Captura de tela de 2021-05-06 18-59-20

**_(Tracktion developers: notice the bug in the download progress counter above :slightly_smiling_face: - it's missing a division by 10 I guess?)_**

The download manager app merely opens the downloaded files. When I download apps, it tries to open it like a regular compressed archive - on Ubuntu this is supposed to open GNOME Software. Content downloads are opened as a regular archive, and users are supposed to extract the relevant files on their own:

Captura de tela de 2021-05-06 18-59-30

GeorgesStavracas:
This last push exposes the `~/.vst` and `~/.vst3` folders to the sandbox. Added them because they're read by default by Waveform11.

I tried downloading some content using the Download Manager app, and extract it at `~/.vst`, and Waveform11 was able to identify the new plugins, and scan and show them:

Captura de tela de 2021-05-06 19-28-12

So I think at least this part of the Download Manager apps is not broken anymore? :slightly_smiling_face: 

(It's becoming increasingly harder to justify not exposing the entire home folder though.)

hfiguiere:
Are these VST detected by Ardour (flatpak) ?
I haven't tested if it find stuff in `~/.vst`

I know VST3 would.

GeorgesStavracas:
It does, if I manually add `~/.vst` to the list of VST2 search paths.

hfiguiere:
> It does, if I manually add `~/.vst` to the list of VST2 search paths.

cool. for the path, feels like the expected behaviour in Ardour.

GeorgesStavracas:
This last allows this Flatpak to read and write to `~/Tracktion` by default. The reason I added is that this is the path which Waveform11 uses by default.

Feature-wise, this is close to be 100% usable. The download manager is able to function as it does outside the Flatpak sandbox; Waveform11 is able to run through audio backends using ALSA, and JACK (via PipeWire). I'm already producing songs with it.

com.tracktion.Waveform11.yml
```diff
@@ -0,0 +1,137 @@
...
+tags:
+  - proprietary
+command: waveform11
+finish-args:
+  - --device=all
+  - --env=LV2_PATH=$HOME/.lv2:/app/extensions/Plugins/lv2
+  - --env=LXVST_PATH=/app/extensions/Plugins/lxvst
+  - --env=VST3_PATH=/app/extensions/Plugins/vst3
+  - --env=LADSPA_PATH=/app/extensions/Plugins/ladspa
+  - --filesystem=home/.vst:create
```
hfiguiere, line 14: you probably also want `~/.lxvst` as it's the other path for VST(2) on Linux

GeorgesStavracas: Excellent point. Added.

GeorgesStavracas:
So, on the technical side, this is working fantastically now. I've been using it to produce my own songs for a few weeks, and don't notice any drawbacks compared to running it in an Ubuntu machine.

So for us to move forward with this Flatpak, we'll need the following:

 * [ ]  Clarify if the appropriate naming should be `com.tracktion.Waveform11`, or `com.tracktion.Waveform`. This depends on whether Tacktion treats new versions as different products (e.g. if someone has a license to use Waveform11, it will not work on Waveform12), or if it's a new version of the same product.
 * [ ] An SVG icon (ideally), and/or 512x512 + 64x64 + 16x16 icons.
 * [ ] AppData information. Ideally, Tracktion's marketing team would provide the file itself, or at least the exact wording they want us to use. It's also missing screenshots.

com.tracktion.Waveform11.yml
hfiguiere: it would be nice to know which one to remove. This is helpful down the line for maintenance.

GeorgesStavracas:
I took the liberty to write the app metadata myself, using information and screenshots I could find on the website. Since I had no response yet, I reasoned that using `com.tracktion.Waveform` (without `11`) is the safer option; if major releases are different products, we can easily start suffixing new repositories with numbers; or we can also just keep updating this app directly.

com.tracktion.Waveform.appdata.xml
```diff
@@ -0,0 +1,62 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id>com.tracktion.Waveform</id>
+  <name>Waveform11</name>
+  <summary>Waveform is a</summary>
```
hfiguiere, line 5: incomplete sentence?

com.tracktion.Waveform.yml
```diff
@@ -0,0 +1,138 @@
...
+  - /include
+  - /cache
+  - /include
+
+add-extensions:
+  org.freedesktop.LinuxAudio.Plugins:
+    directory: extensions/Plugins
+    version: '20.08'
+    add-ld-path: lib
+    merge-dirs: ladspa;lv2;lxvst;vst3
```
hfiguiere, line 44: still does it support LADSPA and LV2? if not then this should be shorten and the env should be reduced too.

hfiguiere (changes requested):

com.tracktion.Waveform.yml
GeorgesStavracas: Waveform11 supports LADSPA, but not LV2 apparently.

com.tracktion.Waveform.appdata.xml
GeorgesStavracas: Oopsie

com.tracktion.Waveform11.yml
GeorgesStavracas: Cleaned them up now.

com.tracktion.Waveform11.yml
GeorgesStavracas: I'll keep an eye on bug reports, I prefer relaxing constrained filesystem permissions than giving up at start

PR #2319, opened 2021-05-09
Add com.corsixth.corsixth

rmnvgr:

    - Read-only home directory access so that users can import the game assets
- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link: https://github.com/CorsixTH/CorsixTH/issues/1416#issuecomment-835540527**
    - Lua patch: They don't want to allow building shared library from their Makefile (http://lua-users.org/lists/lua-l/2006-03/msg00088.html)
    - CorsixTH patches: The app ID changed after the latest release

cc @TheCycoONE

### DIFF
--- a/corsixth-appdata.patch
+++ b/corsixth-appdata.patch
@@ -1,0 +1,55 @@
+From baec9533810026e55554ce7a6dc8e1817173ca88 Mon Sep 17 00:00:00 2001
+From: Toby <mrtobylane@gmail.com>
+Date: Wed, 26 Aug 2020 00:43:31 +0100
+Subject: [PATCH] Update appdata.xml (#1693)
+
+---
+ ...ppdata.xml => com.corsixth.corsixth.appdata.xml} | 13 +++++++------
+ 1 file changed, 7 insertions(+), 6 deletions(-)
+ rename CorsixTH/{com.corsixth.CorsixTH.appdata.xml => com.corsixth.corsixth.appdata.xml} (88%)
+
+diff --git a/CorsixTH/com.corsixth.CorsixTH.appdata.xml b/CorsixTH/com.corsixth.corsixth.appdata.xml
+similarity index 88%
+rename from CorsixTH/com.corsixth.CorsixTH.appdata.xml
+rename to CorsixTH/com.corsixth.corsixth.appdata.xml
+index 68819ccd94..b252d5bc78 100644
+--- a/CorsixTH/com.corsixth.CorsixTH.appdata.xml
++++ b/CorsixTH/com.corsixth.corsixth.appdata.xml
+@@ -1,6 +1,6 @@
+ <?xml version="1.0" encoding="UTF-8"?>
+ <component type="desktop">
+-  <id>com.corsixth.CorsixTH</id>
++  <id>com.corsixth.corsixth</id>
+   <!-- Copyright 2019 Artem Polishchuk <ego.cordatus@gmail.com> -->
+   <metadata_license>CC0</metadata_license>
+   <project_license>BSD-3-Clause AND MIT</project_license>
+@@ -24,14 +24,14 @@
+     <binary>corsix-th</binary>
+   </provides>
+   <releases>
+-    <release version="0.63" date="2019-05-24"></release>
++    <release version="0.64" date="2020-06-16"></release>
+   </releases>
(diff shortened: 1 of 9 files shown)


rmnvgr:
bot, build com.corsixth.corsixth

corsixth-appdata.patch
```diff
@@ -0,0 +1,55 @@
...
+rename from CorsixTH/com.corsixth.CorsixTH.appdata.xml
+rename to CorsixTH/com.corsixth.corsixth.appdata.xml
+index 68819ccd94..b252d5bc78 100644
+--- a/CorsixTH/com.corsixth.CorsixTH.appdata.xml
++++ b/CorsixTH/com.corsixth.corsixth.appdata.xml
+@@ -1,6 +1,6 @@
+ <?xml version="1.0" encoding="UTF-8"?>
+ <component type="desktop">
+-  <id>com.corsixth.CorsixTH</id>
++  <id>com.corsixth.corsixth</id>
```
hfiguiere, line 22: So why not using this ID instead to not even have to rename?

corsixth-appdata.patch
```diff
@@ -0,0 +1,55 @@
...
+   <!-- Copyright 2019 Artem Polishchuk <ego.cordatus@gmail.com> -->
+   <metadata_license>CC0</metadata_license>
+   <project_license>BSD-3-Clause AND MIT</project_license>
+@@ -24,14 +24,14 @@
+     <binary>corsix-th</binary>
+   </provides>
+   <releases>
+-    <release version="0.63" date="2019-05-24"></release>
++    <release version="0.64" date="2020-06-16"></release>
```
hfiguiere, line 31: Did you submit these upstream?

hfiguiere (changes requested):

corsixth-appdata.patch
rmnvgr: Upstream changed their ID from `com.corsixth.CorsixTH` to `com.corsixth.corsixth`, but it happened after their latest release: https://github.com/CorsixTH/CorsixTH/issues/1416#issuecomment-835636263

I rename for this release, but it won't be necessary for future releases.

corsixth-appdata.patch
rmnvgr: The two CorsixTH patches are actually upstream commits that I cherry-picked:
- https://github.com/CorsixTH/CorsixTH/commit/baec9533810026e55554ce7a6dc8e1817173ca88
- https://github.com/CorsixTH/CorsixTH/commit/8962dfc7880d08ee77c2f81dacdb30c9114f2204

They will be dropped with the next upstream release.

corsixth-appdata.patch
hfiguiere: ok. make sense.

corsixth-appdata.patch
hfiguiere: Excellent.

TheCycoONE: Did you also need https://github.com/CorsixTH/CorsixTH/commit/b595595439726f35b36004cfd6f41b3f61a3a009

corsixth-appdata.patch
bilelmoussaoui: You should add the old id to `<provides><id>oldId</id></provides>`  so that the stores don't show the application twice

rmnvgr: Done, the patch has been submitted upstream here: https://github.com/CorsixTH/CorsixTH/pull/1871

rmnvgr:
> I would consider adding innoextract (example: [flathub/io.openrct2.OpenRCT2@5ef5a99](https://github.com/flathub/io.openrct2.OpenRCT2/commit/5ef5a9924ca90a5193d43c6b81b81599759adcdd)) just so it's easier to extract the installer.

OpenRCT2 uses `innoextract` itself, I don't think that's the case with CorsixTH (@TheCycoONE?), users have to run it themselves. Is there a point of including it given that users may not know that it's included in the flatpak?

barthalion:
Up to you; I ended up installing OpenRCT2 just to extract the installer here for testing. Probably doesn't make much sense if it's too opaque. Can be handled later anyway.

/merge @TheCycoONE

PR #2323, opened 2021-05-12
Add net.sourceforge.dmidiplayer

pedrolcl:
initial commit

- [x] All assets referenced in the manifest are redistributable by any party.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "shared-modules"]
+	path = shared-modules
+	url = https://github.com/flathub/shared-modules/
--- a/net.sourceforge.dmidiplayer.json
+++ b/net.sourceforge.dmidiplayer.json
@@ -1,0 +1,94 @@
+{
+  "app-id": "net.sourceforge.dmidiplayer",
+  "runtime": "org.kde.Platform",
+  "runtime-version": "5.15",
+  "sdk": "org.kde.Sdk",
+  "command": "dmidiplayer",
+  "rename-icon": "dmidiplayer",
+  "finish-args": [
+    /* X11 + XShm access */
+    "--share=ipc",
+    "--socket=x11",
+    /* Note playback */
+    "--socket=pulseaudio",
+    /* MIDI */
+    "--device=all",
+    /* Allow loading, saving files from anywhere (portals don’t work yet) */
+    "--filesystem=home",
+    /* Allow other instances to see lockfiles */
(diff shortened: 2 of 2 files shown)


pedrolcl:
bot, build net.sourceforge.dmidiplayer

nedrichards:
This is a Qt app right? Any special reason it doesn't support wayland or file access portals? Not a blocker, just interested. Other than that lgtm.

hfiguiere:
A lot of Qt app bug with Wayland (at least within Mutter).

Actually by default Wayland QPA should be disabled on such sessions except that the KDE sdk revert the patch.

(I haven't check this one)

pedrolcl:
@nedrichards: This flatpak is modelled after the VMPK one by @hfiguiere, both applications depending on Drumstick. So, I really didn't think too much about Wayland or Portals, I've simply duplicated that part. What I know for sure is that VMPK needs XCB/X11 as a hard dependency, but this probably can be relaxed in dmidiplayer in the future (will need some work, though). On the other hand, dmidiplayer reads more file types. Exactly like VMPK, this program needs to load soundfont files (usually without opening the file open dialog or any other user interaction) but also MIDI/Karaoke and Cakewalk files provided in the command line, contained in playlist files, after an open file dialog, or a drag&drop operation from the file manager.

hfiguiere:
I had disabled wayland on VMPK because there was a bug with menus not being position properly bug that occured only with Wayland.

However
1. it's possible the bug is fixed in more recent Qt 
2. I tried dmidiplayer with `--socket=wayland` and it seemed to work ok.

In the latter case, there is still the classic `qt.qpa.wayland: Wayland does not support QWindow::requestActivate()`, but beside spaming the console it's usually harmless (and absolutely not specific to this app)

pedrolcl:
@hfiguiere : when running vmpk with `--socket=wayland`, the worst problem is that the option "Raw Computer Keyboard" does not work at all. This option is very important for users using non-English national keyboard layouts with diacritics (dead keys), because it allows mapping such keys for note input. On the other hand, in dmidiplayer that is not an issue, and I can confirm it works well with it.

I've tested the flatpak in Fedora 34 and Ubuntu 21.04, with different results. The sound in Fedora 34 doesn't work at all most of the time, and changing the synth backend freezes the application. In Ubuntu 21.04 I've not observed this problem.

barthalion:
/merge

Feel free to address adding `--socket=wayland` in the newly created repo. Thanks!

PR #2390, opened 2021-06-23
Add org.kde.keysmith

espidev:

* dbus is needed for activating the window on an attempt to open a second instance
- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://invent.kde.org/utilities/keysmith 

### DIFF
--- a/org.kde.keysmith.json
+++ b/org.kde.keysmith.json
@@ -40,6 +40,10 @@
                 {
                     "type": "patch",
                     "path": "0001-add-url.patch"
+                },
+                {
+                    "type": "patch",
+                    "path": "0002-lower-case-dbus.patch"
                 }
             ]
         }
--- a/org.kde.keysmith.json
+++ b/org.kde.keysmith.json
@@ -1,0 +1,45 @@
+{
+    "id": "org.kde.keysmith",
+    "runtime": "org.kde.Platform",
+    "runtime-version": "5.15",
+    "sdk": "org.kde.Sdk",
+    "command": "keysmith",
+    "rename-icon": "keysmith",
+    "finish-args": [
+        "--share=ipc",
+        "--share=network",
+        "--socket=x11",
+        "--socket=wayland",
+        "--device=dri",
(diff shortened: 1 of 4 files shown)


espidev:
bot, build org.kde.keysmith

espidev:
bot, build org.kde.keysmith

org.kde.keysmith.json
```diff
@@ -0,0 +1,48 @@
...
+    "runtime-version": "5.15",
+    "sdk": "org.kde.Sdk",
+    "command": "keysmith",
+    "rename-icon": "keysmith",
+    "finish-args": [
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--device=dri",
+        "--own-name=org.kde.Keysmith"
```
hfiguiere, line 13: Maybe if you set the app-id to match exactly, you don't need to "own-name". That's probably the best.

hfiguiere (changes requested):

org.kde.keysmith.json
espidev: Do you mean the "id" field or in the application itself?

hfiguiere: This probably require to rename the appdata file.

espidev: org.kde.Keysmith for the "id" seems inconsistent with other kde apps, not sure if there are any implications of that

hfiguiere: the issue here is that by default the app own the dbus name that match the id, so the two should match. Now the question is whether you are sure of the dbus name?

nedrichards: Consistency between apps is much less important than matching whatever the app itself expresses - although if we want to be consistent in KDE going into the codebase and changing it isn't *that* hard ;-)

espidev:
bot, build org.kde.keysmith

espidev:
bot, build org.kde.keysmith

PR #2419, opened 2021-07-11
Add org.sigxcpu.Livi

agx:

- [x] I am using only the minimal set of permissions - unfortunately full device access is needed for the video/media devices needed by gstreamer's v4l stateless codecs
- [x] All assets referenced in the manifest are redistributable by any party
- [x] I am an upstream contributor to the project. \
- [x] I own the domain used in the application ID `sigxcpu.org`
- [x] Any additional patches or files have been submitted to the upstream projects concerned. https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/-/merge_requests/2398

### DIFF
--- a/build-aux/0001-play-Emit-correct-signal.patch
+++ b/build-aux/0001-play-Emit-correct-signal.patch
@@ -1,0 +1,28 @@
+From a122f67365dc593c362d00dfc8a29f737c6e3c51 Mon Sep 17 00:00:00 2001
+Message-Id: <a122f67365dc593c362d00dfc8a29f737c6e3c51.1625836942.git.agx@sigxcpu.org>
+From: =?UTF-8?q?Guido=20G=C3=BCnther?= <agx@sigxcpu.org>
+Date: Fri, 9 Jul 2021 14:55:43 +0200
+Subject: [PATCH] play: Emit correct signal
+
+SIGNAL_MEDIA_INFO_UPDATED should be emitted on media info changes,
+not SIGNAL_VIDEO_DIMENSIONS_CHANGED.
+---
+ gst-libs/gst/play/gstplay-signal-adapter.c | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/gst-libs/gst/play/gstplay-signal-adapter.c b/gst-libs/gst/play/gstplay-signal-adapter.c
+index cc5f0e0ce4..cf8b57e198 100644
+--- a/gst-libs/gst/play/gstplay-signal-adapter.c
++++ b/gst-libs/gst/play/gstplay-signal-adapter.c
+@@ -168,7 +168,7 @@ gst_play_signal_adapter_emit (GstPlaySignalAdapter * self,
+       GstPlayMediaInfo *media_info;
+       gst_structure_get (message_data, GST_PLAY_MESSAGE_DATA_MEDIA_INFO,
+           GST_TYPE_PLAY_MEDIA_INFO, &media_info, NULL);
+-      g_signal_emit (self, signals[SIGNAL_VIDEO_DIMENSIONS_CHANGED], 0,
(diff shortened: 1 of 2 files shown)


agx:
Is there a way to restrict this package to amd64 since that's the main target?

hfiguiere:
bot, build org.sigxcpu.Livi

hfiguiere:
> Is there a way to restrict this package to amd64 since that's the main target?

hfiguiere:
but why restricting to PC?

hfiguiere:
upstream appdata is incomplete.

agx:
metainfo validates with `appstreamcli` *and* `appstream-util` now. I had used the former only before.

agx:
> but why restricting to PC?

@hfiguiere i meant to say arm64 aka aarch64. My main focus is to make video smooth on linux based mobile phones which are all aarch64 atm. It will work in principle on PC but we e.g. don't build libav etc so playback there would be slower.

hfiguiere:
> metainfo validates with `appstreamcli` _and_ `appstream-util` now. I had used the former only before.

You should use the one provided as a flatpak as `org.freedesktop.appstream-glib`.
It has more checks that match what flathub expect.

hfiguiere:
> > but why restricting to PC?
> 
> @hfiguiere i meant to say arm64 aka aarch64. My main focus is to make video smooth on linux based mobile phones which are all aarch64 atm. It will work in principle on PC but we e.g. don't build libav etc so playback there would be slower.

That's a first. If it builds and run on x86_64 I see no reason to not build it.

agx:
@hfiguiere Dropped the aarch64 restriction

barthalion:
bot, build org.sigxcpu.Livi

barthalion:
It doesn't seem to work on x86_64:

```
ERROR:../src/livi-gst-paintable.c:128:gtk_gst_paintable_video_renderer_create_video_sink: assertion failed: (self->context != NULL && ctx != NULL)
Bail out! ERROR:../src/livi-gst-paintable.c:128:gtk_gst_paintable_video_renderer_create_video_sink: assertion failed: (self->context != NULL && ctx != NULL)
```

agx:
@barthalion it works here, both the one i built locally and the one from the `flatpakref`. Are you on wayland and OpenGL works?

hfiguiere:
If I run it wit `--h264-demo`, I get
```
(livi:2): livi-window-WARNING **: 12:13:13.180: V4L stateless codec not in use, playback will likely be slow
```

I get a failure passing any video filename to the command line.

agx:
@hfiguiere 

> (livi:2): livi-window-WARNING **: 12:13:13.180: V4L stateless codec not in use, playback will likely be slow

that's expected, see my remark about focusing on linux phones above. The demo does play though, right?

As for command line: It does handle videos on the command line *but* I did not set `--filesystem=home` on purpose since the main use case is being launched by "gio open <file|URL>" (which is what happens when you e.g. click on a video in nautilus etc). I can set the though if it's a requirement.

agx:
@hfiguiere  I added `--filesystem=xdg-videos` which i think make sense for a video player so:

`flatpak run org.sigxcpu.Livi Videos/<whatever>mp4` should work now

hfiguiere:
bot, build org.sigxcpu.Livi

barthalion:
I'd consider adding some aliases to the desktop file like `Livi` and `uPlayer` as it's difficult to trigger the right search keyword in GNOME Shell to find it.

> it works here, both the one i built locally and the one from the `flatpakref`. Are you on wayland and OpenGL works? It can't work without it (the error basically means it couldn't great the GL context)

I switched to Wayland and it indeed works there. Does it mean X11 is unsupported? If so, `fallback-x11` should be dropped from finish-args. Other than this, LGTM.

agx:
@barthalion x11 is indeed unsupported, dropped `fallback-x11`. Thanks for reviewing!

PR #2461, opened 2021-08-13
Add com.virustotal.VirusTotalUploader

tim77:

  `--filesystem=host:ro` because drag-n-drop is the primary way in app for adding files.
- [ ] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://github.com/VirusTotal/qt-virustotal-uploader/issues/14
  - https://github.com/VirusTotal/qt-virustotal-uploader/pull/15

---

Handy app which could useful for community. No longer actively developed.

### DIFF
--- a/com.virustotal.VirusTotalUploader.appdata.xml
+++ b/com.virustotal.VirusTotalUploader.appdata.xml
@@ -10,6 +10,9 @@
             it for uploading and scanning.
         </p>
     </description>
+    <releases>
+        <release date="2014-07-10" version="1.2"></release>
+    </releases>
     <categories>
         <category>Network</category>
         <category>Security</category>
@@ -24,4 +27,5 @@
             <image>https://virustotalcloud.appspot.com/static/img/osx-uploader/osx-app-screen.png</image>
         </screenshot>
     </screenshots>
+    <content_rating type="oars-1.1" />
 </component>
--- a/com.virustotal.VirusTotalUploader.appdata.xml
+++ b/com.virustotal.VirusTotalUploader.appdata.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <component type="desktop-application">
-    <id>com.virustotal.Uploader</id>
+    <id>com.virustotal.VirusTotalUploader</id>
     <name>VirusTotal Uploader</name>
     <summary>Upload files to virustotal.com</summary>
     <description>
@@ -18,7 +18,7 @@
     <project_license>Apache-2.0</project_license>
     <url type="homepage">https://www.virustotal.com/ru/documentation/desktop-applications/linux-oss-uploader</url>
     <url type="bugtracker">https://github.com/VirusTotal/qt-virustotal-uploader/issues</url>
-    <launchable type="desktop-id">com.virustotal.Uploader.desktop</launchable>
+    <launchable type="desktop-id">com.virustotal.VirusTotalUploader.desktop</launchable>
     <screenshots>
         <screenshot type="default">
             <image>https://virustotalcloud.appspot.com/static/img/osx-uploader/osx-app-screen.png</image>
--- a/com.virustotal.VirusTotalUploader.yml
+++ b/com.virustotal.VirusTotalUploader.yml
@@ -5,19 +5,9 @@ runtime-version: '5.15'
 command: VirusTotalUploader
 finish-args:
   - --device=dri
-  - --filesystem=/media:ro
-  - --filesystem=/mnt:ro
-  - --filesystem=/opt:ro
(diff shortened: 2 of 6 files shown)


tim77:
bot, build com.virustotal.VirusTotalUploader

com.virustotal.VirusTotalUploader.yml
```diff
@@ -0,0 +1,55 @@
+id: com.virustotal.VirusTotalUploader
+sdk: org.kde.Sdk
+runtime: org.kde.Platform
+runtime-version: '5.15'
+command: VirusTotalUploader
+finish-args:
+  - --device=dri
+  - --filesystem=host:ro
```
hfiguiere, line 8: wouldn't `home:ro` be enough?

com.virustotal.VirusTotalUploader.appdata.xml
```diff
@@ -0,0 +1,27 @@
...
+    <metadata_license>CC0-1.0</metadata_license>
+    <project_license>Apache-2.0</project_license>
+    <url type="homepage">https://www.virustotal.com/ru/documentation/desktop-applications/linux-oss-uploader</url>
+    <url type="bugtracker">https://github.com/VirusTotal/qt-virustotal-uploader/issues</url>
+    <launchable type="desktop-id">com.virustotal.VirusTotalUploader.desktop</launchable>
+    <screenshots>
+        <screenshot type="default">
+            <image>https://virustotalcloud.appspot.com/static/img/osx-uploader/osx-app-screen.png</image>
+        </screenshot>
+    </screenshots>
```
hfiguiere, line 26: missing `releases` and `oars` tags.

hfiguiere (changes requested):

com.virustotal.VirusTotalUploader.yml
tim77: OK, but since there is not much difference from security side maybe instead of host/home even better to list all `xdg-*` dirs, removable devices `filesystem=/media:ro`, `filesystem=/mnt:ro`, `filesystem=/run/media` and probably `/opt`?

tim77:
bot, build com.virustotal.VirusTotalUploader

com.virustotal.VirusTotalUploader.yml
gasinvein: @hfiguiere I think not, a common use case is to check files on removable drives, so `host` can be replaced with `home` + `/run/media` + `/media`, which is largely the same as `host`.

tim77: Probably `filesystem=xdg-data:ro` and `filesystem=xdg-cache:ro` to much and should removed. Need your opinion here.

gasinvein: Well, if the goal is secure-by-default, then it's certainly too much, as `~/.local/share` and `~/.cache` contain a lot of sensitive data.

barthalion: Really, the proper way is to use portals. Direct filesystem permissions are just hit&miss for non-obvious apps like this one. Why wouldn't user want to upload a file directly from their top-level home directory here?

vchernin: [This is apparently a QT 5 app](https://github.com/VirusTotal/qt-virustotal-uploader#generic-instructions-for-compiling-and-install), does that include the relevant portals?

barthalion: I have removed all filesystem permissions and it seems to work fine for me.

tim77: @barthalion which permission settings you advice using here? So maybe allow just `Downloads` if portals works in this app?

barthalion: No flesystem permissions are needed when portal is used.

gasinvein: @barthalion You can select file from the app menu even if no filesystem permissions, yes, but drag-and-dropping files to the app window won't work.

tim77: Yep. Tested now and portals works in this app, but DnD not if we remove all FS permissions. That's why i mentioned in OP why we need more wide restriction since app even have title with DnD mention. `host` is to much for sure, so can we find middle ground here?

gasinvein: I'd say `host:ro` is fine. Otherwise, we're guaranteed to have users for whom DnD would be broken, no matter how many smaller permissions we add.

tim77: After some discussion reverted `host:ro` with some commentary. We ready for merge it?

PR #2462, opened 2021-08-14
Add ru.proninyaroslav.libretrack

proninyaroslav:

 - `--persist=.dart_tool`: required for storing Floor database https://pub.dev/packages/floor. Proposing a change on the XDG path https://github.com/vitusortner/floor/issues/593


flathub.json
```diff
@@ -0,0 +1,6 @@
+{
+  "only-arches": [
```
hfiguiere, line 2: you don't need this these are the only arch supported today.

ru.proninyaroslav.libretrack.metainfo.xml
```diff
@@ -0,0 +1,65 @@
...
+        <screenshot>
+            <image type="source">https://github.com/proninyaroslav/libretrack/blob/master/art/linux_screenshots/s7.png</image>
+        </screenshot>
+    </screenshots>
+    <launchable type="desktop-id">ru.proninyaroslav.libretrack.desktop</launchable>
+    <releases>
+        <release version="1.2.0" date="2021-08-12"/>
+    </releases>
+    <update_contact>proninyaroslav@gmail.com</update_contact>
+</component>
```
hfiguiere, line 65: it's missing the oars rating. (required)

ru.proninyaroslav.libretrack.yaml
```diff
@@ -0,0 +1,99 @@
...
+finish-args:
+  # Required due to being a GUI application
+  - --socket=x11
+  - --socket=fallback-x11
+  # Required to make sure x11 performance is achived on all platforms
+  # At least that's what the legends tell. it might be worth experimenting
+  # with dropping this permission.
+  - --share=ipc
+  # Required for experimental wayland support
+  - --socket=wayland
```
hfiguiere, line 16: it's either `wayland` + `fallback-x11` or just `x11`.

ru.proninyaroslav.libretrack.yaml
```diff
@@ -0,0 +1,99 @@
...
+  # Required for notifications in various desktop environments
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.kde.StatusNotifierWatcher
+  # Required due to KDE's special way of handling notifications
+  - --own-name=org.kde.StatusNotifierItem-2-1
+  # Required for advanced input methods e.g. writing CJK languages
+  - --talk-name=org.freedesktop.portal.Fcitx
+  # Required for correct work of Flutter packages, which use system DBus
+  # for example https://pub.dev/packages/connectivity_plus
+  - --socket=system-bus
```
hfiguiere, line 31: can't you just allow the proper dbus names?

ru.proninyaroslav.libretrack.yaml
```diff
@@ -0,0 +1,99 @@
...
+  # TODO: Temporary fix https://github.com/canonical/dbus.dart/issues/257
+  - --socket=session-bus
```
hfiguiere, line 33: and here too?

Wildcard are allowed.

`--log-session-bus` and `--log-system-bus` are usefull to figure it out.

ru.proninyaroslav.libretrack.yaml
```diff
@@ -0,0 +1,99 @@
...
+      - -DBUILD_SHARED_LIBS=ON
+    sources:
+      - type: archive
+        url: https://github.com/open-source-parsers/jsoncpp/archive/refs/tags/1.7.5.tar.gz
+        sha256: 4338c6cab8af8dee6cdfd54e6218bd0533785f552c6162bb083f8dd28bf8fbbe
+  - name: libretrack
+    buildsystem: simple
+    only-arches:
+      - x86_64
+      - aarch64
```
hfiguiere, line 57:
```suggestion
```

ru.proninyaroslav.libretrack.yaml
```diff
@@ -0,0 +1,99 @@
...
+    build-commands:
+      - mkdir -p /app
+      - cp -r libretrack /app/libretrack
+      - chmod +x /app/libretrack/libretrack
+      - install -Dm644 /app/libretrack/data/icons/app-icon.svg /app/share/icons/hicolor/scalable/apps/ru.proninyaroslav.libretrack.svg
+      - mkdir -p /app/bin /app/lib
+      - ln -s /usr/lib/x86_64-linux-gnu/libsqlite3.so.0 /app/lib/libsqlite3.so
```
hfiguiere, line 64: this will fail on aarch64.
Also why do you need it?

ru.proninyaroslav.libretrack.yaml
```diff
@@ -0,0 +1,99 @@
...
+      - ln -s /app/libretrack/libretrack /app/bin/libretrack
+      - install -Dm644 ru.proninyaroslav.libretrack.desktop /app/share/applications/ru.proninyaroslav.libretrack.desktop
+      - install -Dm644 ru.proninyaroslav.libretrack.metainfo.xml /app/share/appdata/ru.proninyaroslav.libretrack.metainfo.xml
+      - install libretrack.sh /app/bin/libretrack
+    sources:
+      - type: archive
+        only-arches:
+          - x86_64
+        url: https://github.com/proninyaroslav/libretrack/releases/download/1.2.0/libretrack-1.2.0-linux-x86_64.tar.gz
```
hfiguiere, line 73: can't you build from source?

hfiguiere (changes requested):

flathub.json
proninyaroslav: In other words, arm64 is not supported?

ru.proninyaroslav.libretrack.yaml
proninyaroslav: No, because `dbus` package tries to open the system socket file physically and fails if it's not available.

ru.proninyaroslav.libretrack.yaml
proninyaroslav: I have given the reason why I can't use `talk-name` at the moment https://github.com/canonical/dbus.dart/issues/257

ru.proninyaroslav.libretrack.yaml
proninyaroslav: `libsqlite3` is not automatically linked to the `/lib` dir. At runtime, I will get an exception because the app can't find the library. I will add a check.

ru.proninyaroslav.libretrack.yaml
proninyaroslav: Build requires internet connection as I get Dart packages from pub.dev as well as some Linux-specific dependencies for Flutter. In addition, I need to download Flutter itself. As far as I know, Flatpak build system does not allow internet connection during project build, only to get dependencies.

ru.proninyaroslav.libretrack.yaml
proninyaroslav: https://github.com/canonical/dbus.dart/blob/main/lib/src/dbus_client.dart#L213

proninyaroslav: I'm getting this error:

`flutter: SocketException: Connection failed (OS Error: No such file or directory, errno = 2), address = /run/dbus/system_bus_socket, port = 0`

proninyaroslav: First of all, they try to get `DBUS_SYSTEM_BUS_ADDRESS` env variable, and if it's not set then `unix:path=/run/dbus/system_bus_socket` is used.

flathub.json
hfiguiere: only `aarch64` and `x86_64` so its not needed to say they are the only one.

(`arm64` is `aarch64` in case it wasn't clear)

ru.proninyaroslav.libretrack.yaml
hfiguiere: fair enough. :-(

hfiguiere:
bot, build ru.proninyaroslav.libretrack

ru.proninyaroslav.libretrack.yaml
```diff
@@ -0,0 +1,96 @@
+app-id: ru.proninyaroslav.libretrack
+runtime: org.gnome.Platform
```
hfiguiere, line 2: btw do you really need the GNOME SDK/Platform?

proninyaroslav:
@hfiguiere 
Hmm, what could be wrong with the screenshots? As I understand it, the problem is in the presence of the alpha channel?

proninyaroslav:
@hfiguiere 
I removed the alpha channel but it didn't help. Are there any size or aspect ratio restrictions? I haven't found any documentation on this matter.

hfiguiere:
```
$wget https://github.com/proninyaroslav/libretrack/blob/master/art/linux_screenshots/s1.png
[...]
$ file s1.png 
s1.png: HTML document, UTF-8 Unicode text, with very long lines
```

hfiguiere:
tl;dr this doesn't return the PNG.

proninyaroslav:
Yes, my mistake, I forgot to specify `raw`. Thank you.

proninyaroslav:
Fixed.

proninyaroslav:
bot, build ru.proninyaroslav.libretrack

ru.proninyaroslav.libretrack.yaml
proninyaroslav: I have automated the build, so it shouldn't cause any problems in the future https://github.com/proninyaroslav/libretrack/blob/master/.github/workflows/flutter_release_build.yml

proninyaroslav:
@hfiguiere 
I updated the tarballs and I removed the `session-bus` option which was used as a temporary solution.

hfiguiere (changes requested):

ru.proninyaroslav.libretrack.yaml
proninyaroslav: I opened an issue on this problem, but it will be fixed in the future https://github.com/canonical/dbus.dart/issues/260

nedrichards:
bot, build ru.proninyaroslav.libretrack

nedrichards:
I see there's a new release of the dbus package that fixes the issue you reported. Will that allow you to remove the broad permission? That's the only thing standing in the way of merging for me.

proninyaroslav:
@nedrichards 
Yes, it allowed to remove `session-bus`. But `system-bus` still remains https://github.com/canonical/dbus.dart/issues/260. Maybe there is no solution to this problem, I don't know for sure.

barthalion:
I posted a comment there.

barthalion:
@proninyaroslav You can track all DBus activity with `flatpak run --log-system-bus`. Blanket access to the system bus is a no-go without a valid reason, and apparently you are well aware of what may be required to make it work.

proninyaroslav:
@barthalion 
I understand. As I said earlier, I'm using the `connectivity_plus` package to check the Internet connection https://github.com/proninyaroslav/libretrack/blob/1.2.0/lib/core/work_manager/fallback_work_manager.dart#L267
This package, like several others, uses the system bus https://github.com/fluttercommunity/plus_plugins/blob/package_info_plus_linux-v1.0.3/packages/connectivity_plus/connectivity_plus_linux/lib/src/network_manager.dart#L20
Could this be a valid reason?

barthalion:
You can specify which system D-Bus names should be allowed with `--system-talk-name=` (it can be passed multiple times). That should make it work.

proninyaroslav:
@barthalion 
Yes, I tried this, but it still threw an exception. Now I checked it again and it seems `dbus` has fixed it. I'll commit it.

proninyaroslav:
bot, build ru.proninyaroslav.libretrack

proninyaroslav:
@barthalion  @nedrichards
Currently `system-bus` has been removed and only `--persist=.dart_tool` remains from non-standard permissions. Maybe there is another way of access, but this is the only one that I found in the documentation.

PR #2478, opened 2021-08-24
Add org.keshavnrj.wonderwall

keshavbhatt:


org.keshavnrj.wonderwall.yaml
```diff
@@ -0,0 +1,46 @@
+app-id: org.keshavnrj.wonderwall
+base: io.qt.qtwebkit.BaseApp
+base-version: '5.11'
+runtime: org.kde.Platform
+runtime-version: '5.11'
```
hfiguiere, line 5: This runtime version is obsolete. 5.15 is the current one.

org.keshavnrj.wonderwall.yaml
```diff
@@ -0,0 +1,46 @@
...
+sdk: org.kde.Sdk
+command: wonderwall
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=x11
+  - --socket=wayland
+  - --filesystem=host
```
hfiguiere, line 13: why `host`?

org.keshavnrj.wonderwall.yaml
```diff
@@ -0,0 +1,46 @@
...
+base: io.qt.qtwebkit.BaseApp
+base-version: '5.11'
+runtime: org.kde.Platform
+runtime-version: '5.11'
+sdk: org.kde.Sdk
+command: wonderwall
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=x11
```
hfiguiere, line 11:
```suggestion
  - --socket=fallback-x11
```

If you do not support wayland then remove the wayland line and keep this.

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64"]
```
hfiguiere, line 2: why only x86_64?

org.keshavnrj.wonderwall.yaml
```diff
@@ -0,0 +1,46 @@
+app-id: org.keshavnrj.wonderwall
+base: io.qt.qtwebkit.BaseApp
+base-version: '5.11'
+runtime: org.kde.Platform
+runtime-version: '5.11'
+sdk: org.kde.Sdk
+command: wonderwall
+finish-args:
+  - --share=network
```
hfiguiere, line 9: are network permissions necessary?

org.keshavnrj.wonderwall.yaml
```diff
@@ -0,0 +1,46 @@
...
+  - --talk-name=org.gtk.vfs
+  - --talk-name=org.gtk.vfs.*
+  - --talk-name=org.freedesktop.FileManager1
+
+  #- --filesystem=/var/lib/lightdm-data:rw
+
+  - --filesystem=xdg-run/dconf
+  - --filesystem=~/.config/dconf:ro
+  - --talk-name=ca.desrt.dconf
+  - --env=DCONF_USER_CONFIG_DIR=.config/dconf
```
hfiguiere, line 26:
```suggestion
```

This is not needed at all

org.keshavnrj.wonderwall.yaml
```diff
@@ -0,0 +1,46 @@
...
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=x11
+  - --socket=wayland
+  - --filesystem=host
+  - --filesystem=xdg-download
+  - --device=dri
+  
+  - --talk-name=org.gtk.vfs
```
hfiguiere, line 17:
```suggestion
```

This one has always been a mistake

org.keshavnrj.wonderwall.yaml
```diff
@@ -0,0 +1,46 @@
+app-id: org.keshavnrj.wonderwall
```
hfiguiere, line 1: This is supposed to match a domain that exist our you are in control (or a namespace if it's a hosting like github/gitlab/sourceforge).

hfiguiere (changes requested):

org.keshavnrj.wonderwall.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - DEFINES+=USE_DBUS
+      - INCLUDEPATH+="${FLATPAK_DEST}/include/QtWebKitWidgets"
+      - INCLUDEPATH+="${FLATPAK_DEST}/include/QtWebKit"
+      - LRELEASE=lrelease
+    build-options:
+      env:
+        QMAKEPATH: "/app/lib"
+    sources:
+      - type: archive
+        url: "http://ktechpit.com/USS/ww_flathub_build.zip"
```
hfiguiere, line 45: and one last thing: an unversioned archive (ie the version isn't the filename) is a source of trouble because if there is a new version, the build will break.

org.keshavnrj.wonderwall.yaml
keshavbhatt: I thought QtWebkit module may not be available in Latest versions. But looks like there is base app for 5.15 available. I will use 5.15
Thanks.

org.keshavnrj.wonderwall.yaml
keshavbhatt: Removed `- --filesystem=host`

org.keshavnrj.wonderwall.yaml
keshavbhatt: If i user 5.15 where wayland is supported, should I keep wayland? Also What if I keep all three of these  
  ```
- --socket=x11
  - --socket=fallback-x11
  - --socket=wayland
```

flathub.json
keshavbhatt: I will remove this in next PR. Thanks

org.keshavnrj.wonderwall.yaml
keshavbhatt: Yes, the pulls wallpapers from web. That's the only reason my application require network.

org.keshavnrj.wonderwall.yaml
keshavbhatt: Will remove this in next PR, I found this line in other flathub flatpaks too :(

org.keshavnrj.wonderwall.yaml
keshavbhatt: I don't know what each of these lines do in flatpak but, my application need to call gsettings to set wallpaper on gnome based desktop environments. I found a wallpaper app in flathub that uses these lines, I also tested removing these lines and my app was not able to set wallpaper.

org.keshavnrj.wonderwall.yaml
keshavbhatt: Can i put my website's domain name here? com.ktechpit.wonderwall would be fine?
If this is optional i would prefer the keep the previous values since i will require to change it in many places. :(

org.keshavnrj.wonderwall.yaml
keshavbhatt: Will try to fix this in upcoming PR.

Thanks a lot for all the suggestions. I am actually very new to flatpak.

keshavbhatt:
Since the app domain is changed to new one, I think i should make a new PR for the app submission.

org.keshavnrj.wonderwall.yaml
hfiguiere: it is not optional. app-id are meant to be unique, so if you don't own the namespace, it's just open for trouble.

hfiguiere:
you didn't need to make a new PR. pushing to the existing branch just work.

keshavbhatt:
I am sorry if that cause any form of trouble. I thought the new namespace for app domain will give issue in the same PR. Made a new PR with mentioned improvements. If anything need review i am open to make changes.

PR #2481, opened 2021-08-25
Add com.github.Matoking.protontricks

Matoking:
Convert Protontricks into a stand-alone Flatpak application. Most of the build manifest is based on the existing com.valvesoftware.Steam.Utility.protontricks.

---

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** N/A

Until now, Protontricks has been maintained as a Flatpak extension on [com.valvesoftware.Steam.Utility.protontricks](github.com/flathub/com.valvesoftware.Steam.Utility.protontricks). However, Protontricks 1.6.0 now ships additional desktop entry files to launch  the GUI and add Protontricks specific actions for EXE files. Those desktop files can't be installed into the host system inside a Flatpak extension, making it necessary to convert Protontricks into a stand-alone application.

This comes with a set of changes into the build manifest:

* We copy most of the `finish-args` and `add-extensions` from the [Steam Flatpak](https://github.com/flathub/com.valvesoftware.Steam/blob/beta/com.valvesoftware.Steam.yml) instead of inheriting them automatically as a Flatpak extension. This is to ensure that Windows executables launched with Protontricks behave in a similar way to executables added inside the Steam client and launched there.
* ID is renamed from `com.valvesoftware.Steam.Utility.protontricks` to `com.github.Matoking.protontricks`
* Access is granted to Steam installation directory since we need to detect Steam apps and use Proton binaries.
* Desktop entry files are installed

The old com.valvesoftware.Steam.Utility.protontricks application will be EOL'd and updated to point toward this application. Flatpak can suggest the user to replace the application with the new one when this happens, although [it seems to be unclear](https://github.com/flathub/com.valvesoftware.Steam.Utility.protontricks/issues/30#issuecomment-899714721) whether this has been tested when the application type is also changed from an extension to an application.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+.flatpak-builder
+/protontricks
+build
--- a/README.md
+++ b/README.md
@@ -1,0 +1,46 @@
+Protontricks for Flatpak
+========================
+
+[Protontricks](https://github.com/Matoking/protontricks) compatible with the Flatpak version of Steam.
+
+# Installation
+
+You can install Protontricks for Flatpak from the Flathub repository. You need to have Flatpak installed with the Flathub repository configured. For installing Flatpak and configuring Flathub, see the [official installation instructions](https://flatpak.org/setup/).
+
+Install Protontricks for Flatpak using the following command:
+
+```sh
+flatpak install flathub com.github.Matoking.protontricks
+```
+
+Add an alias that allows you to call Protontricks using the `protontricks-flat` alias:
+
+```sh
+echo "alias protontricks-flat='flatpak run com.github.Matoking.protontricks'" >> ~/.bashrc
+```
+
+**You will need to restart any terminal emulators you have open for the alias to take effect.**
+
+# Usage
+
+## Command-line
+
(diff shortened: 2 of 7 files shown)


gasinvein:
I think you shouldn't copy _all_ of the Steam flatpak finish-apps. Large portion of them are overrides for native games (mostly the `--env=` ones), others won't do anything (e.g. related to `/app/utils` directory), and `--persist=.` is a very special one known to confuse users and developers alike.
Wine (and Proton), IIRC, mostly needs only `--allow=devel`, `--allow=multiarch`, `--device=all` and `--system-talk-name=org.freedesktop.UDisks2`.

gasinvein:
I've just realized that .exe's is launched via this app likely won't be able to communicate with the Steam client due to separate pid namespace. How much of a problem this would be?

barthalion:
LGTM, but I'd rather delegate final approval to @nanonyme and @gasinvein.

gasinvein:
I don't see any significant problems with the manifest. Some more `finish-args` are probably useless and can be dropped, but it's not a big deal.

Matoking:
bot, build com.github.Matoking.protontricks

PR #1298, opened 2019-12-16
add gthumb photo manager

muelli:
This is a simple package for https://gitlab.gnome.org/GNOME/gthumb/

I guess the upstream appdata file could need some love. Let's see what the bot says.

PR #2500, opened 2021-09-11
Add org.gnome.gThumb - !1298 follow-up

fedelibre:
This is a follow-up of !1298.

It's still a work in progress. I'll try to address the issues listed by @paobac [here](https://github.com/flathub/flathub/pull/1298#issuecomment-568210978). Let's make a checklist of it and mark those solved:

- [x] System bookmarks not shown in the bookmarks menu, they are read from the file XDG_CONFIG_HOME/gtk-3.0/bookmarks
- [x] Mount points not shown
- [ ] <s>The 'Open with...' menu doesn't show the applications associated with the file type</s> - cannot be solved?
- [ ] <s>Pressing 'g' should open the selected image with Gimp</s> - cannot be solved?
- [x] Image thumbnails should be read from and saved into ~/.cache/thumbnails
- [x] No video thumbnails, generated executing the scripts in /usr/share/thumbnailers/
- [x] No audio/video playback

Let's see if the build completes successfully at this stage, then I'll add new commits later.


fedelibre:
bot, build org.gnome.gThumb

org.gnome.gThumb.yml
```diff
@@ -0,0 +1,81 @@
...
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DEXIV2_BUILD_EXIV2_COMMAND=OFF
+      - -DEXIV2_BUILD_SAMPLES=OFF
+    sources:
+      - type: archive
+        url: https://exiv2.org/builds/exiv2-0.27.4-Source.tar.gz
+        sha256: 84366dba7c162af9a7603bcd6c16f40fe0e9af294ba2fd2f66ffffb9fbec904e
+
+  - name: lcms2
```
hfiguiere, line 32: lcms2 2.10 is in the runtime / SDK

hfiguiere:
for the mount point you should allow gvfs. It needs some dbus and filesystem permissions

org.gnome.gThumb.yml
```diff
@@ -0,0 +1,81 @@
...
+      - --disable-examples
+    build-commands:
+      - autoreconf --install --force
+      - ./configure --prefix=/app
+      - make
+      - make install
+    sources:
+      - type: archive
+        url: https://www.libraw.org/data/LibRaw-0.20.2.tar.gz
+        sha256: dc1b486c2003435733043e4e05273477326e51c3ea554c6864a4eafaff1004a6
```
hfiguiere, line 69:
```suggestion
  - name: libraw
    config-opts:
      - --disable-examples
    sources:
      - type: archive
        url: https://www.libraw.org/data/LibRaw-0.20.2.tar.gz
        sha256: dc1b486c2003435733043e4e05273477326e51c3ea554c6864a4eafaff1004a6
      - type: shell
        commands:
        - autoreconf -fi
```

With this it works. I'm not sure what the problem was your were having, and why `--force --install` didn't seem to do it.

org.gnome.gThumb.yml
```diff
@@ -0,0 +1,81 @@
...
+
+finish-args:
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --share=ipc
+  - --talk-name=org.freedesktop.secrets
+  - --filesystem=xdg-public-share
+  - --filesystem=xdg-pictures
+
```
hfiguiere, line 16:
```suggestion
  - --filesystem=xdg-run/gvfs
  - --filesystem=xdg-run/gvfsd
  - --talk-name=org.gtk.vfs.*
```

This will provide gvfs access and display the mounts.

hfiguiere (changes requested):

org.gnome.gThumb.yml
fedelibre: You are right, it works

org.gnome.gThumb.yml
fedelibre: Yes, I did try it before opening the PR. It lets you see the mounts but you cannot open it. Did you try it?

org.gnome.Photos has the same problem.

org.gnome.gThumb.yml
fedelibre: Thanks, I should have checked:

```
[📦 org.gnome.Sdk ~]$ pkg-config --list-all | grep lcms2
lcms2                          lcms2 - LCMS Color Management Library
```

fedelibre:
bot, build org.gnome.gThumb

fedelibre:
>  No video thumbnails, generated executing the scripts in /usr/share/thumbnailers/

I'm trying to fix this, but I need some help.
These scripts and the binary needed to run them (/usr/bin/gdk-pixbuf-thumbnailer) are under /usr in the GNOME SDK.
I guess there's some security rule which prevents the execution of the script?
How to work around it?

fedelibre:
>     * System bookmarks not shown in the bookmarks menu, they are read from the file XDG_CONFIG_HOME/gtk-3.0/bookmarks

I confirm the problem and I see that the flatpak has its own bookmarks, which seem to be saved in this XML file: `~/.var/app/org.gnome.gThumb/config/gthumb/bookmarks.xbel`.

I don't know what to do. Any advice is welcome.

fedelibre:
It doesn't seem a matter of permissions. I mean, this edit doesn't seem to help:

```
$ git diff
diff --git a/org.gnome.gThumb.yml b/org.gnome.gThumb.yml
index 27039ca..26fc22a 100644
--- a/org.gnome.gThumb.yml
+++ b/org.gnome.gThumb.yml
@@ -12,6 +12,8 @@ finish-args:
   - --device=dri
   - --share=ipc
   - --talk-name=org.freedesktop.secrets
+  - --filesystem=xdg-cache
+  - --filesystem=xdg-config
   - --filesystem=xdg-public-share
   - --filesystem=xdg-pictures
   - --filesystem=xdg-run/gvfs
```

fedelibre:
I take back my last comment. I've just added a new image in my ~/Pictures folder and I can see the thumbnail in `~/.cache/thumbnails/large`. So xdg-cache permission seems to work.

I added xdg-config permission, but still the app is not reading from `~/.config/gtk-3.0/bookmarks`.

fedelibre:
>     * [ ]  The 'Open with...' menu doesn't show the applications associated with the file type
> 
>     * [ ]  Pressing 'g' should open the selected image with Gimp
> 

These two problems are related, but I don't think there's any good solution. See also [this open issue on xdg-desktop-portal](https://github.com/flatpak/xdg-desktop-portal/issues/283).

fedelibre:
bot, build org.gnome.gThumb

fedelibre:
> > No video thumbnails, generated executing the scripts in /usr/share/thumbnailers/
> 
> I'm trying to fix this, but I need some help.
> These scripts and the binary needed to run them (/usr/bin/gdk-pixbuf-thumbnailer) are under /usr in the GNOME SDK.
> I guess there's some security rule which prevents the execution of the script?
> How to work around it?

I guess I should have said that these files are present in GNOME **platform**:

```
$ flatpak run --command=sh org.gnome.Platform//40
[📦 org.gnome.Platform ~]$ which gdk-pixbuf-thumbnailer
/usr/bin/gdk-pixbuf-thumbnailer
[📦 org.gnome.Platform ~]$ ls /usr/share/thumbnailers/
gdk-pixbuf-thumbnailer.thumbnailer  librsvg.thumbnailer                
[📦 org.gnome.Platform ~]$ cat /usr/share/thumbnailers/gdk-pixbuf-thumbnailer.thumbnailer 
[Thumbnailer Entry]
TryExec=/usr/bin/gdk-pixbuf-thumbnailer
Exec=/usr/bin/gdk-pixbuf-thumbnailer -s %s %u %o
MimeType=image/png;image/jpeg;image/bmp;image/x-bmp;image/x-MS-bmp;image/gif;image/x-icon;image/x-ico;image/x-win-bitmap;image/vnd.microsoft.icon;application/ico;image/ico;image/icon;text/ico;application/x-navi-animation;image/x-portable-anymap;image/x-portable-bitmap;image/x-portable-graymap;image/x-portable-pixmap;image/tiff;image/x-xpixmap;image/x-xbitmap;image/x-tga;image/x-icns;image/x-quicktime;image/qtif;
```

fedelibre:
> I added xdg-config permission, but still the app is not reading from `~/.config/gtk-3.0/bookmarks`

I should have used `xdg-config/gtk-3.0`. I'll force-push the last commit (amended).

barthalion:
> No video thumbnails, generated executing the scripts in /usr/share/thumbnailers/

I assume this needs totem-video-thumbnailer to work?

fedelibre:
All the actionable issues listed by upstream author @paobac have been fixed.

This PR is now ready for the final review.
I'd like to contribute a patch upstream to handle the update of release tags automatically and avoid the patch file. But I wouldn't hold the approval of this PR until it's ready.

fedelibre:
bot, build org.gnome.gThumb

org.gnome.gThumb.yml
```diff
@@ -0,0 +1,102 @@
...
+finish-args:
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --socket=wayland
+  - --device=dri
+  - --share=ipc
+  - --talk-name=org.freedesktop.secrets
+  - --filesystem=xdg-cache
+  - --filesystem=xdg-config/gtk-3.0
```
barthalion, line 16: Excuse my laziness (I guess it was explained already in the old PR?), but what these two are needed for?

PR #2509, opened 2021-09-16
Create io.github.prateekmedia.appimagepool

prateekmedia:


prateekmedia:
Taking the discussion from https://github.com/flathub/flathub/pull/2492, I either need to show user a warning regarding container limitation to execution of program or I will develop the app a special `--devel` permission to get outside container enviroment and execute the program.

But I proposed a third way which dolphin uses to execute programs even in flatpak. Now I don't know how dolphin does that but here is its [flatpak manifest](https://github.com/flathub/org.kde.dolphin/blob/master/org.kde.dolphin.json),

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

hfiguiere:
`cp: target '/usr/lib/sdk/flutter/' is not a directory`

`/usr` is read-only. So some path should be set.

The prefix is `/app`.

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,69 @@
...
+modules:
+  - name: flutter
+    buildsystem: simple
+    sources:
+      - type: archive
+        url: https://storage.googleapis.com/flutter_infra_release/releases/beta/linux/flutter_linux_2.6.0-5.2.pre-beta.tar.xz
+        sha256: 1a2d82b9a7934d00d680b6a0a9e09697264dd86ba37d984812c73297f33458e2
+    build-commands:
+      - mkdir -p /usr/lib/sdk/flutter/
+      - cp -rpv * /usr/lib/sdk/flutter/
```
hfiguiere, line 36: `/usr` is read-only. You can't copy anything there.

use `/app`

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,69 @@
...
+  - name: scripts
+    sources:
+      - type: script
+        commands:
+          - export FLUTTER_SDK=/usr/lib/sdk/flutter
+          - export PATH=$PATH:$FLUTTER_SDK/bin
+        dest-filename: enable.sh
+    buildsystem: simple
+    build-commands:
+      - cp enable.sh /usr/lib/sdk/flutter/
```
hfiguiere, line 47: same as above.

hfiguiere (changes requested):

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

io.github.prateekmedia.appimagepool.yml
prateekmedia: done

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
@hfiguiere I have used golang extension from freedesktop as a base for flutter, but I don't know how to set environment variables for flutter, otherwise it's reporting command not found.

hfiguiere:
sdk extensions are a bit of a different case, and this is not the approach to take here. It's actually simplier.

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,60 @@
...
+  - --socket=wayland
+  - --socket=pulseaudio
+  - --share=network
+  - --device=dri
+  - --persist=Applications
+  - --talk-name=org.freedesktop.Notifications
+build-option:
+  - append-path: /usr/lib/sdk/rust-stable/bin
+    build-args:
+      - --share=network
```
hfiguiere, line 24: this won't work. there is no network access during the build phase.

you need all the crates downloaded first. There is a tool in https://github.com/flatpak/flatpak-builder-tools to generate a manifest to include.

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,60 @@
...
+        sha256: 1a2d82b9a7934d00d680b6a0a9e09697264dd86ba37d984812c73297f33458e2
+    build-commands:
+      - install -d /app/lib/sdk/flutter
+      - cp -rpv * /app/lib/sdk/flutter/
+
+  - name: appimagepool
+    buildsystem: simple
+    build-commands:
+      - export FLUTTER_SDK=/app/lib/sdk/flutter
+      - export PATH=$PATH:$FLUTTER_SDK/bin
```
hfiguiere, line 42: You should just had a prepend-path: /app/lib/sdk/flutter/bin in the build-options (or add it to the `append-path`)

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,60 @@
...
+        url: https://storage.googleapis.com/flutter_infra_release/releases/beta/linux/flutter_linux_2.6.0-5.2.pre-beta.tar.xz
+        sha256: 1a2d82b9a7934d00d680b6a0a9e09697264dd86ba37d984812c73297f33458e2
+    build-commands:
+      - install -d /app/lib/sdk/flutter
+      - cp -rpv * /app/lib/sdk/flutter/
+
+  - name: appimagepool
+    buildsystem: simple
+    build-commands:
+      - export FLUTTER_SDK=/app/lib/sdk/flutter
```
hfiguiere, line 41: you should instead put this in the `env` section of the `build-options` at the top.

hfiguiere (changes requested):

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,60 @@
...
+    build-args:
+      - --share=network
+    env:
+      CARGO_HOME: /run/build/rust-flatpak/cargo
+modules:
+  - name: flutter
+    buildsystem: simple
+    sources:
+      - type: archive
+        url: https://storage.googleapis.com/flutter_infra_release/releases/beta/linux/flutter_linux_2.6.0-5.2.pre-beta.tar.xz
```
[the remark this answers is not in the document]
prateekmedia: yes.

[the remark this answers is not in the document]
hfiguiere: because
```
% file bin/cache/dart-sdk/bin/dart
bin/cache/dart-sdk/bin/dart: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, stripped
```

doesn't look like it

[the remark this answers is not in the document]
prateekmedia: you are right, let me look into this.

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,58 @@
...
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --socket=pulseaudio
+  - --share=network
+  - --device=dri
+  - --persist=Applications
+  - --talk-name=org.freedesktop.Notifications
+build-option:
+  prepend-path: /app/lib/sdk/flutter/bin
+  append-path: /usr/lib/sdk/rust-stable/bin
```
hfiguiere, line 23:
```suggestion
  append-path: /usr/lib/sdk/rust-stable/bin:/app/lib/sdk/flutter/bin
```

this ought to be enough

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,58 @@
...
+  env:
+    CARGO_HOME: /run/build/popsicle/cargo
```
hfiguiere, line 25:
```suggestion
    CARGO_HOME: /run/build/appimagepool/cargo
```

cut&paste mistake

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,58 @@
...
+        sha256: 1a2d82b9a7934d00d680b6a0a9e09697264dd86ba37d984812c73297f33458e2
+    build-commands:
+      - install -d /app/lib/sdk/flutter
+      - cp -rpv * /app/lib/sdk/flutter/
+
+  - name: appimagepool
+    buildsystem: simple
+    build-commands:
+      - flutter config --enable-linux-desktop
+      - cargo build --release
```
hfiguiere, line 41: the readme in the cargo part of the flatpak-builder-tools will tell you what to do here. This isn't sufficient. (but it hasn't been reached yet)

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,58 @@
...
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --socket=pulseaudio
+  - --share=network
+  - --device=dri
+  - --persist=Applications
+  - --talk-name=org.freedesktop.Notifications
+build-option:
```
hfiguiere, line 21:
```suggestion
build-options:
```

This is why it's failing. flatpak-builder gives you a warning.

hfiguiere (changes requested):

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

io.github.prateekmedia.appimagepool.yml
prateekmedia: done

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,57 @@
...
+      - install -Dm644  assets/appimagepool.svg /app/share/icons/hicolor/128x128/apps/appimagepool.svg
+      - install -Dm644  assets/appimagepool.desktop /app/share/applications/appimagepool.desktop
+      - install -Dm644  assets/appimagepool.appdata.xml /app/share/metainfo/appimagepool.appdata.xml
+      - sed -i 's/<id>appimagepool/<id>io.github.prateekmedia.appimagepool/' /app/share/metainfo/appimagepool.appdata.xml
+      - mkdir -p /app/bin/ /app/lib/
+      - ln -s /usr/lib/${FLATPAK_ARCH}-linux-gnu/libgio-2.0.so.0 /app/lib/libgio-2.0.so
+      - ln -s /app/appimagepool/appimagepool /app/bin/appimagepool
+    sources:
+      - type: git
+        dest: appimagepool
```
hfiguiere, line 53:
```suggestion
```

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,57 @@
...
+        url: https://storage.googleapis.com/flutter_infra_release/releases/beta/linux/flutter_linux_2.6.0-5.2.pre-beta.tar.xz
+        sha256: 1a2d82b9a7934d00d680b6a0a9e09697264dd86ba37d984812c73297f33458e2
+    build-commands:
+      - install -d /app/lib/sdk/flutter
+      - cp -rpv * /app/lib/sdk/flutter/
+
+  - name: appimagepool
+    buildsystem: simple
+    build-commands:
+      - flutter config --enable-linux-desktop
```
hfiguiere, line 39:
```suggestion
      - cp -rpv * /app/lib/sdk/flutter/
      - flutter config --enable-linux-desktop

  - name: appimagepool
    buildsystem: simple
    build-commands:
```

You must to the flutter config in the flutter module, because after that the filesystem is read only.

hfiguiere (changes requested):
but it will still fail after building most of the rust code, not sure what can be tweaked, flutter seems to think it's a good idea to write in random location on the filtesystem.

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,56 @@
...
+modules:
+  - name: flutter
+    buildsystem: simple
+    sources:
+      - type: archive
+        url: https://storage.googleapis.com/flutter_infra_release/releases/beta/linux/flutter_linux_2.6.0-5.2.pre-beta.tar.xz
+        sha256: 1a2d82b9a7934d00d680b6a0a9e09697264dd86ba37d984812c73297f33458e2
+    build-commands:
+      - install -d /app/lib/sdk/flutter
+      - cp -rpv * /app/lib/sdk/flutter/
```
hfiguiere, line 34:
```suggestion
      - cp -rpv * .git /app/lib/sdk/flutter/
```

also it's missing this, I forgot. For some reason flutter want the `.git` and `*` doesn't catch it.

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

io.github.prateekmedia.appimagepool.yml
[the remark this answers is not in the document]
hfiguiere:
```
/app/lib/sdk/flutter/bin/internal/shared.sh: line 227: /app/lib/sdk/flutter/bin/cache/dart-sdk/bin/dart: cannot execute binary file: Exec format error
```

as I said, it doesn't work on `aarch64`.

hfiguiere:
the build failure is caused by flutter wanting to write `/usr/lib/sdk/flutter/version` despite already being there. it can't.

one solution I see is to merge the flutter and the appimagepool

prateekmedia:
How can they be merged? In the flatpak manifest or the upstream.

hfiguiere:
in the manifest.

hfiguiere:
it's about how flatpak works: you can write on top of an existing module.

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,56 @@
...
+  - --share=network
+  - --device=dri
+  - --persist=Applications
+  - --talk-name=org.freedesktop.Notifications
+build-options:
+  append-path: /usr/lib/sdk/rust-stable/bin:/app/lib/sdk/flutter/bin
+  env:
+    CARGO_HOME: /run/build/appimagepool/cargo
+modules:
+  - name: flutter
```
barthalion, line 26: I don't know anything about Flutter, but if the resulting binary is fully static, you should add `cleanup: ["*"]`.

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,56 @@
...
+command: appimagepool
+separate-locales: false
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --socket=pulseaudio
+  - --share=network
+  - --device=dri
+  - --persist=Applications
```
barthalion, line 19: Given how appimages are traditionally used, you probably want to use `--filesystem=~/Applications:create` here, or move it completely to `$XDG_DATA_HOME` (applications have write access here by default). Probably the `--filesystem` makes more sense to make it easier for users to find and execute download applications outside the sandbox.

barthalion:
> Taking the discussion from #2492, I either need to show user a warning regarding container limitation to execution of program or I will develop the app with a special `--devel` permission to get outside container environment and execute the program.
> 
> But I proposed a third way which dolphin uses to execute programs even in flatpak. Now I don't know how dolphin does that but here is its [flatpak manifest](https://github.com/flathub/org.kde.dolphin/blob/master/org.kde.dolphin.json). Any help regarding this would be appreciated.

This is not ideal for Dolphin either, and it will be limited the same way your app would be. If a user tries to execute "system-level" like gparted, or an app requiring access to `/dev`, it will simply fail.

That's why you need to either use `--devel`, or `--talk-name=org.freedesktop.Flatpak` and change your app to execute appimages with `flatpak-spawn --host` – which, of course, throws any sandboxing away but there's no other way to make it work properly.

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
@barthalion I have tried with `--devel` and the other way but it looks like it is still throwing the same exception
```bash
**$ cd /home/prateeksu/Applications
$ ./Eagle.AppImage
ProcessException: No such file or directory
  Command: ./Eagle.AppImage 
$ ./Eagle.AppImage
workingDirectory: /home/prateeksu/Applications

  Check that ./Eagle.AppImage exists
    command: ./Eagle.AppImage

[ERROR:flutter/lib/ui/ui_dart_state.cc(209)] Unhandled Exception: ShellException(./Eagle.AppImage, error: ProcessException: No such file or directory
  Command: ./Eagle.AppImage , workingDirectory: /home/prateeksu/Applications)
**
```

But I can use it normally.

barthalion:
I'm fairly sure the problem is you need to `chmod +x` the file before trying to execute it. But it still means it will be executed inside the sandbox, so the command should be `flatpak-spawn --host ~/Applications/filename` after you add `--talk-name` I mentioned above.

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

io.github.prateekmedia.appimagepool.yml
```diff
@@ -14,6 +14,7 @@ finish-args:
   - --socket=pulseaudio
   - --share=network
   - --device=dri
+  - --allow=devel
```
barthalion, line 17: This shouldn't be needed. The app should use `flatpak-spawn --host` to execute binaries when running as Flatpak.

barthalion:
Current build fails to start for me:

```
> flatpak run --command=bash io.github.prateekmedia.appimagepool
[📦 io.github.prateekmedia.appimagepool ~]$ appimagepool 
Gtk-Message: 09:11:49.054: Failed to load module "canberra-gtk-module"
Gtk-Message: 09:11:49.055: Failed to load module "canberra-gtk-module"
Segmentation fault (core dumped)
```

prateekmedia:
@barthalion do you mean to use `flatpak-spawn --host ~/Applications/run-my.AppImage` command instead of regular `~/Applications/run-my.AppImage` in the app?

barthalion:
Yeah, exactly. It means AppImage is executed directly on the host, without sandboxing – it can't work reliably any other way.

nedrichards:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
@nedrichards @barthalion Can you re-review it, I think everything even downloaded app launching is working due to `flatpak-spawn --host $appname`.

barthalion:
Does it start for you? It still produces a segmentation fault on my laptop.

prateekmedia:
@barthalion You are right, looks like opening from command line doesn't work, can you try to open from your launcher.

prateekmedia:
@barthalion Aha, looks like owning name as `appimagepool` fixes the issue.
```bash
flatpak run io.github.prateekmedia.appimagepool --own-name=appimagepool
```

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,42 @@
...
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --socket=pulseaudio
+  - --share=network
+  - --device=dri
+  - --filesystem=~/Applications
+  - --talk-name=org.freedesktop.Flatpak
+  - --talk-name=org.freedesktop.Notifications
+  - --own-name=appimagepool
```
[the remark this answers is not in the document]
prateekmedia: Thanks, changed it to com.prateekmedia.appimagepool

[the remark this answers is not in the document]
hfiguiere: it should match the app-id.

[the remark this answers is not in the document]
prateekmedia: @hfiguiere Ok I will change it to the app id.

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,42 @@
...
+  - --own-name=io.github.prateekmedia.appimagepool
+modules:
+  - name: appimagepool
+    buildsystem: simple
+    build-commands:
+      - cp -R appimagepool/usr/bin/ /app/appimagepool
+      - cp -R appimagepool/usr/share/ /app/share
+      - sed -i 's/<id>appimagepool/<id>io.github.prateekmedia.appimagepool/' /app/share/metainfo/appimagepool.appdata.xml
+      - mkdir -p /app/bin/ /app/lib/
+      - ln -s /usr/lib/${FLATPAK_ARCH}-linux-gnu/libgio-2.0.so.0 /app/lib/libgio-2.0.so
```
barthalion, line 29: Is it actually needed?

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,42 @@
...
+command: appimagepool
+separate-locales: false
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --socket=pulseaudio
+  - --share=network
+  - --device=dri
+  - --filesystem=~/Applications
```
barthalion, line 17: From my limited testing with heavily lifted sandbox, it seems to be trying to save files in `XDG_CONFIG_HOME` but misses a slash when concatenating the path and filename, so it ends up as `/var/home/bpiotrowski/.var/app/io.github.prateekmedia.appimagepool/confignvim.appimage` on my system.

I thought the idea was to use `~/Applications` so it's easily discoverable outside flatpak environment?

io.github.prateekmedia.appimagepool.yml
```diff
@@ -0,0 +1,42 @@
...
+  - --talk-name=org.freedesktop.Flatpak
+  - --talk-name=org.freedesktop.Notifications
+  - --own-name=io.github.prateekmedia.appimagepool
```
barthalion, line 20: This is wrong, your application `--own-name=ID` by default. There's no need to specify it.

barthalion (commented):
Still segfaults:

```
           PID: 314048 (appimagepool)
           UID: 1000 (bpiotrowski)
           GID: 1000 (bpiotrowski)
        Signal: 11 (SEGV)
     Timestamp: Tue 2021-09-28 17:44:36 CEST (5min ago)
  Command Line: ./appimagepool
    Executable: /app/appimagepool/appimagepool
 Control Group: /user.slice/user-1000.slice/user@1000.service/app.slice/app-flatpak-io.github.prateekmedia.appimagepool-314031.scope
          Unit: user@1000.service
     User Unit: app-flatpak-io.github.prateekmedia.appimagepool-314031.scope
         Slice: user-1000.slice
     Owner UID: 1000 (bpiotrowski)
       Boot ID: c165e0ee4b514edabe614f951b507ad2
    Machine ID: 7ba96b996cba407a881e678c260fb38c
      Hostname: risuku
       Storage: /var/lib/systemd/coredump/core.appimagepool.1000.c165e0ee4b514edabe614f951b507ad2.314048.1632843876000000.zst (present)
     Disk Size: 4.5M
       Message: Process 314048 (appimagepool) of user 1000 dumped core.
                
                Stack trace of thread 4:
                #0  0x00007f484eb4410a n/a (/usr/lib/x86_64-linux-gnu/libgdk-3.so.0.2404.26 + 0x7a10a)
```

I don't know why you think it's D-Bus related. It segfaults the same way for me if I pass `--socket=system-bus` and `--socket=session-bus`.

io.github.prateekmedia.appimagepool.yml
prateekmedia: Yes without it app fails to start, complaining libgio not found

io.github.prateekmedia.appimagepool.yml
prateekmedia: Yes that was the idea, but I don't know why it's saving there.

io.github.prateekmedia.appimagepool.yml
prateekmedia: This was the build https://github.com/flathub/flathub/pull/2492#issuecomment-913108032 that fails to load libgio

io.github.prateekmedia.appimagepool.yml
prateekmedia: Got it, will fix this immediately.

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

io.github.prateekmedia.appimagepool.yml
prateekmedia: Ok removed it eventually.

barthalion:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
actually mkdir /app/bin was also needed, also in the next version I will try to remove the libgio dependency as it is due to an old package.

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

barthalion:
Err, yeah, I removed it by mistake. I guess I still can't use vim after all these years.

barthalion:
bot, build io.github.prateekmedia.appimagepool

prateekmedia:
bot, build io.github.prateekmedia.appimagepool

PR #2516, opened 2021-09-19
com.github.alexkdeveloper.desktop-files-creator 1.0.0

alexkdeveloper:

### DIFF
--- a/com.github.alexkdeveloper.desktop-files-creator.json
+++ b/com.github.alexkdeveloper.desktop-files-creator.json
@@ -9,7 +9,8 @@
         "--share=ipc",
         "--socket=fallback-x11",
         "--socket=wayland",
-        "--filesystem=home"
+        "--filesystem=home:ro",
+        "--filesystem=xdg-data/applications:create"
     ],
     "cleanup" : [
         "/include",
@@ -33,7 +34,7 @@
                     "type" : "git",
                     "url" : "https://github.com/alexkdeveloper/desktop-files-creator.git",
                     "tag": "1.0.0",
-                    "commit": "154e94a3856a307a68a9d5b2297e368f698d157e"
+                    "commit": "5aa178f9109da9c334a42383243fe77f8f0cf3e9"
                 }
             ]
         }
--- a/com.github.alexkdeveloper.desktop-files-creator.json
+++ b/com.github.alexkdeveloper.desktop-files-creator.json
@@ -1,0 +1,41 @@
+{
+    "app-id" : "com.github.alexkdeveloper.desktop-files-creator",
+    "runtime" : "org.gnome.Platform",
+    "runtime-version" : "40",
+    "sdk" : "org.gnome.Sdk",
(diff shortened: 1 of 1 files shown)


com.github.alexkdeveloper.desktop-files-creator.json
```diff
@@ -0,0 +1,41 @@
...
+    "runtime" : "org.gnome.Platform",
+    "runtime-version" : "40",
+    "sdk" : "org.gnome.Sdk",
+    "command" : "desktop-files-creator",
+    "finish-args" : [
+        "--share=network",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--filesystem=home"
```
gasinvein, line 12: Why does the app need access to the whole home directory?
If it needs to write files only to `~/.local/share/applications`, add `--filesystem=xdg-data/applications` instead.
If it always use file chooser dialog to open/save files, then it probably shouldn't need static filesystem permissions at all.

alexkdeveloper: Hi! I just tried to configure the permissions as you showed (--filesystem=xdg-data/applications) and the program does not see the files. Chooser is empty. The user will not be able to specify the files.

gasinvein: If the chooser is empty, given it's a GTK3 app, my guess is that you're using `FileChooserDialog` instead of `FileChooserNative`. The former doesn't support portals. Unless you add custom widgets or otherwise extend the dialogs, consider switching your app to use `FileChooserNative`.

alexkdeveloper: I use image preview in chooser. It was not possible to add it in Native Chooser. Is it possible to add the following options :ro to existing permissions? That is, to allocate the application directory for writing, and everything else is read-only?

gasinvein: Sure, you can replace `home` with `home:ro` + `xdg-data/application`

alexkdeveloper: It's done! Updated the file.

com.github.alexkdeveloper.desktop-files-creator.json
```diff
@@ -0,0 +1,42 @@
...
+    ],
+    "modules" : [
+        {
+            "name" : "Desktop Files Creator",
+            "builddir" : true,
+            "buildsystem" : "meson",
+            "sources" : [
+                {
+                    "type" : "git",
+                    "url" : "https://github.com/alexkdeveloper/desktop-files-creator.git",
```
barthalion, line 35: You should consider configuring https://github.com/flathub/flatpak-external-data-checker/ to automate updates but it's not a blocker.

barthalion:
bot, build com.github.alexkdeveloper.desktop-files-creator

PR #2605, opened 2021-10-29
Add org.gnome.NautilusPreviewer (aka Sushi)

fedelibre:

- [X] I contacted upstream developers about submitting their software to Flathub: https://gitlab.gnome.org/GNOME/sushi/-/issues/27

### DIFF
--- a/org.gnome.NautilusPreviewer.json
+++ b/org.gnome.NautilusPreviewer.json
@@ -1,0 +1,117 @@
+{
+    "app-id": "org.gnome.NautilusPreviewer",
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "41",
+    "sdk": "org.gnome.Sdk",
+    "build-options": {
+        "cflags": "-O2 -g",
+        "cxxflags": "-O2 -g",
+        "env": {
+            "V": "1"
+        }
+    },
+    "command": "sushi",
+    "cleanup": [
+        "/include",
+        "/lib/pkgconfig",
+        "/share/gtk-doc",
+        "/share/man",
+        "/share/pkgconfig",
+        "*.a"
+    ],
+    "finish-args": [
+        "--socket=pulseaudio",
+        "--socket=x11",
+        "--socket=wayland",
+        "--share=ipc",
+        "--share=network",
+        "--device=dri",
+        "--talk-name=org.freedesktop.FileManager1",
+        "--filesystem=xdg-run/dconf",
+        "--filesystem=~/.config/dconf:ro",
+        "--filesystem=host:ro",
+        "--filesystem=home",
+        "--env=DCONF_USER_CONFIG_DIR=.config/dconf"
+    ],
+    "modules": [
+        {
+            "name": "popplerdata",
+            "no-autogen": true,
+            "make-install-args": ["prefix=/app"],
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://poppler.freedesktop.org/poppler-data-0.4.11.tar.gz",
+                    "sha256": "2cec05cd1bb03af98a8b06a1e22f6e6e1a65b1e2f3816cb3069bb0874825f08c"
+                }
+            ]
+        },
+        {
+            "name": "poppler",
+            "buildsystem": "cmake-ninja",
+            "config-opts": ["-DCMAKE_INSTALL_LIBDIR=/app/lib", "-DCMAKE_INSTALL_INCLUDEDIR=/app/include",
+                            "-DENABLE_LIBOPENJPEG=none", "-DENABLE_BOOST=OFF"],
+            "cleanup": [ "/bin" ],
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://poppler.freedesktop.org/poppler-21.10.0.tar.xz",
+                    "sha256": "964b5b16290fbec3fae57c2a5bcdea49bb0736bd750c3a3711c47995c9efc394"
+                }
+            ]
+        },
+        {
+            "name": "evince",
+            "buildsystem": "meson",
+            "cleanup": [ "/share/GConf", "/share/help" ],
+            "config-opts": [ "-Dnautilus=false", "-Dviewer=false",
+                             "-Dpreviewer=false", "-Ddbus=false",
+                             "-Dintrospection=true", "-Dcomics=disabled",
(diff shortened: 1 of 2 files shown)


fedelibre:
bot, build org.gnome.NautilusPreviewer

fedelibre:
I've started from the manifest in the source, then I updated all the dependencies to the latest versions.
I wonder if neon27 is really needed. It's not mentioned at all in sushi's meson.build.

As I think Sushi is never supposed to edit a file, what about appending `:ro` to `--filesystem=home`? Or maybe just deleting that line and leaving `"--filesystem=host:ro"` would achieve the same?

fedelibre:
bot, build org.gnome.NautilusPreviewer

org.gnome.NautilusPreviewer.json
```diff
@@ -0,0 +1,117 @@
+{
+    "app-id": "org.gnome.NautilusPreviewer",
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "41",
+    "sdk": "org.gnome.Sdk",
+    "build-options": {
```
hfiguiere, line 6: you shouldn't need these.

org.gnome.NautilusPreviewer.json
```diff
@@ -0,0 +1,117 @@
...
+        "/include",
+        "/lib/pkgconfig",
+        "/share/gtk-doc",
+        "/share/man",
+        "/share/pkgconfig",
+        "*.a"
+    ],
+    "finish-args": [
+        "--socket=pulseaudio",
+        "--socket=x11",
```
hfiguiere, line 24:
```suggestion
        "--socket=fallback-x11",
```

org.gnome.NautilusPreviewer.json
```diff
@@ -0,0 +1,117 @@
...
+        "--socket=wayland",
+        "--share=ipc",
+        "--share=network",
+        "--device=dri",
+        "--talk-name=org.freedesktop.FileManager1",
+        "--filesystem=xdg-run/dconf",
+        "--filesystem=~/.config/dconf:ro",
+        "--filesystem=host:ro",
```
hfiguiere, line 32: does it really need host?

nedrichards: it can be invoked on any file that nautilus can see and has to work seamlessly without any kind of user confirmation so yes, it does.

hfiguiere: if it can work `:ro` when why isn't home `:ro` too? Does it write stuff in HOME that it doesn't elswhere?

fedelibre: I've removed the permission to home directory and it seems to work fine.

What never worked on my computer (Fedora 35 Silverblue, btrfs partition) is the preview on any file outside of the home directory. I don't think it's a sandbox issue, as I tried to grant all host permissions temporarily and it didn't help. In the log I see the error: `org.gnome.Nauti[50136]: JS ERROR: Failed to query default handler for file:///var/www/html/index.html: Gio.IOErrorEnum: Error when getting information for file “/var/www/html/index.html”: No such file or directory`.

fedelibre:
bot, build org.gnome.NautilusPreviewer

org.gnome.NautilusPreviewer.json
nedrichards: Basically, those files aren't mounted there when seen from inside the sandbox - they're available but at a different path. Sushi will need to be able to detect being inside a flatpak and then do some path translation for that case. Check the docs but basically you need that at /var/run/var etc. https://docs.flatpak.org/en/latest/sandbox-permissions.html?highlight=host#filesystem-access

hfiguiere: I don't think `/var` here is accessible from the sandbox.

org.gnome.NautilusPreviewer.json
```diff
@@ -0,0 +1,109 @@
...
+    "finish-args": [
+        "--socket=pulseaudio",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--share=ipc",
+        "--share=network",
+        "--device=dri",
+        "--talk-name=org.freedesktop.FileManager1",
+        "--filesystem=xdg-run/dconf",
+        "--filesystem=~/.config/dconf:ro",
```
barthalion, line 24: Are these and DCONF_USER_CONFIG_DIR needed?

felipeborges: No.

fedelibre:
> I've started from the manifest in the source, then I updated all the dependencies to the latest versions.
> I wonder if neon27 is really needed. It's not mentioned at all in sushi's meson.build.

It's not in [Fedora spec file](https://src.fedoraproject.org/rpms/sushi/blob/rawhide/f/sushi.spec) either.

@felipeborges I can safely remove neon27, right?

fedelibre:
Wait, neon is in the [Fedora flatpak manifest](https://src.fedoraproject.org/flatpaks/sushi/blob/stable/f/sushi.yaml#_38).
Maybe it's needed only for flatpak?

I'll use a recent tarball from [github repository](https://github.com/notroj/neon).

org.gnome.NautilusPreviewer.json
fedelibre: Ok, removed

fedelibre:
bot, build org.gnome.NautilusPreviewer

org.gnome.NautilusPreviewer.json
fedelibre: Can I leave this to the user?

I guess that most people will use Sushi to preview files in the home directory.
Those who need access to other directories can add a permission manually.

fedelibre:
Ok, I've removed neon and made little modifications based on your !27 MR.

fedelibre:
bot, build org.gnome.NautilusPreviewer

org.gnome.NautilusPreviewer.json
fedelibre: Eventually I decided to follow what Felipe did on [upstream MR 27](https://gitlab.gnome.org/GNOME/sushi/-/merge_requests/27) and use `--filesystem=home:ro` only.

fedelibre:
I'm happy with the current state of this PR.
If you don't have any more feedback, feel free to squash the commits and merge.
Thanks

PR #2610, opened 2021-10-31
Add WatchFlower flatpak manifest

emericg:

(permissions usage is exeplained in the manifest)

### DIFF
--- a/io.emeric.watchflower.yml
+++ b/io.emeric.watchflower.yml
@@ -13,7 +13,7 @@ finish-args:
   - --device=dri
   # X11 + XShm access
   - --share=ipc
-  - --socket=x11
+  - --socket=fallback-x11
   # Wayland access
   - --socket=wayland
   # We need Bluetooth support
--- a/io.emeric.watchflower.yml
+++ b/io.emeric.watchflower.yml
@@ -70,6 +70,7 @@ modules:
         url-template: https://www.kernel.org/pub/linux/bluetooth/bluez-$version.tar.xz
 
 - name: qtconnectivity
+  buildsystem: simple
   cleanup-platform:
     - /bin
     - /mkspecs
@@ -82,8 +83,9 @@ modules:
         type: anitya
         project-id: 153467
         tag-template: v$version
-  buildsystem: qmake
-  post-install:
+  build-commands:
+    - qmake
+    - make -j $FLATPAK_BUILDER_N_JOBS
     - cp -r -n {bin,include,lib} /app # we need to deploy libs at the right place
     - mkdir -p /app/src/bluetooth
     - cp -r src/bluetooth /app/src/ # header files too
--- a/io.emeric.watchflower.yml
+++ b/io.emeric.watchflower.yml
@@ -86,8 +86,8 @@ modules:
   build-commands:
     - qmake
     - make -j $FLATPAK_BUILDER_N_JOBS
-    - cp -r -n {bin,include,lib} /app # we need to deploy libs at the right place
     - mkdir -p /app/src/bluetooth
+    - cp -r -n {bin,include,lib} /app # we need to deploy libs at the right place
     - cp -r src/bluetooth /app/src/ # header files too
 
 - name: watchflower
@@ -96,6 +96,9 @@ modules:
     - type: archive
       url: https://github.com/emericg/WatchFlower/archive/refs/tags/v3.1.tar.gz
       sha256: c40a23ba7d962b10b2d35e65fe5f0b55f3b97f1773d4304bfb25df6f42a33cac
+    - type: file
+      path: watchflower.appdata.xml
   post-install:
     - mkdir -p /app/share/icons/hicolor/scalable/apps/
     - mv /app/share/pixmaps/watchflower.svg /app/share/icons/hicolor/scalable/apps/
+    - mv -f watchflower.appdata.xml /app/share/appdata/watchflower.appdata.xml
--- a/io.emeric.watchflower.yml
+++ b/io.emeric.watchflower.yml
@@ -1,12 +1,12 @@
 app-id: io.emeric.watchflower
 runtime: org.kde.Platform
-runtime-version: '5.15'
+runtime-version: '5.15-21.08'
 sdk: org.kde.Sdk
 command: watchflower
 
 rename-desktop-file: watchflower.desktop
 rename-appdata-file: watchflower.appdata.xml
-#rename-icon: watchflower
+rename-icon: watchflower
 
 finish-args:
   # Our UI is GPU accelerated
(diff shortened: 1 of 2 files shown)


io.emeric.watchflower.yml
```diff
@@ -0,0 +1,98 @@
...
+#rename-icon: watchflower
+
+finish-args:
+  # Our UI is GPU accelerated
+  - --device=dri
+  # X11 + XShm access
+  - --share=ipc
+  - --socket=x11
+  # Wayland access
+  - --socket=wayland
```
hfiguiere, line 18:
```suggestion
  # X11 + XShm access
  - --share=ipc
  - --socket=fallback-x11
  # Wayland access
  - --socket=wayland
```

hfiguiere:
bot, build io.emeric.watchflower

io.emeric.watchflower.yml
```diff
@@ -0,0 +1,98 @@
+app-id: io.emeric.watchflower
+runtime: org.kde.Platform
+runtime-version: '5.15'
```
hfiguiere, line 3:
```suggestion
runtime-version: '5.15-21.08'
```

io.emeric.watchflower.yml
```diff
@@ -0,0 +1,98 @@
...
+      x-checker-data:
+        type: anitya
+        project-id: 153467
+        tag-template: v$version
+  build-commands:
+    - qmake
+    - make -j $FLATPAK_BUILDER_N_JOBS
+    - cp -r -n {bin,include,lib} /app # we need to deploy libs at the right place
+    - mkdir -p /app/src/bluetooth
+    - cp -r src/bluetooth /app/src/ # header files too
```
hfiguiere, line 91: you could use `buildsystem: qmake` and add the rest into the `post-install` phase.

hfiguiere:
the build fails because the id in the appadata file doesn't match:
https://github.com/emericg/WatchFlower/blob/v3.0/assets/linux/watchflower.appdata.xml#L5

hfiguiere:
you can patch it here before there is a new release upstream. (but it should be fixed upstream too)

io.emeric.watchflower.yml
```diff
@@ -0,0 +1,98 @@
+app-id: io.emeric.watchflower
+runtime: org.kde.Platform
+runtime-version: '5.15'
+sdk: org.kde.Sdk
+command: watchflower
+
+rename-desktop-file: watchflower.desktop
+rename-appdata-file: watchflower.appdata.xml
+#rename-icon: watchflower
```
hfiguiere, line 9: I'm under the impression you'll need this, from a quick glance at the source.

io.emeric.watchflower.yml
emericg: Can I ask why? I just need Qt 5.15, the version of the KDE SDK doesn't matter at all for this application.

io.emeric.watchflower.yml
emericg: Ok so "rename-icon: watchflower.svg" ? The SVG file is next to the desktop and appdata file already.

emericg:
> the build fails because the id in the appdata file doesn't match: https://github.com/emericg/WatchFlower/blob/v3.0/assets/linux/watchflower.appdata.xml#L5

Yeah I though this might happen. I'm renaming this upstream and probably do a release anyway.
These App ID are just an awful idea...

Anyway thanks for your time reviewing this.

io.emeric.watchflower.yml
hfiguiere: this is just the versiobn based of the latest freedesktop-sdk. Policy: use the latest.

io.emeric.watchflower.yml
hfiguiere: just `rename-icon: watchflower`

emericg: thanks, I'll push a release tomorrow and update this pull request.

emericg:
Ok this should be better!

emericg:
I'm reverting the buildsystem: qmake change.
For instance it is (exactly) how subsurface is doing this. And it is building locally too.
https://github.com/flathub/org.subsurface_divelog.Subsurface/blob/b2d00ee5155911af773aa0f8de7a8ae9a0bf7785/org.subsurface_divelog.Subsurface.yaml#L130

hfiguiere:
bot, build io.emeric.watchflower

hfiguiere:
as long as it builds...

hfiguiere:
The icon is installed in ` /app/share/pixmaps/watchflower.svg` which is not where it is being looked in.

`/app/share/icons/hicolor/scalable/apps` is an appropriate destination (you can do that in the post-install)

emericg:
This should do the trick.
By the way, on my local install the runtime-version: '5.15-21.08' isn't recognized, I have version 5.15 of the org.kde.* runtimes, and I cannot make sense of where to get infos about these runtimes and the versions available.

hfiguiere:
by default flatpak builder doesn't install the needed sdk. it will just tell you they are missing.

emericg:
By the way I pushed the patch for the icon earlier.

hfiguiere:
bot, build io.emeric.watchflower

barthalion:
```

**Comment by emericg** [comment]
I included the patched appdata file here, that should do it until the next release.
These checks on the appdata could really be batched right? Instead of stopping on the first error, that could save some time (and CPU time too).

**Comment by emericg** [comment]
Any news?

**Comment by barthalion** [comment]
bot, build io.emeric.watchflower

FYI, you can trigger new builds yourself, too.

File: watchflower.appdata.xml [hunk 5]
```diff
@@ -0,0 +1,86 @@
+<?xml version="1.0" encoding="utf-8"?>
```
barthalion, line 1: FYI, ideally, you should maintain this file in your repository and ship it as the part of tarball.

barthalion:
LGTM. Can you confirm it properly works with sensors?

emericg:
> bot, build io.emeric.watchflower
> 
> FYI, you can trigger new builds yourself, too.

Sorry I didn't know that.

> LGTM. Can you confirm it properly works with sensors?

Yes! With the command from the buildbot (flatpak install --user https://dl.flathub.org/build-repo/64251/io.emeric.watchflower.flatpakref) we have a working build!

watchflower.appdata.xml
emericg: Yes everything is already in the repo:
https://github.com/emericg/WatchFlower/tree/master/assets/linux

I even build the flatpak file with a github action:
https://github.com/emericg/WatchFlower/actions/workflows/flatpak.yml

But because we need to build a tag or release, containing the flatpak files, it's always too late to fix the issue we discover on a tag or release...

PR #2615, opened 2021-11-03
Add Twine flatpak

SISheogorath:
This patch provides an initial draft of a Twine flatpak. It will provide
an offline app that can be used to create games and alike. It's
carefully crafted to contain Twine while not interfere with its regular
functionality.

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link: https://github.com/klembot/twinejs/issues/651**
- [ ] I own the domain used in the application ID or the domain has a policy for delegating subdomains (e.g. GitHub, SourceForge). --> Used the official upstream page domain


org.twinery.Twine.yaml
```diff
@@ -0,0 +1,69 @@
...
+      - install -Dm755 twine.sh /app/bin/twine-wrapper
+      - install -Dm644 org.twinery.Twine.desktop "${FLATPAK_DEST}/share/applications/org.twinery.Twine.desktop"
+      - install -Dm644 org.twinery.Twine.metainfo.xml "${FLATPAK_DEST}/share/metainfo/org.twinery.Twine.metainfo.xml"
+      - install -Dm644 org.twinery.Twine.svg "${FLATPAK_DEST}/share/icons/hicolor/scalable/apps/${FLATPAK_ID}.svg"
+    sources:
+      - type: file
+        only-arches:
+          - x86_64
+        dest-filename: twine.zip
+        url: "https://github.com/klembot/twinejs/releases/download/2.3.15/twine_2.3.15_linux64.zip"
```
hfiguiere, line 52: since this is x86_64 only (which make me sad) you need a flathub.json to limit architectures.

hfiguiere:
bot, build org.twinery.Twine

org.twinery.Twine.svg
```diff
@@ -0,0 +1,133 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
hfiguiere, line 1: why not using the svg in the repository?

hfiguiere (changes requested):
have the appdata and .desktop files been submitted upstream?

org.twinery.Twine.yaml
SISheogorath: This should be fixed now, I'll ask upstream to provide aarch64 builds :)

hfiguiere: or the flatpak could be built from source....

org.twinery.Twine.svg
SISheogorath: And it's removed. Good catch!

org.twinery.Twine.yaml
SISheogorath: Nope, not gonna happen from me as long as node-based applications are not self-contained (which might happens with yarn2 at some point). Maintaining npm dependencies and debugging broken builds just to come to the same results as upstream is none of my hobbys. 🤷🏻

org.twinery.Twine.yaml
```diff
@@ -0,0 +1,72 @@
...
+runtime: "org.freedesktop.Platform"
+runtime-version: "21.08"
+sdk: "org.freedesktop.Sdk"
+sdk-version: "21.08"
+command: "twine-wrapper"
+copy-icon: true
+separate-locales: false
+finish-args:
+  # Required due to being a GUI application
+  - --socket=x11
```
tinywrkb, line 14: Missing `--device=dri`.

org.twinery.Twine.yaml
```diff
@@ -0,0 +1,72 @@
...
+        sha256: "6e2b40c997fcf167ea280c1d6b65edaadc6f806a00468cb1f65b0fd2c98d3a62"
+        x-checker-data:
+          type: anitya
+          project-id: 236298
+          stable-only: true
+          url-template: "https://github.com/klembot/twinejs/releases/download/$version/twine_$version_linux64.zip"
+      - type: "script"
+        dest-filename: "twine.sh"
+        commands:
+          - export TMPDIR="$XDG_RUNTIME_DIR/app/${FLATPAK_ID:-org.twinery.Twine}"
```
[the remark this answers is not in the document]
SISheogorath: This is currently necessary to be compatible with flatpak version 1.0.9 which is shipped with CentOS 7, as this version of flatpak doesn't support `FLATPAK_ID` as environment variable.

See: https://github.com/flathub/im.riot.Riot/issues/116

org.twinery.Twine.yaml
```diff
@@ -0,0 +1,72 @@
...
+  - name: twine
+    buildsystem: simple
+    build-commands:
+      - mkdir twine
+      - cd twine && unzip ../twine.zip
+      - mv twine "${FLATPAK_DEST}"
+      - install -Dm755 twine.sh /app/bin/twine-wrapper
+      - install -Dm644 org.twinery.Twine.desktop "${FLATPAK_DEST}/share/applications/org.twinery.Twine.desktop"
+      - install -Dm644 org.twinery.Twine.metainfo.xml "${FLATPAK_DEST}/share/metainfo/org.twinery.Twine.metainfo.xml"
+      - install -Dm644 org.twinery.Twine.svg "${FLATPAK_DEST}/share/icons/hicolor/scalable/apps/${FLATPAK_ID}.svg"
```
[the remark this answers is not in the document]
SISheogorath: > "Applications must provide application icons in at least 64×64px and 128×128px sizes. Application icons should either be included as scalable SVGs or high-resolution PNGs."

The way I read this is that application icons have to be at least 128x128px and can be either SVGs or PNG (the latter of higher resolution). But I might misread that. If I do, that means I should through imagemagic against these icons, create them in first place?

org.twinery.Twine.yaml
```diff
@@ -0,0 +1,72 @@
+---
+app-id: "org.twinery.Twine"
+base: org.electronjs.Electron2.BaseApp
+base-version: '21.08'
+runtime: "org.freedesktop.Platform"
+runtime-version: "21.08"
+sdk: "org.freedesktop.Sdk"
+sdk-version: "21.08"
```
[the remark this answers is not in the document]
SISheogorath: Fixed!

org.twinery.Twine.yaml
SISheogorath: Uhm, why? I don't see where I would need hardware acceleration in a text editior or does Electron use this by default these days?

org.twinery.Twine.yaml
```diff
@@ -0,0 +1,72 @@
...
+  - "/share/doc"
+  - "/share/info"
+  - "/share/man"
+  - "/man"
+modules:
+  - name: twine
+    buildsystem: simple
+    build-commands:
+      - mkdir twine
+      - cd twine && unzip ../twine.zip
```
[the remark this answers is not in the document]
SISheogorath: With the tooling available in the freesoftware-sdk I actually don't see much room to use something else. But I'm open for suggestions. Be aware that this is a plain zip file, no tar wrapped in zip.

org.twinery.Twine.yaml
tinywrkb: According to [this](https://github.com/electron/electron/blob/main/docs/tutorial/offscreen-rendering.md), hardware rendering needs to be explicitly disabled in order to enable software rendering, or like what happening here, fail creating the GL context by not having a DRI device.  
With intel_gpu_top, I can see significant change in the render engine usage when the app is run with DRI device available in the sandbox.

org.twinery.Twine.yaml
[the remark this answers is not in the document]
tinywrkb: bsdtar is there, and while it can't handle everything, it's pretty close to.

```
bsdtar -xf twine.zip
```

org.twinery.Twine.yaml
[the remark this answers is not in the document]
tinywrkb: Thanks! I wasn't aware.

org.twinery.Twine.yaml
SISheogorath: 👍🏻

org.twinery.Twine.yaml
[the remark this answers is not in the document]
SISheogorath: Thanks! Done.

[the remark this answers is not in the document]
hfiguiere: can't use use "type": "archive" and let flatpak-builder handle the decompression?

org.twinery.Twine.yaml
hfiguiere: I thought the whole point of Electron was to be cross platform and portable. I must be old school in what this means...

SISheogorath: My concern is not electron itself, but how flatpak and the node ecosystem work. From my perspective there is no point in regenerating a flatpak dependency tree from multiple package-lock.json to run the exact same commands as upstream just to have the same files around as them.

Not to mention that in either case there will be binary blobs unless I build electron as part of this flatpak, which basically includes a complete chromium build, which again takes hours to complete, as you may already know from the chromium flatpak. I don't want to put on that shoe. If that's what we we will suddenly require on flathub, I will rather withdraw this submission and build this on my GitLab and push the flatpak to a s3 storage.

hfiguiere: there is no requirement whatsoever.

(also for the record, I have no say in what gets accepted or reject anyway)

org.twinery.Twine.yaml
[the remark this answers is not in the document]
barthalion: This is definitely enough. `appstream-compose` will create proper icons from the SVG.

org.twinery.Twine.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
```
barthalion, line 1: Should be upstreamed, if isn't already.

org.twinery.Twine.metainfo.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
barthalion, line 1: Should be upstreamed, if isn't already.

barthalion:
Any chance we can get explicit approval from the upstream developer?

SISheogorath:
@klembot it would be great if we could get approval for this flatpak :) I would also be interested in pushing the mentioned files upstream, but not sure where to put them yet.

For orientation: https://github.com/klembot/twinejs/issues/651

PR #2638, opened 2021-11-14
Com.doycho.euterpe.gtk

ironsmile:


com.doycho.euterpe.gtk.json
```diff
@@ -0,0 +1,68 @@
+{
+    "app-id" : "com.doycho.euterpe.gtk",
+    "runtime" : "org.gnome.Sdk",
+    "runtime-version" : "3.38",
```
nedrichards, line 4: is there a reason this doesn't work with a new gnome sdk? 41 is the latest release and will be in security support much longer.

com.doycho.euterpe.gtk.json
```diff
@@ -0,0 +1,68 @@
...
+    "sdk" : "org.gnome.Sdk",
+    "sdk-extensions": [
+        "org.freedesktop.Sdk.Extension.rust-stable"
+    ],
+    "command" : "euterpe-gtk",
+    "finish-args" : [
+        "--share=network",
+        "--share=ipc",
```
nedrichards, line 12: I haven't done a build yet to check but you're very likely to want --device=dri here for gpu acceleration given the stack you're using.

nedrichards:
bot, build com.doycho.euterpe.gtk

nedrichards:
The appdata doesn't pass validation. Tips for that and appdata usage are available https://github.com/flathub/flathub/wiki/AppData-Guidelines

com.doycho.euterpe.gtk.json
hfiguiere: also 3.38 is actually EOL.

shared-modules/maturin/fedc.json
```diff
@@ -0,0 +1,8 @@
+{
```
hfiguiere, line 1: `shared-modules` as a directory is usually used for the `shared-modules` as a git submodule. So I don't recommend using that directory name to organise things in a flathub project.

com.doycho.euterpe.gtk.json
nedrichards: There is that. ;-)

ironsmile: This is what Gnome Builder did when creating a new project and it seems to work just fine on all of my devices so I left it as it is :D Will give it a try with the latest SDK and I guess it will be alright too.

com.doycho.euterpe.gtk.json
```diff
@@ -0,0 +1,68 @@
...
+                    "url" : "https://github.com/ironsmile/euterpe-gtk.git",
+					"tag": "v0.1.0",
+					"commit": "2333d22a60742b1d4a864cfab68c2f458325ce83"
+                }
+            ]
+        }
+    ],
+    "build-options" : {
+        "env" : {        }
+    }
```
hfiguiere, line 67:
```suggestion
    ]
```

com.doycho.euterpe.gtk.json
ironsmile: Done. It seems to be working just fine with SDK 41.

com.doycho.euterpe.gtk.json
ironsmile: I've added it. To be honest I am pretty new to Flatpak so I am absolutely sure what I am doing. But from reading the docs I would love OpenGL rendering if that is an option. The above were what Gnome Builder generated and what else I needed to make the app do its job like access to pulseaudio.

shared-modules/maturin/fedc.json
ironsmile: Move them to `includes`.

com.doycho.euterpe.gtk.json
ironsmile: Are you sure? This will make `build-options` a key into the `modules` array. I am pretty sure it must in the top level object.

hfiguiere: I'm pretty this doesn't do what you think it does.

*hint* it's empty, it does nothing, it is not needed and this is what the "suggestion" is about: removing it.

ironsmile: Oh, I see. I misunderstood the suggestion diff. Anyway, removed it.

ironsmile:
> The appdata doesn't pass validation. Tips for that and appdata usage are available https://github.com/flathub/flathub/wiki/AppData-Guidelines

Done. I've added everything the validator requires. At least when I run the same validation command the builtbot runs then it passes. I read [the wiki](https://github.com/flathub/flathub/wiki/AppData-Guidelines) originally but figured some fields are optional, especially when there's nothing to be added to them. Like `<content_rating type="oars-1.0" />`. And for some reason I was left with the impression that validation is done as part of `flatpak install`. Guess not.

nedrichards:
bot, build com.doycho.euterpe.gtk

com.doycho.euterpe.gtk.json
```diff
@@ -0,0 +1,66 @@
+{
+    "app-id" : "com.doycho.euterpe.gtk",
+    "runtime" : "org.gnome.Sdk",
```
nedrichards, line 3: do you mean to be using the Sdk as the runtime? This is more normally org.gnome.Platform (unless you explicitly need dev tools at runtime)

com.doycho.euterpe.gtk.json
nedrichards: Yeah, device=dri will give you opengl or vulkan acceleration 'for free' with recent gtk.

nedrichards:
Other than the SDK/platform question this lgtm

com.doycho.euterpe.gtk.json
ironsmile: Good catch. I totally did not notice that when submitting. Changed the runtime to "org.gnome.Platform".

nedrichards:
bot, build com.doycho.euterpe.gtk

PR #2645, opened 2021-11-17
studio.kx.carla

Bleuzen:

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. [**Link**](https://github.com/falkTX/Carla/pull/1535)

### DIFF
--- a/appdata.xml
+++ b/appdata.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <component type="desktop">
-  <id>com.github.falkTX.Carla</id>
+  <id>studio.kx.carla</id>
   <metadata_license>CC0-1.0</metadata_license>
   <project_license>GPL-2.0</project_license>
   <name>Carla</name>
@@ -15,7 +15,7 @@
     </screenshot>
   </screenshots>
   <content_rating type="oars-1.1" />
-  <launchable type="desktop-id">com.github.falkTX.Carla.desktop</launchable>
+  <launchable type="desktop-id">studio.kx.carla.desktop</launchable>
   <url type="homepage">https://kx.studio/Applications:Carla</url>
   <url type="bugtracker">https://github.com/falkTX/Carla/issues</url>
   <releases>
--- a/appdata.xml
+++ b/appdata.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <component type="desktop">
  <id>com.github.falkTX.Carla</id>
- <metadata_license>CC0</metadata_license>
+ <metadata_license>CC0-1.0</metadata_license>
  <project_license>GPL-2.0</project_license>
  <name>Carla</name>
  <summary>Audio Plugin Host</summary>
--- a/appdata.xml
+++ b/appdata.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+ <id>com.github.falkTX.Carla</id>
+ <metadata_license>CC0</metadata_license>
+ <project_license>GPL-2.0</project_license>
+ <name>Carla</name>
+ <summary>Audio Plugin Host</summary>
+ <description>Carla is a fully-featured audio plugin host, with support for many audio drivers and plugin formats.</description>
+ <content_rating type="oars-1.1" />
+ <launchable type="desktop-id">com.github.falkTX.Carla.desktop</launchable>
+ <update_contact>supgesu_AT_gmail.com</update_contact>
+ <url type="homepage">https://kx.studio/Applications:Carla</url>
+ <url type="bugtracker">https://github.com/falkTX/Carla/issues</url>
+ <releases>
+   <release version="2.4.0" date="2021-08-19" />
+ </releases>
+</component>
--- a/appdata.xml
+++ b/appdata.xml
@@ -5,7 +5,9 @@
  <project_license>GPL-2.0</project_license>
  <name>Carla</name>
  <summary>Audio Plugin Host</summary>
- <description>Carla is a fully-featured audio plugin host, with support for many audio drivers and plugin formats.</description>
+ <description>
+  <p>Carla is a fully-featured audio plugin host, with support for many audio drivers and plugin formats.</p>
+ </description>
  <content_rating type="oars-1.1" />
  <launchable type="desktop-id">com.github.falkTX.Carla.desktop</launchable>
  <update_contact>supgesu_AT_gmail.com</update_contact>
--- a/appdata.xml
+++ b/appdata.xml
@@ -1,19 +1,24 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <component type="desktop">
- <id>com.github.falkTX.Carla</id>
- <metadata_license>CC0-1.0</metadata_license>
- <project_license>GPL-2.0</project_license>
- <name>Carla</name>
- <summary>Audio Plugin Host</summary>
- <description>
-  <p>Carla is a fully-featured audio plugin host, with support for many audio drivers and plugin formats.</p>
- </description>
- <content_rating type="oars-1.1" />
- <launchable type="desktop-id">com.github.falkTX.Carla.desktop</launchable>
- <update_contact>supgesu_AT_gmail.com</update_contact>
- <url type="homepage">https://kx.studio/Applications:Carla</url>
- <url type="bugtracker">https://github.com/falkTX/Carla/issues</url>
- <releases>
-   <release version="2.4.1" date="2021-10-15" />
- </releases>
+  <id>com.github.falkTX.Carla</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>GPL-2.0</project_license>
+  <name>Carla</name>
+  <summary>Audio Plugin Host</summary>
+  <description>
+    <p>Carla is a fully-featured audio plugin host, with support for many audio drivers and plugin formats.</p>
+  </description>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://kx.studio/screenshots/carla.png</image>
+      <image>https://kx.studio/screenshots/carla-patchbay.png</image>
+    </screenshot>
+  </screenshots>
+  <content_rating type="oars-1.1" />
+  <launchable type="desktop-id">com.github.falkTX.Carla.desktop</launchable>
+  <url type="homepage">https://kx.studio/Applications:Carla</url>
+  <url type="bugtracker">https://github.com/falkTX/Carla/issues</url>
+  <releases>
+    <release version="2.4.1" date="2021-10-15" />
+  </releases>
 </component>
--- a/appdata.xml
+++ b/appdata.xml
@@ -12,6 +12,6 @@
  <url type="homepage">https://kx.studio/Applications:Carla</url>
  <url type="bugtracker">https://github.com/falkTX/Carla/issues</url>
  <releases>
-   <release version="2.4.0" date="2021-08-19" />
+   <release version="2.4.1" date="2021-10-15" />
  </releases>
 </component>
--- a/com.github.falkTX.Carla.json
+++ b/com.github.falkTX.Carla.json
@@ -107,11 +107,6 @@
       ],
       "build-options": {
         "arch": {
-          "arm": {
-            "env": {
-              "NOOPT": "true"
-            }
-          },
           "aarch64": {
             "env": {
               "NOOPT": "true"
--- a/com.github.falkTX.Carla.json
+++ b/com.github.falkTX.Carla.json
@@ -21,7 +21,7 @@
     "--env=TMPDIR=/var/tmp",
     /* On SuSE it seems to be necessary for sound to work */
     "--env=ALSA_CONFIG_PATH=",
-    "--env=LV2_PATH=$HOME/.lv2:/app/extensions/Plugins/lv2"
+    "--env=LV2_PATH=$HOME/.lv2:/app/extensions/Plugins/lv2",
     "--env=DSSI_PATH=/app/extensions/Plugins/dssi",
     "--env=VST_PATH=/app/extensions/Plugins/lxvst",
     "--env=VST3_PATH=$HOME/.vst3:/app/extensions/Plugins/vst3"
--- a/com.github.falkTX.Carla.json
+++ b/com.github.falkTX.Carla.json
@@ -15,7 +15,6 @@
     "--device=all",
     /* For pipewire */
     "--filesystem=xdg-run/pipewire-0",
-    "--system-talk-name=org.freedesktop.RealtimeKit1",
     /* Allow loading, saving files from the home directory (portals don’t work yet) */
     "--filesystem=home",
     "--env=TMPDIR=/var/tmp",
--- a/com.github.falkTX.Carla.json
+++ b/com.github.falkTX.Carla.json
@@ -1,0 +1,140 @@
+{
+  "app-id": "com.github.falkTX.Carla",
+  "runtime": "org.kde.Platform",
+  "runtime-version": "5.15-21.08",
+  "sdk": "org.kde.Sdk",
+  "command": "carla",
+  "rename-desktop-file": "carla.desktop",
+  "rename-icon": "carla",
+  "finish-args": [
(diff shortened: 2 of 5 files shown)


Bleuzen:
Still WIP, because:
Carla in flatpak can't see the audio devices connected to the system. I'm not sure what is missing.

Also building with PyQT 5.15 fails, but 5.14 works for now.
If you are interested, I can also commit the 5.15 version, but have no idea how to fix it yet.

com.github.falkTX.Carla.json
```diff
@@ -0,0 +1,140 @@
...
+    "--filesystem=xdg-run/pipewire-0",
+    "--system-talk-name=org.freedesktop.RealtimeKit1",
+    /* Allow loading, saving files from the home directory (portals don’t work yet) */
+    "--filesystem=home",
+    "--env=QT_QPA_PLATFORM=xcb",
+    "--env=TMPDIR=/var/tmp",
+    /* On SuSE it seems to be necessary for sound to work */
+    "--env=ALSA_CONFIG_PATH=",
+    "--env=LD_LIBRARY_PATH=/app/lib:/app/lib/libgig",
+    "--env=LV2_PATH=/app/extensions/Plugins/lv2",
```
hfiguiere, line 25:
```suggestion
    "--env=LV2_PATH=$HOME/.lv2:/app/extensions/Plugins/lv2",
```

com.github.falkTX.Carla.json
```diff
@@ -0,0 +1,140 @@
...
+        "make PREFIX=/app -j $FLATPAK_BUILDER_N_JOBS",
+        "make PREFIX=/app install"
+      ],
+      "build-options": {
+        "arch": {
+          "arm": {
+            "env": {
+              "NOOPT": "true"
+            }
+          },
```
hfiguiere, line 115:
```suggestion
```
This arch is no longer support so you don't need it.

appdata.xml
```diff
@@ -0,0 +1,19 @@
...
+  <p>Carla is a fully-featured audio plugin host, with support for many audio drivers and plugin formats.</p>
+ </description>
+ <content_rating type="oars-1.1" />
+ <launchable type="desktop-id">com.github.falkTX.Carla.desktop</launchable>
+ <update_contact>supgesu_AT_gmail.com</update_contact>
+ <url type="homepage">https://kx.studio/Applications:Carla</url>
+ <url type="bugtracker">https://github.com/falkTX/Carla/issues</url>
+ <releases>
+   <release version="2.4.1" date="2021-10-15" />
+ </releases>
```
hfiguiere, line 18: missing `oars`, `screenshots`. so it won't validate.

hfiguiere: also the file should be submitted upstream.

com.github.falkTX.Carla.json
```diff
@@ -0,0 +1,140 @@
...
+    /* For pipewire */
+    "--filesystem=xdg-run/pipewire-0",
+    "--system-talk-name=org.freedesktop.RealtimeKit1",
+    /* Allow loading, saving files from the home directory (portals don’t work yet) */
+    "--filesystem=home",
+    "--env=QT_QPA_PLATFORM=xcb",
+    "--env=TMPDIR=/var/tmp",
+    /* On SuSE it seems to be necessary for sound to work */
+    "--env=ALSA_CONFIG_PATH=",
+    "--env=LD_LIBRARY_PATH=/app/lib:/app/lib/libgig",
```
hfiguiere, line 24:
```suggestion
```

There is no libgig in this package.

hfiguiere (changes requested):

hfiguiere:
bot, build com.github.falkTX.Carla

com.github.falkTX.Carla.json
Bleuzen: Is there a reason you only suggested this for lv2 and not also vst3?
Already saw this on Ardour: https://github.com/flathub/org.ardour.Ardour/pull/32
Is there something bad about loading vst3 plugins from home?

appdata.xml
Bleuzen: added screenshots and submitted it upstream: https://github.com/falkTX/Carla/pull/1535
not sure about oars, since there is a xml node for it. Any additional info needed?

com.github.falkTX.Carla.json
Bleuzen: added both lv2 and vst3 plugin paths for home now

Bleuzen:
bot, build com.github.falkTX.Carla

Bleuzen:
bot, build com.github.falkTX.Carla

Bleuzen:
bot, build com.github.falkTX.Carla

appdata.xml
hfiguiere: nevermind about oars, I missed seeing it here. It's good. Sorry abou that.

com.github.falkTX.Carla.json
hfiguiere: Following the spec for VST I believe it's not necessary, at least for Ardour (I remember the conversation with Robin about it since VST3_PATH was ignored).

I have to check with Carla though.

hfiguiere: also loading binaries from out of the sandbox is doomed to fail - even more when developers assume Linux = Ubuntu (from a few years ago) x86_64. But that's not what I was getting at here.

Bleuzen:
Tested this build. Works as described. It can load plugins, operates as a jack client, but can't see devices and other clients.

Native Carla:
image

Carla running in flatpak:
image

Still not sure what to do about it (missing permissions?).

hfiguiere:
do you have pipewire running on the *host* ?

Bleuzen:
@hfiguiere Yes, using it as replacement for pulse and jack:
```
❯ pactl info
Server String: /run/user/1000/pulse/native
Library Protocol Version: 35
Server Protocol Version: 35
Is Local: yes
Client Index: 58
Tile Size: 65472
User Name: elias
Host Name: archlinux
Server Name: PulseAudio (on PipeWire 0.3.40)
Server Version: 15.0.0
Default Sample Specification: float32le 2ch 48000Hz
Default Channel Map: front-left,front-right
Default Sink: alsa_output.usb-MOTU_M2_M20000044780-00.analog-stereo
Default Source: alsa_input.usb-MOTU_M2_M20000044780-00.analog-stereo
Cookie: 8193:3e3d
```

com.github.falkTX.Carla.json
```diff
@@ -0,0 +1,134 @@
...
+  "finish-args": [
+    "--share=ipc",
+    "--socket=x11",
+    "--socket=wayland",
+    "--socket=pulseaudio",
+    /* MIDI */
+    "--device=all",
+    /* For pipewire */
+    "--filesystem=xdg-run/pipewire-0",
+    "--system-talk-name=org.freedesktop.RealtimeKit1",
```
tinywrkb, line 18: I don't think this can work considering the PID namespace.  
Apps should use the [Realtime Portal](https://flatpak.github.io/xdg-desktop-portal/portal-docs.html#gdbus-org.freedesktop.portal.Realtime).  
If you added this for PipeWire (JACK) then I believe it's wrong and should be dropped.

hfiguiere: Some app crash if they don't have access to it - speaking from experience here. 

(can't speak for this one)

com.github.falkTX.Carla.json
```diff
@@ -0,0 +1,134 @@
...
+      "sources": [
+        {
+          "type": "archive",
+          "url": "https://files.pythonhosted.org/packages/4d/81/b9a66a28fb9a7bbeb60e266f06ebc4703e7e42b99e3609bf1b58ddd232b9/PyQt5-5.14.2.tar.gz",
+          "sha256": "bd230c6fd699eabf1ceb51e13a8b79b74c00a80272c622427b80141a22269eb0"
+        },
+        {
+          "type": "script",
+          "commands": [
+            "python3 configure.py --assume-shared --confirm-license --no-designer-plugin --no-qml-plugin --sysroot=/app --qsci-api --qsci-api-destdir=/app/qsci --sipdir=/app/share/sip --sip=/app/bin/sip --sip-incdir=/app/include QMAKE_CFLAGS_RELEASE='-I/usr/include/python3.9/' QMAKE_CXXFLAGS_RELEASE='-I/usr/include/python3.9/'"
```
tinywrkb, line 71: Please look at how I [packaged pyqt](https://github.com/flathub/org.qutebrowser.qutebrowser/blob/master/pyqt5/pyqt5.json), and the [configure script](https://github.com/flathub/org.qutebrowser.qutebrowser/blob/master/pyqt5/sip/sip-build-configure.in). This will help you avoid manually changing the Python version string when switching to a new runtime version.

com.github.falkTX.Carla.json
tinywrkb: @hfiguiere 
Thanks for pointing this out, I haven't hit this issue yet with the few JACK apps I packaged.

hfiguiere: maybe it got fixed since. But back when it was starting to work that was the case. And I don't really have the time to regress everything.

com.github.falkTX.Carla.json
hfiguiere: maybe we should make a shared-module for this.

tinywrkb: I started working on this, but I'm not sure what is the best way to let maintainer choose different configure options (enabled PyQt bindings) while not breaking automated updates/f-e-d-c of Flathub's shared-modules or the app packaging.

hfiguiere: We shouldn't need to choose different configure option. You can't when using distro package.

Bleuzen: Looking at all the files needed for pyqt5 in your flatpak, this looks way too complicated for building one simple dependency. Also I'm still a beginner to all of this, so not sure what exactly is really needed for Carla (to copy the pyqt5 build files from your project may be too much right?).

com.github.falkTX.Carla.json
Bleuzen: I found this in most other jack apps, so put it here. Not sure if it is really still needed. Could try without. But it is hard to test for now because something isn't working anyway (Carla can't see other clients and devices).

Bleuzen: removed it for now because it doesn't crash Carla

hfiguiere:
bot, build com.github.falkTX.Carla

hfiguiere:
I am not sure what's expected to be visible or not, but Carla as an audio plugin patch bay works for me. I don't think you have access to the whole graph from inside the sandbox. But with qjackctl outside the sandbox I can.

hfiguiere:
LGTM

hfiguiere:
short of the few changes mentionned

(yes I will update the documentation and other flatpaks)

Bleuzen:
@hfiguiere 

> I am not sure what's expected to be visible or not

Everything should be visible. Just like in native running Carla.
My main use case for Carla is to insert plugins in the graph and route applications through them. This is not possible without Carla having access to the full graph, so in this state it is pretty much useless.

> but Carla as an audio plugin patch bay works for me

Yes but need access to individual plugins, which is not possible currently because they are all running under one Carla jack client, which is a hefty limitation.


> I don't think you have access to the whole graph from inside the sandbox.

This should be possible. I even checked it with Ardour before starting this flatpak to make sure.
Example:

Here I have Guitarix and VLC running natively on the host (NO flatpak) and also Ardour running (IN flatpak).
Now look at what Ardour as a flatpak can see:

image

Also Ardour as a flatpak can see all connected audio devices:
image


So flatpaks can see the whole graph. But I'm not sure why Carla doesn't.

com.github.falkTX.Carla.json
Bleuzen: now that i got Carla fully working, I can confirm that this is not needed

com.github.falkTX.Carla.json
Bleuzen: > I have to check with Carla though.

I tested it now and turns out that Carla needs this path in env to find plugins in home. However it doesn't like to "$HOME" variable. It doesn't resolve to a full path in Carla.
I have to add
`--env=VST3_PATH=/home/elias/.vst3`
for it to work and if I do:
`--env=VST3_PATH=$HOME/.vst3`
Carla won't find the plugins in my user directory.

Bleuzen: > also loading binaries from out of the sandbox is doomed to fail

Not exactly. Most devs compile against LTS version of Ubuntu. This is old enough to work in Flatpak, so most plugins work just fine.

com.github.falkTX.Carla.json
```diff
@@ -0,0 +1,133 @@
...
+    /* MIDI */
+    "--device=all",
+    /* For pipewire */
+    "--filesystem=xdg-run/pipewire-0",
+    /* Allow loading, saving files from the home directory (portals don’t work yet) */
+    "--filesystem=home",
+    "--env=TMPDIR=/var/tmp",
+    /* On SuSE it seems to be necessary for sound to work */
+    "--env=ALSA_CONFIG_PATH=",
+    "--env=LV2_PATH=$HOME/.lv2:/app/extensions/Plugins/lv2",
```
hfiguiere, line 23:
```suggestion
    "--env=LV2_PATH=$HOME/.lv2:/app/extensions/Plugins/lv2:/app/lib/lv2",
```

You need that one too (documentation needs update).
Also TIL that `$HOME` isn't expanded. :-(

tinywrkb: > You need that one too (documentation needs update). Also TIL that `$HOME` isn't expanded. :-(

Maybe Flatpak should bind mount homedir to `/var/home` like done with the xdg-dirs to /var/{config,data,cache}.

com.github.falkTX.Carla.json
hfiguiere: And then they link against `libcurl-gnutls.so` because why not, everybody run Ubuntu, right?

See https://github.com/flathub/org.freedesktop.LinuxAudio.BaseExtension/issues/7

com.github.falkTX.Carla.json
hfiguiere: I filed https://github.com/flatpak/flatpak/issues/4596

com.github.falkTX.Carla.json
Bleuzen: I will leave it at this since it is the best we can do for now.

Bleuzen:
Removed the "[WIP]" from title because this is considered done and tested now.

There are only improvements left, namely:
- build with newer version of pyqt5, possibly use a shared-module when available (could need help from @tinywrkb)
- use app metainfo from upstream when merged

but both don't hinder Carla from working and can be done afterwards. So I think it is ready to merge now if no other regressions come up.

barthalion:
Has upstream developer been contacted about this submission, i.e. maintenance specifically?

Bleuzen:
@barthalion In the main post there is a pull request linked, so the developer knows about this submission. However I haven't asked about maintenance.

@falkTX Do you want to (co-)/maintain this flatpak?

falkTX:
> @falkTX Do you want to (co-)/maintain this flatpak?

Not really, I am yet to be convinced that flatpak is a good way to do plugins. The possibility of mixing different library versions is a major issue.

Plus I don't see myself ever using flatpaks myself, so I rather not have something labelled as supported that I don't use.

falkTX:
btw please don't use `com.github.falkTX.Carla` as id, github hosting is likely temporary.
Use `studio.kx.carla` instead, thanks.

hfiguiere:
> Not really, I am yet to be convinced that flatpak is a good way to do plugins. The possibility of mixing different library versions is a major issue.

Which is something that Flatpak actually take care of - unless random binaries on random distro. (not trying to convince you her)

BTW this is just the standalone version of Carla. Not Carla as a plugin.

Bleuzen:
updated now the app id

Bleuzen:
bot, build studio.kx.carla

barthalion:
If you don't mind @falkTX, we would like you to grant write access to the Flathub repo, nonetheless, just so you are not blocked from closing issues or merging PRs.

/merge @falkTX

falkTX:
Ok, so we can continue conversation there. I will create tickets as action items for individual things

PR #2655, opened 2021-11-25
new app: org.cockpit_project.CockpitClient

allisonkarlitskaya:

  - `--talk-name=org.freedesktop.Flatpak` is used to spawn the `ssh` command on the host (in order to take advantage of the user's ssh config to make outgoing connections, by explicit request of the user).  Note in particular that we don't do this from inside the sandbox by accessing the user's files because we don't have access to the network (and depend on a private network namespace to avoid exposing access to the app to other users on the system) and to ensure that we can access features like PKCS11 tokens.
  - LGPL 2.1+
  -  **Link:** https://github.com/cockpit-project/cockpit/graphs/contributors
  - https://cockpit-project.org/ (build from https://github.com/cockpit-project/cockpit-project.github.io/)
  - [x] `org.cockpit_project.CockpitClient.metainfo.xml` included here is autogenerated from a template in the upstream project: https://github.com/cockpit-project/cockpit/blob/main/containers/flatpak/cockpit-client.metainfo.xml.in
  - [x] `org.cockpit_project.CockpitClient.yml` included here is autogenerated from https://github.com/cockpit-project/cockpit/blob/main/containers/flatpak/cockpit-client.yml.in

### DIFF
--- a/org.cockpit_project.CockpitClient.metainfo.xml
+++ b/org.cockpit_project.CockpitClient.metainfo.xml
@@ -1,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<component type="desktop-application">
+  <id>org.cockpit_project.CockpitClient</id>
+
+  <name>Cockpit Client</name>
+  <summary>Connect via ssh to servers with Cockpit</summary>
+
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>LGPL-2.1-or-later</project_license>
+
+  <description>
+    <p>
+      Cockpit Client provides a graphical interface to your servers, containers, and virtual machines.  Connections are made over SSH, using the SSH configuration of the local user (including aliases, known hosts, key files, hardware tokens, etc).
+    </p>
+    <p>
+      The server needs only to have cockpit-bridge installed.  The Cockpit webserver doesn&apos;t need to be enabled, and no extra ports need to be opened.
+    </p>
+  </description>
+
+  <releases>
+    <release date="2021-11-24" version="257" type="stable"/>
+  </releases>
+
+  <launchable type="desktop-id">cockpit-client.desktop</launchable>
+
+  <categories>
+    <category>Network</category>
+    <category>RemoteAccess</category>
+  </categories>
+
+  <provides>
+    <binary>cockpit-client</binary>
+  </provides>
+
+  <requires>
+    <control>keyboard</control>
+    <control>pointing</control>
+    <display_length compare="ge">medium</display_length>
+  </requires>
+
+  <content_rating type="oars-1.0"/>
+
+  <url type="homepage">https://cockpit-project.org/</url>
+  <url type="bugtracker">https://github.com/cockpit-project/cockpit/issues</url>
+  <url type="help">https://cockpit-project.org/documentation.html</url>
+  <url type="translate">https://translate.fedoraproject.org/projects/cockpit</url>
+</component>
--- a/org.cockpit_project.CockpitClient.metainfo.xml
+++ b/org.cockpit_project.CockpitClient.metainfo.xml
@@ -18,6 +18,21 @@
     </p>
   </description>
 
+  <screenshots>
+    <screenshot>
+      <caption>Cockpit Client login screen</caption>
+      <image>https://cockpit-project.org/images/flatpak/flatpak-screenshot-login.png</image>
+    </screenshot>
+    <screenshot type="default">
+      <caption>Cockpit overview</caption>
+      <image>https://cockpit-project.org/images/flatpak/flatpak-screenshot-overview.png</image>
+    </screenshot>
+    <screenshot>
+      <caption>Cockpit terminal</caption>
+      <image>https://cockpit-project.org/images/flatpak/flatpak-screenshot-terminal.png</image>
+    </screenshot>
+  </screenshots>
+
   <releases>
     <release date="2021-11-24" version="257" type="stable"/>
   </releases>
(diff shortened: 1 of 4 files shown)


allisonkarlitskaya:
bot, build org.cockpit_project.CockpitClient

org.cockpit_project.CockpitClient.metainfo.xml
```diff
@@ -18,6 +18,21 @@
     </p>
   </description>
 
+  <screenshots>
+    <screenshot>
+      <caption>Cockpit Client login screen</caption>
+      <image>http://cockpit-project.org/images/flatpak/flatpak-screenshot-login.png</image>
```
martinpitt, line 24: Same as in https://github.com/cockpit-project/cockpit/pull/16646 -- https please, these are redirected anyway.

martinpitt (changes requested):
Thanks!

org.cockpit_project.CockpitClient.metainfo.xml
allisonkarlitskaya: fixed, thanks

org.cockpit_project.CockpitClient.yml
```diff
@@ -0,0 +1,38 @@
+app-id: org.cockpit_project.CockpitClient
+runtime: org.gnome.Platform
+runtime-version: '41'
+sdk: org.gnome.Sdk
+default-branch: stable
```
hfiguiere, line 5:
```suggestion
```

org.cockpit_project.CockpitClient.yml
```diff
@@ -0,0 +1,38 @@
...
+      - install -Dt /app/share/metainfo ${FLATPAK_ID}.metainfo.xml
+
+    sources:
+      - type: archive
+        url: https://github.com/cockpit-project/cockpit/releases/download/258/cockpit-258.tar.xz
+        sha256: 5806a53218da6b4679eb6087251fb1c0421347fd31c9f87e31c9ac86978c09cc
+
+      - type: file
+        path: org.cockpit_project.CockpitClient.metainfo.xml
+finish-args:
```
hfiguiere, line 34: finish-args are usually above modules as a common practice

martinpitt (approved):
Thanks!

hfiguiere (changes requested):
don't you need a `--share=network` too in the finish-args?

org.cockpit_project.CockpitClient.metainfo.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
allisonkarlitskaya: This is the one from upstream, automatically generated by a script which lives in the git repository.  It doesn't get included as part of the tarball, however: the reason for that is the version history.  It's easier to keep track of that here. when we push the new release, instead of having to explicitly write a commit for each version upstream.

org.cockpit_project.CockpitClient.yml
allisonkarlitskaya: This is here because this file is generated out of upstream git and we sometimes want to write `default-branch: devel` here (in case it's not for a released tarball).  We could maybe change that around to use a commandline argument to `flatpak-builder` but this seemed a lot easier.

org.cockpit_project.CockpitClient.yml
allisonkarlitskaya: OK.  I can move this.

org.cockpit_project.CockpitClient.yml
```diff
@@ -0,0 +1,38 @@
...
+  - --talk-name=org.freedesktop.Flatpak
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --device=dri
```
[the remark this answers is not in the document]
allisonkarlitskaya: I'll add that.

allisonkarlitskaya:
> don't you need a `--share=network` too in the finish-args?

No.  We run ssh on the host, and it's the only thing that talks to the outside world.  Meanwhile, inside of the flatpak sandbox, `cockpit-ws` is listening on a TCP port that we specifically want *not* to be exposed to the world.  This is a case of using flatpak's container to protect the app from the system, instead of protecting the system from the app.  We even have an explicit check to ensure that the user doesn't do a `--share=network` override.

org.cockpit_project.CockpitClient.metainfo.xml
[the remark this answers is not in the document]
hfiguiere: it should be part of the tarball. and you should use the one shipping when that happen.

nedrichards:
Are there any other upstream authors who should be added to this repo when its created?

org.cockpit_project.CockpitClient.metainfo.xml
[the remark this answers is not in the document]
nedrichards: Yeah, this guideline is intended to make sure that the whole Linux ecosystem gets better since we absolutely do not live in an 'ideal' world where flatpak is the only way of getting apps! I do understand some of the potentially annoying workflow issues this can put into 'doing a release' though.

nedrichards:
bot, build org.cockpit_project.CockpitClient

allisonkarlitskaya:
> Are there any other upstream authors who should be added to this repo when its created?

@martinpitt and myself should be a good start.

We intend to at least semi-automate version updates as part of our heavily automated release process, so we'll have to see if we need anything else.  I was thinking about a deploy key, but I actually start to favour our bot opening a PR and getting a test build from @flathubbot first.

allisonkarlitskaya:
Thanks very much!

PR #2686, opened 2021-12-15
Add org.love2d.love2d

JakobDev:

Love2d is simple Game Engine. This Flatpak should serve 2 purposes (if allowed)
1. Use as standalone. After installing this Flatpak, you can right click a .love file and use Open with>LÖVE 11 to run any Love2d Game you want.
2. It should also be used as base app for people who want to publish their love2d Game on Flathub, so they need only to include the .love file.

Love2d is not fully compatible between main versions e.g. version 11.x has some API changes to version 10.x, so you may need to use a older version to run a older .love file. I don't know if it's possible to have multiple versions of a App, but when it's possible, this should the way to go.


JakobDev:
bot, build org.love2d.love2d

JakobDev:
bot, build org.love2d.love2d

JakobDev:
bot, build org.love2d.love2d

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64"]
```
hfiguiere, line 2: luajit 2.1.0 allow building on the other 64-bits arch as well.

hfiguiere (changes requested):

org.love2d.love2d.yml
```diff
@@ -0,0 +1,42 @@
+app-id: org.love2d.love2d
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+command: love
+branch: "11"
```
hfiguiere, line 6:
```suggestion
```

nedrichards:
What's the upstream involvement status here? That's the only checkbox left unchecked on the initial template.

JakobDev:
bot, build org.love2d.love2d

JakobDev:
> What's the upstream involvement status here? That's the only checkbox left unchecked on the initial template.

I have not talk to upstream yet, but in a [Issue for a Snap package](https://github.com/love2d/love/issues/1482) they say that AppImages are enough in their opinion.

flathub.json
hfiguiere: so you can remove this file now.

JakobDev:
bot, build org.love2d.love2d

org.love2d.love2d.yml
```diff
@@ -0,0 +1,41 @@
+app-id: org.love2d.love2d
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+command: love
+finish-args:
+  - --socket=x11
```
barthalion, line 7: Doesn't it work under Wayland?

barthalion:
The problem with using it as "base app" is that LOVE is a moving target, so at some point it will stop being compatible.

JakobDev:
> The problem with using it as "base app" is that LOVE is a moving target, so at some point it will stop being compatible.

This problem also occurs in the normal usage. Let's say a user had made a .love file for version 11.x. Love 12 come and the Flatpak gets updated to Version 12. The .love file will no longer work. So I think it's important to distribute this Flatpak in multiple versions. So if a user wants to run a .love file for Version 11, he installs Version 11. If a user wants to run a .love file for Version 0.10, he installs Version 0.10. If a programmer wants to create a Flatpak for Version 11, he uses Version 11 of this Flatpak as Base.

nedrichards:
You can do that - and many of the baseapps around here do the same - it's slightly non-aligned with the user side of just tracking 'stable', makes more sense for developers so this may end up being a bit complicated.

barthalion:
The ideal solution would be to submit LOVE to https://github.com/flathub/shared-modules/ and then include it here. This way, it can be easily included for games packaged in a standalone way. I'm not a fan of BaseApps for codebases that are relatively fast to build, like we have here.

nedrichards:
Yeah, that's fair. Put it in both places.

JakobDev:
If everyone things that I should only target the latest version and not use this as baseapp, then I will do it that way.

JakobDev:
bot, build org.love2d.love2d

JakobDev:
I totally forgot the add the Wayland permission

org.love2d.love2d.yml
```diff
@@ -0,0 +1,42 @@
...
```
hfiguiere, line 7:
```suggestion
  - --socket=fallback-x11
```

joshua-stone:
I'm inclined to agree with @barthalion that love2d should be added to https://github.com/flathub/shared-modules. Based on my own experiences, love2d had a very unstable API, even among minor versions. For example, a couple titles I packaged some time ago specifically require 10.2 for running without glitches:

https://github.com/MarvellousSoft/MarvInc/tree/dev/dist/flatpak
https://github.com/joshua-stone/BYTEPATH/tree/master/dist/flatpak

A shared module for a number of love2d releases should be easy to maintain, especially considering how many releases there are:

https://love2d.org/wiki/Version_History

nedrichards:
The standalone use case is still valid though, right? Game creators are users too!

JakobDev:
bot, build org.love2d.love2d

JakobDev:
> I'm inclined to agree with @barthalion that love2d should be added to https://github.com/flathub/shared-modules. Based on my own experiences, love2d had a very unstable API, even among minor versions. For example, a couple titles I packaged some time ago specifically require 10.2 for running without glitches:
> 
> https://github.com/MarvellousSoft/MarvInc/tree/dev/dist/flatpak https://github.com/joshua-stone/BYTEPATH/tree/master/dist/flatpak
> 
> A shared module for a number of love2d releases should be easy to maintain, especially considering how many releases there are:
> 
> https://love2d.org/wiki/Version_History

I don't know that the Love2 is so different between minor versions. The only problem I had so far was the change of RGB values from 0-255 to 0-1 in Version 11. I will make a PR against shared modules.

> The standalone use case is still valid though, right? Game creators are users too!

Of course. And it can be also used by users who just play a downloaded .love file. Not only developers.

org.love2d.love2d.metainfo.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>org.love2d.love2d</id>
+  <name>LÖVE</name>
+  <summary>LÖVE is an awesome 2D game framework for Lua.</summary>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>Zlib</project_license>
+  <content_rating type="oars-1.0"/>
+  <description>
+    <p>LÖVE is an awesome 2D game framework for Lua. </p>
```
nedrichards, line 10: What's the expected way for a developer to actually run their game in this? A bit more explanation here would ease confusion.

nedrichards:
Yeah, let's get this in as it's useful. I think we need a bit more explanation as to the expected workflow (there's no filesystem permissions right?)

JakobDev:
> Yeah, let's get this in as it's useful. I think we need a bit more explanation as to the expected workflow (there's no filesystem permissions right?)

It's pretty easy: Just download a .love file. [Here is a example Game I found in the Love2d Forums](https://mikeeybikeey.itch.io/volcano-star). Just get the .love file from the linked site. After that, just right click it in your Filemanger. Open with > LÖVE. And you will see it works.

nedrichards:
bot, build org.love2d.love2d

nedrichards:
Yeah, I think we could add some of that in the description but that's fine by me.

Enderteck:
Hey you should get in contact with the love devs as they are gonna see if this might become an official release:

https://github.com/love2d/love/issues/2035

PR #2692, opened 2021-12-18
add com.seventhstring.Transcribe

MichaHoffmann:

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:https://www.seventhstring.com/xscribe/**


MichaHoffmann:
Hey,

first flatpak package for me! I contacted the developer of this app if its ok to distribute it on flathub and he agreed. Since i'm not the developer, im not sure how to handle the License, Application-ID, etc. Are there any recommendations for this scenario? I have been using this for myself locally until now.

org.seventhstring.Transcribe.yml
```diff
@@ -0,0 +1,40 @@
...
+  - name: transcribe
+    buildsystem: simple
+    build-commands: 
+    - mkdir -p /app/bin
+    - cp transcribe libgstvideosection.so xschelp.htb /app/bin
+    - mkdir -p /app/share/icons/hicolor/16x16/apps && mv gtkicons/xsc16x16.png /app/share/icons/hicolor/16x16/apps/org.seventhstring.Transcribe.png
+    - mkdir -p /app/share/icons/hicolor/32x32/apps && mv gtkicons/xsc32x32.png /app/share/icons/hicolor/32x32/apps/org.seventhstring.Transcribe.png
+    - mkdir -p /app/share/icons/hicolor/48x48/apps && mv gtkicons/xsc48x48.png /app/share/icons/hicolor/48x48/apps/org.seventhstring.Transcribe.png
+    - mkdir -p /app/share/icons/hicolor/128x128/apps && mv gtkicons/xsc128x128.png /app/share/icons/hicolor/128x128/apps/org.seventhstring.Transcribe.png
+    - |
```
hfiguiere, line 25: Why don't you run the install script instead?

org.seventhstring.Transcribe.yml
```diff
@@ -0,0 +1,40 @@
+app-id: org.seventhstring.Transcribe
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+command: transcribe
+finish-args:
+#TODO: figure out if its possible to only grant permissions 
+# on ~/.Transcribe!7 and xdg-music:ro, for now we use home
+# as a functional compromise
```
hfiguiere, line 9: please do.

org.seventhstring.Transcribe.yml
```diff
@@ -0,0 +1,40 @@
...
+- --share=ipc
+- --socket=x11
+- --socket=wayland
```
hfiguiere, line 12: does it support wayland? If yes, then it's "fallback-x11". If not then the wayland socket is removed.

org.seventhstring.Transcribe.yml
```diff
@@ -0,0 +1,40 @@
...
+      Encoding=UTF-8
+      Version=1.0
+      Type=Application
+      Exec=/app/bin/transcribe
+      Categories=Audio;
+      Terminal=false
+      Icon=org.seventhstring.Transcribe
+      Name=Transcribe!
+      " > /app/share/applications/org.seventhstring.Transcribe.desktop
+    sources:
```
hfiguiere, line 37: There is no appstream file. One is required.

org.seventhstring.Transcribe.yml
```diff
@@ -0,0 +1,40 @@
+app-id: org.seventhstring.Transcribe
```
hfiguiere, line 1: seventhstring.org redirect to seventhstring.com, and I think it's more appropriate to use that unless upstream already use something else.

hfiguiere (changes requested):

org.seventhstring.Transcribe.yml
```diff
@@ -0,0 +1,40 @@
...
+      Type=Application
+      Exec=/app/bin/transcribe
+      Categories=Audio;
+      Terminal=false
+      Icon=org.seventhstring.Transcribe
+      Name=Transcribe!
+      " > /app/share/applications/org.seventhstring.Transcribe.desktop
+    sources:
+      - type: archive
+        url: https://www.seventhstring.com/xscribe/downlo/xscsetup-9.10.0.tar.gz
```
hfiguiere, line 39: The documentation say "64-bit linux" but this is only x86_64 so you should make it `x86_64` only.

org.seventhstring.Transcribe.yml
MichaHoffmann: I initially tried to and then rename the icon/.desktop file using the `rename-icon` and `rename-desktop-file` option but that did not work. Also running the install script requires `xdg-utils` to be installed and it seemed simpler just to copy the files.

Having said that, should i try again using the install script?

org.seventhstring.Transcribe.yml
MichaHoffmann: The problem i faced here is that when i use `--filesystem=~/.Transcribe!7` or `persist=.Transcribe!7` it does not work as expected. The application will not persist that file in the host then and ask for a license key in the next startup again.

org.seventhstring.Transcribe.yml
MichaHoffmann: will do!

org.seventhstring.Transcribe.yml
MichaHoffmann: will do!

org.seventhstring.Transcribe.yml
hfiguiere: actually this is why I was asking about using the install script. Because otherwise this will be out of sync from upstream.

MichaHoffmann: Sorry, i did not understand that, could you explain again please?

hfiguiere: wrong thread

org.seventhstring.Transcribe.yml
MichaHoffmann: i tested it and it runs just fine with `fallback-x11` and `x11`

MichaHoffmann: removed wayland

org.seventhstring.Transcribe.yml
MichaHoffmann: done

org.seventhstring.Transcribe.yml
MichaHoffmann: added `flathub.json` such this is built only for `x86_64`

com.seventhstring.Transcribe.yml
```diff
@@ -0,0 +1,61 @@
...
+command: /app/bin/transcribe/transcribe
+rename-desktop-file: seventhstring-transcribe.desktop
+rename-icon: seventhstring-transcribe
+finish-args:
+#TODO: figure out if its possible to only grant permissions 
+# on ~/.Transcribe!7 and xdg-music:ro, for now we use home
+# as a functional compromise
+- --share=ipc
+- --socket=x11
+- --socket=fallback-x11
```
hfiguiere, line 14: This is not what I said.

org.seventhstring.Transcribe.yml
MichaHoffmann: ok i figured out how to make `rename-icon` and `rename-desktop-file` work. now only the install script is called

com.seventhstring.Transcribe.yml
hfiguiere:
```suggestion
- --socket=fallback-x11
- --socket=wayland
```

The binary links against Gtk3 so this should work.

com.seventhstring.Transcribe.yml
```diff
@@ -0,0 +1,61 @@
...
+      # the build script uses xdg-utils
+      #
+      # taken from https://github.com/flathub/com.calibre_ebook.calibre
+      - name: xdg-utils
+        buildsystem: autotools
+        cleanup: ['*']
+        sources:
+          - type: archive
+            url: https://portland.freedesktop.org/download/xdg-utils-1.1.3.tar.gz
+            sha1: 98294cf332c341b85e481d98e9ea59357fe1efc7
```
hfiguiere, line 42:
```suggestion
```

com.seventhstring.Transcribe.yml
MichaHoffmann: Oh sorry, i misunderstood! What did you mean?

I just had a look at a few other flatpaks i have installed and 
```
  - --share=ipc
  - --socket=x11
  - --socket=wayland
```
seems to be a fairly common configuration. should i go with that?

com.seventhstring.Transcribe.yml
```diff
@@ -0,0 +1,61 @@
+app-id: com.seventhstring.Transcribe
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+command: /app/bin/transcribe/transcribe
```
hfiguiere, line 5: this is never an absolute path. Just create a symlink in `/app/bin`

com.seventhstring.Transcribe.yml
MichaHoffmann: works just fine, can i drop `--share=ipc` too then?

hfiguiere:
bot, build com.seventhstring.Transcribe

com.seventhstring.Transcribe.yml
hfiguiere: I never mentionned it.

NO

hfiguiere:
still missing the appstream file.

MichaHoffmann:
> still missing the appstream file.

actively working on it! the current one is still wip

com.seventhstring.Transcribe.yml
MichaHoffmann: > I never mentionned it.
> 
> NO

ok! i have read somewhere that its needed for `socket=x11` permissions, so thats why i asked!

com.seventhstring.Transcribe.yml
MichaHoffmann: Why should the hash be removed? should it be sha256 instead?

com.seventhstring.Transcribe.yml
MichaHoffmann: done

org.seventhstring.Transcribe.yml
MichaHoffmann: added a WIP appstream file ( missing description still )

com.seventhstring.Transcribe.yml
MichaHoffmann: ah sorry, misread what you meant

hfiguiere:
bot, build com.seventhstring.Transcribe

MichaHoffmann:
> The appstream needs to be installed.

sorry forgot to install it :facepalm:, could you please start the build again?

hfiguiere:
you don't need any special permission to start a build.

MichaHoffmann:
bot, build com.seventhstring.Transcribe

MichaHoffmann:
bot, build com.seventhstring.Transcribe

MichaHoffmann:
I tested the build artifact just now, it works for me (after correcting a typo..)

com.seventhstring.Transcribe.yml
MichaHoffmann: the application does not seem to support wayland. running with `wayland + fallback-x11` on my local system ( using wayland ) gives: `13:34:59: Error: Unable to initialize GTK+, is DISPLAY set properly?`. running just with `x11` works fine, so thats what im inclined to do after some testing.

MichaHoffmann:
bot, build com.seventhstring.Transcribe

org.seventhstring.Transcribe.yml
MichaHoffmann: Is there a way to only allow access to the file `.Transcribe!7` in home? i could not figure it out i have to admit.

barthalion:
bot, build com.seventhstring.Transcribe

com.seventhstring.Transcribe.yml
```diff
@@ -0,0 +1,62 @@
...
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+command: transcribe
+rename-desktop-file: seventhstring-transcribe.desktop
+rename-icon: seventhstring-transcribe
+finish-args:
+#TODO: figure out if its possible to only grant permissions 
+# on ~/.Transcribe!7 and xdg-music:ro, for now we use home
+# as a functional compromise
```
barthalion, line 11: It would be great to figure this out before we publish it on Flathub. You can change it to '--persist=.Transcribe!7` and Flatpak will create proper bind-mount.

MichaHoffmann: Hey, i'll try it today after work!

MichaHoffmann: That seems to mount a directory called `.Transcribe!7` into the container. After entering the license key it looks like this on my machine: 
```
mhoffm@fedora:~$ flatpak enter com.seventhstring.Transcribe ls -lha
total 8.0K
drwx------. 6 mhoffm mhoffm  160 Dec 22 16:22  .
drwx------. 3 mhoffm mhoffm   60 Dec 22 16:22  ..
drwxr-xr-x. 1 mhoffm mhoffm    0 Dec 19 12:17 '.Transcribe!7'
-rwxr-xr-x. 1 mhoffm mhoffm 2.0K Dec 22 16:22 '.Transcribe!7j3O7ay'
drwxr-xr-x. 3 mhoffm mhoffm   60 Dec 22 16:22  .local
drwxr-xr-x. 3 mhoffm mhoffm   60 Dec 22 16:22  .var
-rw-r--r--. 1 mhoffm mhoffm   39 Dec 22 16:22  .yeqwqauh
drwxr-xr-x. 1 mhoffm mhoffm  120 Dec 13 18:28  Music
```

The application seems to create a file with some random suffix if the filename is already present. Is it possible to use files for `--persist`?

MichaHoffmann: Is it possible? I found this: https://github.com/flatpak/flatpak/blob/412c15772f794fd300991a91f2e0fe51f615bf79/common/flatpak-context.c#L2555 

It seems to me that the argument of `--persist` will always appear as a directory inside the container.

barthalion: Yeah, that's the case. Is `.Transcribe!7` supposed to be a text file? If yes, I would suggest looking for it in some kind of subdirectory (e.g. `.config/Transcribe/Transcribe!7`) instead. This way, if application respects `XDG_CONFIG_HOME` environment variable, there will be no conflicts.

There's also `--filesystem:path:create`, but it will also create a directory and not a file.

MichaHoffmann: Its a file yes! Its also not configurable, and the application does not respect XDG_CONFIG_HOME. After trying and playing around with it, I dont think there is a better way really.

barthalion: It would be possible to provide a wrapper script to work around these limitations, but as it probably doesn't use XDG portals anyway, let's keep it with `--filesystem=home`.

barthalion:
Does anyone from upstream team would like to have write access here?

MichaHoffmann:
> Does anyone from upstream team would like to have write access here?

Do you mean from the transcribe upstream?

barthalion:
Yeah, my understanding was that you are not directly affiliated with Transcribe team? In general, we want to involve upstream teams when possible, but it's not required if their answer is "we have no interest or capacity".

MichaHoffmann:
> Yeah, my understanding was that you are not directly affiliated with Transcribe team? In general, we want to involve upstream teams when possible, but it's not required if their answer is "we have no interest or capacity".

Thats true, i just like the program! Let me quickly ask them via email.

com.seventhstring.Transcribe.yml
MichaHoffmann: ok i utilized a wrapper script and that works! thanks

MichaHoffmann:
The upstream developer does not wish to be involved in the linux packaging.

MichaHoffmann:
bot, build com.seventhstring.Transcribe

MichaHoffmann:
bot, build com.seventhstring.Transcribe

PR #2694, opened 2021-12-19
Add org.syphon.Syphon

0x1a8510f2:
First attempt at making a flatpak so I may have overlooked something. Based largely on `im.fluffychat.Fluffychat` due to similar technologies and features.

As far as I know. I based the selection on Fluffychat which uses similar technologies and provides similar features.

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. https://github.com/syphon-org/syphon/issues/216

I personally don't, but Syphon's main author does.


0x1a8510f2:
bot, build org.syphon.Syphon

0x1a8510f2:
bot, build org.syphon.Syphon

0x1a8510f2:
bot, build org.syphon.Syphon

org.syphon.Syphon.desktop
```diff
@@ -0,0 +1,8 @@
+[Desktop Entry]
+Type=Application
+Name=Syphon
+Icon=org.syphon.Syphon
+Exec=env GDK_GL=gles syphon %U
```
hfiguiere, line 5: this should only be a command name. If you need to set some specific environment, you should put a wrapper shell script.

org.syphon.Syphon.json
```diff
@@ -0,0 +1,97 @@
...
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "41",
+    "sdk": "org.gnome.Sdk",
+    "command": "syphon",
+    "separate-locales": false,
+    "finish-args": [
+        "--share=ipc",
+        "--socket=x11",
+        "--socket=fallback-x11",
+        "--socket=wayland",
```
hfiguiere, line 12:
```suggestion
        "--socket=fallback-x11",
        "--socket=wayland",
```

This assume it actually support wayland

org.syphon.Syphon.json
```diff
@@ -0,0 +1,97 @@
...
+        "--socket=pulseaudio",
+        "--share=network",
+        "--device=all",
```
hfiguiere, line 15: why do you need "device all"

flathub.json
```diff
@@ -0,0 +1,7 @@
+{
+  "only-arches": [
+    "x86_64"
```
hfiguiere, line 3: this is sad given it's an open source project. Can't you build from source?

org.syphon.Syphon.json
```diff
@@ -0,0 +1,97 @@
...
+        "--filesystem=xdg-download",
+        "--talk-name=org.freedesktop.Notifications",
+        "--own-name=org.syphon.Syphon"
+    ],
+    "modules": [
+        {
+            "name": "libjsoncpp",
+            "buildsystem": "cmake",
+            "config-opts": [
+                "-DCMAKE_BUILD_TYPE=release",
```
hfiguiere, line 25:
```suggestion
                "-DCMAKE_BUILD_TYPE=RelWithDebInfo",
```

org.syphon.Syphon.json
```diff
@@ -0,0 +1,97 @@
...
+                {
+                    "type": "archive",
+                    "url": "https://github.com/open-source-parsers/jsoncpp/archive/refs/tags/1.9.5.tar.gz",
+                    "sha256": "f409856e5920c18d0c2fb85276e24ee607d2a09b5e7d5f0a371368903c275da2"
+                }
+            ]
+        },
+        {
+            "name": "olm",
+            "buildsystem": "cmake",
```
hfiguiere, line 38:
```suggestion
            "buildsystem": "cmake-ninja",
```

org.syphon.Syphon.json
```diff
@@ -0,0 +1,97 @@
...
+        "--share=network",
+        "--device=all",
+        "--filesystem=xdg-download",
+        "--talk-name=org.freedesktop.Notifications",
+        "--own-name=org.syphon.Syphon"
+    ],
+    "modules": [
+        {
+            "name": "libjsoncpp",
+            "buildsystem": "cmake",
```
hfiguiere, line 23:
```suggestion
            "buildsystem": "cmake-ninja",
```

org.syphon.Syphon.json
```diff
@@ -0,0 +1,97 @@
...
+            ]
+        },
+        {
+            "name": "syphon",
+            "buildsystem": "simple",
+            "only-arches": [
+                "x86_64"
+            ],
+            "build-commands": [
+                "ls syphon",
```
hfiguiere, line 54:
```suggestion
```

hfiguiere (changes requested):

org.syphon.Syphon.json
```diff
@@ -0,0 +1,97 @@
...
+        "--own-name=org.syphon.Syphon"
+    ],
+    "modules": [
+        {
+            "name": "libjsoncpp",
+            "buildsystem": "cmake",
+            "config-opts": [
+                "-DCMAKE_BUILD_TYPE=release",
+                "-DBUILD_SHARED_LIBS=ON"
+            ],
```
hfiguiere, line 27:
```suggestion
            ],
            "builddir": true,
```

This is what I would try to solve the build problem.

BTW I'm still surprised you need to build these dependencies since you unpack a binary distribution.

org.syphon.Syphon.json
0x1a8510f2: Syphon currently supports sending photos from the webcam and support for video/calls and voice messages is planned. I believe this is required to access the webcam and mic, correct?

org.syphon.Syphon.json
0x1a8510f2: Doesn't work as this is an array not a map. Should this be somewhere else?

hfiguiere: This is not what the "suggestion" says.

org.syphon.Syphon.json
hfiguiere: webcam. fair enough.

org.syphon.Syphon.json
0x1a8510f2: Aha, indeed it does not. Got confused because GitHub highlighted too much in red.

image

In local testing, that fixes it, thanks! :D

flathub.json
0x1a8510f2: Official ARM64 Linux builds of Syphon exist, but they're currently a bit irregular. I will add ARM64 eventually though, and probably quite soon.

org.syphon.Syphon.json
0x1a8510f2: >BTW I'm still surprised you need to build these dependencies since you unpack a binary distribution.

For some reason, flutter seems to include some libs and not others. On that topic also, using version 1.7.5 works but 1.9.5 produces `syphon: error while loading shared libraries: libjsoncpp.so.1: cannot open shared object file: No such file or directory`, even if I add `ln -s -r /app/lib/libjsoncpp.so /app/lib/libjsoncpp.so.1`. Any idea what could be going wrong?

org.syphon.Syphon.json
0x1a8510f2: Out of curiosity, why this change?

org.syphon.Syphon.desktop
0x1a8510f2: Yeah, this is almost certainly not needed either, just left over from the Fluffychat desktop file. Removed.

org.syphon.Syphon.json
```diff
@@ -0,0 +1,98 @@
...
+    "runtime-version": "41",
+    "sdk": "org.gnome.Sdk",
+    "command": "syphon",
+    "separate-locales": false,
+    "finish-args": [
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--socket=pulseaudio",
+        "--socket=session-bus",
```
0x1a8510f2, line 13: According to https://github.com/flathub/flathub/wiki/App-Requirements#dbus-access, I shouldn't be using this. However, without this the app dies with:

```
Failed to register: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: org.freedesktop.DBus.Error.ServiceUnknown
```
According to https://github.com/elementary/docs/issues/120, this is the solution.

Please advise.

org.syphon.Syphon.json
0x1a8510f2: Turns out, it's just incompatible

org.syphon.Syphon.json
nedrichards: Debug info gets automatically removed from the release that most users get, but it's packaged in a .Debug extension and can be added back in to help make development and bug fixing easier.

org.syphon.Syphon.json
nedrichards: run with `--log-session-bus` (this is super verbose) and look at the address' that are actually called. allowlist the required ones.

org.syphon.Syphon.json
0x1a8510f2: Ah, okay! :)

nedrichards:
bot, build org.syphon.Syphon

org.syphon.Syphon.json
0x1a8510f2: I've figured out that the `Failed to register` error was down to the binary having a different package name (old name) to the flatpak. Patch submitted upstream: https://github.com/syphon-org/syphon/pull/450. However, I'm still having issues with some features breaking without `--socket=system-bus`. I believe I found a possible cause, this being the [`connectivity_plus`](https://github.com/fluttercommunity/plus_plugins/blob/main/packages/connectivity_plus/connectivity_plus_linux/lib/src/connectivity.dart) dep which depends on [`nm.dart`](https://github.com/canonical/nm.dart/blob/442cac055ac44d5610e55f45e8360eb6e57c6ac7/lib/src/network_manager_client.dart) which, in turn, attempts to use dbus to communicate with NetworkManager. I haven't been able to track down exactly what permissions it needs though. Would you mind taking a look?

Likely worth noting is the following error from the app logs:
```
[ERROR:flutter/lib/ui/ui_dart_state.cc(209)] Unhandled Exception: SocketException: Connection failed (OS Error: No such file or directory, errno = 2), address = /var/run/dbus/system_bus_socket, port = 0
```

hfiguiere: Like the session bus you can log the system bus with `--log-system-bus --socket=system-bus` to find which `--system-talk-name=`

0x1a8510f2: Aha, I was using `--talk-name` instead of `--system-talk-name`. That solved it. Thanks!

0x1a8510f2:
Just the following left to do now:
- Wait for new Syphon release with updated app ID (`org.syphon.Syphon`)
- Fix storage issues where Syphon can't seem to be able to save anything
- Bit of QA to make sure everything works

flathub.json
```diff
@@ -0,0 +1,7 @@
...
+  ],
+  "publish-delay-hours": 1,
+  "automerge-flathubbot-prs": false
```
barthalion, line 6:
```suggestion
```

barthalion:
Once your submission is ready, should we also invite the main developer to the Flathub repo?

org.syphon.Syphon.json
```diff
@@ -0,0 +1,96 @@
...
+                    "dest": "syphon",
+                    "sha256": "7b4a2f5929b04131eb9b74e28cf97a1ef70c3ac3e3224e41092f12ac42c7c6fa",
+                    "x-checker-data": {
+                        "type": "json",
+                        "is-main-source": true,
+                        "url": "https://api.github.com/repos/syphon-org/syphon/releases/latest",
+                        "version-query": ".tag_name",
+                        "tag-query": ".tag_name",
+                        "timestamp-query": ".published_at",
+                        "url-query": ".html_url"
```
gasinvein, line 78: The html url isn't a link is something like [this](https://github.com/syphon-org/syphon/releases/tag/0.2.5), I assume it isn't what you wanted.
What you did want is probably this:
```suggestion
                        "url-query": ".assets[] | select(.name==\"syphon.linux.x64.tar.gz\") | .browser_download_url"
```

barthalion:
@0x1a8510f2 Please ping me when the new release is out and you consider it ready to merge.

0x1a8510f2:
New release is out already (which reminds me, I should update the manifest), but I'm having some issues with a random segfault :/

hfiguiere:
bot, build org.syphon.Syphon

0x1a8510f2:
bot, build org.syphon.Syphon

nedrichards:
@0x1a8510f2 is this OK to go now?

0x1a8510f2:
Sadly not. It seems that after every issue I solve, there's another around the corner. Finished dealing with the segfault but now I can't get persistence unless I use `--filesystem=host` so will need to figure out where stuff is saved.

I'm considering closing this PR and re-opening when I've confirmed everything is working. Would you say that would be the preferred course of action or should I leave it open?

nedrichards:
leave it open and update when you can. it's always good to have the context and if nothing happens it will end up getting automatically closed eventually.

hfiguiere:
bot, build org.syphon.Syphon

barthalion:
@0x1a8510f2 Is the new build better? FYI you can trigger new builds yourself by posting a comment like we do #

0x1a8510f2:
It turns out one lib is still broken because we accidentally copied a symlink. Working on it, but may need to wait for next Syphon release

0x1a8510f2:
bot, build org.syphon.Syphon

0x1a8510f2:
At last, I think Syphon is ready to be published! :D Thank you for waiting

0x1a8510f2:
Ping @barthalion as requested

org.syphon.Syphon.json
gasinvein: Please, either fix the `x-checker-data` metadata or remove it.

0x1a8510f2: Aha, sorry, forgot about this. I'll commit your suggestion.

0x1a8510f2:
bot, build org.syphon.Syphon

0x1a8510f2:
Thanks! Can @ereio be invited to that repo also?

barthalion:
Hi @ereio, just invited you.

ereio:
Accepted, thanks @0x1a8510f2, @barthalion and everyone for making this work! :)

PR #2705, opened 2021-12-26
com.github.christianrauch.Jahresarbeit-2003

christianrauch:

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:**https://github.com/christianrauch/Jahresarbeit-2003

### DIFF
--- a/com.github.christianrauch.Jahresarbeit-2003.yaml
+++ b/com.github.christianrauch.Jahresarbeit-2003.yaml
@@ -101,3 +101,21 @@ modules:
     builddir: true
     config-opts:
     - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+
+  - name: launcher
+    sources:
+    - type: file
+      path: ja2003.desktop
+    - type: file
+      url: https://raw.githubusercontent.com/christianrauch/Jahresarbeit-2003/master/icon1.svg
+      sha256: 911dbd92eb76cc8534e2210b68836b7e3a1e2a3672bfe47e6e211134d81eeffe
+    - type: file
+      path: ja2003.appdata.xml
+    buildsystem: simple
+    build-commands:
+    - install -d /app/share/applications
+    - install -m644 ja2003.desktop /app/share/applications/com.github.christianrauch.Jahresarbeit-2003.desktop
+    - install -d /app/share/icons/hicolor/scalable/apps
+    - install -m644 icon1.svg /app/share/icons/hicolor/scalable/apps/com.github.christianrauch.Jahresarbeit-2003.svg
+    - install -d /app/share/metainfo
+    - install -m644 ja2003.appdata.xml /app/share/metainfo/com.github.christianrauch.Jahresarbeit-2003.appdata.xml
--- a/com.github.christianrauch.Jahresarbeit-2003.yaml
+++ b/com.github.christianrauch.Jahresarbeit-2003.yaml
@@ -1,0 +1,20 @@
+app-id: com.github.christianrauch.Jahresarbeit-2003
+runtime: org.freedesktop.Platform
+runtime-version: "21.08"
+sdk: org.freedesktop.Sdk
+command: /app/ja2003
+finish-args:
+- --socket=x11
+- --device=dri
+- --share=ipc
+
+modules:
+  - name: JA2003
+    sources:
+    - type: git
+      url: https://github.com/christianrauch/Jahresarbeit-2003.git
+      tag: "2.0"
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+    - -DCMAKE_BUILD_TYPE=RelWithDebInfo
--- a/com.github.christianrauch.Jahresarbeit-2003.yaml
+++ b/com.github.christianrauch.Jahresarbeit-2003.yaml
@@ -4,7 +4,8 @@ runtime-version: "21.08"
 sdk: org.freedesktop.Sdk
 command: /app/ja2003
 finish-args:
-- --socket=x11
+- --socket=wayland
+- --socket=fallback-x11
 - --device=dri
 - --share=ipc
 
--- a/com.github.christianrauch.Jahresarbeit-2003.yaml
+++ b/com.github.christianrauch.Jahresarbeit-2003.yaml
@@ -8,8 +8,90 @@ finish-args:
 - --socket=fallback-x11
 - --device=dri
 - --share=ipc
+cleanup:
+- /include
+- /bin
+- /lib/cmake
+- /lib/pkg-config
+- /share
+- "*.a"
 
 modules:
+  - name: libdecor
+    sources:
+    - type: git
+      url: https://gitlab.gnome.org/jadahl/libdecor.git
+      tag: "0.1.0"
+    buildsystem: meson
+    builddir: true
+
+  - name: SDL2
+    sources:
+    - type: archive
+      url: https://github.com/libsdl-org/SDL/archive/refs/tags/release-2.0.18.tar.gz
+      sha256: b7061be3d86d14cdddbc696ab5b377ed28418f33be256c92215a57363329aa5f
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+    - -DCMAKE_BUILD_TYPE=Release
+    # subsystems
+    - -DSDL_ATOMIC=OFF
+    - -DSDL_AUDIO=OFF
+    - -DSDL_VIDEO=ON
+    - -DSDL_RENDER=ON
+    - -DSDL_EVENTS=ON
+    - -DSDL_JOYSTICK=ON
+    - -DSDL_HAPTIC=ON
+    - -DSDL_HIDAPI=OFF
+    - -DSDL_POWER=OFF
+    - -DSDL_THREADS=OFF
+    - -DSDL_TIMERS=OFF
+    - -DSDL_FILE=OFF
+    - -DSDL_LOADSO=ON
+    - -DSDL_CPUINFO=OFF
+    - -DSDL_FILESYSTEM=ON
+    - -DSDL_DLOPEN=OFF
+    - -DSDL_SENSOR=OFF
+    - -DSDL_LOCALE=OFF
+    # Wayland build with libdecor support
+    - -DSDL_WAYLAND_LIBDECOR=ON
+
+  - name: SDL2-ttf
+    sources:
+    - type: archive
+      url: https://github.com/libsdl-org/SDL_ttf/archive/refs/tags/release-2.0.15.tar.gz
+      sha256: 02e887b560faf398cbd60f56ce0a1cbaf96012dd4ddaa455f8307cb4911c86cc
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+    - -DCMAKE_BUILD_TYPE=Release
+
+  - name: SDL2-image
+    sources:
+    - type: git
+      url: https://github.com/libsdl-org/SDL_image.git
+      commit: a31da6b
+    buildsystem: autotools
+    config-opts:
+    - --disable-bmp
+    - --enable-gif
+    - --enable-jpg
+    - --disable-jpg-shared
+    - --disable-lbm
+    - --disable-pcx
+    - --disable-png
+    - --disable-png-shared
+    - --disable-pnm
+    - --disable-svg
+    - --disable-tga
+    - --disable-tif
+    - --disable-tif-shared
+    - --disable-xcf
+    - --disable-xpm
+    - --disable-xv
+    - --disable-webp
+    - --disable-webp-shared
+
   - name: JA2003
     sources:
     - type: git
--- a/com.github.christianrauch.Jahresarbeit-2003.yaml
+++ b/com.github.christianrauch.Jahresarbeit-2003.yaml
@@ -13,7 +13,7 @@ cleanup:
 - /bin
 - /lib/cmake
 - /lib/pkg-config
-- /share
+- /share/aclocal
 - "*.a"
 
 modules:
--- a/ja2003.appdata.xml
+++ b/ja2003.appdata.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.github.christianrauch.Jahresarbeit-2003</id>
+  
+  <name>Jahresarbeit 2003</name>
+  <summary xml:lang="de">Find die größte Fläche von zusammenhängenden Quadraten</summary>
+  <summary>Find the largest area of connected squares</summary>
+
+  <url type="homepage">https://github.com/christianrauch/Jahresarbeit-2003</url>
+  
+  <project_license>GPL-2.0 AND OFL-1.1 AND Apache-2.0</project_license>
+  <metadata_license>CC0-1.0</metadata_license>
+  
+  <recommends>
+    <display_length compare="ge">1024</display_length>
+  </recommends>
+  
+  <description xml:lang="de">
+    <p>
+      Das Programm &quot;Jahresarbeit 2003&quot; entstand als Jahresarbeit im Fach Informatik. Nach Spielstart sieht man ein 9x9 Feld mit verschiedenfarbigen Quadraten. Am Rand des Feldes sind Pfeile zu erkennen, welche genutzt werden um die Reihen und Spalten in denen sich die Quadrate befinden zu verschieben. Zusammenhängende gleichfarbige Quadrate bilden eine Fläche, welche durch draufklicken entfernt wird. Ziel des Spieles ist es eine möglichst große Fläche in kürzester Zeit weg zu klicken. Das Spiel ist zu Ende wenn alle Quadrate entfernt sind oder es keine Möglichkeit mehr gibt um die restlichen Quadrate zu entfernen. Jedoch sollte man um alle Quadrate zu entfernen so wenig Klicks wie möglich gebrauchen, da diese die Endpunktzahl verringern.
+    </p>
+  </description>
+  <description>
+    <p>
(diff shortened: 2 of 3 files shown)


christianrauch:
bot, build

christianrauch:
Can someone trigger the buildbot on this?

The AppStream Metadata is referencing local files for the screenshots. I am not sure if this is correct and would like to run the buildbot over the manifest to check for any errors that I cannot catch locally.

com.github.christianrauch.Jahresarbeit-2003.yaml
```diff
@@ -0,0 +1,121 @@
+app-id: com.github.christianrauch.Jahresarbeit-2003
+runtime: org.freedesktop.Platform
+runtime-version: "21.08"
+sdk: org.freedesktop.Sdk
+command: /app/ja2003
```
hfiguiere, line 5: command is never an absolute path. Also the executable shoud be in `/app/bin`

com.github.christianrauch.Jahresarbeit-2003.yaml
```diff
@@ -0,0 +1,121 @@
...
+finish-args:
+- --socket=wayland
+- --socket=fallback-x11
+- --device=dri
```
hfiguiere, line 9:
```suggestion
- --device=dri
- --share=ipc
```

this is needed for X11 anyway.

com.github.christianrauch.Jahresarbeit-2003.yaml
```diff
@@ -0,0 +1,121 @@
...
+cleanup:
+- /include
+- /bin
```
hfiguiere, line 12:
```suggestion
```

You need the main program to be in there.

com.github.christianrauch.Jahresarbeit-2003.yaml
```diff
@@ -0,0 +1,121 @@
...
+
+  - name: JA2003
+    sources:
+    - type: git
+      url: https://github.com/christianrauch/Jahresarbeit-2003.git
+      tag: "2.0"
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+    - -DCMAKE_BUILD_TYPE=Release
```
hfiguiere, line 103:
```suggestion
    - -DCMAKE_BUILD_TYPE=RelWithDebInfo
```

ja2003.appdata.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
hfiguiere, line 1: 1. did you submit these upstream
2. ideally the file should have the proper name from start.

ja2003.appdata.xml
```diff
@@ -0,0 +1,39 @@
...
+  <id>com.github.christianrauch.Jahresarbeit-2003</id>
+  
+  <name>Jahresarbeit 2003</name>
+  <summary>Find die größte Fläche von zusammenhängenden Quadraten</summary>
+
+  <url type="homepage">https://github.com/christianrauch/Jahresarbeit-2003</url>
+  
+  <project_license>GPL-2.0</project_license>
+  <metadata_license>OFL-1.1</metadata_license>
+  <metadata_license>Apache-2.0</metadata_license>
```
hfiguiere, line 12: 1. only one `metadata_license`
2. CC0-1.0 is prefered. I don't think any of those two is accepted.

ja2003.appdata.xml
```diff
@@ -0,0 +1,39 @@
...
+    <p>
+      Das Programm &quot;Jahresarbeit 2003&quot; entstand als Jahresarbeit im Fach Informatik. Nach Spielstart sieht man ein 9x9 Feld mit verschiedenfarbigen Quadraten. Am Rand des Feldes sind Pfeile zu erkennen, welche genutzt werden um die Reihen und Spalten in denen sich die Quadrate befinden zu verschieben. Zusammenhängende gleichfarbige Quadrate bilden eine Fläche, welche durch draufklicken entfernt wird. Ziel des Spieles ist es eine möglichst große Fläche in kürzester Zeit weg zu klicken. Das Spiel ist zu Ende wenn alle Quadrate entfernt sind oder es keine Möglichkeit mehr gibt um die restlichen Quadrate zu entfernen. Jedoch sollte man um alle Quadrate zu entfernen so wenig Klicks wie möglich gebrauchen, da diese die Endpunktzahl verringern.
+    </p>
+  </description>
+  
+  <launchable type="desktop-id">com.github.christianrauch.Jahresarbeit-2003.desktop</launchable>
+  <screenshots>
+    <screenshot type="default">
+      <caption>Hauptmenü</caption>
+      <image>screenshots/menu.png</image>
```
hfiguiere, line 28: screenshot should be accessible from a public URL. This will fail

ja2003.desktop
```diff
@@ -0,0 +1,8 @@
+[Desktop Entry]
+Version=1.0
+Type=Application
+Name=Jahresarbeit 2003
+Comment=Find die größte Fläche von zusammenhängenden Quadraten
```
hfiguiere, line 5: any chance to get these description in English too?
desktop files support localisation

ja2003.desktop
```diff
@@ -0,0 +1,8 @@
...
+Categories=Game;LogicGame;
+Icon=com.github.christianrauch.Jahresarbeit-2003
+Exec=/app/ja2003
```
hfiguiere, line 8: `Exec` is never an absolute path.

ja2003.appdata.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.github.christianrauch.Jahresarbeit-2003</id>
+  
+  <name>Jahresarbeit 2003</name>
+  <summary>Find die größte Fläche von zusammenhängenden Quadraten</summary>
```
hfiguiere, line 6: any chance to get these in English? AppStream files do support localisation.

ja2003.desktop
```diff
@@ -0,0 +1,8 @@
+[Desktop Entry]
```
hfiguiere, line 1: Has this been submitted upstream?
Also ideally it would have the proper name from start

com.github.christianrauch.Jahresarbeit-2003.yaml
```diff
@@ -0,0 +1,121 @@
...
+
+  - name: launcher
```
hfiguiere, line 105: usually the metadata file are declared / installed with the main module.

hfiguiere (changes requested):

hfiguiere:
bot, build com.github.christianrauch.Jahresarbeit-2003

hfiguiere:
you need to tell the bot what to build. It doesn't guess.

hfiguiere:
missing `content_rating` and `releases`. Also the screenshot.

com.github.christianrauch.Jahresarbeit-2003.yaml
```diff
@@ -0,0 +1,121 @@
...
+command: /app/ja2003
+finish-args:
+- --socket=wayland
+- --socket=fallback-x11
+- --device=dri
+cleanup:
+- /include
+- /bin
+- /lib/cmake
+- /lib/debug
```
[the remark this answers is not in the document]
christianrauch: I think in my case there are no debug symbols. This folder will be empty so I gathered it is not used.

com.github.christianrauch.Jahresarbeit-2003.yaml
christianrauch: Placing the executable in bin will break the application since it is searching for resource files starting at the directory of the executable. SDL explicitly mentions this for [`SDL_GetBasePath`](https://wiki.libsdl.org/SDL_GetBasePath).

com.github.christianrauch.Jahresarbeit-2003.yaml
christianrauch: Do I need to build with Debug symbols? In this case, it makes sense to keep `/lib/debug`.

com.github.christianrauch.Jahresarbeit-2003.yaml
christianrauch: As mentioned in the other commit, this will break the search path for resources (graphics, fonts, ...).

ja2003.appdata.xml
christianrauch: Well, "I am upstream". The AppData is not in the main repo. It is quite OS specific and only used for the flatpak for now.

ja2003.appdata.xml
christianrauch: I may have misinterpreted what "metadata" is. Is this the AppData XML itself? Or is this for "resources" like images? How do I specify that a project is covered by multiple licences? E.g. GPL for code, OFL for fonts, etc.?

ja2003.appdata.xml
hfiguiere: no, it is not used only by Flatpak. It is as specific as a .desktop file, that if you don't ship one with your Linux app, it's a bug.

com.github.christianrauch.Jahresarbeit-2003.yaml
hfiguiere: what people do in that case is they put a wrapper script in bin.

Standard on UNIX-like is to have executables in `bin`. Otherwise they can't be found..

com.github.christianrauch.Jahresarbeit-2003.yaml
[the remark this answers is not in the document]
hfiguiere: it's always wrong to remove it.

com.github.christianrauch.Jahresarbeit-2003.yaml
hfiguiere: hence the suggestion.

ja2003.appdata.xml
christianrauch: Well, they are part of this PR. So they will be public once it is merged and a repo was created. I don't know how I can reference them otherwise.

ja2003.desktop
christianrauch: The program itself is mostly hardcoded in German. Does it make sense to have the description multi-lingual if the application only supports a single language?

ja2003.appdata.xml
hfiguiere: You mentionned above in https://github.com/flathub/flathub/pull/2705#issuecomment-1001226835 "AppStream metadata". This is exactly that.

ja2003.appdata.xml
hfiguiere: they have to be available at the time of the build.

ja2003.desktop
christianrauch: No. This is the same as for the AppData. "I am upstream" and this desktop file is specific for the flatpak.

ja2003.desktop
christianrauch: This path has to be absolut because the executable is not in the PATH and is has to be in the root of the folder to discover the resources.

com.github.christianrauch.Jahresarbeit-2003.yaml
christianrauch: Most of the metadata (AppData, *.desktop files) encode the flatpak specific ID. So I am maintaining them in the dedicated flatpak repo outside of the generic main repo.

ja2003.desktop
hfiguiere: desktop file are NOT specific to flatpak either.

com.github.christianrauch.Jahresarbeit-2003.yaml
hfiguiere: 1, this is irrelevant to what I just said here.
2. lot of package don't have to touch the appdata or desktop file and just use upstream, sometime with just a renae. Since you are upstream maybe a future release can work using upstream metadata.

ja2003.desktop
christianrauch: Sorry, I mean that I created this very desktop file just for flatpak, because of the icon and executable path.

ja2003.appdata.xml
christianrauch: When the flatpak repo is checked out during build time, the files will be available locally, just like the icon or desktop file. How can I reference those files in the XML?

christianrauch:
bot, build com.github.christianrauch.Jahresarbeit-2003

christianrauch:
bot, build com.github.christianrauch.Jahresarbeit-2003

christianrauch:
bot, build com.github.christianrauch.Jahresarbeit-2003

christianrauch:
bot, build com.github.christianrauch.Jahresarbeit-2003

christianrauch:
bot, build com.github.christianrauch.Jahresarbeit-2003

christianrauch:
I added the debug symbols because they might be useful if crashes occur. I am keeping the filesystem layout. I haven't found a way to set the `PATH` environment variable for runtime use, so the launchers etc. will stay flatpak specific.

I am getting error messages like `Docker login for 'ghcr.io' failed with exit code 1` and the buildbot will not queue the build. Can someone else start the build? Other than this, I am fine with the flatpak and it works as expected.

ja2003.appdata.xml
```diff
@@ -0,0 +1,47 @@
...
+
+  <url type="homepage">https://github.com/christianrauch/Jahresarbeit-2003</url>
+  
+  <project_license>GPL-2.0</project_license>
+  <project_license>OFL-1.1</project_license>
+  <project_license>Apache-2.0</project_license>
```
hfiguiere, line 12: there should only be one of these tags.

ja2003.desktop
hfiguiere: this comment still stand and there are ways to do it properly with a wrapper script.

ja2003.appdata.xml
christianrauch: Ok, in this case I combined them with `AND` as suggested at https://freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-project_license.

ja2003.desktop
christianrauch: I still do not see the advantage of adding yet another file just for changing the path. I will not be able to reuse the *.desktop file anyway because of the clunky icon name `com.github.christianrauch.Jahresarbeit-2003`. The app will work perfectly fine and for a user it does not matter where the file is inside the flatpak.

Am I missing something here? Is there some drawback when starting the executable from the root path? AFAIK, a lot of Steam games work this way and there is no problem with this.

christianrauch: I tried to change the `PATH` using:
```yaml
finish-args:
- --env=PATH=/app
```
and this seems to work for the *.desktop launcher, but it will not be recognised by `command: /app/ja2003`. If there is a way to set `PATH` such that it will work in the launcher and the manifest, I am happy to do so.

christianrauch:
bot, build com.github.christianrauch.Jahresarbeit-2003

com.github.christianrauch.Jahresarbeit-2003.yaml
barthalion: This is actually fine, and Flatpak will not complain about it.

ja2003.appdata.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.github.christianrauch.Jahresarbeit-2003</id>
+  
+  <name>Jahresarbeit 2003</name>
+  <summary>Find die größte Fläche von zusammenhängenden Quadraten</summary>
```
barthalion, line 6: I need to agree with @hfiguiere here. Currently, we expect the default description to be in English, as we are unable to guarantee availability of translations for all applications. You can do something like `<description xml:lang="de">` (the same for summary) and it will be correctly interpreted by native frontends like GNOME Software. I'd also include a comment that this application is available only in German.

ja2003.desktop
barthalion: I'm going to mark it as resolved, because other than presumably bad style, it's actually going to work.

ja2003.desktop
barthalion: I think I wouldn't bother with the desktop file, as long as appdata is in English and mentions the application is only in German.

ja2003.desktop
barthalion: FTR, these can be re-used by any Linux distro. Files can be easily modified with `desktop-file-edit` to set proper values. This is not a blocker, though.

ja2003.appdata.xml
```diff
@@ -0,0 +1,45 @@
...
+      <caption>Hauptmenü</caption>
+      <image>https://res.cloudinary.com/canonical/image/fetch/f_auto,q_auto,fl_sanitize,w_800,h_600/https://dashboard.snapcraft.io/site_media/appmedia/2019/07/menu.png</image>
+    </screenshot>
+    <screenshot>
+      <caption>Spiel</caption>
+      <image>https://res.cloudinary.com/canonical/image/fetch/f_auto,q_auto,fl_sanitize,w_800,h_600/https://dashboard.snapcraft.io/site_media/appmedia/2019/07/game.png</image>
+    </screenshot>
+    <screenshot>
+      <caption>Spiel Beschreibung</caption>
+      <image>https://res.cloudinary.com/canonical/image/fetch/f_auto,q_auto,fl_sanitize,w_800,h_600/https://dashboard.snapcraft.io/site_media/appmedia/2019/07/info.png</image>
```
barthalion, line 42: I think it would be nice to host these screenshots elsewhere (even directly on GitHub) instead.

ja2003.appdata.xml
christianrauch: This works for `<summary xml:lang="de">` and `<summary xml:lang="en">` but `<description xml:lang="de">` gives me the error: `com.github.christianrauch.Jahresarbeit-2003: AppData problem: tag-missing : <description> required`. So, I am just doing the translation for the `summary`.

ja2003.appdata.xml
christianrauch: That means I have to create a new repo or add them to the code repo. I may do this in the future by adding a README with the screenshots to the upstream repo, but for now, I don't see the advantage of this extra work. Hosting them on https://dashboard.snapcraft.io will be as stable as any other cloud service.

ja2003.appdata.xml
barthalion: I believe you can simply skip `xml:lang` for English altogether and leave it without any tag. This should fix the problem for the description (and I'd drop it from the summary as well).

christianrauch: Ok, this seems to work. Only one of the `description` tags must be without the `xml:lang`. I now have `summary` and `description` in both languages.

nedrichards:
bot, build com.github.christianrauch.Jahresarbeit-2003

PR #2707, opened 2021-12-27
Add org.freedesktop.Platform.VulkanLayer.OBSVkCapture

SuNNjek:

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://github.com/nowrep/obs-vkcapture/pull/49#issuecomment-1000823524

----

This is the Vulkan layer needed for [this OBS plugin](https://github.com/nowrep/obs-vkcapture) to work. The OBS plugin needs to be its own flatpak for which I will also create a PR shortly.
Edit: the PR for the OBS plugin is #2708

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,4 +3,4 @@ Vulkan Capture for OBS
 
 Tools to capture Vulkan/OpenGL applications with OBS. For instructions on how to use, see [this page](https://github.com/nowrep/obs-vkcapture#usage).
 
-For this to work you also need the corresponding OBS VkCapture plugin for OBS Studio. This plugin is available as [`com.obsproject.Studio.Plugin.VkCapture` on Flathub](https://github.com/flathub/com.obsproject.Studio.Plugin.VkCapture). 
+For this to work you also need the corresponding OBS VkCapture plugin for OBS Studio. This plugin is available as [`com.obsproject.Studio.Plugin.OBSVkCapture` on Flathub](https://github.com/flathub/com.obsproject.Studio.Plugin.OBSVkCapture).
--- a/README.md
+++ b/README.md
@@ -3,4 +3,4 @@ Vulkan Capture for OBS
 
 Tools to capture Vulkan/OpenGL applications with OBS. For instructions on how to use, see [this page](https://github.com/nowrep/obs-vkcapture#usage).
 
-For this to work you also need the corresponding OBS plugin (should be com.obsproject.Studio.Plugin.VkCapture for the OBS flatpak).
+For this to work you also need the corresponding OBS VkCapture plugin for OBS Studio. This plugin is available as [`com.obsproject.Studio.Plugin.VkCapture` on Flathub](https://github.com/flathub/com.obsproject.Studio.Plugin.VkCapture). 
--- a/README.md
+++ b/README.md
@@ -4,3 +4,21 @@ Vulkan Capture for OBS
 Tools to capture Vulkan/OpenGL applications with OBS. For instructions on how to use, see [this page](https://github.com/nowrep/obs-vkcapture#usage).
 
 For this to work you also need the corresponding OBS VkCapture plugin for OBS Studio. This plugin is available as [`com.obsproject.Studio.Plugin.OBSVkCapture` on Flathub](https://github.com/flathub/com.obsproject.Studio.Plugin.OBSVkCapture).
+
+Permissions
+-----------
+
+The capture tools communicate with OBS via an abstract socket. Therefore it is necessary for both OBS Studio and the captured application to have networking permissions. Luckily OBS Studio already has this, as well as Steam.
+
+To check if the application you're trying to capture has permission to use networking, use the following command:
+```
+flatpak info --show-permissions <APP ID>
+```
+where `<APP ID>` is the ID of the application that you want to capture.
+If your output contains a line looking like `shared=network,...`, then
+networking is already enabled for that application.
+
+If the application you're trying to capture does not currently have networking permissions, you can add them with the following command:
+```
+flatpak override --share=network <APP ID>
+```
--- a/README.md
+++ b/README.md
@@ -1,0 +1,6 @@
+Vulkan Capture for OBS
+======================
+
+Tools to capture Vulkan/OpenGL applications with OBS. For instructions on how to use, see [this page](https://github.com/nowrep/obs-vkcapture#usage).
+
+For this to work you also need the corresponding OBS plugin (should be com.obsproject.Studio.Plugin.VkCapture for the OBS flatpak).
--- a/org.freedesktop.Platform.VulkanLayer.VkCapture.metainfo.xml
+++ b/org.freedesktop.Platform.VulkanLayer.VkCapture.metainfo.xml
@@ -2,8 +2,8 @@
 <component type="addon">
   <id>org.freedesktop.Platform.VulkanLayer.VkCapture</id>
   <extends>org.freedesktop.Platform</extends>
-  <name>Vulkan/OpenGL Capture tools</name>
-  <summary>Vulkan/OpenGL game capture tools to use with OBS</summary>
+  <name>OBS VkCapture tools</name>
+  <summary>Vulkan/OpenGL capture tools to use with the OBS VkCapture plugin</summary>
   <url type="homepage">https://github.com/nowrep/obs-vkcapture</url>
   <url type="bugtracker">https://github.com/nowrep/obs-vkcapture/issues</url>
   <metadata_license>CC0-1.0</metadata_license>
--- a/org.freedesktop.Platform.VulkanLayer.VkCapture.metainfo.xml
+++ b/org.freedesktop.Platform.VulkanLayer.VkCapture.metainfo.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="addon">
+  <id>org.freedesktop.Platform.VulkanLayer.VkCapture</id>
+  <extends>org.freedesktop.Platform</extends>
+  <name>Vulkan/OpenGL Capture tools</name>
+  <summary>Vulkan/OpenGL game capture tools to use with OBS</summary>
+  <url type="homepage">https://github.com/nowrep/obs-vkcapture</url>
+  <url type="bugtracker">https://github.com/nowrep/obs-vkcapture/issues</url>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>GPL-2.0</project_license>
+  <releases>
+    <release version="1.1.1" date="2021-12-26"/>
+  </releases>
+</component>
--- a/org.freedesktop.Platform.VulkanLayer.VkCapture.yml
+++ b/org.freedesktop.Platform.VulkanLayer.VkCapture.yml
@@ -17,6 +17,8 @@ build-options:
 modules:
   - name: VkCapture
     buildsystem: cmake-ninja
+    build-options:
+      libdir: lib/x86_64-linux-gnu
     config-opts: &vkcapture-config-opts
       - -DBUILD_PLUGIN=off
       - -DCMAKE_BUILD_TYPE=Release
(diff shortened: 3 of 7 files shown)


org.freedesktop.Platform.VulkanLayer.VkCapture.yml
```diff
@@ -0,0 +1,48 @@
...
+    config-opts: &vkcapture-config-opts
+      - -DBUILD_PLUGIN=off
+      - -DCMAKE_BUILD_TYPE=Release
+    sources: &vkcapture-sources
+      - type: git
+        url: https://github.com/nowrep/obs-vkcapture.git
+        tag: v1.1.1
+        commit: 8e7bfdbe2248664adb7a8c92805e329408e51f6e
+
+  - name: VkCapture-32bit
```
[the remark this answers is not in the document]
SuNNjek: I tested this with Steam/Proton, which if I recall correctly, needs multiarch for games that don't support 64 bit. Other similar flatpaks such as https://github.com/flathub/org.freedesktop.Platform.VulkanLayer.MangoHud and https://github.com/flathub/org.freedesktop.Platform.VulkanLayer.vkBasalt that are meant to be used with games also have this

GeorgesStavracas:
Will check this and the other one tonight, thanks!

SuNNjek:
Had to change the branch to 21.08 because I just found out that it wouldn't work otherwise. I assume it needs to match the version of the runtime?

org.freedesktop.Platform.VulkanLayer.VkCapture.metainfo.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="addon">
+  <id>org.freedesktop.Platform.VulkanLayer.VkCapture</id>
+  <extends>org.freedesktop.Platform</extends>
+  <name>Vulkan/OpenGL Capture tools</name>
```
GeorgesStavracas, line 5: I think the name should be `OBS VkCapture tools`

org.freedesktop.Platform.VulkanLayer.VkCapture.metainfo.xml
```diff
@@ -0,0 +1,14 @@
...
+  <summary>Vulkan/OpenGL game capture tools to use with OBS</summary>
```
GeorgesStavracas, line 6:
```suggestion
  <summary>Vulkan/OpenGL capture tools to use with the OBS VkCapture plugin</summary>
```

README.md
```diff
@@ -0,0 +1,6 @@
+Vulkan Capture for OBS
+======================
+
+Tools to capture Vulkan/OpenGL applications with OBS. For instructions on how to use, see [this page](https://github.com/nowrep/obs-vkcapture#usage).
+
+For this to work you also need the corresponding OBS plugin (should be com.obsproject.Studio.Plugin.VkCapture for the OBS flatpak).
```
GeorgesStavracas, line 6:
```suggestion
For this to work you also need the corresponding OBS VkCapture plugin for OBS Studio. This plugin is available as [`com.obsproject.Studio.Plugin.VkCapture` on Flathub](https://github.com/flathub/com.obsproject.Studio.Plugin.VkCapture]. 
```

GeorgesStavracas (changes requested):
Some comments, but I'm not entirely sure about my suggestions either. Hope we can find better wording there.

org.freedesktop.Platform.VulkanLayer.VkCapture.metainfo.xml
SuNNjek: Yeah, I think that makes it clearer that it's supposed to be used with the OBS plugin

README.md
SuNNjek: I have committed the changes (don't know if GitHub notifies if I just add a commit, so I'm just gonna leave a comment)

GeorgesStavracas:
Thanks. I took the liberty to ask @gasinvein for an extra pair of eyes. If gasinvein is too busy, we can timeout and land this by the end of the day.

gasinvein:
@GeorgesStavracas  Sorry, I'm completely unfamiliar with OBS Studio and don't even know how to test it. But overall the multiarch build looks valid.

org.freedesktop.Platform.VulkanLayer.VkCapture.yml
```diff
@@ -0,0 +1,50 @@
+id: org.freedesktop.Platform.VulkanLayer.VkCapture
```
gasinvein, line 1: Isn't this ID a bit too generic? Maybe `OBSVkCapture` could be better?

GeorgesStavracas: Good point - now that you mention, I think `OBSVkCapture` is a better namespace for this plugin / Vulkan layer / etc.

SuNNjek: Yeah, hadn't considered that. I have changed it to `org.freedesktop.Platform.VulkanLayer.OBSVkCapture` now

gasinvein:
I see there are components for OpenGL capture, that I assume work by `LD_PRELOAD`'ing the `libobs_glcapture.so` library. It probably won't work OOTB, because the extension libdirs aren't on the `LD_LIBRARY_PATH` inside the sandbox by default. So, you may want to insert the install libdirs into `LD_LIBRARY_PATH` in the `obs-glcapture` script.
If OpenGL capture is out of scope of this extension - just ignore my comment.

README.md
```diff
@@ -0,0 +1,6 @@
+Vulkan Capture for OBS
+======================
+
+Tools to capture Vulkan/OpenGL applications with OBS. For instructions on how to use, see [this page](https://github.com/nowrep/obs-vkcapture#usage).
+
+For this to work you also need the corresponding OBS VkCapture plugin for OBS Studio. This plugin is available as [`com.obsproject.Studio.Plugin.VkCapture` on Flathub](https://github.com/flathub/com.obsproject.Studio.Plugin.VkCapture).
```
GeorgesStavracas, line 6: Need to change these ones to `OBSVkCapture` too

SuNNjek:
@gasinvein Since I'm not the developer of the plugin, I don't know exactly how the OpenGL stuff works unfortunately. Going by the script I would say you're right though. I have only tested it with Vulkan, but I'll see if I can get it to work with a patch.

README.md
SuNNjek: Done

SuNNjek:
@gasinvein I've added a patch now that modifies the OpenGL capture script. I've tested it with the OBS and Steam flatpaks with the game Slime Rancher, which has a native Linux OpenGL build and it works. Without the patch it doesn't work, you were right about that.

patches/obs-glcapture.patch
```diff
@@ -0,0 +1,21 @@
...
+index 2efcf51..aecf1ef 100755
+--- a/src/obs-glcapture
++++ b/src/obs-glcapture
+@@ -1,6 +1,9 @@
+ #!/bin/sh
+ 
+ OBS_GLCAPTURE_LIB="libobs_glcapture.so"
++LIBDIR_BASE="/usr/lib/extensions/vulkan/OBSVkCapture/lib"
++LIBDIR_32="${LIBDIR_BASE}/i386-linux-gnu/"
++LIBDIR_64="${LIBDIR_BASE}/x86_64-linux-gnu/"
```
gasinvein, line 11: It could be a bit cleaner to have a single arch-independent path with `$LIB` in it, e.g.
```
LIBDIR="/usr/lib/extensions/vulkan/OBSVkCapture/\$LIB"
```
(note that the escaping - `\$`)

Although it feels a bit like a hack, it seems to work for e.g. MangoHud.

SuNNjek: Didn't know you could do that but it works, thanks

patches/obs-glcapture.patch
```diff
@@ -0,0 +1,19 @@
...
++LIBDIR="/usr/lib/extensions/vulkan/OBSVkCapture/\$LIB"
+ 
+ if [ "$#" -eq 0 ]; then
+     programname=`basename "$0"`
+@@ -11,4 +12,5 @@ if [ "$#" -eq 0 ]; then
+ fi
+ 
+ LD_PRELOAD="${LD_PRELOAD}:${OBS_GLCAPTURE_LIB}"
+-exec env LD_PRELOAD="${LD_PRELOAD}" "$@"
++LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:${LIBDIR}"
```
gasinvein, line 18: Nitpick: check if the current `LD_LIBRARY_PATH` is empty, append `${LIBDIR}` to the current one if it isn't, replace the current one if it is. E.g.
```bash
if [ -z "${LD_LIBRARY_PATH}" ]; then
  export LD_LIBRARY_PATH="${LIBDIR}"
else
  export LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:${LIBDIR}"
fi
```

SuNNjek: Done

nedrichards:
bot, build org.freedesktop.Platform.VulkanLayer.OBSVkCapture

patches/obs-glcapture.patch
gasinvein: Does it still work? :-)
Not a blocker for the flatpak, just a suggestion: if you're familiar with cmake, consider configuring the `obs-glcapture` script via cmake at build time, putting in it the actual install prefix (instead of hardcoded `/usr/lib/extensions/vulkan/OBSVkCapture`). This way these changes could be submitted upstream.

SuNNjek: Yes, it still works, otherwise I wouldn't have committed it :smile: Not too familiar with CMake unfortunately, but maybe I'll take a look at it sometime and make a PR

gasinvein:
As discussed in the [matrix room](https://matrix.to/#/#flatpak:matrix.org), communication between the vulkan layer and the OBS plugin seems to happen over an abstract socket. AFAIK, for this to work, both ends of the socket need to be in the same network namespace; in case of flatpak, this means that both OBS Studio and the app that's being capturedneed to have `--share=network` permission. 
I expect this plugin to be mainly used with Steam flatpak, where it would work ootb, but still I think this is something worth documenting somewhere.

SuNNjek:
I can put it in the readme file, would that be alright?

gasinvein:
As you see fit. The readme would probably suffice.

README.md
```diff
@@ -3,4 +3,22 @@ Vulkan Capture for OBS
...
+Permissions
+-----------
+
+The capture tools communicate with OBS via an abstract socket. Therefore it is necessary for both OBS Studio and the captured application to have networking permissions. Luckily OBS Studio already has this, as well as Steam.
+
+To check if the application you're trying to capture has permission to use networking, use the following command:
+```
+flatpak info --show-permissions <APP ID>
+```
+where `<APP ID>` is the ID of the application that you want to capture.
```
GeorgesStavracas, line 17: Nitpick: I think you need empty lines between the code block and the paragraph for Markdown not to break

SuNNjek: Do you mean between `flatpak info --show-permissions <APP ID>` and "where ..."? If so, that was intentional, I wanted the code block to stand out.

GeorgesStavracas: I mean this:

```markdown
To check if the application you're trying to capture has permission to use networking, use the following command:

` ` `
flatpak info --show-permissions <APP ID>
` ` `

where `<APP ID>` is the ID of the application that you want to capture.
```

Notice the newlines between the code block, and the paragraphs above and below it

SuNNjek: Yeah, like I said, that was intended. Since it's a code block, there's no need for a newline there, it won't be inlined. Is there a problem with this formatting?

GeorgesStavracas: Alright then, as long as the Markdown is rendered correctly, no strong opinions there

GeorgesStavracas (approved):
LGTM now :+1:

PR #2712, opened 2021-12-28
org.qt_project.qtcreator

christianrauch:

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:**https://github.com/ros-industrial/ros_qtc_plugin

There is already a Qt Creator flatpak [io.qt.QtCreator](https://github.com/flathub/io.qt.QtCreator). However, I am submitting an alternative version with the following benefits:
- won't crash GNOME Shell 3.36 on Wayland
- is much smaller in size (116.5 MB vs 16 GB, i.e. less than 1%)
- builds on arm64
- reuses as much as possible of the runtime and only adds a minimal subset of Qt modules that are required
- installs the development packages during the build for plugin development
- integrates the [ROS Qt Creator plugin](https://github.com/ros-industrial/ros_qtc_plugin)
- I am willing to accept additional PR for third-party plugins

### DIFF
--- a/org.qt_project.qtcreator.yaml
+++ b/org.qt_project.qtcreator.yaml
@@ -1,0 +1,94 @@
+app-id: org.qt_project.qtcreator
+runtime: org.kde.Sdk
+runtime-version: "6.2"
+sdk: org.kde.Sdk
+command: qtcreator
+# Qt Creator already provides desktop launchers and AppData metadata
+# but since we cannot use the original AppID, we have to rename the original files
+rename-desktop-file: org.qt-project.qtcreator.desktop
+rename-appdata-file: org.qt-project.qtcreator.appdata.xml
+rename-icon: QtProject-qtcreator
+finish-args:
+# Qt Creator crashes GNOME Shell 3.36 on Wayland
+# - --socket=wayland
+# - --socket=fallback-x11
+- --socket=x11
+- --share=ipc
+- --allow=devel
+- --share=network
+- --device=dri
+- --talk-name=org.freedesktop.Flatpak # for 'flatpak-spawn'
+- --filesystem=home
+cleanup:
+- /include
+- /lib/arch
+- /lib/cmake
+- /lib/libQt6Core5Compat.prl
+- /lib/libQt6Core5Compat.so # unused symlink
+- /lib/metatypes
+- /lib/qtcreator/objects-RelWithDebInfo
+- /lib/x86_64-linux-gnu
+- /mkspecs
+- /modules
+- /qml
+- /share/cmake
+- /share/pkgconfig
+- "*.a"
+cleanup-commands:
+- mv /app/lib/*-linux-gnu/* /app/lib/
+
+modules:
+# addition Qt modules that are not part of the KDE SDK
+- name: qt5compat
+  sources:
+  - type: git
+    url: git://code.qt.io/qt/qt5compat.git
+    tag: v6.2.1
+  buildsystem: cmake-ninja
+  builddir: true
+  post-install:
+  # move cmake files into a known path because we cannot determine the arch triplet
+  # (i.e. "x86_64-linux-gnu") using shell commands or environment variables
+  - mkdir /app/lib/arch/
+  - mv /app/lib/*-linux-gnu/cmake /app/lib/arch/
+
+- name: QtCreator
+  sources:
+  - type: git
+    url: git://code.qt.io/qt-creator/qt-creator.git
+    tag: v6.0.1
+  buildsystem: cmake-ninja
+  builddir: true
+  config-opts:
+  - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+  - -DQt6Core5Compat_DIR=/app/lib/arch/cmake/Qt6Core5Compat/
+  post-install:
+  # run cmake manually to install the "Devel" component for plugin development
+  - cmake --install . --prefix=/app --component Devel
+  # set release data manually
+  # ideally, this can be extracted via 'git for-each-ref --format="%(creatordate:format:%Y-%m-%d)" "refs/tags/v6.0.1"'
+  - sed -i 's/<release version="6.0.1">/<release date="2021-12-26" version="6.0.1">/g' /app/share/metainfo/org.qt-project.qtcreator.appdata.xml
+
+- name: yaml-cpp
+  sources:
+  - type: git
+    url: https://github.com/jbeder/yaml-cpp.git
+    tag: "yaml-cpp-0.7.0"
+  buildsystem: cmake-ninja
+  builddir: true
+  config-opts:
+  - -DCMAKE_BUILD_TYPE=Release
+  - -DBUILD_SHARED_LIBS=OFF
+  - -DCMAKE_POSITION_INDEPENDENT_CODE=ON
+  - -DYAML_CPP_BUILD_TESTS=OFF
+
+- name: ROS-QtCreator-Plugin
+  sources:
+  - type: git
+    url: https://github.com/ros-industrial/ros_qtc_plugin.git
+    tag: "0.5.0"
+  buildsystem: cmake-ninja
+  builddir: true
+  config-opts:
+  - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+  - -DQt6Core5Compat_DIR=/app/lib/arch/cmake/Qt6Core5Compat/
--- a/org.qt_project.qtcreator.yaml
+++ b/org.qt_project.qtcreator.yaml
@@ -92,3 +92,23 @@ modules:
   config-opts:
   - -DCMAKE_BUILD_TYPE=RelWithDebInfo
   - -DQt6Core5Compat_DIR=/app/lib/arch/cmake/Qt6Core5Compat/
+
+- name: scripts
+  sources:
+  - type: script
+    dest-filename: cmake
+    commands:
+    - flatpak-spawn --host "cmake" "${@}";
+  - type: script
+    dest-filename: catkin
+    commands:
+    - flatpak-spawn --host "bash" "-c" "source /opt/ros/*/setup.bash && catkin ${@}";
+  - type: script
+    dest-filename: colcon
+    commands:
+    - flatpak-spawn --host "bash" "-c" "source /opt/ros/*/setup.bash && colcon ${@}";
+  buildsystem: simple
+  build-commands:
+  - cp * /app/bin/
+#  - install -D cmake /app/bin/cmake
+


christianrauch:
bot, build org.qtproject.qtcreator

hfiguiere:
why not submitting a PR to fix the other package?

christianrauch:
> why not submitting a PR to fix the other package?

Because the scope of both projects is different. The alternative version is supposed to contain "everything", whereas this version only contains a minimal amount of dependencies (see the manifests).

org.qtproject.qtcreator.yaml
```diff
@@ -0,0 +1,85 @@
+app-id: org.qtproject.qtcreator # id "org.qt-project.qtcreator" cannot be used
```
hfiguiere, line 1: the rule would be to use `org.qt_project.qtcreator`

org.qtproject.qtcreator.yaml
```diff
@@ -0,0 +1,85 @@
...
+  - type: git
+    url: git://code.qt.io/qt-creator/qt-creator.git
+    tag: v6.0.1
+  buildsystem: simple
+  build-commands:
+  # run cmake manually to install the "Devel" component for plugin development
+  - cmake -GNinja -B ${FLATPAK_BUILDER_BUILDDIR} -DCMAKE_BUILD_TYPE=RelWithDebInfo -DQt6Core5Compat_DIR=/app/lib/arch/cmake/Qt6Core5Compat/
+  - cmake --build ${FLATPAK_BUILDER_BUILDDIR}
+  - cmake --install ${FLATPAK_BUILDER_BUILDDIR} --prefix=/app
+  - cmake --install ${FLATPAK_BUILDER_BUILDDIR} --prefix=/app --component Devel
```
hfiguiere, line 61: what's wrong with using `buildsystem: cmake-ninja`

christianrauch:
@hfiguiere I reused the upstream AppData but it is missing a timestamp for `<release>`. What is the correct way to add that timestamp information? Is there a command-line tool that can modify the original file?

org.qtproject.qtcreator.yaml
christianrauch: I have to pass `--component Devel` to the last install step. I haven't found an alternative way to do this with flatpak-biulder.

hfiguiere: `make-install-args`

hfiguiere:
> @hfiguiere I reused the upstream AppData but it is missing a timestamp for `<release>`. What is the correct way to add that timestamp information? Is there a command-line tool that can modify the original file?

some people use sed, some use a patch.

org.qtproject.qtcreator.yaml
christianrauch: > `make-install-args`

This would only work for a single install step, right? Passing `--component Devel` will only install the `Devel` component and not the actual program. This is why I have to call the install step twice with different arguments. How would I have the install step called twice with different components?

Also, the documentation says "An array of extra arguments that will be passed to make install". Does this still apply to `cmake --install` and/or `ninja install`?

hfiguiere: you can also run it in the `post-install` phase.

christianrauch:
bot, build org.qt_project.qtcreator

org.qtproject.qtcreator.yaml
christianrauch: If I use `cmake-ninja` with
```yaml
post-install:
  - cmake --install ${FLATPAK_BUILDER_BUILDDIR} --prefix=/app --component Devel
```
then the "normal installation" of the runtime and executables works, but the post-install step fails with only `CMake Error: Error processing file: /run/build/QtCreator/cmake_install.cmake`.

The verbose flatpak-builder output `FB: Running: flatpak build --die-with-parent --env=FLATPAK_BUILDER_BUILDDIR=/run/build/QtCreator [...] /bin/sh -c 'cmake --install ${FLATPAK_BUILDER_BUILDDIR} --prefix=/app --component Devel'` looks normal and verbose make output also does not provide additional information.

Is it possible to halt the build process at this stage, enter a shell and try to debug the command by changing the generated `cmake_install.cmake`?

christianrauch: The same `post-install` step works if I am calling cmake manually in the `simple` build system like before, excluding what was split into the post-install step:
```yaml
buildsystem: simple
build-commands:
- cmake -GNinja -B ${FLATPAK_BUILDER_BUILDDIR} -DCMAKE_BUILD_TYPE=RelWithDebInfo -DQt6Core5Compat_DIR=/app/lib/arch/cmake/Qt6Core5Compat/ -DBUILD_PLUGINS_BY_DEFAULT:BOOL=OFF -DBUILD_EXECUTABLE_BUILDOUTPUTPARSER:BOOL=OFF -DBUILD_EXECUTABLE_CPASTER:BOOL=OFF -DBUILD_EXECUTABLE_CPLUSPLUS-KEYWORDGEN:BOOL=OFF -DBUILD_EXECUTABLE_QML2PUPPET:BOOL=OFF -DBUILD_LIBRARY_AGGREGATION=ON -DBUILD_LIBRARY_EXTENSIONSYSTEM=ON -DBUILD_LIBRARY_UTILS=ON
- cmake --build ${FLATPAK_BUILDER_BUILDDIR}
- cmake --install ${FLATPAK_BUILDER_BUILDDIR} --prefix=/app
post-install:
- cmake --install ${FLATPAK_BUILDER_BUILDDIR} --prefix=/app --component Devel
```
But using `cmake-ninja`:
```yaml
buildsystem: cmake-ninja
builddir: true
config-opts:
- -DQt6Core5Compat_DIR=/app/lib/arch/cmake/Qt6Core5Compat/
[...]
post-install:
- cmake --install ${FLATPAK_BUILDER_BUILDDIR} --prefix=/app --component Devel
```
will fail.

So generally, this `post-install` step would work, but calling cmake manually behaves differently to the `cmake-ninja` procedure.

christianrauch:
bot, build org.qt_project.qtcreator

christianrauch:
bot, build org.qt_project.qtcreator

christianrauch:
@hfiguiere Thanks for your review so far.

The last build works as expected. I haven't found a way to make this work with `cmake-ninja`, so I am going to stick with the `simple` `buildsystem`.

christianrauch:
bot, build org.qt_project.qtcreator

org.qt_project.qtcreator.yaml
```diff
@@ -0,0 +1,91 @@
...
+  sources:
+  - type: git
+    url: git://code.qt.io/qt-creator/qt-creator.git
+    tag: v6.0.1
+  buildsystem: simple
+  build-commands:
+  # run cmake manually to install the "Devel" component for plugin development
+  - cmake -GNinja -B ${FLATPAK_BUILDER_BUILDDIR} -DCMAKE_BUILD_TYPE=RelWithDebInfo -DQt6Core5Compat_DIR=/app/lib/arch/cmake/Qt6Core5Compat/
+  - cmake --build ${FLATPAK_BUILDER_BUILDDIR}
+  - cmake --install ${FLATPAK_BUILDER_BUILDDIR} --prefix=/app
```
barthalion, line 62: Technically speaking, these post-install commands shouldn't interfere with `buildsystem: cmake`.

org.qt_project.qtcreator.yaml
```diff
@@ -0,0 +1,91 @@
...
+cleanup-commands:
+- mv /app/lib/*-linux-gnu/* /app/lib/
+
+modules:
+# addition Qt modules that are not part of the KDE SDK
+- name: qt5compat
+  sources:
+  - type: git
+    url: git://code.qt.io/qt/qt5compat.git
+    tag: v6.2.1
```
barthalion, line 43: This is a nitpick, but the usual convention is that `sources` is the last element of module.

org.qt_project.qtcreator.yaml
barthalion: Looks like it has been discussed above already.

barthalion:
In general, I don't see the point of having the same application maintained twice with small differences. It would be better to simply submit a pull request to https://github.com/flathub/io.qt.QtCreator – if you are going to use it daily, we can make you a co-maintainer there.

org.qt_project.qtcreator.yaml
christianrauch: I didn't know that. My preference is that the name and source of a module are close together because this is the most important information about the software that makes up the flatpak. Everything about the buildsystem is exchangable.

christianrauch:
> In general, I don't see the point of having the same application maintained twice with small differences.

I agree that there is no point in having duplicated flatpaks with the same functionality and scope.

But I think in this case, the scope of the flatpaks is different. I tried to make my arguments above in the opening of the PR. From my point of view, the differences are huge. The `io.qt.QtCreator` flatpak contains much more than just the Qt Creator IDE. It installs another Qt version on top of the runtime and tools like bluez and clang. Hence the large 16GB size of the flatpak. I am using 50GB root partitions and giving away 16GB of this instead of only 116MB to duplicate toolkits and tools is unreasonable and often not possible for me. This becomes even more critical in arm64 environments.

> It would be better to simply submit a pull request to https://github.com/flathub/io.qt.QtCreator – if you are going to use it daily, we can make you a co-maintainer there.

I discussed the huge flatpak size and the Wayland crashing issue with `io.qt.QtCreator` and the opinion there is to [include everything into the flatpak](https://github.com/flathub/io.qt.QtCreator/issues/23#issuecomment-901244203) and to effectively [disable the flatpak on Wayland with GNOME 3.36](https://github.com/flathub/io.qt.QtCreator/issues/48#issuecomment-1002536626). Both are not an option for me and likely also not an option for consumers of this "Qt Creator with ROS plugin" flatpak. Hence I think that the scope of both flatpaks is different.

I am the co-developer of the [ROS Qt Creator plugin](https://github.com/ros-industrial/ros_qtc_plugin). I would like to have a much easier way to distribute the plugin with a compatible Qt Creator version and I think that a flatpak is the perfect choice. However, because of the limitations of `io.qt.QtCreator` and the different scope of these flatpaks, I cannot use it and recommend it to others. Creating a new flatpak that installs fast, does not consume an unreasonable amount of disk space and does not crash my desktop is therefore the only flatpak option I have right now.

There are other options to distribute software (binary installer, ...), but I really would prefer to distribute this as a flatpak as I believe that this is the best way to distribute Linux software.

hfiguiere:
Could the "ROS Qt Creator plugin" just be packagd as a plugin ? would probably require some work on the flatpak, but that's a an option.

christianrauch:
> Could the "ROS Qt Creator plugin" just be packagd as a plugin ?

The plugin could be added to the other flatpak (after installing the Devel component), but this would still leave me with a too large package that would crash GNOME 3.36.

IMHO, there should be a minimal flatpak with just the Qt Creator IDE and the compatible plugins. Everything else (toolkits, libraries, compiler, etc.) should be accessed through the host filesystem.

hfiguiere:
if you use Flatseal https://flathub.org/apps/details/com.github.tchx84.Flatseal you can disable  wayland locally. (or update to a less broken distro, your choice)

As for accessing plugins through the host filesystem, it just defeat the whole thing. Feel free to propose solution to splite QtCreator so that option plugins are optional.

christianrauch:
The plugin should be part of the flatpak because of API incompatibility between major versions.

But other than this, I think that an IDE needs access to the host filesystem to access projects and development dependencies. It's impossible to integrate every possible library and headers into a flatpak.

org.qt_project.qtcreator.yaml
```diff
@@ -0,0 +1,91 @@
...
+- /lib/cmake
+- /lib/libQt6Core5Compat.prl
+- /lib/metatypes
+- /lib/qtcreator/objects-RelWithDebInfo
+- /lib/x86_64-linux-gnu
+- /mkspecs
+- /modules
+- /qml
+- /share/cmake
+- /share/pkgconfig
```
tinywrkb, line 32: IMO it's a bad practice to remove development resources from a development related app.

org.qt_project.qtcreator.yaml
```diff
@@ -0,0 +1,91 @@
...
+- "*.a"
+cleanup-commands:
+- mv /app/lib/*-linux-gnu/* /app/lib/
```
tinywrkb, line 35: This is wrong. Very likely that you're breaking the CMake configs of the Qt module. Look inside `/app/lib/<ARCH>-linux-gnu/cmake/Qt6*/Qt6*Targets*.cmake`  
I suggest instead to create symlinks like this
```
ln -sr ${FLATPAK_DEST}/lib/${FLATPAK_ARCH}-linux-gnu/libQt*.so* -t ${FLATPAK_DEST}/lib/
```

org.qt_project.qtcreator.yaml
```diff
@@ -0,0 +1,91 @@
...
+  # (i.e. "x86_64-linux-gnu") using shell commands or environment variables
+  - mkdir /app/lib/arch/
+  - mv /app/lib/*-linux-gnu/cmake /app/lib/arch/
+
+- name: QtCreator
+  sources:
+  - type: git
+    url: git://code.qt.io/qt-creator/qt-creator.git
+    tag: v6.0.1
+  buildsystem: simple
```
tinywrkb, line 57: You should be able to avoid the simple buildsystem. My pQt6 WebEngine PR](https://github.com/flathub/io.qt.qtwebengine.BaseApp/pull/29) might give you some hints.  
Specifically, look at how I fixed the includes dirs in the `pri` files from `mkspecs` of Qt modules installed in `/app`.

org.qt_project.qtcreator.yaml
christianrauch: This is not an SDK. The dependencies are for building the IDE.

What would be the purpose of bundling only yaml-cpp and the Qt5 compatibility module with an IDE? You can hardly build a useful program out of those.

org.qt_project.qtcreator.yaml
christianrauch: Well, the building worked. So this should not be a problem. But using `FLATPAK_ARCH` is probably the better option to resolve these architecture-dependent paths when setting the Qt5 compatibility module path when building Qt Creator. I will consider this later.

org.qt_project.qtcreator.yaml
christianrauch: Are you referring to the `qt6-qtwebengine` module in your PR's [io.qt.qtwebengine.BaseApp.json](https://github.com/tinywrkb/io.qt.qtwebengine.BaseApp/blob/rt62/io.qt.qtwebengine.BaseApp.json)? I cannot relate this to the problem that I have with the "Devel" component in the install step. The reason for using the `simple` build system is that I cannot control the installed components with `cmake-ninja` since I have to install the default component for the actual IDE and the Devel component for the plugin development. Using the same `post-install` command when with `cmake-ninja` does fail for unknown reasons. I already discussed this in https://github.com/flathub/flathub/pull/2712#discussion_r776100057.

org.qt_project.qtcreator.yaml
christianrauch: This is after the build when all the cmake files are removed. Yes, it will break the cmake script if I would do this during the build, but this does not matter at this point because the cmake script does not exist anymore after the cleanup.

I am moving the library files to `/app/lib` because the linker will not find the so files otherwise and the flatpak will just crash because `qtcreator` cannot find `libQt6Core5Compat.so.6`. I could symlink them but the effect will be the same.

christianrauch:
bot, build org.qt_project.qtcreator

4c0n:
Hi,

It makes me a bit sad to see that you're trying to go this route. There are pretty decent solutions to all of your problems, but you don't seem very receptive to them. For example:

- In order to reduce the size of the io.qt.QtCreator flatpak it's possible to split off the Qt6 (and possibly Qt5) dependencies into extensions, so that they become optional, the reason I haven't done that, is because I feel it should probably be done in collaboration with the KDE team as we could work on them together, instead of maintaining the same things separately
- Access to the host system can be provided by using flatseal to alter the permissions if you must, any libraries you would like to include can be part of your project as well
- Your ROS plugin should most definitely be an extension, but that requires you to make sure it is up-to-date with the latest Qt/Qt Creator version
- The problem of the compositor/window manager crashing on your OS is a problem of the OS and should be resolved there if that version is still supported, as discussed we could potentially produce a warning or prevent the crash, however it's not very desirable

You mention bluez being included as an issue somewhere, this is included because that is a dependency of the qt-connectivity module. clang is a modified version that is required for certain features of Qt Creator to work properly, you can look at the Qt Creator README for more information on that.

Currently there are quite a few users of the io.qt.QtCreator flatpak already, so your opinion on it being unusable in its current state is in fact your opinion. See: https://klausenbusk.github.io/flathub-stats/#ref=io.qt.QtCreator&interval=infinity&downloadType=installs%2Bupdates

Time is an issue for me as I also have a job (and a personal life) that is not related to any of this, so I could use some help achieving these things. Especially considering the massive build times (which will grow even more out of control when adding some other features like android support and python support). Please feel free to open a PR or make suggestions that will work to both our advantage, but also respect the fact that they might not always get accepted (with good reason).

christianrauch:
> It makes me a bit sad to see that you're trying to go this route. 

I did not mean to be disruptive with my flatpak proposal here or hurt anyone. All I wanted was a Qt Creator flatpak with the needed plugins that I can use and also recommend to others. The stated limitations of the `io.qt.QtCreator` flatpak (storage space, Wayland) are hard limitations for me as I either cannot install it due to the storage requirements or I cannot use it because I am running a Wayland desktop. This is the reason why I created this "IDE-only" flatpak.

> In order to reduce the size of the io.qt.QtCreator flatpak it's possible to split off the Qt6 (and possibly Qt5) dependencies into extensions, so that they become optional

Back when I stated that the huge size of the flatpak is an issue for me, bundling Qt with Qt Creator was intended (https://github.com/flathub/io.qt.QtCreator/issues/23#issuecomment-901244203) and splitting this out of the flatpak and only keeping Qt Creator was not an option (https://github.com/flathub/io.qt.QtCreator/issues/23#issuecomment-915006062).

If you want to bundle Qt Creator and all of Qt into one "IDE+SDK" flatpak, then the scope is just different from an "IDE" flatpak that just bundles Qt Creator. I don't see an issue with publishing two flatpaks with two different scopes.

If this opinion changed since then, and removing everything except Qt Creator and plugins from `io.qt.QtCreator` is an option now, then of course publishing two flatpaks with the same scope does not make sense.

> Access to the host system can be provided by using flatseal to alter the permissions if you must, any libraries you would like to include can be part of your project as well

This was the idea behind this flatpak. Only Qt Creator and its dependencies are bundled together. Everything else, such as SDKs, build tools or other libraries are supposed to be used from the host via `flatpak-spawn` or via an "SDK" flatpak.

I don't think that installing another application (Flatseal) to manage access to libraries and tools is a nice user experience. The IDE should "just work" as expected out of the box, in the same way the [`org.gnome.Builder`](https://flathub.org/apps/details/org.gnome.Builder) flatpak works.

> Your ROS plugin should most definitely be an extension, but that requires you to make sure it is up-to-date with the latest Qt/Qt Creator version

Since the Qt Creator API plugin changes with every major release and is not backward compatible, plugins depend in specific Qt Creator versions. As far as I can tell, flathub does not support publishing multiple versions of a flatpak. If plugins become extensions, then the release of Qt Creator and the plugins on flathub have to be synchronised manually. If this manual synchronisation can be guaranteed, then plugins can become extensions.

> The problem of the compositor/window manager crashing on your OS is a problem of the OS and should be resolved there if that version is still supported, as discussed we could potentially produce a warning or prevent the crash, however it's not very desirable

Yes, it is an issue with the OS and solving it there would be the best solution. But switching from a stable LTS version to something else is not possible in a lot of applications. A simple workaround is to disable Wayland. I don't see a major disadvantage in doing that. For non-flatpak apps, Qt disables Wayland on GNOME anyway. So the experience will be the same as installing Qt Creator from the official binaries.

> clang is a modified version that is required for certain features of Qt Creator to work properly, you can look at the Qt Creator README for more information on that.

I saw that Qt Creator uses a custom clang version. In this case, clang is primarily a dependency of Qt Creator and not part of the SDK.

> being unusable in its current state is in fact your opinion.

Of course, crashing my desktop and consuming 16GB instead of <200MB is an issue for me. This is the reason why I decided to create a "Qt Creator only" flatpak. However, finding someone that prefers to install a 200MB flatpak over a 16GB flatpak when only the IDE is required, will not be difficult. A smaller flatpak will just be more accessible than a larger one. Not everyone has large disks and a high-speed connection.

> Please feel free to open a PR or make suggestions that will work to both our advantage, but also respect the fact that they might not always get accepted (with good reason).

My suggestion would essentially be to remove the Qt SDK part of the `io.qt.QtCreator` flatpak. If this is an option now, I am happy to send a PR implementing this.

4c0n:
I don't think having to figure out how to configure Qt Creator, after installing it, to actually make it do anything would make it more user friendly and that's what would happen without a SDK. It would also not have any of the bundled examples, which is also not very ideal.
So just removing the SDK does not seem like a great option, hence the suggestion of making it more optional by having it available as extensions (hopefully that would allow to bundle the extensions by default, but allow you to choose to install it without any of the extensions if you wish to do so).
Also doing that in collaboration with KDE would require some convincing, which I already took a shot at, but didn't get very far (https://github.com/flathub/io.qt.QtCreator/issues/40).

If you can help out with the above, that would be great.

christianrauch:
>  So just removing the SDK does not seem like a great option, hence the suggestion of making it more optional by having it available as extensions

By removing the SDK, I mean removing it from the Qt Creator flatpak such that it is not bundled with Qt Creator. The SDK should be optional in my opinion. And there are multiple ways to install Qt, you can use the distribution packages, the official binary installer, or probably a flatpak SDK. Here, the "flatpak SDK" would not be the SDK that you use to build flatpaks and of which you can only use one, but something that can be used by IDE flatpaks to access libraries, headers and tools from one flatpak to another. But I have no idea how this can be implemented. In any case, using these SDKs should be optional. Not every user of Qt Creator writes Qt applications.

I envision that the Qt Creator flatpak should behave like the `org.gnome.Builder`. This does not bundle an SDK and is very moderate in size (125MB download, 350MB installed). Instead it is using the flatpak SDKs to build applications:
Screenshot from 2022-01-26 16-05-53

Implementing this will probably take many more engineering hours than you and me are willing to put into this (and I actually think that the Qt Creator developers should be tasked with this). But for the time being, my suggestion is to remove any SDK from the flatpak such that there is no hard dependency and users can chose on their own which SDK to use and how to install this.

Erick555:
@christianrauch I found it's possible to make qtcreater in flathub 10x smaller with little changes. You may take a look [here](https://github.com/flathub/io.qt.QtCreator/pull/57#issuecomment-1046097406)

marbub:
@christianrauch I removed all the Qt versions from the QtCreator package. It still copies the LLVM package, but I hope we can make the extension automatically downloadable.

PR #2724, opened 2022-01-03
add com.cuperino.qprompt

Cuperino:

"--share=ipc", is used to enable OpenGL accelerated graphics, which are essential to QPrompt
"--share=network" is used in the About page, to download contributor avatars.
"--device=all" is not used at the time but it will be used in an update to enable receiving inputs from USB knob controllers (such as Contour's ShuttleXpress), and video game controllers.
- [X] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. (I am the project's author)
- [X] I own the domain used in the application ID or the domain has a policy for delegating subdomains (e.g. GitHub, SourceForge). (I own the domain name.)


com.cuperino.qprompt.json
```diff
@@ -0,0 +1,40 @@
...
+               "type": "git",
+               "url": "https://github.com/Cuperino/QPrompt.git",
+               "tag": "v1.0.0",
+               "commit": "c886f9b692dc8900af69f01b6af26f19f220090f"
+            }
+         ]
+      }
+   ],
+   "runtime": "org.kde.Platform",
+   "runtime-version": "5.15",
```
barthalion, line 38:
```suggestion
   "runtime-version": "5.15-21.08",
```

com.cuperino.qprompt.json
```diff
@@ -0,0 +1,40 @@
+{
+   "command": "qprompt",
+   "finish-args": [
+      "--share=ipc",
+      "--socket=x11",
```
barthalion, line 5:
```suggestion
      "--socket=fallback-x11",
```

com.cuperino.qprompt.json
```diff
@@ -0,0 +1,40 @@
...
+      "--socket=wayland",
+      "--filesystem=host",
+      "--share=network",
```
barthalion, line 8: Does it really need both host and network access?

Cuperino: --filesystem=host is necessary because users need to be able to read and write documents with QPrompt. These documents can be in plain text, HTML, and markdown formats. Future versions will also include read only support for PDF.

--share=network is only necessary to be able to load avatars from the about page's list of contributors. Users could disable this functionality without any implications. Nevertheless, I'd rather have networking on because, in some systems, failing to download the avatar results in QPrompt crashing. This is likely an upstream bug with KDE's Kirigami framework.

com.cuperino.qprompt.json
Cuperino: I don't know why, but QPrompt silently fails to start when use the 5.15-21.08 runtime instead of just 5.15...

com.cuperino.qprompt.json
Cuperino: Removing --share=network does not crash QPrompt while using the 5.15 runtime. I think it can be safely removed.

com.cuperino.qprompt.json
```diff
@@ -0,0 +1,40 @@
...
+               "type": "git",
+               "url": "https://github.com/Cuperino/QPrompt.git",
+               "tag": "v1.0.0",
+               "commit": "c886f9b692dc8900af69f01b6af26f19f220090f"
+            }
+         ]
+      }
+   ],
+   "runtime": "org.kde.Platform",
+   "runtime-version": "5.15",
```
hfiguiere, line 38: this whol section is usually at the top of the file

com.cuperino.qprompt.json
```diff
@@ -0,0 +1,40 @@
...
+   "command": "qprompt",
+   "finish-args": [
+      "--share=ipc",
+      "--socket=fallback-x11",
+      "--socket=wayland",
+      "--filesystem=host",
+      "--share=network",
+      "--device=all"
+   ],
+   "id": "com.cuperino.qprompt",
```
hfiguiere, line 11: id is usually the first in the manifest.

com.cuperino.qprompt.json
```diff
@@ -0,0 +1,40 @@
...
+      "/share/gtk-doc",
+      "/share/man",
+      "/share/pkgconfig",
+      "*.la",
+      "*.a"
+   ],
+   "modules": [
+      {
+         "buildsystem": "cmake-ninja",
+         "name": "qprompt",
```
hfiguiere, line 26: name is usually first.

com.cuperino.qprompt.json
hfiguiere: you mean it's crashing on startup? if you run the flatpak with `--command=sh` and `--devel` you can start the executable in gdb and figure out a stack trace.

hfiguiere (changes requested):

com.cuperino.qprompt.json
Cuperino: Moved id to the start, following convention.

com.cuperino.qprompt.json
Cuperino: Moved name up, following convention.

com.cuperino.qprompt.json
Cuperino: Moved section to the top, right below id.

com.cuperino.qprompt.json
Cuperino: This is strange... When I get a shell inside the application's sandbox `flatpak --command=sh --devel run com.cuperino.qprompt` and run QPrompt I can see that it clearly is crashing with a segmentation fault. But when I get a shell inside an application sandbox without having to install it `flatpak-builder --run ./build com.cuperino.qprompt.json sh`, QPrompt runs almost perfectly and only icons are missing. I hypothesize these might be separate issues and the crash may be related to permissions. Or this may be related to [this CMake install bug which I thought was specific to Windows](https://bugs.kde.org/show_bug.cgi?id=445248), but I doubt it.

I haven't been able to run the debugger with the first approach because the build always lacks debug symbols irregardless of what I try. I've tried setting `"config-opts": "--DCMAKE_BUILD_TYPE=RelWithDebInfo",` inside the module in the manifest, and defining `CMAKE_BUILD_TYPE=Debug` before performing the build.

Could someone point me to the correct way of building the Flatpak with debug symbols?

hfiguiere: `flatpak-builder --run` as less syscall filtering than `flatpak run`

And `--devel` still grant a few more than without.

`RelWithDebInfo` build with debug symbols, but there is a `.Debug` package that you need to install.

com.cuperino.qprompt.json
barthalion: Ideally, it should use a portal, which would let you drop any filesystem permissions. See https://docs.flatpak.org/en/latest/portals-qt.html

nedrichards:
bot, build com.cuperino.qprompt

com.cuperino.qprompt.json
Cuperino: @barthalion Thanks for letting me know. It helped me realize the portals where already in use, thus filesystem=host permissions are not necessary.

I also removed network permissions because the ability to download avatar pictures is non-essential to QPrompt, and its implementation has to be changed to use portals upstream, at the Kirigami framework.

Cuperino:
Updated the manifest file to a newer QPrompt commit, which adds the AppData that was missing to complete the build.

Cuperino:
bot, build com.cuperino.qprompt

PR #2745, opened 2022-01-08
Add net.gaijin.WarThunder

Doomsdayrs:
### Criteria

- [x] I am using only the minimal set of permissions. (If not, please explain each non-standard permission.)
- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://forum.warthunder.com/index.php?/topic/544815-permisson-to-add-to-flathub/
I do not own the domain, Gaijin Entertainment owns the game.
- [x] Any additional patches or files have been submitted to the upstream projects concerned. (If not, explain why.)
The game works fine without any patches.

#### Extra Permissions:

* `--device=all`: To allow WarThunder access to controllers

#### Current Issues

- GTK failing to load modules
```
Gtk-Message: 01:12:55.662: Failed to load module "xapp-gtk3-module"
Gtk-Message: 01:12:55.720: Failed to load module "canberra-gtk-module"
Gtk-Message: 01:12:55.720: Failed to load module "pk-gtk-module"
Gtk-Message: 01:12:55.720: Failed to load module "canberra-gtk-module"
Gtk-Message: 01:12:55.720: Failed to load module "pk-gtk-module"
```
This does not effect game play, but it would be nice to rid of these outputs.
 
- ALSA not being able to open /dev/dsp
```
ALSA lib ../../oss/pcm_oss.c:397:(_snd_pcm_oss_open) Cannot open device /dev/dsp
```

This does not effect gameplay. It is an odd error that does not occur in the non-flatpak version.

- ipc_gid being invalid
```
ALSA lib pcm_direct.c:2003:(snd1_pcm_direct_parse_open_conf) The field ipc_gid must be a valid group (create group audio)
```
This does not effect gameplay. It is another odd audio output.

- Game crashes if you plug in a controller
The game will instantly crash if you plug in a controller. I speculate this is because WT automatically tries to access the controller but is blocked for some reason.
If you plug in the controller before the game starts there will be no issues.


net.gaijin.warthunder.yaml
```diff
@@ -0,0 +1,31 @@
...
+  - --share=network
+#  - --filesystem=~/.config/WarThunder/
+#  - --persist=.config
+
+modules:
+  - name: warthunder
+    buildsystem: simple
+    build-commands: 
+      - mkdir /app/bin/
+      - install -D ./run.sh /app/bin/
```
hfiguiere, line 22:
```suggestion
      - install -D -t ${FLATPAK_DEST}/bin/ ./run.sh 
```

otherwise it doesn't work (dest is a filename unless using -t, and -D will create dirs as needed)

hfiguiere (changes requested):

net.gaijin.warthunder.yaml
Doomsdayrs: -D creates directories

hfiguiere: you confuse with `-d`.

```
 -D                  create all leading components of DEST except the last,
                        or all components of --target-directory,
                        then copy SOURCE to DEST
```

Doomsdayrs: If I am confusing it, Then why does it run perfectly?

hfiguiere: if you remove the mkdir your command will say 
```
install: target '/app/bin/' is not a directory: No such file or directory
```

Doomsdayrs:
Build script runs fine before I added the exporting of the launcher icon.

net.gaijin.WarThunder.yaml
```diff
@@ -0,0 +1,31 @@
+app-id: net.gaijin.WarThunder
+runtime: org.gnome.Platform
+runtime-version: '41'
+sdk: org.gnome.Sdk
+command: run.sh
```
hfiguiere, line 5: I think you are confusing the script to run the application and the script to apply the extra data.

Also I don't think you can use the extra data for the source of the icons. Definitely `/app` is read only.

TheEvilSkeleton: `/app` is only read-only during runtime. It should work during install time and build time.

hfiguiere: what I'm getting it is that run.sh contain things that belong to the `apply_extra`.

There should be `/app/bin/apply_extra` that will extract the extra_data. and `run.sh` listed above is the the script that whill launch the app (which in that case has `/app` in read only)

run.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -e
+
+SRC="/app/extra/wt.tar.gz"
+
+# Unload launcher into data directory
+tar -xv --gzip -f $SRC -C $XDG_DATA_HOME
+
+# Export icon
+mkdir "/app/extra/export/share/icons/128x128/apps/"
+mv "$XDG_DATA_HOME/WarThunder/launcher.ico" "/app/extra/export/share/icons/128x128/apps/net.gaijin.WarThunder.png"
```
hfiguiere, line 10: all of this belong to the apply_extra that is supposed to be run after the download, which happen once.

net.gaijin.WarThunder.yaml
Doomsdayrs: Added apply_extra to extract needed files during install

run.sh
Doomsdayrs: This must occur, as the game launcher uses the directory it is placed in as data storage.

/app/bin is write only

run.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -e
+
+SRC="/app/extra/wt.tar.gz"
+
+# Extract
+tar -xv --gzip -f $SRC -C $XDG_DATA_HOME
```
TheEvilSkeleton, line 6: This ensures that the script doesn't extract at runtime if the `WarThunder` directory exists.
```suggestion
if [[ ! -d "${XDG_DATA_HOME}/WarThunder" ]]; then
	tar -xv --gzip -f $SRC -C $XDG_DATA_HOME
fi
```

Doomsdayrs: Negative, This must exist because the launcher itself could be updated.

TheEvilSkeleton: Oh yikes.

TheEvilSkeleton: @Doomsdayrs try with this patch. I just simulated on my end and it works. You'll notice `"$XDG_DATA_HOME"/sha_check`, a file that holds the current archive's hash. The new script checks the archive's hash, compares with the hash in the file and extracts if they're different.

So it should work with updates now while avoiding redundantly extracting every time the user launches.

net.gaijin.WarThunder.desktop
```diff
@@ -0,0 +1,13 @@
+[Desktop Entry]
+Version=1.0
+Name=War Thunder
+Exec=run.sh
+Terminal=false
+Type=Application
+Icon=net.gaijin.WarThunder
+Categories=GNOME;GTK;Game;
```
barthalion, line 8: `GNOME;GTK`?

net.gaijin.WarThunder.desktop
```diff
@@ -0,0 +1,13 @@
...
+StartupNotify=true
+Name[en_US.UTF-8]=War Thunder
+X-GNOME-FullName[en_US.UTF-8]=War Thunder
+NoDisplay=false
+X-GNOME-UsesNotifications=false
```
barthalion, line 13: Are these actually needed?

net.gaijin.WarThunder.desktop
```diff
@@ -0,0 +1,13 @@
+[Desktop Entry]
+Version=1.0
+Name=War Thunder
+Exec=run.sh
+Terminal=false
```
barthalion, line 5: Isn't false the default here?

net.gaijin.WarThunder.metainfo.xml
```diff
@@ -0,0 +1,65 @@
...
+     <li>Regular content updates including new vehicles, maps, missions and nations.</li>
+     <li>Astonishing graphics, authentic sound effects and beautiful music creating an atmosphere to fully immerse yourself in.</li>
+     <li>Create custom content for War Thunder and share it on War Thunder Live, with the prospect of earning real money through the Revenue Share Partner System!</li>
+    </ul>
+   <p>NOTE: This package is not verified by, affiliated with, or supported by Gaijin.</p>
+   <p>NOTE: This installs the launcher into the user data directory. Expect storage usage of 14GB+</p>
+  </description>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://lh3.google.com/u/0/d/1eBTdcQuiconkJJnC8d768gxIcFl_ai6H=w1920-h1080-iv1</image>
```
barthalion, line 29: Would be nice to use some actual screenshots taken on Linux, but it's not a blocker.

apply_extra.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -e
```
barthalion, line 1: This is all commented-out now. If it's on purpose, please remove it.

run.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -e
+
+SRC="/app/extra/wt.tar.gz"
+
+# Extract
+if [[ ! -f "$XDG_DATA_HOME/sha_check" ]] || [[ "$(sha256sum /app/extra/wt.tar.gz | cut -d' ' -f1)" != "$(cat "$XDG_DATA_HOME"/sha_check 2>/dev/null)" ]]; then
+	sha256sum /app/extra/wt.tar.gz | cut -d' ' -f1 > "$XDG_DATA_HOME"/sha_check
+	tar -xv --gzip -f "$SRC" -C "$XDG_DATA_HOME"
```
barthalion, line 8: Is it because the launcher updates itself?

TheEvilSkeleton: No. It's because the launcher is located at `$XDG_DATA_HOME` due to it being coded by the game developers. We can't have it in /app/extra because the launcher (i.e. executable) lives in the same directory as the configuration files, so we were forced to use this workaround.

net.gaijin.WarThunder.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - install -D -t "${FLATPAK_DEST}/share/applications/" ./net.gaijin.WarThunder.desktop
+      - install -D ./net.gaijin.WarThunder-128.png "/app/share/icons/hicolor/128x128/apps/net.gaijin.WarThunder.png"
+      - install -D -t "${FLATPAK_DEST}/share/metainfo/" ./net.gaijin.WarThunder.metainfo.xml
+    sources:
+      - type: extra-data
+        filename: wt.tar.gz
+        url: https://cdnnow-distr.gaijinent.com/wt_launcher_linux_0.9.4.74.tar.gz
+        sha256: 68d4e26c170091f9ede93730f8d258e72af5f325c256ffa6e44fdb9ad0c7d638
+        # curl -s -L -I "URL" | gawk -v IGNORECASE=1 '/^Content-Length/ { print $2 }'
+        size: 19275822
```
TheEvilSkeleton, line 31: This is a template.
```yaml
        url: https://aws-yup-distr-03.gaijinent.com/wt_launcher_linux_0.9.4.74.tar.gz
        sha256: 68d4e26c170091f9ede93730f8d258e72af5f325c256ffa6e44fdb9ad0c7d638
        # curl -s -L -I "URL" | gawk -v IGNORECASE=1 '/^Content-Length/ { print $2 }'
        size: 19275822
        x-checker-data:
          type: rotating-url
          url: 'https://warthunder.com/en/download/launcherLinux'
          pattern: 'https://cdnnow-distr.gaijinent.com/wt_launcher_linux_([0-9.]+).tar.gz'
```

net.gaijin.WarThunder.desktop
Doomsdayrs: It is how the game itself Categorizes itself.

These do not throw errors, it seems to be valid.

net.gaijin.WarThunder.yaml
Doomsdayrs: For what exactly?

TheEvilSkeleton: It's a template to implement flatpak-external-data-checker for WarThunder. Essentially what that does is check if the launcher has an update, submits an MR and merges an hour later if it wasn't manually merged.

This should in theory prevent the self-updater from complaining as much as possible.

net.gaijin.WarThunder.metainfo.xml
Doomsdayrs: There isn't really much difference between Linux / Windows / MacOS of this game

apply_extra.sh
Doomsdayrs: It has been commented out, might become useful again in the future so I shall leave it there

net.gaijin.WarThunder.yaml
```diff
@@ -30,6 +30,10 @@ modules:
         sha256: 68d4e26c170091f9ede93730f8d258e72af5f325c256ffa6e44fdb9ad0c7d638
         # curl -s -L -I "URL" | gawk -v IGNORECASE=1 '/^Content-Length/ { print $2 }'
         size: 19275822
+        x-checker-data:
+          type: rotating-url
+          url: https://warthunder.com/en/download/launcherLinux
+          pattern: 'https://cdnnow-distr.gaijinent.com/wt_launcher_linux_([0-9.]+).tar.gz'
```
TheEvilSkeleton, line 36: Remove all of this. It randomizes the download server which will make f-e-d-c think it has an update all the time. https://matrix.to/#/!RfXaBjokqHAbzZrgHz:matrix.org/$WrH9ulD92oRe0NEHFEuDQ9_f6kPbppug8zucZiWS-RA?via=matrix.org&via=fedora.im&via=kde.org
```suggestion
```

Doomsdayrs:
bot, build net.gaijin.WarThunder

barthalion:
Let's give upstream some time to respond to the request.

Doomsdayrs:
> Let's give upstream some time to respond to the request.

Of course, time to sitback and sip some tea

TheEvilSkeleton:
If they don't respond, you can try contacting them directly: https://gaijinent.com/en/contacts/

Doomsdayrs:
Screenshot from 2022-01-11 20-11-29

Adding this screenshot to the email.

Doomsdayrs:
Sent the email to Marketing, Let's hope they jump on this free advertisement.

Doomsdayrs:
bot, build net.gaijin.WarThunder

net.gaijin.WarThunder.desktop
barthalion: This is less about validity and more about relevancy. It doesn't make any sense to have it marked as GNOME/GTK.

net.gaijin.WarThunder.yaml
barthalion: @Doomsdayrs please add it to make sure the app is not suddenly broken if that URL disappears.

barthalion (commented):
I'm generally fine with merging it as it-is now, but it needs x-checker-data set to ensure it's installable.

Doomsdayrs:
> I'm generally fine with merging it as it-is now, 

I would like this to not be merged for a bit longer. The WarThunder developers said they were asking about this.

>but it needs x-checker-data set to ensure it's installable.
https://matrix.to/#/!RfXaBjokqHAbzZrgHz:matrix.org/$WrH9ulD92oRe0NEHFEuDQ9_f6kPbppug8zucZiWS-RA?via=matrix.org&via=fedora.im&via=kde.org
It randomizes the download server, so we removed it

Doomsdayrs:
Woops that was not what I wanted to do

net.gaijin.WarThunder.yaml
Doomsdayrs: I will have to wait until they update their launcher to see.

Currently I am betting it will not work.

Doomsdayrs:
I would rather wait for them to decide. Currently the situation has been quiet from Gaijin due to the war.

They have been decent to Linux users for a long time. I rather wait and respect them.

net.gaijin.WarThunder.yaml
Doomsdayrs: Is it allowed to create a script that would run daily on the repo to update a file with the sha256 sum of the latest launcher?

barthalion: I believe it's something flatpak-external-data-checker supports as well.

barthalion:
It's been another two months now.

Doomsdayrs:
bot, build net.gaijin.WarThunder

Doomsdayrs:
> It's been another two months now.

Alright, They are not getting to us. This is good to merge.

PR #2780, opened 2022-01-18
Add manifest for org.kde.kdf

flyingcakes85:

  - `dolphin.appdata.patch` : https://invent.kde.org/system/dolphin/-/merge_requests/325
  - `kdf.appdata.patch` : https://invent.kde.org/utilities/kdf/-/merge_requests/3

<hr>

Notes
Dolphin manifest borrowed from https://github.com/flathub/org.kde.dolphin

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "shared-modules"]
+	path = shared-modules
+	url = https://github.com/flathub/shared-modules.git
--- a/dolphin.appdata.patch
+++ b/dolphin.appdata.patch
@@ -1,0 +1,23 @@
+From d754f8d4f91fe41bdf4237586b2ca66c8409b823 Mon Sep 17 00:00:00 2001
+From: Snehit Sah <snehitsah@protonmail.com>
+Date: Tue, 18 Jan 2022 12:47:08 +0530
+Subject: [PATCH] Add content rating
+
+Signed-off-by: Snehit Sah <snehitsah@protonmail.com>
+---
+ src/org.kde.dolphin.appdata.xml | 1 +
+ 1 file changed, 1 insertion(+)
+
+diff --git a/src/org.kde.dolphin.appdata.xml b/src/org.kde.dolphin.appdata.xml
+index 59a2d59c9..8bc232b74 100644
+--- a/src/org.kde.dolphin.appdata.xml
++++ b/src/org.kde.dolphin.appdata.xml
+@@ -418,4 +418,5 @@
+     <release version="21.08.3" date="2021-11-04"/>
+     <release version="21.08.2" date="2021-10-07"/>
+   </releases>
++  <content_rating type="oars-1.1"/>
+ </component>
+-- 
+2.34.1
+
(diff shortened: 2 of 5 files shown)


flyingcakes85:
bot, build org.kde.kdf

flyingcakes85:
Dolphin (and its dependencies) removed from manifest.

flyingcakes85:
bot, build org.kde.kdf

nedrichards:
@flathub/kde FYI

I wonder about the open in file manager stuff. Without a call to flatpak-spawn (and the necessary permission) it's not going to work that well, but also doing it seamlessly would ideally require some changes in the main codebase. What does the rest of the KDE team think?

tsdgeos:
My opinion is that we need to patch kdf to disable the feature of letting you chose a random file manager when running in flatpak and instead just a QDesktopServices::openUrl call

flyingcakes85:
> My opinion is that we need to patch kdf to disable the feature of letting you chose a random file manager when running in flatpak and instead just a QDesktopServices::openUrl call

I'll take care of KDF source code, and update this pull request with a patch. Till then this PR is on hold.

travier:
CC @flathub/kde

flyingcakes85:
bot, build org.kde.kdf

flyingcakes85:
@travier 

Opening in file manager *still* doesn't work on my machines. The only scenario where I can get it to work is when having Thunar as default file manager and opening a thumb drive. Opening internal drives don't work. And with Dolhpin, nothing opens.

Could someone please test the build from flathubbot in last comment?

nedrichards:
Absolutely does not work on Fedora Silverblue 35 under GNOME.

tsdgeos:
Seems you have found a bug in Qt's portal integration, looping @grulja in

We're doing QDesktopServices::openUrl("file:///"), i.e. open the root folder, but Qt is xdgDesktopPortalOpenFile, trying to open a fd out of / and passing it to the portal OpenFile method

It is my understanding the portal OpenDirectory method would make more sense in this case

travier:
Tested this one and it looks like it needs `--device=all` to figure out the names and mount points, etc.
Apart from that, it worked for me so LGTM. If it does not work well for others then let's merge it as Beta and figure things out progressively?

tsdgeos:
going to beta seems reasonable to me

flyingcakes85:
bot, build org.kde.kdf

grulja:
> Seems you have found a bug in Qt's portal integration, looping @grulja in
> 
> We're doing QDesktopServices::openUrl("file:///"), i.e. open the root folder, but Qt is xdgDesktopPortalOpenFile, trying to open a fd out of / and passing it to the portal OpenFile method
> 
> It is my understanding the portal OpenDirectory method would make more sense in this case

Yes and I think back then when I implemented this method didn't exist. I can try to look into this once I find some time or if anyone wants to do it, I will happily help.

It needs to just check here [1] that the URL is a directory and make a different call.

[1] - https://invent.kde.org/qt/qt/qtbase/-/blob/dev/src/gui/platform/unix/qgenericunixservices.cpp#L301

travier:
I think we should merge this one as Beta as it works. We will fix the remaining issues as we go.

travier:
CC @flathub/kde

tsdgeos:
makes sense

Issue #2736, opened 2022-01-07
builds failing for com.jetbrains.GoLand


patriziobruno:
The build for com.jetbrains.GoLand fails on `mirror-screenshots`.

https://flathub.org/builds/#/builders/35/builds/8806/steps/6/logs/stdio

This is blocking https://github.com/flathub/com.jetbrains.GoLand/pull/30

barthalion:
It is really bizarre. I will investigate on Monday (or over the weekend if time allows).

akurtakov:
org.eclipse.Java fails too https://github.com/flathub/org.eclipse.Java/pull/55

patriziobruno:
@barthalion I'll be happy to help if you can point me in the right direction

patriziobruno:
@barthalion is there any way I can help? Is the failing code public?

barthalion:
Fixed via https://github.com/flathub/org.flatpak.Builder/pull/70.

I want to make something clear, though: there's absolutely no reason why GoLand should have access to the entire session bus. You should really drop it.

PR #2786, opened 2022-01-20
Add io.qt.qdbusviewer

JakobDev:

Another useful tool that is already included in the KDE Sdk.


JakobDev:
bot, build io.qt.qdbusviewer

JakobDev:
Why does the mirroring of the Screenshots fail? The Screenshot URL is valid.

JakobDev:
Maybe the failing mirror of the Screenshots was a bug. Let's try it again.

JakobDev:
bot, build io.qt.qdbusviewer

barthalion:
This is caused by recent change in build infra: https://github.com/flathub/flathub/issues/2736 

I will look into it this week.

barthalion:
bot, build io.qt.qdbusviewer

barthalion:
Have you contacted upstream developers about this submission?

io.qt.qdbusviewer.yml
```diff
@@ -0,0 +1,29 @@
+app-id: io.qt.qdbusviewer
+runtime: org.kde.Sdk
```
barthalion, line 2: Is it on purpose that it uses Sdk as the runtime?

barthalion: Ok, I just realized the app actually comes from the Sdk. It would be better to figure out how to build it from source, so it can be updated separately from Sdk.

JakobDev:
> Have you contacted upstream developers about this submission?

I don't know how what's the best way to do this. And to be honest: The Qt Company didn't want to maintain Qt Creator, which is a way bigger program that has more focus of the Qt Company on Flathub. Qdbusviewer didn't have a .desktop file and is not promoted on their webpage, so I don#T think they are interested in doing this.

> It would be better to figure out how to build it from source, so it can be updated separately from Sdk.

This isn't needed. Qdbusviewer is as part of Qt Tools part of Qt itself. It is released with Qt. There are no releases outside a Qt release. Moreover Qdbusviewer don't get any big changes. It only get's a few bugfixes.

barthalion:
> There are no releases outside a Qt release.

That's the problem here. How do you plan to manage updates of the app if there's no explicit way to determine the available releases? You are bound to whatever is available in the runtime, and you need to remember to periodically rebuild the app even though nothing in the manifest changes. It's just at odds with how Flathub works. Gentoo and Arch seen to build qt-tools separately from Qt and this is what should be done here.

nedrichards:
@barthalion you don't need to rebuild the app to get updates do you? Surely flatpak will be keeping the KDE SDK up to date separately so the only updates here that would be required are targeting a new runtime release, which needs the same QA and remembering as any other app updating. We already have a couple of other apps that provide entry points to SDK tools like this and I'm not against it generally, unless there's something I've missed (there very may well be).

PR #2796, opened 2022-01-23
Add org.gnupg.GPA

nmaggioni:

- [x] I am using only the minimal set of permissions. **Full disk access is required to open arbitrary files.**
- [ ] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub.


nmaggioni:
bot, build org.gnupg.GPA

org.gnupg.GPA.yaml
```diff
@@ -0,0 +1,54 @@
...
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+command: gpa
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  #- --filesystem=~/.gnupg:rw
+  #- --filesystem=xdg-run/gnupg:rw
+  - --filesystem=host
```
hfiguiere, line 12: host remains a broad permission

org.gnupg.GPA.yaml
```diff
@@ -0,0 +1,54 @@
...
+  - --filesystem=/tmp
+
+modules:
+
+  - shared-modules/gtk2/gtk2.json
+
+  - name: gpgme
+    sources:
+      - type: archive
+        url: https://www.gnupg.org/ftp/gcrypt/gpgme/gpgme-1.16.0.tar.bz2
```
hfiguiere, line 22: this is already in the runtime.

org.gnupg.GPA.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - "/lib/*.a"
+      - "/lib/pkgconfig"
+      - "/man"
+      - "/share/aclocal"
+      - "/share/devhelp"
+      - "/share/gtk-doc"
+      - "/share/man"
+      - "/share/pkgconfig"
+    buildsystem: simple
+    sources:
```
hfiguiere, line 40: usually sources are last. it doesn't matter to flatpak-builder but it improves readability

org.gnupg.GPA.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - type: git
+        url: git://git.gnupg.org/gpa.git
+        commit: faaf0b7b72bd3c4ed02490132a693d0692b45bfb
+
+    build-commands:
+      - ./autogen.sh
```
hfiguiere, line 46: why not using the default autotools buildsystem?

org.gnupg.GPA.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - ./configure --prefix=${FLATPAK_DEST}
+      - make
+      - make install
+      - make -k check
+    post-install:
+      - install -Dm644 ${FLATPAK_DEST}/share/pixmaps/gpa.png ${FLATPAK_DEST}/share/icons/hicolor/48x48/apps/org.gnupg.GPA.png
+      - install -Dm644 ${FLATPAK_DEST}/share/applications/gpa.desktop ${FLATPAK_DEST}/share/applications/org.gnupg.GPA.desktop
```
hfiguiere, line 53: you an use `rename-desktop-file` instead

org.gnupg.GPA.yaml
```diff
@@ -0,0 +1,54 @@
...
+        commit: faaf0b7b72bd3c4ed02490132a693d0692b45bfb
+
+    build-commands:
+      - ./autogen.sh
+      - ./configure --prefix=${FLATPAK_DEST}
+      - make
+      - make install
+      - make -k check
+    post-install:
+      - install -Dm644 ${FLATPAK_DEST}/share/pixmaps/gpa.png ${FLATPAK_DEST}/share/icons/hicolor/48x48/apps/org.gnupg.GPA.png
```
hfiguiere, line 52: a 128x128 icon is required (or a SVG)

Also I would just copy without renaming and just `rename-icon` making the `desktop-file-edit` unecessary.

org.gnupg.GPA.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - install -Dm644 ${FLATPAK_DEST}/share/applications/gpa.desktop ${FLATPAK_DEST}/share/applications/org.gnupg.GPA.desktop
+      - desktop-file-edit --set-icon=org.gnupg.GPA ${FLATPAK_DEST}/share/applications/org.gnupg.GPA.desktop
```
hfiguiere, line 54: I don't see the appstream file. is it installed by default with the right name?

org.gnupg.GPA.yaml
```diff
@@ -0,0 +1,54 @@
+app-id: org.gnupg.GPA
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+command: gpa
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
```
hfiguiere, line 9: are you sure? gtk2 does not support wayland.

hfiguiere (changes requested):

org.gnupg.GPA.yaml
nmaggioni: You should be able to open any arbitrary file in GPA to crypt/decrypt/sign it, how else could this be achieved?

org.gnupg.GPA.yaml
nmaggioni: GPA has historically supplied only [a 48x48 icon](https://git.gnupg.org/cgi-bin/gitweb.cgi?p=gpa.git;a=blob;f=gpa.png;h=5d612356a1498ebc40816841fa8d809ffb908b62;hb=HEAD) as far as I remember, and upscaling it would make it even worse.

Choosing a 3rd party icon like [the one from the Papirus theme](https://github.com/PapirusDevelopmentTeam/papirus-icon-theme/blob/49074f0cdecbda3f463f68a8af2e95804093b44e/Papirus/64x64/apps/gpa.svg) could be a better alternative but wouldn't be true to the original project sources.

org.gnupg.GPA.yaml
nmaggioni: Wouldn't that get routed through Xwayland internally anyway?

hfiguiere: no. 

Did you even try it?

nmaggioni:
@hfiguiere Thanks for the super quick review, let me know how you'd want the icon and fs permissions to be handled.

hfiguiere:
yep, can't find properly sized icon.

nmaggioni:
bot, build org.gnupg.GPA

org.gnupg.GPA.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - "/lib/pkgconfig"
+      - "/man"
+      - "/share/aclocal"
+      - "/share/devhelp"
+      - "/share/gtk-doc"
+      - "/share/man"
+      - "/share/pkgconfig"
+    buildsystem: autotools
+    post-install:
+      - install -Dm644 ${FLATPAK_ID}.png ${FLATPAK_DEST}/share/icons/hicolor/128x128/apps/${FLATPAK_ID}.png
```
hfiguiere, line 36: you need to edit the .desktop file to set the icon name. Like it was done before.

nmaggioni:
bot, build org.gnupg.GPA

org.gnupg.GPA.yaml
nmaggioni: I've applied your previous `rename-icon` suggestion, `test -f build/*/share/app-info/icons/flatpak/128x128/org.gnupg.GPA.png` now passes locally.

org.gnupg.GPA.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - "/share/gtk-doc"
+      - "/share/man"
+      - "/share/pkgconfig"
+    buildsystem: autotools
+    post-install:
+      - install -Dm644 ${FLATPAK_ID}.png ${FLATPAK_DEST}/share/icons/hicolor/128x128/apps/gpa.png
+      - install -Dm644 ${FLATPAK_ID}.metainfo.xml ${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.metainfo.xml
+    sources:
+      - type: git
+        url: git://git.gnupg.org/gpa.git
```
barthalion, line 41: Would be probably better to use https://dev.gnupg.org/source/gpa.git as the URL here instead.

org.gnupg.GPA.yaml
barthalion: It's not possible with GTK2 app.

barthalion:
Have you contacted upstream developers about this submission?

nmaggioni:
I wrote to the official mailing list but it bounced, and their bug tracker has disk space issues.
GPA isn't actively released since 2018, there has been very few commits since then, and it's licensed under the GPL; I figured out I'd go ahead and create this PR anyway.

Worst case I can keep my fork out of FlatHub, if that's a problem.

nedrichards:
Both of the above are clear reasons why you can't link to an upstream request!

nedrichards:
bot, build org.gnupg.GPA

nedrichards:
When trying to create a key:

> The GPGME library returned an unexpected
error at gpagenkeysimpleop.c:224. The error was:

>	No pinentry

> This is either an installation problem or a bug in GPA.
GPA will now try to recover from this error.

nmaggioni:
The error still appears even though `pinentry` is now installed and works fine when manually invoked:

```plaintext
$ flatpak run --command=bash --devel org.gnupg.GPA

[📦 org.gnupg.GPA flathub]$ which pinentry
/app/bin/pinentry
[📦 org.gnupg.GPA flathub]$ ls -la /app/bin/pinentry*
lrwxrwxrwx. 1 nmaggioni nmaggioni    14 25 gen 23.40 /app/bin/pinentry -> pinentry-gtk-2
-rwxr-xr-x. 5 nmaggioni nmaggioni 84672  1 gen  1970 /app/bin/pinentry-gtk-2
[📦 org.gnupg.GPA flathub]$ pinentry-gtk-2 
Gtk-Message: 23:41:09.153: Failed to load module "canberra-gtk-module"
Gtk-Message: 23:41:09.154: Failed to load module "pk-gtk-module"
OK Pleased to meet you
CONFIRM
OK
^C
[📦 org.gnupg.GPA flathub]$ 
```

The same happens when building on top of the Gnome platform and runtime which already include the appropriate `pinentry` binaries, thus I suspect that this error is due to an interaction with the host's running `gpg-agent`.
Disabling access to the host's filesystem (commenting [L11-12](https://github.com/flathub/flathub/pull/2796/files#diff-b5f4797b1a9befadf406b1b020867a0acc633cb13f3aeef785df907c442868f9R11)) didn't have any effect on the issue, though.

Does anyone have any clues on how to debug such an integration? Killing and manually re-launching the agent with variations of the `--pinentry-program` flag didn't have any effect, as didn't editing (or creating the whole file, in my case) `~/.gnupg/gpg-agent.conf` to include `pinentry-program /usr/bin/pinentry-[...]` and then running `gpg-connect-agent reloadagent /bye` to reload the config.

Erick555:
Why --filesystem=xdg-run/gnupg:rw is commented out? Is this app supposed to run it's own gpg agent or use the host one?

nmaggioni:
Those two lines are commented out as the reference for minimal permissions in case you only want to show/edit keys and manually crypt/decript text instead of opening files from the main filesystem.

GPA won't run an agent on its own since it's basically just a GUI on top of `gpg` itself.

muelli:
you're building the GTK2 version of Pinentry, no? That's know to not integrate well. The GNOME3 version works much better, generally speaking. Have you tried building that?

nmaggioni:
@muelli Switching to the GNOME SDK and building the gnome3 variant of pinentry seems to only work when manually calling `/app/bin/pinentry-gnome3` from inside the flatpak and allowing unlimited DBus access by setting `--socket=session-bus` in the manifest: when invoking pinentry from GPA it either defaults back to the ncurses frontend (if the flatpak was launched from a terminal) or crashes with a classical "inappropriate ioctl for device" (when launched from a desktop file).

Something is amiss with either DBus or X11, but I can't figure out why or what.
The GTK2 variant seems to be still crashing with an obscure "Input/output error", probably due to the same causes.

hfiguiere:
you'd need to change `command: gpa` to point to the right command.

as for d-bus, you should establish which one are needed.

nmaggioni:
`gpa` itself works fine, it's just its pinentry dependency that throws a fit about only being able to run in a TTY, apparently.

At the moment only operations that need a pin/passphrase are broken, the rest works.

muelli:
the bot is silly. Who is responsible for its behaviour? It would be better to not close issues when there has been some activity. Can anybody reopen?

hfiguiere:
feel free to resubmit. The OP didn't seem to be interested to go further.

nmaggioni:
I currently do not intend to troubleshoot the pinentry/dbus issue further, if anybody else is interested in doing so please feel free to use my code to get started.

Just to reiterate the progress:
> At the moment only operations that need a pin/passphrase are broken, the rest works.

muelli:
yeah, I have the feeling that it's good enough for having the app on flathub.
I would commit to co-maintaining the app and trying to find some spare cycles to make it work better.

org.gnupg.GPA.yaml
```diff
@@ -0,0 +1,66 @@
+app-id: org.gnupg.GPA
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
```
barthalion, line 3: Should be switched to 22.08.

nmaggioni:
Runtime updated and application re-tested. As per the [previous comments](https://github.com/flathub/flathub/pull/2796#issuecomment-1021703969), most of the functionality is there but everything that requires a PIN to be entered is broken. Filesystem access is currently unlimited in order to be able to load/write encrypted files arbitrarily from the GUI, but the manifest contains commented out flags to restrict such access to the GPG daemon only.

If the current state is satisfactory and noboby else wants to chime in and troubleshoot pinentry, this is as good a time as any to merge this.

barthalion:
bot, build org.gnupg.GPA

PR #2816, opened 2022-02-01
add io.github.lxi_tools.lxi-gui

lundmar:


io.github.lxi-tools.yaml
```diff
@@ -0,0 +1,142 @@
+---
+app-id: io.github.lxi-tools
+runtime: org.gnome.Platform
+runtime-version: '41'
+sdk: org.gnome.Sdk
+command: lxi-gui
+finish-args:
+  - "--share=ipc"
+  - "--socket=x11"
```
hfiguiere, line 9:
```suggestion
  - "--socket=fallback-x11"
```

io.github.lxi-tools.yaml
```diff
@@ -0,0 +1,142 @@
...
+        path: lua-makefile-is-shit.patch
+      - type: file
+        path: lua-5.3.pc
+    buildsystem: simple
+    build-commands:
+      - make linux INSTALL_TOP=${FLATPAK_DEST} MYCFLAGS=-fPIC V=5.3 R=5.3.6 install
+      - mkdir -p ${FLATPAK_DEST}/lib/pkgconfig
+      - cp lua-5.3.pc ${FLATPAK_DEST}/lib/pkgconfig
+
+  - name: libxml2
```
hfiguiere, line 30: this is already in the runtime.

io.github.lxi-tools.yaml
```diff
@@ -0,0 +1,142 @@
...
+  - "--device=dri"
+  - "--filesystem=host"
+modules:
+  - name: lua
+    sources:
+      - type: archive
+        url: https://www.lua.org/ftp/lua-5.3.6.tar.gz
+        sha256: fc5fd69bb8736323f026672b1b7235da613d7177e72558893a0bdcd320466d60
+      - type: patch
+        path: lua-makefile-is-shit.patch
```
hfiguiere, line 21: maybe a less offensive patch name?

lua-makefile-is-shit.patch
```diff
@@ -0,0 +1,60 @@
...
+diff --git a/src/luaconf.h b/src/luaconf.h
+index 9eeeea6..2a7deef 100644
+--- a/src/luaconf.h
++++ b/src/luaconf.h
+@@ -200,7 +200,7 @@
+ 
+ #else			/* }{ */
+ 
+-#define LUA_ROOT	"/usr/local/"
++#define LUA_ROOT	"/usr/"
```
hfiguiere, line 57: this is even more likely to be incorrect since the prefix is `/app`

io.github.lxi-tools.yaml
```diff
@@ -0,0 +1,142 @@
+---
+app-id: io.github.lxi-tools
```
hfiguiere, line 2: A github based app-id include the user/org name AND the app name. So it is made of 4 components.

io.github.lxi-tools.yaml
```diff
@@ -0,0 +1,142 @@
...
+runtime-version: '41'
+sdk: org.gnome.Sdk
+command: lxi-gui
+finish-args:
+  - "--share=ipc"
+  - "--socket=x11"
+  - "--socket=wayland"
+  - "--share=network"
+  - "--device=dri"
+  - "--filesystem=host"
```
hfiguiere, line 13: why "host" filesystem access?

hfiguiere (changes requested):

hfiguiere:
bot, build io.github.lxi-tools

hfiguiere:
it can't find the appdata file.

io.github.lxi-tools.yaml
lundmar: Fixed.

io.github.lxi-tools.yaml
lundmar: Fixed - removed libxml2.

io.github.lxi-tools.yaml
lundmar: Ha ha - sorry about that!

When I named the patch I was furious because I had just found that that the lua Makefile STILL does not support building lua as a shared library. That is a problem I ran into more than a decade ago and still lua upstream has not be able to fix this despite the fact that literally every single Linux distribution out there needs to build and distribute lua as a shared library and consequently everyone carries around these stupid hacky lua makefile patches forever and now with the rise of Meson also hacky lua pkg-config files. It's all just very silly.

lundmar: Fixed.

lua-makefile-is-shit.patch
lundmar: Don't worry about that - it is not used. I'm overriding it using the INSTALL_TOP variable. The files end up correctly in FLATPAK_DEST aka /app.

io.github.lxi-tools.yaml
lundmar: Right, I overlooked that. I did actually name it correctly in the appdata.xml file that my application installs. Fixed.

io.github.lxi-tools.yaml
lundmar: Fixed - change to 'home'.

io.github.lxi-tools.yaml
```diff
@@ -0,0 +1,133 @@
...
+command: lxi-gui
+finish-args:
+  - "--share=ipc"
+  - "--socket=fallback-x11"
+  - "--socket=wayland"
+  - "--share=network"
+  - "--device=dri"
+  - "--filesystem=home"
+modules:
+  - name: lua
```
barthalion, line 15: Consider switching to https://github.com/flathub/shared-modules/tree/master/lua5.3.

io.github.lxi-tools.yaml
```diff
@@ -0,0 +1,133 @@
...
+      - cp lua-5.3.pc ${FLATPAK_DEST}/lib/pkgconfig
+
+  - name: libtirpc
+    sources:
+      - type: archive
+        url: https://downloads.sourceforge.net/libtirpc/libtirpc-1.3.2.tar.bz2
+        sha256: e24eb88b8ce7db3b7ca6eb80115dd1284abc5ec32a8deccfed2224fc2532b9fd
+    buildsystem: autotools
+
+  - name: readline
```
barthalion, line 37: It's already in the runtime.

io.github.lxi-tools.yaml
```diff
@@ -0,0 +1,133 @@
...
+      - "--disable-mono"
+
+  - name: liblxi
+    sources:
+      - type: archive
+        url: https://github.com/lxi-tools/liblxi/releases/download/v1.15/liblxi-1.15.tar.xz
+        sha256: c334ee3c4cc668821f356181afa1cd33e7113aef5464b187ac5b842059abb8b4
+    buildsystem: meson
+
+  - name: pango
```
barthalion, line 80: It's already in the runtime.

io.github.lxi-tools.yaml
```diff
@@ -0,0 +1,133 @@
...
+    sources:
+      - type: archive
+        url: https://download.gnome.org/sources/pango/1.50/pango-1.50.3.tar.xz
+        sha256: 4add05edf51c1fb375a1ccde7498914120e23cb280dd7395b1aeb441f1838a4c
+    buildsystem: meson
+
+  - name: gtk4
```
barthalion, line 87: I understand 4.4.1 from the runtime is too old?

barthalion:
bot, build io.github.lxi-tools

io.github.lxi-tools.yaml
```diff
@@ -0,0 +1,133 @@
...
+runtime-version: '41'
+sdk: org.gnome.Sdk
+command: lxi-gui
+finish-args:
+  - "--share=ipc"
+  - "--socket=fallback-x11"
+  - "--socket=wayland"
+  - "--share=network"
+  - "--device=dri"
+  - "--filesystem=home"
```
barthalion, line 13: I can't test it as it doesn't build yet, but if it operates on single files, and could use GtkNativeFileChooser to open and save files, this permission could be dropped.

barthalion:
```
error: 'io.github.lxi-tools.lxi-gui' is not a valid application name: Only last name segment can contain -
```

Should be renamed to `io.github.lxi-tools.lxi_gui' due to backwards compatibility with D-Bus names.

io.github.lxi-tools.yaml
lundmar: Done.

io.github.lxi-tools.yaml
lundmar: Fixed - removed readline.

io.github.lxi-tools.yaml
lundmar: Sorry, can't remove that, I need most recent GTK4 which requires newer pango that what is in your runtime.

io.github.lxi-tools.yaml
lundmar: Correct, my application needs very recent version.

io.github.lxi-tools.yaml
lundmar: My app allows users via GtkFileChooser to save .lua scripts, .csv files, and .png image files and I would like users to be able to save those anywhere in their home - as everyone would.

bilelmoussaoui: GtkFileNativeChooser should allow you to do all of that

lundmar: Yeah, but you see I don't really want that because that means that the user may be presented with a non-gtk4 native file chooser widget. I would much prefer to have the GTK4 file chooser working on the users home, hence the need for 'home' access.

bilelmoussaoui: Sure, the next gnome release at least will have a gtk4 file chooser (thanks to xdg-desktop-portal-gnome) :)

lundmar:
Changed application name as suggested.

io.github.lxi-tools.yaml
```diff
@@ -0,0 +1,113 @@
+---
+app-id: io.github.lxi-tools.lxi_gui
```
bilelmoussaoui, line 2: You forgot to rename the manifest, it should match the app-id

lundmar: Fixed.

bilelmoussaoui:
bot, build io.github.lxi-tools.lxi_gui

barthalion:
Doh, my apologies, I've got it backwards. `io.github.lxi_tools.lxi-gui' is the name that will work.

io.github.lxi-tools.yaml
hfiguiere: well you could just use the shared-modules.

lundmar:
Listen guys, I'm not going to change the perfectly sane application ID of my application just to satisfy some filter forced by flathub because it wants to force my application to work on legacy systems with old DBus installations.

I don't want my application to support legacy systems AND let me remind you that using '-' in any of the application ID elements is officially allowed and works perfectly fine with all distributions not featuring old legacy DBus installations. My application ID is even strongly(!) suggested by Gnomes own guidelines (ref. https://wiki.gnome.org/HowDoI/ChooseApplicationID):

 * "each element must contain one or more of the alphanumeric characters (A-Z, a-z, 0-9) plus underscore ('_') and hyphen ('-') and must not start with a digit"

 * ".. it is very strongly recommended to choose a name based on the global public DNS system."

I was hoping for a better flatpak experience but this is crazy. I can't even hack the app ID just for the purpose of flathub because I am forced to make the changes in my application for DBus communication to work according to the hacked app ID else the flatpak fails at runtime. There is no way I will go with a second grade fake application ID which does not reflect my true DNS domain.

This issue combined with the incredibly cumbersome command-line interfaces of the flatpak tools and the constant switcheroo between yaml and json file formats and the fact that flatpaks are only good for distributing single binary GUI applications has made me realize that flatpak is not the future of distributing Linux applications, and certainly not my project which features both a GUI application and a command-line tool.

I've been through the exact same process snapping my lxi-tools application binaries and it was a lot less involved and less painful than this and it allows my users to use both my GUI app and command-line app.

Sorry. I'm bailing out of this mess.

bilelmoussaoui:
No, you are lookig at outdated wiki pages. The developer guidelines at https://developer.gnome.org/documentation/tutorials/application-id.html definitely suggests avoiding the usage of - even if the application id itself accept that

barthalion:
I understand your frustration, but as for Flathub, we don't really get to decide what Flatpak does or doesn't. The community is tightly-knit, but it's not as with Snapcraft where there's a single store and a single tool to use, and there are more stores like ours. All I can essentially offer is guidance how to make your submission compatible with Flatpak.

io.github.lxi-tools.lxi_gui.appdata.xml
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id>io.github.lxi-tools.lxi_gui.desktop</id>
```
hfiguiere, line 3: This is definitly wrong:

```suggestion
  <id>io.github.lxi-tools.lxi_gui</id>
```

hfiguiere (changes requested):

hfiguiere:
The main problem is that than app-id also happen to needing to be a valid d-bus name, even more since it gets passed to the GtkApplication creation:

https://github.com/lxi-tools/lxi-tools/blob/master/src/lxi_gui-main.c#L53

Then with flatpak it is used to automatically grant permission to your own d-bus name.

`g_application_id_is_valid()` documentation is quite clear that's it is strongly discouraged and @barthalion suggestion https://github.com/flathub/flathub/pull/2816#issuecomment-1026742234 applies here.

io.github.lxi-tools.yaml
nedrichards: In flatpak the file chooser is part of the system and not the application surface area - that's why it's a legitimate security boundary. You should expect that a KDE or Qt file chooser (or anything else!) will show up if that's the base OS that the user has selected, the app author just gets or passes off a file. This reasoning is not compelling to allow the home permission.

nedrichards:
bot, build io.github.lxi_tools.lxi-gui

PR #2822, opened 2022-02-02
org.freedesktop.weston

christianrauch:

This adds [Weston](https://gitlab.freedesktop.org/wayland/weston), the Wayland reference compositor, and its examples. At the moment, it can only be used as a nested compositor.

I would appreciate any help that would make it possible to run it from a VTE. This will require access to hardware and ttys and probably some launchd service, but I don't know which permissions have to be set for this.

### DIFF
--- a/org.freedesktop.weston.yaml
+++ b/org.freedesktop.weston.yaml
@@ -9,6 +9,26 @@ finish-args:
 - --socket=wayland
 - --device=dri
 - --filesystem=/run/user/
+cleanup:
+- /bin/checkmk
+- /bin/libevdev-tweak-device
+- /bin/libinput
+- /bin/mouse-dpi-tool
+- /bin/mtdev-test
+- /bin/touchpad-edge-detector
+- /etc
+- /include
+- /lib/*.a
+- /lib/*.la
+- /lib/cmake
+- /lib/debug
+- /lib/pkgconfig
+- /lib/udev
+- /libexec/libinput
+- /share/libweston-10
+- /share/man
+- /share/pkgconfig
+- /share/zsh
 
 modules:
 - name: mtdev
--- a/org.freedesktop.weston.yaml
+++ b/org.freedesktop.weston.yaml
@@ -3,6 +3,8 @@ runtime: org.freedesktop.Platform
 runtime-version: "21.08"
 sdk: org.freedesktop.Sdk
 command: weston
+rename-desktop-file: weston.desktop
+rename-icon: wayland
 finish-args:
 - --socket=wayland
 - --device=dri
@@ -79,3 +81,21 @@ modules:
   - -Dwcap-decode=false
   - -Dtest-junit-xml=false
   - -Dtest-gl-renderer=false
+  post-install:
+  # launcher
+  - install -Dm644 /app/share/wayland-sessions/weston.desktop /app/share/applications/weston.desktop
+  # SVG icon, "is not a valid icon: Format not recognized"
+  # - install -Dm644 /app/share/weston/wayland.svg /app/share/icons/hicolor/scalable/apps/wayland.svg
+  # PNG icon
+  - install -Dm644 /app/share/weston/wayland.png /app/share/icons/hicolor/128x128/apps/wayland.png
+  # set icon
+  - desktop-file-edit --set-icon=wayland /app/share/applications/weston.desktop
+
+- name: appstream
+  sources:
+  - type: file
+    path: org.freedesktop.weston.appdata.xml
+  buildsystem: simple
+  build-commands:
+  # AppStream meta data
+  - install -Dm644 org.freedesktop.weston.appdata.xml /app/share/metainfo/org.freedesktop.weston.appdata.xml
--- a/org.freedesktop.weston.yaml
+++ b/org.freedesktop.weston.yaml
@@ -99,3 +99,24 @@ modules:
   build-commands:
   # AppStream meta data
   - install -Dm644 org.freedesktop.weston.appdata.xml /app/share/metainfo/org.freedesktop.weston.appdata.xml
+
+- name: weston-seatd
+  sources:
+  - type: script
+    dest-filename: weston-seatd
+    commands:
+    - export SEATD_LOGLEVEL=debug
+    - seatd-launch -- weston
+  buildsystem: simple
+  build-commands:
+  - install -Dm755 weston-seatd /app/bin/weston-seatd
+
+- name: weston-logind
+  sources:
+  - type: script
+    dest-filename: weston-logind
+    commands:
+    - dbus-launch weston
+  buildsystem: simple
+  build-commands:
+  - install -Dm755 weston-logind /app/bin/weston-logind
--- a/org.freedesktop.weston.yaml
+++ b/org.freedesktop.weston.yaml
@@ -1,0 +1,81 @@
+app-id: org.freedesktop.weston
+runtime: org.freedesktop.Platform
+runtime-version: "21.08"
+sdk: org.freedesktop.Sdk
+command: weston
+finish-args:
+- --socket=wayland
+- --device=dri
+- --filesystem=/run/user/
+
+modules:
+- name: mtdev
+  sources:
+  - type: git
+    url: http://bitmath.org/git/mtdev.git
+    tag: "v1.1.6"
+    # fatal: dumb http transport does not support shallow capabilities
+    disable-shallow-clone: true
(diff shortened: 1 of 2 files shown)


christianrauch:
bot, build org.freedesktop.weston

christianrauch:
bot, build org.freedesktop.weston

org.freedesktop.weston.yaml
```diff
@@ -0,0 +1,98 @@
...
+
+- name: launcher
+  sources:
+  - type: file
+    path: weston.xml
+  buildsystem: simple
+  build-commands:
+  # launcher
+  - install -d /app/share/applications/
+  - install -m644 /app/share/wayland-sessions/weston.desktop /app/share/applications/weston.desktop
```
hfiguiere, line 87:
```suggestion
  - install -Dm644 /app/share/wayland-sessions/weston.desktop /app/share/applications/weston.desktop
```

org.freedesktop.weston.yaml
```diff
@@ -0,0 +1,98 @@
...
+  - -Dsystemd=false
+  - -Dpipewire=false
+  - -Dcolor-management-colord=false
+  - -Dlauncher-logind=false
+  - -Dlauncher-libseat=true
+  - -Dshell-fullscreen=false
+  - -Dshell-ivi=false
+  - -Dshell-kiosk=false
+
+- name: launcher
```
hfiguiere, line 79: this module should be merged with the one above. Use `post-install` for the install steps.

christianrauch:
bot, build org.freedesktop.weston

org.freedesktop.weston.yaml
```diff
@@ -0,0 +1,98 @@
...
+  - install -d /app/share/applications/
+  - install -m644 /app/share/wayland-sessions/weston.desktop /app/share/applications/weston.desktop
+  # SVG icon, "is not a valid icon: Format not recognized"
+  # - install -d /app/share/icons/hicolor/scalable/apps/
+  # - install -m644 /app/share/weston/wayland.svg /app/share/icons/hicolor/scalable/apps/wayland.svg
+  # PNG icon
+  - install -d /app/share/icons/hicolor/128x128/apps/
+  - install -m644 /app/share/weston/wayland.png /app/share/icons/hicolor/128x128/apps/wayland.png
+  # set icon
+  - echo "Icon=wayland" >> /app/share/applications/weston.desktop
```
hfiguiere, line 95: use `edit-desktop-file` for this kind of changes.

org.freedesktop.weston.yaml
```diff
@@ -0,0 +1,98 @@
...
+  build-commands:
+  # launcher
+  - install -d /app/share/applications/
+  - install -m644 /app/share/wayland-sessions/weston.desktop /app/share/applications/weston.desktop
+  # SVG icon, "is not a valid icon: Format not recognized"
+  # - install -d /app/share/icons/hicolor/scalable/apps/
+  # - install -m644 /app/share/weston/wayland.svg /app/share/icons/hicolor/scalable/apps/wayland.svg
+  # PNG icon
+  - install -d /app/share/icons/hicolor/128x128/apps/
+  - install -m644 /app/share/weston/wayland.png /app/share/icons/hicolor/128x128/apps/wayland.png
```
hfiguiere, line 93:
```suggestion
  - install -Dm644 /app/share/weston/wayland.png /app/share/icons/hicolor/128x128/apps/wayland.png
```

org.freedesktop.weston.yaml
```diff
@@ -0,0 +1,98 @@
...
+  # set icon
+  - echo "Icon=wayland" >> /app/share/applications/weston.desktop
+  # AppStream meta data
+  - install -d /app/share/metainfo
+  - install -m644 weston.appdata.xml /app/share/metainfo/org.freedesktop.weston.appdata.xml
```
hfiguiere, line 98:
```suggestion
  - install -Dm644 weston.appdata.xml /app/share/metainfo/org.freedesktop.weston.appdata.xml
```

hfiguiere (changes requested):

org.freedesktop.weston.yaml
christianrauch: Will `install` automatically create the path `/app/share/applications/`?

christianrauch:
bot, build org.freedesktop.weston

org.freedesktop.weston.yaml
```diff
@@ -0,0 +1,94 @@
...
+  - -Dlauncher-logind=false
+  - -Dlauncher-libseat=true
+  - -Dshell-fullscreen=false
+  - -Dshell-ivi=false
+  - -Dshell-kiosk=false
+
+- name: launcher
+  sources:
+  - type: file
+    path: weston.appdata.xml
```
hfiguiere, line 82: I'd vote for having this fie named properly from the get go.

christianrauch:
bot, build org.freedesktop.weston

org.freedesktop.weston.yaml
christianrauch: Renamed to `org.freedesktop.weston.appdata.xml`.

christianrauch:
This currently only runs as a nested compositor. It is not possible to start it from a tty / VTE due to insufficient permissions to `/dev/tty0`.

If I try to run this with `all` `device` access:
```
flatpak run --command=sh --device=all org.freedesktop.weston
```
and list the devices inside the sandbox, then only some of the devices have user permissions, e.g.:
```
[📦 org.freedesktop.weston ~]$ ls -al /dev/tty* | grep $USER
crw--w---- 1 christian nfsnobody 4,  2  2. Feb 01:16 /dev/tty2
crw------- 1 christian nfsnobody 4,  3  4. Feb 01:41 /dev/tty3
```
but the device of interest (`/dev/tty0`) has set `nfsnobody` as user.

It also appears as if `--device=all` does not have an effect at all, since running without will show the same permissions.

Can someone tell me how I have to set the sandbox permissions to access `/dev/tty0`?

hfiguiere:
`--device=all` just expose device, it doesn't handle permissions.

Also `/dev/tty0` is probably already in use. I'm not sure what you intention is with the flaptak.

tinywrkb:
Is the DRM backend still enabled? Does it make sense to keep it?

christianrauch:
> `--device=all` just expose device, it doesn't handle permissions.

If this is the case, then I am not sure why I am already seeing the devices (like `/dev/tty0`) without specifying `--device=all`.

> I'm not sure what you intention is with the flaptak.

Well, it's like for any flatpak - packaging software with all its dependencies so that others can use the most recent version without compiling from source and hunting down all dependencies manually.

christianrauch:
> Is the DRM backend still enabled? Does it make sense to keep it?

Yes, it's enabled. The intention was that weston in the flatpak can be used nested in another desktop session or from a tty / VTE. I was hoping I can also get some help here to set the required permissions. If this is not possible at all, then removing the native backends would make sense.

hfiguiere:
> > `--device=all` just expose device, it doesn't handle permissions.
> 
> If this is the case, then I am not sure why I am already seeing the devices (like `/dev/tty0`) without specifying `--device=all`.

See the device doesn't mean you have access to it. It's likely owned by `root`.

> > I'm not sure what you intention is with the flaptak.
> 
> Well, it's like for any flatpak - packaging software with all its dependencies so that others can use the most recent version without compiling from source and hunting down all dependencies manually.

Again I'm not sure. Techincally weston is a Wayland graphics server, it's part of the system. Flatpak is for applications. Application needs the system to provide the graphics server.

christianrauch:
> Techincally weston is a Wayland graphics server, it's part of the system. Flatpak is for applications. Application needs the system to provide the graphics server.

I see what you mean. It's not an application like an "app" on mobile phone operating systems like Android. But running as a nested compositor in a desktop environment could be interpreted as an "app".

But then, flatpak is just a packaging format that runs processes inside a sandbox. Apart from the sandbox permission interface, there is nothing that presents to run arbitrary processes inside the flatpak sandbox, right? You could run services inside a flatpak that only communicate outside the sandbox using the network and otherwise do not have access to the host. This would be useful if the service is compromised due to a bug, in which case an outside attacker would not be able to escape the sandbox.

barthalion:
Flatpak is really designed to be used for standard desktop applications, and CLI/TUI at best. It can be used the way Docker say, but why would it? That's what Docker is for. And you can't completely disable sandboxing to use this flatpak as it were a host package.

PR #2828, opened 2022-02-05
Add org.msweet.htmldoc

michaelrsweet:


org.msweet.htmldoc.yml
```diff
@@ -0,0 +1,36 @@
...
+sdk: org.freedesktop.Sdk
+command: htmldoc
+rename-icon: htmldoc.png
+rename-desktop-file: htmldoc.desktop
+finish-args:
+  # X11 + XShm access
+  - --share=ipc
+  - --socket=x11
+  # Wayland access
+  - --socket=wayland
```
hfiguiere, line 13: fltk doesn't support wayland, does it?

org.msweet.htmldoc.appdata.xml
```diff
@@ -0,0 +1,17 @@
...
+  <supports>
+    <control>pointing</control>
+    <control>keyboard</control>
+    <control>touch</control>
+  </supports>
+  <description>
+    <p>HTMLDOC is a program that reads HTML and Markdown source files or web pages and generates corresponding EPUB, HTML, PostScript, or PDF files with an optional table of contents.</p>
+  </description>
+  <launchable type="desktop-id">org.msweet.htmdoc.desktop</launchable>
+</component>
```
hfiguiere, line 17: file is missing `<content_rating>`, `<releases>`, `<screenshots>`

can we assume to will be part of the next upstream release too?

org.msweet.htmldoc.yml
```diff
@@ -0,0 +1,36 @@
+app-id: org.msweet.htmldoc
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+command: htmldoc
+rename-icon: htmldoc.png
```
hfiguiere, line 6:
```suggestion
rename-icon: htmldoc
```

hfiguiere:
bot, build org.msweet.htmldoc

hfiguiere (changes requested):

org.msweet.htmldoc.yml
hfiguiere: this is why the build failed at that point.

org.msweet.htmldoc.yml
michaelrsweet: 1.4 will, 1.3 doesn't.

org.msweet.htmldoc.appdata.xml
michaelrsweet: I don't post screenshots, are they necessary?

The `content_rating` and `releases` elements were not part of the generated XML from the appdata web form... I guess I can convert the current (markdown) changelog into a releases/release tree, but I'm not sure what to put for the rating since this is just a conversion utility so any rating depends on the HTML files/web pages that are being converted.

org.msweet.htmldoc.yml
michaelrsweet: I assume I should comment out the wayland stuff for now???

michaelrsweet:
@hfiguiere Updated both files as requested.

org.msweet.htmldoc.yml
hfiguiere: that's the idea. if it doesn't need it it shouldn't be.

hfiguiere:
bot, build org.msweet.htmldoc

nedrichards:
You're not actually installing the appdata file anywhere as part of your build process, which is why it's failing. If you're adding it here (and as mentioned above it's better to have it in your main release proper) then you need to declare it and put it in the correct place for the system to process.

org.msweet.htmldoc.appdata.xml
nedrichards: https://github.com/flathub/flathub/wiki/App-Requirements#Appstream shows that you can get the tool that Flathub uses (I see the instructions don't actually tell you to *install* it, but you can get that from flathub). That'll help, and yes screenshots are required for applications that declare themselves to be a desktop-application.

You can answer the OARS questions https://hughsie.github.io/oars/generate.html from the perspective of your app and anything that you provide in it - not what users bring to the table in their own files. So in your case I'd imagine the answer would be 'there's nothing scary here.

https://github.com/flathub/flathub/wiki/AppData-Guidelines has some more tips and best practice.

michaelrsweet:
@nedrichards Given that I'm not using this appdata file for any other platform/distribution method, I'd rather not clutter the main sources with it. It looks like I need to copy and gzip the file to the `.../share/app-info/xmls` directory, but it isn't obvious how I should determine what the prefix should be - are there any example projects that I could look at that do this step in their YAML file?

michaelrsweet:
I should also note that using the flatpak-builder command on Ubuntu 20.04 (what I have handy at the moment) leads to 100% CPU usage to the point that the system is completely unresponsive and I have to do a hard reset to get control back...

hfiguiere:
it should be installed in `${FLATPAK_DEST}/share/metainfo`

also it is not specific to flatpak so that's why we prefer that upstream manages it.

michaelrsweet:
@hfiguiere I have never heard of or needed the appdata file on any platform that HTMLDOC has run on before... So for now it stays in the flathub repo...

I'll see if I can get the file installed in the proper directory...

michaelrsweet:
@hfiguiere OK, I think I have the changes made now.

org.msweet.htmldoc.yml
```diff
@@ -0,0 +1,40 @@
...
+sdk: org.freedesktop.Sdk
+command: htmldoc
+rename-icon: htmldoc
+rename-desktop-file: htmldoc.desktop
+finish-args:
+  # X11 + XShm access
+  - --share=ipc
+  - --socket=x11
+  # Wayland access
+  #- --socket=wayland
```
hfiguiere, line 13:
```suggestion
```

org.msweet.htmldoc.yml
```diff
@@ -0,0 +1,40 @@
...
+    buildsystem: autotools
+    sources:
+      - type: archive
+        url: https://www.fltk.org/pub/fltk/1.3.8/fltk-1.3.8-source.tar.bz2
+        sha256: 1d9832409fc08c908c64ab40ad9fd8b0e8af5d203651eb9690b810f41e2f412d
+  - name: htmldoc
+    buildsystem: autotools
+    post-install:
+      - mkdir -p ${FLATPAK_DEST}/share/metainfo
+      - cp org.msweet.htmldoc.metainfo.xml ${FLATPAK_DEST}/share/metainfo
```
hfiguiere, line 34:
```suggestion
      - install -Dm644 -t ${FLATPAK_DEST}/share/metainfo org.msweet.htmldoc.metainfo.xml 
```

hfiguiere:
BTW you can also trigger the build. So special permissions :-)

bot, build org.msweet.htmldoc

hfiguiere:
At least now it tries to validate it:
```
builddir/files/share/appdata/org.msweet.htmldoc.appdata.xml: FAILED:
• style-invalid         : <caption> cannot end in '.' [The HTMLDOC main window.]
Validation of files failed
```

michaelrsweet:
bot, build org.msweet.htmldoc

hfiguiere:
url can be the URL to the github project. It there is an official website, even better.

it's not "bullshit". It's part of the user experience.

hfiguiere:
apparently it can't find an 128x128 or scalable icon.

michaelrsweet:
@hfiguiere This is getting frustrating - now the build claims I don't have a 128x128 icon, but I *am* installing it (mind you, in a legacy location because I still support older Linux desktops, but it works on all current distros).

I guess I need to move files around with another post-install command, but NONE of this is documented and there is no simple example for me to follow....

hfiguiere:
> @hfiguiere This is getting frustrating - now the build claims I don't have a 128x128 icon, but I _am_ installing it (mind you, in a legacy location because I still support older Linux desktops, but it works on all current distros).

"legacy location" doesn't sound right.

`${FLATPAK_DEST}/share/icons/hicolor/128x128/apps/` is the xdg conformant location.
(`scalable` for the SVG instead of 128x128, and same for other sizes)

> 
> I guess I need to move files around with another post-install command, but NONE of this is documented and there is no simple example for me to follow....

michaelrsweet:
bot, build org.msweet.htmldoc

org.msweet.htmldoc.yml
michaelrsweet: Since FLTK 1.4 will support Wayland, I want to leave this in, commented out.

hfiguiere: but you don't want to "clutter" your repository with standard appstream file. ok.

michaelrsweet: Again, until trying to package HTMLDOC with Flatpak (as a service to users that are only able to install Flatpak applications on shared servers) I had never heard of appstream files. They aren't required for Linux software in general or for any other distribution mechanism, and until that isn't the case I am not adding the file to HTMLDOC itself.

That said, this Wayland stuff is part of the (limited) Flatpak documentation, so I included it thinking it might be necessary. When FLTK 1.4 is released later this year I will need to uncomment that line for Linux systems that support Wayland, and I don't want to have to find that example again. I'll already have enough of a hassle to figure out how to support both X11 and Wayland-based systems... :/

michaelrsweet:
bot, build org.msweet.htmldoc

michaelrsweet:
Seems to be working for me when installing on Ubuntu...

org.msweet.htmldoc.metainfo.xml
```diff
@@ -0,0 +1,59 @@
...
+  <project_license>GPL-2.0-only</project_license>
+  <supports>
+    <control>pointing</control>
+    <control>keyboard</control>
+    <control>touch</control>
+  </supports>
+  <description>
+    <p>HTMLDOC is a program that reads HTML and Markdown source files or web pages and generates corresponding EPUB, HTML, PostScript, or PDF files with an optional table of contents.</p>
+  </description>
+  <launchable type="desktop-id">org.msweet.htmdoc.desktop</launchable>
```
hfiguiere, line 18:
```suggestion
  <launchable type="desktop-id">org.msweet.htmldoc.desktop</launchable>
```

org.msweet.htmldoc.yml
```diff
@@ -0,0 +1,40 @@
...
+    buildsystem: autotools
+    sources:
+      - type: archive
+        url: https://www.fltk.org/pub/fltk/1.3.8/fltk-1.3.8-source.tar.bz2
+        sha256: 1d9832409fc08c908c64ab40ad9fd8b0e8af5d203651eb9690b810f41e2f412d
+  - name: htmldoc
+    buildsystem: autotools
+    post-install:
+      - install -Dm644 -t ${FLATPAK_DEST}/share/metainfo org.msweet.htmldoc.metainfo.xml
+      - install -Dm644 desktop/htmldoc-128.png ${FLATPAK_DEST}/share/app-info/icons/flatpak/128x128/org.msweet.htmldoc.png
```
hfiguiere, line 34:
```suggestion
```

See my other suggestion in the metainfo. It is the typo in the desktop file name that caused `appstream-compose` to not find the icons. (this is when the get copied)

hfiguiere (changes requested):

org.msweet.htmldoc.yml
hfiguiere: Just to add, the icons are in the right place as they are.

michaelrsweet: OK, that does explain things... Thanks!

michaelrsweet:
LGTM (again on Ubuntu)

PR #2832, opened 2022-02-06
com.github.louis77.tuner Update to 1.5.0

louis77:


com.github.louis77.tuner.yml
```diff
@@ -0,0 +1,56 @@
+---
+app-id: com.github.louis77.tuner
+runtime: io.elementary.Platform
+runtime-version: '6.1'
+sdk: io.elementary.Sdk
```
hfiguiere, line 5: this runtime isn't available on Flathub.

com.github.louis77.tuner.yml
```diff
@@ -0,0 +1,56 @@
...
+command: com.github.louis77.tuner
+finish-args:
+- "--share=ipc"
+- "--socket=fallback-x11"
+- "--socket=wayland"
+- "--talk-name=org.gtk.vfs"
```
hfiguiere, line 11:
```suggestion
```

louis77:
bot, build com.github.louis77.tuner

com.github.louis77.tuner.yml
```diff
@@ -0,0 +1,56 @@
...
+- "--talk-name=org.gtk.vfs.*"
+- "--share=network"
+- "--metadata=X-DConf=migrate-path=/com/github/louis77/tuner/"
```
hfiguiere, line 14:
```suggestion
```

You don't need this. It's only for apps what used to be available with older runtimes

com.github.louis77.tuner.yml
```diff
@@ -0,0 +1,56 @@
...
+    url: https://gitlab.gnome.org/GNOME/geocode-glib/-/archive/3.26.2/geocode-glib-3.26.2.tar.gz
+    sha256: 589ed8cf890fb47619ad1062b7117d16104554078b837344496d603d0896ec20
+- name: tuner
+  buildsystem: meson
+  config-opts:
+  - "--buildtype=release"
+  post-install:
+  - install -Dm644 /app/share/icons/hicolor/64x64/apps/${FLATPAK_ID}.svg  -t /app/share/icons/hicolor/128x128/apps/
+  sources:
+  - type: dir
```
hfiguiere, line 55: you need to actually get the source from somewhere.

hfiguiere (changes requested):
It can't build for now:

com.github.louis77.tuner.yml
technosf: Hubert,
Is there a way to get that runtime into the buildbot at buildtime? Can the app be built locally and uploaded to flathub, bypassing the buildbot?

hfiguiere: short answer no.

longer answer, I suggested it be submitted, but the idea didn't get traction.

There is a base app for "Juno" but I'm not sure if that's a recent enough Elementary platform. Otherwise it's my understanding that the Elementary runtime is just the GNOME runtime + a few packages, so that could be approach taken.

nanonyme: How about that "a few packages" be provided through shared-modules to make conversion to what Flathub supports more obvious?

bilelmoussaoui: What I would strongly suggest here is to ask if the elementary developers are fine publishing their runtime here, duplicating the work when we already have too many things on our plate is not ideal.

nanonyme: I strongly disagree. Same runtime should not have multiple sources, that will break.

nanonyme: If this really needs Elementary SDK rather than gnome, it should be in Elementary repo.

hfiguiere:
I don't understand why you opened this PR

This app is already on flathub https://github.com/flathub/com.github.louis77.tuner

and that's where the updates happen.

PR #2840, opened 2022-02-08
org.freedesktop.Sdk.Extension.llvm13

Erick555:

Recently someone asked on matrix flatpak channel about new version so here it is. It is based on existing [LLVM12 extension](https://github.com/flathub/org.freedesktop.Sdk.Extension.llvm12) with update commit added on top in order to preserve history.

Note that LLVM 14 is scheduled for mid-march so within 2 months this extension may be obsolete. I'm not sure if it's better to withhold this and wait for next version or publish it now.


Erick555:
bot, build org.freedesktop.Sdk.Extension.llvm13

barthalion:
I think the main question is whether it's going to be useful for @abique. Otherwise I would wait for LLVM 14 instead.

abique:
> I think the main question is whether it's going to be useful for @abique. Otherwise I would wait for LLVM 14 instead.

We don't count on it, and we have prepared everything we need.
If it is there maybe we could try to use it but, we can try that later as well (llvm >=14).

Don't do it just for us.

Thank you for taking the time to upgrade LLVM, I hope it is not a wasted effort.

That being said, I think it is cool to have the latest compilers ready ;-)

abique:
Actually it is now 10+ times that I rebuilt my Docker image, and I can tell you that building 10 times llvm+clang sucks your time.
Even just for building Bitwig it'd be worth to have clang 13 ready.

abique:
I think I'd use your build of clang-13 if you publish it.
Also let say that in 3 months you get llvm-14, it is possible that it contains bugs which forces certain applications to be built against llvm-13. So it is definitely a good thing to have the option.

abique:
It seems that llvm-12, currently doesn't do multi-lib compilation right?
I had the error that it could not find `#include <cstdint>` and I believe it was for our 32 bit plugin host.

Erick555:
Right it doesn't. I don't thin any extension does and there is no extension of extension concept in flatpak.

PR #2845, opened 2022-02-13
Add org.winehq.Wine

gasinvein:
This flatpak contains Wine with the most useful dependencies.
It is primarily intended as base app for various game launchers, like Lutris, Minigalaxy, Bottles, etc.; but also can be useful as a standalone app for advanced users.
Current plan is to ship only Wine stable channel releases (e.g. 5.0.x, 7.0.x) on separate flatpak branches, unless there is a high demand for other versions.
This is _not_ intended for repackaged Windows apps; those will probably want to bundle a custom Wine version anyway.

Please merge it to `branch/stable-21.08` git branch.

  - `xdg-*` filesystems are used by Wine for default desktop integration
  - Upstream didn't show any interest in shipping a flatpak so far; [forum thread](https://forum.winehq.org/viewtopic.php?p=130824&sid=b8579645b1f2636da3f7869f84afdf0f).
  - Wine is patched to prefer dlls from `WINEDLLPATH` over those shipped with Wine itself; this is the only way I see to implement pluggable extensions with 3rd-party dlls like DXVK, wine-nine-standalone, etc.


gasinvein:
bot, build org.winehq.Wine

gasinvein:
bot, build org.winehq.Wine

Erick555:
> Current plan is to ship only Wine stable channel releases (e.g. 5.0.x, 7.0.x) on separate flatpak branches, unless there is a high demand for other versions.

As wine development moves very fast and stable will quickly miss tons of fixes and features I think Development branch (currently at 7.2) would be very useful and popular. Or is it better to have separate appid for stable & devel?

Would apps be able to use multiple stable versions of wine baseapp? If not then I'm not sure if separate stable branches are worth to accumulate as anything beside latest stable will obviously rot and I don't see any launcher would want to stay on old wine version. Maybe rolling stable/development branch split would be better then?

Erick555:
I don't see wine update affecting launcher itself. There may be regressions for specific apps but then solution is to switch back from system to pinned bundled version from winehq or elsewhere which is usually default behavior anyway. To sum up, apps will likely switch immediately to latest stable when someone sends PR  which sounds like avoidable burden.

Erick555:
I believe it is _proven_ in wine ecosystem already. In distros there is either one wine version or two (stable & devel) in repos and all launchers use the same version. When distro updates wine then all launchers use the same new wine version again. This works because usage of wine from perspective of launcher(s) is stable and requires no change from the launcher. There is no dependency requirement between wine & launcher. Versioning wine may make sense when app could use multiple ones at same time otherwise I don't see any _proven_ benefits of such split.

gasinvein:
My point of avoiding sudden breaking changes still stands, and there is no downside in keeping versioned stable branches. Unless some launcher maintainer asks to add a rolling release branch, I don't see a point in adding it. That's it.

Erick555:
It could stand when you provide example of those _breaking changes_ in past years or decades. Typical launcher can run any stable, development or staging wine version so it's ready for new wine stable release before it even exist thus holding it until someone makes PR is pointless.

Again, launcher could pin particular windows app to any wine release in history which will be installed in app prefix or use wine provided by system (baseapp in flathub). The former is what guarantees stability. Pinning system version additionally which can be only one at time for all apps doesn't do any favor.

Erick555:
In general I think hardcoding one wine version is too limited usecase for apps and users. Baseapp providing wine deps and ldconfig setup + different wine flavors provided as extensions usable simultaneously and interchangeably are much more flexible. Something like you described in https://github.com/flathub/flathub/pull/1830#issuecomment-716423438

barthalion:
While it would be more flexible, there's only one gasinvein doing the maintenance. I would prefer a simpler solution over a comprehensive one that will burn out the maintainer.

@gasinvein I'd go with single stable release branch for now. If someone requests an "oldstable" release (are 5.x even maintained?), then you can decide if it makes sense.

gasinvein:
Ok, let's go with rolling-stable branch for now.
>are 5.x even maintained?

If I'm not mistaken, current stable and old-stable versions are maintained at a time.

bot, build org.winehq.Wine

gasinvein:
Looks like the bot was triggered by message edits, sorry for the noise. Latest commit build is [this](https://github.com/flathub/flathub/pull/2845#issuecomment-1041305783).

PR #2861, opened 2022-02-19
io.github.Qalculate.qalculate-qt

hanna-kn:

### DIFF
--- a/io.github.Qalculate.qalculate-qt.yaml
+++ b/io.github.Qalculate.qalculate-qt.yaml
@@ -1,0 +1,86 @@
+app-id: io.github.Qalculate.qalculate-qt
+runtime: org.kde.Platform
+runtime-version: '5.15'
+sdk: org.kde.Sdk
+command: qalculate-qt
+rename-appdata-file: qalculate-qt.appdata.xml
+rename-desktop-file: qalculate-qt.desktop
+rename-icon: qalculate-qt
+copy-icon: true
+finish-args:
+  - --device=dri
+  # X11 + XShm access
+  - --share=ipc
+  - --socket=fallback-x11
+  # Wayland access
+  - --socket=wayland
+  # Network access
+  # for Update Exchange Rates
+  - --share=network
+  # Filesystem access
+  # Import CSV File
+  # Export CSV File
+  # Save Result Image
+  - --filesystem=xdg-desktop
+  - --filesystem=xdg-documents
+  - --filesystem=xdg-pictures
+  - --filesystem=xdg-public-share
+  # Session Bus access
+cleanup:
+  - /include
+  - /lib/pkgconfig
+  - /share/man
+  - '*.la'
+modules:
+  - shared-modules/intltool/intltool-0.51.json
+
+  - name: mpfr
+    config-opts:
+      - --disable-static
+    sources:
+      - type: archive
+        url: https://ftp.gnu.org/gnu/mpfr/mpfr-4.1.0.tar.xz
+        sha256: 0c98a3f1732ff6ca4ea690552079da9c597872d30e96ec28414ee23c95558a7f
+
+  - name: gnuplot
+    config-opts:
+      - --without-gd
+      - --without-libcerf
+      - --without-lua
+      - --with-readline=gnu
+      - --without-linux-vga
+      - --with-qt
+      - --enable-maintainer-mode
+    sources:
+      - type: archive
+        url: https://downloads.sourceforge.net/gnuplot/gnuplot-5.2.8.tar.gz
+        sha256: 60a6764ccf404a1668c140f11cc1f699290ab70daa1151bb58fed6139a28ac37
+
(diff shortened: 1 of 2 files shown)


io.github.Qalculate.qalculate-qt.yaml
```diff
@@ -0,0 +1,86 @@
+app-id: io.github.Qalculate.qalculate-qt
+runtime: org.kde.Platform
+runtime-version: '5.15'
```
gasinvein, line 3:
```suggestion
runtime-version: '5.15-21.08'
```

io.github.Qalculate.qalculate-qt.yaml
```diff
@@ -0,0 +1,86 @@
...
+  # for Update Exchange Rates
+  - --share=network
+  # Filesystem access
+  # Import CSV File
+  # Export CSV File
+  # Save Result Image
+  - --filesystem=xdg-desktop
+  - --filesystem=xdg-documents
+  - --filesystem=xdg-pictures
+  - --filesystem=xdg-public-share
```
gasinvein, line 27: Since it's a Qt app, it should seamlessly open/save files through portals. Does something not work properly without direct filesystem access?

gasinvein:
bot, build io.github.Qalculate.qalculate-qt

io.github.Qalculate.qalculate-qt.yaml
hanna-kn: Those lines were copied from the flatpak for the GTK GUI (io.github.Qalculate). I tested without those lines and, except for some minor functionality (user import/export of CSV file using mathematical function instead of GUI dialog), it seems to work properly.

gasinvein:
bot, build io.github.Qalculate.qalculate-qt

io.github.Qalculate.qalculate-qt.yaml
gasinvein: The GTK UI app doesn't seem to use portals (probably calls `FileChooser` instead of `FileChooserNative`?).

gasinvein:
The appdata.xml file isn't installed. From a quick look at `qalculate-qt.pro` in the app source, it seems like most of the resources aren't installed when `COMPILE_RESOURCES` is set to `yes`, including appdata.xml. Is it intentional?

io.github.Qalculate.qalculate-qt.yaml
hanna-kn: I have not removed direct file system access. I will attempt to use FileChooserNative in the GTK UI.

hanna-kn:
> The appdata.xml file isn't installed. From a quick look at `qalculate-qt.pro` in the app source, it seems like most of the resources aren't installed when `COMPILE_RESOURCES` is set to `yes`, including appdata.xml. Is it intentional?

Probably (although obviously not in this case). I have now removed COMPILE_RESOURCES=yes from config-opts.

gasinvein:
bot, build io.github.Qalculate.qalculate-qt

qalculate-qt-flatpak.patch
```diff
@@ -0,0 +1,14 @@
...
+index 4c2e95b..2085e45 100644
+--- a/data/qalculate-qt.appdata.xml
++++ b/data/qalculate-qt.appdata.xml
+@@ -63,9 +63,6 @@
+   <kudo>UserDocs</kudo>
+  </kudos>
+  <launchable type="desktop-id">qalculate-qt.desktop</launchable>
+- <provides>
+-  <binary>qalculate-qt</binary>
+- </provides>
```
gasinvein, line 11: Why is this patch needed? I guess it was copied from the GTK UI app, but I'm not sure why it was there either.

hanna-kn: I found this in the log of the GTK GUI flatpak (from before my own involvement and before it was added to flathub):
```
Fix AppData file
    
The provides tag describes the public interfaces this program provides.
The binary tag is a child tag of the provides tag.
It provides the name of a binary installed into a location in PATH.
However, flatpak doesn't export any binaries on the host.
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-provides
```
I have a faint memory (possible wrong) of the tag causing an error when building the flatpak previously, but it is at least not the case anymore.

It would only make my life easier if the patch can safely be removed.

gasinvein: I believe it's ok to have this tag if the metainfo file comes from upstream source and doesn't cause errors, but asked the question in https://matrix.to/#/#flatpak:matrix.org just in case.

TingPing: It is only informational and shouldn't be a problem.

hanna-kn: I've now removed the patch.

nedrichards:
bot, build io.github.Qalculate.qalculate-qt

PR #2865, opened 2022-02-20
Add me.simakis.dropboxignore

sp1thas:


hfiguiere:
There is no .desktop file installed, nor icon.

Also this seems to be a command line script, whic isn't exactly the best candidate for flatpak.

me.simakis.dropboxignore.yml
```diff
@@ -0,0 +1,17 @@
+app-id: me.simakis.dropboxignore
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+command: dropboxignore.sh
+cleanup: src
```
hfiguiere, line 6:
```suggestion
```

me.simakis.dropboxignore.yml
```diff
@@ -0,0 +1,17 @@
...
+
+modules:
+ - name: dropboxingore
+   buildsystem: simple
+   build-commands:
+     - install -D src/bin/dropboxignore.sh /app/bin/dropboxignore.sh
```
hfiguiere, line 12:
```suggestion
     - install -D bin/dropboxignore.sh /app/bin/dropboxignore.sh
```

me.simakis.dropboxignore.yml
```diff
@@ -0,0 +1,17 @@
...
+   sources:
+     - type: git
+       url: https://github.com/sp1thas/dropboxignore.git
+       commit: 2684ace4ec1fc95b0c6ec4102f1d9f21ed295d4c
+       dest: src
```
hfiguiere, line 17:
```suggestion
```

hfiguiere (changes requested):

sp1thas:
> Also this seems to be a command line script, whic isn't exactly the best candidate for flatpak.

Hey @hfiguiere! You are totally right, this is a CLI tool (I think that the desktop launcher won't be applicable). On the occasion of [Cointop](https://flathub.org/apps/details/com.github.miguelmota.Cointop), I thought that flathub could be a good choice for command line interfaces too. If you think this is not the case, feel free to close this PR, otherwise I will commit your suggestions

hfiguiere:
it also has a .desktop and an icon. Which are mandatory.

hfiguiere:
bot, build me.simakis.dropboxignore

me.simakis.dropboxignore.yml
```diff
@@ -0,0 +1,19 @@
...
+
+modules:
+ - name: dropboxingore
+   buildsystem: simple
+   build-commands:
+     - install -D bin/dropboxignore.sh /app/bin/dropboxignore.sh
+     - install -Dm644 icons/64.png /app/share/app-info/icons/flatpak/64x64/me.simakis.dropboxigmore
+     - install -Dm644 icons/128.png /app/share/app-info/icons/flatpak/128x128/me.simakis.dropboxigmore
+     - install -Dm644 icons/256.png /app/share/app-info/icons/flatpak/256x256/me.simakis.dropboxigmore
+     - install -Dm644 icons/512.png /app/share/app-info/icons/flatpak/512x512/me.simakis.dropboxigmore
```
hfiguiere, line 15: you NEVER install in app-info. Also when installing the icon file you keep its extension

Neither the metainfo nor the desktop files are installed.

sp1thas: Almost certain that I fixed the issues.

me.simakis.dropboxignore.desktop
```diff
@@ -0,0 +1,6 @@
+[Desktop Entry]
+Name=dropboxignore
+Exec=/app/bin/dropboxignore.sh
```
hfiguiere, line 3:
```suggestion
Exec=dropboxignore.sh
```

it's never an absolute path

hfiguiere:
bot, build me.simakis.dropboxignore

me.simakis.dropboxignore.metainfo.xml
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
```
nedrichards, line 2: I think this may actually want to declare itself as a console-applicatio rather than desktop-application. https://www.freedesktop.org/software/appstream/docs/sect-Metadata-ConsoleApplication.html there might be a little knock on from that but this will allow software stores to present it more accurately to users.

nedrichards:
bot, build me.simakis.dropboxignore

nedrichards:
This is fine but I think it might need a --filesystem=home by default, or better documentation about adding specific paths otherwise people are going to be sadly confused about how to find their dropbox folders.

barthalion:
I don't think the desktop launcher does anything here? Cointop is TUI so it can get away with `Terminal=true`, but it doesn't work for CLI apps. I think you should drop it, change the appstream file to use `console-application`, and then it will be good to merge.

sp1thas:
@nedrichards good catch. the extra finish arg is added.

@barthalion I totally agree, from my pov, the desktop launcher doesn't make sense for this kind of CLI tools. Just removed. I think that `console-application` is already [there](https://github.com/flathub/flathub/pull/2865/files#diff-16fcc5662598e4940da3209a38b48ab27f7d25a24c5f042114b09036a65d3bacR2), did I miss something?

barthalion:
All good, I just didn't check it.

bot, build me.simakis.dropboxignore

sp1thas:
@barthalion forgot to remove some stuff related to the removed `.desktop` file. Feel free to re-build.

nedrichards:
bot, build me.simakis.dropboxignore

@sp1thas fwiw you can use this command too if there's any more issues to check

sp1thas:
bot, build me.simakis.dropboxignore

sp1thas:
@nedrichards seems it's ready to move forward

PR #2881, opened 2022-03-02
Add Pushupmaster

tronta:

- [xI have read the [App Requirements][reqs] and [App Maintenance][maint] pages.


org.gnome.pushupmaster.json
```diff
@@ -0,0 +1,52 @@
+{
+    "app-id" : "org.gnome.pushupmaster",
```
hfiguiere, line 2: It this an official gnome app?

org.gnome.pushupmaster.json
```diff
@@ -0,0 +1,52 @@
...
+    "runtime" : "org.gnome.Platform",
+    "runtime-version" : "master",
```
hfiguiere, line 4: this version is not available on flathub

org.gnome.pushupmaster.json
```diff
@@ -0,0 +1,52 @@
...
+        "org.freedesktop.Sdk.Extension.rust-stable"
+    ],
+    "command" : "pushupmaster",
+    "finish-args" : [
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--device=dri",
+        "--socket=wayland",
+        "--socket=pulseaudio",
+        "--filesystem=/sys/bus/iio/devices/"
```
hfiguiere, line 16: this is not valid.

org.gnome.pushupmaster.json
```diff
@@ -0,0 +1,52 @@
...
+    ],
+    "build-options" : {
+        "append-path" : "/usr/lib/sdk/rust-stable/bin",
+        "build-args" : [
+            "--share=network"
```
hfiguiere, line 21: this won't work. You need to use flatpak-builder-tools to generate a manifest for the dependencies. Or have all the crates vendored in the tarball.

org.gnome.pushupmaster.json
```diff
@@ -0,0 +1,52 @@
...
+    "modules" : [
+        {
+            "name" : "pushupmaster",
+            "builddir" : true,
+            "buildsystem" : "meson",
+            "sources" : [
+                {
+                    "type" : "git",
+                    "url" : "https://gitlab.gnome.org/peter/pushupmaster.git",
+                    "branch": "main"                }
```
hfiguiere, line 48: you need a revision or tag. Not a branch.

hfiguiere (changes requested):

org.gnome.pushupmaster.json
tronta: ah. ok. I need to check then, if it works with an older version, or I have to wait,

org.gnome.pushupmaster.json
tronta: actually no, but I put it is on the git.gnome repo.
If I should change sth there, please tell me. It's my first app :-)

org.gnome.pushupmaster.json
hfiguiere: it's not about this change. it's about `/sys`

tronta: I need this to access the proximity sensor on the phone. What would be the correct way?

nedrichards: Does your device expose information via https://gitlab.freedesktop.org/hadess/iio-sensor-proxy/ ?

org.gnome.pushupmaster.json
nedrichards: Yup, this isn't GNOME software as released by the GNOME release team. Check out https://circle.gnome.org/ for ways to get your app a 'seal of approval'.

tronta: Thanks for the hint. I wanted to do this anyway. Let me work on the other topics and some additional improvements on the SW and I will go for it.

PR #2886, opened 2022-03-05
Add Chromium BaseApp

refi64:

- [ ] I have read the [App Requirements][reqs] and [App Maintenance][maint] pages.
- [ ] My pull request follows the instructions at [App Submission][submission].

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+.flatpak-builder/
+_build/
--- a/flathub.json
+++ b/flathub.json
@@ -1,0 +1,3 @@
+{
+  "skip-appstream-check": true
+}
--- a/gtk-settings.ini
+++ b/gtk-settings.ini
@@ -1,0 +1,2 @@
+[Settings]
+gtk-print-backends=file,cups
--- a/gtk3-werror.patch
+++ b/gtk3-werror.patch
@@ -1,0 +1,12 @@
+diff --git a/meson.build b/meson.build
+index cbd820da44..0438f6cc7d 100644
+--- a/meson.build
++++ b/meson.build
+@@ -309,7 +309,6 @@ elif cc.get_id() == 'gcc' or cc.get_id() == 'clang'
+     '-Werror=sequence-point',
+     '-Werror=return-type',
+     '-Werror=trigraphs',
+-    '-Werror=array-bounds',
+     '-Werror=write-strings',
+     '-Werror=address',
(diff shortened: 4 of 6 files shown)


Erick555:
Should zypak be there as well?

refi64:
@Erick555 no, the goal is to share this w/ Chromium, which doesn't need Zypak.

Erick555:
Ha, I was sure chromium has zypak but it doesn't, nvm.

BTW, potential problem with baseapp is that it makes more difficult to test & fix issues as you have to merge them first in baseapp then rebuild chromium in order to test the result instead of just spinning chromium build.

Similar setup affects firefox right now where people need to wait till fix makes it to baseapp then wait till ff will be rebuild against new baseapp then find out inital fix was insufficient and the process must be repeated :smile:

refi64:
It's definitely unfortunate, but these modules are going to be split across 6 Flatpaks (Chromium, UCG, Chrome, Chrome Dev [soon], Edge, Brave), so it felt like it was past time to make this a baseapp. IIRC I can still overwrite files in the base, so for quick testing I could copy-paste the module to the individual apps.

nedrichards:
These seem pretty 'restful' dependencies with not super active release cycles but you're sure that these won't cause problems if any of the above apps get out of sync with the main upstream? (I'm always slightly suspicious about baseapps because 'specifying all your dependencies is the point' but I do understand their utility)

refi64:
> but you're sure that these won't cause problems if any of the above apps get out of sync with the main upstream

Almost everything in the Chrome family is dynamically loaded at runtime for compat reasons, so if any dependencies had a major release change or similar, Chrome would likely support both for a *very* long time.

In addition, all the apps this would be used on follow a release cycle very close to upstream, with maybe a 2mo discrepancy between Chrome Dev and Chrome stable being the largest gap.

nedrichards:
That's good to know, and yeah I'm aware my feelings could be accurately be described as "old man rages at cloud" so don't consider this a blocker ;-) 

Reviewing more generally, I assume that the chromium world will at some point move to libsecret 0.20> and at that point it'll get an extra data checker interface.

refi64:
> I assume that the chromium world will at some point move to libsecret 0.20> and at that point it'll get an extra data checker interface.

The older version is used because Chromium requires APIs that the portal backend does not support.

Any updates on merging this? :sweat_smile: would like to get the Chrome-specific version up so the Dev Flatpak can be posted soon.

PR #2893, opened 2022-03-07
Add org.freedesktop.Sdk.Extension.avr-gcc

ashahir05:

### DIFF
--- a/flathub.json
+++ b/flathub.json
@@ -1,0 +1,4 @@
+{
+  "skip-icons-check": true,
+  "automerge-flathubbot-prs": true
+}
--- a/org.freedesktop.Sdk.Extension.avr-gcc.appdata.xml
+++ b/org.freedesktop.Sdk.Extension.avr-gcc.appdata.xml
@@ -1,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2022 Lorenz Wildberg <lorenz@wild-fisch.de> -->
+<component type="addon">
+  <extends>org.freedesktop.Sdk</extends>
+  <id>
+    org.freedesktop.Sdk.Extension.avr-gcc
+  </id>
+  <metadata_license>
+    CC0-1.0
+  </metadata_license>
+  <name>
+    AVR-GCC SDK extension
+  </name>
+  <summary>
+    GCC compiler for AVR platforms
+  </summary>
+  <description>
+    <p>
+      GCC compiler for AVR platforms.
+    </p>
+  </description>
+  <project_license>
+    LGPL-2.1-only
+  </project_license>
+  <url type="homepage">
+  </url>
+  <url type="bugtracker">
+  </url>
+  <translation/>
+  <update_contact>
+  </update_contact>
+  <releases>
+    <release version="0.1" date="2022-03-04"/>
+  </releases>
+</component>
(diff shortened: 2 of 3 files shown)


ashahir05:
bot, build org.freedesktop.Sdk.Extension.avr-gcc

ashahir05:
bot, build org.freedesktop.Sdk.Extension.avr-gcc

ashahir05:
bot, build org.freedesktop.Sdk.Extension.avr-gcc

hfiguiere:
what is the intent of this SDK since the avr compiler doesn't target any flatpak supported target?

tinywrkb:
I don't see the point of having just the toolchain.  
If you're packaging an AVR SDK, then ship everything that is needed for AVR development, and it would be better to just name it `org.freedesktop.Sdk.Extension.avr`.   
Specifically, this is missing gdb, avrdude, and if it were me packaging this, then I would add a simavr for simulation.  
In fact, I packaged this [here](https://github.com/tinywrkb/flatpaks/blob/master/org.freedesktop.Sdk.Extension.avr/org.freedesktop.Sdk.Extension.avr.yaml) a while back, and tried to make it host architecture agnostic, but I haven't tested that it builds on aarch64.

@hfiguiere this could be useful for any embedded development or simulation app that targets AVR or Arduino, though AFAIK the Arduino IDE Flathub package currently uses a pre-built toolchain.

ping @A6GibKm

A6GibKm:
Yes, arduino requires a very specific set of dependencies shipped by themselves.

hfiguiere:
> @hfiguiere this could be useful for any embedded development or simulation app that targets AVR or Arduino, though AFAIK the Arduino IDE Flathub package currently uses a pre-built toolchain.

Maybe it's not a SDK extension then but rather a "runtime". Not sure we have an extension point for that though.

tinywrkb:
> Yes, arduino requires a very specific set of dependencies shipped by themselves.

And if I understand it correctly, IDE v2 switches to use arudino-cli, the toolchain will be downloaded during run-time and won't be packaged.  

> > @hfiguiere this could be useful for any embedded development or simulation app that targets AVR or Arduino, though AFAIK the Arduino IDE Flathub package currently uses a pre-built toolchain.
> 
> Maybe it's not a SDK extension then but rather a "runtime". Not sure we have an extension point for that though.

I don't think there is a `subdirectories` extension point for the runtime, which would've been nice to have and avoid deploying from SDK extensions to apps (e.g. Java) or using a base app (e.g. Wine).

So if this is not appropriate as SDK extension, and there are no apps that are expected to make use of this extension, the maybe this should be packaged as an app or a base app?

tinywrkb:
I think the best course of action is the following:

1. Show that this would be useful for an actual application.   
It could be a CLI app, avrdude comes to mind.  
It could also be a GUI app, avrdudess seems to be alive, though not very active. I got it to run, though only tesedt that the main UI is loading, you just need to build libgdiplus (with giflib), and manually deploy mono from the mono6 extension (the packaged install script does a poor job). 

2. Create a generic embedded development base app that will have a sdk/toolchain mount point with extension subdirectories enabled (see `man:flatpak-metadata(5)`). This base app could be used to build other toolchains, e.g. `riscv{32,64}-unknown-elf` and `arm-none-eabi`. I might be interested in riscv toolchain for package [Ripes](https://github.com/mortbopet/Ripes).  
I suggest using the KDE runtime, as it will give a good coverage of widget toolkits.  
Also add a few shared libs that are likely will be used by apps. libusb for example.

3. Submit the sdk/toolchain as an extension of the baseapp.  
Try to avoid binary dependency of extension files on libs from the base app to avoid ABI breakage on user installations.

ashahir05:
@hfiguiere this is meant to be a tool used in flatpak'd IDEs and text editors like vscode.

PR #2895, opened 2022-03-07
Ch.threema.threema web desktop

ljrk0:

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link: Done and got OK as per https://github.com/threema-ch/threema-web-electron/pull/4**
- [x] I own the domain used in the application ID or the domain has a policy for delegating subdomains (e.g. GitHub, SourceForge). **The domain is owned by upstream. See above**
- [x] Any additional patches or files have been submitted to the upstream projects concerned. *(If not, explain why.)* **The .desktop is generated upstream as part of the electron-installer-debian package inside a `.deb`.  We could extract it using `dpkg -x` but I think tracking a minimal .deskop is fine.**

I've already opened a previous PR at https://github.com/flathub/flathub/pull/2630, however the AppID was changed to `threema-web-desktop`, hence the new branch & PR.

Semi-open issues:

- [ ] Investigate arm64 build, shouldn't be *that* difficult, but can be done later.
- [ ] The condition for the OK is:
    > This is OK from our side, as long as it's clear from the package description that the packaging is being done by the community, and not by Threema. We are currently trying to focus our efforts on the development of the new standalone Threema desktop client, and at this point in time do not want to maintain / support the Flatpak package as well. (Once the new version is released, this may change.) However, we are open to packages created by the community.
    How is this best done? Patch/Amend the metadata before inclusion?
 
Does this otherwise look fine?


ch.threema.threema-web-desktop.yml
```diff
@@ -0,0 +1,74 @@
...
+    build-options:
+      env:
+        XDG_CACHE_HOME: /run/build/threema-web-desktop/flatpak-node/cache
+        npm_config_cache: /run/build/threema-web-desktop/flatpak-node/npm-cache
+        npm_config_nodedir: /usr/lib/sdk/node14
+        npm_config_offline: 'true'
+    sources:
+      - type: git
+        url: https://github.com/threema-ch/threema-web-electron.git
+        branch: develop
```
hfiguiere, line 35:
```suggestion
```

Never a branch

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64"]
```
hfiguiere, line 2: Still :-(

ch.threema.threema-web-desktop.yml
```diff
@@ -0,0 +1,74 @@
...
+      - cp -a app/build/dist-electron/packaged/Threema-linux-x64 /app/main
+      # Install app wrapper
+      - install -Dm755 -t /app/bin/ run.sh
+      # Appstream Metainfo
+      - install -Dm644 ch.threema.threema-web-desktop.metainfo.xml /app/share/metainfo/ch.threema.threema-web-desktop.metainfo.xml
+      # .desktop file
+      - install -Dm644 ch.threema.threema-web-desktop.desktop /app/share/applications/ch.threema.threema-web-desktop.desktop
+      # Icons
+      - mkdir -p /app/share/icons/hicolor/128x128/apps
+      - mkdir -p /app/share/icons/hicolor/512x512/apps
```
hfiguiere, line 72:
```suggestion
```

install -D will do that for you

ch.threema.threema-web-desktop.yml
```diff
@@ -0,0 +1,74 @@
...
+  - --share=ipc
+  - --socket=x11
+  - --socket=pulseaudio
+  - --share=network
+  - --filesystem=xdg-download
+build-options:
+  append-path: /usr/lib/sdk/node14/bin
+  env:
+    NPM_CONFIG_LOGLEVEL: info
+    DEV_ENV: development
```
hfiguiere, line 21: should that be something else since it's not a development build?

hfiguiere (changes requested):

ch.threema.threema-web-desktop.yml
ljrk0: Thanks for the quick review!

I went after the Flatpak Building guide which suggested adding a branch when using a commit to make sure the branch is "at this commit":

> The commit to use from the git repository. If branch is also specified, then it is verified that the branch/tag is at this specific commit. This is a readable way to document that you're using a particular tag, but verify that it does not change.

(https://docs.flatpak.org/en/latest/flatpak-builder-command-reference.html?highlight=branch#)

I don't have particular feelings on including it though :-)

ch.threema.threema-web-desktop.yml
ljrk0: Upstream confirmed that this is required "for now", but I wanted to document this oddity but forgot about it.

ch.threema.threema-web-desktop.yml
hfiguiere: no. this is outdated. (can't fix it). Build will fail if any commit is added to the branch, because it's a branch,

ljrk0: I see, will change this.

ljrk0: Ah, makes sense.  There are differences in the requirements/goals of Flatpak the tool and Flathub as a repo.  Could perhaps be worded differently, maybe I get around opening a PR there some day to clarify this.

ch.threema.threema-web-desktop.yml
ljrk0: Addressed this with a comment.

ljrk0:
bot, build ch.threema.threema-web-desktop

ljrk0:
bot, build ch.threema.threema-web-desktop

flathub.json
ljrk0: Is it possible to check which architecture is currently built?

ljrk0:
bot, build ch.threema.threema-web-desktop

flathub.json
ljrk0: I currently check `DESTCPU`, hopefully that's the "right" way to do things :)

barthalion:
Keep it mind that attempt of building Electron apps offline (as required by Flathub) is generally pain in the neck, so you may want to consider repackaging official binaries instead.

ljrk0:
> Keep it mind that attempts of building Electron apps offline (as required by Flathub) are generally a pain in the neck, so you may want to consider repackaging official binaries instead.

The building actually works quite fine, but apparently detecting build arch using `DESTCPU` doesn't work so the final packaging step fails since my shell script expands to `""`.  Using the `generated.json` dependencies works quite fine for building!  Also, they don't build aarch64 :-)

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
```
hfiguiere, line 1: with these values you can just remove the flathub.json file

ch.threema.threema-web-desktop.yml
```diff
@@ -0,0 +1,79 @@
...
+  - --socket=pulseaudio
+  - --share=network
+  - --filesystem=xdg-download
+build-options:
+  append-path: /usr/lib/sdk/node14/bin
+  env:
+    NPM_CONFIG_LOGLEVEL: info
+    # Must be set according to upstream: https://github.com/threema-ch/threema-web-electron/pull/4#issuecomment-995968724
+    DEV_ENV: development
+    threema_web_version: threema-web-2.3.19
```
hfiguiere, line 23: maybe you can add `DESTCPU` defined depending on the architecture?

hfiguiere (changes requested):

ch.threema.threema-web-desktop.yml
ljrk0: Hm, but how do I do that?  That is, how can I set values/execute code depending on the architecture?  I know that `flatpak-builder` is passed the `--arch` flag and it seems that indeed Electron seems to correctly "autodetect" this and builds for aarch64/amd64, but I have no idea how this detection works in order to actually retrieve the correct file.

flathub.json
ljrk0: Ah, so we don't build for i386 etc anymore?

hfiguiere: not for a few years

ch.threema.threema-web-desktop.yml
hfiguiere: something like that (adjust with your indent spacing):
```yaml
arch:
  x86_64:
    env:
      DESTCPU: x86_64
  aarch64:
    env:
      DESTCPU: aarch64
```

This should set `DESTCPU` to the right value.

hfiguiere: just to clarify the `arch` above is at the same level as `env` and `append-path` in the `build-options`

ch.threema.threema-web-desktop.yml
```diff
@@ -0,0 +1,79 @@
...
+      # Build Threema Electron
+      - npm run app:install
+      # Patch #3 for Threema Web (linux-deb really just means "linux")
+      - node tools/patches/post-patch-threema-web.js linux-deb consumer
+      # Here, linux:deb also just means "linux", effectively
+      - npm run electron:dist:linux:deb:consumer
+      # This would actually package .deb
+      #- npm run electron:package:linux:deb:consumer
+      # Bundle app and dependencies
+      - cp -a app/build/dist-electron/packaged/Threema-linux-$(case "$DESTCPU" in 'amd64') printf 'x64' ;; 'aarch64') printf 'arm64' ;; *) printf "$DESTCPU" ;; esac) /app/main
```
hfiguiere, line 70: Wouldn't this work?

```suggestion
      - cp -a app/build/dist-electron/packaged/Threema-linux-* /app/main
```

ch.threema.threema-web-desktop.yml
ljrk0: I see, thanks!

barthalion:
bot, build ch.threema.threema-web-desktop

ljrk0:
Ok, this seems to work, I have a typo in the `.desktop` and the `Social Media` category as specified in the packaging script doesn't actually exist, I will adjust that.  Outstanding tasks:

1. Patch the metadata to note that this package is "unofficial"
2. Fix typos / invalid categories in .desktop
3. Add per-arch environment variables.  Perhaps even go back to the shell-magic approach as globbing for files doesn't "feel right" IMHO.
4. Long-term:  Upstream the patches we currently carry
    a) metadata caption final dot remove
    b) aarch64 support
    c) invalid category "Social Media" in `.desktop`

Did I miss anything?  I could remove the `only-arches` property but indeed these are the only arches we support, so if at some point an additional architecture would be added to the Flathub build rack it might still be good to restrict the build until we've modified the packaging scripts accordingly as well.

ljrk0:
> No new architecture will magically appear. Just remove the `flathub.json`. Why? Because its absence cleary indicate we build all architecture.

Ah, I see.  Just asking in order to understand the process better and (hopefully) need less reviewing in future PRs :-)

ljrk0:
bot, build ch.threema.threema-web-desktop

ljrk0:
bot, build ch.threema.threema-web-desktop

ljrk0:
bot, build ch.threema.threema-web-desktop

ljrk0:
Hm, when installing I get:

    Warning: Not exporting file ch.threema.threema-web-desktop.appdata.xml of unsupported type.

barthalion:
Yeah, it's a bug in Flatpak that is not relevant to your manifest or the way the application will work. Just ignore it.

barthalion:
I don't use Threema so I can't really test it. Can you confirm it works as expected?

ljrk0:
Ah, good, then everything should be fine?

Yes, I've used it as a daily driver every time the build was successful.

PR #2897, opened 2022-03-07
Add Heroic Games Launcher

flavioislima:


flavioislima:
Question: Since the AppImage, after extraction has the app icons, do I need to copy it somewhere?

com.heroicgameslauncher.yml
```diff
@@ -0,0 +1,145 @@
...
+    buildsystem: simple
+    build-commands:
+      - install -Dm755 data/gamemoderun -t /app/bin
+    sources: *gamemode_sources
+
+  # --- Heroic ---
+  - name: heroic
+    buildsystem: simple
+    build-commands:
+      - unappimage Heroic-2.2.2.AppImage
```
GeorgesStavracas, line 108: What's the rationale for pulling the AppImage version here, instead of e.g. pulling the source and building it?

com.heroicgameslauncher.yml
```diff
@@ -0,0 +1,145 @@
...
+  - --allow=devel
+  - --device=all
+  - --talk-name=org.freedesktop.Notifications
+  - --persist=.wine
+  - --filesystem=home
+  - --filesystem=xdg-data/Steam:ro
+  - --filesystem=~/.var/app/com.valvesoftware.Steam:ro
+  - --require-version=1.1.2
+  - --env=XDG_CONFIG_DIRS=/etc/xdg:/usr/lib/x86_64-linux-gnu/GL:/usr/lib/i386-linux-gnu/GL
+  - --env=GST_PLUGIN_SYSTEM_PATH=/app/lib/gstreamer-1.0:/app/lib32/gstreamer-1.0:/usr/lib/extensions/gstreamer-1.0:/usr/lib/x86_64-linux-gnu/gstreamer-1.0:/usr/lib/i386-linux-gnu/gstreamer-1.0
```
GeorgesStavracas, line 30: Nitpick: it's a good practice to sort these alphabetically

com.heroicgameslauncher.yml
```diff
@@ -0,0 +1,145 @@
...
+
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=x11
+  - --socket=wayland
+  - --socket=pulseaudio
+  - --allow=multiarch
+  - --allow=devel
+  - --device=all
```
GeorgesStavracas, line 22: Is `all` required because of joysticks and controllers?

com.heroicgameslauncher.hgl.metainfo.xml
```diff
@@ -0,0 +1,59 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.heroicgameslauncher.hgl</id>
+  
+  <name>Heroic Games Launcher</name>
+  <summary>An Open Source GOG and Epic Games Launcher. </summary>
+  <url type="homepage">https://heroicgameslauncher.com</url>
```
GeorgesStavracas, line 7: Consider adding the following extra URL fields:
 * `<url type="bugtracker">...</url>`
 * `<url type="donation">...</url>`
 * `<url type="translate">...</url>`

GeorgesStavracas (commented):
Some comments so far

com.heroicgameslauncher.yml
flavioislima: Simplicity and effectiveness since pulling the whole source code, requiring NodeJS runtime, Python, and several other dependencies. Making the build completely bloated.
The AppImage runs pretty well, its compiled and optimized already.
So, unless there is something bad about using it, I believe we could stick to it :)

com.heroicgameslauncher.yml
flavioislima: Yes. It is a game launcher so we need at least Joysticks. 
I could not find a way of limiting this permission though.
Also, I believe that for devices like the Steam Deck that would be good to avoid issues.

com.heroicgameslauncher.yml
GeorgesStavracas: I can't make a compelling case neither in favour or against this choice because I don't know AppImage well enough to catch subtleties. However, I have the following concerns about it:

 * How does the AppImage work with regards to libc? Does it bundle it? Is mixing the unpacked AppImage with Flatpak platform libraries safe?
 * Which distro / platform is this AppImage built against?
 * If AppImage statically links dependencies, are you sure they wouldn't clash with dynamically linked libraries from the Flatpak runtime?

(To be abundantly clear, I am not playing devil's advocate; I'm sure you have good reasons to have opted for this approach, I just want to understand these reasons better.)

com.heroicgameslauncher.yml
GeorgesStavracas: There's no way to limit these to certain device types, it would need new portals.

JakobDev:
You should also add a translate and a help URL
```xml
<url type="translate">https://hosted.weblate.org/projects/heroic-games-launcher</url>
<url type="help">https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/wiki</url>
```
A list of all available types can be found [here](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-url).

JakobDev:
bot, build com.heroicgameslauncher.hgl

JakobDev:
bot, build com.heroicgameslauncher

JakobDev:
You should rename your YML file to com.heroicgameslauncher.hgl.yml The name of the manifest have to match the ID.

flavioislima:
> ```
> <url type="help">https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/wiki</url>
> ```

You are right, just fixed it.

GeorgesStavracas:
bot, build com.heroicgameslauncher.hgl

com.heroicgameslauncher.yml
flavioislima: Yes, no worries.
I am not an AppImage expert as well. :)
But this is an Electron App built with Electron Builder.
In summary, all Linux packages that we have today, also the Windows and the Mac ones, are the same thing, just a folder with all files build without external dependencies. so if I extract the AppImage and the Deb or the Pacman file, the content will be pretty similar, respecting only the package type differences.

JakobDev:
Have you installed the metainfo file in /app/share/metainfo?

GeorgesStavracas:
Build failed with `stat: cannot stat 'builddir/*/share/app-info/xmls/com.heroicgameslauncher.hgl.xml.gz': No such file or directory`

You need to manually copy the metainfo file with

```
  - name: heroic
      ...
      - type: file
        path: com.heroicgameslauncher.hgl.metainfo.xml
```

GeorgesStavracas:
bot, build com.heroicgameslauncher.hgl

flavioislima:
> - type: file
>         path: com.heroicgameslauncher.hgl.metainfo.xml

> Have you installed the metainfo file in /app/share/metainfo?
no, just fixed it

com.heroicgameslauncher.yml
TingPing: > How does the AppImage work with regards to libc? Does it bundle it? Is mixing the unpacked AppImage with Flatpak platform libraries safe?

It does nothing with libc and does not contain it. It may be broken if the developer builds against a newer libc but generally they develop against old libc to be safe. None

JakobDev:
This looks like a Bug of the Builder

com.heroicgameslauncher.hgl.metainfo.xml
```diff
@@ -0,0 +1,62 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.heroicgameslauncher.hgl</id>
+  
+  <name>Heroic Games Launcher</name>
+  <summary>An Open Source GOG and Epic Games Launcher. </summary>
```
TheEvilSkeleton, line 6:
```suggestion
  <summary>An Open Source GOG and Epic Games Launcher.</summary>
```

com.heroicgameslauncher.hgl.metainfo.xml
```diff
@@ -0,0 +1,62 @@
...
+  <url type="bugtracker">https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues</url>
+  <url type="donation">https://ko-fi.com/heroicgames</url>
+  <url type="translate">https://hosted.weblate.org/projects/heroic-games-launcher</url>
+  <url type="help">https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/wiki</url>
+  <metadata_license>GPL-3.0</metadata_license>
+  <project_license>GPL-3.0</project_license>
+  
+  <description>
+    <p>
+      Heroic is an Open Source Games Launcher. Right now it supports launching games from the Epic Games Store using Legendary and GOG Games using our custom implementation with gogdl.
```
TheEvilSkeleton, line 17:
```suggestion
      Heroic is an Open Source Games Launcher. Right now it supports launching games from the Epic Games Store using Legendary and GOG Games using our custom implementation with gogdl.
```

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,148 @@
...
+finish-args:
+  - --allow=devel
+  - --allow=multiarch
+  - --device=all
+  - --env=XDG_CONFIG_DIRS=/etc/xdg:/usr/lib/x86_64-linux-gnu/GL:/usr/lib/i386-linux-gnu/GL
+  - --env=GST_PLUGIN_SYSTEM_PATH=/app/lib/gstreamer-1.0:/app/lib32/gstreamer-1.0:/usr/lib/extensions/gstreamer-1.0:/usr/lib/x86_64-linux-gnu/gstreamer-1.0:/usr/lib/i386-linux-gnu/gstreamer-1.0
+  - --filesystem=home
+  - --filesystem=xdg-data/Steam:ro
+  - --filesystem=~/.var/app/com.valvesoftware.Steam:ro
+  - --persist=.wine
```
TheEvilSkeleton, line 23: This line is not going to work because of https://github.com/Heroic-Games-Launcher/flathub/blob/d63760503eb1bcf83d98caebf70446629f2ce227/com.heroicgameslauncher.hgl.yml#L20. It will ignore that `persist` option and continue writing to `~/.wine` (unless this was fixed).

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,148 @@
...
+      - -Dwith-sd-bus-provider=no-daemon
+    sources: &gamemode_sources
+      - type: archive
+        url: https://github.com/FeralInteractive/gamemode/releases/download/1.6.1/gamemode-1.6.1.tar.xz
+        sha256: 10c2a3f142eae472f5a09e42616e38c666c05b25ca3e61e562a543bb3fda66c5
+
+  - name: gamemode-32bit
+    build-options:
+      arch:
+        x86_64: &compat_i386_opts
```
gasinvein, line 93: This alias doesn't resolve. You've probably forgot to copy `x-compat-i386-opts: &compat_i386_opts` anchor.
Since invalid properties are ignored by flatpak-builder, you're effectively building 64-bit gamemode twice.

TheEvilSkeleton (changes requested):

com.heroicgameslauncher.hgl.yml
gasinvein, line 23: `--persist=` has no effect together with `--filesystem=home`

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,148 @@
...
+        cp /usr/bin/addr2line /app/bin/
+        cp /usr/lib/x86_64-linux-gnu/libbfd-*.so /app/lib/
+        install -Dm644 -t /app/etc ld.so.conf
+        mkdir -p /app/lib{,32}/ffmpeg
+        mkdir -p /app/utils /app/share/vulkan
+        ln -srv /app/{utils/,}share/vulkan/explicit_layer.d
+        ln -srv /app/{utils/,}share/vulkan/implicit_layer.d
+        mkdir -p /app/links/lib
+        ln -srv /app/lib /app/links/lib/x86_64-linux-gnu
+        ln -srv /app/lib32 /app/links/lib/i386-linux-gnu
```
gasinvein, line 145: Since you don't have an extension point at `/app/utils`, you don't need this.

com.heroicgameslauncher.hgl.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Name=Heroic Games Launcher
+Exec=/app/bin/heroic-run
```
hfiguiere, line 3:
```suggestion
Exec=heroic-run
```

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,148 @@
...
+
+sdk-extensions:
+  - org.freedesktop.Sdk.Compat.i386
+  - org.freedesktop.Sdk.Extension.toolchain-i386
+
+finish-args:
+  - --allow=devel
+  - --allow=multiarch
+  - --device=all
+  - --env=XDG_CONFIG_DIRS=/etc/xdg:/usr/lib/x86_64-linux-gnu/GL:/usr/lib/i386-linux-gnu/GL
```
hfiguiere, line 18: I doubt this is correct.

gasinvein (commented):
In addition to comments below: Doesn't Heroic Games Launcher need Wine to actually run games? Or some prebuilt Wine is downloaded at run time?

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,148 @@
...
+        commands:
+          - zypak-wrapper /app/bin/heroic/heroic "$@"
+
+      - type: file
+        filename: Heroic-2.2.2.AppImage
+        url: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/releases/download/v2.2.2/Heroic-2.2.2.AppImage
+        sha256: e2ea705052abd99fd100c00b18a5ce084c4548bc6a748bdd29bd739cc74d8f94
+     
+      - type: file
+        path: com.heroicgameslauncher.hgl.metainfo.xml
```
hfiguiere, line 124: this file isn't installed.

TheEvilSkeleton:
Any chance we could include Wine base app as well?

mirkobrombin:
> Any chance we could include Wine base app as well?

Heroic need zypak which is provided by the current base https://github.com/flathub/flathub/blob/d63760503eb1bcf83d98caebf70446629f2ce227/com.heroicgameslauncher.hgl.yml#L116 
If we can't use multiple bases we should install zyper independently.

TheEvilSkeleton:
I see... I think Flatpak should support multiple base apps. I'll open an issue.

flavioislima:
> Any chance we could include Wine base app as well?

I am not sure. Heroic has its own Wine/Proton Downloader, so I believe would be one more thing to be downloaded during Heroic build.

com.heroicgameslauncher.hgl.yml
flavioislima: So how do I fix that so it could install?

flavioislima:
> In addition to comments below: Doesn't Heroic Games Launcher need Wine to actually run games? Or some prebuilt Wine is downloaded at run time?

Heroic has a built-in wine/proton downloader that can be used to download several versions of Wine-GE and Proton-GE.
It will also use Proton if you have Steam installed.

com.heroicgameslauncher.hgl.yml
flavioislima: Which part? the `--env` ?

com.heroicgameslauncher.hgl.yml
barthalion: You can add something like `install -Dm644 $FLATPAK_ID.metainfo.xml /app/share/metainfo/$FLATPAK_ID.metainfo.xml` to `build-commands` in the same module.

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,148 @@
...
+    add-ld-path: lib
+    merge-dirs: vulkan/icd.d;glvnd/egl_vendor.d;OpenCL/vendors;lib/dri;lib/d3d;vulkan/explicit_layer.d;vulkan/implicit_layer.d
+    download-if: active-gl-driver
+    enable-if: active-gl-driver
+
+modules:
+
+  # --- Tools ---
+  - name: unappimage
+    buildsystem: simple
```
barthalion, line 58: I assume this is not needed after build? If so, it can be removed by adding `cleanup: ["*"]`.

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,148 @@
...
+      - mv squashfs-root /app/bin/heroic
+      - install -D heroic-run -t /app/bin
+    sources:
+      - type: script
+        dest-filename: heroic-run
+        commands:
+          - zypak-wrapper /app/bin/heroic/heroic "$@"
+
+      - type: file
+        filename: Heroic-2.2.2.AppImage
```
barthalion, line 119: Consider configuring https://github.com/flathub/flatpak-external-data-checker/ (see GitHub API example in README), unless you plan to automate it somehow else/maintain it fully manually.

barthalion:
bot, build com.heroicgameslauncher.hgl

com.heroicgameslauncher.hgl.yml
flavioislima: Done

com.heroicgameslauncher.hgl.yml
flavioislima: I will check this out.

com.heroicgameslauncher.hgl.yml
flavioislima: Done

com.heroicgameslauncher.hgl.yml
flavioislima: You were right, I removed it and everything still works :)

com.heroicgameslauncher.hgl.yml
flavioislima: Yes, removed it and everything seems to still work fine

com.heroicgameslauncher.hgl.desktop
flavioislima: done

com.heroicgameslauncher.hgl.yml
flavioislima: fixed

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,149 @@
...
+    enable-if: active-gl-driver
+  
+  x-compat-i386-opts: &compat_i386_opts
+  prepend-pkg-config-path: /app/lib32/pkgconfig:/usr/lib/i386-linux-gnu/pkgconfig
+  ldflags: -L/app/lib32
+  append-path: /usr/lib/sdk/toolchain-i386/bin
+  env:
+    CC: i686-unknown-linux-gnu-gcc
+    CXX: i686-unknown-linux-gnu-g++
+  libdir: /app/lib32
```
gasinvein, line 59:
```suggestion
  x-compat-i386-opts: &compat_i386_opts
    prepend-pkg-config-path: /app/lib32/pkgconfig:/usr/lib/i386-linux-gnu/pkgconfig
    ldflags: -L/app/lib32
    append-path: /usr/lib/sdk/toolchain-i386/bin
    env:
      CC: i686-unknown-linux-gnu-gcc
      CXX: i686-unknown-linux-gnu-g++
    libdir: /app/lib32
```

flavioislima: Removed more unnecessary commands and the `ld.so.conf` file that I believe was not necessary as well.

flavioislima:
Removed more unnecessary commands and the `ld.so.conf` file that I believe was not necessary as well.

com.heroicgameslauncher.hgl.yml
gasinvein: Basically, add `x-checker-data` object property
```suggestion
        filename: Heroic-2.2.2.AppImage
        x-checker-data:
          type: electron-updater
          url: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/releases/latest/download/latest-linux.yml
```
And change `sha256` to `sha512`

flavioislima: Yes, thats one me because I found a bug on Heroic after the release that was affecting the Flatpak, so I fixed it and uploaded a new AppImage.
But this is not common to happen and I will release a new version of Heroic today with the fixes.

gasinvein: This is not a blocker, just the update checker will not actually submit updates until the metadata is fixed or a new release comes out.

gasinvein:
> > Any chance we could include Wine base app as well?
> 
> I am not sure. Heroic has its own Wine/Proton Downloader, so I believe would be one more thing to be downloaded during Heroic build.

Well, if you use `org.winehq.Wine` as the base app, Wine will be available right after HGL is installed. It's somehow similar to having system (distro-provided) Wine package installed for non-flatpak HGL.

flavioislima:
Did most of the fixes, can we try to run the build again? :)

gasinvein:
> > > > Any chance we could include Wine base app as well?
> > > 
> > > 
> > > I am not sure. Heroic has its own Wine/Proton Downloader, so I believe would be one more thing to be downloaded during Heroic build.
> > 
> > 
> > Well, if you use `org.winehq.Wine` as the base app, Wine will be available right after HGL is installed. It's somehow similar to having system (distro-provided) Wine package installed for non-flatpak HGL.
> 
> Yes, but which wine version? Is it staging? Because for gaming should be at least that. Default wine won't work for the majority of them I believe.

It's Wine stable branch, 7.0.x currently. In my experience, latest stable is usually recent enough for most games, and it's that, stable, i.e. minimum regressions. Although some very recent games may not work without some staging patches indeed.

TheEvilSkeleton:
> Did most of the fixes, can we try to run the build again? :)

LGTM.

barthalion:
bot, build com.heroicgameslauncher.hgl

barthalion:
FYI you don't need anyone to trigger a new build, just post a comment like mine.

barthalion:
```
builddir/files/share/appdata/com.heroicgameslauncher.hgl.appdata.xml: FAILED:
• tag-invalid           : <metadata_license> is not valid [GPL-3.0]
```

It's usually `CC0-1.0`.

flavioislima:
> ```
> builddir/files/share/appdata/com.heroicgameslauncher.hgl.appdata.xml: FAILED:
> • tag-invalid           : <metadata_license> is not valid [GPL-3.0]
> ```
> 
> It's usually `CC0-1.0`.

But even for GPL?

flavioislima:
Ah ok, thats not the project license.

flavioislima:
bot, build com.heroicgameslauncher.hgl

JakobDev:
I think the build bot is just waiting for a free slot. This can happen when some build before takes a long time. You can view all ruining build [here](https://flathub.org/builds/). Currently there are 5 builds are running. There might also be other builds in the Queue before you.

JakobDev:
You have not installed the Icon. If you have a svg, install it under /app/share/icons/hicolor/scalable/apps. If you have a png, install it under /app/share/icons/hicolor/resolution/apps.

flavioislima:
> You have not installed the Icon. If you have a svg, install it under /app/share/icons/hicolor/scalable/apps. If you have a png, install it under /app/share/icons/hicolor/resolution/apps.

Since the Appimage already have the icons, to avoid duplication I added a line to move the icons from the heroic folder to the /app/share folder.

flavioislima:
bot, build com.heroicgameslauncher.hgl

JakobDev:
The Icons have to be named com.heroicgameslauncher.hgl.png not heroic.png. Everything that is exported must match the AppID.

hfiguiere:
the `.desktop` file doesn't seem to be installed.

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,153 @@
...
+      - mv squashfs-root /app/bin/heroic
+      - install -D heroic-run -t /app/bin
+    sources:
+      - type: script
+        dest-filename: heroic-run
+        commands:
+          - zypak-wrapper /app/bin/heroic/heroic "$@"
+
+      - type: file
+        filename: Heroic-2.2.2.AppImage
```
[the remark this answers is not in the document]
flavioislima: which name would be then?

flavioislima:
> the `.desktop` file doesn't seem to be installed.

where it should be installed?

JakobDev:
> > the `.desktop` file doesn't seem to be installed.
> 
> where it should be installed?

/app/share/applications/com.heroicgameslauncher.hgl.desktop

Please make sure that the Icon in the .desktop is also com.heroicgameslauncher.hgl.

flavioislima:
> > > the `.desktop` file doesn't seem to be installed.
> > 
> > 
> > where it should be installed?
> 
> /app/share/applications/com.heroicgameslauncher.hgl.desktop
> 
> Please make sure that the Icon in the .desktop is also com.heroicgameslauncher.hgl.

locally I am having this issue with the icons:
`Not exporting share/icons/hicolor/resolution/apps, wrong extension`
I am installing like this:
`install -Dm644 com.heroicgameslauncher.hgl.png /app/share/icons/hicolor/resolution/apps`

TheEvilSkeleton:
Try `install -Dm644 com.heroicgameslauncher.hgl.png /app/share/icons/hicolor/resolution/apps/com.heroicgameslauncher.hgl.png`.

GeorgesStavracas:
You need to replace `resolution` with the actual resolution of the icon, e.g. `/app/share/icons/hicolor/128x128/apps/com.heroicgameslauncher.hgl.png`

flavioislima:
Ok, trying here :D

Do you guys have an app that I could use as an example?
Some things like the desktop and icons I could not find on the documentation.

TheEvilSkeleton:
[UnityHub](https://github.com/flathub/com.unity.UnityHub/) maybe?

TheEvilSkeleton:
Are you sure the icon is correct?

image

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,158 @@
...
+      - |
+        set -e
+        mkdir -p /app/bin
+        mkdir -p /app/lib/i386-linux-gnu
+        mkdir -p /app/lib/debug/lib/i386-linux-gnu
+        mkdir -p /app/lib/i386-linux-gnu/GL
+        mkdir -p /app/lib/i386-linux-gnu/dri/intel-vaapi-driver
+        cp /usr/bin/addr2line /app/bin/
+        cp /usr/lib/x86_64-linux-gnu/libbfd-*.so /app/lib/
+        install -Dm644 com.heroicgameslauncher.hgl.png /app/share/icons/hicolor/128x128/apps/com.heroicgameslauncher.hgl.png
```
TheEvilSkeleton, line 146: Are you sure the icon is correct?

image

flavioislima:
ah, it should be exactly 128x128. Ok, I will get from the appimage

com.heroicgameslauncher.hgl.yml
flavioislima: fixed

TheEvilSkeleton: Feel free to mark it as resolved.

JakobDev:
Just take all the Icons in different sizes from the AppImage and rename them

JakobDev:
bot, build com.heroicgameslauncher.hgl

JakobDev:
Let's see if the builder accepts the new Icon size

com.heroicgameslauncher.hgl.yml
TheEvilSkeleton, line 146: I believe the "Check for 128x128 icon" check in the Flathub build is misleading because you can use an icon that isn't 128x128. I suggest to use 256x256 since it's higher resolution.

FTR, the MS Edge Flatpak does the same.
https://github.com/flathub/com.microsoft.Edge/blob/f147cdd39d5bf08a29b6a19acc96b934a23b5603/com.microsoft.Edge.yaml#L137

TheEvilSkeleton (changes requested):

hfiguiere:
> Just take all the Icons in different sizes from the AppImage and rename them

there is `rename-icon` for that.

flavioislima:
Build Success 🎉 
Will run another one because I released an update for Heroic

flavioislima:
bot, build com.heroicgameslauncher.hgl

TheEvilSkeleton:
Having this little error:
```
TheEvilSkeleton@TheMainDesktop ~> flatpak install --user https://dl.flathub.org/build-repo/78742/com.heroicgameslauncher.hgl.flatpakref

com.heroicgameslauncher.hgl permissions:
    ipc            network       pulseaudio       wayland               x11
    devices        devel         multiarch        file access [1]       dbus access [2]

    [1] home, xdg-data/Steam:ro, ~/.var/app/com.valvesoftware.Steam:ro
    [2] org.freedesktop.Notifications


        ID                               Branch     Op     Remote         Download
 1. [✓] com.heroicgameslauncher.hgl      test       i      hgl-origin     116.6 MB / 133.2 MB

Warning: Not exporting file com.heroicgameslauncher.hgl.appdata.xml of unsupported type.
Installation complete.
```

flavioislima:
> Having this little error:
> 
> ```
> TheEvilSkeleton@TheMainDesktop ~> flatpak install --user https://dl.flathub.org/build-repo/78742/com.heroicgameslauncher.hgl.flatpakref
> 
> com.heroicgameslauncher.hgl permissions:
>     ipc            network       pulseaudio       wayland               x11
>     devices        devel         multiarch        file access [1]       dbus access [2]
> 
>     [1] home, xdg-data/Steam:ro, ~/.var/app/com.valvesoftware.Steam:ro
>     [2] org.freedesktop.Notifications
> 
> 
>         ID                               Branch     Op     Remote         Download
>  1. [✓] com.heroicgameslauncher.hgl      test       i      hgl-origin     116.6 MB / 133.2 MB
> 
> Warning: Not exporting file com.heroicgameslauncher.hgl.appdata.xml of unsupported type. <!------------------------- this one
> Installation complete.
> ```
> 
> The launcher launches. I don't have an EGS account so I can't test any further.

you can use a GOG one as well.
About the warning, I have no clue.

TheEvilSkeleton:
I don't have a GOG account either :sweat_smile:.

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,158 @@
...
+      - mv squashfs-root /app/bin/heroic
+      - install -D heroic-run -t /app/bin
+    sources:
+      - type: script
+        dest-filename: heroic-run
+        commands:
+          - zypak-wrapper /app/bin/heroic/heroic "$@"
+
+      - type: file
+        path: Heroic-2.2.3.AppImage
```
hfiguiere, line 127:
```suggestion
```

`path` is to indicate where to find the file locally but you download it (see `url`)
I said previously you don't need that.

hfiguiere (changes requested):

JakobDev:
The Installation works so far in GOG, but no WIne or Proton Version is found.  Steam is installed as system package and not as Flatpak.

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,158 @@
...
+    build-commands:
+      - install -Dm755 data/gamemoderun -t /app/bin
+    sources: *gamemode_sources
+
+  # --- Heroic ---
+  - name: heroic
+    buildsystem: simple
+    build-commands:
+      - unappimage Heroic-2.2.3.AppImage
+      - rm Heroic-2.2.3.AppImage
```
hfiguiere, line 117:
```suggestion
      - unappimage Heroic-*.AppImage
```

1. otherwise the build will fail at the next versioj
2. you don't need to remove the file.

JakobDev:
>  About the warning, I have no clue.

This warning comes from the file /app/share/appdata/com.heroicgameslauncher.hgl.appdata.xml. This file is not needed.

flavioislima:
> The Installation works so far in GOG, but no WIne or Proton Version is found. Steam is installed as system package and not as Flatpak.
> 
> Another thing: I was promoted for a Update at startup. You should disable the Update checker completely when Heroic is installed by a package manager or Flatpak.

OK, will disable the update checker then.
It should find proton and wine. But ok. Have you tried to install using the wine manager?

com.heroicgameslauncher.hgl.yml
flavioislima: But then need to remove the line with the type as well?

flavioislima:
> > About the warning, I have no clue.
> 
> This warning comes from the file /app/share/appdata/com.heroicgameslauncher.hgl.appdata.xml. This file is not needed.

So I should remove the file completely or just the install part?

Torbuntu:
Was able to login to both EGS and GOG, I installed a game from EGS but Heroic cannot find any of the wine/proton runtimes/packages I have installed as flatpaks.

flavioislima:
> 

it wont find wine or proton flatpaks for now. I don't know yet on how to do it. For now you will need to use the Wine Manager to install Wine-GE and Proton-GE.
It should find Proton from Steam Flatpak though.

flavioislima:
> Was able to login to both EGS and GOG, I installed a game from EGS but Heroic cannot find any of the wine/proton runtimes/packages I have installed as flatpaks.


it wont find wine or proton flatpaks for now. I don't know yet on how to do it. For now you will need to use the Wine Manager to install Wine-GE and Proton-GE.
It should find Proton from Steam Flatpak though.

TheEvilSkeleton:
You need to add the Wine base app so they can see each other.

JakobDev:
I tired to install The latest Wine version with the Manager, and tried to launch Ocean Hert with it. It failed with a FileNotFound Error.
```
Launch Command: "/app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl" launch "/home/jakob/Games/Heroic/Ocean's Heart" 1819561805 --wine '/home/jakob/.var/app/com.heroicgameslauncher.hgl/config/heroic/tools/wine/Wine-GE-Proton7-6/bin/wine' --wine-prefix '/home/jakob/.var/app/com.heroicgameslauncher.hgl/Games/Heroic/Prefixes/Oceans-Heart' --os windows
        System Info:
        
  Heroic Version: 2.2.2 Oden
  Legendary Version:  0.20.25 Our Benefactors
  OS: Freedesktop.org KERNEL: 5.10.102-1-MANJARO ARCH: x64
  CPU: AMD Ryzen 5 2600 Six-Core Processor @3.4 GOVERNOR: schedutil
  RAM: Total: 16.8 GB Available: 6.6 GB
  GRAPHICS: GPU0: TU106 [GeForce RTX 2060 SUPER]  VRAM: 32MB DRIVER:  
  PROTOCOL: x11
  
        Game Settings: {
	"autoInstallDxvk": false,
	"autoInstallVkd3d": false,
	"maxSharpness": 2,
	"nvidiaPrime": false,
	"otherOptions": "",
	"savesPath": "",
	"showFps": false,
	"targetExe": "",
	"useGameMode": false,
	"wineCrossoverBottle": "Heroic",
	"winePrefix": "/home/jakob/.var/app/com.heroicgameslauncher.hgl/Games/Heroic/Prefixes/Oceans-Heart",
	"wineVersion": {
		"bin": "'/home/jakob/.var/app/com.heroicgameslauncher.hgl/config/heroic/tools/wine/Wine-GE-Proton7-6/bin/wine'",
		"name": "Wine - Wine-GE-Proton7-6",
		"type": "wine",
		"wineserver": "'/home/jakob/.var/app/com.heroicgameslauncher.hgl/config/heroic/tools/wine/Wine-GE-Proton7-6/bin/wineserver'",
		"wineboot": "'/home/jakob/.var/app/com.heroicgameslauncher.hgl/config/heroic/tools/wine/Wine-GE-Proton7-6/bin/wineboot'"
	},
	"useSteamRuntime": false
}

        Game Log:
  File "gogdl/cli.py", line 44, in <module>
  File "gogdl/cli.py", line 41, in main
  File "gogdl/launch.py", line 61, in launch
  File "subprocess.py", line 951, in __init__
  File "subprocess.py", line 1821, in _execute_child
FileNotFoundError: [Errno 2] No such file or directory: '/home/jakob/.var/app/com.heroicgameslauncher.hgl/config/heroic/tools/wine/Wine-GE-Proton7-6/bin/wine'
[869] Failed to execute script 'cli' due to unhandled exception!
```


>  It should find Proton from Steam Flatpak though.

Would be nice, if Heroic finds Proton if Steam is installed as native package (like in my case) and not as Flatpak.

> So I should remove the file completely or just the install part?

You should try to find out what puts the file there and then fix it.

flavioislima:
> You need to add the Wine base app so they can see each other.

yes but I need Electron base app as well. 
Like I said, for now its completely OK to use the Builtin Wine Manager to download them.
Then this can be improved later.

Also, Wine Stable is not the best version to play games.

TheEvilSkeleton:
Yeah no worries. I wasn't asking you to add the base app, I was just explaining how we might do it.

flavioislima:
> Yeah no worries. I wasn't asking you to add the base app, I was just explaining how we might do it.

if we have in the future support for multiple bases, then I believe will be a good thing to have :)

barthalion:
> Would be nice, if Heroic finds Proton if Steam is installed as native package (like in my case) and not as Flatpak.

This is not going to reliably work. HGL has to download its own vetted Wine, know how to re-use what Steam ships, or include a Wine build. Anything coming from the host will work only coincidentally.

barthalion:
bot, build com.heroicgameslauncher.hgl

flavioislima:
the build failed after removing the `metainfo` install. So I just added it again.

flavioislima:
> > Would be nice, if Heroic finds Proton if Steam is installed as native package (like in my case) and not as Flatpak.
> 
> This is not going to reliably work. HGL has to download its own vetted Wine, know how to re-use what Steam ships, or include a Wine build. Anything coming from the host will work only coincidentally.

Yes, Ideally you should use Proton from Steam Flatpak and Wine/Proton downloaded with Heroic.
Because Heroic flatpak uses different paths and configurations than the non-flatpak version.
So, I am not sure if it will work.

JakobDev:
> @JakobDev I had this issue yesterday and in my case was because org.freedesktop.Platform.Compat.i386 was not installed. I needed to install it manually.

That fixed it, but it's a big problem. If some installs Heroic through a Software Center e.g. on a Steam Deck, it should work.

> This is not going to reliably work. HGL has to download its own vetted Wine, know how to re-use what Steam ships, or include a Wine build. Anything coming from the host will work only coincidentally.

If you install heroic outside Flatpak e.g. with  the AUR, it can use the Proton version from Steam.

> the build failed after removing the metainfo install. So I just added it again.

The metainfo must be present. It's the .appdate file that causes the warning.

flavioislima:
@JakobDev so I need to remove the .appdate on a cleanup?

Also, the extension is on the build file, I don't know why it is not installing correctly sometimes.
There is any way to check if it is installed before the build?

JakobDev:
> @JakobDev so I need to remove the .appdate on a cleanup?

You can do that, but I think it should be better to find out what placing the file there.

> Also, the extension is on the build file, I don't know why it is not installing correctly sometimes.

You don't say that it should install. It says only, use that extension if installed. I'm sorry, but I don't know how to tell Flatpak it should install a extension. I have never used extension in one of my programs.

barthalion:
@flavioislima Please wait for the next test build to finish and I will have a look.

flavioislima:
> @flavioislima Please wait for the next test build to finish and I will have a look.

Thanks 👍

hfiguiere:
> @JakobDev so I need to remove the .appdate on a cleanup?

There should only be one. So **it's likely** the one you have in this repository is not needed and therefore be removed. (.metainfo.xml or .appdata.xml are the same thing)

JakobDev:
> (.metainfo.xml or .appdata.xml are the same thing)

I know, but .appdata is deprecated and .metainfo should be used. Quote from the [specification](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location):

> Component metadata of type desktop-application as described in [Section 2.2, “Desktop Applications”](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html) can be installed with an .appdata.xml extension as well for historical reasons. AppStream implementations will read the XML files as long as they end up in the right location on the filesystem.

> AppStream tools scan the /usr/share/appdata/ path for legacy compatibility as well. It should not be used anymore by new software though, even on older Linux distributions (like RHEL 7 and Ubuntu 16.04 LTS) the metainfo path is well supported. Support for the legacy path will likely be dropped completely with a future AppStream 1.0 release.

TheEvilSkeleton:
Same warning as before:
```
Warning: Not exporting file com.heroicgameslauncher.hgl.appdata.xml of unsupported type.
```

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,156 @@
...
+sdk-extensions:
+  - org.freedesktop.Sdk.Compat.i386
+  - org.freedesktop.Sdk.Extension.toolchain-i386
+
+finish-args:
+  - --allow=devel
+  - --allow=multiarch
+  - --device=all
+  - --env=GST_PLUGIN_SYSTEM_PATH=/app/lib/gstreamer-1.0:/app/lib32/gstreamer-1.0:/usr/lib/extensions/gstreamer-1.0:/usr/lib/x86_64-linux-gnu/gstreamer-1.0:/usr/lib/i386-linux-gnu/gstreamer-1.0
+  - --filesystem=home
```
TheEvilSkeleton, line 19: Why does it need that permission?

flavioislima: The default install folder is `$HOME/Games/Heroic`.

flavioislima:
@hfiguiere @JakobDev 
If you guys look at the build files, there is only one: metainfo.xml.
Looking for an `appdata` and I cant find it. Don't know what is generating it.

barthalion:
It's not immediately clear to me what causes this problem right now.

com.heroicgameslauncher.hgl.yml
barthalion: Do you know how often users change this path? If it's a reasonable default, it would be great to limit it to this path and restore `--persist` line for Wine.

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,156 @@
...
+        mkdir -p /app/lib/i386-linux-gnu
+        mkdir -p /app/lib/debug/lib/i386-linux-gnu
+        mkdir -p /app/lib/i386-linux-gnu/GL
+        mkdir -p /app/lib/i386-linux-gnu/dri/intel-vaapi-driver
+        cp /usr/bin/addr2line /app/bin/
+        cp /usr/lib/x86_64-linux-gnu/libbfd-*.so /app/lib/
+        install -Dm644 com.heroicgameslauncher.hgl.png /app/share/icons/hicolor/128x128/apps/com.heroicgameslauncher.hgl.png
+        install -Dm644 com.heroicgameslauncher.hgl.desktop /app/share/applications/${FLATPAK_ID}.desktop
+        install -Dm644 -t /app/etc ld.so.conf
+        install -Dm644 $FLATPAK_ID.metainfo.xml /app/share/metainfo/$FLATPAK_ID.metainfo.xml
```
barthalion, line 147:
```suggestion
        install -Dm644 $FLATPAK_ID.metainfo.xml /app/share/appdata/$FLATPAK_ID.appdata.xml
```

Mind changing it to this? It's a long shot but it seems like something is not quite right with the way the new name is handled. `appdata` is the old name and is the default code path.

flavioislima: Done

com.heroicgameslauncher.hgl.yml
flavioislima: What does persist does exactly?
I belive I can limit to a specific games folder then.
But probably lost users will install on other partitions but I've heard it's not ideal to add the host permission so I will let the users change the permissions with something like flatseal.
The thing is that the users can install games anywhere. Heroic will suggest the folder I said by default for both the games and wine prefixes.

barthalion:
bot, build com.heroicgameslauncher.hgl

com.heroicgameslauncher.hgl.yml
gasinvein: `--persist` maps given directory inside the sandbox to a "virtual home directory". It won't give access to `~/.wine` in the real home dir, but will cause anything written to `~/.wine` by the app to remain in place between app restarts.

barthalion:
OK, that's clearly irrelevant, and I'm somewhat sure it's some regression in Flatpak. It shouldn't make any difference in practice.

gasinvein:
Indeed, I'm starting to see the appdata warning from most of the other apps. I guess it can be ignored here.

com.heroicgameslauncher.hgl.yml
barthalion: `--persist` will set up a symlink/bind mount to a location writable by default instead of exposing it from the host, so something like `--persist=.wine` for Wine and `--filesystem=~/Games/Heroic:create` (or maybe `--persist=Games`? it requires testing) should generally work. We tend to prefer limited permissions as they can be changed relatively easily, especially when you (as the app maintainer) control the default location of where files are stored, but if you insist on `--filesystem=home`, it's ultimately your decision.

flavioislima: I see. Heroic can use Wine and Proton installed by non-flatpak versions of lutris and steam.
So it will need access for reading both folders inside home.
I can limit the permissions then, sounds more reasonable.

flavioislima:
ok, now I have limited the permissions to access only folders that Heroic really needs to access. So I believe makes more sense.

flavioislima:
bot, build com.heroicgameslauncher.hgl

travier:
I did not review the entire manifest but I can confirm that this works well on Fedora 35 KDE on Wayland. Thanks!

com.heroicgameslauncher.hgl.yml
TheEvilSkeleton: I think `--filesystem=~/Games/Heroic:create` would be best. Lutris does the same:

https://github.com/flathub/net.lutris.Lutris/blob/323bab375bdb02e4fecafc280bd6e25df26a1d35/net.lutris.Lutris.yml#L21

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,158 @@
...
+        dest-filename: heroic-run
+        commands:
+          - zypak-wrapper /app/bin/heroic/heroic "$@"
+
+      - type: file
+        x-checker-data:
+          type: electron-updater
+          url: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/releases/latest/download/latest-linux.yml
+        url: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/releases/download/v2.2.3/Heroic-2.2.3.AppImage
+        sha512: 952524d5f7e208eee25e5ffc53ed4b9dc240b0ffb7ea38b465114f4d65e636ecc47f0a4669e5c5f62f6744df9f9c087d07dc6690c1cc0afb6d8baaa9005df1d6
```
gasinvein, line 132: Nitpick: `x-checker-data` property is usually added last:
```suggestion
        url: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/releases/download/v2.2.3/Heroic-2.2.3.AppImage
        sha512: 952524d5f7e208eee25e5ffc53ed4b9dc240b0ffb7ea38b465114f4d65e636ecc47f0a4669e5c5f62f6744df9f9c087d07dc6690c1cc0afb6d8baaa9005df1d6
        x-checker-data:
          type: electron-updater
          url: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/releases/latest/download/latest-linux.yml
```

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,158 @@
...
+        mkdir -p /app/lib/i386-linux-gnu/dri/intel-vaapi-driver
+        cp /usr/bin/addr2line /app/bin/
+        cp /usr/lib/x86_64-linux-gnu/libbfd-*.so /app/lib/
+        install -Dm644 com.heroicgameslauncher.hgl.png /app/share/icons/hicolor/128x128/apps/com.heroicgameslauncher.hgl.png
+        install -Dm644 com.heroicgameslauncher.hgl.desktop /app/share/applications/${FLATPAK_ID}.desktop
+        install -Dm644 -t /app/etc ld.so.conf
+        install -Dm644 $FLATPAK_ID.metainfo.xml /app/share/appdata/$FLATPAK_ID.appdata.xml
+    sources:
+      - type: file
+        path: ld.so.conf
```
gasinvein, line 152: You can use recent `inline` type source to embed small files instead of storing them separately.
```suggestion
      - type: inline
        dest-filename: ld.so.conf
        contents: |
          /app/lib32
          /app/lib/i386-linux-gnu
```

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,158 @@
...
+
+sdk-extensions:
+  - org.freedesktop.Sdk.Compat.i386
+  - org.freedesktop.Sdk.Extension.toolchain-i386
+
+finish-args:
+  - --allow=devel
+  - --allow=multiarch
+  - --device=all
+  - --env=GST_PLUGIN_SYSTEM_PATH=/app/lib/gstreamer-1.0:/app/lib32/gstreamer-1.0:/usr/lib/extensions/gstreamer-1.0:/usr/lib/x86_64-linux-gnu/gstreamer-1.0:/usr/lib/i386-linux-gnu/gstreamer-1.0
```
gasinvein, line 18: Unless you plan to bundle gstreamer modules, you can remove `/app/lib/gstreamer-1.0` and `/app/lib32/gstreamer-1.0` from here.

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,158 @@
...
+  - --filesystem=~/.local/share/lutris:ro
```
gasinvein, line 19: Unless you hardcode the `~/.local/share/lutris` path in the HGL app (i.e. not XDG-compliant),
```suggestion
  - --filesystem=xdg-data/lutris:ro
```

com.heroicgameslauncher.hgl.yml
flavioislima: @TheEvilSkeleton changed that already on latest build 👍🏽

com.heroicgameslauncher.hgl.yml
flavioislima: We need that for proton to work.

gasinvein: `/app` is where the HGL app bundle is mounted at run time. There are no gstreamer modules in it. You need the `/usr/.../gstreamer-1.0` paths, but not `/app/.../gstreamer-1.0` paths.

ld.so.conf
```diff
@@ -0,0 +1,5 @@
+# We just make any GL32 extension have higher priority
+include /run/flatpak/ld.so.conf.d/app-*-org.freedesktop.Platform.GL32.*.conf
+/app/lib32
+/app/lib/i386-linux-gnu
+/lib64
```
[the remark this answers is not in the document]
flavioislima: Done for all files

com.heroicgameslauncher.hgl.yml
flavioislima: fixed on latest commit

com.heroicgameslauncher.hgl.yml
flavioislima: fixed

com.heroicgameslauncher.hgl.yml
flavioislima: Actually not, removed and proton still works :)

com.heroicgameslauncher.hgl.yml
flavioislima: fixed

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,158 @@
...
+    build-commands:
+      - |
+        set -e
+        mkdir -p /app/bin
+        mkdir -p /app/lib/i386-linux-gnu
+        mkdir -p /app/lib/debug/lib/i386-linux-gnu
+        mkdir -p /app/lib/i386-linux-gnu/GL
+        mkdir -p /app/lib/i386-linux-gnu/dri/intel-vaapi-driver
+        cp /usr/bin/addr2line /app/bin/
+        cp /usr/lib/x86_64-linux-gnu/libbfd-*.so /app/lib/
```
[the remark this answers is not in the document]
flavioislima: yes, removed

flavioislima:
bot, build com.heroicgameslauncher.hgl

ld.so.conf
```diff
@@ -0,0 +1,5 @@
+# We just make any GL32 extension have higher priority
```
gasinvein, line 1: This file isn't needed [anymore](https://github.com/flathub/flathub/pull/2897#discussion_r822753402).

flavioislima: ah ok, so the inline will write a file with the content. got it

flavioislima:
bot, build com.heroicgameslauncher.hgl

flavioislima:
So I believe we are good to go, now? :D

mwleeds:
> @hfiguiere @JakobDev If you guys look at the build files, there is only one: metainfo.xml. Looking for an `appdata` and I cant find it. Don't know what is generating it.

Line 146 of your manifest: 
`install -Dm644 $FLATPAK_ID.metainfo.xml /app/share/appdata/$FLATPAK_ID.appdata.xml`

flavioislima:
> > @hfiguiere @JakobDev If you guys look at the build files, there is only one: metainfo.xml. Looking for an `appdata` and I cant find it. Don't know what is generating it.
> 
> Line 146 of your manifest: `install -Dm644 $FLATPAK_ID.metainfo.xml /app/share/appdata/$FLATPAK_ID.appdata.xml`

it wasn't like that before, we changed to see if it would fix it. Before it was metadata. 
As said above, this seems like a flatpak regression.
I will revert to how it was before

flavioislima:
bot, build com.heroicgameslauncher.hgl

axtloss:
When trying the latest build (80983), it seems like heroic can't access the wine/proton installations. Proton-ge installed with flatpak doesn't get recognized at all, while the ones installed with the heroic wine manager do get recognized but fail to launch with the attached error message
screenshot_from_2022-03-08_23-51-00
the specified path does exist though:
screenshot_from_2022-03-09_00-13-27

---
Another thing I've noticed, but I think it may be an issue with my installation, is that after signing out I'm not able to log in anymore, once I pass the epic games log in I just get stuck at the heroic "logging in" screen
Screenshot from 2022-03-09 21-39-44

mirkobrombin:
> When trying the latest build (80983), it seems like heroic can't access the wine/proton installations. Proton-ge installed with flatpak doesn't get recognized at all, while the ones installed with the heroic wine manag..

The problem should be caused by the GE-Proton structure, files are located in `files/` and not in the root.

flavioislima:
@axtloss heroic won't find wine or proton flatpak.
For now you will need to use the ones installed on steam or using the wine manager.
This will be fixed in future versions of Heroic.

About the other issue, check if the Freedesktop.compat.i386 is installed on your machine.

Unfortunately we could not find a way of automate the installation but it fixes the error.

Proton should work even without it.

axtloss:
> @axtloss heroic won't find wine or proton flatpak.
> For now you will need to use the ones installed on steam or using the wine manager.
Yes I am aware of that, however using the wine manager also doesn't work, it says in the logs that the path doesn't exist, even though it does (1st and 2nd screenshot)

JakobDev:
> About the other issue, check if the Freedesktop.compat.i386 is installed on your machine.
> 
> Unfortunately we could not find a way of automate the installation but it fixes the error.

Thats really bad. You can show a message what to do, but it's still a bad user experience. The Heroic Games Launcher will be used by Linux newbies. Especially the Steam Deck use Flatpaks as main source for installing. This will bring many new Users to Linux and to Heroic. It's a Game machine, so a lot of Users. Linux already has a bad reputation that it's complex and only for nerds, so I don't know if it's a good idea to welcome new users by telling them they have to open a Terminal and run a random command just for using a Game Launcher, after they installed Heroic through a Software center.

gasinvein:
> About the other issue, check if the Freedesktop.compat.i386 is installed on your machine.
> 
> Unfortunately we could not find a way of automate the installation but it fixes the error.

This issue will be fixed by itself when the app is released on Flathub. Flatpak doesn't follow dependencies across repos, and the test build repo is technically a different one.

flavioislima:
> > About the other issue, check if the Freedesktop.compat.i386 is installed on your machine.
> > Unfortunately we could not find a way of automate the installation but it fixes the error.
> 
> Thats really bad. You can show a message what to do, but it's still a bad user experience. The Heroic Games Launcher will be used by Linux newbies. Especially the Steam Deck use Flatpaks as main source for installing. This will bring many new Users to Linux and to Heroic. It's a Game machine, so a lot of Users. Linux already has a bad reputation that it's complex and only for nerds, so I don't know if it's a good idea to welcome new users by telling them they have to open a Terminal and run a random command just for using a Game Launcher, after they installed Heroic through a Software center.

Yes so good that I'm wrong then.🙂
This will happens only if installing from the terminal.
Installing heroic from the store will install the extension.
So the user won't need to go to the terminal.
This is a pretty common extension and softwares like Bottles uses it as well. @mirkobrombin knows better.

flavioislima:
> > @axtloss heroic won't find wine or proton flatpak.
> > For now you will need to use the ones installed on steam or using the wine manager.
> 
> Yes I am aware of that, however using the wine manager also doesn't work, it says in the logs that the path doesn't exist, even though it does (1st and 2nd screenshot)

Yes. Like was said above. Since you installed from the terminal, you will need to install the extensions as well.
Aftet the flathub release it will work just fine.

axtloss:
Yeah sorry I misread your comment, however I'm not able to try that since as mentioned before I'm not able to sign in at all, neither with the external login nor with the built in one, it's always stuck at the logging in screen as seen in the third screenshot of my first comment.

flavioislima:
> Yeah sorry I misread your comment, however I'm not able to try that since as mentioned before I'm not able to sign in at all, neither with the external login nor with the built in one, it's always stuck at the logging in screen as seen in the third screenshot of my first comment.

Check the app logs on the settings screen.
I cannot reproduce the issue.

hojjatabdollahi:
@flavioislima I installed the latest build. I can log in with gog, but not with epic. I can see this error in the terminal:
```
(19:14:44) INFO:    [Legendary]:       Logging with Legendary...
(19:14:44) ERROR:   [Backend]:         Error: spawn "/app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/legendary" ENOENT
```
In the settings it says it's using the built-in Legendary binary.

barthalion:
@flavioislima The submission looks good to me, so if you consider it ready, I can merge it on your signal.

flavioislima:
> @flavioislima The submission looks good to me, so if you consider it ready, I can merge it on your signal.

Glad to hear. I will test a bit more since some people are having issues with Login and I will let you know once is ready :)

flavioislima:
> @flavioislima I installed the latest build. I can log in with gog, but not with epic. I can see this error in the terminal:
> 
> ```
> (19:14:44) INFO:    [Legendary]:       Logging with Legendary...
> (19:14:44) ERROR:   [Backend]:         Error: spawn "/app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/legendary" ENOENT
> ```
> 
> In the settings it says it's using the built-in Legendary binary.

I saw this message as well but for me, it works and that's strange because ENOENT should be not found.
Can you check if the path exists and the legendary binary is executable?

flavioislima:
bot, build com.heroicgameslauncher.hgl

axtloss:
After installing the newest build (81081) logging in properly works and playing games too, however in the logs I found this error:
Screenshot from 2022-03-10 15-17-43
I am not sure how important that is since the game worked with no issues.

flavioislima:
yes, I updated the appimage to test and now everything works but this message about hostname is something that is happening only on flatpak. But at least everything seems to be working fine now.

axtloss:
After playing around with the settings a bit, I noticed that it can't detect mangohud even though it is installed as flatpak, is this a limitation of not being on flathub yet?

flavioislima:
> After playing around with the settings a bit, I noticed that it can't detect mangohud even though it is installed as flatpak, is this a limitation of not being on flathub yet?

No, I believe heroic cant find Mangohud using the default command `mangohud`.
How should heroic run it? How do you do with other places like steam?

mirkobrombin:
> > After playing around with the settings a bit, I noticed that it can't detect mangohud even though it is installed as flatpak, is this a limitation of not being on flathub yet?
> 
> No, I believe heroic cant find Mangohud using the default command `mangohud`. How should heroic run it? How do you do with other places like steam?

To reach mongohud, heroic should support com.valvesoftware.Steam.Utility extension

axtloss:
Adding `MANGOHUD=1` as an envvar for the game properly launches mangohud, with no issues.

gasinvein:
> To reach mongohud, heroic should support com.valvesoftware.Steam.Utility extension

No! The `com.valvesoftware.Steam.Utility.MangoHud` extension is long EOL and replaced by the runtime-wide `org.freedesktop.Platform.VulkanLayer.MangoHud` extension that works with all apps. Setting `MANGOHUD=1` env var is enough for Vulkan apps. For the `mangohud` command to work, the app needs to add `/usr/lib/extensions/vulkan/MangoHud/bin/` to `PATH`.

flavioislima:
> org.freedesktop.Platform.VulkanLayer.MangoHu

Ok, how can I add the extension to Heroic?

flavioislima:
> adding `--env=PATH=/app/bin:/usr/bin:/usr/lib/extensions/vulkan/MangoHud/bin/` to com.heroicgameslauncher.hgl.yml in the finish-args should work.

nice, no need to add the extension itself? `org.freedesktop.Platform.VulkanLayer.MangoHud`
also, do you guys know if there is a way to use Winetricks with flatpak?

axtloss:
> nice, no need to add the extension itself? org.freedesktop.Platform.VulkanLayer.MangoHud

no, as said by [gasinvein](https://github.com/flathub/flathub/pull/2897#issuecomment-1064176504) mangohud is a runtime-wide extension, meaning it's available in every flatpak.

> also, do you guys know if there is a way to use Winetricks with flatpak?

It probably will have to be bundled with heroic or the installed wine from the wine manager itself to work.

flavioislima:
ok, so testing a new build now with v2.2.4 but I will release 2.2.5 later today and then I believe we can merge it

axtloss:
Tried the newest build (81108), and everything seems to work fine, I can log into epic games (I can't test gog), installing games and wine works with no issues, games can be launched and mangohud as well as gamemode work with no issues. Seems to be ready to merge!

flavioislima:
bot, build com.heroicgameslauncher.hgl

flavioislima:
@barthalion I believe we are good to go now :D

hojjatabdollahi:
@flavioislima I tested the latest build, the login to epic works.
I was able to install GE Proton using wine manager and install a game from epic. But running the game gives this error:
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/hojjat/.var/app/com.heroicgameslauncher.hgl/config/heroic/tools/wine/Wine-GE-Proton7-6/bin/wine'
[200] Failed to execute script 'cli' due to unhandled exception!
```
Although that file exists and is executable.

Also, it still gave me the prompt that "a new version is available" because you released 2.2.5. I thought you disabled that for the flatpak version.

flavioislima:
Actually wait a minute before the merge. I might have found a new bug.
I changed some commands to run inside a shell (to fix some stuff on windows). And now they are falling only on flatpak, not in AppImage or the pacman I installed.
How can I enable a shell inside the container?

flavioislima:
Nevermind, I believe we don't need the shell anymore after my changes today so I removed it and updated the app image.

So we can proceed with the merge.

axtloss:
> Actually wait a minute before the merge. I might have found a new bug. I changed some commands to run inside a shell (to fix some stuff on windows). And now they are falling only on flatpak, not in AppImage or the pacman I installed. How can I enable a shell inside the container? And it needs to be `BASH`, `SH` wont work.

bash should be in /bin/bash inside the flatpak, but if you don't need it anymore it works too I guess

flavioislima:
> > Actually wait a minute before the merge. I might have found a new bug. I changed some commands to run inside a shell (to fix some stuff on windows). And now they are falling only on flatpak, not in AppImage or the pacman I installed. How can I enable a shell inside the container? And it needs to be `BASH`, `SH` wont work.
> 
> bash should be in /bin/bash inside the flatpak, but if you don't need it anymore it works too I guess

Well, for some reason it doesn't work.
Actually is nodeJS that calls the shell, and we use bash as a parameter on linux, but inside the flatpak the legendary binary could not be executed.
Removing the shell parameter works fine now.

flavioislima:
bot, build com.heroicgameslauncher.hgl

axtloss:
Tried build 81121, also works with no issues but I found this when closing/quitting a game:
image
game still runs fine though

hojjatabdollahi:
I deleted the data from the old build. Installed the new build. Can not log in to epic again.
```
(12:02:25) INFO:    [Legendary]:       Logging with Legendary...
(12:02:25) ERROR:   [Backend]:         Error: spawn legendary ENOENT
```

flavioislima:
> I deleted the data from the old build. Installed the new build. Can not log in to epic again.
> 
> ```
> (12:02:25) INFO:    [Legendary]:       Logging with Legendary...
> (12:02:25) ERROR:   [Backend]:         Error: spawn legendary ENOENT
> ```
> 
> I also see this in the terminal:
> 
> ```
> (12:00:24) ERROR:   [Legendary]:       Error: Command failed: legendary --version
> /bin/sh: line 1: legendary: command not found
> 
> libva error: vaGetDriverNameByIndex() failed with unknown libva error, driver_name = (null)
> hostname: invalid option -- 'f'
> ```

Yeah, somehow those commands are running on SH and this breaks heroic.
Let's hold the merge while I fix those stuff again.
Maybe it's better just to get back to the latest working commit while I figure this out.
I will do that and update the appimage.

axtloss:
do you know where legendary is stored in the flatpak?

flavioislima:
> do you know where legendary is stored in the flatpak?

`.local/share/flatpak/app/com.heroicgameslauncher.hgl/current/active/files/bin/heroic/resources/app.asar.unpacked/build/bin/linux/`

flavioislima:
> 



> changing `--env=PATH=/app/bin:/usr/bin:/usr/lib/extensions/vulkan/MangoHud/bin/` to `--env=PATH=/app/bin:/usr/bin:/usr/lib/extensions/vulkan/MangoHud/bin:/app/bin/heroic/resources/app.asar.unpacked/build/bin/linux` will probably fix it
> 
> Edit: tested that and it works

Ah, great than so I don't need to make a new release. :)

flavioislima:
bot, build com.heroicgameslauncher.hgl

axtloss:
no you misunderstood, the line should be `--env=PATH=/app/bin:/usr/bin:/usr/lib/extensions/vulkan/MangoHud/bin:/app/bin/heroic/resources/app.asar.unpacked/build/bin/linux`

flavioislima:
I just copy pasted and tested here making a local build. Tried to login and it worked

flavioislima:
ah damn, now I see it, ahahahah

flavioislima:
bot, build com.heroicgameslauncher.hgl

flavioislima:
@barthalion finally everything seems to be fine now. 👍🏽

axtloss:
Tried build 81138, the error about legendary not being found doesn't exist anymore, everything runs with no issues.

hojjatabdollahi:
It works fine on my end. Logged in. Downloaded a game and GE Proton. (had to open the setting for the game once for it to find GE) and ran the game. GOG logs in too. Everything works.

flavioislima:
Yes, I am testing here as well. Tried native games, windows games using wine and proton, mangohud, etc. Seems to be working good now finally. 🎉

flavioislima:
> /merge
> 
> Note that it will take up to 4 hours for the app to be actually published.

Thanks a lot!

com.heroicgameslauncher.hgl.yml
```diff
@@ -0,0 +1,160 @@
+id: com.heroicgameslauncher.hgl
```
TheEvilSkeleton, line 1: Shouldn't this be `app-id`?

hfiguiere: either are accepted and equivalent.

PR #2908, opened 2022-03-11
Add org.sil.FieldWorks

marksvc:

  The app metainfo file passes other than that the validator doesn't seem to know about license "OFL-1.1-RFN".
  Permission usage is described in manifest comments.
  xchm/no-config.patch - The upstream project fixed the trouble I was experiencing with config files, but I can't upgrade to that newer xchm version until some other dependencies are sorted out.
  libxklavier/..Drop-intltool..patch - This is from Debian and helps build successfully.
  mono/mono-ping-path.patch - I haven't decided how to handle this with respect to upstream.


org.sil.FieldWorks.yml
```diff
@@ -0,0 +1,787 @@
...
+        # Branch develop
+        tag: FieldWorks9.0.17
+        commit: 6d09370772ecd99f3efbe776c55c6561d1b7af88
+      - type: file
+        url: https://github.com/sillsdev/FwLocalizations/releases/download/v9.0.17-14-g37c0f8f/crowdin.zip
+        sha512: 08586f7479d5c99d668d4ec92a1824e02f2982683abf3d13549e07dc1378eaf1204547a1ce6cf9a29f4a9ee418aeaf558e00a13e25e9bd41a0de880a20de1f39
+        dest: fw/Downloads/
+      - type: shell
+        commands:
+          - mv fw/Downloads/crowdin.zip fw/Downloads/Crowdin.zip
```
hfiguiere, line 152:
```suggestion
      dest-filename: Crowdin.zip
```

org.sil.FieldWorks.yml
```diff
@@ -0,0 +1,787 @@
...
+              - |
+                tee nuget.config <<< '<?xml version="1.0" encoding="utf-8"?>
+                <configuration>
+                  <packageSources>
+                    <add key="LocalNugetPackages" value="'$(pwd)/nuget-sources'" />
+                  </packageSources>
+                  <disabledPackageSources>
+                    <add key="nuget.org" value="true" />
+                  </disabledPackageSources>
+                </configuration>'
```
hfiguiere, line 656: you could use a "type: inline" to have this as an inline source.

org.sil.FieldWorks.yml
```diff
@@ -0,0 +1,787 @@
...
+  # FieldWorks allows playing sound files and playing and recording in audio
+  # writing systems.
+  - --socket=pulseaudio
+  # FieldWorks can import language and dictionary data, restore from backup
+  # files, and link to pictures or other media to appear in a dictionary.
+  # Users need to save project data, project backups, Encoding Converter
+  # configurations, and configuration and settings of FieldWorks and included
+  # components.
+  # Users need to be able to export dictionaries to arbitrary locations.
+  - --filesystem=host
```
hfiguiere, line 47: why would it need to write outside 'home' ?

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64"]
```
hfiguiere, line 2: any specific reason? Mono works on aarch64

hfiguiere (changes requested):

hfiguiere:
bot, build org.sil.FieldWorks

org.sil.FieldWorks.yml
marksvc: I was thinking if I limited it to home, users wouldn't be able to export to a USB Flash Drive.

flathub.json
marksvc: I would be interested to try it some time. I know FieldWorks bundles a specific gecko for rendering html, and I don't think that includes aarch64 at present.

marksvc:
The build failed building gtkmm2.4, but it did not appear to build its modules first.

org.sil.FieldWorks.yml
marksvc: Thank you :)

org.sil.FieldWorks.yml
marksvc: Thank you. I was not aware of inline.

nedrichards:
bot, build org.sil.FieldWorks

hfiguiere:
> The build failed building gtkmm2.4, but it did not appear to build its modules first.

gtk2 appear AFTER gtkmm2.4 from what I am seeing.

marksvc:
But it's in the modules list, right? I don't see that the builder tried to build atkmm1.6 either.
```
- name: gtkmm2.4
  modules:
  - name: atkmm1.6
    ...
  - shared-modules/gtk2/gtk2.json
```

marksvc:
Ok the problem is it couldn't deserialize modules. I'm checking on what happened with the submodule.

hfiguiere:
> But it's in the modules list, right? I don't see that the builder tried to build atkmm1.6 either.
> 
> ```
> - name: gtkmm2.4
>   modules:
>   - name: atkmm1.6
>     ...
>   - shared-modules/gtk2/gtk2.json
> ```

yaml is basically unreadable. But I would put gtk2 FIRST in that list.

org.sil.FieldWorks.yml
```diff
@@ -0,0 +1,786 @@
...
+# SOFTWARE.
+
+app-id: org.sil.FieldWorks
+
+# org.gnome.Platform provides some pieces that we would need to manually specify
+# if we use org.freedesktop.Platform.
+runtime: org.gnome.Platform
+# Target a runtime that is most similar to Ubuntu 20.04, for which we can easily
+# build and test on without flatpak.
+runtime-version: "3.36"
```
hfiguiere, line 30:
```suggestion
runtime-version: "41"
```

3.36 is EOL.

marksvc:
I think the newer runtime has resulted in some deprecation problems. Looking into it.

marksvc:
Okay, the problems with not having gtk, and deprecation, should both be fixed.

hfiguiere:
bot, build org.sil.FieldWorks

marksvc:
The newer runtime is resulting in some compilation trouble. Looking.

marksvc:
It will be good to fix the build failure that happens with the newer runtime, but I haven't learned its cause. I have set the runtime back to one where everything should build.

barthalion:
Is the 3.36 runtime the last one that worked? There is also 3.38 and 40 that are slightly newer.

marksvc:
@barthalion I checked again this morning. Runtime 3.38 and 40 both fail with the same problem. (And one needs to adjust the glibmm version back or forth to see.) I notice from some past work that mono 5 may not easily compile with newer versions of GCC. I see that org.gnome.Sdk 3.36 has GCC 9.3.0, while 3.38 and 40 have 10.2.0 and 41 has 11.2.0. So if that's the problem, then it may be a matter of getting GCC to allow for older code. (Although the error doesn't look like a _compilation error_. It's ... not clear what's gone wrong yet. So GCC may not be related to the trouble.)

hfiguiere:
Not sure what glibmm issue you are having, but I found that in some app like Inkscape, adding 
```json
            "build-options": {
                "cxxflags": "-DGLIB_VERSION_MIN_REQUIRED=GLIB_VERSION_2_66"
            },
```

helped if building failed because of glib.

marksvc:
@hfiguiere Thank you. When I used newer org.gnome.Sdk v41, and older glibmm 2.64.2, the error in this situation was
```
glib/glibmm/binding.cc: In member function ‘void Glib::Binding::unbind()’:
glib/glibmm/binding.cc:131:27: error: ‘GObject* g_binding_get_source(GBinding*)’ is deprecated: Use 'g_binding_dup_source' instead [-Werror=deprecated-declarations]
  131 |   if (g_binding_get_source(gobj()))
      |       ~~~~~~~~~~~~~~~~~~~~^~~~~~~~
```
(And if I use an older org.gnome.Sdk with a newer glibmm, then I get an error like
> meson.build:3:0: ERROR: Meson version is 0.53.2 but project requires >= 0.54.0

:)

I can try a CXX flag like you showed if I need to do further convincing of the build to work with these various versions.

marksvc:
`appstream-glib validate`'s list of licenses is not up to date, and it does not recognize the "OFL-1.1-RFN" license. I opened an issue for this at https://github.com/hughsie/appstream-glib/issues/435.

barthalion:
bot, build org.sil.FieldWorks

barthalion:
I don't think the license list will be really updated in a reasonable time, so please either drop the unknown license or try disabling the validation altogether like this: https://github.com/flathub/org.kde.katomic/blob/06a8d48244bcbdb113b18e1fffe9ad766d5b218e/flathub.json

nedrichards:
bot, build org.sil.FieldWorks

marksvc:
It had trouble fetching pieces of mono using git://. I'll look at how to work around that.

hfiguiere:
> It had trouble fetching pieces of mono using git://. I'll look at how to work around that.

You should use `https://` for git repositories

nedrichards:
> I don't think the license list will be really updated in a reasonable time, so please either drop the unknown license or try disabling the validation altogether like this: https://github.com/flathub/org.kde.katomic/blob/06a8d48244bcbdb113b18e1fffe9ad766d5b218e/flathub.json

Absolutely agreed. I got a [patch into appstream-glib master just now](https://github.com/hughsie/appstream-glib/commit/2e23a7780ea9c01be337d9bf05c7fdfbcd7d6ecc) but the line between there and 'trusted and tested code running on flathub' is not a clear or straight path.

barthalion:
bot, build org.sil.FieldWorks

barthalion:
Just to reiterate, it should be updated to the latest runtime eventually. I will merge it as-is for now.

/merge

PR #2918, opened 2022-03-14
Add Flatpak recipe for AppStream

ximion:
Hi!

- [x] I own the domain used in the application ID or the domain has a policy for delegating subdomains.

This adds a bundle for the `appstreamcli` utility, to make validating things or composing metadata easier for users on older distros which do not ship with an up-to-date AppStream out of the box.

One thing I am not sure about is that AppStream ships two metainfo files: `org.freedesktop.appstream.cli.metainfo.xml` for the main `appstreamcli` app, and `org.freedesktop.appstream.compose.metainfo.xml`, which is an `addon` component to the CLI utility. I haven't found a way to explicitly tell Flatpak which one of the files should be used though (besides simply removing one of the files, which feels wrong) - so, some advice on this would be appreciated!
The other question is, due to this basically exposing the `appstreamcli` utility, but not really the library and documentation and other stuff, whether the bundle's ID should rather be `org.freedesktop.appstream.cli`, and not just `org.freedesktop.appstream` - is there any preference on that?

Cheers,
    Matthias

### DIFF
--- a/flathub.json
+++ b/flathub.json
@@ -1,0 +1,3 @@
+{
+    "skip-appstream-check": true
+}
--- a/org.freedesktop.appstream.cli.yaml
+++ b/org.freedesktop.appstream.cli.yaml
@@ -1,0 +1,76 @@
+---
+app-id: org.freedesktop.appstream.cli
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+command: appstreamcli
+separate-locales: false
+
+finish-args:
+ - --filesystem=host
+ - --filesystem=~/.local/share/flatpak:ro
+ - --filesystem=/var/lib/flatpak:ro
+ - --share=network
+
+cleanup:
+ - /include
+ - /lib/pkgconfig
+ - /lib/*.a
+ - /lib/*.la
+
+modules:
+- name: libyaml
+  sources:
+  - type: git
+    url: https://github.com/yaml/libyaml.git
+    tag: '0.2.5'
+
+- name: snowball
+  buildsystem: simple
+  build-commands:
+   - make libstemmer.so -j$FLATPAK_BUILDER_N_JOBS
+   - install -Dm644 include/libstemmer.h /app/include/libstemmer.h
+   - cp -d libstemmer.so.* -t /app/lib/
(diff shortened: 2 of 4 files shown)


nedrichards:
bot, build org.freedesktop.appstream

nedrichards:
@ximion thanks for this! I've got a bit of a cold so my mind is full of sludge but this is absolutely a thing that can be done with a bit of tweaking (LibreOffice does it for example to export multiple 'apps') but I can't for the life of me explain it coherently right now. I'll either come back to you when I feel better or a more talented person will have helped you by then!

ximion:
> @ximion thanks for this! I've got a bit of a cold so my mind is full of sludge but this is absolutely a thing that can be done with a bit of tweaking (LibreOffice does it for example to export multiple 'apps') but I can't for the life of me explain it coherently right now. I'll either come back to you when I feel better or a more talented person will have helped you by then!

Looks like the build bot has spoken and wants this to be named `org.freedesktop.appstream.cli`, so I just made that change :-)
The addon not showing up is a really tiny issue that I am completely fine with ignoring for now :-)

barthalion:
We went with `org.freedesktop.appstream-glib` even though `appstream-util` was the entrypoint command. I guess it doesn't particularly matter, as long as obvious what it is.

bot, build org.freedesktop.appstream.cli

barthalion:
You need to add a `flathub.json` file like this one: https://github.com/flathub/org.kde.katomic/blob/06a8d48244bcbdb113b18e1fffe9ad766d5b218e/flathub.json as by default we validate for desktop apps.

ximion:
> We went with org.freedesktop.appstream-glib even though appstream-util was the entrypoint command. I guess it doesn't particularly matter, as long as obvious what it is.

Previously the build failed because the bot complained that the AppStream ID and Flatpak ID didn't match, so the easiest way to fix that for me was to just makes them identical :-)

> You need to add a flathub.json file as by default we validate for desktop apps.

Added :-)

barthalion:
bot, build org.freedesktop.appstream

barthalion:
bot, build org.freedesktop.appstream.cli

ximion:
Nice, thank you!
The Flathub process is very neat!

PR #2931, opened 2022-03-17
Add se.cendio.tlclient

Newbytee:

- [X] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** I have talked to them over email and in person (they have offices near the city I live in).
- [X] I own the domain used in the application ID or the domain has a policy for delegating subdomains (e.g. GitHub, SourceForge). **Clarification:** No, but this was the app ID that they suggested.

Draft because I want to let them get a final look before we merge.


se.cendio.tlclient.metainfo.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
```
hfiguiere, line 2: you also need screenshots.

se.cendio.tlclient.json
```diff
@@ -0,0 +1,62 @@
...
+				"mv tl-*-client-linux-dynamic-x86_64/* /app",
+				"mv /app/lib/tlclient/share /app",
+				"mv /app/share/applications/thinlinc-client.desktop /app/share/applications/${FLATPAK_ID}.desktop",
+				"mv /app/share/mime/packages/thinlinc-client-mime.xml /app/share/mime/packages/${FLATPAK_ID}.xml",
+				"desktop-file-edit --set-key=TryExec --set-value=/app/bin/tlclient-openconf /app/share/applications/${FLATPAK_ID}.desktop",
+				"desktop-file-edit --set-key=Exec --set-value='/app/bin/tlclient-openconf %f' /app/share/applications/${FLATPAK_ID}.desktop",
+				"desktop-file-edit --set-key=Icon --set-value=${FLATPAK_ID} /app/share/applications/${FLATPAK_ID}.desktop",
+				"rename thinlinc-client ${FLATPAK_ID} /app/share/icons/hicolor/*/apps/*",
+				"rename application-vnd.cendio.thinlinc.clientconf ${FLATPAK_ID} /app/share/icons/hicolor/*/mimetypes/*",
+				"mkdir -p /app/share/metainfo",
```
hfiguiere, line 36:
```suggestion
```

`install -D` does that for you

se.cendio.tlclient.json
```diff
@@ -0,0 +1,62 @@
...
+		"--persist=.fltk",
+		"--persist=.thinlinc",
+		"--share=ipc",
+		"--share=network",
+		"--socket=x11",
+		"--socket=pulseaudio"
+	],
+	"cleanup": [
+		"thinlinc.tar.gz"
+	],
```
hfiguiere, line 20:
```suggestion
```


cleanup is to remove install things, not source or buid artifacts.

se.cendio.tlclient.metainfo.xml
Newbytee: The problem is that upstream doesn't really provide any screenshots. The closest would be this: https://www.cendio.com/website/var/assets/firstpage/anatomy_thinlinc_session.png

Newbytee:
bot, build se.cendio.tlclient

se.cendio.tlclient.json
```diff
@@ -0,0 +1,58 @@
...
+	"tags": [
+		"proprietary"
+	],
+	"finish-args": [
+		"--persist=.fltk",
+		"--persist=.thinlinc",
+		"--share=ipc",
+		"--share=network",
+		"--socket=x11",
+		"--socket=pulseaudio"
```
CendioOssman, line 16: File system access is also needed for local drives to work. And for public key authentication.

Can this be interactive? So the user can choose how much access they want to give?

CendioOssman, line 16: Smart card support seems to need `--socket=pcsc`:

https://github.com/flatpak/flatpak/issues/4723

CendioOssman, line 16: How about Kerberos support?

se.cendio.tlclient.json
```diff
@@ -0,0 +1,58 @@
+{
+	"app-id": "se.cendio.tlclient",
+	"runtime": "org.freedesktop.Platform",
+	"runtime-version": "21.08",
```
CendioOssman, line 4: The ThinLinc client has very mild system requirements, so perhaps this can be set so it always picks the latest?

se.cendio.tlclient.json
```diff
@@ -0,0 +1,58 @@
...
+				"mv tl-*-client-linux-dynamic-x86_64/* /app",
+				"mv /app/lib/tlclient/share /app",
+				"mv /app/share/applications/thinlinc-client.desktop /app/share/applications/${FLATPAK_ID}.desktop",
+				"mv /app/share/mime/packages/thinlinc-client-mime.xml /app/share/mime/packages/${FLATPAK_ID}.xml",
+				"desktop-file-edit --set-key=TryExec --set-value=/app/bin/tlclient-openconf /app/share/applications/${FLATPAK_ID}.desktop",
+				"desktop-file-edit --set-key=Exec --set-value='/app/bin/tlclient-openconf %f' /app/share/applications/${FLATPAK_ID}.desktop",
+				"desktop-file-edit --set-key=Icon --set-value=${FLATPAK_ID} /app/share/applications/${FLATPAK_ID}.desktop",
+				"rename thinlinc-client ${FLATPAK_ID} /app/share/icons/hicolor/*/apps/*",
+				"rename application-vnd.cendio.thinlinc.clientconf ${FLATPAK_ID} /app/share/icons/hicolor/*/mimetypes/*",
+				"install -Dm644 ${FLATPAK_ID}.metainfo.xml -t /app/share/metainfo"
```
CendioOssman, line 33: We could probably start including this with the ThinLinc client if that helps.

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+	"only-arches": ["x86_64"]
```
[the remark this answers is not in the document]
Newbytee: Flathub does also support 64-bit ARM (ARMv8), but from what I see ThinLinc only has downloads for 32-bit ARM (ARMv7). 32-bit ARM support is unfortunately no longer provided by Flathub.

se.cendio.tlclient.json
Newbytee: > Can this be interactive? So the user can choose how much access they want to give?

Sort of. There is [Flatseal](https://github.com/tchx84/Flatseal) which is a GUI application for configuring this, but the proper way is to use desktop portals for filesystem access. See https://github.com/flatpak/xdg-desktop-portal. Unless you want to go about integrating that (I'm assuming FLTK does not provide support for this for you), it would probably be easiest to just always allow access to these directories to not confuse users. What kind of access is necessary?

se.cendio.tlclient.json
Newbytee: That would be nice. This can also be used by the .deb and .rpm packages. However, note that you'd then ideally also make sure the manifest contains the latest releases in the `<releases>` section.

se.cendio.tlclient.json
Newbytee: I'm not sure that's possible, and I'm not sure it's desired. There's always the risk that something stops working when you upgrade to the next runtime, even if it is unlikely. For that reason I think it's best to manually do this and test upgrades before you release them.

se.cendio.tlclient.json
barthalion: I assume krb5 and basic config needs bundling like Chromium does: https://github.com/flathub/org.chromium.Chromium/blob/master/org.chromium.Chromium.yaml#L112

CendioOssman: For local drives constant file access is needed, which I don't think there is a portal for.

I'd say the user's home drive is a minimum. Some user want to export things outside of that though, so complete filesystem access would be needed for them.

barthalion: …and this needs adding too, probably? https://github.com/flathub/org.chromium.Chromium/blob/master/org.chromium.Chromium.yaml#L9

hfiguiere: there is no lpr. if you support cups properly it can be added.

CendioOssman: This issue has more smart card details, and unfortunately support is fragile:

https://github.com/flatpak/flatpak/issues/4066

CendioOssman: > there is no lpr. if you support cups properly it can be added.

Do you have some reference on what is provided? E.g. should cups libraries be included in the flatpak and there is some socket mapped?

Newbytee: I could add this, but I'm not sure how to test whether it is working. Once we have screenshots the bot will produce a test build you can try though, @CendioOssman.

Newbytee: There's a `--socket=cups`, but I'm not sure if adding that is sufficient for it to work properly.

se.cendio.tlclient.json
```diff
@@ -0,0 +1,58 @@
...
+	"app-id": "se.cendio.tlclient",
+	"runtime": "org.freedesktop.Platform",
+	"runtime-version": "21.08",
+	"sdk": "org.freedesktop.Sdk",
+	"command": "tlclient-openconf",
+	"tags": [
+		"proprietary"
+	],
+	"finish-args": [
+		"--persist=.fltk",
```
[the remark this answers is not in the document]
Newbytee: Dropped.

se.cendio.tlclient.json
Newbytee: I added host filesystem access.

Newbytee: I added the pcsc and cups sockets, but I am not sure how to test whether they work.

CendioOssman: I assume that cups and pcsc-lite needs to be built and put inside the flatpak. That seems to be what the linked issues do.

@Newbytee, you can test smart card support without an actual smart card. The client will say "No smart card available" if smart card authentication is enabled. But it will say "No smart card support" if PC/SC support is missing.

Likewise you should be able to test printing by setting up some dummy printer queue locally.

CendioOssman: Sure. Kerberos is a bit of a hassle to set up so we can have a look once there is a test build.

Newbytee: > [@Newbytee](https://github.com/Newbytee), you can test smart card support without an actual smart card. The client will say "No smart card available" if smart card authentication is enabled. But it will say "No smart card support" if PC/SC support is missing.

Where would I see this?

image


image

barthalion:
bot, build se.cendio.tlclient

se.cendio.tlclient.json
CendioOssman: If you close the options after switching to smart card authentication the password field will be replaced with a combo box with that message.

Newbytee:
> bot, build se.cendio.tlclient

It's going to fail since there are no screenshots

Newbytee:
Oh, come on.

se.cendio.tlclient.json
Newbytee: Thanks. Unfortunately:

image

CendioOssman: Yes, your local system needs to have some basic smart card support as well. pcsc-lite needs to be installed and pcscd needs to be running (or autostarted by systemd).

The ThinLinc client's log file might have some details as well.

se.cendio.tlclient.metainfo.xml
CendioOssman: We've produced a set of screenshots now:

https://www.cendio.com/resources/images/screenshots/thinlinc/linux-en-intro.png
https://www.cendio.com/resources/images/screenshots/thinlinc/linux-en-local-drives.png
https://www.cendio.com/resources/images/screenshots/thinlinc/linux-en-fullscreen-mode.png
https://www.cendio.com/resources/images/screenshots/thinlinc/linux-en-profile-chooser.png
https://www.cendio.com/resources/images/screenshots/thinlinc/linux-en-login-window.png
https://www.cendio.com/resources/images/screenshots/thinlinc/linux-en-graphics-acceleration.png 

Our intent is to keep those URLs stable for things like flatpaks. The above order is also the order they're intended to be shown.

Newbytee:
bot, build se.cendio.tlclient

se.cendio.tlclient.metainfo.xml
Newbytee: Thanks! I've integrated them and used the captions on the images as captions in the manifest if that works for you. The screenshots look great!

Newbytee: Also, there's a test build available now down below.

barthalion:
@Newbytee  @CendioOssman LGTM, let me know when you want us to merge it.

se.cendio.tlclient.json
CendioOssman: We can confirm this works in our testing.

CendioOssman: So we've tested here now, and it looks like the following need to be included in the app (or with some other runtime?):

 * `libpcsclite.so.1`
 * cups-client, and probably cups-libs

CendioOssman: I'm afraid it does not work. So the app needs to include at least krb5-libs (specifically `libgssapi_krb5.so.2`, no idea about configuration).

Newbytee: Thanks, I'll look into that.

Newbytee:
> Also note that screenshots don't currently work. The links that flathub has injected in to the metainfo all give 404.
> 
> Perhaps these don't work until the flatpak is fully published?

Really? :thinking: They load fine here:

image

barthalion:
> Perhaps these don't work until the flatpak is fully published?

Yeah, that's correct. They will be fine once it's published.

se.cendio.tlclient.json
Erick555: As for lpr you may build just it instead of whole cups, see https://github.com/flathub/org.kde.okular/blob/404f59630e7d43bcddb97a76181e8f99f90422fd/org.kde.okular.json#L451-L470

Newbytee: After adding  libpcsclite.so.1, PCSC seems to work on my system: :partying_face: 

image

I've also added lpr, but I'm not sure how to test whether it works.

Newbytee: libgssapi_krb5.so.2 should now be present.

Newbytee:
bot, build se.cendio.tlclient

williamsjoblom:
Is there a thought behind the decision of using `tlclient-openconf` as `command` instead of `tlclient`? `tlclient-openconf` will not allow any arguments besides a configuration file to tlclient.

Newbytee:
> Is there a thought behind the decision of using `tlclient-openconf` as `command` instead of `tlclient`? `tlclient-openconf` will not allow any command line arguments besides a configuration file.

No, not really. I was not aware of the difference between them. As I recall I had some issue with the application not starting when I used the `tlclient` command, but it seems to start up just fine now like that, so I'll change it.

Newbytee:
bot, build se.cendio.tlclient

Newbytee:
bot, build se.cendio.tlclient

williamsjoblom:
Sweet! I have done some shallow testing of smart card support and things do not still seem to work as intended. The client reports "No smart cards available" and 0 slots even though I have a smart card connected (while the non-flatpak client correctly detects the smart card). I'll do some digging and see what I come up with :)

Erick555:
> Sweet! I have done some shallow testing of smart card support and things do not still seem to work as intended. The client reports "No smart cards available" and 0 slots even though I have a smart card connected (while the non-flatpak client correctly detects the smart card). I'll do some digging and see what I come up with :)

smartcard support in flatpak is very brittle especially if you test this on fedora.

williamsjoblom:
> smartcard support in flatpak is very brittle especially if you test this on fedora.

I have noticed. With `pcsc-lite-1.9.1-maxreaders.patch` from Fedora applied inside the flatpak things seem to work as intended, at least at the surface (see https://bugzilla.redhat.com/show_bug.cgi?id=2054826).

Erick555:
Right although applying this patch in flatpak would break all other distros so it's no solution. At least if this patch worked for you then it's a sign the problem isn't with app and everything possible was done.

williamsjoblom:
Besides the shaky platform-specific behavior mentioned, both smart card authentication and tunneling seemingly work as intended. So that's as good as it gets as this is an upstream RHEL/Fedora bug.

williamsjoblom:
Local printer redirection also seems to work as intended after the inclusion of cups :+1:

Newbytee:
> Besides the shaky platform-specific behavior mentioned, both smart card authentication and tunneling seemingly work as intended. So that's as good as it gets as this is an upstream RHEL/Fedora bug.
> Local printer redirection also seems to work as intended after the inclusion of cups :+1: 

Great! Let me know if you think it's ready to merge/publish.

williamsjoblom:
I have now given Kerberos some love as well and things seem to work relatively well. There are a few rough corners though. These all seem to be due to flatpak-limitations, so I guess this is as good as it gets.

**System Kerberos Configuration**

Ideally, we would like the contained krb5 to use the host system `/etc/krb5.conf` to get access to the configured `[domain_realm]`s. I could not find a simple solution to this. As `/etc/` is mounted under =/run/host/etc/= inside the flatpak sandbox the solution becomes fragile as many distributions use absolute path includes in `krb5.conf`.

We potentially have a similar printer-related problem with `/etc/cups` as well since this file can contain domain socket paths which also become fragile when sandboxed.

**Kerberos Credentials Cache**
`KCM` seems to work as expected given the following (see https://github.com/flathub/org.chromium.Chromium/blob/master/org.chromium.Chromium.yaml#L9):
```diff
diff --git a/se.cendio.tlclient.json b/se.cendio.tlclient.json
index 91741a8..b11ca3c 100644
--- a/se.cendio.tlclient.json
+++ b/se.cendio.tlclient.json
@@ -9,6 +9,7 @@
        ],
        "finish-args": [
                "--filesystem=host",
+               "--filesystem=/run/.heim_org.h5l.kcm-socket",
                "--persist=.thinlinc",
                "--share=ipc",
                "--share=network",
```

On the other hand, `FILE` and `DIR` seem broken as these require access to what is generally randomly named temporaries in `/tmp/`.

Similarly, `KEYRING` seems broken as well. See https://github.com/flatpak/flatpak/issues/4281

So, `KCM` is currently the only working credentials cache (with the minor addition above) and from what I can tell, the only one that we can expect to get working as a flatpak. Looking at other flatpaks using Kerberos such as Chrome and Firefox, this seems to be their conclusion as well.

So if you apply the above patch, Kerberos support can be considered as good as it gets :smiley:

Newbytee:
Thanks for looking into Kerberos! I applied your suggested change, so I assume this can be considered ready now?

Newbytee:
bot, build se.cendio.tlclient

williamsjoblom:
Just let me try out the build when it's ready and double-check that we didn't forget anything. Then everything's green from our side.

Newbytee:
@williamsjoblom Do you also want to be invited as a collaborator in the repository that will be created? Per email I was told you'd like to have @CendioOssman and @samhed invited.

williamsjoblom:
> @williamsjoblom Do you also want to be invited as a collaborator in the repository that will be created? Per email from Pierre Ossman I was told you'd like to have @CendioOssman and @samhed invited.
Yea, sure!

se.cendio.tlclient.metainfo.xml
```diff
@@ -0,0 +1,72 @@
...
+			It consists of both server-side components (session-broker,
+			load-balancer, administration interface, etc.) and end-user
+			components (client software, customization tools, etc.) to provide a
+			complete solution.
+		</p>
+		<p>
+			NOTE: This package is not verified by, affiliated with, or supported
+			by Cendio AB.
+		</p>
+	</description>
```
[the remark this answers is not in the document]
Newbytee: Done!

Newbytee:
bot, build se.cendio.tlclient

williamsjoblom:
Kerberos is working fine with the latest build. I'm heading home for the day now, but I'll make sure to look at the remaining bits first thing in the morning.

williamsjoblom:
Looks good to me! :+1:

barthalion:
@Newbytee @CendioOssman Would you consider it good to be merged now?

Newbytee:
> @Newbytee @CendioOssman Would you consider it good to be merged now?

It's a yes from me :+1:

CendioOssman:
I think everything is tested and fixed or bug reported, so we should be good to go.

Newbytee:
@barthalion You forgot @samhed D;

CendioOssman:
It seems to up and available now. Nice.

Unfortunately some bugs in flathub are affecting our package:

https://github.com/flathub/flathub/issues/2994
https://github.com/flathub/flathub/issues/2995

PR #2933, opened 2022-03-18
Add Chrome Dev Flatpak

refi64:
Yes I gutted the checklist, because this is mostly a carbon copy of Chrome but pointing to Dev (and with fewer permissions, gonna bet that the portal is now good enough for our filesystem use case? :crossed_fingers: )


refi64:
Forgot to mention, see the context behind wanting to push out a Dev Flatpak [here](https://discourse.flathub.org/t/plans-for-chrome-flatpak-stabilization-and-request-for-testers/2266)

barthalion:
This should go to beta, right?

refi64:
@barthalion 

> This should go to beta, right?

Originally I was thinking of this, but there were some replies on the Discourse thread that raised interesting points, in particular:

> Other Major Platforms have the beta & canary Chromes (and analogous Firefoxes, Edges, and other obscurely-named browsers) in their main repo.

> I suspect that web developers would appreciate having access to non-stable Chrome releases, and having them listed in the web store or the desktop app manager, and without having to add the flathub-beta repo.

Personally I'm not really committed to either direction, so I wouldn't mind an authoritative "no" here, but I *have* moved over into the non-beta repo camp overall.

barthalion:
I will merge it to beta for now to match the stable Chrome, and leave the decision to you about publishing it to regular Flathub.

/merge:beta

PR #2935, opened 2022-03-18
Add br.gov.fazenda.receita.irpf

guihkx:

- [ ] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** (I didn't)

---

### Disclaimer

I'm not affiliated with the Brazilian government in any way, I'm just your average Linux user tired of having to install this software every year and see it litter my home folder with a bunch of files installed in legacy paths. ¯\\\_(ツ)_/¯

And I'm sure that **if** this gets distributed on Flathub, it will be much appreciated by other Brazilians too. :)

### About the software

This is a Java-based software developed by the Brazilian Government to assist Brazilian citizens do their taxes.

Their official page is here (in Portuguese): https://www.gov.br/receitafederal/pt-br/assuntos/meu-imposto-de-renda

### Runtime used

Unfortunately, unless I use the `org.gnome.Platform` runtime, any clickable link within the program won't do anything. That's the only reason why I'm not using `org.freedesktop.Platform`.

### Filesystem permissions

Also adding to the point above, I had to set both `--talk-name=org.gtk.vfs.*` and `--filesystem=xdg-run/gvfsd` in order to properly open links with the web browser. And the `--filesystem=~/ProgramasRFB:create` permission is necessary for the app to import tax data from current or previous years.

Let me know if you have any more questions. Thanks.

### OpenJDK 11 instead of 17

Unfortunately, if we use OpenJDK 17, the program displays an annoying warning after you launch it, reminding you that "they recommend utilizing Java 17". I deem this as bad UX, so that's why I decided to stick with OpenJDK 11.


guihkx:
I've been packaging this app myself for about a year now, but just now I noticed that the namespace changed. 🤦 

I think I would make sense to change it to `br.gov.receitafederal.IRPF`, but I'm not sure.

Nevertheless, before I actually do any more changes, please let me know if you guys actually plan to include this app on Flathub.

Thanks!

barthalion:
I believe we already had government applications submitted before by "third parties" so I don't think there's any reason not to accept yours.

br.gov.economia.receita.IRPF.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Name=IRPF 2022
+GenericName=Imposto de Renda Pessoa Física
+Comment=Programa de Declaração de Imposto sobre a Renda da Pessoa Física
+Icon=br.gov.economia.receita.IRPF
+Exec=/app/bin/irpf
```
barthalion, line 6:
```suggestion
Exec=irpf
```

br.gov.economia.receita.IRPF.yaml
```diff
@@ -0,0 +1,134 @@
+app-id: br.gov.economia.receita.IRPF
+branch: stable
```
barthalion, line 2:
```suggestion
```

br.gov.economia.receita.IRPF.yaml
```diff
@@ -0,0 +1,134 @@
...
+      - type: file
+        path: icon-64.png
+      - type: file
+        path: icon-128.png
+      - type: file
+        path: icon-256.png
+  - name: irpf_xml
+    buildsystem: simple
+    build-commands:
+      - install -Dm644 *.xml "${FLATPAK_DEST}/share/br.gov.economia.receita.IRPF/lib/resources"
```
barthalion, line 57: As a workaround, you could also symlink this location to a writable path, e.g. a directory under `$XDG_DATA_HOME`.

br.gov.economia.receita.IRPF.yaml
```diff
@@ -0,0 +1,134 @@
...
+command: irpf
+finish-args:
+  - --share=network
+  - --socket=x11
+  - --share=ipc
+  - --env=PATH=/app/jre/bin:/app/bin:/usr/bin
+  - --env=JAVA_HOME=/app/jre
+  - --talk-name=org.gtk.vfs.*
+  - --filesystem=xdg-run/gvfsd
+  - --filesystem=~/ProgramasRFB:create
```
barthalion, line 17: Does flatpak complain if you drop `:create` but the directory does not exist?

barthalion:
bot, build br.gov.economia.receita.IRPF

br.gov.economia.receita.IRPF.yaml
guihkx: That's a good idea! About the implementation though, I assume I'd have to create and populate this writable directory during runtime, i.e. in the `irpf` script? 🤔

guihkx:
Thanks for the review, I'm addressing the issues locally and I'll push the changes shortly.

Regarding the namespace of the app, I think I will rename it from `br.gov.economia.receita.IRPF` to `br.gov.fazenda.receita.irpf`, which is the same ID used by their official Android app: https://play.google.com/store/apps/details?id=br.gov.fazenda.receita.irpf

Any objections?

br.gov.economia.receita.IRPF.yaml
guihkx: It doesn't complain, but then the directory doesn't get created when you open the app.

nedrichards:
My goodness, I remember investigating packaging this at Endless many years ago, nice work. We'd certainly accept this - as and when it meets all the standards of course.

br.gov.economia.receita.IRPF.yaml
guihkx: Actually, I'm not sure I understand this correctly. I'm thinking I'd have to point `$XDG_DATA_HOME/resources` to `/app/share/br.gov.economia.receita.IRPF/lib/resources` in order for the app to be capable of updating these XML files. But how am I supposed to do that? Because when there is an update, the app tries to overwrite the XML files sitting under `/app/share/br.gov.economia.receita.IRPF/lib/resources` .

hfiguiere: This tree is read-only.

I wonder if it is not better to get the data as "extra-data".

guihkx: So, something like JDownloader does [currently](https://github.com/flathub/org.jdownloader.JDownloader/blob/7e0c5072d35329d52303940d02c1b3415d6ae72f/org.jdownloader.JDownloader.yaml#L56), I presume?

Everything is "installed" under [`${XDG_DATA_HOME}/jdownloader`](https://github.com/flathub/org.jdownloader.JDownloader/blob/b52b4a27026ca486f53bccb40ccacea38115da38/jd-wrapper.sh#L3) (including the main .jar file), which allows the app to auto-update, but I personally am not a big fan of that approach.

At least for this app, it just silently fails when it can't overwrite these XML files. Besides, I've made a script to help keep them up-to-date with ease.

guihkx: Nevertheless, let me know what approach you guys think it's best.

nedrichards:
bot, build br.gov.economia.receita.IRPF

nedrichards:
bot, build br.gov.fazenda.receita.irpf

br.gov.economia.receita.IRPF.yaml
barthalion: Yeah, that's just a suggestion. Whatever makes your maintenance easier.

PR #2950, opened 2022-03-23
Add org.apache.directory.studio

wjt:

- [ ] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://issues.apache.org/jira/browse/DIRSTUDIO-1153?focusedCommentId=17511282&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17511282
  - I do not own the domain, but the app ID is based on the upstream site.
  - I have not yet submitted the .desktop or metainfo files upstream, though I have mentioned them in the ticket about a Flatpak package linked above.


flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64"]
```
hfiguiere, line 2: :-( I thought Java was "run everywhere"

wjt:
I'm sure a sourceful build would be possible (gaining aarch64 support) but I'm afraid the limit of my interest in this app is to be able to create LDAP accounts for new team members at Endless OS Foundation, on Endless OS, with a UI. The easiest way to run it was to make a Flatpak with the OpenJDK extension, so the marginal cost of adding the `.desktop` and `.metainfo.xml` files was quite low.

org.apache.directory.Studio.desktop
```diff
@@ -0,0 +1,8 @@
+[Desktop Entry]
+Type=Application
+Name=Apache Directory Studio
+Comment=A complete directory tooling platform, intended to be used with any LDAP server
+Icon=org.apache.directory.Studio
+Exec=/app/ApacheDirectoryStudio/ApacheDirectoryStudio
```
hfiguiere, line 6: you should have a symlink in `/app/bin` and just put that (no path)

org.apache.directory.Studio.yaml
```diff
@@ -0,0 +1,54 @@
+app-id: org.apache.directory.Studio
+
+sdk: org.gnome.Sdk
+runtime: org.gnome.Platform
```
hfiguiere, line 4: does this really need the GNOME platform?

org.apache.directory.Studio.yaml
```diff
@@ -0,0 +1,54 @@
...
+runtime-version: '42'
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.openjdk17
+
+command: /app/ApacheDirectoryStudio/ApacheDirectoryStudio
+finish-args:
+  - --device=dri
+  - --device=shm
```
hfiguiere, line 12: Are you sure it needs that?

org.apache.directory.Studio.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+  - name: ApacheDirectoryStudio
+    buildsystem: simple
+    build-commands:
+      - install -Dt /app/share/applications org.apache.directory.Studio.desktop
+      - install -Dt /app/share/metainfo org.apache.directory.Studio.metainfo.xml
+      - install -D studio-icon_128x128.png /app/share/icons/hicolor/128x128/apps/org.apache.directory.Studio.png
+    sources:
+      - type: file
+        path: org.apache.directory.Studio.desktop
```
hfiguiere, line 50: so Apache ship pre builds for Linux and don't even provide a .desktop file nor an icon?

org.apache.directory.Studio.yaml
```diff
@@ -0,0 +1,54 @@
+app-id: org.apache.directory.Studio
+
+sdk: org.gnome.Sdk
+runtime: org.gnome.Platform
+runtime-version: '42'
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.openjdk17
+
+command: /app/ApacheDirectoryStudio/ApacheDirectoryStudio
```
hfiguiere, line 9: like for the desktop. You should have a symlink in `/app/bin` and use the command name, no path.

hfiguiere (changes requested):

flathub.json
wjt: It uses (and bundles) [SWT](https://en.wikipedia.org/wiki/Standard_Widget_Toolkit) which links to GTK.

org.apache.directory.Studio.yaml
wjt: I couldn't find a `.desktop` file in the bundle.

There is a 48×48 XPM at the top level of the bundle… Oh, there actually is a 128×128 PNG buried in there.

org.apache.directory.Studio.yaml
wjt: I'm not!

org.apache.directory.Studio.yaml
wjt: I swear I tried this and it didn't work, but now it seems to.

org.apache.directory.Studio.yaml
wjt: Seems you're right.

flathub.json
wjt: It doesn't look like a build from source would be particularly difficult, assuming the existence of a maven manifest generator or some manual work to fake it… But I'm afraid I'm not about to work on it.

https://gitbox.apache.org/repos/asf?p=directory-studio.git;a=blob;f=build.sh;h=698c8e80d4a89c1aeddef7794831c7b3a39b359f;hb=HEAD

barthalion:
bot, build org.apache.directory.Studio

wjt:
```
<groupId>org.apache.directory.studio</groupId>
```

looks like I should lowercase it.

barthalion:
I don't think it particularly matters if the application does not use D-Bus, but I will leave the decision to you.

nedrichards:
I think it's likely to ease any upstream adoption of app-id named assets to have it actually match what they think the app-id is so worth doing.

org.apache.directory.Studio.yaml
wjt: It actually does need the GNOME Platform for webkit reasons.

```
SWT Webkit: Warning, You are using an old version of webkitgtk. (pre 2.4) BrowserFunction functionality will not be avaliable
SWT WebKit: error initializing DBus server, dBusServer == 0
SWT.CHROMIUM style is not supported in Wayland.
```

With this change the welcome page does not work.

org.apache.directory.studio.yaml
```diff
@@ -0,0 +1,54 @@
...
+command: ApacheDirectoryStudio
+finish-args:
+  - --device=dri
+  - --env=PATH=/app/bin:/app/jre/bin:/usr/bin
+  - --persist=.ApacheDirectoryStudio
+  - --persist=.eclipse
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --socket=fallback-x11
```
wjt, line 18: FWIW while the app does *work* with Wayland, it's definitely buggier than with X11. For instance, the "are you sure you want to quit?" dialog is unusably small when first opened.

However I think it's probably good enough.

barthalion:
bot, build org.apache.directory.studio

PR #2966, opened 2022-03-30
Add Needly (io.github.lruzicka.Needly) to Flathub

lruzicka:

- [X] I am using extended permissions to allow the application to read and write to the host system, the entire home directory, and to talk to libvirt sessions because the application takes screenshots from libvirt and edits files that might reside anywhere on the host system.
    - `--device=kvm` # Allow access to a KVM device
    - `--filesystem=home` # Allow read-write access to home dir
    - `--filesystem=host` # Allow read-write access to host dir tree
    - `--filesystem=xdg-run/libvirt` # Access to qemu:///session
    - `--filesystem=/run/libvirt` # Access to qemu:///system 


README.md
```diff
@@ -0,0 +1,181 @@
+# Needly (Version 2.5)
```
hfiguiere, line 1: no user will see this README

com.github.lruzicka.Needly.yaml
```diff
@@ -0,0 +1,113 @@
...
+#
+app-id: com.github.lruzicka.Needly
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+command: needly
+finish-args: 
+   - --share=ipc # For X11
+   - --socket=wayland # Enable communication with Wayland
+   - --socket=x11 # Alternatively, with X11
```
hfiguiere, line 11:
```suggestion
   - --socket=fallback-x11 # Alternatively, with X11
```

unless there is a good reason you need simultaneous X11 and wayland.

com.github.lruzicka.Needly.yaml
```diff
@@ -0,0 +1,113 @@
...
+   - --device=kvm # Allow access to a KVM device
+   - --filesystem=home # Allow read-write access to home dir
+   - --filesystem=host # Allow read-write access to host dir tree
```
hfiguiere, line 14: why do you need `host` filesystem? This is very broad. Also it includes `home`

com.github.lruzicka.Needly.yaml
```diff
@@ -0,0 +1,113 @@
...
+app-id: com.github.lruzicka.Needly
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+command: needly
+finish-args: 
+   - --share=ipc # For X11
+   - --socket=wayland # Enable communication with Wayland
+   - --socket=x11 # Alternatively, with X11
+   - --device=kvm # Allow access to a KVM device
```
hfiguiere, line 12: I must misunderstand what the app does to ask why it wants `kvm` access

com.github.lruzicka.Needly.yaml
```diff
@@ -0,0 +1,113 @@
...
+      - type: file
+        url: https://files.pythonhosted.org/packages/2a/74/919462bed158ccc2a8061f31d82c157740c2e597ff50490526336e8aa688/libvirt-python-8.1.0.tar.gz
+        sha256: a21ecfab6d29ac1bdd1bfd4aa3ef58447f9f70919aefecd03774613f65914e43
+
+  # The Needly application
+  - name: needly
+    buildsystem: simple
+    build-options:
+      build-args:
+        - --share=network
```
hfiguiere, line 93: this is not allowed thus will fail.

Any python dependency must be added to the manifest.

bundle/.flatpak-builder/ccache/disabled/ccache.conf
```diff
@@ -0,0 +1 @@
+disable = true
```
hfiguiere, line 1: why do you need this?

com.github.lruzicka.Needly.yaml
```diff
@@ -0,0 +1,113 @@
...
+    build-commands:
+      - pip3 install --prefix=/app needly-2.5.5.tar.gz
+      - mkdir -pv /app/share/applications/
+      - cp -v *.desktop /app/share/applications
+      - mkdir -pv /app/share/icons/hicolor/128x128/apps
+      - cp -v *.png /app/share/icons/hicolor/128x128/apps
+      - mkdir -pv /app/share/appdata
+      - cp -v *.appdata.xml /app/share/appdata
```
hfiguiere, line 101: did you test the build? From what I can gather, it won't be able to find the icon, desktop and appstream file.

Also usually files are installed using `install` which with `-D` make `mkdir` unecessary.

bundle/com.github.lruzicka.Needly.appdata.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.github.lruzicka.Needly.desktop</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>GPL-3.0</project_license>
+  <name>Needly</name>
+  <summary>openQA needle editor</summary>
+  <description>
+    <p>
+Needly is an openQA needle editor written in Python. It creates or modifies needles for the openQA tests.
```
hfiguiere, line 10: by reading this I have no clue what it does.

bundle/com.github.lruzicka.Needly.appdata.xml
```diff
@@ -0,0 +1,46 @@
...
+  <provides>
+    <binary>needly</binary>
+  </provides>
+  <recommends>
+    <control>pointing</control>
+    <control>keyboard</control>
+    <display_length side="longest" compare="ge">1400</display_length>
+  </recommends>
+  <releases>
+    <release version="2.5" date="2022-03-29">
```
hfiguiere, line 34: the tarball has version `2.5.5`

com.github.lruzicka.Needly.yaml
```diff
@@ -0,0 +1,113 @@
...
+      - --disable-gssapi
+      - --libdir=/app/lib
+    sources:
+      - type: archive
+        url: https://downloads.sourceforge.net/sourceforge/libtirpc/libtirpc-1.3.2.tar.bz2
+        sha256: e24eb88b8ce7db3b7ca6eb80115dd1284abc5ec32a8deccfed2224fc2532b9fd
+
+  # This is another dependency for libvirt which is used to
+  # create the necessary documentation.
+  - name: docutils
```
hfiguiere, line 49: can't you just disable building libvirt documentation?

hfiguiere (changes requested):

hfiguiere:
bot, build com.github.lruzicka.Needly

com.github.lruzicka.Needly.yaml
hfiguiere: This is the reason of the build failure.

bundle/.flatpak-builder/ccache/disabled/ccache.conf
lruzicka: I don't. These must be some files that my local build left there. I will delete it.

com.github.lruzicka.Needly.yaml
lruzicka: I copied this from my colleague's manifest at https://github.com/flathub/org.freefilesync.FreeFileSync and for my local build it worked ok. What is wrong about this and how shall I change it?

bundle/com.github.lruzicka.Needly.appdata.xml
lruzicka: Well, this is a highly specific application. When libvirt VM is running, the application can connect to it and take a screenshot from that VM. Then you can draw a rectangle (area) over the screenshot, provide some metadata and save it which creates a "needle", basically a png and a json file. These needles are used for openQA testing. People doing openQA should know what the purpose of the application is, the rest of the world will probably not be interested in it.

bundle/com.github.lruzicka.Needly.appdata.xml
lruzicka: I though I could cover the minor versions with a more general number. Will change it.

com.github.lruzicka.Needly.yaml
lruzicka: On my local build, Libvirt fails to build without documentation.

com.github.lruzicka.Needly.yaml
lruzicka: Because on the real openQA machine, the "needles" (files that this app edits) could be stored anywhere on the system, most frequently in `/var/lib`. 
On my machine, I maintain two repositories, one for testing and one for development - the majority of openQA users will have the same settings - so I need to be able to open the files from basically everywhere.

com.github.lruzicka.Needly.yaml
lruzicka: Ok. I did not know how to use these exactly and thought that both of the sockets (x11 and wayland) will make sure it works on both systems. I will change it to fit your suggestion.

README.md
lruzicka: Which means I need to do something to make it visible? How? Or it does not make sense to include the README.md in the package?

com.github.lruzicka.Needly.yaml
lruzicka: I think it needs it to be able to take the screenshots from a running VM. I can however try and compile without this flag, if it still works.

barthalion:
bot, build com.github.lruzicka.Needly

com.github.lruzicka.Needly.yaml
barthalion: To explain: builds happen offline purposely to enforce some basic reproducibility. Tools like https://github.com/flatpak/flatpak-builder-tools/blob/master/pip/flatpak-pip-generator can help you generate a fully offline manifest.

com.github.lruzicka.Needly.yaml
lruzicka: Maybe if I used `--filesystem=/var/lib/openqa`? Would that enable rw access to that directory and to the subdirectories?

com.github.lruzicka.Needly.yaml
lruzicka: Originally, I had poetry-core added to the manifest, but the local build insisted on installing it again from the pypi and would fail without this turned on. The only time I could build it was, when I let the application be installed from a local file (needle.py) added to the package and leaving pypi out of it. Would that make sense for you?

lruzicka: > To explain: builds happen offline purposely to enforce some basic reproducibility. Tools like https://github.com/flatpak/flatpak-builder-tools/blob/master/pip/flatpak-pip-generator can help you generate a fully offline manifest.

I will check, thanks.

lruzicka: So, I was able to use the pip-generator to get a correct json file and I do not longer need this.

com.github.lruzicka.Needly.yaml
lruzicka: The app does not need the `kvm` access to do what it should. I have removed it.

com.github.lruzicka.Needly.yaml
lruzicka: The system does not allow writing to `/var/lib` anyway, so I have removed the `host` access.

com.github.lruzicka.Needly.yaml
lruzicka: The application uses `tkinter` and `--socket=wayland` with `--socket=fallback-x11` fail with `no display` error. When I only use `--socket=x11` it works even on wayland, so I am just using x11 socket for now.

com.github.lruzicka.Needly.yaml
hfiguiere: you mean passing `-Ddocs=false` as a config opt to libvirt doesn't work?

com.github.lruzicka.Needly.yaml
```diff
@@ -0,0 +1,77 @@
...
+    buildsystem: simple
+    build-commands: 
+      - pip3 install --prefix=/app libvirt-python-8.1.0.tar.gz
+    sources: 
+      - type: file
+        url: https://files.pythonhosted.org/packages/2a/74/919462bed158ccc2a8061f31d82c157740c2e597ff50490526336e8aa688/libvirt-python-8.1.0.tar.gz
+        sha256: a21ecfab6d29ac1bdd1bfd4aa3ef58447f9f70919aefecd03774613f65914e43
+
+  # The Needly application
+  - python3-needly.json
```
hfiguiere, line 75: the main module for the application should be in the main manifest. alternative are difficult to maintain.

python3-needly.json
```diff
@@ -0,0 +1,37 @@
...
+    "name": "python3-needly",
+    "buildsystem": "simple",
+    "build-commands": [
+        "pip3 install --verbose --exists-action=i --no-index --find-links=\"file://${PWD}\" --prefix=${FLATPAK_DEST} \"needly\" --no-build-isolation",
+        "mkdir -pv /app/share/applications/",
+        "cp -v *.desktop /app/share/applications/",
+        "mkdir -pv /app/share/icons/hicolor/128x128/apps/",
+        "cp -v *.png /app/share/icons/hicolor/128x128/apps/",
+        "mkdir -pv /app/share/appdata/",
+        "cp -v *.desktop /app/share/appdata"
```
hfiguiere, line 11: You should use the explicit file name not a glob. And using `install -D` is still less commands and clearer.

bundle/com.github.lruzicka.Needly.appdata.xml
hfiguiere: I'm really not a fan of "if you don't understand then it is not for you". It feels like gatekeeping, and since the files are not coming from upstream, then it's a good occasion to have a much better description.

hfiguiere (changes requested):

hfiguiere:
And please change the PR title to reflect the app you are submitting.

com.github.lruzicka.Needly.yaml
lruzicka: > you mean passing `-Ddocs=false` as a config opt to libvirt doesn't work?

When doing `-Ddocs=false` the compilation complains about the keyword and suggest either `enable`, `disable`, or `auto`. I tried with `disable` and the documentation did not compile, but more and more errors started to appear, so I let it build and with it it works normally.

com.github.lruzicka.Needly.yaml
lruzicka: > the main module for the application should be in the main manifest. alternative are difficult to maintain.

I put the Needly module into the main manifest.

python3-needly.json
lruzicka: I tried to change this to `install -D` and stopped using globs.

lruzicka:
I also tried to change the description to make it more understandable.

hfiguiere:
bor, build io.github.lruzicka.Needly

lruzicka:
> bor, build io.github.lruzicka.Needly

You made a typo here. :)

lruzicka:
bot, build io.github.lruzicka.Needly

README.md
barthalion: It doesn't make sense to include it here. You should add it to your GitHub repo instead.

PR #2968, opened 2022-03-31
Add app.organicmaps.desktop

fgaz:

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link: https://github.com/organicmaps/organicmaps/issues/535**
  - well, organicmaps maintainers own it, not _me_
  - Only the wrapper is not upstreamed, but that will require a bigger patch that just those three lines in the manifest

### DIFF
--- a/app.organicmaps.desktop.yml
+++ b/app.organicmaps.desktop.yml
@@ -1,0 +1,51 @@
+app-id: app.organicmaps.desktop
+runtime: org.kde.Platform
+runtime-version: '5.15-21.08'
+sdk: org.kde.Sdk
+command: OMaps
+rename-desktop-file: OrganicMaps.desktop
+rename-icon: organicmaps
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+modules:
+  - name: rsync # used at build time
+    buildsystem: autotools
+    config-opts:
+      - --disable-md2man
+      - --disable-xxhash
+    sources:
+      - type: git
+        url: https://github.com/WayneD/rsync
+        tag: v3.2.3
+        commit: e94bad1c156fc3910f24e2b3b71a81b0b0bdeb70
+  - name: organicmaps
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+    post-install:
+      - 'mv ${FLATPAK_DEST}/bin/OMaps ${FLATPAK_DEST}/bin/.OMaps-wrapped'
+      - 'install -Dm755 ../wrapper.sh ${FLATPAK_DEST}/bin/OMaps'
+    cleanup:
+      - '/lib' # static libraries and build configs
+      - '/include'
+    sources:
+      - type: git
+        url: https://github.com/organicmaps/organicmaps
+        # TODO on next release, add tag: YYYY.MM.DD-N-android
+        commit: 186c3761c07e9a30e22d2d7890dfd61ad1cdd095
+      - type: shell
+        commands:
+          - 'bash configure.sh'
+      # Tell the program that the read-only and the read-write data locations
+      # are different, and create the read-write one.
+      - type: script
+        commands:
+          - 'data_path="${XDG_DATA_HOME:-${HOME}/.var/app/app.organicmaps.OrganicMaps/data}/OMaps"'
+          - 'mkdir -p "$data_path"'
+          - 'exec -a "$0" /app/bin/.OMaps-wrapped -resources_path /app/share/organicmaps/data -data_path "$data_path" "$@"'
+        dest-filename: 'wrapper.sh'


fgaz:
bot, build app.organicmaps.desktop

app.organicmaps.desktop.yml
```diff
@@ -0,0 +1,39 @@
+app-id: app.organicmaps.desktop
+runtime: org.kde.Platform
+runtime-version: '5.15'
```
hfiguiere, line 3:
```suggestion
runtime-version: '5.15-21.08'
```

app.organicmaps.desktop.yml
```diff
@@ -0,0 +1,39 @@
...
+      - type: shell
+        commands:
+          - 'bash configure.sh'
+      # Tell the program that the read-only and the read-write data locations
+      # are different, and create the read-write one.
+      - type: script
+        commands:
+          - 'data_path="${XDG_DATA_HOME:-${HOME}/.var/app/app.organicmaps.OrganicMaps/data}/OMaps"'
+          - 'mkdir -p "$data_path"'
+          - '/bin/OMaps -resources_path /share/organicmaps/data -data_path "$data_path"'
```
hfiguiere, line 38: there will be no `/bin/OMaps`. Nor will there be a `/share`. The install prefix is `/app`.

app.organicmaps.desktop.yml
```diff
@@ -0,0 +1,39 @@
+app-id: app.organicmaps.desktop
```
hfiguiere, line 1: `.desktop` is a bad idea. Really. It should be something like `app.organigcmaps.OrganicMaps`.

hfiguiere (changes requested):

app.organicmaps.desktop.yml
fgaz: Why is it a bad idea?

@biodranik would switching [back](https://github.com/organicmaps/organicmaps/pull/2206#discussion_r838309052) to `app.organigcmaps.OrganicMaps` be ok?

hfiguiere: because
1. .desktop confuse with desktop files. You end up with `something.something.desktop.desktop` as a desktop file name.
2. it is not an app name. The original issue you linked had it right.

(any example of this on flathub is wrong, but hard to fix)

app.organicmaps.desktop.yml
```diff
@@ -0,0 +1,39 @@
...
+runtime: org.kde.Platform
+runtime-version: '5.15'
+sdk: org.kde.Sdk
+command: OMaps-wrapper
```
hfiguiere, line 5:
```suggestion
command: OMaps-wrapper
rename-desktop-file: OrganicMaps.desktop
```

fgaz:
I've never seen this error before, and locally it does get past this stage, so I have no clue about what's happening here:

```
FB: Running: git rev-parse --verify --quiet c834d06c2739f63e28e5665d4569dca9415e3891:.gitmodules
FB: Running: git ls-tree fbebe70c7a4a67128cd94c6d35f29c7d65e34361 3party/icu/icu
FB: Running: git config submodule.3party/icu/icu.url file:///srv/buildbot/worker/build-x86_64-5/build/.flatpak-builder/git/https_github.com_unicode-org_icu.git
FB: Running: git submodule update --init 3party/icu/icu
Cloning into '/srv/buildbot/worker/build-x86_64-5/build/.flatpak-builder/build/organicmaps-1/3party/icu/icu'...
Downloading icu4j/main/shared/data/icudata.jar (14 MB)
Error downloading object: icu4j/main/shared/data/icudata.jar (44951f8): Smudge error: Error downloading icu4j/main/shared/data/icudata.jar (44951f88294c06e433a3b61238d9bb5f59ba01f091fcfb8fe4966f98f0748ef7): error transferring "44951f88294c06e433a3b61238d9bb5f59ba01f091fcfb8fe4966f98f0748ef7": [0] remote missing object 44951f88294c06e433a3b61238d9bb5f59ba01f091fcfb8fe4966f98f0748ef7
Errors logged to '/srv/buildbot/worker/build-x86_64-5/build/.flatpak-builder/build/organicmaps-1/.git/modules/3party/icu/icu/lfs/logs/20220331T162844.672775813.log'.
Use `git lfs logs last` to view the log.
error: external filter 'git-lfs filter-process' failed
fatal: icu4j/main/shared/data/icudata.jar: smudge filter lfs failed
fatal: Unable to checkout '904cf62457de2440e8526deb75c95a3f7296f517' in submodule path '3party/icu/icu'
]2;flatpak-builder: Cleanup organicmapsError: module organicmaps: Child process exited with code 1
```

app.organicmaps.desktop.yml
```diff
@@ -0,0 +1,39 @@
...
+      - 'install -Dm755 ../wrapper.sh ${FLATPAK_DEST}/bin/OMaps-wrapper'
+    cleanup:
+      - '/lib' # static libraries and build configs
+      - '/include'
+    sources:
+      - type: git
+        # TODO set to upstream when the pr is merged
+        url: https://github.com/fgaz/organicmaps
+        # TODO on next release, add tag: YYYY.MM.DD-N-android
+        commit: fbebe70c7a4a67128cd94c6d35f29c7d65e34361
```
hfiguiere, line 28: I don't see were the AppStream file is installed either. Not in the CMakefile.txt nor here. (the former is prefered, it should be installed regardless)

fgaz:
Let's see

bot, build app.organicmaps.desktop

app.organicmaps.desktop.yml
fgaz: Done, thanks

fgaz:
Nope, same error :-/

app.organicmaps.desktop.yml
biodranik: I don't understand the problem with .desktop, but if it's critical, `app.organicmaps.flathub` is also ok.

hfiguiere: it is not supposed to be flatpak / flathub specific. It's actually supposed to be the same if it's a deb or a rpm as this is the same ID used elsewhere... (like to register a dbus session bus, but I haven't figured out if you do or not, as it is optional with Qt).

**The last component of the app-id is supposed to be the name of the application** in some form.

See https://docs.flatpak.org/en/latest/conventions.html#application-ids

`app.organicmaps.OrganicMaps` is IMHO the best choice, the last part being consistent with your .desktop file. BTW .desktop files also use the app-id in their name, but that's something that gets fixed with the `rename-desktop-file`.

biodranik: Then `app.organicmaps` or `app.organicmaps.linux` should be ok.

hfiguiere: > Then `app.organicmaps` or `app.organicmaps.linux` should be ok.

Neither. The first one is too short, there shall be at least 3 components. The second, well, same as previously. "linux" isn't the name of the application.

fgaz: maybe we could find a compromise? eg. app.organicmaps.OrganicMapsDesktop both satisfies flathub requirements (domain.AppName) and has "desktop" in the name

hfiguiere: It fits the requirements, so yes.

barthalion: There are applications with `.desktop` suffix though, so it's not like the end of the world. Some developers tend to capitalize the last part, so it would be `app.organicmaps.Desktop` in this case.

barthalion:
git-lfs is available, but builds are done offline. Maybe multiple nested git submodules confuse flatpak-builder and it ends up not downloading these files.

```
Downloading icu4j/main/shared/data/icudata.jar (14 MB)
Error downloading object: icu4j/main/shared/data/icudata.jar (44951f8): Smudge error: Error downloading icu4j/main/shared/data/icudata.jar (44951f88294c06e433a3b61238d9bb5f59ba01f091fcfb8fe4966f98f0748ef7): error transferring "44951f88294c06e433a3b61238d9bb5f59ba01f091fcfb8fe4966f98f0748ef7": [0] remote missing object 44951f88294c06e433a3b61238d9bb5f59ba01f091fcfb8fe4966f98f0748ef7
Errors logged to '/srv/buildbot/worker/build-x86_64-2/build/.flatpak-builder/build/organicmaps-1/.git/modules/3party/icu/icu/lfs/logs/20220331T180418.682160461.log'.
Use `git lfs logs last` to view the log.
```

barthalion:
I think the problem is that flatpak-builder has no native support for git-lfs, so it becomes a problem once it's used in a submodule (as opposed to the directly used repository like Veloren does). The submodule itself is cloned, but LFS objects aren't.

biodranik:
LFS part of the ICU submodule is not required to build Organic Maps.

barthalion:
bot, build app.organicmaps.desktop

barthalion:
```
========================================================================
Building module organicmaps in /srv/buildbot/worker/build-x86_64-1/build/.flatpak-builder/build/organicmaps-1
========================================================================
]2;flatpak-builder: Building organicmapsFB: Running: cp -al /srv/buildbot/worker/build-x86_64-1/build/.flatpak-builder/git/https_github.com_fgaz_organicmaps /srv/buildbot/worker/build-x86_64-1/build/.flatpak-builder/build/organicmaps-1/.git
FB: Running: git config --bool core.bare false
FB: Running: git checkout fbebe70c7a4a67128cd94c6d35f29c7d65e34361
fatal: reference is not a tree: fbebe70c7a4a67128cd94c6d35f29c7d65e34361
```

@fgaz Can you repoint the manifest to upstream repo? Looks like your PR was merged and it broke the build.

barthalion:
bot, build app.organicmaps.desktop

barthalion:
Build has failed: https://flathub.org/builds/#/builders/12/builds/10136/steps/6/logs/stdio

fgaz:
That is a weird log. The git lfs error is still there, but the build somehow proceeds, _but_ it gets interrupted after a while for no apparent reason?!

barthalion:
I'm observing identical failure locally, so it's not some build infra quirk. The command used by buildbot can be replicated locally by executing `flatpak run org.flatpak.Builder --verbose --user --sandbox --force-clean --repo=repo --ccache --install-deps-from=flathub --default-branch=localtest --state-dir=/home/bpiotrowski/.cache/flatpak-builder --disable-updates --install --bundle-sources builddir app.organicmaps.desktop.yml`.

barthalion:
```
FAILED: opening_hours_integration_tests 
: && /usr/bin/ccache /run/ccache/bin/c++ -O2 -g -pipe -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -O3 -DNDEBUG -fuse-ld=gold    -fdiagnostics-color=always 3party/opening_hours/opening_hours_integration_tests/CMakeFiles/opening_hours_integration_tests.dir/Unity/unity_0_cxx.cxx.o -o opening_hours_integration_tests  libopening_hours.a && cd /run/build/organicmaps/_flatpak_build/3party/opening_hours/opening_hours_integration_tests && rsync -a /run/build/organicmaps/3party/opening_hours/opening_hours_integration_tests/opening-count.lst /run/build/organicmaps/_flatpak_build/
/bin/sh: line 1: rsync: command not found
```

I assume this is the error.

barthalion:
it fails appdata validation:

```
Error loading AppData file: failed to parse /app/share/appdata/app.organicmaps.desktop.appdata.xml: Error on line 24 char 29: <p> already set 'Organic Maps is a free Android & iOS offline maps app for travelers, tourists, hikers, and cyclists. It uses crowd-sourced OpenStreetMap data and is developed with love by' and tried to replace with ' ('
```

You can check it locally with CLI program `org.freedesktop.appstream-glib` available on Flathub.

hfiguiere:
one doesn't simply put an ampersand in an xml file.

fgaz:
bot, build app.organicmaps.desktop

fgaz:
Looks like `rename-desktop-file: OrganicMaps.desktop` was redundant, since the appstream file already specifies the .desktop file name

bot, build app.organicmaps.desktop

fgaz:
bot, build app.organicmaps.desktop

hfiguiere:
> > to me it looked more like the file was not found (not installed)
> 
> Looking at the log, it was installed and renamed from OrganicMaps.desktop to app.organicmaps.desktop, but the appstream file still specified that the name was OrganicMaps.desktop. I guess we'll see in the next build

So the build system rename the desktop file? (I probably missed it in the log) Then the appstream file should have the proper name as well, or the build with fail. `rename-desktop-file` takes care of that when actually renaming the file.

fgaz:
Ah I see what you mean now, sorry. Indeed it looks like the error came from the renaming, not from the appstream validation:

```
Renaming OrganicMaps.desktop to app.organicmaps.desktop.desktop
Error: Error moving file /srv/buildbot/worker/build-x86_64/build/.flatpak-builder/rofiles/rofiles-eouYf3/files/share/applications/OrganicMaps.desktop: No such file or directory
```

...and now I understand why: I forgot to rebase my branch :facepalm: 

bot, build app.organicmaps.desktop

fgaz:
> Error loading AppData file: failed to parse /app/share/appdata/app.organicmaps.desktop.appdata.xml: Error on line 25 char 29: <p> already set 'Organic Maps is a free Android & iOS offline maps app for travelers, tourists, hikers, and cyclists. It uses crowd-sourced OpenStreetMap data and is developed with love by' and tried to replace with ' ('

Weird. I don't get that error when running `org.freedesktop.appstream-glib validate` from flathub ([evidence](https://github.com/organicmaps/organicmaps/runs/5836861666?check_suite_focus=true)). Could it be that `org.freedesktop.appstream-glib` is outdated on the builders?

fgaz:
Yes it looks like I get the same (buggy?) errors if I downgrade from 0.7.19 to 0.7.18. Would it be possible to upgrade the validator on the builders?

barthalion:
Quite the contrary, we're using almost the latest commit from git. I will check what's going on locally.

barthalion:
OK, yeah, that build error comes from runtime, which might be shipping an older or unpatched appstream-glib release. It's not even the validation step, it fails on merging desktop and metadata files into appstream.

barthalion:
```
> flatpak run --command=appstream-compose org.freedesktop.appstream-glib  --prefix=files --origin=flatpak --basename=app.organicmaps.desktop app.organicmaps.desktop
Processing application app.organicmaps.desktop
Saving AppStream /var/home/bpiotrowski/dev/flathub/flathub/files/share/app-info/xmls/app.organicmaps.desktop.xml.gz
Done!
> flatpak run --filesystem=home --user --command=appstream-compose org.freedesktop.Platform//21.08  --prefix=files --origin=flatpak --basename=app.organicmaps.desktop app.organicmaps.desktop
Processing application app.organicmaps.desktop
Error loading AppData file: failed to parse files/share/metainfo/app.organicmaps.desktop.metainfo.xml: Error on line 25 char 29: <p> already set 'Organic Maps is a free Android & iOS offline maps app for travelers, tourists, hikers, and cyclists. It uses crowd-sourced OpenStreetMap data and is developed with love by' and tried to replace with ' ('
```

Yeah, that's definitely it. I will try to sort it out on our end.

barthalion:
https://github.com/flathub/org.flatpak.Builder/pull/80 should fix it, waiting for @nanonyme's review.

fgaz:
since

> we're using almost the latest commit from git

and https://github.com/flathub/org.flatpak.Builder/pull/80 was merged, I assume the update got through

bot, build app.organicmaps.desktop

barthalion:
bot, build app.organicmaps.desktop

barthalion:
OK, this time we have a legit error, in which CI failed to find a proper icon.

- The desktop file needs to specify `Icon` field following reverse DNS: https://github.com/fgaz/organicmaps/blob/appstream-validation/qt/res/OrganicMaps.desktop. This can be done with `desktop-file-edit` in post-install or build-commands.
- The `<launchable> file in the metainfo file needs to point to the actual desktop file, i.e. following reverse DNS notation: https://github.com/fgaz/organicmaps/blob/appstream-validation/qt/res/app.organicmaps.desktop.metainfo.xml#L96=

fgaz:
bot, build app.organicmaps.desktop

hfiguiere:
I don't think adding the `.desktop` file was the suggestion.

fgaz:
> I don't think adding the .desktop file was the suggestion.

what do you mean? In the last push I only added `rename-icon: organimaps` and a bigger icon

hfiguiere:
> > I don't think adding the .desktop file was the suggestion.
> 
> what do you mean? In the last push I only added `rename-icon: organimaps` and a bigger icon

nenvermind I mis-read it's the commit message in the notification that through me off.

And then you copied my typo :-(

fgaz:
> And then you copied my typo :-(

haha oops

bot, build app.organicmaps.desktop

app.organicmaps.desktop.yml
```diff
@@ -0,0 +1,50 @@
...
+  - name: rsync # used at build time
+    buildsystem: autotools
+    config-opts:
+      - --disable-md2man
+      - --disable-xxhash
+    sources:
+      - type: git
+        url: https://github.com/WayneD/rsync
+        tag: v3.2.3
+        commit: ec1abcfd29d98cc7a4864fd3a8ffc6245770ef0c
```
hfiguiere, line 24: for some reason the tag doesn't match the commit.

image

hfiguiere (changes requested):

app.organicmaps.desktop.yml
hfiguiere: ah it got changed your previous commit. It was correct before.

barthalion:
bot, build app.organicmaps.desktop

app.organicmaps.desktop.yml
fgaz: Oops, I did that commit in a rush and changed the wrong hash. Fixed.

bot, build app.organicmaps.desktop

barthalion:
It also needs more permissions:

```
LOG TID(1) INFO   2.2489e-05 platform/string_storage_base.cpp:26 StringStorageBase() Settings path: /var/home/bpiotrowski/.config/OMaps/settings.ini
LOG TID(1) WARNING 0.000155398 platform/string_storage_base.cpp:48 StringStorageBase() Loading settings: /var/home/bpiotrowski/.config/OMaps/settings.ini; Read; No such file or directory
```

It seems to hardcode `~/.config/OMaps` instead of following XDG_CONFIG_HOME. It probably saves offline maps somewhere, too.

fgaz:
Or, let's fix it upstream!

bot, build app.organicmaps.desktop

barthalion:
bot, build app.organicmaps.desktop

app.organicmaps.desktop.yml
```diff
@@ -0,0 +1,50 @@
+app-id: app.organicmaps.desktop
+runtime: org.kde.Platform
+runtime-version: '5.15-21.08'
+sdk: org.kde.Sdk
+command: OMaps-wrapper
```
barthalion, line 5: You need to use `OMaps-wrapper` also in the desktop file. It starts fine from CLI with `flatpak run`, but the desktop file fails to launch the app correctly.

fgaz: I solved (I think, let's see) the issue by naming the wrapper OMaps and renaming the original executable

fgaz:
bot, build app.organicmaps.desktop

app.organicmaps.desktop.yml
fgaz: looks like it's fixed! https://github.com/flathub/flathub/pull/2968#issuecomment-1103541977

barthalion:
/merge @biodranik

Please let us know if we should invite someone more from OrganicMaps team.

app.organicmaps.desktop.yml
bam80: Funny thing the app now can run on mobiles too, so the `.desktop` makes it really confusing IMO:
https://github.com/flathub/app.organicmaps.desktop/issues/39

PR #2981, opened 2022-04-04
Add org.purei.Play

Zer0xFF:

- [X] I am using only the minimal set of permissions. *(using --device=all for gamepad.)*
- [x] I am an upstream contributor to the project. ~~If not, I contacted upstream developers about submitting their software to Flathub. **Link:**~~
- [x] I own the domain used in the application ID or the domain has a policy for delegating subdomains (e.g. GitHub, SourceForge). (the owner/maintainer owns the domain and he's aware of this)

Side Not:
@jpd002 would also like access to the flathub repo, once created please

### DIFF
--- a/org.purei.Play.yaml
+++ b/org.purei.Play.yaml
@@ -13,7 +13,6 @@ finish-args:
   - --socket=pulseaudio
 modules:
   - "shared-modules/glu/glu-9.json"
-  - "shared-modules/glew/glew.json"
   - name: openal
     buildsystem: cmake-ninja
     builddir: true
--- a/org.purei.Play.yaml
+++ b/org.purei.Play.yaml
@@ -53,5 +53,5 @@ modules:
       - type: git
         url: https://github.com/jpd002/Play-.git
         branch: master
-        commit: eb061f6be52e68ede8f8094372eda66f9b7b925f
+        commit: 10a3d86288b2197580a18183872d6d54cc98c2e0
         disable-shallow-clone: true
--- a/org.purei.Play.yaml
+++ b/org.purei.Play.yaml
@@ -8,7 +8,7 @@ finish-args:
   - --share=network
   - --socket=wayland
   - --socket=fallback-x11
-  - --filesystem=host
+  - --filesystem=host:ro
   - --device=all
   - --socket=pulseaudio
 modules:
--- a/org.purei.Play.yaml
+++ b/org.purei.Play.yaml
@@ -46,7 +46,6 @@ modules:
             - -DUSE_GLES=yes
     post-install:
       - |
-        git fetch --tags --force
         GIT_VERSION=$(git describe --always);
         GIT_DATE=$(git show -s --format=%cs);
         sed -i 's|<releases>|<releases>\n    <release date="'${GIT_DATE}'" version="'${GIT_VERSION}'"/>|' ${FLATPAK_DEST}/share/metainfo/org.purei.Play.metainfo.xml;
@@ -55,3 +54,4 @@ modules:
         url: https://github.com/jpd002/Play-.git
         branch: master
         commit: eb061f6be52e68ede8f8094372eda66f9b7b925f
+        disable-shallow-clone: true
--- a/org.purei.Play.yaml
+++ b/org.purei.Play.yaml
@@ -8,7 +8,10 @@ finish-args:
   - --share=network
   - --socket=wayland
   - --socket=fallback-x11
-  - --filesystem=host:ro
+  - --filesystem=home:ro
+  - --filesystem=/mnt:ro
+  - --filesystem=/media:ro
+  - --filesystem=/run/media:ro
   - --device=all
   - --socket=pulseaudio
 modules:
--- a/org.purei.Play.yaml
+++ b/org.purei.Play.yaml
@@ -1,0 +1,58 @@
+app-id: org.purei.Play
+runtime: org.kde.Platform
+runtime-version: '5.15'
+sdk: org.kde.Sdk
+command: Play
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --filesystem=host
+  - --device=all
+  - --socket=pulseaudio
+modules:
+  - "deps/Dependencies/shared-modules/glu/glu-9.json"
+  - "deps/Dependencies/shared-modules/glew/glew.json"
+  - name: openal
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DALSOFT_REQUIRE_PULSEAUDIO=ON
+      - -DALSOFT_BACKEND_PIPEWIRE=off
+    cleanup: 
+        - /bin
+        - /include
+        - /lib/pkgconfig
+        - /share
+    sources:
+      - type: git
+        url: https://github.com/kcat/openal-soft.git
+  - name: evdev
+    config-opts: 
+      - --disable-static
+    cleanup: 
+      - /bin
+      - /include
+      - /lib/pkgconfig
+      - /lib/*.la
+      - /share/man
+    sources:
+      - type: archive
+        url: https://freedesktop.org/software/libevdev/libevdev-1.5.7.tar.xz
+        sha256: a1e59e37a2f0d397ffd7e83b73af0e638db83b8dd08902ef0f651a21cc1dd422
+  - name: Play!
+    buildsystem: cmake-ninja
+    config-opts:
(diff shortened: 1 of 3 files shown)


org.purei.Play.yaml
```diff
@@ -0,0 +1,58 @@
+app-id: org.purei.Play
+runtime: org.kde.Platform
+runtime-version: '5.15'
```
hfiguiere, line 3:
```suggestion
runtime-version: '5.15-21.08'
```

org.purei.Play.yaml
```diff
@@ -0,0 +1,58 @@
...
+sdk: org.kde.Sdk
+command: Play
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --filesystem=host
```
hfiguiere, line 11: that's a very broad filesystem permission

hfiguiere (changes requested):

hfiguiere:
bot, build org.purei.Play

org.purei.Play.yaml
```diff
@@ -0,0 +1,58 @@
...
+  - --device=all
+  - --socket=pulseaudio
+modules:
+  - "deps/Dependencies/shared-modules/glu/glu-9.json"
+  - "deps/Dependencies/shared-modules/glew/glew.json"
```
hfiguiere, line 16:
```suggestion
  - shared-modules/glu/glu-9.json
  - shared-modules/glew/glew.json
```

org.purei.Play.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - -DALSOFT_REQUIRE_PULSEAUDIO=ON
+      - -DALSOFT_BACKEND_PIPEWIRE=off
+    cleanup: 
+        - /bin
+        - /include
+        - /lib/pkgconfig
+        - /share
+    sources:
+      - type: git
+        url: https://github.com/kcat/openal-soft.git
```
hfiguiere, line 31: you need a specific commit/tag

hfiguiere:
see my suggestion as to why it's not building.

org.purei.Play.yaml
Zer0xFF: i was meant to make this `--filesystem=host:ro` if you have different suggestion, I'd like to hear it.
my decision was based on fact that the user could have their games stored anywhere including external storage.

org.purei.Play.yaml
```diff
@@ -0,0 +1,59 @@
...
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --filesystem=host:ro
+  - --device=all
+  - --socket=pulseaudio
+modules:
+  - "shared-modules/glu/glu-9.json"
+  - "shared-modules/glew/glew.json"
```
j8r, line 16: Not needed

org.purei.Play.yaml
```diff
@@ -0,0 +1,59 @@
...
+    cleanup: 
+      - /bin
+      - /include
+      - /lib/pkgconfig
+      - /lib/*.la
+      - /share/man
+    sources:
+      - type: archive
+        url: https://freedesktop.org/software/libevdev/libevdev-1.5.7.tar.xz
+        sha256: a1e59e37a2f0d397ffd7e83b73af0e638db83b8dd08902ef0f651a21cc1dd422
```
j8r, line 45:
```suggestion
      - type: archive
        url: https://www.freedesktop.org/software/libevdev/libevdev-1.12.1.tar.xz
        sha256: 1dbba41bc516d3ca7abc0da5b862efe3ea8a7018fa6e9b97ce9d39401b22426c
        x-checker-data:
          type: anitya
          project-id: 20540
          stable-only: true
          url-template: https://www.freedesktop.org/software/libevdev/libevdev-$version.tar.xz
```

org.purei.Play.yaml
Zer0xFF: you're right, thanks

org.purei.Play.yaml
j8r: Is it needed @jpd002? I haven't tested much, it did compile on my end without.
I suppose either it is optional or the project build it on its own.

org.purei.Play.yaml
Zer0xFF: is this better?

org.purei.Play.yaml
Zer0xFF: the project doesnt build this no, thought to my surprise just built this locally and it worked with out (aka its part of the kde sdk `Found OpenAL: /usr/lib/x86_64-linux-gnu/libopenal.so`)

org.purei.Play.yaml
j8r: I'd say just `host:ro` at this point. The most sensitive data, which is likely in the home directory, can already be read.

The permissions will be simpler, and people can always restrict more/change to `home` later with e.g. Flatseal.

j8r:
bot, build org.purei.Play

org.purei.Play.yaml
hfiguiere: for removable storage devices there is `/run/media` and `/media`.

j8r: `/mnt` makes sense since some people (like me) mount disks permanently there

org.purei.Play.yaml
```diff
@@ -0,0 +1,50 @@
...
+          url-template: https://www.freedesktop.org/software/libevdev/libevdev-$version.tar.xz
+  - name: Play!
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+    post-install:
+      - |
+        GIT_VERSION=$(git describe --always);
+        GIT_DATE=$(git show -s --format=%cs);
+        sed -i 's|<releases>|<releases>\n    <release date="'${GIT_DATE}'" version="'${GIT_VERSION}'"/>|' ${FLATPAK_DEST}/share/metainfo/org.purei.Play.metainfo.xml;
```
j8r, line 45: The existing version can be replaced, it is essentially a place holder.
Also there is no version, I don't know what `$GIT_VERSION` will be.
Maybe call it `nightly`.

Zer0xFF: GIT_VERSION is meant to be `VERSION-HASH` but its not correct in the build, probably because build system doesnt pull all the tags, latest comment should fix that.

Zer0xFF:
bot, build org.purei.Play

org.purei.Play.yaml
j8r: A commit hash is a developer information, not really one for the end user.
`nightly`, `latest` or similar will look better, and will actually mean something for them.
Note that the commit is already present in the Flatpak metadata (`flatpak info`).

Just in a single line is at the end needed in `post-install`:
`sed -i "s/date=.* v/date=\"$(git show -s --format=%cs)\" version=\"nightly\"/" ${FLATPAK_DEST}/share/metainfo/org.purei.Play.metainfo.xml`

Zer0xFF: this is the versioning we show within the app, and the one we expect during reports

j8r: Ok. Why still keeping the existing release instead of replacing it, which will likely not be updated (unless you handle that).

Zer0xFF: its just easier to remove that line upstream than worry about it from this side.
its also easier to fix the current issue (this way) on this side than upstream (for now), I think this entire post-install can be moved upstream into the cmake script itself.

j8r: Maybe but then it won't pass the `flatpak run org.freedesktop.appstream-glib validate org.purei.Play.metainfo.xml` linter.
It can be considered as a placeholder.

Note also that versions **must** be sorted. Commit hashes not being version, CI errors on the Flathub side may happen in the future if older releases are kept.,

hfiguiere: IMHO upstream should be the proper version when doing a release. Remember, flathub is to distribute releases.

Zer0xFF:
bot, build org.purei.Play

org.purei.Play.yaml
Zer0xFF: even upstream version is set per commit/build, so it works out the same, Im merely saying that managing this upstream makes more sense, since it simplified the script on this side, but it'd also technically needed for appimage so no need for duplicate work, though this is something I'd do down the line (busy for the next few weeks)

j8r: Anyway this shouldn't block the merge?
As @jpd002 whish, this Flatpak package should track master.
@Zer0xFF and @jpd002 what do you think of version like `0.50-10a3d86` (version-shorthash)? That's what [xemu](https://github.com/mborgerson/xemu/releases) does.

PR #2988, opened 2022-04-06
Add com.riverbankcomputing.PyQt.BaseApp

JakobDev:

This is a BaseApp which contains PyQt. It used by many Python Programs. It is not easy to build and take some time to build (around 10 Minutes).  This BaseApp currently has PyQt6. PyQt5 may be added in the future. I'm also planing to make a second BaseApp which is based on [io.qt.qtwebengine.BaseApp](https://github.com/flathub/io.qt.qtwebengine.BaseApp) and contains Qt Web Engine.

### DIFF
--- a/com.riverbankcomputing.PyQt.BaseApp.yaml
+++ b/com.riverbankcomputing.PyQt.BaseApp.yaml
@@ -1,0 +1,96 @@
+app-id: com.riverbankcomputing.PyQt.BaseApp
+runtime: org.kde.Platform
+runtime-version: "6.2"
+sdk: org.kde.Sdk
+command: /usr/bin/python
+finish-args:
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --share=ipc
+
+modules:
+  - name: pyqt6
+    modules:
+
+      - name: sip
+        buildsystem: simple
+        build-commands:
+          - pip3 install --no-deps --no-use-pep517 --prefix=$FLATPAK_DEST .
+        sources:
+          - type: archive
+            url: https://pypi.python.org/packages/source/s/sip/sip-6.4.0.tar.gz
+            sha256: 42ec368520b8da4a0987218510b1b520b4981e4405086c1be384733affc2bcb0
+        modules:
+
+          - name: python-toml
+            buildsystem: simple
+            build-commands:
+              - pip3 install --no-deps --no-use-pep517 --prefix=$FLATPAK_DEST .
+            sources:
+              - type: archive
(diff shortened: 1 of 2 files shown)


JakobDev:
bot, build com.riverbankcomputing.PyQt.BaseApp

hfiguiere:
how do you provide an that needs both QtWebEngine and PyQt?

com.riverbankcomputing.PyQt.BaseApp.yaml
```diff
@@ -0,0 +1,96 @@
+app-id: com.riverbankcomputing.PyQt.BaseApp
+runtime: org.kde.Platform
+runtime-version: "6.2"
+sdk: org.kde.Sdk
+command: /usr/bin/python
```
hfiguiere, line 5: Command is NEVER an aboslute path

JakobDev: This is a BaseApp and never meant to run directly. This is just a placeholder.

hfiguiere:
so you want to maintain 2^4 packages?

JakobDev:
> so you want to maintain 2^4 packages?

All of my programs use PyQt, one PyQt with QtWebengine. This means I have to maintain this anyway. It might be better, to use a separate yml with contains PyQt, so I have to edit only this file and could use git submodules. For PyQt5: Since Qt5 itself is EOL, if you are not a paying customer which has longer support, there will not be much to do here. Only Bugfix updates, which can be delegated to flatpak-external-data-checker.

hfiguiere:
see https://github.com/flathub/shared-modules/pull/145

Also if these modules were not so impossible to build reliably, and if the tarball didn't disappear. but that's an upstream problem.

com.riverbankcomputing.PyQt.BaseApp.yaml
barthalion: I think `command` field is optional though. You can just remove it.

barthalion:
Maintaining two base apps doesn't seem reasonable. I would prefer a single BaseApp with both PyQt and QtWebEngine.

Also please sync with @tinywrkb about your work.

tinywrkb:
@JakobDev my WIP PyQtWebEngine baseapp is [here](https://github.com/tinywrkb/flathub/tree/com.riverbankcomputing.PyQtWebEngine.BaseApp).  
I'm in the middle of polishing it, and sending a new app PR is next on my todo list, but not all the bits that I want in are in place yet.

JakobDev:
I'm closing this in favour of #2998

PR #3012, opened 2022-04-15
Add org.freedesktop.Sdk.Extension.qt6

4c0n:
In order to prevent Qt Creator from rebuilding Qt6 at every change and to make the build of Qt6 more re-usable, I have split off the modules in this Sdk Extension.
It will reduce build time of Qt Creator as requested. It is being used here: https://github.com/flathub/io.qt.QtCreator/blob/use-extension/io.qt.QtCreator.yaml and in order to merge that to the beta branch and test it, I need this package to be available first. 


Erick555:
See https://github.com/flathub/flathub/pull/3013#issuecomment-1100066897

4c0n:
@Erick555 The case is different here, Qt6 needs to be available at all times in the Qt Creator package and not be allowed to  be uninstalled. It's currently part of the Qt Creator package itself and it's causing the build times of that package to be massive and isn't reusable.

Erick555:
Then it absolutely can't be an extension. Extension are optional by flatpak design. You can either bundle it or make a baseapp. Or even better ask for adding missing modules in runtime.

4c0n:
This way it can be separated, decreasing build times and frequency and be useful to others.

Erick555:
There are better ways to solve the problem. You chose one that doesn't fit. This is my point.

4c0n:
We disagree on that.

Erick555:
Ok, that's fine. Let's hear opinion from others.

hfiguiere:
is the build relocatable? I'm not even sure. See my other comment.

Erick555:
There is also stability concern regarding to this one as in current form it's going to ship always latest Qt 6.x release and some apps/tools using it may need adjustments for new version especially if they selectively copy extension parts to avoid mismatches. KDE splits every major release (5.x, 6.x).

4c0n:
According to semantic versioning that would be a minor release and it should be backwards compatible, however in Qt's case it might not always be. However it would be fairly easy to overcome this issue.

4c0n:
bot, build org.freedesktop.Sdk.Extension.qt6

DioEgizio:
Wanted to say that if this gets merged it'd be epic and highly beneficial for projects using qt6

hfiguiere:
> Wanted to say that if this gets merged it'd be epic and highly beneficial for projects using qt6

There is a qt6 sdk/runtime already. This is something else.

4c0n:
It actually would be beneficial IMO as it would provide a complete Qt6 package that allows you to use the parts you need without having to figure out how to keep your own Qt modules compatible (ABI) and in sync with the runtime. It would also reduce build times for every app that uses it.

Thank you for expressing your support.

Erick555:
Mentioned use case is exactly why it was [rejected](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/356#note_98819362) in the past so things are going in circle here:



> Using an SDK extension for dependencies is a really really bad UX.
>
> *  All users of it have to fiddle with paths manually
> *  All users of it have to copy the libs into their app
> * They lose automatic updates of dependencies for fixes
> * It makes deduplication less effective
> * The runtime still has to have the Qt extension points itself anyway for themes, platform modules, etc

4c0n:
Yup you shared one person's opinion and it's not even the actual reason that request was rejected, also we are capable of making other decisions than the Freedesktop project and don't all have to share the same opinion. I'd say those points are pretty debatable as well.
I however have lost hope in this becoming a thing and wish not to waste my time on these type of discussions on PR's with people that are not responsible for merging or denying it, glad you have so much time to spare and I wish you luck with trying to change peoples minds on things where you disagree.

Erick555:
This opinion was from flathub organization [member](https://github.com/orgs/flathub/people).

All I tried is to convince you are wasting your time but somewhat you think I'm wasting it which is quite irony. Anyway the damage was done.

My last suggestion is to use #flatpak matrix channel to get straight to face opinion from flathub members on the matter.

da2ce7:
@4c0n @Erick555 @hfiguiere https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/356#note_1186990257

PR #3013, opened 2022-04-15
Add org.freedesktop.Sdk.Extension.qt5

4c0n:
This is to be used in an App extension for Qt Creator so that Qt5 will become optional/uninstallable. And needs to be available so that: https://github.com/flathub/io.qt.QtCreator/blob/use-extension/io.qt.QtCreator.yaml can be merged to beta and tested.


4c0n:
bot, build org.freedesktop.Sdk.Extension.qt5

Erick555:
I think making an Sdk extension for one app to consume may be questionable especially if you are making a [stub app extension](https://github.com/flathub/flathub/pull/3014) to use this one. Just merging the former into the latter would suffice and make things easier for you.

Long time ago it was decided that maintaining qt belongs to KDE team as they're most competent to do so. At same time Qt as Sdk extension has very limited usecase which may be not worth the effort.

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64"]
```
hfiguiere, line 2: wait what?

Erick555: It's because io.qt.QtCreator is x86 only and this extension targets it. I raised my concern about this approach above.

4c0n:
> I think making an Sdk extension for one app to consume may be questionable especially if you are making a separate [stub app extension](https://github.com/flathub/flathub/pull/3014) to use this one. Just merging the former into the latter would suffice and make things easier for you.

It usually has to start somewhere, there are more and more separate Qt components showing up on flathub, which could be reused from the Sdk extension.

> 
> Long time ago it was decided that maintaining qt belongs to KDE team as they're most competent to do so. At same time Qt as Sdk extension has very limited usecase which may be not worth the effort.

KDE uses Qt, however the SDK and platform they provide is incomplete and tailored to KDE. There are usecases where one would like to combine Qt with Gtk which is now not easy to accomplish.
Personally I think it would be better if the Sdk platform would use a extension to provide a lot more flexibility.

flathub.json
4c0n: Should not be an issue to open it up for other architectures.

Erick555:
> It usually has to start somewhere, there are more and more separate Qt components showing up on flathub, which could be reused from the Sdk extension.

Like which ones? Some apps need Qt forks (even QtCreator itself) so extension doesn't help them. Moreover Sdk extension doesn't help them if they hard depend on Qt and need it either in Platform or /app (copying whole extension at build time is pretty ugly). If you can provide examples of apps beside yours for which this extesnin would be useful then it can make better case for it.

> KDE uses Qt, however the SDK and platform they provide is incomplete and tailored to KDE. There are usecases where one would like to combine Qt with Gtk which is now not easy to accomplish.

KDE runtimes have both Qt and GTK already so I don't understand that part. Did you tried asking for adding missing modules in https://invent.kde.org/packaging/flatpak-kde-runtime . Isn't it the place to start?

4c0n:
> > It usually has to start somewhere, there are more and more separate Qt components showing up on flathub, which could be reused from the Sdk extension.
> 
> Like which ones? Some apps need Qt forks (even QtCreator itself) so extension doesn't help them. Moreover Sdk extension doesn't help them if they hard depend on Qt and need it either in Platform or /app (copying whole extension at build time is pretty ugly). If you can provide examples of apps beside yours for which this extesnin would be useful then it can make better case for it.
> 

Qt Creator does not require any Qt forks, where did you get that from? Which apps require Qt forks? The KDE ones perhaps? Which is kind of the point here isn't it? Actually it's not that different from other Sdk extensions, also you can copy whichever part you require, no need to copy the whole thing (unless you need the whole thing, like in the case of Qt Creator). Like I said you could actually build the KDE Sdk out of modules of the extension if you wanted. It's really not that difficult to think of/find use cases, but you seem to have a pretty strong opinion already ¯\_(ツ)_/¯ 

> > KDE uses Qt, however the SDK and platform they provide is incomplete and tailored to KDE. There are usecases where one would like to combine Qt with Gtk which is now not easy to accomplish.
> 
> KDE runtimes have both Qt and GTK already so I don't understand that part. Did you tried asking for adding missing modules in https://invent.kde.org/packaging/flatpak-kde-runtime . Isn't it the place to start before making your own thing?

There is a clear separation of runtimes here: https://docs.flatpak.org/en/latest/available-runtimes.html besides sometimes multiple Qt versions are required, which cannot be done nicely using the Sdk. I've tried speaking about that and didn't get very far. https://github.com/flathub/io.qt.QtCreator/issues/40
Also it would make sense not wanting to pack certain things into the KDE sdk, as it's KDE, not Qt. The licensing also differs for certain modules, so they might not want to use them for KDE purposes. Qt is not (just) KDE.

4c0n:
bot, build org.freedesktop.Sdk.Extension.qt5

flathub.json
hfiguiere: QtCreator being x86_64 only is another question I had elsewhere on a higher level.

Erick555:
> Qt Creator does not require any Qt forks, where did you get that from? Which apps require Qt forks? The KDE ones perhaps

yeah, sorry it was llvm fork I had in mind so Qt Creator isn't valid example. telegram is example of app which needs [patched Qt](https://github.com/desktop-app/patches/tree/22629a5df509f9833ee98de3aaa0bd9e265da709).

No, neither of KDE or their apps don't need forked/patched Qt. The [runtime patches](https://invent.kde.org/packaging/flatpak-kde-runtime/-/tree/qt5.15lts-21.08/patch) are flatpak specific - they fix Qt behavior in flatpak for all apps. This also means your extensions without those patches can be broken.

> Actually it's not that different from other Sdk extensions, also you can copy whichever part you require, no need to copy the whole thing (unless you need the whole thing, like in the case of Qt Creator).

Or you can build the missing parts from runtime and forget about extension.

> It's really not that difficult to think of/find use cases, but you seem to have a pretty strong opinion already

I didn't find those cases. If you know so many of them then you may just provide few and we will be done with this.

> besides sometimes multiple Qt versions are required, which cannot be done nicely using the Sdk. I've tried speaking about that and didn't get very far. flathub/io.qt.QtCreator#40

Creator is the only app that needs multiple Qt versions which is why it got attention from admins in first place.

> Also it would make sense not wanting to pack certain things into the KDE sdk, as it's KDE, not Qt. The licensing also differs for certain modules, so they might not want to use them for KDE purposes. Qt is not (just) KDE.

KDE runtimes are for general purpose in flathub for all Qt apps. I don't know from where you got impression that KDE runtime is only for KDE apps.

hfiguiere:
I wonder of the usefulness of this at all for anything. In general the build isn't relocatable, and we alredy have issues with QtWebEngine in a different prefix (pyside doesn't like it). So it would be pretty much pointless to use.

4c0n:
> I wonder of the usefulness of this at all for anything. In general the build isn't relocatable, and we alredy have issues with QtWebEngine in a different prefix (pyside doesn't like it). So it would be pretty much pointless to use.

It is relocatable, I'm doing just that in the app extension for Qt Creator (and inside Qt Creator for Qt6). The paths can be altered by adding a qt.conf file in the bin directory.
Also regarding usefulness, it would allow for development using Qt in other IDE's for example, like Gnome builder or Visual Studio.

org.freedesktop.Sdk.Extension.qt5.yaml
```diff
@@ -0,0 +1,790 @@
...
+      - -no-rpath
+      - -no-directfb
+      - -no-kms
+      - -system-proxies
+      - -gtk
+      - -no-use-gold-linker
+    sources:
+      - type: git
+        url: git://code.qt.io/qt/qtbase.git
+        tag: v5.15.3-lts-lgpl
```
Erick555, line 50: Note that Qt company stopped making new releases of v5.15 for general public over one year ago so this version is abandoned. 5.15 is still supported by KDE who cherry-picks patches there.

4c0n: I don't see much of a point for development with the fork. So then the applications people would build with them would have to be linked to the fork and I'm assuming most distros do not ship that fork, limiting it's usefulness. It's not just for development of apps for flathub or using the KDE runtime in flatpak in general (in which case it would be a good idea though).

Erick555: At least ArchLinux and Fedora use fork. I don't know about other distros but most would prefer something active unless they're itself in some maintenance mode. Same question is for flathub - should it provide supported or unsupported software versions?

4c0n: Fedora does not use a (or the KDE) fork:
 https://src.fedoraproject.org/rpms/qt5-qtbase/blob/rawhide/f/qt5-qtbase.spec

they use patches where necessary.

Erick555: They [extract all patches from KDE fork](https://src.fedoraproject.org/rpms/qt5-qtbase/blob/f36/f/qt5-qtbase.spec#_142) and apply them on top of pristine Qt release which essentially means they use the fork. Do we really need to discuss about facts here?

4c0n:
org.freedesktop.Sdk.Extension.qt5

4c0n:
bot, build org.freedesktop.Sdk.Extension.qt5

4c0n:
bot, build org.freedesktop.Sdk.Extension.qt5

da2ce7:
@4c0n @Erick555 @hfiguiere https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/356#note_1186990257

PR #3014, opened 2022-04-15
Add io.qt.QtCreator.Extension.Qt5

4c0n:
This will make Qt5 optional for Qt Creator and needs to be available before we can merge: https://github.com/flathub/io.qt.QtCreator/blob/use-extension/io.qt.QtCreator.yaml to beta and test it.


hfiguiere:
on a higher level, why is this x86_64 only?

4c0n:
> on a higher level, why is this x86_64 only?

In the past the Qt Creator builds for other architectures would fail on the build platform because of resource limitations, out of memory or out of disk space was a common theme. I could try and open it up again and see if those issues are still there if you'd like.

hfiguiere:
On the builders? The builders can handle it.

4c0n:
They couldn't in the past, but I have zero problems with trying it again as long as there is some way of getting support if the worker fails for reasons like the above ones.

PR #3015, opened 2022-04-15
Add se.manyver.Manyverse

atomrc:
Adds the manyverse app to flathub (see https://www.manyver.se/)


se.manyver.Manyverse.yaml
```diff
@@ -0,0 +1,63 @@
...
+rename-desktop-file: manyverse.desktop
+rename-icon: manyverse
+finish-args:
+  - --socket=x11
+  - --socket=wayland
+  - --socket=pulseaudio
+  - --share=network
+  - --allow=bluetooth
+  # full filesystem access is needed in order for the user to be able to browse and upload files from their filesystem
+  - --filesystem=home:ro
```
atomrc, line 18: We need a ro access to the home directory in order for the user to be able to browse his own files when uploading something to the application

atomrc:
bot, build se.manyver.Manyverse

staltz:
Hi FlatHub maintainers, I just want to note that @atomrc is part of our team, so  his PR qualifies as a "maintainer".

If/when this is accepted, could you also add me @staltz to the newly created github repo?

flathub.json
```diff
@@ -0,0 +1,3 @@
+{    
+    "only-arches": ["x86_64"]
```
hfiguiere, line 2: I had comments: why is this x86_64 only?

se.manyver.Manyverse.yaml
```diff
@@ -0,0 +1,63 @@
...
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+command: manyverse
+separate-locales: false
+rename-desktop-file: manyverse.desktop
+rename-icon: manyverse
+finish-args:
+  - --socket=x11
+  - --socket=wayland
```
hfiguiere, line 13: does it support wayland?

Also even as fallback `--share=ipc` is needed for x11

se.manyver.Manyverse.yaml
```diff
@@ -0,0 +1,63 @@
...
+  - --socket=pulseaudio
+  - --share=network
+  - --allow=bluetooth
+  # full filesystem access is needed in order for the user to be able to browse and upload files from their filesystem
+  - --filesystem=home:ro
+  - --filesystem=xdg-config/manyverse
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.freedesktop.portal.Background
+  - --own-name=org.kde.*
```
hfiguiere, line 22: this is a very broad dbus own.

flathub.json
atomrc: This constraint might indeed not be needed.

se.manyver.Manyverse.yaml
atomrc: This is a part where, honestly, I am not actually sure. It's an electron app, does it make sense to add the wayland socket for an electron app, I am not sure. I'd appreciate any suggestion there :)

se.manyver.Manyverse.yaml
atomrc: That's also good point, I might not even need an own name. How do I know exactly if I need a own name?

atomrc:
@hfiguiere Thanks for you review. Quick question, shall I make the changes in the newly create repo now?

se.manyver.Manyverse.yaml
hfiguiere: in general Electron doesn't support wayland.

hfiguiere:
yes. nothing else will happen here.

atomrc:
It seems that the build is currently failing on flathub. I am not sure why it seems to be at the upload step with the error 
`Api call to https://hub.flathub.org/api/v1/build/84342/upload failed with status 408, details: {'status': 408, 'error-type': 'no-error', 'message': 'No json error details from server'}`

here https://flathub.org/builds/#/builders/35/builds/10346

Is it something I need to fix on my end?

atomrc:
@hfiguiere I have made the changes to the manyverse repo. But we seem to have deployment errors on the build script. 
Seems like the build for aarch64 architecture is hanging https://flathub.org/builds/#/builders/32/builds/86510

Any idea what is happening?

hfiguiere:
there is no source for aarch64.... but it'll try to build it. (the question was not about removing the flathub.json but why this wasn't available on aarch64, and it seems there never was an answer)

hfiguiere:
technically the build failed because `ar: manyverse*.deb: No such file or directory` on aarch64.

atomrc:
Right, makes sense. Sorry I got a little confused. So yeah we do not have, on manyverse side, a job that builds an aarch64 binary.
Until then it's probably safer to only have the x86 build, WDYT?

hfiguiere:
why not building the flatpak from source?

staltz:
> why not building the flatpak from source?

Initially, another contributor did it from the .deb and we kept that method because it comes with the `.desktop` file ready.

About aarch64, it might be challenging to build because we have node.js native modules (leveldown, sodium-native, etc) which may fail compilation for various reasons. We want to support aarch64 eventually, but maybe it's not worth forcing it at this moment. We can work iteratively.

hfiguiere:
> > why not building the flatpak from source?
> 
> Initially, another contributor did it from the .deb and we kept that method because it comes with the `.desktop` file ready.

If you don't manage the .destkop in you source repository it's already a bug.

> 
> About aarch64, it might be challenging to build because we have node.js native modules (leveldown, sodium-native, etc) which may fail compilation for various reasons. We want to support aarch64 eventually, but maybe it's not worth forcing it at this moment. We can work iteratively.

You can start by building from source here.

staltz:
Yup, we have hundreds of bugs, this is just one more. Have to prioritize them, and this .desktop management one is not the worst bug.

PR #3022, opened 2022-04-16
Add io.github.garglk.Gargoyle

cspiegel:

There are two permissions which probably require explanation:

1. `--filesystem=home`
    * While Gargoyle, via Qt, can use portals to open files in general, there are sometimes auxiliary files which need to be loaded simultaneously, but which are not specified by the user; instead, they're searched for in the same directory as the "main" file, and as such, C and C++ I/O is directly used to open and read them.
2. `--socket=system-bus`
    * Without this, the file picker on Plasma dumps the following to stderr:
    ```
    kf.solid.backends.udisks2: Failed enumerating UDisks2 objects: "org.freedesktop.DBus.Error.Disconnected"
     "Not connected to D-Bus server"
    ```
    Despite this, as far as I can tell, all disks are still visible in the file chooser, so I'm not entirely sure this is anything but a red herring.  But I'd rather get feedback on this before removing it, in case it's something that's been seen before.

A quick note on the file `fdo.patch`: this fixes a few things up to conform to freedesktop.org standards. This is currently in a branch upstream that I'll merge after Gargoyle goes through the Flatpak process, so that the two are in sync.


io.github.garglk.Gargoyle.yaml
```diff
@@ -0,0 +1,31 @@
...
+command: gargoyle
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=x11
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --socket=pulseaudio
+  - --device=dri
+  - --socket=system-bus
```
hfiguiere, line 14: no. you should figure out which d-bus name it needs. use `--log-system-bus` with `flatpak run` to figure out what is needed.

io.github.garglk.Gargoyle.yaml
```diff
@@ -0,0 +1,31 @@
+id: io.github.garglk.Gargoyle
+runtime: org.kde.Platform
+runtime_version: "5.15-21.08"
+sdk: org.kde.Sdk
+command: gargoyle
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=x11
+  - --socket=fallback-x11
```
hfiguiere, line 10:
```suggestion
  - --socket=fallback-x11
```

fdo.patch
```diff
@@ -0,0 +1,92 @@
+diff --git a/garglk/CMakeLists.txt b/garglk/CMakeLists.txt
+index 2d9b935e..4dd4167d 100644
+--- a/garglk/CMakeLists.txt
++++ b/garglk/CMakeLists.txt
+@@ -239,9 +239,9 @@ elseif(UNIX)
+     endif()
+ 
+     if(WITH_FREEDESKTOP)
+-        install(FILES "gargoyle.desktop" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/applications")
+-        install(FILES "gargoyle-house.png" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons")
```
hfiguiere, line 10: I would just make it install it in the right place

fdo.patch
```diff
@@ -0,0 +1,92 @@
...
++    <launchable type="desktop-id">io.github.garglk.Gargoyle.desktop</launchable>
++    <releases>
++        <release version="2022.1" date="2022-01-30"/>
++        <release version="2019.1.1" date="2019-11-19"/>
++        <release version="2019.1" date="2019-11-03"/>
++        <release version="2011.1" date="2011-10-19"/>
++    </releases>
++    <provides>
++        <binary>gargoyle</binary>
++    </provides>
```
hfiguiere, line 90: the appstream file is missing `content_rating`.

fdo.patch
```diff
@@ -0,0 +1,92 @@
+diff --git a/garglk/CMakeLists.txt b/garglk/CMakeLists.txt
```
hfiguiere, line 1: has this patch been submitted upstream?

io.github.garglk.Gargoyle.yaml
```diff
@@ -0,0 +1,31 @@
...
+  - --socket=wayland
+  - --socket=pulseaudio
+  - --device=dri
+  - --socket=system-bus
+  - --filesystem=home
+modules:
+  - name: gargoyle
+    buildsystem: cmake
```
hfiguiere, line 18:
```suggestion
    buildsystem: cmake-ninja
```

this is usually faster.

hfiguiere (changes requested):

hfiguiere:
bot, build io.github.garglk.Gargoyle

io.github.garglk.Gargoyle.yaml
cspiegel: As mentioned, this was to get feedback on an error message.  There's nothing that's output with `--log-system-bus`, even though the error message only disappears when there's a system bus (session bus isn't enough, though there _is_ session bus logging).  As in, providing access to the system bus is the only way to remove the error, but there's apparently nothing accessing the system bus, according to the logging?  Not sure what's happening here, and it's possible it's some idiosyncratic behavior on my system.

Again, this doesn't _seem_ to affect behavior, but I'd obviously like to determine why the error message is there.

fdo.patch
hfiguiere: As I said:

```
builddir/files/share/appdata/io.github.garglk.Gargoyle.appdata.xml: FAILED:
Validation of files failed
```

fdo.patch
cspiegel: I'm not entirely sure what you mean.  This is a low-resolution PNG that doesn't satisfy the requirements for Flathub, so it's not installed here, and instead a replacement SVG is being installed.  At some point we'll switch the official icon to an SVG, but the current release just doesn't have one.

fdo.patch
cspiegel: As I mentioned in the pull request, this comes from an upstream branch that will be merged after it's synchronized with any changes required for Flathub.

hfiguiere:
bot, build io.github.garglk.Gargoyle

hfiguiere:
bot, build io.github.garglk.Gargoyle

io.github.garglk.Gargoyle.yaml
Erick555: Could you try `--system-talk-name=org.freedesktop.UDisks2`?

cspiegel: That seems to do the trick, thanks!

io.github.garglk.Gargoyle.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - --socket=pulseaudio
+  - --device=dri
+  - --filesystem=home
+  - --system-talk-name=org.freedesktop.UDisks2
+modules:
+  - name: gargoyle
+    buildsystem: cmake-ninja
+    builddir: true
+    post-install:
+      - install -D $FLATPAK_BUILDER_BUILDDIR/gargoyle-house.svg /app/share/icons/hicolor/scalable/apps/io.github.garglk.Gargoyle.svg
```
hfiguiere, line 20:
```suggestion
      - install -D ../gargoyle-house.svg /app/share/icons/hicolor/scalable/apps/io.github.garglk.Gargoyle.svg
```

Ideally this icon would be upstream

cspiegel: At the moment we have a low-res icon upstream.  I intend to include an SVG icon in the next release, but it's not yet been determined what that icon will be.  The icon here is an SVG representation of the current icon (a white house), so fits with the appearance of the current release, but may not be what ends up as the final icon.  When the next release comes out, this icon will be removed from the Flathub repo, and either added to the official release (if it's the one being used), or the release will have another SVG icon that will be suitable for Flathub.

barthalion:
bot, build io.github.garglk.Gargoyle

PR #3066, opened 2022-04-24
Add org.es_de.emulationstation-de

leonstyhre:

### DIFF
--- a/org.es_de.emulationstation-de.yml
+++ b/org.es_de.emulationstation-de.yml
@@ -60,7 +60,7 @@ modules:
                 cxxflags: -fPIC
     sources:
     - type: archive
-      url: http://downloads.sourceforge.net/freeimage/FreeImage3180.zip
+      url: https://downloads.sourceforge.net/freeimage/FreeImage3180.zip
       sha256: f41379682f9ada94ea7b34fe86bf9ee00935a3147be41b6569c9605a53e438fd
     - type: shell
       commands:
--- a/org.es_de.emulationstation-de.yml
+++ b/org.es_de.emulationstation-de.yml
@@ -23,7 +23,6 @@ finish-args:
   - --filesystem=~/.var/app
   - --share=network
   - --talk-name=org.freedesktop.Flatpak
-  - --env=PATH=/app/bin:/var/lib/flatpak/exports/bin:/usr/bin:/usr/local/bin:/usr/local/sbin:/usr/sbin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/var/lib/snapd/snap/bin
 cleanup:
   - /include
   - /share/ffmpeg
@@ -82,5 +81,5 @@ modules:
     sources:
     - type: git
       url: https://gitlab.com/leonstyhre/emulationstation-de.git
-      commit: 3cb464c7adcea8d020cdff92803c64b7f72fb443
+      commit: ed4e832ea591269ea853f26060740782dcaeb2f0
 
--- a/org.es_de.emulationstation-de.yml
+++ b/org.es_de.emulationstation-de.yml
@@ -1,0 +1,82 @@
+#  SPDX-License-Identifier: MIT
+#
+#  EmulationStation Desktop Edition
+#  org.es_de.emulationstation-de.yml
+#
+#  Flatpak manifest file for use with flatpak-builder.
+#
+
+app-id: org.es_de.emulationstation-de
+sdk: org.freedesktop.Sdk
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+command: emulationstation
+branch: stable
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --socket=wayland
+  - --socket=pulseaudio
+  - --device=all
+  - --filesystem=host:ro
+  - --filesystem=/var/lib/flatpak:ro
+  - --filesystem=home
+  - --filesystem=~/.var/app
+  - --share=network
+  - --talk-name=org.freedesktop.Flatpak
+  - --env=PATH=/app/bin:/usr/bin:/var/lib/flatpak/exports/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/var/lib/snapd/snap/bin
+cleanup:
+  - /include
+  - /share/ffmpeg
+  - /lib/cmake
+  - /lib/debug
+  - /lib/pkgconfig
+  - /lib/libfreeimage.a
+
+modules:
+  - name: ffmpeg
+    config-opts:
+    - --disable-static
+    - --disable-programs
+    - --disable-doc
+    - --enable-gpl
+    - --enable-shared
+    - --enable-libvorbis
+    - --enable-libopus
+    - --enable-libvpx
+    - --enable-postproc
+    sources:
+    - type: git
+      url: https://github.com/FFmpeg/FFmpeg.git
+      tag: n5.0
+
+  - name: freeimage
+    no-autogen: true
+    build-options:
+        cxxflags: -std=c++14
+    sources:
+    - type: archive
+      url: http://downloads.sourceforge.net/freeimage/FreeImage3180.zip
+      sha256: f41379682f9ada94ea7b34fe86bf9ee00935a3147be41b6569c9605a53e438fd
+    - type: shell
+      commands:
+      - sed -i 's|-o root -g root ||' ./Makefile.gnu
+      - sed -i 's|/usr|/app|' ./Makefile.gnu
+
+  - name: pugixml
+    buildsystem: cmake
+    config-opts:
+    - -DBUILD_SHARED_LIBS=on .
+    sources:
+    - type: archive
+      url: https://github.com/zeux/pugixml/releases/download/v1.11.4/pugixml-1.11.4.tar.gz
+      sha256: 8ddf57b65fb860416979a3f0640c2ad45ddddbbafa82508ef0a0af3ce7061716
+
+  - name: emulationstation
+    buildsystem: cmake
+    config-opts:
+    - -DFLATPAK_BUILD=on .
+    sources:
+    - type: git
+      url: https://gitlab.com/leonstyhre/emulationstation-de.git
+      commit: 3cb464c7adcea8d020cdff92803c64b7f72fb443
--- a/org.es_de.emulationstation-de.yml
+++ b/org.es_de.emulationstation-de.yml
@@ -53,6 +53,11 @@ modules:
     no-autogen: true
     build-options:
         cxxflags: -std=c++14
+        arch:
+            aarch64:
+                # See https://sourceforge.net/p/freeimage/bugs/325/
+                cflags: -fPIC -DPNG_ARM_NEON_OPT=0
+                cxxflags: -fPIC
     sources:
     - type: archive
       url: http://downloads.sourceforge.net/freeimage/FreeImage3180.zip
--- a/org.es_de.emulationstation-de.yml
+++ b/org.es_de.emulationstation-de.yml
@@ -81,5 +81,5 @@ modules:
     sources:
     - type: git
       url: https://gitlab.com/leonstyhre/emulationstation-de.git
-      commit: 3f1bdf58a9a1a3c7c71195fbf973168463da69f0
+      commit: cb16b510022e95cc1a8e51a6ab26caaa0b73bb58
 
--- a/org.es_de.emulationstation-de.yml
+++ b/org.es_de.emulationstation-de.yml
@@ -30,7 +30,7 @@ cleanup:
   - /lib/cmake
   - /lib/debug
   - /lib/pkgconfig
-  - /lib/*.a
+  - '*.a'
 
 modules:
   - name: ffmpeg
--- a/org.es_de.emulationstation-de.yml
+++ b/org.es_de.emulationstation-de.yml
@@ -28,7 +28,6 @@ cleanup:
   - /include
   - /share/ffmpeg
   - /lib/cmake
-  - /lib/debug
   - /lib/pkgconfig
   - '*.a'
 
--- a/org.es_de.emulationstation-de.yml
+++ b/org.es_de.emulationstation-de.yml
@@ -81,5 +81,5 @@ modules:
     sources:
     - type: git
       url: https://gitlab.com/leonstyhre/emulationstation-de.git
-      commit: ed4e832ea591269ea853f26060740782dcaeb2f0
+      commit: 446cc235cf7bc09b0eb7bd548e14f92e041bbda3
 
--- a/org.es_de.emulationstation-de.yml
+++ b/org.es_de.emulationstation-de.yml
@@ -81,5 +81,5 @@ modules:
     sources:
     - type: git
       url: https://gitlab.com/leonstyhre/emulationstation-de.git
-      commit: 446cc235cf7bc09b0eb7bd548e14f92e041bbda3
+      commit: 3f1bdf58a9a1a3c7c71195fbf973168463da69f0
 
--- a/org.es_de.emulationstation-de.yml
+++ b/org.es_de.emulationstation-de.yml
@@ -18,12 +18,12 @@ finish-args:
   - --socket=pulseaudio
   - --device=all
   - --filesystem=host:ro
-  - --filesystem=/var/lib/flatpak:ro
+  - --filesystem=/var/lib/flatpak/exports/bin:ro
   - --filesystem=home
   - --filesystem=~/.var/app
   - --share=network
   - --talk-name=org.freedesktop.Flatpak
-  - --env=PATH=/app/bin:/usr/bin:/var/lib/flatpak/exports/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/var/lib/snapd/snap/bin
+  - --env=PATH=/app/bin:/var/lib/flatpak/exports/bin:/usr/bin:/usr/local/bin:/usr/local/sbin:/usr/sbin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/var/lib/snapd/snap/bin
 cleanup:
   - /include
   - /share/ffmpeg
--- a/org.es_de.emulationstation-de.yml
+++ b/org.es_de.emulationstation-de.yml
@@ -11,7 +11,6 @@ sdk: org.freedesktop.Sdk
 runtime: org.freedesktop.Platform
 runtime-version: '21.08'
 command: emulationstation
-branch: stable
 finish-args:
   - --share=ipc
   - --socket=x11
@@ -31,7 +30,7 @@ cleanup:
   - /lib/cmake
   - /lib/debug
   - /lib/pkgconfig
-  - /lib/libfreeimage.a
+  - /lib/*.a
 
 modules:
   - name: ffmpeg
@@ -64,19 +63,20 @@ modules:
       - sed -i 's|/usr|/app|' ./Makefile.gnu
 
   - name: pugixml
-    buildsystem: cmake
+    buildsystem: cmake-ninja
     config-opts:
-    - -DBUILD_SHARED_LIBS=on .
+    - -DBUILD_SHARED_LIBS=on
     sources:
     - type: archive
       url: https://github.com/zeux/pugixml/releases/download/v1.11.4/pugixml-1.11.4.tar.gz
       sha256: 8ddf57b65fb860416979a3f0640c2ad45ddddbbafa82508ef0a0af3ce7061716
 
   - name: emulationstation
-    buildsystem: cmake
+    buildsystem: cmake-ninja
     config-opts:
-    - -DFLATPAK_BUILD=on .
+    - -DFLATPAK_BUILD=on
     sources:
     - type: git
       url: https://gitlab.com/leonstyhre/emulationstation-de.git
       commit: 3cb464c7adcea8d020cdff92803c64b7f72fb443
+


org.es_de.emulationstation-de.yml
```diff
@@ -0,0 +1,82 @@
...
+runtime-version: '21.08'
+command: emulationstation
+branch: stable
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --socket=wayland
+  - --socket=pulseaudio
+  - --device=all
+  - --filesystem=host:ro
```
hfiguiere, line 21: why host?

org.es_de.emulationstation-de.yml
```diff
@@ -0,0 +1,82 @@
...
+  - --filesystem=/var/lib/flatpak:ro
```
hfiguiere, line 22: why do you need this one?

org.es_de.emulationstation-de.yml
```diff
@@ -0,0 +1,82 @@
...
+  - --filesystem=home
+  - --filesystem=~/.var/app
```
hfiguiere, line 24: or even why do you need this one?

org.es_de.emulationstation-de.yml
```diff
@@ -0,0 +1,82 @@
...
+  - --share=network
+  - --talk-name=org.freedesktop.Flatpak
+  - --env=PATH=/app/bin:/usr/bin:/var/lib/flatpak/exports/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/var/lib/snapd/snap/bin
```
hfiguiere, line 27: Why this when many of these path don't exist?

org.es_de.emulationstation-de.yml
```diff
@@ -0,0 +1,82 @@
...
+  - --socket=x11
+  - --socket=wayland
+  - --socket=pulseaudio
+  - --device=all
+  - --filesystem=host:ro
+  - --filesystem=/var/lib/flatpak:ro
+  - --filesystem=home
+  - --filesystem=~/.var/app
+  - --share=network
+  - --talk-name=org.freedesktop.Flatpak
```
hfiguiere, line 26: why access to this d-bus?

org.es_de.emulationstation-de.yml
```diff
@@ -0,0 +1,82 @@
...
+  - --env=PATH=/app/bin:/usr/bin:/var/lib/flatpak/exports/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/var/lib/snapd/snap/bin
+cleanup:
+  - /include
+  - /share/ffmpeg
+  - /lib/cmake
+  - /lib/debug
```
hfiguiere, line 32:
```suggestion
```

org.es_de.emulationstation-de.yml
```diff
@@ -0,0 +1,82 @@
...
+  - /lib/pkgconfig
+  - /lib/libfreeimage.a
```
hfiguiere, line 34:
```suggestion
  - *.a
```

org.es_de.emulationstation-de.yml
```diff
@@ -0,0 +1,82 @@
...
+    - type: archive
+      url: http://downloads.sourceforge.net/freeimage/FreeImage3180.zip
+      sha256: f41379682f9ada94ea7b34fe86bf9ee00935a3147be41b6569c9605a53e438fd
+    - type: shell
+      commands:
+      - sed -i 's|-o root -g root ||' ./Makefile.gnu
+      - sed -i 's|/usr|/app|' ./Makefile.gnu
+
+  - name: pugixml
+    buildsystem: cmake
```
hfiguiere, line 67:
```suggestion
    buildsystem: cmake-ninja
```

org.es_de.emulationstation-de.yml
```diff
@@ -0,0 +1,82 @@
...
+    config-opts:
+    - -DBUILD_SHARED_LIBS=on .
+    sources:
+    - type: archive
+      url: https://github.com/zeux/pugixml/releases/download/v1.11.4/pugixml-1.11.4.tar.gz
+      sha256: 8ddf57b65fb860416979a3f0640c2ad45ddddbbafa82508ef0a0af3ce7061716
+
+  - name: emulationstation
+    buildsystem: cmake
```
hfiguiere, line 76:
```suggestion
    buildsystem: cmake-ninja
```
unles it doesn't work

org.es_de.emulationstation-de.yml
```diff
@@ -0,0 +1,82 @@
...
+    config-opts:
+    - -DFLATPAK_BUILD=on .
```
hfiguiere, line 78: I'm not sure the dot is correct.

org.es_de.emulationstation-de.yml
```diff
@@ -0,0 +1,82 @@
...
+#
+#  Flatpak manifest file for use with flatpak-builder.
+#
+
+app-id: org.es_de.emulationstation-de
+sdk: org.freedesktop.Sdk
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+command: emulationstation
+branch: stable
```
hfiguiere, line 14:
```suggestion
```

hfiguiere:
bot, build org.es_de.emulationstation-de

hfiguiere (changes requested):

org.es_de.emulationstation-de.yml
leonstyhre: It's an emulator frontend which needs to be able to launch emulators wherever they are installed on the filesystem.

org.es_de.emulationstation-de.yml
leonstyhre: Yes it's required, if I remove any of these, the application doesn't work as it can't launch the required emulators. And I need access to devices to handle controller input for instance.

org.es_de.emulationstation-de.yml
leonstyhre: This is required to locate RetroArch cores as they are installed in ~/.var/app

org.es_de.emulationstation-de.yml
leonstyhre: The application has to be able to locate emulators on various distributions, and this path is a combination of all the paths required based on my thorough testing across these distributions.

org.es_de.emulationstation-de.yml
leonstyhre: Without it I can't run emulators using ""flatpak-spawn --host" so the application becomes totally unusable without this level of access.

org.es_de.emulationstation-de.yml
leonstyhre: It's the only .a file generated by the build process, but sure I can change it to *.a instead.

org.es_de.emulationstation-de.yml
hfiguiere: or you sure that inside the flatpak sandbox there is any of these? Are you really sure?

org.es_de.emulationstation-de.yml
hfiguiere: I never talked about device here.

hfiguiere: I still don't understand the justification to give acces to this.

org.es_de.emulationstation-de.yml
hfiguiere: so basically you want to grant access to all apps for one.

org.es_de.emulationstation-de.yml
leonstyhre: No problem, I'll change this.

org.es_de.emulationstation-de.yml
leonstyhre: No problem, I'll change this.

org.es_de.emulationstation-de.yml
leonstyhre: Ok, I'll remove the dot.

org.es_de.emulationstation-de.yml
leonstyhre: I'll remove this.

org.es_de.emulationstation-de.yml
hfiguiere: `--filesystem=host` does not give access the `/usr`, `/bin`, `/sbin` among others.

And if you goal is to run host binaries inside the sandbox, then you should revisit this idea.

org.es_de.emulationstation-de.yml
leonstyhre: Do you mean this line specifically?
- --filesystem=/var/lib/flatpak:ro

org.es_de.emulationstation-de.yml
leonstyhre: Yes, without this access my application will not be able to launch all the emulators that are installed as Flatpaks, which nowadays is the preferred way to install emulators. Especially so on the Steam Deck where many users run my application today (it's currently distributed as an AppImage).

org.es_de.emulationstation-de.yml
leonstyhre: Everything I've attempted to run works with these settings enabled, for example emulators installed as Snaps, Flatpaks and source ports installed to /usr/games/ etc.

org.es_de.emulationstation-de.yml
hfiguiere: that is where the comment is.

org.es_de.emulationstation-de.yml
hfiguiere: 1. it's included in the broad `--filesystem=home`
2. I think you misunderstand how flatpak works. It's a sandbox. If you want to launch another flatpak it's done outside of that sandbox.

org.es_de.emulationstation-de.yml
hfiguiere: I doubt. The `/usr` is provided by the faltpak platform and is inside the sandbox. It is unrelated to the distribution.

org.es_de.emulationstation-de.yml
leonstyhre: Actually, this is the same reason as my comment below:
Yes, without this access my application will not be able to launch all the emulators that are installed as Flatpaks, which nowadays is the preferred way to install emulators. Especially so on the Steam Deck where many users run my application today (it's currently distributed as an AppImage).

org.es_de.emulationstation-de.yml
leonstyhre: Yes that is a bit strange as ~/.var/app is a subdirectory in home, but if I remove this entry then it doesn't work. I can relocate the RetroArch cores to any location outside of ~/.var/app, for instance ~/.var/cores but as soon as they are placed inside and subdirectory in ~/.var/app they can no longer be found. Unless I add this entry, then it works.

No I think I understand how it works, I launch emulators using "flatpak-spawn --host" so they run outside the sandbox, but I need to be able to check which cores are available before triggering the launch of the emulator as I warn the user with a popup notification about missing cores. So therefore I need to be able to see the files inside this directory.

org.es_de.emulationstation-de.yml
leonstyhre: Aha, so you mean that /usr is shadowed by Flatpak itself and can't be accessed by any means from inside the sandbox?

leonstyhre: Sorry, I mean /usr/bin as for sure /usr/games is accessible as I can launch source ports installed there.

hfiguiere: There is no `/usr/games` inside the sandbox. and `/usr/bin` is the one from the sandbox.

leonstyhre: No I don't mean /usr/games is inside the sandbox, but I use the PATH variable to get a list of locations to look for emulators, including /usr/games. I don't necessarily need to have this PATH variable set in the manifest file, but it makes life easier for me since I would otherwise need to hardcode the paths inside the application.

leonstyhre: Maybe this will make it clearer:
https://gitlab.com/leonstyhre/emulationstation-de/-/blob/master/es-app/src/FileData.cpp

hfiguiere: this is not what `PATH` is for.

leonstyhre: The findEmulatorPath() function is where I locate emulators, and it works perfectly fine to find for example those installed in /usr/games or those installed as Flatpaks, Snaps etc. But it requires the PATH information and the security settings in my manifest file to work.

leonstyhre: The rule set is defined here:
https://gitlab.com/leonstyhre/emulationstation-de/-/blob/master/resources/systems/unix/es_find_rules.xml

leonstyhre: Setting the PATH using env= provides me with the paths to look for emulators, without this setting there are only two entries defined (I think /usr/bin and /app/bin or something like that?). So I read the PATH variable inside the application to expand the places to look for emulators.

org.es_de.emulationstation-de.yml
```diff
@@ -0,0 +1,82 @@
...
+    - --enable-postproc
+    sources:
+    - type: git
+      url: https://github.com/FFmpeg/FFmpeg.git
+      tag: n5.0
+
+  - name: freeimage
+    no-autogen: true
+    build-options:
+        cxxflags: -std=c++14
```
hfiguiere, line 55:
```suggestion
        cxxflags: -std=c++14
        arch:
            aarch64:
                # See https://sourceforge.net/p/freeimage/bugs/325/
                cflags: -fPIC -DPNG_ARM_NEON_OPT=0
                cxxflags: -fPIC
```

org.es_de.emulationstation-de.yml
```diff
@@ -0,0 +1,82 @@
...
+    sources:
+    - type: archive
+      url: http://downloads.sourceforge.net/freeimage/FreeImage3180.zip
```
hfiguiere, line 58:
```suggestion
      url: https://downloads.sourceforge.net/freeimage/FreeImage3180.zip
```

barthalion:
bot, build org.es_de.emulationstation-de

org.es_de.emulationstation-de.yml
barthalion: The manifest also defines `--talk-name=org.freedesktop.Flatpak`. The app doesn't need access to `/var/lib/flatpak` or `~/.var/app` if it launches commands with `flatpak-spawn --host`.

org.es_de.emulationstation-de.yml
barthalion: This does seem excessive though, and I'm not sure if I understand the point. If this is about cores downloaded by ES itself, it should really use $XDG_DATA_HOME to store them.

org.es_de.emulationstation-de.yml
barthalion: OK, I see below it's about the `exports/bin` subdirectory. Why not limit it to that directly, then?

org.es_de.emulationstation-de.yml
barthalion: I would consider moving it elsewhere, e.g. to a wrapper script, or the application itself.

leonstyhre: If it's an issue for you I can hardcode it inside the application although it would be a bit ugly. I added these entries to the PATH variable because that's how I look for emulators (well one of multiple ways I look for emulators) so it would require no code changes on my side. If you could confirm that using the environment variables is not acceptable then I'll change, but I'm not sure I understand the concern as it's not a security risk to add entries to the PATH variable?

org.es_de.emulationstation-de.yml
leonstyhre: No I don't download any cores, my application only acts as a frontend. So what I'll do is that on game launch check if the core that is configured for the certain game/system exists, and if it doesn't I display an error to the user. If it does exist, then I trigger a launch of the emulator using flatpak-spawn --host and pass the core location as an argument. If you have a suggestion to solve this without using - --filesystem=~/.var/app then I'm very happy to accommodate that, but it has to be a solution that still makes the application work correctly.

org.es_de.emulationstation-de.yml
leonstyhre: Yes I always check that the binary exists before attempting a launch of an emulator and I provide an error message to the users if the emulator does not exist. Just so that I understand, do you suggest changing the configuration to this?

- --filesystem=/var/lib/flatpak/exports/bin:ro

org.es_de.emulationstation-de.yml
leonstyhre: Btw, the reason I don't set explicit links to the RetroArch config directory is that my application is very configurable and the user may create their own custom systems configuration entries. So they could use another application/emulator potentially that works with emulator cores or equivalent functionality, and just pointing to a specific RetroArch directory would cripple the ability to customize systems. As well, my frontend is not limited to games and could launch any application installed by the users, even desktop applications.

org.es_de.emulationstation-de.yml
leonstyhre: And btw, thanks for fixing the aarch64 build!

leonstyhre: Thinking more about the PATH variable, the best solution would be if I could access this variable from the "real" system outside the sandbox, then none of this custom or hardcoded stuff would be needed. Is there a reliable way to do this, i.e. reading the PATH of the user that is executing the Flatpak?

org.es_de.emulationstation-de.yml
leonstyhre: I've now committed an update where I've restricted the /var/lib/flatpak:ro further to /var/lib/flatpak/exports/bin:ro

leonstyhre:
I've tested the automatic build you made and it works fine, all the emulators and source ports I've tested launch correctly and the video player and scraper work fine as well. I just added a further filesystem restriction as per your request but the application should still work correctly even with this implemented.

As well I tested an aarch64 build on a Raspberry Pi 400 with this manifest file and it also built and ran correctly.

So if we could just resolve the final outstanding issue with the PATH variable I guess we have an application that could be published? Thanks!

org.es_de.emulationstation-de.yml
```diff
@@ -0,0 +1,87 @@
...
+  - --share=network
+  - --talk-name=org.freedesktop.Flatpak
+  - --env=PATH=/app/bin:/var/lib/flatpak/exports/bin:/usr/bin:/usr/local/bin:/usr/local/sbin:/usr/sbin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/var/lib/snapd/snap/bin
+cleanup:
+  - /include
+  - /share/ffmpeg
+  - /lib/cmake
+  - /lib/debug
+  - /lib/pkgconfig
+  - /lib/*.a
```
hfiguiere, line 33:
```suggestion
  - '*.a'
```

leonstyhre: I did change it, it was the following before:
/lib/libfreeimage.a
And I changed it to:
/lib/*.a
There are no .a files outside the lib directory.

Running this command confirms it:
find ~/.local/share/flatpak/app/org.es_de.emulationstation-de/ -name *.a | wc -l
0

hfiguiere: *sigh*

This will remove ANY `*.a`. Period. It doesn't matter if they are just in `lib`.

org.es_de.emulationstation-de.yml
leonstyhre: I could remove it, but doing that will just leave me with an empty lib/debug directory after building the package. If that's something you need retained inside the Flatpak for some reason then I have no problem doing it, so please clarify whether it's required.

org.es_de.emulationstation-de.yml
leonstyhre: Yes I'm aware of how regular expressions work. There are however no .a files outside the /lib/ directory. But I've still made the change and committed it now to satisfy your request.

org.es_de.emulationstation-de.yml
hfiguiere: debug symbols are separated into their own flatpak.

leonstyhre: Thanks for clarifying, I have now removed the entry and I've just committed the change.

leonstyhre:
Could you please let me know what additional changes are needed (if any) to get my application published? As far as I know I have restricted filesystem access to the minimum required for the application to still function correctly.
I have tested a workaround to hardcode the PATH variable inside the application but I would prefer not to use it as it would be a special solution only for Flatpaks and it's not a pretty solution. But if using "--env=PATH=" is unacceptable I can still do it so please advice.
If any other changes than these are required, then please also advice. The application has been tested and the Flatpak works fine on both aarch64 and x86 and I think it's ready for publishing. Thanks!

Witherking25:
https://github.com/flatpak/flatpak/issues/4043

leonstyhre:
Thanks for the XCB info, but after reading this information I'm not sure what I could do to resolve it? Seems like it either needs to be fixed in the OS or a Flatpak code change would be needed? Or is there anything I could do inside the package to mitigate this problem?

Witherking25:
can you run it with QT_DEBUG_PLUGINS=1 and see the output?

leonstyhre:
I tried to run it with QT_DEBUG_PLUGINS=1 and it didn't print any additional log output, but I have no Qt libraries in my Flatpak so I guess that makes sense? Any other way to provide more debug output?

Witherking25:
you might be able to get around this by setting some env vars

Witherking25:
I'm in the after q3 club myself

leonstyhre:
He also told me that if he starts the ES-DE Flatpak like this, then emulators can be started:
```
DISPLAY=":0" XAUTHORITY="/run/user/1000/d15ba0cc-cd1d-471b-886d-1e44b2392894" flatpak run org.es_de.emulationstation-de
```

Witherking25:
then add those or similar as finish params

Witherking25:
I've been using es de on my manjaro desktop that I use for testing steam deck stuff.

leonstyhre:
Sorry, I'm not sure I follow, could you give an example of what I could add to the manifest file?

leonstyhre:
That's nice to hear! Yes I have added ES-DE to the AUR or you could use the AppImage of course. But many users ask for Flatpak support to make it convenient to install and keeping the application updated over time.

Witherking25:
"--env=$(stuff to find the /run/user/1000/* directory)"

Witherking25:
add it to finish args

Witherking25:
flatpak mounts the host at /var/run/host

Witherking25:
the $PATH would still point to the flatpak directory unless you prepend /var/run/host

Witherking25:
but $(echo $PATH) prints the flatpak $PATH not the host $PATH

Witherking25:
there is a way though let me check...

Witherking25:
it will post a command for installing said build once it's finished building

leonstyhre:
Aha, I thought only admins could trigger the build bot, thanks for clarifying that!

Witherking25:
so you can start a build by typing the thing i put above

Witherking25:
bot is the keyword it searches for btw

Witherking25:
with a comma too

leonstyhre:
Thanks, that's great. I'll working on this tonight and hopefully I can find some way to make this work. If you figure out a way to inject the host PATH into the Flatpak please let me know as that would be very helpful.

Witherking25:
$(flatpak-spawn --host echo $PATH) works

Witherking25:
also the host is only mounted if you give the flatpak the filesystem permission "host"

Witherking25:
so 
"--env=PATH=$PATH:$(flatpak-spawn --host echo $PATH)"

leonstyhre:
Thanks a lot, I'll give it a go.

Witherking25:
flatpak-spawn --host echo $PATH just prints the sandbox $PATH i guess

Witherking25:
ok i made a boo boo
$() is still static
`command` is 100% dynamic

leonstyhre:
Thanks, yes I tried to set the PATH but then I couldn't even run the Flatpak. So I'll have to continue with the hardcoded PATH. But that works although it's quite ugly, the more important issue at the moment is the XCB stuff. I'll see if I can get that to work somehow.

Witherking25:
Path must be hard coded because sandbox has no access to host path

Witherking25:
The XCB env vars can be assigned by adding the following to the finish args:
```
“--env=XAUTORITY=`script`”,
“--env=DISPLAY=\”:0\””
```

leonstyhre:
It seems as if overriding these variables is impossible. No matter how I enter the env= entries they show up with the same values inside the Flatpak. My guess is they are initialized on application startup somehow and that the variables passed in finish-args are getting overwritten.

Witherking25:
i was wrong about \`\` then. can you try a wrapper script that exports the env vars then calls emulationstation?

Witherking25:
```bash
#!/bin/bash
export XAUTORITY=$(/app/bin/script or just generate it in the wrapper script and set it here)
export DISPLAY=":0"
emulationstation
```

Witherking25:
> It seems as if overriding these variables is impossible. No matter how I enter the env= entries they show up with the same values inside the Flatpak. My guess is they are initialized on application startup somehow and that the variables passed in finish-args are getting overwritten.

my guess is that it is being overriden by the host variable

leonstyhre:
I'm not willing to continue spending time on this at the moment. The Flatpak works fine on other distributions like Ubuntu, Fedora and Manjaro, it's a waste of energy and a lost cause as long as the problem persists in SteamOS and apparently also OpenSUSE.

Let's just hope it gets resolved in these operating systems and/or in the Flatpak framework, then I may continue my efforts of making a Flatpak release of ES-DE. In the meanwhile the AppImage works fine, or the DEB, RPM and AUR releases I make available.

I have one additional Flatpak-specific change I'll commit and then I'll test build that to see that it works fine on the fully functional operating systems. After that I'll put this on hold.

leonstyhre:
No wrapper script should be required, the package works perfectly on Ubuntu, Fedora and Manjaro. I need to be able to pass the variables into the script and it seems to be broken. I'm not willing to spend more time on this but if you are able to get it to run reliably then please let me know. I think you should be able to test it on OpenSUSE as that has reportedly the same problem as SteamOS.

Witherking25:
i guess im fireing up an OpenSUSE vm lol

leonstyhre:
bot, build org.es_de.emulationstation-de

org.es_de.emulationstation-de.yml
barthalion: Bear with me because I clearly don't know enough about ES. By these comments, do you mean that ES will try to re-use RetroArch cores that were downloaded with flatpaked RetroArch and saved within ~/.var/app?

barthalion: OK, I think I understand that correctly based on https://gitlab.com/leonstyhre/emulationstation-de/-/blob/master/resources/systems/unix/es_find_rules.xml#L23. This could really be limited to `~/.var/app/org.libretro.RetroArch` then.

barthalion: That being said, I placed a GBA rom in `~/ROMs/gba`, and it shows up on the list, but fails to launch, saying it couldn't find an emulator.

```
Apr 28 10:23:59 Error:	Couldn't launch game, emulator binary not found
Apr 28 10:23:59 Error:	Raw emulator launch command:
Apr 28 10:23:59 Error:	%EMULATOR_RETROARCH% -L %CORE_RETROARCH%/mgba_libretro.so %ROM%
```

The XML file defines an entry for mGBA though, so I think I understand even less what these definitions are about.

barthalion: …and if it defaults to RetroArch, why not bundle it with the app?

leonstyhre: Installing RetroArch as Flatpak is only one possibility, the user may have installed it as a Snap, AppImage or via the OS repository, or possibly built it from source code. As well the cores could be relocated to wherever, possibly to an SD card, external drive, mounted NFS share or whatever. I've seen users doing all kinds of stuff by now and ES-DE is fully configurable and if customizing your es_find_rules.xml file almost anything is possible.

Many users apparently don't like RetroArch at all and only want to use ES-DE with standalone emulators, and as well I don't plan to release updates that frequently, so it would be a bad idea to ship RetroArch in the same Flatpak. I would basically have to do the same for all supported emulators. There is actually a separate project to do precisely that which is called RetroDECK, their plan is to bundle all supported emulators in the same Flatpak with ES-DE. I collaborate with this project and I think they also have an open PR here.

leonstyhre: When you launch a game ES-DE checks if the emulator is installed, and if not it shows an error notification inside the application. It tries to locate emulators via all the rules defined in es_find_rules.xml in sequential order. Only after the emulator binary is found it attempts to locate the cores configured using the %CORE_% variable. So in your case the emulator binary is not found for some reason, and it would be quite interesting to understand why. Starting ES-DE with the --debug flag should provide some more detailed log output. I guess you tried it with the lastest Flatpak I had built yesterday?

I had to make quite an ugly hack to get around the Flatpak sandbox limitations or I would not be able to find emulators installed via the repository which could go into directories shadowed by the sandbox, like /usr/bin or /usr/local/bin for instance. As such I call the "which" command using "flatpak-spawn --host" and it actually works although being very far from elegant. If you can let me know a better way to do it I'm very happy to implement that instead, but it really has to be able to locate binaries anywhere in the host. I have actually not implemented the same hack for absolute paths so at the moment the Flatpak build is a bit crippled as you couldn't for instance make an entry like /usr/games/chocolate-doom and have ES-DE launch that when running as a Flatpak. Either I'll add support for that to my ugly hack, or there may be a more elegant solution to get around the shadowed directories. If you have any ideas how to, then please advice!

leonstyhre: In general, sure ES-DE could be locked down further in the manifest, but the more restrictions are added, the less flexible the software becomes and at a certain point it's not worth shipping as a Flatpak. This is so as the AppImage, DEB, RPM or AUR releases have no such limitations. I would really like to ship as a Flatpak, but I can't cripple the application too much as I would get a large amount of support requests from users. Nobody should for example have to use Flatseal to get ES-DE running properly even if they have their ROMs on an SD card or their RetroArch cores in a non-standard location.

Witherking25:
for me is refuses to believe that i have any emulators installed...

Witherking25:
i literally have emudeck setup on my laptop so it definitely has emulators. they are all flatpaks, but that shouldn't matter. emudeck uses appimage es_de and it works there...

leonstyhre:
The Flatpak hack I implemented may not be portable across all distributions, I can only tell you that it works for me on Ubuntu, Fedora and Manjaro.

I'm starting to feel that Flatpak support for ES-DE is a lost cause, at least for the time being. If Flatpaks gets improved so that full access is provided to the host OS while not shadowing directories like /usr then it may be feasible. The PATH variable from the host OS also has to be available from inside the sandbox. Rewriting a lot of core application logic to work around restrictions of a single package format is not worth it at this point in time and some restrictions may even be impossible to overcome. And on top of that it doesn't work on SteamOS and OpenSUSE anyway as something is broken on those distributions when it comes to running "flatpak-spawn --host".

I'll think a bit more about it, but likely I'll soon close this PR. ES-DE works perfectly fine as DEB, RPM, AppImage or built from the AUR so there are definitely good alternatives available (and it also runs on FreeBSD, OpenBSD, NetBSD, macOS and Windows but that's another topic). I'm also looking into adding in-application updates, then there will be limited benefits from using a package manager in the first place. But that will be some time into the future.

Still I really appreciate the time you spent looking into getting the Flatpak to run!

Witherking25:
I was on Ubuntu on the laptop test but ok

PR #3070, opened 2022-04-25
Add org.qutebrowser.qutebrowser.Userscripts

tinywrkb:

  The only added patch is to change the installation target of opendht's Python module, so it's a Flatpak specific packaging issue.

qutebrowser userscripts are external scripts that are run by the browser. An userscript can be bound to a key combination, triggered manually with the `:spawn` command, or through URL hints.  
There are a few userscripts that are installed by the app (submitted to the upstream repo), and the user can add more into `xdg-data/qutebrowser/userscripts`.  
These userscripts require a misc of Python modules, and applications.  
Userscripts and running external applications are two features that are important to qutebrowser's users, and one reason the Flatpak app is not very popular among them is that it doesn't support them.  

To resolve this, a flatpak-spawn wrapper and **dependencies for userscripts** were added recently to the app.  
The problem is that build time now exploded to over 40 minutes on the CI, which makes maintenance more awkward, and time-consuming.  
I've already done build optimization for the qutebrowser Flatpak packaging, including submitting a PyQt base application, but build time is still too long.  
Moving userscripts' dependencies into a separate Flatpak extension bring us to around 10 minutes of Flathub CI run for the qutebrowser app.

~~**The branch set for the extension is matching the KDE runtime version**, to help with runtime bumps, and avoid breaking the extension when the app has a runtime version change.  
If the CI only detects git branches that follow Freedesktop version names, meaning `branch/21.08`, and not KDE version names, meaning `branch/5.15-21.08`, then I guess I could use the former, as currently this extension does not depends on KDE-runtime-specific libs.~~ **edit: This was criticized, so the branch was changed back to `stable`. I guess a temporary breakage of the extension for the user is acceptable.**

The last missing pieces to make this work with the app is updating `PATH` and `PYTHONPATH`, and these are handled in flathub/org.qutebrowser.qutebrowser#184, which will be merged post publication of this extension. **edit:** Note that my current half-baked update helper tools (in the `tools/` folder) are expected to be re-written.

The extension submitted with flathubbot checks enabled, but will switch to a custom GitHub workflow, as some modules require updating with flatpak-builder-tools, so currently they are not updated automatically. This will be left for post merge.

Another expected change after merging this is packaging also userscripts. This way we can pick up fixes for userscripts from qutebrowser stable branch, and also add popular userscripts that are not in qutebrowser's repository.

**CC:** @The-Compiler


tinywrkb:
bot, build org.qutebrowser.qutebrowser.Userscripts

tinywrkb:
bot, build org.qutebrowser.qutebrowser.Userscripts

tinywrkb:
This looks fine, but some Python modules are not installed due to existing modules in `/app/lib/python<pyver>`.  
Setting the pip option `--ignored-installed` does more harm than good, so I prefer merging in the current state.  
This means an extension rebuild will be needed post flathub/org.qutebrowser.qutebrowser#184 merge, and there will be a time frame that the userscripts will be missing supporting Python modules.  
I expected this kind of uncertainties with the userscripts feature and the flatpak-spawn wrapper, which is why I haven't documented it, and announced it, so I mind breaking it temporarily.  
I want to avoid the alternative, which is submitting an empty base application that only has an extension mount point.

tinywrkb:
The latest push is just an update for cleanup.

tinywrkb:
bot, build org.qutebrowser.qutebrowser.Userscripts

tinywrkb:
bot, build org.qutebrowser.qutebrowser.Userscripts

barthalion:
As this is an extension for specific app, it should use `stable` as branch instead of following runtime versioning. I also don't see the point of `module/module.json` kind of split, and I would say it actually makes readability of the manifest worse. If you insist on splitting modules to separate files, at least keep them in a single directory.

tinywrkb:
> As this is an extension for specific app, it should use stable as branch instead of following runtime versioning.

Fixed.

tinywrkb:
bot, build org.qutebrowser.qutebrowser.Userscripts

barthalion:
I'm going to make an exception about the form of the manifest because you're a known contributor. In general, it's not going to pass.

/merge @The-Compiler

tinywrkb:
Thanks for merging!

Issue #2533, opened 2021-09-26
Handling ABI dependent extensions


tinywrkb:
The [Fcitx5](https://github.com/flathub/org.fcitx.Fcitx5) and its extensions are outdated quite a while now, so I thought about adding `f-e-d-c` support, but then noticed the dependency of extesnsion on libs from the app's package.  

How to deal with ABI breakage of libs from the app's package?  
The updated app needed to be in the repo in order to build updated extensions, but shouldn't be available to the user.  
Is there a way to have an interim branch and hold off pushing the updated app to the stable branch?  
Is it possible to avoid rebuilding, and just copy the refs from the interim branch to the stable branch after the app and its extensions were updated?

CC @wengxt the developer Fcitx5 and its Flathub app maintainer.

barthalion:
I don't think it's really doable without coordinated (e.g. with me) mass-merge of changes. Re-commiting a non-stable ref to the stable one is possible but also requires my direct involvement on the infrastructure.

tinywrkb:
@barthalion thanks for responding.  
Specifically for Fcitx5, the maintainer of the Flathub app should join the discussion and give his opinion.

For the general case, I hope maybe we can have some guidelines written somewhere, at least mention that this is a possible problem.  

> I don't think it's really doable without coordinated (e.g. with me) mass-merge of changes. Re-commiting a non-stable ref to the stable one is possible but also requires my direct involvement on the infrastructure.

How many of you there are? Adding workload to Flathub's developers seems to me like something that should be avoided.

As I see it, these are the possible solutions that should be added to the guidelines:

* Don't use extensions, package everything with the app.
* The app should manage its plugins/addons instead of using Flatpak extensions. Apps like Steam, Lutris, RetroArch, and Kodi, download and run executable code, so this is nothing new, but will fail if `~/` is mounted with `noexec`.
* Give the developer access to push directly into Flathub. Isn't it time to decide on an official policy? I haven't seen one.

tinywrkb:
I'm closing this as it doesn't look like there's an interest in this, and I suggested to the maintainer of the affected app to use bundled extensions method ([see PR](https://github.com/flathub/org.fcitx.Fcitx5/pull/4), or get authorization to push directly to Flathub.

PR #3077, opened 2022-04-28
Add org.freedesktop.Sdk.Extension.wxwidgets

tinywrkb:

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link: https://github.com/wxWidgets/wxWidgets/issues/22370**

This SDK extension packages wxWidgets, WebKitGTK, and their dependencies.  
I'm not developing wxWidgets apps, but a while back I took over the maintenance of Codeblocks IDE (I was nostalgic), which its maintainer and original submitter was MIA.  
As a wxWidgets app, an important feature of Codeblocks is its support for wxWidgets development, and the previous Flatpak maintainer did not take this into consideration.  

When I started maintaining the app, I quickly switched the runtime to Gnome due to the availability WebKitGTK, and bundled wxWidgets' development files (headers, pkgconfig, ...) to have some sort of wxWidgets development environment, but I don't think this is the way forward.  
Instead, I'm proposing adding a wxWidgets Freedesktop SDK extension which could work with any IDE.

Relevant Codeblocks app changes flathub/org.codeblocks.codeblocks#21

~~This is still WIP, so it's marked as draft.~~

### DIFF
--- a/org.freedesktop.Sdk.Extension.wxwidgets.metainfo.xml
+++ b/org.freedesktop.Sdk.Extension.wxwidgets.metainfo.xml
@@ -7,6 +7,7 @@
   <summary>Cross-Platform GUI Library</summary>
   <url type="homepage">https://www.wxwidgets.org/</url>
   <releases>
+    <release version="3.2.1" date="2022-09-08"/>
     <release version="3.2.0" date="2022-07-07"/>
   </releases>
 </component>
--- a/org.freedesktop.Sdk.Extension.wxwidgets.metainfo.xml
+++ b/org.freedesktop.Sdk.Extension.wxwidgets.metainfo.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="runtime">
+  <id>org.freedesktop.Sdk.Extension.wxwidgets</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>LGPL-2.1-or-later WITH WxWindows-exception-3.1</project_license>
+  <name>WxWidgets Sdk extension</name>
+  <summary>Cross-Platform GUI Library</summary>
+  <url type="homepage">https://www.wxwidgets.org/</url>
+  <releases>
+    <release version="3.2.0" date="2022-07-07"/>
+  </releases>
+</component>
--- a/org.freedesktop.Sdk.Extension.wxwidgets.yaml
+++ b/org.freedesktop.Sdk.Extension.wxwidgets.yaml
@@ -1,7 +1,7 @@
 id: org.freedesktop.Sdk.Extension.wxwidgets
-branch: '21.08'
+branch: '22.08'
 runtime: org.freedesktop.Sdk
-runtime-version: '21.08'
+runtime-version: '22.08'
 sdk: org.freedesktop.Sdk
 sdk-extensions:
   - org.freedesktop.Sdk.Extension.vala
--- a/org.freedesktop.Sdk.Extension.wxwidgets.yaml
+++ b/org.freedesktop.Sdk.Extension.wxwidgets.yaml
@@ -186,7 +186,6 @@ modules:
                 url: https://github.com/google/woff2.git
                 commit: 4721483ad780ee2b63cb787bfee4aa64b61a0446
                 x-checker-data:
-                  is-main-source: true
                   type: json
                   url: https://api.github.com/repos/google/woff2/commits?sha=master&per_page=1
                   commit-query: .[].sha
--- a/org.freedesktop.Sdk.Extension.wxwidgets.yaml
+++ b/org.freedesktop.Sdk.Extension.wxwidgets.yaml
@@ -23,8 +23,8 @@ modules:
   - name: wxwidgets
     sources:
       - type: archive
-        url: https://github.com/wxWidgets/wxWidgets/releases/download/v3.2.0/wxWidgets-3.2.0.tar.bz2
-        sha256: 356e9b55f1ae3d58ae1fed61478e9b754d46b820913e3bfbc971c50377c1903a
+        url: https://github.com/wxWidgets/wxWidgets/releases/download/v3.2.1/wxWidgets-3.2.1.tar.bz2
+        sha256: c229976bb413eb88e45cb5dfb68b27890d450149c09b331abd751e7ae0f5fa66
         x-checker-data:
           is-main-source: true
           type: html
@@ -62,8 +62,8 @@ modules:
           - -DENABLE_BUBBLEWRAP_SANDBOX=OFF
         sources:
           - type: archive
-            url: https://webkitgtk.org/releases/webkitgtk-2.36.4.tar.xz
-            sha256: b6bebe1f85a479d968c19e44a4704622ef8cef61636ad1b2406b77d16ae2e2a8
+            url: https://webkitgtk.org/releases/webkitgtk-2.36.7.tar.xz
+            sha256: 0c260cf2b32f0481d017670dfed1b61e554967cd067195606c9f9eb5fe731743
             x-checker-data:
               is-important: true
               type: anitya
@@ -98,8 +98,8 @@ modules:
               - -Ddocbook_docs=disabled
             sources:
               - type: archive
-                url: https://gitlab.gnome.org/GNOME/libnotify/-/archive/0.7.12/libnotify-0.7.12.tar.gz
-                sha256: dd033c656cdb564debcfeeaf172519b590f270f7a19309a510d2a02eddfc98ff
+                url: https://gitlab.gnome.org/GNOME/libnotify/-/archive/0.8.1/libnotify-0.8.1.tar.gz
+                sha256: 7c0b252edecbf08db50d775f9e720ecc03c742fb97c25f3966a8b7a4bedf8133
                 x-checker-data:
                   type: anitya
                   project-id: 13149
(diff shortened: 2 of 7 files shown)


tinywrkb:
bot, build org.freedesktop.Sdk.Extension.wxwidgets

tinywrkb:
Per wxWidgets' developer recommendation, I switch to wxWidgets 3.1.x development releases, and will switch back to stable after the 3.2.0 release will be out.

What's blocking this now is mainly doing more testing.  
I might revise the deployment script, `install.sh`, though that much lower priority, and might be done post merge.

tinywrkb:
bot, build org.freedesktop.Sdk.Extension.wxwidgets

org.freedesktop.Sdk.Extension.wxwidgets.metainfo.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="runtime">
+  <id>org.freedesktop.Sdk.Extension.wxwidgets</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>wxWindows</project_license>
```
hfiguiere, line 5: This should do it
```suggestion
  <project_license>LGPL-2.1-or-later WITH WxWindows-exception-3.1</project_license>
```

(it's a valid SPDX identifier)

org.freedesktop.Sdk.Extension.wxwidgets.yaml
```diff
@@ -0,0 +1,209 @@
+id: org.freedesktop.Sdk.Extension.wxwidgets
```
hfiguiere, line 1: I would rather take the BaseApp route instead.

tinywrkb: As written in the 1st post, this is designed for wxWidgets development with Codeblocks IDE or any other IDEs.  
Building WebKitGTK without ccache hits takes a substantial amount of time, and I would not bother with it if I didn't want an SDK extension.  
Even if there was a base application, I would still submit this extension.

hfiguiere: ah ok, nevermind then.

org.freedesktop.Sdk.Extension.wxwidgets.metainfo.xml
tinywrkb: Fixed in 10e425720ae72039fb551a46c0d369a6fe2e3dec

tinywrkb:
bot, build org.freedesktop.Sdk.Extension.wxwidgets

tinywrkb:
bot, build org.freedesktop.Sdk.Extension.wxwidgets

tinywrkb:
bot, build org.freedesktop.Sdk.Extension.wxwidgets

tinywrkb:
This should be ready now.  
I went through all the code samples of wxWidgets, and they can be built and run correctly as far as I can tell.  
Also built and run with this extension the Flathub app `net.sourceforge.GrandOrgue`, and Codeblock (svn trunk version).

tinywrkb:
Switching to draft to evaluate building WebKitGTK with full debugging symbols.

4c0n:
Great! Making this available as a SDK extension makes it flexible and re-usable. :smile:
I'm wondering if it would be a good idea to have WebkitGTK as a separate extension as there seem to be some other packages that currently have the `--disable-webkit` or `--without-webkit` flags in their builds, maybe because it's too time consuming to include it, but of course I can't be sure about that.

tinywrkb:
@4c0n 

This is already being handled by the `install.sh` script, and the next push here will have examples.  
I'm currently rewriting the script, as it has a few issues, and I want it to fail if new shared libs or locales were detected to limit the maintenance burden of having to keep track of these changes.  

Next push here will have updated `install.sh`, examples, and will test full debug symbols for WebKitGTK.

tinywrkb:
bot, build org.freedesktop.Sdk.Extension.wxwidgets

tinywrkb:
The contribution of WebKitGTK's debug symbols to the size of this extension:

- `-g0` (no debug symbols): extension total size is ~400MB.
- `-g1`: extension total size is ~1GB
- `-g` (flatpak-builder's default): extension total size is ~4.5GB. Almost an extra 30 minutes to build the extension.  
  Also, creating `.Debug` extension for applications becomes a very costly operation.

I'm in favor of switching back setting `-g1`.

tinywrkb:
bot, build org.freedesktop.Sdk.Extension.wxwidgets

tinywrkb:
> Hence the reason for wondering if maybe it would be better to place it in a separate sdk extension.

No, there's no way to keep these extensions in sync and not break ABI without manual intervention of a Flathub admin keeping ostree refs from being published until both extensions are rebuilt.  
We will be going to need a separate Flatpak branch for each WebKitGTK release that breaks ABI, and after it's publish to update wxWidgets to use the new WebKitGTK branch.  
I asked similar question about handling cross-Flatpak-ref ABI dependency [here](https://github.com/flathub/flathub/issues/2533).

tinywrkb:
Last push just renames the examples.

tinywrkb:
> > Hence the reason for wondering if maybe it would be better to place it in a separate sdk extension.
> 
> No, there's no way to keep these extensions in sync and not break ABI without manual intervention of a Flathub admin keeping ostree refs from being published until both extensions are rebuilt. We will be going to need a separate Flatpak branch for each WebKitGTK release that breaks ABI, and after it's published to update wxWidgets to use the new WebKitGTK branch. I asked similar question about handling cross-Flatpak-ref ABI dependency [here](https://github.com/flathub/flathub/issues/2533).

After considering this again, I see that WebKitGTK has semi-annual schedule for major releases, so maybe if we're not strict as the Freedesktop runtime about adding new symbols, and don't mind stating that we can't promise that downgrading the WebKitGTK extension won't break ABI of code linked against the latest WebKitGTK build, then I guess a WebKitGTK extension is achievable without too much trouble.  
We will need 2 branches for each runtime version, so a total of 4 maintained branches, and maybe retire some of them earlier than the runtime.  
I'm not sure yet what the best way to wire-up ABI checks of minor releases, or should we just trust upstream.

tinywrkb:
> > > Hence the reason for wondering if maybe it would be better to place it in a separate sdk extension.
> > 
> > 
> > No, there's no way to keep these extensions in sync and not break ABI without manual intervention of a Flathub admin keeping ostree refs from being published until both extensions are rebuilt. We will be going to need a separate Flatpak branch for each WebKitGTK release that breaks ABI, and after it's published to update wxWidgets to use the new WebKitGTK branch. I asked similar question about handling cross-Flatpak-ref ABI dependency [here](https://github.com/flathub/flathub/issues/2533).
> 
> After considering this again, I see that WebKitGTK has semi-annual schedule for major releases, so maybe if we're not strict as the Freedesktop runtime about adding new symbols, and don't mind stating that we can't promise that downgrading the WebKitGTK extension won't break ABI of code linked against the latest WebKitGTK build, then I guess a WebKitGTK extension is achievable without too much trouble. We will need 2 branches for each runtime version, so a total of 4 maintained branches, and maybe retire some of them earlier than the runtime. I'm not sure yet what the best way to wire-up ABI checks of minor releases, or should we just trust upstream.

I take this back. I've been criticized for doing something similar in an extension. Avoiding ABI breaking by setting a branch version.  
I'm not going to bother with this when no official Flathub guidelines document how to avoid ABI breakage between different Flatpak refs/packages.  
This is a Flatpak limitation that should be fixed, and I won't try to work around.

I have no further changes planned to this PR, and it's ready to be merged.

4c0n:
> > > > Hence the reason for wondering if maybe it would be better to place it in a separate sdk extension.
> > > 
> > > 
> > > No, there's no way to keep these extensions in sync and not break ABI without manual intervention of a Flathub admin keeping ostree refs from being published until both extensions are rebuilt. We will be going to need a separate Flatpak branch for each WebKitGTK release that breaks ABI, and after it's published to update wxWidgets to use the new WebKitGTK branch. I asked similar question about handling cross-Flatpak-ref ABI dependency [here](https://github.com/flathub/flathub/issues/2533).
> > 
> > After considering this again, I see that WebKitGTK has semi-annual schedule for major releases, so maybe if we're not strict as the Freedesktop runtime about adding new symbols, and don't mind stating that we can't promise that downgrading the WebKitGTK extension won't break ABI of code linked against the latest WebKitGTK build, then I guess a WebKitGTK extension is achievable without too much trouble. We will need 2 branches for each runtime version, so a total of 4 maintained branches, and maybe retire some of them earlier than the runtime. I'm not sure yet what the best way to wire-up ABI checks of minor releases, or should we just trust upstream.
> 
> I take this back. I've been criticized for doing something similar in an extension. Avoiding ABI breaking by setting a branch version.  
> I'm not going to bother with this when no official Flathub guidelines document how to avoid ABI breakage between different Flatpak refs/packages.  
> This is a Flatpak limitation that should be fixed, and I won't try to work around.
> 
> I have no further changes planned to this PR, and it's ready to be merged.

Totally agree that it would be great to have more guidelines and information on this subject.
Too bad communication seems difficult when it comes to matters like this, I recently tried proposing for more guidelines on the discourse forum, but also did not get any useful response so far.
I truly hope communication will improve for Flathub and its community.

tinywrkb:
bot, build org.freedesktop.Sdk.Extension.wxwidgets

tinywrkb:
bot, build org.freedesktop.Sdk.Extension.wxwidgets

tinywrkb:
bot, build org.freedesktop.Sdk.Extension.wxwidgets

PR #3084, opened 2022-04-30
Add io.github.elograf.Elograf

papoteur-mga:


io.github.elograf.Elograf.metainfo.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>io.github.elograf.Elograf</id>
```
hfiguiere, line 3: this app-id is invalid. There is no "elograf" organisation or user on github.

io.github.elograf.Elograf.yaml
```diff
@@ -0,0 +1,58 @@
...
+      url: https://github.com/jordansissel/xdotool/releases/download/v3.20211022.1/xdotool-3.20211022.1.tar.gz
+      sha256: 96f0facfde6d78eacad35b91b0f46fecd0b35e474c03e00e30da3fdd345f9ada
+  - name: PyQtApp
+  - name: nerd-dictation
+    buildsystem: simple
+    build-commands: 
+      - "cp nerd-dictation /app/bin/"
+    sources:
+        - type: git
+          url: https://github.com/ideasman42/nerd-dictation
```
hfiguiere, line 31: you need a release / commit here.

io.github.elograf.Elograf.yaml
```diff
@@ -0,0 +1,58 @@
...
+        --prefix=${FLATPAK_DEST} "vosk" --no-build-isolation
+    sources:
+    - type: file
+      url: https://files.pythonhosted.org/packages/62/d5/5f610ebe421e85889f2e55e33b7f9a6795bd982198517d912eb1c76e1a53/pycparser-2.21-py2.py3-none-any.whl
+      sha256: 8ee45429555515e1f6b185e78100aea234072576aa43ab53aefcae078162fca9
+    - type: file
+      url: https://files.pythonhosted.org/packages/00/9e/92de7e1217ccc3d5f352ba21e52398372525765b2e0c4530e6eb2ba9282a/cffi-1.15.0.tar.gz
+      sha256: 920f0d66a896c2d99f0adbb391f990a84091179542c205fa53ce5787aff87954
+    - type: file
+      url: https://files.pythonhosted.org/packages/d4/e2/25a02e009538ed03371db59fca54a278192562fea523e851b4842b4a182f/vosk-0.3.37-py3-none-manylinux_2_12_x86_64.manylinux2010_x86_64.whl
```
hfiguiere, line 45: this is x86_64 only. You should get the tar.gz instead to have it built from source.

io.github.elograf.Elograf.yaml
```diff
@@ -0,0 +1,58 @@
...
+      sha256: a2bdd3cc30db751af04611088a1658850865ebbccbeaa006c3fe791105250b62
+  - name: elograf
+    buildsystem: simple
+    build-commands:
+      - "python3 setup.py install --prefix=/app"
+    sources:
+        - type: git
+          url: "https://github.com/papoteur-mga/elograf"
```
hfiguiere, line 53: you need a revision / commit here

io.github.elograf.Elograf.yaml
```diff
@@ -0,0 +1,58 @@
...
+finish-args:
```
hfiguiere, line 54: finish-args are usually above module.

io.github.elograf.Elograf.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - --share=ipc
+    - --socket=x11
```
hfiguiere, line 56: it doesn't support wayland?

io.github.elograf.Elograf.yaml
```diff
@@ -0,0 +1,58 @@
...
+  - name: xdotool
+    buildsystem: simple
+    build-commands:
+      - make
+      - make install PREFIX=/app 
+    sources:
+    - type: archive
+      url: https://github.com/jordansissel/xdotool/releases/download/v3.20211022.1/xdotool-3.20211022.1.tar.gz
+      sha256: 96f0facfde6d78eacad35b91b0f46fecd0b35e474c03e00e30da3fdd345f9ada
+  - name: PyQtApp
```
hfiguiere, line 24: what's this empty module?

hfiguiere (changes requested):

io.github.elograf.Elograf.metainfo.xml
papoteur-mga: Fixed

io.github.elograf.Elograf.yaml
papoteur-mga: Fixed with a commit.

io.github.elograf.Elograf.yaml
papoteur-mga: Fixed with a tag

io.github.elograf.Elograf.yaml
papoteur-mga: Moved

io.github.elograf.Elograf.yaml
papoteur-mga: No, it uses xdotool to simulate the keyboard. There is not yet equivalent for Wayland.

io.github.elograf.Elograf.yaml
papoteur-mga: I don't intend to use the compilation way for this part. It would need to include kaldi and the libs needed for it build, like openfst, clapack... at least.
What I expect is to use to part provided by pypi.org. But I don't know if this is possible or allowed.

io.github.papoteur-mga.Elograf.yaml
```diff
@@ -0,0 +1,60 @@
+id: io.github.papoteur-mga.Elograf
```
hfiguiere, line 1: the app-id is still invalid: see https://docs.flatpak.org/en/latest/conventions.html#application-ids

(`-` as not allowed where you have one)

Also it must be changed in the metainfo.xml

io.github.elograf.Elograf.yaml
hfiguiere: so how does this work on aarch64 that we build for?

(hint it DOES NOT)

PR #3100, opened 2022-05-03
Add Flatpak bundle for PoMiDAQ

ximion:
Hi!

This adds a Flatpak bundle for the PoMiDAQ application which enables user to acquire data with the [UCLA Miniscope](https://github.com/Aharoni-Lab/Miniscope-v4) platform on Linux in neuroscientific applications.


io.github.bothlab.pomidaq.yaml
```diff
@@ -0,0 +1,68 @@
...
+runtime-version: 5.15-21.08
+sdk: org.kde.Sdk
+command: pomidaq
+
+finish-args:
+  - '--share=ipc'
+  - '--socket=wayland'
+  - '--socket=fallback-x11'
+  - '--device=all'
+  - '--filesystem=host'
```
hfiguiere, line 12: this is a very broad filesystem permision

hfiguiere:
bot, build io.github.bothlab.pomidaq

hfiguiere (changes requested):

io.github.bothlab.pomidaq.yaml
ximion: Yes, but a lot of scientists have files in a lot of places, for example in our lab data is stored on devices mounted in `/mnt` and directories in `/srv`, other labs have network drives (terrible idea for live recording storage, but still people do that), I have seen people mounting stuff in `/var` etc.
I don't want to limit where data can be exported to, as that will most certainly lead to user frustration.
Maybe marking things like `/usr` as `ro` would be okay as compromise? Otherwise I'd need to add quite a large whitelist of possible locations where scientists could store data...

hfiguiere:
`x86_64`:
```
builddir/files/share/appdata/io.github.bothlab.pomidaq.appdata.xml: FAILED:
Validation of files failed
```

`aarch64`:

opencv need some patches as usual

ximion:
The release info stuff requires `appstreamcli`, which the runtime apparently doesn't include. So I'll have to build that as well for this bundle.

hfiguiere:
this should be generated when the tarball is made, by upstream.

ximion:
This should pass now. I also disabled Carotene (which seems to be ARM specific anyway), so this might even build on ARM now. The performance impact should be minimal, as FFmpeg does most of the expensive work already.

hfiguiere:
bot, build io.github.bothlab.pomidaq

ximion:
Hmm, I do not understand this error. This software installs a vectorgraphic icon as `prefix/share/icons/hicolor/scalable/apps/pomidaq.svg` which should be easily rendered into 64x64 and 128x128px icons by any software that needs it...

hfiguiere:
if you put `rename-icon: pomidaq` in the top-level it will rename the icom properly.

io.github.bothlab.pomidaq.yaml
```diff
@@ -0,0 +1,113 @@
+id: io.github.bothlab.pomidaq
+runtime: org.kde.Platform
+runtime-version: 5.15-21.08
+sdk: org.kde.Sdk
+command: pomidaq
```
hfiguiere, line 5:
```suggestion
command: pomidaq
rename-icon: pomidaq
```

hfiguiere:
see my suggestion. The desktop file need the same.

ximion:
Ah yes, that does make sense! I forgot about this, and probably would have looked for a solution for quite a while, given that the error message would have sent me in a completely different direction...
Thanks for the hint and fix! :-)

ximion:
This should build now (maybe even on ARM, but I couldn't test that locally).

hfiguiere:
bot, build io.github.bothlab.pomidaq

hfiguiere:
it seems that despite using Qt, it tries to use openGL directly, and sadly the KDE runtime is built with only GLES.

that's a valid reason to skip `aarch64`

ximion:
> it seems that despite using Qt, it tries to use openGL directly, and sadly the KDE runtime is built with only GLES.

Yes indeed, I use OpenGL to render visualizations directly - could maybe be ported to GLES, but I don't think I have the time to work on that, especially since this software isn't currently used on any arm device.

> that's a valid reason to skip `aarch64`

Can I just ignore the build failure, or is there any way to specify unsupported architectures? If the KDE runtime changes in future, this might build on aarch64, but not right now...

hfiguiere:

ximion:
I've added this, and will check in future whether I can compile this with GLES (GLES compatibility wasn't a priority during development, but it may just work...).

hfiguiere:
you can also


bot, build io.github.bothlab.pomidaq

ximion:
Ah, I didn't know the bot would listen to me :-)

ximion:
This indeed would need some serious GLES porting work, so let's build only on platforms supporting desktop OpenGL for now.

bot, build io.github.bothlab.pomidaq

PR #3111, opened 2022-05-07
Add org.kryogenix.Pick

stuartlangridge:

### DIFF
--- a/org.kryogenix.Pick.yml
+++ b/org.kryogenix.Pick.yml
@@ -1,0 +1,37 @@
+app-id: org.kryogenix.Pick
+runtime: org.gnome.Platform
+runtime-version: '42'
+sdk: org.gnome.Sdk
+command: pick-colour-picker
+
+finish-args:
+  - --share=ipc
+  - --socket=pulseaudio
+  - --socket=x11
+
+modules:
+  - name: pick-colour-picker
+    buildsystem: simple
+    build-commands:
+      - install -D pick-colour-picker-flatpak-runscript.sh /app/bin/pick-colour-picker
+      - chmod a+x /app/bin/pick-colour-picker
+      - cp -r pick /app/pick
+      - rm -rf /app/pick/__pycache__
+      - mkdir -p /app/share/metainfo/
+      - sed 's/pick-colour-picker.desktop/org.kryogenix.Pick.desktop/' pick-colour-picker.appdata.xml > /app/share/metainfo/org.kryogenix.Pick.appdata.xml
+      - install -D data/icons/scalable/apps/pick-colour-picker-symbolic.svg /app/share/icons/hicolor/scalable/apps/org.kryogenix.Pick.svg
+      - install -D data/icons/512x512/apps/pick-colour-picker.png /app/share/icons/hicolor/512x512/apps/org.kryogenix.Pick.png
+      - install -D data/icons/128x128/apps/pick-colour-picker.png /app/share/icons/hicolor/128x128/apps/org.kryogenix.Pick.png
+      - install -D data/icons/64x64/apps/pick-colour-picker.png /app/share/icons/hicolor/64x64/apps/org.kryogenix.Pick.png
+      - install -D data/icons/48x48/apps/pick-colour-picker.png /app/share/icons/hicolor/48x48/apps/org.kryogenix.Pick.png
+      - install -D data/icons/32x32/apps/pick-colour-picker.png /app/share/icons/hicolor/32x32/apps/org.kryogenix.Pick.png
+      - install -D data/icons/24x24/apps/pick-colour-picker.png /app/share/icons/hicolor/24x24/apps/org.kryogenix.Pick.png
+      - install -D data/icons/22x22/apps/pick-colour-picker.png /app/share/icons/hicolor/22x22/apps/org.kryogenix.Pick.png
+      - install -D data/icons/16x16/apps/pick-colour-picker.png /app/share/icons/hicolor/16x16/apps/org.kryogenix.Pick.png
+      - mkdir -p /app/share/applications/
+      - sed 's/Icon=.*/Icon=org.kryogenix.Pick/' pick-colour-picker.desktop > /app/share/applications/org.kryogenix.Pick.desktop
+    sources:
+      - type: git
+        url: https://github.com/stuartlangridge/ColourPicker.git
+        branch: app
+        commit: 946322bae8cf8805d3ede179fb92cce2a00a2d26
--- a/org.kryogenix.Pick.yml
+++ b/org.kryogenix.Pick.yml
@@ -9,6 +9,10 @@ finish-args:
   - --socket=pulseaudio
   - --socket=x11
 
+rename-desktop-file: pick-colour-picker.desktop
+rename-appdata-file: pick-colour-picker.appdata.xml
+rename-icon: pick-colour-picker
+
 modules:
   - name: pick-colour-picker
     buildsystem: simple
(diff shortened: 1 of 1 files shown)


org.kryogenix.Pick.yml
```diff
@@ -0,0 +1,37 @@
...
+      - install -D data/icons/512x512/apps/pick-colour-picker.png /app/share/icons/hicolor/512x512/apps/org.kryogenix.Pick.png
+      - install -D data/icons/128x128/apps/pick-colour-picker.png /app/share/icons/hicolor/128x128/apps/org.kryogenix.Pick.png
+      - install -D data/icons/64x64/apps/pick-colour-picker.png /app/share/icons/hicolor/64x64/apps/org.kryogenix.Pick.png
+      - install -D data/icons/48x48/apps/pick-colour-picker.png /app/share/icons/hicolor/48x48/apps/org.kryogenix.Pick.png
+      - install -D data/icons/32x32/apps/pick-colour-picker.png /app/share/icons/hicolor/32x32/apps/org.kryogenix.Pick.png
+      - install -D data/icons/24x24/apps/pick-colour-picker.png /app/share/icons/hicolor/24x24/apps/org.kryogenix.Pick.png
+      - install -D data/icons/22x22/apps/pick-colour-picker.png /app/share/icons/hicolor/22x22/apps/org.kryogenix.Pick.png
+      - install -D data/icons/16x16/apps/pick-colour-picker.png /app/share/icons/hicolor/16x16/apps/org.kryogenix.Pick.png
+      - mkdir -p /app/share/applications/
+      - sed 's/Icon=.*/Icon=org.kryogenix.Pick/' pick-colour-picker.desktop > /app/share/applications/org.kryogenix.Pick.desktop
```
hfiguiere, line 32: there is `desktop-file-edit` to avoid having to use sed.

Also if you use `rename-icon` then you don't have do bother.

Similarly, just install the desktop file as is and use `rename-desktop-file`, this will also avoid having to do the sed command  line 21.

ie all of this can be made much shorter and simpler

org.kryogenix.Pick.yml
```diff
@@ -0,0 +1,37 @@
...
+    sources:
+      - type: git
+        url: https://github.com/stuartlangridge/ColourPicker.git
+        branch: app
```
hfiguiere, line 36:
```suggestion
```
never branches.

org.kryogenix.Pick.yml
```diff
@@ -0,0 +1,37 @@
+app-id: org.kryogenix.Pick
+runtime: org.gnome.Platform
+runtime-version: '42'
+sdk: org.gnome.Sdk
+command: pick-colour-picker
+
+finish-args:
+  - --share=ipc
+  - --socket=pulseaudio
+  - --socket=x11
```
hfiguiere, line 10: it doesn't support wayland? If you use the GNOME SDK I'll assume it's because it uses pygtk, which mean that it's gtk3 or 4, so wayland.

hfiguiere (changes requested):

hfiguiere:
bot, build org.kryogenix.Pick

org.kryogenix.Pick.yml
stuartlangridge: I don’t understand what this means? The app is in a git branch called “app”; that’s where it is. This isn’t a feature branch or anything; there isn’t a “main” or “master” branch.

org.kryogenix.Pick.yml
stuartlangridge: It picks colours off the screen. As I understand it, this doesn’t work in wayland and can’t be done unless you’re the shell, so Pick doesn’t do Wayland. If that’s incorrect and gtk glosses over the difference then I’m happy to look at supporting wayland in a future release.

org.kryogenix.Pick.yml
cassidyjames: Yeah this could use a bit more explanation, but my understanding is if you're pointing to a specific commit, then specifying a branch isn't strictly necessary, at least.

hfiguiere: worse. branch get new commits which mean that the build will fail.

we want to reproduce build so it has to be a specific commit and if you specify a branch it will breach.

I personally wished we just made this a fatal error to use branches, but I have no control over this.

org.kryogenix.Pick.yml
hfiguiere: nah that's a reason good enough,

org.kryogenix.Pick.yml
stuartlangridge: Just so I understand this: I do currently specify a commit, and that’s fine, but I also specify a branch, and I should not do so; it’s not just irrelevant, as @cassidyjames  suggests, but it’s actually _wrong_ because newer pushes to that branch will get built and ignore the commit I specified?

hfiguiere: exactly what I'm saying.

ideally you'd specify a tag for a release, but I'm not hung on that part. just a commit is perfectly fine.

org.kryogenix.Pick.yml
stuartlangridge: OK, I don't think I understand `rename-icon`.
I can use `rename-desktop-file` and `rename-appdata-file` OK, but rename-icon doesn't seem to work in my testing.

I have icons in the source tree in `data/icons/{sizes}/apps/pick-colour-picker.png`, which I copy into the flatpak build tree with `cp -r data/icons /app/share/icons`. So, at that point in the build process, `...share/icons/24x24/apps/pick-colour-picker.png` exists, as does `.../share/icons/48x48/apps/pick-colour-picker.png`, etc, etc; there are many copies of this icon at different sizes. Which value do I need to put in `rename-icon` to get it to rename these icons?

I have tried `rename-icon: pick-colour-picker.png` and `rename-icon: apps/pick-colour-picker.png` and neither work: both throw this error:

`Error: icon apps/pick-colour-picker.png not found below /path/to/.flatpak-builder/rofiles/rofiles-PhUUsC/files/share/icons`

so I'm a bit lost as to why that isn't working. If you have suggestions I'd be happy to hear them. (The flatpak docs don't explain a great deal about the rename- properties other than to name them :-))

hfiguiere: the documentation say to use the **icon name** which doesn't include the extension. Because you could have a .svg icon too (or only)

stuartlangridge: the docs do say that. Unfortunately, it doesn't help; I tried "rename-icon: pick-colour-picker" first, and I get the same error. (This is the part I don't understand.)

`Error: icon pick-colour-picker not found below .../.flatpak-builder/rofiles/rofiles-KzpfEQ/files/share/icons`

hfiguiere: it only work if the icons are installed in the right place, aka in `${FLATPAK_DEST}/share/icons/hicolor/SIZE/apps/ICON_NAME`

stuartlangridge: aha! `hicolor`! Thank you. Now it works. Incoming update to the PR to fix all these issues :-)

stuartlangridge: I believe the below pushed update should resolve these issues, and thank you for the patience!

org.kryogenix.Pick.yml
stuartlangridge: I believe the below pushed update should resolve these issues, and thank you for the patience!

org.kryogenix.Pick.yml
```diff
@@ -0,0 +1,33 @@
...
+
+modules:
+  - name: pick-colour-picker
+    buildsystem: simple
+    build-commands:
+      - install -D pick-colour-picker-flatpak-runscript.sh /app/bin/pick-colour-picker
+      - chmod a+x /app/bin/pick-colour-picker
+      - cp -r pick /app/pick
+      - rm -rf /app/pick/__pycache__
+      - mkdir -p /app/share/metainfo/
```
hfiguiere, line 24:
```suggestion
```

install will do that for you

org.kryogenix.Pick.yml
```diff
@@ -0,0 +1,33 @@
...
+      - mkdir -p /app/share/icons
```
hfiguiere, line 25:
```suggestion
      - install -d /app/share/icons
```
it's equivalent though.

org.kryogenix.Pick.yml
```diff
@@ -0,0 +1,33 @@
...
+rename-desktop-file: pick-colour-picker.desktop
+rename-appdata-file: pick-colour-picker.appdata.xml
+rename-icon: pick-colour-picker
+
+modules:
+  - name: pick-colour-picker
+    buildsystem: simple
+    build-commands:
+      - install -D pick-colour-picker-flatpak-runscript.sh /app/bin/pick-colour-picker
+      - chmod a+x /app/bin/pick-colour-picker
```
hfiguiere, line 21:
```suggestion
      - install -Dm755 pick-colour-picker-flatpak-runscript.sh /app/bin/pick-colour-picker
```

install can do that for you

hfiguiere (changes requested):

hfiguiere:
bot, build org.kryogenix.Pick

org.kryogenix.Pick.yml
stuartlangridge: updated as suggested for these changes.

hfiguiere:
btw you also can:

bot, build org.kryogenix.Pick

stuartlangridge:
> btw you also can: (run the bot build command)

I'm happy to, if I know why,. I don't know what the bot is for or who needs its output. Is the idea that it runs and builds a flatpak and then you test that to see if it works? Is it mostly to check that the build works successfully? I'm happy to kick off a build command every time I push an updated flatpak manifest if that's helpful!

hfiguiere:
if it doesn't build it doesn't get merged. so it needs to be built. that's just CI expect that for the submissions it's not automatic.

hfiguiere:
So I have installed it and beside the few deprecation warnings, it actually doesn't really work when running wayland.

It doesn't pick colour from wayland client (Firefox, gnome-terminal)  and for X11 (Xwayland) it's always 0,0,0.

Just FYI.

stuartlangridge:
> So I have installed it and beside the few deprecation warnings, it actually doesn't really work when running wayland.
> It doesn't pick colour from wayland client (Firefox, gnome-terminal) and for X11 (Xwayland) it's always 0,0,0.
> Just FYI.

Thank you for that test! I suspected that was the case -- Pick can't be done in a Wayland world without a shell API to call to do it, and we're not all that near standardisation on that front across desktops yet as I understand it -- but it's useful to have it confirmed, thank you :-)

PR #3122, opened 2022-05-08
Add Return to the Roots

Mailaender:

- [X] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://github.com/Return-To-The-Roots/s25client/issues/1524
- [X] Any additional patches or files have been submitted to the upstream projects concerned. https://github.com/Return-To-The-Roots/s25client/pull/1523, https://github.com/Return-To-The-Roots/s25client/pull/1518 https://github.com/Return-To-The-Roots/s25client/pull/1529


Mailaender:
bot, build com.github.Return-to-the-Roots

Mailaender:
bot, build com.github.Return-to-the-Roots

com.github.Return-to-the-Roots.yaml
```diff
@@ -0,0 +1,91 @@
...
+    sources:
+    - type: archive
+      url: https://libsdl.org/release/SDL2-2.0.22.tar.gz
+      sha256: fe7cbf3127882e3fc7259a75a0cb585620272c51745d3852ab9dd87960697f2e
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+    - -DCMAKE_BUILD_TYPE=Release
+
+  - name: SDL2_mixer
```
hfiguiere, line 47: SDL2_mixer 2.0.4 is in the runtime

com.github.Return-to-the-Roots.yaml
```diff
@@ -0,0 +1,91 @@
...
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DNOWIDE_SYSTEM_INCLUDE=ON
+    sources:
+      - type: git
+        url: https://github.com/boostorg/nowide.git
+        tag: v11.1.4
+        commit: b4c10155171b782fc858ae1aadcfdbe34e4940d4
+
+  - name: SDL2
```
hfiguiere, line 37: sdl 2.0.16 is in the runtime. isn't it enough?

hfiguiere (changes requested):

com.github.Return-to-the-Roots.yaml
Mailaender: I got CMake errors about it not being found.

Mailaender: > CMake Error at /usr/share/cmake-3.22/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find SDL2 (missing: SDL2_SDL2MAIN_LIBRARY) (found version
  "2.0.16")
Call Stack (most recent call first):
  /usr/share/cmake-3.22/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
  cmake/Modules/FindSDL2.cmake:87 (find_package_handle_standard_args)
  external/s25edit/SGE/CMakeLists.txt:6 (FIND_PACKAGE)

com.github.Return-to-the-Roots.yaml
Mailaender: Indeed it is not required.

Mailaender:
bot, build com.github.Return-to-the-Roots

Mailaender:
bot, build com.github.Return-to-the-Roots

com.github.Return-to-the-Roots.yaml
Flow86: I only have a guess, but could it be that its the same issue like in nixpkgs, that SDLmain.a was removed accidently? see https://github.com/Return-To-The-Roots/s25client/pull/1511

barthalion: We generally want dynamic linking for Flatpak packages too, as it improves deduplication for updates. From downstream perspective, CMake shouldn't bail out completely only because static library is missing.

barthalion:
```
[bpiotrowski@risuku ~]$ flatpak run com.github.Return-to-the-Roots
Return To The Roots v20220508-397f2b2315e997504d4958bfbdea0af815ce559a
Compiled with GNU C++ version 11.3.0 for linux 64 Bit

Starting in "/app"
Directory for user data (config etc.): "/var/home/bpiotrowski/.s25rttr"


WARNING: Could not use settings from "/var/home/bpiotrowski/.s25rttr/CONFIG.INI", using default values. Reason: File missing or invalid
Searching for drivers in "/app/lib/s25rttr/driver/video"
1 video drivers found!
Loaded video driver "(SDL2) OpenGL via SDL2-Library"
OpenGL 4.6 supported
Searching for drivers in "/app/lib/s25rttr/driver/audio"
1 audio drivers found!
Loaded audio driver "(SDL2) Audio via SDL2_mixer-Library"

Starting the game
Loading "/app/share/s25rttr/RTTR/assets/base/splash.bmp": done in 2ms
Exception caught: File or directory does not exist: "/app/share/s25rttr/S2/GFX/PALETTE/PAL5.BBM"

Failed to load "/app/share/s25rttr/S2/GFX/PALETTE/PAL5.BBM"


ERROR: Some files failed to load.
Please ensure that the Settlers 2 Gold-Edition is installed 
in the same directory as Return to the Roots.
Clearing textures: Finished in 0ms
```

This is suboptimal UX if I need to launch it from CLI to see what it's looking for. E.g. CorsixTH allows to pick location with assets on the first start, alternatively some Quake ports on Flathub use Zenity: https://github.com/flathub/io.github.ezQuake/blob/89ebaefda34778483ba471b86d0979c92a3d32d3/io.github.ezQuake.yaml#L84=

On that note @Mailaender, how did you test it? If it looks for files in `~/.s25rttr`, any way to access this directory is missing from `finish-args`. Ideally, it should be `--persist.s25rttr`, and the error visible in UI should display the host path to that directory.

hfiguiere:
you could just download the data as extra data. and symlink, or something.

Mailaender:
The extra data is still copyrighted and sold on https://www.gog.com/de/game/the_settlers_2_gold_edition plus there is an agreement with Ubisoft about what is allowed and what not.

hfiguiere:
nevermind then.

barthalion:
Anything at `~/.local/share/flatpak/app/` **is not** supposed to be modified by user. RttR needs to start reading files from somewhere configurable like `$XDG_DATA_HOME`. Then you can do a check like ezQuake does, which is somewhat better than nothing; appdata should also mention that user needs to buy the game to use the app.

Mailaender:
[AppData](https://github.com/Return-To-The-Roots/s25client/pull/1523/files#diff-2d16ae74eb151a284e3c1916ad7dadda3f6e9f2e56ca2084d986ed77c2e84c4eR23) already does mention that.

com.github.Return-to-the-Roots.yaml
Mailaender: I experimented with it. It does not only check for the static version, it also requires it to build.

Mailaender:
Updated it to look in `~/.s25rttr` instead.

Mailaender:
bot, build com.github.Return-to-the-Roots

barthalion:
bot, build com.github.Return-to-the-Roots

barthalion:
bot, build com.github.Return-to-the-Roots

Mailaender:
I think that should be done upstream as it loads a GUI for a brief moment which then immediately closes. However, I am not skilled in C++ so this won't happen anytime soon.

bilelmoussaoui:
> I think that should be done upstream as it loads a GUI for a brief moment which then immediately closes. However, I am not skilled in C++ so this won't happen anytime soon.

other games uses plain zenity for that iirc.

Mailaender:
I mean this is a problem on all platforms, not just Flathub. Added a README for documentation.

bilelmoussaoui:
> I mean this is a problem on all platforms, not just Flathub. Added a README for documentation.

README is only read by developers, not end-users that installs the application

barthalion:
Just to be clear, the lack of this message is a blocker. Would it be better if the application displayed assets' location natively? Yes. Does it do it? No. Zenity is a quick workaround to actually inform users what the problem is and how to fix it. I guarantee you regardless of how many other places you update to include the same information, they're unlikely to be read by anyone.

Mailaender:
image

Sorry for the delay. This was my first ever C++ programming in a significant code base.

Mailaender:
bot, build com.github.Return-to-the-Roots

com.github.Return-to-the-Roots.yaml
```diff
@@ -0,0 +1,81 @@
...
+rename-appdata-file: s25rttr.metainfo.xml
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --socket=pulseaudio
+  - --share=network
+  - --device=dri
+  - --filesystem=~/.s25rttr
+
+modules:
```
barthalion, line 17: Misses a global or per-module `cleanup` to remove `/include` and static libs.

com.github.Return-to-the-Roots.yaml
```diff
@@ -0,0 +1,81 @@
...
+rename-desktop-file: s25rttr.desktop
+rename-icon: s25rttr
+rename-appdata-file: s25rttr.metainfo.xml
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --socket=pulseaudio
+  - --share=network
+  - --device=dri
+  - --filesystem=~/.s25rttr
```
barthalion, line 15: I assume this can be removed now?

Mailaender:
bot, build com.github.Return-to-the-Roots

com.github.Return-to-the-Roots.yaml
barthalion: Any reason this is not `--persist`, then?

Mailaender:
bot, build info.rttr.Return-To-The-Roots

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64"]
```
barthalion, line 2: Is it on purpose it's limited to x86_64?

Mailaender: Yes, it fails to build https://buildbot.flathub.org/#/builders/20/builds/6948

PR #3134, opened 2022-05-11
Add net.sourceforge.Kino

tinywrkb:

Kino was declared dead ~10y ago, and the latest release is from 2009, but this still works.  
If you need to capture DV, want to monitor the captured video stream, and not fiddle around with a complicated full fledge video editor (e.g. Kdenlive), then Kino is still a good choice.  
Other simple dvgrab GUIs, like [DVGrabGUI](http://dvgrabgui.sourceforge.net/en/main.html) and [QDvGrab](https://github.com/olielvewen/QDvGrab), are also not being developed anymore.

I've been given a few DV tapes this week to capture.  
I had an old Core 2 board and a VIA VT6307 PCI card lying around, installed Fedora Kinote 35, added this Flatpak, and voilà! Everything worked perfectly, and without any extra step (udev rules, rebuilding kernel, etc...).

The available IEEE 1394 controllers seem to only support PCI, but I've seen reports that PCIe to PCI adapters work, so I suspect that you don't need to source an old board with PCI slots.

~~This might not build on aarch64, considering that Kino brings its own FFmpeg version. **edit: aarch64 is disabled.**~~ This now builds on aarch64, but there are no optimizations for this architecture in the bundled FFmpeg that Kino has in its source. Hopefully it's not too bad when capturing in raw DV mode and previewing the video stream.  

I patched in XDG dirs support, and it seems like it's working correctly.

I'm only using the capture feature, in raw DV mode, which is what I expect that users will want to do with this. Editing and re-encoding should be left for a more modern tool.


tinywrkb:
bot, build net.sourceforge.Kino

tinywrkb:
bot, build net.sourceforge.Kino

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+    "only-arches": ["x86_64"]
```
hfiguiere, line 2: this is disappointing...

hfiguiere (changes requested):

flathub.json
tinywrkb: I can't promise anything, but I'll try to cross-compile locally and see how far I can get. I agree that it would be nice to have this working on aarch64, considering the RPi CM4 exposes PCIe.

hfiguiere: why don't you let the builders handle it?

tinywrkb: I can't stop the build job if make fails, and the CI keeps the job alive for ~2 hours.  
I had to fix 3 modules locally already. That's 6 CI hours.

tinywrkb:
bot, build net.sourceforge.Kino

PR #3149, opened 2022-05-15
Add Pentobi

Mailaender:

- [X] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://github.com/enz/pentobi/issues/10

### DIFF
--- a/io.sourceforge.pentobi.yml
+++ b/io.sourceforge.pentobi.yml
@@ -1,0 +1,28 @@
+app-id: io.sourceforge.pentobi
+runtime: org.kde.Platform
+runtime-version: '5.15-21.08'
+sdk: org.kde.Sdk
+command: pentobi
+rename-icon: pentobi
+finish-args:
+  - --device=dri
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+
+modules:
+  - name: pentobi
+    buildsystem: cmake
+    config-opts:
+      - -DDOCBOOKXSL_DIR=/usr/share/xml/docbook/xml/xsl-stylesheets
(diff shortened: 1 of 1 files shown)


Mailaender:
bot, build io.sourceforge.pentobi

io.sourceforge.pentobi.yml
```diff
@@ -0,0 +1,28 @@
...
+finish-args:
+  - --device=dri
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+
+modules:
+  - name: pentobi
+    buildsystem: cmake
```
hfiguiere, line 16: Cmake-ninja doesn't work?
```suggestion
    buildsystem: cmake-ninja
```

hfiguiere (changes requested):

io.sourceforge.pentobi.yml
Mailaender: No it fails due to cyclic dependencies.

enz:
Great. If you are interested in maintaining the flatpack config, I could add a link from the Pentobi homepage.

Also note that  PENTOBI_OPEN_HELP_EXTERNALLY is no longer neede with Pentobi >= 19.0 because it always uses an external browser (and it won't be needed in the future because Pentobi >= 21.0 will include a custom help viewer impementation, which neither needs QtWebView nor an external browser).

Mailaender:
Removed the option. I think for the help to work you need a browser installed in the sandbox as well (untested) which is a bit unfortunate.

barthalion:
bot, build io.sourceforge.pentobi

barthalion:
Yeah, opening help definitely doesn't work. I know GNOME apps somehow handle this, but I don't remember how. Alternatively, help button could be changed to open an external URL.

PR #3159, opened 2022-05-18
Add com.obsproject.Studio.Plugin.StreamFX

rmnvgr:

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://github.com/Xaymar/obs-StreamFX/issues/781#issuecomment-1121237494
- [x] I own the domain used in the application ID or the domain has a policy for delegating subdomains (e.g. GitHub, SourceForge). (OBS plugin)
We agreed with the developer to ship the patches to the build system downstream.

### DIFF
--- a/flathub.json
+++ b/flathub.json
@@ -1,0 +1,4 @@
+{
+  "skip-icons-check": true,
+  "only-arches": ["x86_64"]
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+.flatpak-builder
+build-dir
--- a/com.obsproject.Studio.Plugin.StreamFX.metainfo.xml
+++ b/com.obsproject.Studio.Plugin.StreamFX.metainfo.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="addon">
+	<id>com.obsproject.Studio.Plugin.StreamFX</id>
+	<extends>com.obsproject.Studio</extends>
+	<name>StreamFX</name>
+	<summary>Additional sources, filters, transitions and encoders for OBS Studio</summary>
+	<description>
+	<p>StreamFX is a plugin to libOBS (OBS Studio, StreamElements OBS Live) that adds new sources, filters and transitions to give you the ability to drastically improve your stream and recording setup past what was possible in OBS Studio until now. Whether it is blurring out a region, moving something in 3D, or just writing your own shaders as a new transition, source or filter - StreamFX has it all.</p>
+	</description>
+	<url type="homepage">https://streamfx.xaymar.com/</url>
+	<url type="bugtracker">https://github.com/Xaymar/obs-StreamFX/issues</url>
+	<url type="contact">https://s.xaymar.com/streamfx-dc</url>
+	<url type="donation">https://github.com/Xaymar/obs-StreamFX/wiki#support-streamfx-on-github-sponsors-or-patreon</url>
+	<url type="translate">https://crowdin.com/project/obs-stream-effects</url>
+	<metadata_license>CC0-1.0</metadata_license>
+	<project_license>GPL-2.0</project_license>
+</component>
--- a/com.obsproject.Studio.Plugin.StreamFX.yml
+++ b/com.obsproject.Studio.Plugin.StreamFX.yml
@@ -1,0 +1,33 @@
+id: com.obsproject.Studio.Plugin.StreamFX
+branch: stable
+runtime: com.obsproject.Studio
+runtime-version: stable
+sdk: org.kde.Sdk//6.4
+build-extension: true
+separate-locales: false
+appstream-compose: false
+build-options:
+  prefix: /app/plugins/StreamFX
+
+modules:
+  - name: streamfx
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DSTRUCTURE_PACKAGEMANAGER=true
+      - -DENABLE_UPDATER=false
+    post-install:
+      - install -Dm644 --target-directory=${FLATPAK_DEST}/share/metainfo ${FLATPAK_ID}.metainfo.xml
+      - appstream-compose --basename=${FLATPAK_ID} --prefix=${FLATPAK_DEST} --origin=flatpak ${FLATPAK_ID}
+    sources:
+      - type: git
+        url: https://github.com/Xaymar/obs-StreamFX.git
+        tag: 0.12.0a173
+        commit: fcb441bdf6e5c70d5eb5ebf6795f2ddd43f70a23
+        disable-shallow-clone: true
+        x-checker-data:
+          type: json
+          url: https://api.github.com/repos/Xaymar/obs-StreamFX/releases/latest
+          tag-query: .tag_name
+          timestamp-query: .published_at
+      - type: file
+        path: com.obsproject.Studio.Plugin.StreamFX.metainfo.xml


streamfx-force-frontend.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
```
rmnvgr, line 1: This patch is necessary to enable the front-end. The build system [checks for the presence](https://github.com/Xaymar/obs-StreamFX/blob/master/CMakeLists.txt#L991) of `${libobs_SOURCE_DIR}/cmake/obs-frontend-api/obs-frontend-apiConfig.cmake`, which apparently isn't shipped in the Flatpak of OBS.

@GeorgesStavracas do you know if something can be done about it in OBS?

barthalion:
@GeorgesStavracas Let me know when it can be merged.

GeorgesStavracas:
I'll review it properly next week

streamfx-force-frontend.patch
Xaymar: This is more of an artifact of two conflicting PRs sent to upstream, one which only added the header file and the other which turned obs-frontend-api into its own module separate from libobs/obs. Unfortunately the former was merged and the latter was rejected.

There was/is also a separate PR that completely redesigns the CMake file for obs-studio, which is what I am waiting for before I redesign the CMake file for StreamFX again. I'm currently failing at finding the link for the CMake rework PR.

streamfx-fix-library-name.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index acaa87c..a0adab6 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -1079,7 +1079,7 @@ if(D_PLATFORM_WINDOWS) # Windows Support
+ endif()
+ 
+ # Minimum Dependencies
+-list(APPEND PROJECT_LIBRARIES libobs)
++list(APPEND PROJECT_LIBRARIES obs)
```
Xaymar, line 10: I'm still fairly sure this link target is incorrect and targets the UI instead of the `libobs` library. StreamFX's build system requires the presence of `LibObsConfig.cmake`, which defines `libobs` not `obs`. The future 27.2.4+ CMakeLists build script also entirely revolves around the existence of libobs, as well as the necessary cmake files.

rmnvgr: I added this patch before setting the `libobs_SOURCE_DIR` option, and didn't try to build without it afterwards. Indeed, it is not necessary, I removed it.

rmnvgr:
bot, build com.obsproject.Studio.Plugin.StreamFX

Xaymar:
For your information, there is a work-in-progress [upstream PR](https://github.com/Xaymar/obs-StreamFX/pull/806) which changes the CMake build system slightly. It will now allow specifying lookup paths for all three dependencies (if the platform needs them), so you can specify `-DOBS_DOWNLOAD=OFF` and then specify the path to libOBS+obs-frontend-api with `-DOBS_PATH=...`. This will apply to all necessary dependencies in the future, such as libaom and svt-av1.

Xaymar:
> Nice! Unfortunately I can't test it now due to the OBS Flatpak not having the `obs-frontend-apiConfig.cmake` file. More worrisome, the Flatpak from OBS `master` branch doesn't appear to ship any CMake file.

See [this](https://github.com/flathub/flathub/pull/3159#discussion_r882223436) comment, it is an artifact from two conflicting PRs. The build system will be enhanced to also check for obs-frontend-api.lib and obs-frontend-api.h if that cmake file is missing.

rmnvgr:
> The build system will be enhanced to also check for obs-frontend-api.lib and obs-frontend-api.h if that cmake file is missing.

That'd be great, as we have the libraries and headers in the Flatpak.

rmnvgr:
@Xaymar Since both OBS and StreamFX will have major changes in their build system, I think it'll be better to wait for their next respective release before publishing the StreamFX Flatpak, to avoid any disruption in the updates while the manifest is adapted. Is that fine with you?

Xaymar:
@rmnvgr v0.12 has no set release date, so it might release before or after OBS Studio 28.0 is out. So I don't mind, since I already release non-flatpak packages anyway.

Xaymar:
It's been a few months, here's the status from my end:

- OBS Studio 28.0 is supported on Windows and Linux, MacOS is still failing with amissing INTERFACE_INCLUDE_DIRECTORIES entry. Due to requiring Qt6 and the changed build system for OBS Studio 28.0, support for OBS Studio 27.2 and below is entirely dropped.
- The Auto-Dependency system was trashed, as it became more and more fragile over time. With recent CMake versions, it would outright fail at times with no known solution for it. Developers are urged to adopt the newer build process instead, which relies on building obs_libraries and obs-frontend-api directly.
- Packages which have a FindModule now always use it, instead of conditionally opting to manually specify link targets. This resolves some platform-dependent link issues entirely.
- Most of the important parts of the build system was either removed with no easy alternative, or is now/will soon handled by scripts.

As I don't plan to make any other massive changes to the build system for now, this should be as stable as it gets until there's another massive change necessary.

barthalion:
@GeorgesStavracas @Xaymar Can you please give me simple yes or no whether it's good to be merged?

rmnvgr:
@barthalion No, I will rework it when OBS 28 releases. I'll put the MR in draft until then.

@Xaymar Thank you for the update!

rmnvgr:
bot, build com.obsproject.Studio.Plugin.StreamFX

rmnvgr:
bot, build com.obsproject.Studio.Plugin.StreamFX

rmnvgr:
Now that OBS 28 is out, I did a test build with the latest StreamFX alpha, it builds without any patches, which is nice. (I built it locally, I don't know why the bot doesn't start the build).

I only needed to force the `REQUIRE_JSON` option, as I had [pointed out](https://github.com/Xaymar/obs-StreamFX/discussions/840#discussioncomment-3288666), since `nlohmann-json` is needed for the frontend (in the about dialog) but the option is only set to `true` when the updater is enabled.

GeorgesStavracas (approved):
The manifest and the metainfo files look good to me. Unless @Xaymar has any objection to landing this, I think it's ready to go!

Xaymar:
LGTM if Flathub has multiple update channels like StreamFX does. Alpha, Beta and Candidate versions shouldn't be used by end-users without explicitly opting into them

GeorgesStavracas:
> LGTM if Flathub has multiple update channels like StreamFX does. Alpha, Beta and Candidate versions shouldn't be used by end-users without explicitly opting into them

There is Flathub Beta, which could hold Beta and RC releases, but not alpha. I'll leave the decision to use Flathub Beta or not to you and @rmnvgr's discretion  :slightly_smiling_face:

rmnvgr:
I'd prefer to avoid maintaining the beta channel and stick to releases. I'll update the manifest and mark the PR as ready when StreamFX 0.12 releases.

flathub.json
```diff
@@ -0,0 +1,4 @@
+{
+  "skip-icons-check": true,
+  "only-arches": ["x86_64"]
```
Xaymar, line 3: Technically, StreamFX supports arm64 if the dependencies are all built for arm64 as well.

hfiguiere: but is there a flatpak of OBS on aarch64? I don't think.

orowith2os: There is not, judging from the [beta.flathub.org page](https://beta.flathub.org/apps/details/com.obsproject.Studio). Maybe upstream could look into building for aarch64 as well?

GeorgesStavracas: aarch64 builds are planned, but there are challenges that prevented it from happening, and it shouldn't block plugins from being accepted into Flathub. Whenever aarch64 builds are published to Flathub by the OBS community, plugin authors will be notified

rmnvgr:
bot, build com.obsproject.Studio.Plugin.StreamFX

Xaymar:
There aren't many bugs left to squash, and many features left to implement:

- HDR/sRGB support for all sources and filters
- Fix Blur's double-free crash
- Find a fix for OBS's variant of FFmpeg, which appears to not understand how pools work and pretend-allocates memory. Mainly affects Windows.

PapyElGringo:
bot, build com.obsproject.Studio.Plugin.StreamFX

rmnvgr:
bot, build com.obsproject.Studio.Plugin.StreamFX

rmnvgr:
bot, build com.obsproject.Studio.Plugin.StreamFX

hanny00:
I tested the new Build 16102 on Flatpak 29, it seems to be working, i see a lot of Filters more than before

PapyElGringo:
I Installed it under `with --user` and `without` but it still missing from OBS is there a step im missing?

rmnvgr:
@PapyElGringo Have you updated OBS to version 29?

PapyElGringo:
Yes I have from Flathub beta repository
```
OBS Studio                                         com.obsproject.Studio                                            29.0.0                        beta                    flathub-beta                system
StreamFX                                           com.obsproject.Studio.Plugin.StreamFX                                                          stable                  streamfx-origin             system
StreamFX                                           com.obsproject.Studio.Plugin.StreamFX                                                          stable                  streamfx-origin             user
```
I'm on `Fedora Silverblue 37`

rmnvgr:
From what I can see it doesn't even try to load the module. I'm not sure if extensions can apply to the beta branch, can you try with OBS stable branch?

PapyElGringo:
Installing from `stable branch` does work :+1:

PapyElGringo:
The build seem to be unavailable can I ask for a rebuild?

mlofjard:
bot, build com.obsproject.Studio.Plugin.StreamFX

mlofjard:
bot, build com.obsproject.Studio.Plugin.StreamFX

Xaymar:
It appears that flathub is not able to clone [this commit from `cmake/clang`](https://github.com/Xaymar/cmake-clang/commit/48516b3c0f5efb96cfc511f58869add8f73f0d32). I'm unsure if I can help resolve this issue, or if this is problem with the flathub buildbot itself.

hfiguiere:
try adding `disable-shallow-clone: true` for the git source.

rmnvgr:
bot, build com.obsproject.Studio.Plugin.StreamFX

rmnvgr:
bot, build com.obsproject.Studio.Plugin.StreamFX

rmnvgr:
@hfiguiere Thanks, that appears to fix it.

kohrias:
Build 25378 does not seem to be available anymore (Can't load uri). Could you start a rebuild?

kohrias:
bot, build com.obsproject.Studio.Plugin.StreamFX

kohrias:
The plugin is loading. Nonetheless there is an error about a missing library in the startup messages: 
```
warning: [StreamFX] <encoder::aom::av1> Loading of '/app/plugins/share/obs/obs-plugins/StreamFX/libaom.so' failed.
warning: [StreamFX] <encoder::aom::av1> Loading of 'libaom' failed.
error: [StreamFX] <encoder::aom::av1> Failed to initialize AOM AV1 encoder: Unable to load AOM library.
```

orowith2os:
The library needed might be here: https://aomedia.googlesource.com/aom/. Though, it might not be too much of a worry, since not many platforms support AV1 yet.

Xaymar:
AOM encoder in StreamFX is deprecated and scheduled to be removed in a future revision. Even if you provide the library, it will just stay hidden.

kohrias:
> AOM encoder in StreamFX is deprecated and scheduled to be removed in a future revision. Even if you provide the library, it will just stay hidden.

Ok, I am going to ignore the message then. Thanks for letting me know.

blusewill:
bot, build com.obsproject.Studio.Plugin.StreamFX

Xaymar:
Just a heads-up, in case things break, the reason is that I had to remove a GPLv2 "or later" incompatible submission to the plugin retroactively. I've mostly restored the content of the repository to the best of my ability, but still have to do a lot of work to fix up the remaining branches, and then rebuild binaries without the unwanted code.

penyuan:
Sorry I might be missing something, but when I tried to install this via @flathubbot's [comment above](https://github.com/flathub/flathub/pull/3159#issuecomment-1475180541), I get a 404 error.

Is there a Flatpak of the StreamFX plugin that I can install and test now @Xaymar?

hfiguiere:
the test build expired, and the current PR must be updated as the upstream source repository doesn't carry the revisions that is expected.

penyuan:
Thanks @hfiguiere, sounds like we need OP @rmnvgr to update this pull request?

Xaymar:
> Is there a Flatpak of the StreamFX plugin that I can install and test now @Xaymar?

At the current time, I do not maintain a flatpak package however in the future official packages for various Linux distributions may be considered. However due to the CEF and Wayland incompatibility that causes OBS Studio to crash, it is also likely that Linux support is dropped entirely unless someone has a known way to fix it.

Sid127:
Is this still being worked on?

kir68k:
```patch
diff --git a/com.obsproject.Studio.Plugin.StreamFX.yml b/com.obsproject.Studio.Plugin.StreamFX.yml
index 269ad8f..c040e58 100644
--- a/com.obsproject.Studio.Plugin.StreamFX.yml
+++ b/com.obsproject.Studio.Plugin.StreamFX.yml
@@ -21,8 +21,8 @@ modules:
     sources:
       - type: git
         url: https://github.com/Xaymar/obs-StreamFX.git
-        tag: 0.12.0a173
-        commit: fcb441bdf6e5c70d5eb5ebf6795f2ddd43f70a23
+        tag: 0.12.0b299
+        commit: 8b97c2b23ddd931eee538e6feec71d76d42e494d
         disable-shallow-clone: true
         x-checker-data:
           type: json
```
After changing this I was able to run `flatpak-builder` on the YAML file, install and run OBS Studio with StreamFX.

Xaymar:
I'm a bit late on the updates here, but here's the current status:

1. I've split StreamFX into a free variant and a paid variant in order to actually have any funds for development. The free variant, which this is based on, is no longer maintained by anyone and relies on pull request submissions.
2. I've registered a company ("Vhonowslend") and transferred StreamFX and a few other projects, as well as all rights and technologies to it. Thanks to the funding, we're currently at 2 part-time employees, and one temporary hire.
3. We're working on a Flatpak bundle for the paid variant. As Flathub lacks authorization means, we're not going to publish this bundle on Flathub. Users will instead have to install it using the normal way.

FWIW, I'm not going to reverse any of these decisions. While it hurts to let go of a pet project, the pet project has an unhealthy ratio of users to maintainers, so going paid was the only reasonable option.

ItsRogueRen:
Totally understand going paid with the amount of work you've put in. I think the transition was definitely rough and could have been done better, but I don't disagree with the end goal. 

That said, will the plugin still be open source, just with pre-compiled binaries being paid? Or will the free version eventually get what paid version does, but after a delay of time?

Xaymar:
> I think the transition was definitely rough and could have been done better, but I don't disagree with the end goal.

Maybe, but you'd likely have done the same in my place. Attempted Murder, Assault, Doxxing, Bullying/Harassment, Mailbombing, Impersonation, name it and it likely still has an open lawsuit against a person which may or may not be a content creator. I had to move to a new address for the orders in my name to stop. Sure, it wasn't perfect, but I'll take good enough. The poorly timed DMCA request arriving just a week prior to the switch wasn't great, but eh - I added a massive text banner to the releases, can't do much more about it.

> That said, will the plugin still be open source, just with pre-compiled binaries being paid? Or will the free version eventually get what paid version does, but after a delay of time?

We'll be following the spirit and letter of the GPL license, but beyond that there aren't any plans for the free version. It's unlikely that we'll ever be updating the free version again, as I've seen hell, and I don't want to go back to it.

PR #3170, opened 2022-05-20
Add RigelEngine

lethal-guitar:

Notes:

I'm using `device=all`, since the application can make use of game controllers and I'd like to allow this. As far as I could tell, there is no finer-grained option to provide access to only the graphics card and game controllers, but if there is, I'd be happy to restrict this permission more.

This application needs read-only access to a directory with game data, which the user needs to provide themselves.
I decided to give read-only access to the home directory, as that seems like the most user-friendly option. The alternative would be to require the user to place the required data into a specific location, or into an XDG directory. But that seems less convenient to me, and since the application needs only read-only access, providing full home directory access seems ok to me.

### DIFF
--- a/io.github.lethal_guitar.RigelEngine.yaml
+++ b/io.github.lethal_guitar.RigelEngine.yaml
@@ -12,7 +12,6 @@ rename-icon: rigelengine
 
 finish-args:
   - --device=all # to access game controllers
-  - --filesystem=~/rigelengine-game-data:ro
   - --share=ipc
   - --socket=wayland
   - --socket=fallback-x11
@@ -22,7 +21,7 @@ modules:
   - name: rigelengine
     buildsystem: cmake-ninja
     config-opts:
-      - -DCMAKE_BUILD_TYPE=Release
+      - -DCMAKE_BUILD_TYPE=RelWithDebInfo
     build-commands:
       - install -D -t "${FLATPAK_DEST}/bin/" ./run.sh
     sources:
--- a/io.github.lethal_guitar.RigelEngine.yaml
+++ b/io.github.lethal_guitar.RigelEngine.yaml
@@ -1,0 +1,30 @@
+app-id: io.github.lethal_guitar.RigelEngine
+
+runtime: org.freedesktop.Platform
+sdk: org.freedesktop.Sdk
+runtime-version: "21.08"
+
+command: RigelEngine
+
+rename-appdata-file: rigelengine.metainfo.xml
+rename-desktop-file: rigelengine.desktop
+rename-icon: rigelengine
+
+finish-args:
+  - --device=all # to access game controllers
+  - --filesystem=home:ro # to let the user place game data in a desired location
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+
+modules:
+  - name: rigelengine
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+    sources:
+      - type: git
+        url: https://github.com/lethal-guitar/RigelEngine.git
+        tag: v0.8.4-flatpak
+        commit: ce389030cdcc6647eef75a97da46c50c1e61c600
--- a/io.github.lethal_guitar.RigelEngine.yaml
+++ b/io.github.lethal_guitar.RigelEngine.yaml
@@ -4,7 +4,7 @@ runtime: org.freedesktop.Platform
 sdk: org.freedesktop.Sdk
 runtime-version: "21.08"
 
-command: RigelEngine
+command: run.sh
 
 rename-appdata-file: rigelengine.metainfo.xml
 rename-desktop-file: rigelengine.desktop
@@ -12,7 +12,7 @@ rename-icon: rigelengine
 
 finish-args:
   - --device=all # to access game controllers
-  - --filesystem=home:ro # to let the user place game data in a desired location
+  - --filesystem=~/rigelengine-game-data:ro
   - --share=ipc
   - --socket=wayland
   - --socket=fallback-x11
@@ -23,8 +23,14 @@ modules:
     buildsystem: cmake-ninja
     config-opts:
       - -DCMAKE_BUILD_TYPE=Release
+    build-commands:
+      - install -D -t "${FLATPAK_DEST}/bin/" ./run.sh
     sources:
       - type: git
         url: https://github.com/lethal-guitar/RigelEngine.git
         tag: v0.8.4-flatpak
         commit: ce389030cdcc6647eef75a97da46c50c1e61c600
+      - type: file
+        path: run.sh
+      - type: patch
+        path: fixed-data-location-for-flatpak.patch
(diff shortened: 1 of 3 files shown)


lethal-guitar:
bot, build

lethal-guitar:
bot, build io.github.lethal_guitar.RigelEngine

io.github.lethal_guitar.RigelEngine.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+command: RigelEngine
+
+rename-appdata-file: rigelengine.metainfo.xml
+rename-desktop-file: rigelengine.desktop
+rename-icon: rigelengine
+
+finish-args:
+  - --device=all # to access game controllers
+  - --filesystem=home:ro # to let the user place game data in a desired location
```
hfiguiere, line 15: it would be better to restrict to a location.

io.github.lethal_guitar.RigelEngine.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - --share=ipc
+  - --socket=wayland
```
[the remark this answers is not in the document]
lethal-guitar: It's using SDL which does support Wayland, and I don't make any use of X11 APIs in my own code, so it should work. I haven't tested it  though, so I can also drop this if you prefer.

io.github.lethal_guitar.RigelEngine.yaml
lethal-guitar: I'll look into it

io.github.lethal_guitar.RigelEngine.yaml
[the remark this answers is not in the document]
lethal-guitar: Ok, I've just tested it - it does run fine on Wayland.

lethal-guitar:
@hfiguiere ok, I've changed to a fixed location within the user's home dir. This required adding some patches, which don't make sense to upstream since they are specific to the Flatpak version.

Let me know what you think! And thanks for the quick review, btw.

lethal-guitar:
bot, build io.github.lethal_guitar.RigelEngine

lethal-guitar:
bot, build io.github.lethal_guitar.RigelEngine

lethal-guitar:
bot, build io.github.lethal_guitar.RigelEngine

lethal-guitar:
bot, build io.github.lethal_guitar.RigelEngine

lethal-guitar:
bot, build io.github.lethal_guitar.RigelEngine

io.github.lethal_guitar.RigelEngine.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+command: run.sh
+
+rename-appdata-file: rigelengine.metainfo.xml
+rename-desktop-file: rigelengine.desktop
+rename-icon: rigelengine
+
+finish-args:
+  - --device=all # to access game controllers
+  - --filesystem=~/rigelengine-game-data:ro
```
barthalion, line 15: I suggest using one of the locations your app has write access by default, like $XDG_DATA_HOME.

lethal-guitar: Hi @barthalion, thanks for your review.
I don't need write access, I just need a read-only location that's convenient for the user to place some files into. But they only need to do that once.

I've had a look at other apps similar to mine, i.e. game engines which need user-provided game data, and found the following approaches:

* [io.github.fabiangreffrath.Doom](https://github.com/flathub/io.github.fabiangreffrath.Doom/blob/fa194e02405fb4b1d59b9ceea3d0423b0fa86995/io.github.fabiangreffrath.Doom.json#L14) uses `home:ro`. This is how I originally did it, but changed to the current approach based on https://github.com/flathub/flathub/pull/3170#discussion_r878125921.
* [net.dengine.Doomsday](https://github.com/flathub/net.dengine.Doomsday/blob/0a63af4d33e44caa258d525dc798e994fc5f505e/net.dengine.Doomsday.json#L13) uses `home`, which seems too broad as I don't need write access.
* [org.openttd.OpenTTD](https://github.com/flathub/org.openttd.OpenTTD/blob/68647fe5bd8d8e43cb3e54dac8bea9d73299e6c6/org.openttd.OpenTTD.yaml#L10) also uses `home`.

So it actually seems a common approach for apps similar to mine to map the entire home directory. My current solution is more restricted, and thus should be preferable if I understand correctly. The current solution also seems more user-friendly to me compared to using `XDG_DATA_HOME`, since the latter is inside a hidden folder whereas my approach uses a regular folder. This makes a difference for me because I want my project to be easy to use for less technical users as well. Most graphical file managers do not show dot-files by default, so using `XDG_DATA_HOME` adds an extra step for the user.

Does that make sense?

lethal-guitar: Looking at some more examples, I see that [EDuke32](https://github.com/flathub/com.eduke32.EDuke32/blob/0a257174e16be13913375647e40b24aea5609ee7/com.eduke32.EDuke32.yaml#L13) uses a subdirectory of `XDG_DOCUMENTS`. I would also be fine with that approach, since the documents folder is not a hidden folder.

barthalion: Yeah, I can live with xdg-documents. I'm just not a fan of top-level directories in $HOME.

lethal-guitar:
bot, build io.github.lethal_guitar.RigelEngine

lethal-guitar:
@barthalion after spending some time trying to make a subdirectory of `xdg-documents` work, I've switched it to `XDG_DATA_HOME` now. Please let me know what you think of the latest state.

My issue was that I couldn't find a way to display the concrete expected path to the user from within `run.sh`, since using `--filesystem=xdg-documents/bla` maps `bla` into `/home/username/bla` inside the container environment. So trying to show a meaningful error message to the user wasn't possible. I find this unfortunate, since I was really hoping that my users wouldn't need to mess with hidden directories. But in the interest of getting this merged, I'm ok with the current solution for now.

Is there any way to get the host-side version of the xdg-documents path from within a flatpak app?

barthalion:
I need to admit that I actually don't know how to query the path to specific XDG directories after landing `--filesystem`. I will ask on our Matrix channel and get back to you.

/merge

barthalion:
You can read it from `$XDG_CONFIG_HOME/user-dirs.dirs`. E.g.:

```
[bpiotrowski@risuku ~]$ flatpak run --command=bash --filesystem=xdg-documents io.github.lethal_guitar.RigelEngine 
[📦 io.github.lethal_guitar.RigelEngine ~]$ cat $XDG_CONFIG_HOME/user-dirs.dirs 
XDG_DOCUMENTS_DIR="/var/home/bpiotrowski/docs"
```

lethal-guitar:
> You can read it from `$XDG_CONFIG_HOME/user-dirs.dirs`. E.g.:
> 
> ```
> [bpiotrowski@risuku ~]$ flatpak run --command=bash --filesystem=xdg-documents io.github.lethal_guitar.RigelEngine 
> [📦 io.github.lethal_guitar.RigelEngine ~]$ cat $XDG_CONFIG_HOME/user-dirs.dirs 
> XDG_DOCUMENTS_DIR="/var/home/bpiotrowski/docs"
> ```

Awesome, thanks for checking! 👍🏻

PR #3175, opened 2022-05-20
Add org.kde.kaffeine

Mailaender:

- [X] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://invent.kde.org/multimedia/kaffeine/-/merge_requests/6
- [X] Any additional patches or files have been submitted to the upstream projects concerned. *(If not, explain why.)* https://invent.kde.org/multimedia/kaffeine/-/merge_requests/6 https://invent.kde.org/multimedia/kaffeine/-/merge_requests/7 https://invent.kde.org/multimedia/kaffeine/-/merge_requests/8

### DIFF
--- a/org.kde.kaffeine.json
+++ b/org.kde.kaffeine.json
@@ -1,0 +1,267 @@
+{
+  "app-id": "org.kde.kaffeine",
+  "runtime": "org.kde.Platform",
+  "runtime-version": "5.15-21.08",
+  "sdk": "org.kde.Sdk",
+  "add-extensions": {
+    "org.freedesktop.Platform.ffmpeg-full": {
+        "directory": "lib/ffmpeg",
+        "version": "21.08",
+        "add-ld-path": ".",
+        "no-autodownload": true,
+        "autodelete": false
+    }
+  },
+  "command": "kaffeine",
+  "rename-icon": "kaffeine",
+  "finish-args": [
+    "--share=ipc",
+    "--share=network",
+    "--socket=x11",
+    "--socket=pulseaudio",
+    "--filesystem=home",
+    "--filesystem=/run/media",
+    "--filesystem=/media",
+    "--device=dri",
+    "--talk-name=com.canonical.AppMenu.Registrar",
+    "--talk-name=org.kde.kconfig.notify",
+    "--talk-name=org.kde.KGlobalSettings",
+    "--talk-name=org.kde.StatusNotifierWatcher",
+    "--talk-name=org.mpris.MediaPlayer2.Player",
+    "--own-name=org.mpris.MediaPlayer2.org.kde.kaffeine",
+    "--system-talk-name=org.freedesktop.UDisks2",
+    "--system-talk-name=org.freedesktop.UPower"
+  ],
+  "modules": [
+    {
+      "name": "ffmpeg",
+      "buildsystem" : "simple",
+      "build-commands": [
+        "install -dm755 ${FLATPAK_DEST}/lib/ffmpeg"
+      ]
+    },
+    {
+      "name": "libdvdread",
+      "rm-configure": true,
+      "sources": [
+        {
+          "type": "archive",
+          "url": "https://download.videolan.org/videolan/libdvdread/6.1.2/libdvdread-6.1.2.tar.bz2",
+          "sha256": "cc190f553758ced7571859e301f802cb4821f164d02bfacfd320c14a4e0da763"
+        },
+        {
+          "type": "script",
+          "commands": [
+            "autoreconf -fiv"
+          ],
+          "dest-filename": "autogen.sh"
+        }
+      ]
+    },
+    {
+      "name": "libdvdnav",
+      "rm-configure": true,
+      "sources": [
+        {
+          "type": "archive",
+          "url": "https://download.videolan.org/videolan/libdvdnav/6.1.1/libdvdnav-6.1.1.tar.bz2",
+          "sha256": "c191a7475947d323ff7680cf92c0fb1be8237701885f37656c64d04e98d18d48"
+        },
+        {
+          "type": "script",
+          "commands": [
+            "autoreconf -fiv"
+          ],
+          "dest-filename": "autogen.sh"
+        }
+      ]
+    },
+    {
+      "name": "libdvdcss",
+      "rm-configure": true,
+      "sources": [
+        {
+          "type": "archive",
+          "url": "https://download.videolan.org/videolan/libdvdcss/1.4.3/libdvdcss-1.4.3.tar.bz2",
+          "sha256": "233cc92f5dc01c5d3a96f5b3582be7d5cee5a35a52d3a08158745d3d86070079"
+        },
+        {
+          "type": "script",
+          "commands": [
+            "autoreconf -fiv"
+          ],
+          "dest-filename": "autogen.sh"
+        }
+      ]
+    },
+    {
+      "name": "libbluray",
+      "config-opts": [
+        "--disable-bdjava-jar"
+      ],
+      "cleanup": [
+        "/include",
+        "*.a",
+        "*.la",
+        "*.pc",
+        "/lib/*.so"
+      ],
+      "sources": [
+        {
+          "type": "archive",
+          "url": "https://download.videolan.org/pub/videolan/libbluray/1.3.0/libbluray-1.3.0.tar.bz2",
+          "sha256": "e2dbaf99e84e0a9725f4985bcb85d41e52c2261cc651d8884b1b790b5ef016f9"
+        }
+      ]
+    },
+    {
+      "name": "libdvbpsi",
+      "rm-configure": true,
+      "sources": [
+        {
+          "type": "archive",
+          "url": "https://download.videolan.org/pub/libdvbpsi/1.3.3/libdvbpsi-1.3.3.tar.bz2",
+          "sha256": "02b5998bcf289cdfbd8757bedd5987e681309b0a25b3ffe6cebae599f7a00112"
+        },
+        {
+          "type": "script",
+          "commands": [
+            "autoreconf -fiv"
+          ],
+          "dest-filename": "autogen.sh"
+        }
+      ]
+    },
+    {
+      "name": "libmpeg2",
+      "config-opts": [
+        "--enable-shared",
+        "--disable-static"
+      ],
+      "rm-configure": true,
+      "cleanup": [
+        "/bin/*mpeg2*"
+      ],
+      "sources": [
+        {
+          "type": "archive",
+          "url": "http://libmpeg2.sourceforge.net/files/libmpeg2-0.5.1.tar.gz",
+          "sha256": "dee22e893cb5fc2b2b6ebd60b88478ab8556cb3b93f9a0d7ce8f3b61851871d4"
+        },
+        {
+          "type": "script",
+          "commands": [
+            "autoreconf -fiv"
+          ],
+          "dest-filename": "autogen.sh"
+        }
+      ]
+    },
+    {
+      "name": "libmodplug",
+      "cleanup": [
+        "/include",
+        "/lib/pkgconfig",
+        "/lib/*.la"
+      ],
+      "sources": [
+        {
+          "type": "archive",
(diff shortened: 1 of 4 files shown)


Mailaender:
bot, build org.kde.kaffeine

hfiguiere:
cc @flathub/kde

Mailaender:
bot, build org.kde.kaffeine

Mailaender:
bot, build org.kde.kaffeine

travier:
Manifest seems good to me. Note that you can also add x-data-checker config for dependencies too, not just the application source. But we can also add it after the merge.

barthalion:
It doesn't seem to play videos from local filesystem for me:

```
23-05-22 08:46:45.732 [Warning ] qt.qpa.wayland: Wayland does not support QWindow::requestActivate()
[00007f90a80013b0] gl vout display error: parent window not available
[00007f90a80013b0] xcb vout display error: window not available
[00007f90a80013b0] xcb vout display error: window not available
[00007f90a80013b0] fb vout display error: cannot get terminal mode (Inappropriate ioctl for device)
[00007f90b0007610] main video output error: video output creation failed
[00007f90ba557370] main decoder error: failed to create video output
[libopenh264 @ 0x7f90ba557b40] get_buffer() failed
[libopenh264 @ 0x7f90ba557b40] Unable to allocate buffer
[00007f90ba557370] avcodec decoder error: avcodec_send_packet critical error
```

Can someone else check as well?

org.kde.kaffeine.json
```diff
@@ -0,0 +1,238 @@
...
+  "command": "kaffeine",
+  "finish-args": [
+    "--share=ipc",
+    "--share=network",
+    "--socket=wayland",
+    "--socket=fallback-x11",
+    "--socket=pulseaudio",
+    "--filesystem=home",
+    "--filesystem=/run/media",
+    "--filesystem=/media"
```
travier, line 15: Missing `--device=dri`

travier:
Plays the sound but not the image for videos for me. Maybe because of the missing `dri` perm.

We also have a lot of:
```
23-05-22 08:58:55.888 [Warning ] kf.solid.backends.udisks2: /run/build-runtime/solid/src/solid/devices/backends/udisks2/udisksmanager.cpp#160: void Solid::Backends::UDisks2::Manager::introspect(const QString&, bool): Failed enumerating UDisks2 objects: "org.freedesktop.DBus.Error.Disconnected" 
 "Not connected to D-Bus server"
```

Maybe need some DBus access here.

org.kde.kaffeine.json
Mailaender: Added

Mailaender:
Video playback worked fine for me even without `dri`. Maybe this doesn't play well with Wayland.

nedrichards:
Yup, there's an array of dbus calls getting filtered due to no allowlist if you run with --log-session-bus (mpris etc.)

Mailaender:
I added some, but probably need help with this.

travier:
bot, build org.kde.kaffeine

travier:
Still can not play videos for me:

```
[libopenh264 @ 0x7fdd6f664a40] get_buffer() failed
[libopenh264 @ 0x7fdd6f664a40] Unable to allocate buffer
```

tinywrkb:
> [libopenh264 @ 0x7fdd6f664a40] get_buffer() failed

This app is linked against FFmpeg? Then it can use the full FFmpeg extension, and avoid the OpenH264 decoder, if that's the source of the issue.

```
add-extensions:
  org.freedesktop.Platform.ffmpeg-full:
    directory: lib/ffmpeg
    version: '21.08'
    add-ld-path: .
    no-autodownload: false
...
modules:
- name: ...
   build-commands/post-install:
      - install -dm755 ${FLATPAK_DEST}/lib/ffmpeg

```

org.kde.kaffeine.json
```diff
@@ -0,0 +1,244 @@
...
+    {
+      "config-opts": [
+        "BUILDCC=gcc",
+        "--disable-lua",
+        "--disable-a52",
+        "--disable-qt",
+        "--disable-ncurses",
+        "--disable-udev"
+      ],
+      "name": "libvlc",
```
hfiguiere, line 213: usually name is the first property

Mailaender:
bot, build org.kde.kaffeine

travier:
Still does not work for me 😕

tinywrkb:
For some reason, it doesn't work in Wayland, using `--socket=x11` removing the Wayland socket works. Maybe also unset `QT_QPA_PLATFORM` for users who have this set in the environment.

You might also need to patch VLC, as the plugins cache test fails. The following works for me, but you should look at what the VLC Flatpak has.

```
diff --git a/src/modules/bank.c b/src/modules/bank.c
index 2e67a0d..1e106e9 100644
--- a/src/modules/bank.c
+++ b/src/modules/bank.c
@@ -273,16 +273,6 @@ static int AllocatePluginFile (module_bank_t *bank, const char *abspath,
     if (bank->mode & CACHE_READ_FILE)
     {
         plugin = vlc_cache_lookup(&bank->cache, relpath);
-
-        if (plugin != NULL
-         && (plugin->mtime != (int64_t)st->st_mtime
-          || plugin->size != (uint64_t)st->st_size))
-        {
-            msg_Err(bank->obj, "stale plugins cache: modified %s",
-                    plugin->abspath);
-            vlc_plugin_destroy(plugin);
-            plugin = NULL;
-        }
     }
 
     if (plugin == NULL)
```

Mailaender:
bot, build org.kde.kaffeine

tinywrkb:
I looked at what the VLC Flatpak maintainer did. He kept the plugin size test (timestamp test still removed of course), but regenerated the plugins cache in the cleanup-commands step, which I'm guessing it's because cleanup runs after symbols extraction, so a plugin's size won't change anymore.  
I've no idea why the devs decided to cache timestamps and sizes, just version the plugins if you're worried about loading incorrect binaries, it's not like these values are tested if `VLC_PLUGIN_PATH` is set.

Mailaender:
bot, build org.kde.kaffeine

AsciiWolf:
bot, build org.kde.kaffeine

AsciiWolf:
Tested and seems to work fine. :-)

The only problem that I noticed is desktop icon that is missing for some reason.

travier:
I can play videos with the latest build! 🎉

travier:
OK, the manifest needs a quick check/update to the latest releases and some x-checker-data config addition for dependencies but we should be good.

AsciiWolf:
What about the missing desktop icon? :-)

travier:
Sorry, I've not checked that.

Mailaender:
I checked

> /app/share/icons/hicolor/scalable/apps/kaffeine.svg

is definitely installed

AsciiWolf:
Then there is most likely something wrong with the AppData file. I took a quick look at the file and did not find any obvious issues. I think that there should not be a .desktop suffix in the [app id](https://github.com/KDE/kaffeine/blob/master/src/org.kde.kaffeine.appdata.xml#L3), but it should not do any harm (it is just not needed). Everything else looks good.

org.kde.kaffeine.json
```diff
@@ -0,0 +1,263 @@
...
+  "add-extensions": {
+    "org.freedesktop.Platform.ffmpeg-full": {
+        "directory": "lib/ffmpeg",
+        "version": "21.08",
+        "add-ld-path": ".",
+        "no-autodownload": true,
+        "autodelete": false
+    }
+  },
+  "command": "kaffeine",
```
AsciiWolf, line 15:
```suggestion
  "command": "kaffeine",
  "rename-icon": "kaffeine",
```

This could help, but I am not sure.

hfiguiere: I think it would.
Toward the end of the build I get:
```
WARNING: Icon not matching app id in /var/home/hub/source/flathub/org.kde.kaffeine/repo/export/share/applications/org.kde.kaffeine.desktop: kaffeine
WARNING: Icon referenced in desktop file but not exported: kaffeine
```

hfiguiere: The installed destkop file has `Icon=kaffeine` which `rename-icon` will fix.

hfiguiere:
> Then there is most likely something wrong with the AppData file. I took a quick look at the file and did not find any obvious issues. I think that there should not be a .desktop suffix in the [app id](https://github.com/KDE/kaffeine/blob/master/src/org.kde.kaffeine.appdata.xml#L3), but it should not do any harm (it is just not needed). Everything else looks good.

While it's better to not have it as it is deprecated we actually tolerate it.

hfiguiere:
```
21-08-22 15:34:02.955 [Warning ] kf.solid.backends.udisks2: /run/build-runtime/solid/src/solid/devices/backends/udisks2/udisksmanager.cpp#160: void Solid::Backends::UDisks2::Manager::introspect(const QString&, bool): Failed enumerating UDisks2 objects: "org.freedesktop.DBus.Error.Disconnected" 
 "Not connected to D-Bus server"
```
it probably want udisk dbus

Mailaender:
bot, build org.kde.kaffeine

hfiguiere:
adding  `"--system-talk-name=org.freedesktop.UDisks2"`, fixes the error

Mailaender:
I know. Figured it out yesterday in https://github.com/flathub/org.kde.kompare/pull/9.

hfiguiere:
it also need     "--system-talk-name=org.freedesktop.UPower"

Mailaender:
bot, build org.kde.kaffeine

AsciiWolf:
I can confirm that there is now a correct desktop icon displayed. :-)

travier:
The AppData patch lists `22.04.1` but the sources are at `2.0.18`. Commented there: https://invent.kde.org/multimedia/kaffeine/-/merge_requests/6 too.

travier:
Should be good to go apart from that.

Mailaender:
https://invent.kde.org/multimedia/kaffeine/-/merge_requests/7

Mailaender:
bot, build org.kde.kaffeine

travier (approved):

Mailaender:
Whoever had this brilliant idea that XML had to be manually sorted.

tsdgeos:
22-08-22 21:24:20.340 [Warning ] kf.notifications: /run/build-runtime/knotifications/src/kstatusnotifieritem.cpp#998: void KStatusNotifierItemPrivate::setLegacySystemTrayEnabled(bool): env says KDE is running but SNI unavailable -- check KDE_FULL_SESSION and XDG_CURRENT_DESKTOP

Probably want --talk-name=org.kde.StatusNotifierWatcher ?

Mailaender:
bot, build org.kde.kaffeine

tsdgeos:
Have you checked if the digital tv stuff works? I wonder if there's special hardware access needed for that

AsciiWolf:
> Have you checked if the digital tv stuff works? I wonder if there's special hardware access needed for that

I am also interested in this (but don't have any hardware that I could test with at the moment), but I am afraid that it will require custom udev rules on the host.

tsdgeos:
I'm not 100% sure it makes a lot of sense to commit this if we can't verify the features work well, maybe merge to beta branch?

AsciiWolf:
I don't personally consider the digital tv functionality as primary function of Kaffeine. Everything else seems to work fine. Also, every dvb-t stick that has its udev rules upstreamed in udevd (systemd) should theoretically work fine out-of-box.

hfiguiere:
if we blocked merge any app that has stuff not working properly, then we'd have nothing

travier:
I think we're in a state where the app is completely fine for beta.
I'd also argue that it has enough features working for stable.
Given that it's not even part of the release service, I'm not sure how much work will happen on this one so I'd be inclined to move it to stable and see if folks complain.

tsdgeos:
> if we blocked merge any app that has stuff not working properly, then we'd have nothing

Sad though, giving people things that work worse than they could.

I won't block this going to regular release.

travier:
Let's give it try. @nedrichards @barthalion Can you merge this one? Thanks!

PR #3182, opened 2022-05-21
Add Kleopatra

Mailaender:

- [X] Any additional patches or files have been submitted to the upstream projects concerned. *(If not, explain why.)* https://invent.kde.org/pim/kleopatra/-/merge_requests/31


Mailaender:
bot, build org.kde.kleopatra

Mailaender:
bot, build org.kde.kleopatra

hfiguiere:
cc @flathub/kde

org.kde.kleopatra.json
```diff
@@ -0,0 +1,91 @@
...
+            "type": "anitya",
+            "project-id": 6845,
+            "stable-only": true,
+            "url-template": "https://boostorg.jfrog.io/artifactory/main/release/$version/source/boost_${major}_${minor}_$patch.tar.bz2"
+          }
+        }
+      ]
+    },
+    {
+      "name": "gpgme",
```
hfiguiere, line 38: this is in the SDK runtime:
```
$ pkg-config --modversion gpgme
1.16.0
```

hfiguiere (changes requested):

org.kde.kleopatra.json
Mailaender: I need `QGpgme` which you omitted.

travier:
I'll review it later too but I'm afraid this one won't be that useful as a Flatpak. See https://hackmd.io/NGP9BybWSre-juT6tI4c2A

org.kde.kleopatra.json
```diff
@@ -0,0 +1,91 @@
...
+  "runtime-version": "5.15-21.08",
+  "sdk": "org.kde.Sdk",
+  "command": "kleopatra",
+  "rename-icon": "kleopatra",
+  "finish-args": [
+    "--share=ipc",
+    "--share=network",
+    "--socket=wayland",
+    "--socket=fallback-x11",
+    "--filesystem=~/.gnupg:create"
```
travier, line 13: I'm not sure this is safe. What happens when the keyring gets accessed from outside and inside a Flatpak at the same time?

travier:
See also https://github.com/flatpak/flatpak/issues/2301

org.kde.kleopatra.json
travier: Interesting. Maybe this will work.

org.kde.kleopatra.json
```diff
@@ -0,0 +1,91 @@
...
+          }
+        }
+      ]
+    },
+    {
+      "name": "QGpgme",
+      "sources": [
+        {
+          "type": "archive",
+          "url": "https://gnupg.org/ftp/gcrypt/gpgme/gpgme-1.16.0.tar.bz2",
```
tsdgeos, line 42: Use 1.17.1?

org.kde.kleopatra.json
```diff
@@ -0,0 +1,91 @@
...
+    },
+    {
+      "name": "kleopatra",
+      "buildsystem": "cmake-ninja",
+      "sources": [
+        {
+          "type": "archive",
+          "url": "https://download.kde.org/stable/release-service/22.04.1/src/kleopatra-22.04.1.tar.xz",
+          "sha256": "d30e6335b78c6fce6ad22e5482032d0f8a02740b46fe9fdc3263d39ceaeaae23",
+          "x-checker-data": {
```
tsdgeos, line 77: Please use the same x-checker-data for the other 2 KDE dependencies ↑↑

Mailaender:
bot, build org.kde.kleopatra

tsdgeos:
This also needs the status notifier thing so you can close it with the [x] and it can go to the system tray area in Plasma

Mailaender:
bot, build org.kde.kleopatra

org.kde.kleopatra.json
```diff
@@ -0,0 +1,111 @@
...
+            "type": "anitya",
+            "project-id": 8763,
+            "stable-only": true,
+            "url-template": "https://download.kde.org/stable/kleopatra/$version/kleopatra-$version.tar.xz"
```
tsdgeos, line 101: url is wrong

org.kde.kleopatra.json
```diff
@@ -0,0 +1,111 @@
...
+      "sources": [
+        {
+          "type": "archive",
+          "url": "https://download.kde.org/stable/release-service/22.04.1/src/kmime-22.04.1.tar.xz",
+          "sha256": "a7f94a60d204f7546a711551feafb05549caf25d5831ecb9942cad32caddca05",
+          "x-checker-data": {
+            "type": "anitya",
+            "project-id": 8763,
+            "stable-only": true,
+            "url-template": "https://download.kde.org/stable/release-service/$version/kmime-$version.tar.xz"
```
tsdgeos, line 84: missing src in url

org.kde.kleopatra.json
```diff
@@ -0,0 +1,111 @@
...
+      "sources": [
+        {
+          "type": "archive",
+          "url": "https://download.kde.org/stable/release-service/22.04.1/src/libkleo-22.04.1.tar.xz",
+          "sha256": "96d6058bab59c4b807548443d20dfc950f0a916c02aa84a96bf46ea2ecfe1807",
+          "x-checker-data": {
+            "type": "anitya",
+            "project-id": 8763,
+            "stable-only": true,
+            "url-template": "https://download.kde.org/stable/release-service/$version/libkleo-$version.tar.xz"
```
tsdgeos, line 67: missing src in url

tsdgeos:
libkleo says

-- The following OPTIONAL packages have not been found:

 * KF5PimTextEdit (required version >= 5.20.1), A textedit with PIM-specific features., <https://commits.kde.org/kpimtextedit>
   Improved audit log viewer.


maybe you can add it?

Mailaender:
Added it.

bot, build org.kde.kleopatra

barthalion:
All good here @flathub/kde?

travier:
This one needs testing for the GPG support. I haven't got to it yet.

tsdgeos:
It can see my gpg keys, but when i tried signing it failed, it complained pinentry is missing

ImBearChild:
> It can see my gpg keys, but when i tried signing it failed, it complained pinentry is missing

I [tried](https://invent.kde.org/bearchild/flatpak-kde-applications) to fix Kleopatra in flatpak repo of KDE, and I met the same situation. So I've done some research for this:

Pinentry  is basically a part of GnuPG, as a safe way to let user input their passphase safely. Kleopatra talks to `gpg-agent`, and `gpg-agent` will use pinentry for every place that needs a passphase. The problem here is we don't have `pinentry` in the Runtime, and build and bundle it with the application won't solve this. Because `gpg-agent` don't find `pinentry`executable in `$PATH`, instead, it use a hard-coded path `/usr/bin/pinentry`(only runtime can add this, but Freedesktop SDK and KDE sdk both not have this thing), but this behaviour can be changed by put `pinentry-program /app/bin/pinentry` in `~/.gnupg/gpg-agent.conf` or a commdline flag.

So there's three way to deal with this issue:

1. Suggest developers to build a KDE runtime with `pinentry` and `pinentry-qt`.
    Kleopatra is not the only application that needs `pinentry`, Kontact and KMail will use it too, and both of them will complain `pinentry is missing` in the same case. So making `pinentry` avaliabe in KDE runtime will benefit all application using `gpg-agent`
2. Modify the conf
   As I wrote above, add a line in configuration file will solve this. But this will break `gpg-agent`on the host, unless we isolate Kleopatra from `~./gunpg ` of host, and this will break some aspect of function. 
3. Modify the code of GnuPG
  Let gpg find `pinentry` in `$PATH`. Since the developer use a nearly hard-coded path rather than search it in system, there may be some strong reasons for them to do this. So this way may be harmful.

Personally, I prefer the first way. Because It's won't break anything. If this looks good to you, I will open a issue at the [repo](https://invent.kde.org/packaging/flatpak-kde-runtime) of KDE runtime.

travier:
Interesting. I'm a little bit uneasy about having multiple versions of gnupg from different flatpaks access the same gnupg config folder & keys at the same time as the format changed in the past and required a migration. But it's the gnupg mess anyway and I don't expect us to be able to fix it.

I would personally prefer we do 3 and only fallback to 1 if not possible.

ImBearChild:
> I'm a little bit uneasy about having multiple versions of gnupg from different flatpaks access the same gnupg config folder & keys at the same time 

Actually, Freedesktop runtime contains GunPG (this is what Kleopatra using) but not pinentry, and I think there this the place where whole problem start. They used to ship the runtime with pinentry, but only commandline version of it. (and applications like Kleopatra and Kontact need a qt version) This casused some strange behavior, so they removed pinentry totally, and ships gpg without pinentry (I think this is very strange too, just like gpg without gpg-agent)

image

From my aspect, making KDE runtime carry a pinentry-qt and (if necessary) making Gnome runtime carry pinentry-gtk is good for both side. Packagers won't have to make gpg and pinentry for every application, and users won't take risk having multiple versions of gnupg. (Only two version at most, one from host, one from sdk.) And if we want to modify the code of GnuPG for an application, we have to replace the one from Freedesktop runtime, or at least make sure application use packager's gpg, that's another problem.

travier:
You'll have one version of gnupg from the host and one for each version of the FreeDesktop.org runtime installed on your system.

ImBearChild:
> each version of the FreeDesktop.org runtime installed on your system.

:laughing:  You're right, i just forgot this is also a multiple version problem. But other parts of reason still make sense, right?

travier:
> But other parts of reason still make sense, right?

If you ask me, nothing in the gnupg ecosystem makes any sense and it should all go away. gnupg should be a DBus session service that exposes a versioned API that clients could rely on. Instead we have strings over pipes and no sane error handling.

So I'm not in favor of adding support for anything gnupg into the runtime and I would prefer that individual applications carry this mess on their own, even if it means that we have a few lines of duplicated manifests for some apps.

ImBearChild:
> carry this mess on their own

They maybe can not carry this mess. There're some reasons for this:

At present, `gpg` is provided by FreeDesktop.org runtime, and flatpak packager can't mask or remove things from FreeDesktop.org runtime. (or packagers actually can remove it? I'm not very sure :rofl: ). And KDE runtime  needs `pinentry` and `pinentry-qt` to make gpg from FreeDesktop.org runtime fully functional. 

Compiling GnuPG for every apps is just a few lines of duplicated, but patch it is not. Making every packager to patch `gpg` and `gpg-agent` (maybe even more) is making things more complex, without bring any benefits to users and packagers.

Gnome have a application called Seahorse to store PGP keys and passwords, and Evolution can handle PGP mails. I found Gnome runtime have `pinentry` and `pinentry-gtk3` [included](https://gitlab.gnome.org/GNOME/gnome-build-meta/-/blob/master/elements/sdk/pinentry.bst). Since Gnome people are one of the world's first class systemd and DBus lovers and they add `pinentry` to their runtime, there's no reason to say "KDE should not do this because GPG is a old-styled stuff without DBus" from my view. :smiley: 

We're not talking about adding GPG to KDE runtime, it's already there (from FreeDesktop.org runtime. Modifying it is beyond packager's reach). Instead, this is just a missing dependency for graphical application.

barthalion:
You can just bundle pinentry in the app.

ImBearChild:
> You can just bundle pinentry in the app.

Yes, but sadly it won't work. As I said above:

> gpg-agent don't find pinentry executable in $PATH, instead, it use a hard-coded path `/usr/bin/pinentry` (only runtime can add this, but Freedesktop SDK and KDE sdk both not have this thing)

So app bundled pinentry won't work. It's installed in `/app/bin`, and gpg-agent just ignore it. I made s screenshot for this. In the picture, you can see `pinentry` is bundled, when kleopatra complain `missing pinentry` ( `缺少` means missing in Chinese )

image

In case that you want to verify my words, you can use `org.kde.kleopatra.json` and `gpgme.json` from this [commit](https://invent.kde.org/bearchild/flatpak-kde-applications/-/commit/52497ef97c96f82e71d0e6c1957fb5f21e3a4cc2) to build kleopatra with pinentry bundled.

I think everyone here will be happy to see any solution that solve this with a few lines of manifest. At least from my computer, it won't work.

ImBearChild:
Ok, I've done some more investigation .

Gpg get path of pinentry on configure it (before build it). Build it using `--with-pinentry-pgm=/app/bin/pinentry` may can fix this. But gpg of Flatpak is bundle with FreeDesktop.org runtime.

travier:
There is https://github.com/flatpak/xdg-desktop-portal/issues/178 & https://github.com/flathub/org.kde.kontact/issues/12

ImBearChild:
They eventually made a `--socket=gpg-agent` at [here](https://github.com/flatpak/flatpak/pull/4958) instead of a portal.

Erick555:
Manifest lacks `--filesystem=xdg-run/gnupg:ro` permission (the `--socket=gpg-agent` is brand new replacement). The gpg-agent should run on host (on some distros it may need to be activated) and call pinentry from host.

Mailaender:
bot, build org.kde.kleopatra

travier:
OK, I've tested this one and it works well for me now. Thanks for the tip @Erick555 ! We should do the same for Kmail & Kontact.

travier:
@barthalion Ready to merge :)

Erick555:
@travier I'm not sure about having both `--socket=gpg-agent` and `--filesystem=xdg-run/gnupg:ro` at same time.

At best one of them is redundant. At worst they would conflict with each other (did you tested with flatpak 1.14?)

I would drop `--socket=gpg-agent` and revisit it in a few years when users update to flatpak that support it. I don;t see value of using it when host socket is exposed at same time.

Erick555:
On other hand it may not hurt having both and it makes easier for users who have latest flatpak to disable filesystem permission by flatseal etc. So if latest flatpak have no issues with this then it may be fine. I'm curious which one takes precedence.

travier:
> @travier I'm not sure about having both `--socket=gpg-agent` and `--filesystem=xdg-run/gnupg:ro` at same time.
> 
> At best one of them is redundant. At worst they would conflict with each other (did you tested with flatpak 1.14?)
> 
> I would drop `--socket=gpg-agent` and revisit it in a few years when users update to flatpak that support it. I don;t see value of using it when host socket is exposed at same time.

Thanks for the info. Made an issue to track that: https://github.com/flathub/org.kde.kleopatra/issues/1

PR #3224, opened 2022-05-31
Add tv.plex.PlexHTPC

sztomi:

  - `--device=all`: Plex HTPC needs access to gamepads and LIRC controls because it is typically used as a kiosk-type application connected to the big screen. So instead of keyboard and mouse, these alternative controls are used.
  - `--filesystem=xdg-cache:rw`: Plex HTPC reads and writes this location for app-specific cache-able runtime data, such as widewine support library, on-demand codecs etc.
  - `--filesystem=xdg-config:rw`: Plex HTPC uses this location to read and write its own configuration
  - `--filesystem=xdg-data:rw`: Plex HTPC uses this location to store downloads (sync) data
  - `--talk-name=org.freedesktop.PowerManagement`: Plex HTPC is typicakky used as a fullscreen kiosk-type application and as such, displays an option to users to shut down the device. This is importand because HTPC setups usually don't have keyboard and mouse attached to them.
  - `--talk-name=org.freedesktop.ScreenSaver`: required to be able to disable the screensaver while playback is running
- [x] I am an upstream contributor to the project (I am employed by Plex)

Opening as draft because we need to do some further testing on our side, but in the meantime, feedback is welcome.


tv.plex.PlexHTPC.yml
```diff
@@ -0,0 +1,109 @@
+app-id: tv.plex.PlexHTPC
+runtime: org.gnome.Platform
+runtime-version: '42'
+sdk: org.gnome.Sdk
+command: /app/Plex.sh
+finish-args:
+  - --socket=x11
+  - --socket=wayland
```
hfiguiere, line 8: does it really support wayland?

tv.plex.PlexHTPC.yml
```diff
@@ -0,0 +1,109 @@
...
+  - --socket=pulseaudio
+  - --share=ipc
+  - --share=network
+  - --device=dri
+  - --device=all
```
hfiguiere, line 13:
```suggestion
  - --device=all
```
all means all.

tv.plex.PlexHTPC.yml
```diff
@@ -0,0 +1,109 @@
...
+  - --talk-name=org.freedesktop.ScreenSaver
+modules:
+  - name: plex-icons
+    buildsystem: simple
+    build-commands:
+      - install -Dm644 $FLATPAK_ID.desktop /app/share/applications/$FLATPAK_ID
+      - install -Dm664 $FLATPAK_ID.png /app/share/icons/hicolor/256x256/apps/$FLATPAK_ID.png
+    sources:
+      - type: dir
+        path: .
```
hfiguiere, line 32: please make this a `file` entry.

tv.plex.PlexHTPC.yml
```diff
@@ -0,0 +1,109 @@
...
+
+  - name: libjpeg # with libjpeg.so.8
```
hfiguiere, line 34: the libjpeg in the runtime isn't enough?

tv.plex.PlexHTPC.yml
```diff
@@ -0,0 +1,109 @@
...
+      - type: git
+        url: https://github.com/libevent/libevent/
+        branch: release-2.1.12-stable
+    cleanup:
+      - /bin
+      - /include
+      - /lib/pkgconfig
+      - /lib/cmake
+
+  - name: libwebp
```
hfiguiere, line 60: There is libwebp 1.2.1 in the runtime

tv.plex.PlexHTPC.yml
```diff
@@ -0,0 +1,109 @@
...
+    cleanup:
+      - '*.la'
+
+  - name: plex-binaries
+    buildsystem: simple
+    build-commands:
+      - cp -a * /app
+    sources:
+      - type: archive
+        url: https://artifacts.plex.tv/plex-htpc-experimental/1.18.0.3028-34988b94/linux/PlexHTPC-1.18.0.3028-34988b94-linux-x86_64.tar.bz2
```
hfiguiere, line 108: they don't have a aarch64?

tv.plex.PlexHTPC.yml
hfiguiere: also usually the missing icon is installed part of the main module

tv.plex.PlexHTPC.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
+Type=Application
+Name=Plex HTPC
+Comment=Plex HTPC for Linux is your client for playing on your computer connected to the big screen
+Exec=/usr/bin/flatpak run --branch=master --arch=x86_64 tv.plex.Htpc
```
hfiguiere, line 5: definitely NOT. This line should be the plain executable name that can be found in `/app/bin`

hfiguiere (changes requested):
you never install the appdata or desktop file

tv.plex.PlexHTPC.yml
```diff
@@ -0,0 +1,109 @@
...
+  - --device=dri
+  - --device=all
+  - --filesystem=xdg-videos:ro
+  - --filesystem=xdg-music:ro
+  - --filesystem=xdg-pictures:ro
+  - --filesystem=xdg-download:create
+  - --filesystem=xdg-cache:rw
+  - --filesystem=xdg-config:rw
+  - --filesystem=xdg-data:rw
+  - --filesystem=home:ro
```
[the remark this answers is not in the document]
sztomi: I've explained in the PR text why the non-standard ones are needed. If you have any further questions, let me know. But I wasn't aware this could be a point of contention since there are other closed source applications on flathub.

tv.plex.PlexHTPC.yml
sztomi: Not at the moment. We might add it in the future.

tv.plex.PlexHTPC.yml
sztomi: Maybe a different ABI version? It definitely wouldn't pick it up from the runtime until I added it. Same goes for libwebp.

hfiguiere: webp 0.5.1 is SIX YEARS old.

sztomi:
> you never install the appdata or desktop file

I install them here: https://github.com/flathub/flathub/blob/3fc9f26436a8c42add38b1901e606c1455af6c39/tv.plex.PlexHTPC.yml#L25..L32

tv.plex.PlexHTPC.yml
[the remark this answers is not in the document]
sztomi: We'll review the permissions used, but there might still be a good number of them that we'll need for the application to function.

sztomi:
@hfiguiere I couldn't find an example off the appdata xml being installed. Can you tell where it should be installed to?

hfiguiere:
Since you are upstream maybe you should consider building using the Freedesktop-sdk, so that you have a sane runtime and don't need to build 6 year old version of image decoder.

hfiguiere:
https://github.com/flathub/flathub/wiki/App-Requirements#appstream
https://github.com/flathub/flathub/wiki/AppData-Guidelines

tv.plex.PlexHTPC.yml
```diff
@@ -0,0 +1,109 @@
+app-id: tv.plex.PlexHTPC
+runtime: org.gnome.Platform
+runtime-version: '42'
+sdk: org.gnome.Sdk
```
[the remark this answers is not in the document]
nedrichards: Sorry that the docs are confusing. The 'basic' choice is the freedesktop platform. If you use Qt or are a KDE application then you should use the KDE platform (which is those things layered on top of freedesktop). If you use a bunch of GNOME libraries then you should use the GNOME platform, which has the same concept.

For historical and common dependency reasons GTK3 is in the freedesktop platform and it's usually worth you using the smallest surface area you can. The current version of FDo is 21.08.

tv.plex.PlexHTPC.yml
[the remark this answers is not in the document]
nedrichards: Yeah, the context of use being basically 'an operating system UI' is going to need more things than you'd otherwise want. Worth checking for sure though.

tv.plex.PlexHTPC.yml
hfiguiere: so 
```suggestion
  - --socket=fallback-x11
  - --socket=wayland
```

hfiguiere: but I'll add a note that I have seen many problems with Qt and Wayland, and if you build Qt into your package (which I think you do other it wouldn't work here, even more reason to use the smaller freedesktop sdk) then make sure this works when running a GNOME session running Wayland.

tv.plex.PlexHTPC.yml
hfiguiere: this is disappointing

tv.plex.PlexHTPC.yml
```diff
@@ -0,0 +1,109 @@
...
+command: /app/Plex.sh
```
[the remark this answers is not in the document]
sztomi: Like I said, I can manually navigate there in the shell and execute it. This is the full contents, copied from the flatpak:

```bash
#!/bin/sh
BASE_DIR=$(dirname "$(readlink -f "$0")")/..
export LD_LIBRARY_PATH=$BASE_DIR/lib/:$BASE_DIR:$LD_LIBRARY_PATH
export LC_ALL=C.UTF-8
export QML_IMPORT_PATH=$BASE_DIR/qml/:$QML_IMPORT_PATH
export QML2_IMPORT_PATH=$BASE_DIR/qml/:$QML2_IMPORT_PATH
export QT_PLUGIN_PATH=$BASE_DIR/plugins/:$QT_PLUGIN_PATH
export QTWEBENGINEPROCESS_PATH=$BASE_DIR/bin/QtWebEngineProcess
export QTDIR=$BASE_DIR
export QT_QPA_PLATFORM_PLUGIN_PATH=$BASE_DIR/plugins/platforms:$QT_QPA_PLATFORM_PLUGIN_PATH
export FONTCONFIG_PATH=${FONTCONFIG_PATH:-/etc/fonts}

# In snap: /home/<user>/snap/plex/common
# Flatpak: /home/<user>/.var/app/$FLATPAK_ID/data/plex
# Native:  /home/<user>/.local/share/plex
export PLEX_USERDATA_DIR=${SNAP_USER_COMMON:-${XDG_DATA_HOME:-$HOME/.local/share}/plex}
# In snap: /snap/plex/<revision>/resources
# Flatpak:  <this file's location>/resources
# Native:  <this file's location>/resources
export PLEX_RESOURCE_PATH=$BASE_DIR/resources
# In snap: /home/<user>/snap/plex/common
# Flatpak: /home/<user>/.var/app/$FLATPAK_ID/cache/plex
# Native:  /home/<user>/.cache/plex
export PLEX_CACHE_DIR=${SNAP_USER_COMMON:-${XDG_CACHE_HOME:-$HOME/.cache/plex}}
# In snap: /home/<user>/snap/plex/<revision>
# Flatpak: /home/<user>/.var/app/$FLATPAK_ID/data/plex
# Native:  /home/<user>/.local/share/plex
export PLEX_STORAGE_INI_DIR=$PLEX_USERDATA_DIR
# In snap: /home/<user>/snap/plex/common/Logs
# Flatpak: /home/<user>/.var/app/$FLATPAK_ID/data/plex/Logs
# Native:  /home/<user>/.local/share/plex/Logs
export PLEX_LOG_DIR=$PLEX_USERDATA_DIR/Logs

export PLEX_MEDIA_SERVER_RESOURCES=$BASE_DIR/resources

# For MPV to get hardware decoding
export LIBVA_DRIVERS_PATH=${SNAP:-${BASE_DIR}}/lib/dri

"$BASE_DIR"/bin/Plex "$@"
```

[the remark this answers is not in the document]
sztomi: I pushed the current state (only some, not all comments were addressed so far)

tv.plex.PlexHTPC.yml
```diff
@@ -0,0 +1,106 @@
+app-id: tv.plex.PlexHTPC
+runtime: org.freedesktop.Platform.GL.default
```
hfiguiere, line 2: this is not a valid runtime.

tv.plex.PlexHTPC.yml
sztomi: I'm not able to replace libwebp with a newer version. We are building Qt 5.15.3 from source and it vendors an old version. 0.5.1 seems to be the last version with ABI 6 (I could be wrong, but that's what I found).

tv.plex.PlexHTPC.yml
sztomi: This is where I got it from: https://docs.flatpak.org/en/latest/available-runtimes.html?highlight=runtimes#freedesktop

Are the docs incorrect?

hfiguiere:
```suggestion
runtime: org.freedesktop.Platform
```

It says **extension**

tv.plex.PlexHTPC.yml
sztomi: Because it doesn't install it for some reason and cqtdeployer doesn't collect it.

tv.plex.PlexHTPC.yml
sztomi: @hfiguiere You are not exactly being helpful here. It also says *"Available Freedesktop runtimes:"* right before that list. It is not explained what an extension means, and I assumed we need that if we need GL support (which we do). Is that not the case?

tv.plex.PlexHTPC.yml
sztomi: As discussed in https://github.com/flathub/flathub/pull/3224#discussion_r886928228

tv.plex.PlexHTPC.yml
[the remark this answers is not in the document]
sztomi: Still stuck on this. I straced the bwrap-args contents and it seems to correctly bind /app. I can also see the files on the host filesystem. https://gist.github.com/sztomi/18217510135f165f0326a690352923d8

tv.plex.PlexHTPC.yml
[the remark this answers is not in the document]
sztomi: Replaced with the freedesktop runtime.

tv.plex.PlexHTPC.yml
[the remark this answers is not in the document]
sztomi: This was resolved by using the correct runtime (`org.freedesktop.Platform`). This error is a red herring. If the runtime is invalid, it should error during the build.

tv.plex.PlexHTPC.yml
sztomi: Done

tv.plex.PlexHTPC.yml
sztomi: I'm resolving this because I'm unable to do anything about it right now. It is noted that it would be beneficial to ensure that the binary package works standalone.

tv.plex.PlexHTPC.yml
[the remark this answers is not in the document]
hfiguiere: I didn't realize you had completely ignored my review comment where I said it was an invalid runtime.

tv.plex.PlexHTPC.yml
sztomi: It turns out that we are missing the wayland Qt plugin from our build, so I've commented out wayland for now.

tv.plex.PlexHTPC.desktop
sztomi: Fixed

tv.plex.PlexHTPC.yml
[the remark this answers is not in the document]
sztomi: I've removed a number of permissions. See the updated post for explanations for the remaining ones.

nedrichards:
> @nedrichards the plexamp flatpak is not maintained by us. Currently we only have an official AppImage release. We might want to release an official flatpak for plexamp later. As for the maintainers for the repo, it would be best if you could add this user: https://github.com/plexappbuilder
> 
> I'll update this PR soon, addressing the feedback.

You might want to have a chat with @simonxciv and @elan who've been in contact with the flathub team directly over plexamp. But all good on adding that extra user, and keep your eyes open for some of the new direct publishing workflows coming soon!

nedrichards:
bot, build tv.plex.PlexHTPC

tv.plex.PlexHTPC.yml
```diff
@@ -0,0 +1,122 @@
+app-id: tv.plex.PlexHTPC
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+command: Plex
+finish-args:
+  - --socket=x11
+  # currently now working - --socket=wayland
```
nedrichards, line 8: do you mean currently not working here?

sztomi: Currently not working because we missed shipping the Qt wayland plugin. Will be corrected in the next release, but until then, the application doesn't run under wayland.

nedrichards: that's fine - it's just that the comment says 'now working' which means the opposite. Fine to add it next release, when you do switch x11 to fallback-x11 to get wayland picked by default if on a wayland native system (it will have no effect on an x11 native system which will continue to use x11 correctly).

sztomi: Oh! I misread your question and didn't notice the typo until now. I'll fix that comment to avoid confusion.

tv.plex.PlexHTPC.yml
```diff
@@ -0,0 +1,122 @@
...
+command: Plex
+finish-args:
+  - --socket=x11
+  # currently not working - --socket=wayland
+  - --socket=pulseaudio
+  - --share=ipc
+  - --share=network
+  - --device=all
+  - --filesystem=xdg-cache:rw
+  - --filesystem=xdg-config:rw
```
hfiguiere, line 14: why do you need to access other app (non flatpak) configuration or worse write into it? same for `xdg-data` and `xdg-cache`

hfiguiere (changes requested):

tv.plex.PlexHTPC.yml
```diff
@@ -0,0 +1,122 @@
...
+  - --filesystem=xdg-data:rw
```
barthalion, line 15:
```suggestion
```

At least in theory, these three shouldn't be needed. Flatpak exports `$XDG_CONFIG_HOME`, `$XDG_CACHE_HOME` and `$XDG_DATA_HOME` environment variables that are writable by default, and it would be the best to make the app respect these if they aren't. If not, please limit these to specific directories needed by Plex. e.g. `--filesystem=xdg-config/plex` (just an example as I did not check it!(

barthalion:
bot, build tv.plex.PlexHTPC

tv.plex.PlexHTPC.yml
barthalion: I've just checked and it seems to properly use all three variables by default.

sztomi: Correct, it respects these variables. But what are the permissions for then?

barthalion: They can be dropped altogether. These permissions are for vendors who want their apps to keep using existing configuration outside the sandbox, or access files belonging to other programs (e.g. mod manager for a game).

tv.plex.PlexHTPC.yml
sztomi: Yeah I don't actually need that, I just misunderstood what these permissions do. Thanks for your help! 🙏

PR #3231, opened 2022-06-02
add net.kubux.Muehle

k-u-bux:


net.kubux.Muehle.yml
```diff
@@ -0,0 +1,23 @@
+app-id: net.kubux.Muehle
+branch: stable
```
hfiguiere, line 2:
```suggestion
```

net.kubux.Muehle.yml
```diff
@@ -0,0 +1,23 @@
...
+    buildsystem: simple
+    build-commands:
+      - tar -xjpvf muehle.tbz -C /
+    sources:
+      - type: file
+        url: http://www.kubux.net/muehle/muehle.tbz
+        dest-filename: muehle.tbz
+        sha256: f6d78747981113595145a326eded221afda857bfdd3cf92acaa90b313ed26300
+        size: 994654
+finish-args:
```
hfiguiere, line 18: finish args are usually above modules.

net.kubux.Muehle.yml
```diff
@@ -0,0 +1,23 @@
...
+sdk: org.freedesktop.Sdk
+command: muehle
+modules:
+  - name: muehle
+    buildsystem: simple
+    build-commands:
+      - tar -xjpvf muehle.tbz -C /
+    sources:
+      - type: file
+        url: http://www.kubux.net/muehle/muehle.tbz
```
hfiguiere, line 14: this as a file without a version. Which might not be stable when it gets updated for a new version. (changing the sha256)

net.kubux.Muehle.yml
```diff
@@ -0,0 +1,23 @@
...
+        dest-filename: muehle.tbz
+        sha256: f6d78747981113595145a326eded221afda857bfdd3cf92acaa90b313ed26300
+        size: 994654
```
hfiguiere, line 17:
```suggestion
```

totally unecessary

net.kubux.Muehle.yml
```diff
@@ -0,0 +1,23 @@
...
+branch: stable
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+command: muehle
+modules:
+  - name: muehle
+    buildsystem: simple
+    build-commands:
+      - tar -xjpvf muehle.tbz -C /
```
hfiguiere, line 11: so you install a binary and build a flatpak for both x86_64 and aarch64. Not sure that will work well.

net.kubux.Muehle.yml
```diff
@@ -0,0 +1,23 @@
...
+        url: http://www.kubux.net/muehle/muehle.tbz
+        dest-filename: muehle.tbz
+        sha256: f6d78747981113595145a326eded221afda857bfdd3cf92acaa90b313ed26300
+        size: 994654
+finish-args:
+  # X11 + XShm access
+  - --share=ipc
+  - --socket=x11
+  # store two numbers in muehle.rc: AI strength and boardsize in pixels
+  - --filesystem=xdg-config:create
```
hfiguiere, line 23: I don't think this is needed

hfiguiere:
bot, build net.kubux.Muehle

hfiguiere (changes requested):

net.kubux.Muehle.yml
k-u-bux: My intent was not to build for x85_64 and aarch64. In fact, I don't want to build for aarch64 as I could not test that anyway. What do I need to say in the manifest to prevent the build?

net.kubux.Muehle.yml
k-u-bux: In order to make settings persistent, this is what I found works. I tried to be more specific with the file path but that did not create the file and the settings ended up not being persistent. Please advise.

k-u-bux: OK, thanks for the hint. Turns out that
  --filesystem=xdg-config:rw
does indeed suffice.

net.kubux.Muehle.yml
k-u-bux: I learned about flathub.json and committed a file that should disable the aarch64 build.

hfiguiere: you could otherwiser build everything on the flathub infrastructure. aarch64 would just work.

it's disappointing to not build aarch64 for no apparent reason.

net.kubux.Muehle.yml
hfiguiere: > In order to make settings persistent, this is what I found works. I tried to be more specific with the file path but that did not create the file and the settings ended up not being persistent. Please advise.

by using XDG_CONFIG_HOME properly instead of `~/.config`. But I can't see the source code, so I'll just assume that's what's being done wrong. Also `--filesystem=xdg-config` give access to any config from any other app outside of the sandbox.

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": [ "x86_64" ]
```
hfiguiere, line 2: I'll reiterate my disappointment here.

nedrichards:
bot, build net.kubux.Muehl

hfiguiere:
bot, build net.kubux.Muehle

net.kubux.Muehle.yml
k-u-bux: The reason is that I cannot test the resulting binary. I don't own such hardware. I would not be able to support that built and fix bugs.

hfiguiere: it sure doesn't open the door to that.

k-u-bux: Well, the source code is not available online; and at this time, I have no plans of changing that. That is the rationale behind the build command.

k-u-bux:
I installed the build 94555 as per the given command. Installation gave a warning:

Warning: Not exporting file net.kubux.Muehle.appdata.xml of unsupported type.
Installation complete.

The resulting application, however, behaves as expected.

PR #3242, opened 2022-06-08
Add tv.plex.PlexDesktop

sztomi:

- [ ] I am using only the minimal set of permissions. *(If not, please explain each non-standard permission.)*:
  - `--talk-name=org.freedesktop.ScreenSaver`: used to inhibit screensaver during playback

This is a very similar package to tv.plex.PlexHTPC (https://github.com/flathub/flathub/pull/3224), except it requires a smaller set of permissions.


hfiguiere (changes requested):
there is no manifest.

farribeiro:
bot, build tv.plex.PlexDesktop

tv.plex.PlexDesktop.yml
```diff
@@ -0,0 +1,117 @@
...
+      - install -Dm644 tv.plex.PlexDesktop.desktop /app/share/applications/tv.plex.PlexDesktop.desktop
+      - install -Dm664 tv.plex.PlexDesktop.png /app/share/icons/hicolor/256x256/apps/tv.plex.PlexDesktop.png
+      - install -Dm664 tv.plex.PlexDesktop.metainfo.xml /app/share/metainfo/tv.plex.PlexDesktop.metainfo.xml
+    sources:
+      - type: file
+        path: tv.plex.PlexDesktop.desktop
+      - type: file
+        path: tv.plex.PlexDesktop.png
+      - type: file
+        path: tv.plex.PlexDesktop.metainfo.xml
```
hfiguiere, line 36: seriously all three files should be part of your upstream build. A desktop app that doesn't ship a .desktop or an icon file, it's basically a bug....

sztomi: Like I said in the other PR, we are only shipping snaps and flatpaks, and both builds add the icons and desktop files. What you consider upstream build (the tarball containing the binaries) is not a package we are handing to users. It is simply the output directory from cqtdeployer, archived. We are also not making it artificially difficult to grab that tarball, but it should be clear to anyone downloading it that it's not an official release. In fact, I added the tarball build precisely to make it easily downloadable (instead of pulling the snap file like spotify does, which is a pain).

hfiguiere: I will rephrase: _they should be part of the main build artifacts._ 

**Neither is specific to Flathub** (can't speak of snap, but MANY MANY packages ship and install these files and we can use them *verbatim*).

sztomi: Plenty of packages do it this way, including tv.plex.PlexHTPC that was merged just days ago (that you also reviewed). Not sure why you are insisting on changing it in this case.

tv.plex.PlexDesktop.yml
```diff
@@ -0,0 +1,117 @@
+app-id: tv.plex.PlexDesktop
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+command: Plex
+finish-args:
+  - --socket=x11
```
nedrichards, line 7: I assume this still has the same wayland issues as the HTPC version.

sztomi: Correct. Both will get the fix at the same time (they share a lot of code).

tv.plex.PlexDesktop.yml
barthalion: We have talked about this with Nick, and we both think that as an upstream, you decide where you want to store these files. The requirement of shipping them with upstream code/tarball is pretty much for downstream submitters.

barthalion:
It generally looks good to me. Please look into improving the description after the merge.

/merge @plexappbuilder

tv.plex.PlexDesktop.metainfo.xml
```diff
@@ -0,0 +1,35 @@
...
+  <launchable type="desktop-id">tv.plex.PlexDesktop.desktop</launchable>
+  <name>Plex</name>
+  <developer_name>Plex</developer_name>
+  <summary>Plex client for desktop computers</summary>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>LicenseRef-proprietary=https://www.plex.tv/en-gb/about/privacy-legal/plex-terms-of-service/</project_license>
+  <url type="homepage">https://plex.tv</url>
+  <url type="help">https://support.plex.tv/articles/</url>
+  <description>
+    <p>Plex for Linux is your client for playing on your Linux desktop computer</p>
```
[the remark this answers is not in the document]
sztomi: Ah yeah, for some reason I thought this description needs to be shorter. I'll follow up with a fix.

PR #3275, opened 2022-06-19
Add manifest for PSPP.

blp:

### DIFF
--- a/org.fsf.pspp.yml
+++ b/org.fsf.pspp.yml
@@ -1,6 +1,6 @@
 app-id: org.fsf.pspp
 runtime: org.gnome.Platform
-runtime-version: '41'
+runtime-version: '42'
 sdk: org.gnome.Sdk
 command: psppire
 modules:
--- a/org.fsf.pspp.yml
+++ b/org.fsf.pspp.yml
@@ -1,0 +1,33 @@
+app-id: org.fsf.pspp
+runtime: org.gnome.Platform
+runtime-version: '41'
+sdk: org.gnome.Sdk
+command: psppire
+modules:
+  - name: spread-sheet-widget
+    buildsystem: autotools
+    sources:
+      - type: archive
+        url: https://alpha.gnu.org/gnu/ssw/spread-sheet-widget-0.8.tar.gz
+        sha256: 8589d8298fcf3b5850d0968b04801a4f40faf0555544f6cc9d954b0162e9954b
+  - name: gsl
+    buildsystem: autotools
+    sources:
+      - type: archive
+        url: https://ftp.gnu.org/gnu/gsl/gsl-2.7.1.tar.gz
+        sha256: dcb0fbd43048832b757ff9942691a8dd70026d5da0ff85601e52687f6deeb34b
+  - name: pspp
+    buildsystem: autotools
+    sources:
+      - type: archive
+        url: https://ftp.gnu.org/gnu/pspp/pspp-1.6.0.tar.gz
+        sha256: 8d5b51dba8ad812ee93288ad664d8184f164d3b394d2f9c09fcb39f86e0b0a12
+finish-args:
+  # X11 + XShm access
+  - --share=ipc
+  - --socket=x11
+  # Wayland access
+  - --socket=wayland
+  - --own-name=gnu.pspp
+  - --talk-name=org.gtk.vfs.*
+  - --filesystem=xdg-run/gvfsd
--- a/org.fsf.pspp.yml
+++ b/org.fsf.pspp.yml
@@ -25,7 +25,7 @@ modules:
 finish-args:
   # X11 + XShm access
   - --share=ipc
-  - --socket=x11
+  - --socket=fallback-x11
   # Wayland access
   - --socket=wayland
   - --own-name=gnu.pspp
--- a/org.fsf.pspp.yml
+++ b/org.fsf.pspp.yml
@@ -3,6 +3,15 @@ runtime: org.gnome.Platform
 runtime-version: '42'
 sdk: org.gnome.Sdk
 command: psppire
+finish-args:
+  # X11 + XShm access
+  - --share=ipc
+  - --socket=fallback-x11
+  # Wayland access
+  - --socket=wayland
+  - --own-name=gnu.pspp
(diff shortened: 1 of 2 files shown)


org.fsf.pspp.yml
```diff
@@ -0,0 +1,33 @@
+app-id: org.fsf.pspp
+runtime: org.gnome.Platform
+runtime-version: '41'
```
hfiguiere, line 3:
```suggestion
runtime-version: '42'
```

org.fsf.pspp.yml
```diff
@@ -0,0 +1,33 @@
...
+      - type: archive
+        url: https://ftp.gnu.org/gnu/gsl/gsl-2.7.1.tar.gz
+        sha256: dcb0fbd43048832b757ff9942691a8dd70026d5da0ff85601e52687f6deeb34b
+  - name: pspp
+    buildsystem: autotools
+    sources:
+      - type: archive
+        url: https://ftp.gnu.org/gnu/pspp/pspp-1.6.0.tar.gz
+        sha256: 8d5b51dba8ad812ee93288ad664d8184f164d3b394d2f9c09fcb39f86e0b0a12
+finish-args:
```
hfiguiere, line 25: the finish args session is usually above `modules`.

org.fsf.pspp.yml
```diff
@@ -0,0 +1,33 @@
...
+  # X11 + XShm access
+  - --share=ipc
+  - --socket=x11
```
hfiguiere, line 28:
```suggestion
  - --socket=fallback-x11
```

hfiguiere:
bot, build org.fsf.pspp

hfiguiere (changes requested):

org.fsf.pspp.yml
```diff
@@ -0,0 +1,33 @@
...
+  # Wayland access
+  - --socket=wayland
+  - --own-name=gnu.pspp
```
hfiguiere, line 31: I have questions about the validity of that D-Bus name. Note that this is not blocking here assuming this is what upstream uses.

org.fsf.pspp.yml
```diff
@@ -0,0 +1,33 @@
+app-id: org.fsf.pspp
```
hfiguiere, line 1: shouldn't it be `org.gnu.pspp` ?

blp: I thought that I'd found other flatpaks that use org.fsf, but I don't see any. I do see org.gnu.emacs. I guess it should be org.gnu.pspp.

org.fsf.pspp.yml
blp: I think this is supposed to match the parameter to gtk_application_new(), which uses `gnu.pspp` as shown here: https://git.savannah.gnu.org/cgit/pspp.git/tree/src/ui/gui/main.c#n525

org.fsf.pspp.yml
hfiguiere: what does upstream think? (ie the developers if that project)

org.fsf.pspp.yml
hfiguiere: I know, but that remain there are questions since `gnu` isn't a valid tld. 

https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names-bus

org.fsf.pspp.yml
blp: For PSPP? I'm the lead developer. I don't have a strong opinion.

blp:
> Build [96656 successful](https://buildbot.flathub.org/#/builders/32/builds/96656) To test this build, install it from the testing repository:
> 
> ```
> flatpak install --user https://dl.flathub.org/build-repo/94387/org.fsf.pspp.flatpakref
> ```

At first glance, at least, this build works.

org.fsf.pspp.yml
hfiguiere: ok so then we can discuss the app-id for `gtk_application_new()`. and that should be the one used here too.

hfiguiere: tbf `org.fsf` or `org.gnu` isn't different to me, but I'd consider the GNU project -> `org.gnu`.

org.fsf.pspp.yml
hfiguiere: and just to add, by default we grant the `app-id` valid as the `own-name`, hence the idea that they should match.

org.fsf.pspp.yml
blp: OK, I pushed a commit to adjust the order.

hfiguiere:
bot, build org.fsf.pspp

blp:
The second build (96659) works OK too.

org.fsf.pspp.yml
blp: OK... I guess I'll come back when I can get an upstream release with all this updated then.

barthalion:
Marking as a draft for now, good to merge once it's renamed to `org.gnu.pspp`.

org.gnu.pspp.yml
```diff
@@ -0,0 +1,33 @@
...
+runtime-version: '42'
+sdk: org.gnome.Sdk
+command: psppire
+finish-args:
+  # X11 + XShm access
+  - --share=ipc
+  - --socket=fallback-x11
+  # Wayland access
+  - --socket=wayland
+  - --own-name=org.gnu.pspp
```
hfiguiere, line 12: this is automatically granted
```suggestion
```

hfiguiere:
bot, build org.gnu.pspp

hfiguiere (changes requested):

barthalion:
On my end, some icons appear missing:

image

This is on Fedora Silverblue 35.

blp:
Yes. Oops. I'll fix it in another release in a few days.

blp:
In this case it was not a GNOME problem, it was an incomplete transition from "gnu.pspp" to "org.gnu.pspp" because there was another place I hadn't found.

blp:
My recent push fixes the icons.

barthalion:
bot, build org.gnu.pspp

PR #3280, opened 2022-06-21
add slomoVideo, an app to create slow-motion (or fast-motion) videos

muelli:
http://slowmovideo.granjow.net/

The app seems to work, except for the rendered video not appearing
anywhere. This also happens with the "native" version, i.e. the one
Ubuntu ships, so I doubt it's a side effect of running it as a flatpak.

I whipped up an appdata file with the help of https://librebob.gitlab.io/appdata-generator/ but it fails validation:
```
FAILED:
• tag-missing           : <description> required
Validation of files failed
```
unfortunately, it doesn't tell me *where* it expects that description tag or why the current one is not sufficient.

I needed to compile ffmpeg, because the extension doesn't work! It doesn't seem to have the null or image2 filter which the app requires. I took it from the other app I maintain: https://github.com/flathub/no.mifi.losslesscut/blob/master/ffmpeg.yaml.

- [ ] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://github.com/slowmoVideo/slowmoVideo/issues/201
- [ ] I own the domain used in the application ID or the domain has a policy for delegating subdomains (e.g. GitHub, SourceForge). **No, but I mentioned flathub in https://github.com/slowmoVideo/slowmoVideo/issues/201**

### DIFF
--- a/cmake.diff
+++ b/cmake.diff
@@ -1,0 +1,15 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 38ff055..b1b8867 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -44,6 +44,10 @@ else ()
+ 	set(PROJECT_VERSION_PATCH "0")
+ endif ()
+ 
++set(PROJECT_VERSION_MAJOR "0")
++set(PROJECT_VERSION_MINOR "6")
++set(PROJECT_VERSION_PATCH "0")
++
+ set(PROJECT_VERSION "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}")
+ message(STATUS "Building slowmoVideo ${PROJECT_VERSION}")
+ configure_file(src/version.h.in version.h)
--- a/ffmpeg.yaml
+++ b/ffmpeg.yaml
@@ -1,0 +1,59 @@
+name: ffmpeg
+config-opts:
+  - --enable-shared
+  - --disable-static
+  - --disable-doc
+  - --disable-manpages
+  - --disable-stripping
+  - --disable-ffplay
+  - --enable-ffprobe
+
+  - --disable-securetransport
+  - --enable-runtime-cpudetect
+
+  - --enable-gpl
+  - --enable-version3
+  - --enable-optimizations
+  - --enable-postproc
+  - --enable-pthreads
+  #- --enable-gnutls
+  - --enable-libxcb-xfixes
+  - --enable-opengl
+  #- --enable-sdl2
+  - --enable-vaapi
+  - --enable-vdpau
+  #- --enable-zlib
+  #- --enable-bzlib
+  #- --enable-lzma
+
+  - --enable-fontconfig
+  - --enable-libfontconfig
+  - --enable-libfreetype
+
+  # copied from another manifest
+  - --enable-avcodec
+  - --enable-avfilter
+  - --enable-filters
+  - --enable-avdevice
+  - --enable-avresample
+  #- --enable-libxml2
+  - --enable-swscale
+
+  - --enable-libxcb
+
+  - --enable-protocol=file
+
+#post-install:
+    # This is what losslesscut tried to execute:
+    #- /app/bin/ffmpeg -hide_banner -f lavfi -i nullsrc=s=256x256:d=1 -f null -
+
+sources:
+  - type: archive
+    url: https://ffmpeg.org/releases/ffmpeg-4.4.tar.xz
+    sha256: 06b10a183ce5371f915c6bb15b7b1fffbe046e8275099c96affc29e17645d909
+    x-checker-data:
+          type: anitya
+          project-id: 5405
+          url-template: https://ffmpeg.org/releases/ffmpeg-$version.tar.xz
+cleanup:
+  - /share/ffmpeg/examples
--- a/net.granjow.slowmovideo.appdata.xml
+++ b/net.granjow.slowmovideo.appdata.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+    <id>net.granjow.slowmovideo</id>
+    <launchable type="desktop-id">net.granjow.slowmovideo.desktop</launchable>
+    <name>slomoVideo</name>
+    <developer_name>Simon A. Eugster</developer_name>
+    <project_license>GPL-3.0</project_license>
+    <metadata_license>CC0-1.0</metadata_license>
+    <summary>slowmoVideo is an OpenSource program that creates slow-motion videos from your footage.</summary>
+    <description><p>But it does not simply make your videos play at 0.01x speed. You can smoothly slow down and speed up your footage, optionally with motion blur.
+    
+    How does slow motion work? slowmoVideo tries to find out where pixels move in the video (this information is called Optical Flow), and then uses this information to calculate the additional frames between the ones recorded by your camera.
+    </p></description>
+    <url type="homepage">http://slowmovideo.granjow.net/</url>
+    <content_rating type="oars-1.1" />
+    <screenshots>
+        <screenshot type="">
+            <image type="source">http://slowmovideo.granjow.net/pngGUI/newProject.png</image>
+            <image type="source">http://slowmovideo.granjow.net/pngGUI/simpleCurveAnnotated.png</image>
+            <image type="source">http://slowmovideo.granjow.net/pngGUI/bezierCurve.png</image>
+        </screenshot>
+    </screenshots>
+    <releases>
+        <release version="v0.6" date="2020-04-12"/>
+    </releases>
+</component>
--- a/net.granjow.slowmovideo.yaml
+++ b/net.granjow.slowmovideo.yaml
@@ -53,7 +53,7 @@ modules:
     builddir: true
     post-install:
         - install -D -ma+r /app/share/icons/AppIcon.png /app/share/icons/hicolor/256x256/apps/AppIcon.png
-        - install -D -ma+r net.granjow.slowmovideo.appdata.xml /app/share/appdata/net.granjow.slowmovideo.appdata.xml
+        - install -D -m644 ../net.granjow.slowmovideo.appdata.xml /app/share/appdata/net.granjow.slowmovideo.appdata.xml
     sources:
         ## Unfortunately, the archive does not contain the necessary submodules
         #- type: archive
--- a/net.granjow.slowmovideo.yaml
+++ b/net.granjow.slowmovideo.yaml
@@ -10,8 +10,8 @@ finish-args:
   - --device=dri
   - --socket=x11
   - --socket=pulseaudio
-  # Because the file chooser somehow doesn't show the files on the host, we export some directory for reading pcap files
-  - --filesystem=xdg-public-share
+  # The app creates a directory which does not work through the portal
+  - --filesystem=xdg-videos
 
 
 modules:
--- a/net.granjow.slowmovideo.yaml
+++ b/net.granjow.slowmovideo.yaml
@@ -1,6 +1,6 @@
 app-id: net.granjow.slowmovideo
 runtime: org.kde.Platform
-runtime-version: '5.15-22.08'
+runtime-version: '5.15-21.08'
 sdk: org.kde.Sdk
 
 command: slowmoUI
--- a/net.granjow.slowmovideo.yaml
+++ b/net.granjow.slowmovideo.yaml
@@ -8,6 +8,7 @@ rename-icon: AppIcon
 rename-desktop-file: slowmoUI.desktop
 finish-args:
   - --device=dri
+  - --share=ipc
   - --socket=x11
   - --socket=pulseaudio
   # The app creates a directory which does not work through the portal
--- a/net.granjow.slowmovideo.yaml
+++ b/net.granjow.slowmovideo.yaml
@@ -1,6 +1,6 @@
 app-id: net.granjow.slowmovideo
 runtime: org.kde.Platform
-runtime-version: '5.15-21.08'
+runtime-version: '5.15-22.08'
 sdk: org.kde.Sdk
 
 command: slowmoUI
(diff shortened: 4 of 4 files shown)


net.granjow.slowmovideo.appdata.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+    <id>net.granjow.slowmovideo</id>
+    <launchable type="desktop-id">net.granjow.slowmovideo.desktop</launchable>
+    <name>slomoVideo</name>
+    <developer_name>Simon A. Eugster</developer_name>
+    <project_license>GPL-3.0</project_license>
+    <metadata_license>CC0-1.0</metadata_license>
+    <summary>slowmoVideo is an OpenSource program that creates slow-motion videos from your footage.</summary>
+    <description>But it does not simply make your videos play at 0.01x speed. You can smoothly slow down and speed up your footage, optionally with motion blur.
```
hfiguiere, line 10: the content of `<description>` should wrapped in at least a `<p>`

net.granjow.slowmovideo.yaml
```diff
@@ -0,0 +1,84 @@
+app-id: net.granjow.slowmovideo
+runtime: org.kde.Platform
+runtime-version: '5.15-21.08'
+sdk: org.kde.Sdk
+x-add-extensions:
```
hfiguiere, line 5: `x-add-extensions` ?

net.granjow.slowmovideo.yaml
```diff
@@ -0,0 +1,84 @@
...
+        version: "21.08"
+        add-ld-path: "."
+
+cleanup-commands:
+    - mkdir -p /app/lib/ffmpeg
+
+command: slowmoUI
+rename-icon: AppIcon
+rename-desktop-file: slowmoUI.desktop
+#rename-appdata-file: hugin.appdata.xml
```
hfiguiere, line 17:
```suggestion
```

net.granjow.slowmovideo.yaml
```diff
@@ -0,0 +1,84 @@
...
+        - type: git
+          url:  https://github.com/slowmoVideo/slowmoVideo.git
+          tag: v0.6
+          commit: 08536c2e7887e3bf12973994f0365716528c7f16
+          x-checker-data:
+            type: git
+
+        # https://github.com/slowmoVideo/slowmoVideo/issues/200
+        - type: patch
+          path: cmake.diff
```
hfiguiere, line 80: patch file is missing here.

net.granjow.slowmovideo.yaml
```diff
@@ -0,0 +1,84 @@
...
+modules:
+
+  - name: opencv
+    # Taken from https://github.com/flathub/org.d3cod3.Mosaic/blob/be9aea900361940120341b2ad6fe7d97bec39c66/org.d3cod3.Mosaic.yml#L152
+    buildsystem: cmake-ninja
+    builddir: true
+    cleanup:
+        - '/include/'
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Debug
```
hfiguiere, line 35:
```suggestion
      - -DCMAKE_BUILD_TYPE=RelWithDebInfo
```

A debug build is unlikely what you want.

ffmpeg.yaml
```diff
@@ -0,0 +1,59 @@
...
+sources:
+  - type: archive
+    url: https://ffmpeg.org/releases/ffmpeg-4.4.tar.xz
+    sha256: 06b10a183ce5371f915c6bb15b7b1fffbe046e8275099c96affc29e17645d909
+    x-checker-data:
+          type: anitya
+          project-id: 5405
+          url-template: https://ffmpeg.org/releases/ffmpeg-$version.tar.xz
+cleanup:
+  - /share/ffmpeg/examples
```
hfiguiere, line 59: you could as well disable examples at build time.

net.granjow.slowmovideo.yaml
```diff
@@ -0,0 +1,84 @@
...
+      - -DWITH_1394=OFF
+      - -DWITH_PROTOBUF=ON
+      # - -DOPENCV_EXTRA_MODULES_PATH=../opencv_contrib-master/modules ../opencv-master
+      #- -DBUILD_LIST=calib3d,core,dnn,features2d,flann,highgui,imgcodecs,imgproc,java_bindings_generator,ml,objdetect,photo,python_bindings_generator,python_tests,stitching,video,videoio
+    sources:
+        - type: archive
+          url: https://github.com/opencv/opencv/archive/3.4.18.tar.gz
+          sha256: d6a69aa4a775a074662685b9bc2b07b88fb867f19d939b03ad281571b62f2620
+
+  - ffmpeg.yaml
```
hfiguiere, line 58: I see no justification for a separate file.

net.granjow.slowmovideo.yaml
```diff
@@ -0,0 +1,84 @@
...
+command: slowmoUI
+rename-icon: AppIcon
+rename-desktop-file: slowmoUI.desktop
+#rename-appdata-file: hugin.appdata.xml
+finish-args:
+  - --device=dri
+  - --socket=x11
+  - --socket=pulseaudio
+  # Because the file chooser somehow doesn't show the files on the host, we export some directory for reading pcap files
+  - --filesystem=xdg-public-share
```
hfiguiere, line 23: are you sure this is the right one?

hfiguiere (changes requested):

net.granjow.slowmovideo.yaml
muelli: ah, I played around with disabling ffmpeg. Let me remove this. Thanks.

net.granjow.slowmovideo.yaml
hfiguiere: cut and paste?

net.granjow.slowmovideo.appdata.xml
```diff
@@ -0,0 +1,25 @@
...
+    
+    How does slow motion work? slowmoVideo tries to find out where pixels move in the video (this information is called Optical Flow), and then uses this information to calculate the additional frames between the ones recorded by your camera.</description>
+    <url type="homepage">http://slowmovideo.granjow.net/</url>
+    <content_rating type="oars-1.1" />
+    <screenshots>
+        <screenshot type="">
+            <image type="source" width="622" height="452">http://slowmovideo.granjow.net/pngGUI/newProject.png</image>
```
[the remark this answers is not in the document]
muelli: I've removed them.

net.granjow.slowmovideo.yaml
muelli: deleted

net.granjow.slowmovideo.yaml
muelli: ah, that applies to Wireshark as well, I filed a bug to not forget: https://github.com/flathub/org.wireshark.Wireshark/issues/93

cmake.diff
```diff
@@ -0,0 +1,15 @@
...
+index 38ff055..b1b8867 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -44,6 +44,10 @@ else ()
+ 	set(PROJECT_VERSION_PATCH "0")
+ endif ()
+ 
++set(PROJECT_VERSION_MAJOR "0")
++set(PROJECT_VERSION_MINOR "6")
++set(PROJECT_VERSION_PATCH "0")
```
hfiguiere, line 11: why is this necessary?

net.granjow.slowmovideo.yaml
muelli: sure. but copying a file is easier than locating the definition in the file and maintaining indentation.

net.granjow.slowmovideo.yaml
muelli: so I *believe* that *some* folder needs to be shared, because the app creates folders which, AFAIK, doesn't work as expected when selecting a path through the portal. I thought "public" was a good choice because users may be more okay with leaking their data of that folder to malicious apps or maintainers than their potentially more private "videos" folder.

cmake.diff
muelli: because of the bug mentioned in the manifest: https://github.com/slowmoVideo/slowmoVideo/issues/200

hfiguiere: so why not instead make sure it take the "else" branch? You talk about being easy on maintenance elsewhere, but make a patch that makes it harder since it would need to be updated each time if upstream never fix the issue.

muelli: > so why not instead make sure it take the "else" branch?

I can't answer that question. I just wanted to make it work with little to no knowledge about cmake. I'm happy if you produce a better patch!

hfiguiere:
To validate your appdata for flathub use the flatpak `org.freedesktop.appstream-glib`.

JakobDev:
> To validate your appdata for flathub use the flatpak org.freedesktop.appstream-glib.

It is the the best to build the whole App with org.flatpak.Builder instead of the host flatpak-builder. It is also better to check your Appstream with both: appstream-glib (to see if it can be uploaded to Flathub) and appstreamcli (to see if the Appstream is valid). appstream-glib let a lot if invalid Appstream files pass which results in broken Appstream data on the Website and Software Centers., which is what the most people don't want. Also Flathub will switch to appstreamcli soon.

net.granjow.slowmovideo.yaml
muelli: I agree. I put xdg-videos now.

muelli:
this fails because of a git submodule.
what's the best practice to replace `git@` URLs in git submodules?

hfiguiere:
using a `git:` URL in a git submodule is WRONG.  (it's also deprecated on github anyway)

Now you can disable submodules and check them out manually by adding the proper sources. It's more maintenance thought.

barthalion:
bot, build net.granjow.slowmovideo

barthalion:
```
install: cannot stat 'net.granjow.slowmovideo.appdata.xml': No such file or directory
```

muelli:
which is weird, because it exists: https://github.com/flathub/flathub/pull/3280/files#diff-7b11d4b110a2655e7a4f8b9ee656cb267df62afab75bc210dcf721f5bf540912R74

is the CWD different in the post-install phase?
And.. it works locally (flatpak-builder 1.0.11).

net.granjow.slowmovideo.yaml
```diff
@@ -0,0 +1,75 @@
...
+          sha256: d6a69aa4a775a074662685b9bc2b07b88fb867f19d939b03ad281571b62f2620
+
+  - ffmpeg.yaml
+
+  - name: slomoVideo
+    buildsystem: cmake-ninja
+    builddir: true
+    post-install:
+        - install -D -ma+r /app/share/icons/AppIcon.png /app/share/icons/hicolor/256x256/apps/AppIcon.png
+        - install -D -ma+r net.granjow.slowmovideo.appdata.xml /app/share/appdata/net.granjow.slowmovideo.appdata.xml
```
hfiguiere, line 56:
```suggestion
        - install -D -m644 ../net.granjow.slowmovideo.appdata.xml /app/share/appdata/net.granjow.slowmovideo.appdata.xml
```

You said `builddir: true`, post-install is done from it.

barthalion:
bot, build net.granjow.slowmovideo

nbenitez:
bot, build net.granjow.slowmovideo

nbenitez:
Is something blocking this?

net.granjow.slowmovideo.yaml
```diff
@@ -0,0 +1,75 @@
+app-id: net.granjow.slowmovideo
+runtime: org.kde.Platform
+runtime-version: '5.15-21.08'
```
hfiguiere, line 3:
```suggestion
runtime-version: '5.15-22.08'
```

it's out now.

net.granjow.slowmovideo.yaml
```diff
@@ -0,0 +1,75 @@
...
+sdk: org.kde.Sdk
+
+command: slowmoUI
+rename-icon: AppIcon
+rename-desktop-file: slowmoUI.desktop
+finish-args:
+  - --device=dri
+  - --socket=x11
```
hfiguiere, line 11: missing share=ipc

hfiguiere (changes requested):

nbenitez:
It seems I cannot push to [muelli:slomo](https://github.com/muelli/flathub/tree/slomo):
`ERROR: Permission to muelli/flathub.git denied to nbenitez`

@muelli it has to be you to add those changes :-)

hfiguiere:
bot, build net.granjow.slowmovideo

hfiguiere:
build failed  because submodules seems to use `git:` URL. This is doomed to failed. No flathub specific.

hfiguiere:
upstream fixed in https://github.com/slowmoVideo/slowmoVideo/commit/c2e4df05fcc6d7dcd75705e08c5a858ecd3ad298

barthalion:
bot, build net.granjow.slowmovideo

barthalion:
Sorry for the noise here, I was investigating how to fix builds getting stuck on failure.

muelli:
bot, build net.granjow.slowmovideo

net.granjow.slowmovideo.yaml
muelli: unfortunately, it doesn't compile with the new compiler :(
https://github.com/flathub/flathub/pull/3280/commits/3f83be108e76b5c7297e89cb22c6e63cadcdec3a
Maybe https://github.com/slowmoVideo/slowmoVideo/commit/2a743063e38af6b4c747064243fc1f9b431d8c0e helps.
That's something to look at in the future.

Granjow:
@muelli Thanks for asking! I've looked into migrating to Qt6 today, but to get slowmoVideo up-to-date again would require a few days of work which I don't have, also I'm not actively using slowmoVideo anymore for years now. I'm really sorry for that, but life and priorities change. I'll still give advice where I can, and maybe one day I'll be back on development, but I'd guess (without googling) there must be many other/better/newer apps for slow-motion videos nowadays.
Nevertheless, it's an honour for me that you are working on slowmoVideo and I really feel sorry that I cannot give you the appropriate support.

PR #3294, opened 2022-06-27
Initial MAME package

jflatt:


org.mamedev.MAME.metainfo.xml
```diff
@@ -0,0 +1,25 @@
...
+    <category>Emulator</category>
+  </categories>
+  <url type="homepage">https://www.mamedev.org</url>
+  <launchable type="desktop-id">org.mamedev.MAME.desktop</launchable>
+  <releases>
+    <release version="0.244" date="2022-05-25"/>
+  </releases>
+  <provides>
+    <binary>mame</binary>
+  </provides>
```
hfiguiere, line 24: missing `content_rating` and `screenshots`

org.mamedev.MAME.yaml
```diff
@@ -0,0 +1,72 @@
...
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --socket=wayland
+  - --allow=bluetooth
+modules:
+  - name: mame
+    buildsystem: simple
+    build-options:
+      no-debuginfo: true
+      strip: true
```
hfiguiere, line 22: why? Debug symbols are provided in a separate package, so there is no reason to do that.

org.mamedev.MAME.yaml
```diff
@@ -0,0 +1,72 @@
...
+      path: mame32x32.png
+    - type: file
+      path: mame48x48.png
+    - type: file
+      path: mame64x64.png
+    - type: file
+      path: mame128x128.png
+    - type: file
+      path: mame256x256.png
+    - type: file
```
hfiguiere, line 46: these icons aren't square. Also only 128x128 is required.

org.mamedev.MAME.yaml
```diff
@@ -0,0 +1,72 @@
...
+      path: org.mamedev.MAME.desktop
+    - type: file
+      path: org.mamedev.MAME.metainfo.xml
+    - type: patch
+      path: 0.244-ptrdiff_t.patch
+    - type: patch
+      path: use-system-lib-flac.patch
+    - type: patch
+      path: ldopts.patch
+    post-install:
```
hfiguiere, line 56: this is no reason to have this in a post-install given it's a `simple` buildsystem.

org.mamedev.MAME.yaml
```diff
@@ -0,0 +1,72 @@
...
+      - cp COPYING /app/bin/
+      - cp uismall.bdf /app/bin/
+      - mkdir -p /app/share/mame/
+      - cp -r artwork/ /app/share/mame/
+      - cp -r bgfx/ /app/share/mame/
+      - cp -r ctrlr/ /app/share/mame/
+      - mkdir -p /app/share/mame/docs/
+      - cp -r docs/legal/ /app/share/mame/docs/
+      - cp -r docs/man/ /app/share/mame/docs
```
hfiguiere, line 65: is there any viewer for these docs? Othrewise they will simply not be accesssible.

org.mamedev.MAME.yaml
```diff
@@ -0,0 +1,72 @@
...
+separate-locales: false
+finish-args:
+  - --device=all
+  - --persist=.mame
+  - --filesystem=host:ro
+  - --share=network
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --socket=wayland
```
hfiguiere, line 15: I'll assume you tested this on Wayland.

hfiguiere (changes requested):
I am actually suprised that neither icons nor .desktop files are upstream. (I consider this being a bug)

org.mamedev.MAME.metainfo.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>org.mamedev.MAME</id>
+  <metadata_license>CC0-1.0</metadata_license>
```
hfiguiere, line 4: no project license?

hfiguiere:
bot, build org.mamedev.MAME

org.mamedev.MAME.yaml
hfiguiere: The build failed with :

```
error: /srv/buildbot/worker/build-x86_64-2/build/builddir/export/share/icons/hicolor/32x32/apps/org.mamedev.MAME.png is not a valid icon: Expected a square icon but got: 32x11
```

org.mamedev.MAME.metainfo.xml
jflatt: Okay.  Unclear on what is optional and what is required

org.mamedev.MAME.yaml
jflatt: This seems like an unusual requirement, but easy one to fix

org.mamedev.MAME.yaml
jflatt: Okay, moved file copying under build-commands

org.mamedev.MAME.yaml
jflatt: They will be accessible through the host filesystem, no?  Probably best to leave legal disclaimers in the package

org.mamedev.MAME.yaml
jflatt: Yes!

org.mamedev.MAME.metainfo.xml
jflatt: Added

org.mamedev.MAME.yaml
hfiguiere: all icons are square. I don't see that being unusual.

org.mamedev.MAME.yaml
hfiguiere: no they won't be, not in any trivial way. Let alone manpages.

org.mamedev.MAME.yaml
```diff
@@ -0,0 +1,72 @@
+id: org.mamedev.MAME
+runtime: org.kde.Platform
+runtime-version: 5.15-21.08
+sdk: org.kde.Sdk
+command: mame
+separate-locales: false
+finish-args:
+  - --device=all
+  - --persist=.mame
+  - --filesystem=host:ro
```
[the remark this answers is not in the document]
jflatt: The application supports somewhere around 3TB worth of ROMs, CD images, hard drive images, etc. between all sorts of different arcades, game consoles, and computer systems.  On top of that, it makes use of screenshots, advertising flyers, audio samples, and more.  It's fairly common, and has built-in support, for these to be split between filesystems and even removable media

org.mamedev.MAME.yaml
jflatt: I've tested only in KDE, but it has no problem centering the rectangular images

org.mamedev.MAME.yaml
jflatt: I dont understand.  I can do
`man ~/.local/share/flatpak/app/org.mamedev.MAME/current/active/files/share/mame/docs/man/mame.6`
but this is from flatpak-builder output.  Are you saying the flathub contents will be any different?

hfiguiere: if `man mame` doesn't produce a manpage then that's what I mean.

There is strictly no point installing them. There is no mechanism to make them available.

org.mamedev.MAME.metainfo.xml
hfiguiere: there can only be one `project_license`

barthalion:
bot, build org.mamedev.MAME

barthalion:
```
org.mamedev.MAME:                  AppData problem: tag-invalid : <project_license> was duplicated
```

org.mamedev.MAME.metainfo.xml
hfiguiere: you can add multiple licenses separated by `OR` or `AND` in project license, as specified by spdx.

hfiguiere: ok.

nedrichards:
bot, build org.mamedev.MAME

jflatt:
bot, build org.mamedev.MAME

jflatt:
bot, build org.mamedev.MAME

jflatt:
> @jflatt This is fantastic, thanks for creating a Flatpak of MAME that can be used on devices like the Steam Deck where the normal repository is not available!
> 
> I have tested the build and found some issues though, more specifically it seems like some systems don't work. I haven't done extensive testing but at least Bally Astrocade and TI-99 don't work, but other MESS systems like Apple II and Apple IIGS work fine.
> 
> As well it seems as if this build does a resolution switch on startup while the repository release of MAME in Kubuntu 22.04 does not, so I assume the latter runs in borderless windowed mode and the Flatpak build doesn't? Or maybe this behavior is controlled by some configuration file in Kubuntu rather than being baked into the binary itself?

I tried Bally Astrocade and TI99/4A and they both worked fine.  What issues are you seeing with them?  The only thing I can think of is these systems both require BIOS files, so be sure you have them.  Also this is the newest version of MAME so be sure the BIOS files you have are the known good files for this version.  Try running from Konsole ''flatpak run org.mamedev.MAME -verifyroms astrocde'.  You should be able to launch from Konsole using 'flatpak run org.mamedev.MAME -verbose' to see if there's any error output.

I'm also not seeing a resolution switch on my Steam Deck, though I don't think I know what that would even look like?  Maybe the resolution is related to your first issue.  I have mine configured for bgfx with the crt-geom-deluxe shaders.  Maybe you can try that setup and see if you notice a difference?  This is standard MAME, there shouldn't be anything special about the build.

leonstyhre:
The Bally Astrocade ROM is fine according to the -verifyroms flag, there is another issue and I think I know what it is.

When I run MAME installed via the repository it works fine, for example this command:
`/usr/games/mame -rompath /home/leon/ROMs/astrocde astrocde -cart astrobat`

But when using the Flatpak like this it fails:
`flatpak run org.mamedev.MAME -rompath /home/leon/ROMs/astrocde astrocde -cart astrobat`

This is the error message shown:
```
Unable to find the BGFX path bgfx, please install it or fix the bgfx_path setting to use the BGFX renderer.

"astrobat" approximately matches the following
supported software items (best match first):

Fatal error: Device Bally Astrocade Cartridge Slot load (-cartridge astrobat) failed: No such file or directory
```

I found a way to fix it, which is to add the file extension, like this:
`flatpak run org.mamedev.MAME -rompath /home/leon/ROMs/astrocde astrocde -cart astrobat.zip`

This launches the game correctly, but that should not be required and it causes issues when executed via ES-DE as I use the basename to launch MAME standalone and it has always worked with all other MAME releases on all operating systems I tested on. So I think this is a bug.

As for the resolution switch it makes MAME take around 2-3 second to display on my monitor and when existing the game it takes the same amount of time to display the desktop again. With the repository release of MAME this is instantaneous. This is consistent with behavior I've seen before with fullscreen mode vs. borderless window fullscreen mode. Maybe you could try it on a regular display running Xorg instead of Wayland and see if you get the same behavior?

Thanks :)

leonstyhre:
Actually, after some additional testing I've found that it's the -rompath flag that doesn't work correctly, the reason it worked to add the .zip extension was that I was located in the ROMs/astrocde directory when running the command so I guess MAME looked inside the current working directory for the game file I specified.

So if I go to another directory and run the following (same) command then the game will NOT start:
`flatpak run org.mamedev.MAME -rompath /home/leon/ROMs/astrocde astrocde -cart astrobat.zip`

And of course this won't work either:
`flatpak run org.mamedev.MAME -rompath /home/leon/ROMs/astrocde astrocde -cart astrobat`

Both of these commands do however work correctly when I use the repository release of MAME.

jflatt:
> Actually, after some additional testing I've found that it's the -rompath flag that doesn't work correctly, the reason it worked to add the .zip extension was that I was located in the ROMs/astrocde directory when running the command so I guess MAME looked inside the current working directory for the game file I specified.
> 
> So if I go to another directory and run the following (same) command then the game will NOT start: `flatpak run org.mamedev.MAME -rompath /home/leon/ROMs/astrocde astrocde -cart astrobat.zip`
> 
> And of course this won't work either: `flatpak run org.mamedev.MAME -rompath /home/leon/ROMs/astrocde astrocde -cart astrobat`
> 
> Both of these commands do however work correctly when I use the repository release of MAME.

There is a lot to be configured within the mame.ini file, including the rompath, I'd suggest you start there.

I'm a little confused on how you have your filesystem set up, but I would recommend keeping software list media in their own directory, aside from the baseline MAME roms.  For example I have my MAME roms in /mnt/roms/ and software list items in /mnt/sl/.  To use astrobat as an example it looks like:
```
/mnt/roms/astrocde.zip
/mnt/sl/astrocde/astrobat.zip
```
If the game is listed in the software lists and your paths are set correctly you should be able to run just
`flatpak run org.mamedev.MAME astrocde astrobat`
There's no need for the -cart option, as the software list has the data to know which carts, slots, etc. are used.

I'm guessing your repository release is including some pre-configured mame.ini file.

On the resolution switching issue, you might be able to run MAME with the -verbose and -window commands and see where in the initialization it's hanging up on.  So far I haven't seen any issues, even on an Xorg desktop.

leonstyhre:
Sorry I should clarify, I'm the maintainer/developer of the EmulationStation Desktop Edition (ES-DE) frontend and this setup is not so much for my personal use as it's intended for something that will work across all types of operating systems and setups.

The problem is that I need to keep the configuration as simple as possible, I can't expect users to edit the mame.ini file and create separate directories for software lists and such.

Anyway, I think I've found the issue now. You're right that the repository-installed MAME release has a mame.ini file which is located under /etc/mame. If I comment out the "hashpath" option I get the same error with this release as with your Flatpak build. I see that you bundle the hash files inside your Flatpak, but you don't seem to provide a default mame.ini file so I guess MAME simply does not know where to look for the hash files?

I could fix this by adding the absolute path to the ROM after the -cart option instead of the basename, such as the following:
`flatpak run org.mamedev.MAME -rompath /home/leon/ROMs/astrocde astrocde -cart /home/leon/ROMs/astrocde/astrobat.zip`

But I'm hesitant to do this as I think the Flatpak setup is currently not behaving correctly.

As for the resolution switching there are no settings in the bundled mame.ini file that affects this and comparing the verbose output of the repository release and your Flatpak build shows that they are very similar. There are apparently fewer audio drivers in your release but apart from that the only relevant data that differs are these lines (repository release above, Flatpak release below):

```
Available videodrivers: x11 wayland KMSDRM dummy 
Available videodrivers: x11 wayland dummy 

SDL/OpenGL defines: SDL_COMPILEDVERSION=2020 USE_OPENGL=1 
SDL/OpenGL defines: SDL_COMPILEDVERSION=2016 USE_OPENGL=1 

OpenGL: AMD Radeon RX 6600 XT (dimgrey_cavefish, LLVM 13.0.1, DRM 3.42, 5.15.0-41-generic)
OpenGL: AMD DIMGREY_CAVEFISH (DRM 3.42.0, 5.15.0-41-generic, LLVM 12.0.1)
```

I would be very surprised if these are indicating the issue, if you can even call it that. I mean the emulator still works, it's just a small annoyance to wait a couple of seconds every time I start and exit a game as the monitor re-syncs the resolution.

jflatt:
> Anyway, I think I've found the issue now. You're right that the repository-installed MAME release has a mame.ini file which is located under /etc/mame. If I comment out the "hashpath" option I get the same error with this release as with your Flatpak build. I see that you bundle the hash files inside your Flatpak, but you don't seem to provide a default mame.ini file so I guess MAME simply does not know where to look for the hash files?

That's right, no default mame.ini, too many variations to include anything sensible
 
> I could fix this by adding the absolute path to the ROM after the -cart option instead of the basename, such as the following: `flatpak run org.mamedev.MAME -rompath /home/leon/ROMs/astrocde astrocde -cart /home/leon/ROMs/astrocde/astrobat.zip`
> 
> But I'm hesitant to do this as I think the Flatpak setup is currently not behaving correctly.

There exists a -hashpath command line argument also.  As far as file paths go, the Flatpak is behaving exactly as I expect it.

> Edit: And using the -window switch is not applicable as it's then no longer fullscreen mode. Or in other words, using the -window option makes MAME start instantaneously both with the repository-installed MAME and your Flatpak build.

I see, the idea of using -window was to be able to watch the Konsole window in the background to see if any line in the verbose output was taking the longest amount of time.  I suppose you can allow X11Forwarding in ssh to watch the console output and see if that indicates any area of slowdown.  Doing that on the Steam Deck is going to take some effort though as the system is all locked down.  What -video option are you using?  Even using the `-video soft` setting starts up immediately for me.

leonstyhre:
I tried the -hashpath option and it works, the following runs the game correctly:
`flatpak run org.mamedev.MAME -hashpath /app/share/mame/hash -rompath /home/leon/ROMs/astrocde astrocde -cart astrobat
`

But the problem is I can't use this as ES-DE could be run on any number of operating systems that have the hashes at different locations. 

I think this hashpath really needs to be set inside the Flatpak, I mean it's probably the only setting you would need to set in a bundled mame.ini file and since the hashes are kept inside the Flatpak directory structure you're certain that the setting will always work. The workaround of using the ROM path directly only worked on Bally Astrocade, not on the TI-99 for example so that was not a usable solution.

If you're not willing to include a mame.ini file with this setting populated then this Flatpak release of MAME will be crippled compared to all other releases. In this case the user would need to create a mame.ini file and place it in the correct location which is adding a lot of unnecessary work for them. I'm sure the EmuDeck guys can create the file and place it at the correct location as part of the MAME Flatpak installation and I will talk with them about that, but it really seems unnecessary as you would be able to solve this more elegantly by include the setting inside the Flatpak.

hfiguiere:
Remember that by default, this will use Wayland. It only uses X11 if there is no Wayland session on the host.

leonstyhre:
As for the video switching this delay is caused by the GPU driver and monitor resyncing the resolution, it's not anything that happens inside MAME. I had exactly the same problem with ES-DE before I switched to borderless windowed mode. Starting with -video soft makes no difference. This behavior is likely device-dependent as well which may be why you don't see it on some devices like the Steam Deck. Whether it also happens on Wayland or not I'm not sure of.

leonstyhre:
I can confirm that creating a mame.ini file in the following location:
`~/.mame/mame.ini`

and including the following line:
`hashpath /app/share/mame/hash`

resolves the issue and the Flatpak release of MAME works correctly.

I'm not going to repeat myself, except to say that it's an extra step and it would be nice if it could be avoided by including it in the Flatpak instead.

jflatt:
> I think this hashpath really needs to be set inside the Flatpak, I mean it's probably the only setting you would need to set in a bundled mame.ini file and since the hashes are kept inside the Flatpak directory structure you're certain that the setting will always work. The workaround of using the ROM path directly only worked on Bally Astrocade, not on the TI-99 for example so that was not a usable solution.

There's also samples, ctrlr, cheat, crosshair, plugins, language.  The samples directory is not fixed like the hash directory, it contains included files and it's expected the user will also have additional files.  So would that be included in a default mame.ini or not?  The filesystem works the same way, whatever your workaround was it should work the same for either Astrocade or TI99.

> If you're not willing to include a mame.ini file with this setting populated then this Flatpak release of MAME will be crippled compared to all other releases. In this case the user would need to create a mame.ini file and place it in the correct location which is adding a lot of unnecessary work for them. I'm sure the EmuDeck guys can create the file and place it at the correct location as part of the MAME Flatpak installation and I will talk with them about that, but it really seems unnecessary as you would be able to solve this more elegantly by include the setting inside the Flatpak.

The package released at mamedev.org does not come with a default mame.ini.  This flatpak is behaving in the same way as the official release, the documentation from mamedev.org matches, and there are no hidden surprises like you had with your "repository release".  There are just way too many options, and going through the work and properly setting up the mame.ini file is necessary to get it running properly.  All that being said, almost everything in MAME can be run through the command line, isn't your frontend passing these options through?  This same work would need to be done to configure the frontend?  Or are you making assumptions and not allowing users to configure the hash directory?

As far as your video switching, are you sure you have all of your paths set up?  There may be some unnecessary file scanning being done by MAME if it can't find the files it's looking for, that could cause some slowdown.

leonstyhre:
I think the hashes are a bit different because you bundle them inside the Flatpak but you make them unreachable by default since there is no hashpath set. Also keep in mind that the required setting is /app/share/mame/hash which is very unintuitive for a user since that directory is not available on their filesystem but is something internal in the Flatpak sandbox.

I don't know about the other settings you mention, everything seems to run fine if the hashpath is set but I have not done extensive testing so more settings may need to be defined as well.

I just downloaded the official 0.245 release for Windows from mamedev.org, unzipped it and it worked perfect in ES-DE without any customization or configuration, this Flatpak release is the first time I've run into this problem and I test on many different operating systems and architectures.

You can for sure pass any command line option you want from ES-DE but the whole point of my application is to make it easy to use so normally a user should never need to customize something like that, and as mentioned they don't need to with MAME on any of the other operating systems and releases I tested on, it just works.

I'm not going to keep arguing about this, if you're unwilling to implement the hashpath setting in the Flatpak then the EmuDeck script will implement a workaround by creating the mame.ini file with the required setting and I will add documentation to the ES-DE user guide that the Flatpak MAME release is broken and needs additional configuration to launch all games.

I just don't think it's very user friendly and I'm sure there will be a number of bug reports about this issue as it's incredibly hard to figure out what's wrong for most users who don't have a deep understanding on the internals of how MAME and the Flatpak sandbox works.

The video switching issue happens within the GPU driver as MAME tells it to switch to fullscreen mode, it has nothing to do with any paths, it's a window/video mode issue. But it's not very important, it's just a minor annoyance and not a showstopper like the hashpath issue just discussed.

jflatt:
> I just downloaded the official 0.245 release for Windows from mamedev.org, unzipped it and it worked perfect in ES-DE without any customization or configuration, this Flatpak release is the first time I've run into this problem and I test on many different operating systems and architectures.

Yes, that's more due to the flatpak common convention of using `/app/share/<appname>` for extra addon files.  The official release can find those additional files as they're packaged into the application path.

> I'm not going to keep arguing about this, if you're unwilling to implement the hashpath setting in the Flatpak then the EmuDeck script will implement a workaround by creating the mame.ini file with the required setting and I will add documentation to the ES-DE user guide that the Flatpak MAME release is broken and needs additional configuration to launch all games.

Don't get me wrong, I'm not arguing at all, I want to help get this all working.  But let's be clear, this is not broken and not a bug, and should not impede this flathub pull request.

leonstyhre:
At least you are aware of the issue now, and it's of course up to you if you want to fix it or not.

I still consider this a defect as the MAME Flatpak will simply not work with some systems unless the user creates a configuration file which points to a directory inside the Flatpak sandbox.

If you decide to fix this later it will be much appreciated, for now we'll make sure that EmuDeck implements the workaround and I'll document the issue accordingly in the ES-DE user guide.

barthalion:
ES-DE could provide its own mame.ini, couldn't it? That sounds like it would make most sense for something that utilizes multiple external programs.

bot, build org.mamedev.MAME

leonstyhre:
@barthalion No ES-DE only acts as a frontend, it never configures any emulators or includes configuration files for them. The Flatpak release of MAME will require the user to make a custom mame.ini file as described in an earlier reply above or the emulator will be crippled. Note that this is not a specific problem for ES-DE, every frontend will have the same problem and running MAME standalone will also have the same issue.

I have included instructions for this in the ES-DE user guide and I think EmuDeck will generate the config file for the user. But any Linux user that does but use EmuDeck will run into the problem. With that said it doesn't break every system that MAME emulates, only the ones requiring hash files.

org.mamedev.MAME.yaml
```diff
@@ -0,0 +1,61 @@
...
+  - --socket=wayland
+  - --allow=bluetooth
+modules:
+  - name: mame
+    buildsystem: simple
+    build-options:
+      no-debuginfo: true
+      strip: true
+    build-commands:
+      - make #-j ${FLATPAK_BUILDER_N_JOBS}
```
barthalion, line 25: Was it broken when multithreaded?

jflatt: Yes, build failed.  The MAME build system was just recently reworked, I am planning to try to re-enable when the next version releases in the next few days

barthalion:
It doesn't start for me unless I disable Wayland socket:

```
~ ❱ flatpak run org.mamedev.MAME
Ignoring MAME exception: Error: BGFX video does not work with wayland videodriver. Please change either of the options.
Fatal error: Error: BGFX video does not work with wayland videodriver. Please change either of the options.
```

Either drop it from `finish-args`, or it should be configured to work on Wayland out of the box.

hfiguiere:
but that's not default.

jflatt:
> but that's not default.

default configuration, out of the box

hfiguiere:
if you have to pass an argument on the command line it is not default.

jflatt:
> if you have to pass an argument on the command line it is not default.

My hardware (Steam Deck) does not require an argument from the command line.  Am I to configure this flatpak for one single platform, or best to leave it unconfigured?

barthalion:
This is only because Steam Deck does not use Wayland by default. If you don't want to support Wayland out of the box, drop it from `finish-args`. It's not an unrealistic expectation to have a functional software right after installation.

jflatt:
bot, build org.mamedev.MAME

jflatt:
bot, build org.mamedev.MAME

Sterophonick:
I seem to be having trouble writing configuration files with this Flatpak. Every time I attempt to save I get a `** Failed saving <config>.ini **` message onscreen.

jflatt:
> I seem to be having trouble writing configuration files with this Flatpak. Every time I attempt to save I get a `** Failed saving <config>.ini **` message onscreen. This doesn't happen with the native build, so I'm not sure what's up with it whether or not it's my end or a problem with the package.

It's a problem with the package

barthalion:
A fixable problem, though?

jflatt:
bot, build org.mamedev.MAME

jflatt:
bot, build org.mamedev.MAME

jflatt:
> I seem to be having trouble writing configuration files with this Flatpak. Every time I attempt to save I get a `** Failed saving <config>.ini **` message onscreen. This doesn't happen with the native build, so I'm not sure what's up with it whether or not it's my end or a problem with the package.

Can you try this again now on the newer test version?

Sterophonick:
> > I seem to be having trouble writing configuration files with this Flatpak. Every time I attempt to save I get a `** Failed saving <config>.ini **` message onscreen. This doesn't happen with the native build, so I'm not sure what's up with it whether or not it's my end or a problem with the package.
> 
> Can you try this again now on the newer test version?

Problem still persists unfortunately.

Sterophonick:
> > I seem to be having trouble writing configuration files with this Flatpak. Every time I attempt to save I get a `** Failed saving <config>.ini **` message onscreen. This doesn't happen with the native build, so I'm not sure what's up with it whether or not it's my end or a problem with the package.
> 
> Can you try this again now on the newer test version?

Problem still persists unfortunately.

jflatt:
> > > `** Failed saving <config>.ini **` message onscreen
> Problem still persists unfortunately.

If you run it with -showconfig what is the entry for inipath? Does this pop up when you do something specific?
Anything else in the error message?  Can you run with -verbose and put the output somewhere like pastebin?

Sterophonick:
> > > > `** Failed saving <config>.ini **` message onscreen
> > > > Problem still persists unfortunately.
> 
> If you run it with -showconfig what is the entry for inipath? Does this pop up when you do something specific? Anything else in the error message? Can you run with -verbose and put the output somewhere like pastebin?

Output of `-showconfig`:
https://pastebin.com/1hMCAuMt
`inipath` is `$HOME/.mame;/app/share/mame/ini`

Output of `-verbose`, booting up MAME and then immediately attempting to save settings:
https://pastebin.com/4vM7G1fP

Nothing else is shown other than the message.

leonstyhre:
@jflatt Thanks for including a working mame.ini file in the Flatpak! I can confirm that it resolves the issue discussed previously and for example Bally Astrocade games now run correctly from inside ES-DE when using this Flatpak release. So no custom ~/.mame/mame.ini file is required any longer.

jflatt:
> Output of `-showconfig`: https://pastebin.com/1hMCAuMt `inipath` is `$HOME/.mame;/app/share/mame/ini`
That configuration looks like it's all just the default values.  I believe it's the homepath = . that's killing it.  It's most likely trying to write files into the current working directory which is read-only.  If you don't have it, create a ~/.mame directory and copy in the mame.ini from the Files tab here in this pull request

Sterophonick:
This works now, up to a point. By default it works fine. However, for my use case I would really appreciate it if we could get the ability to symbolically link configuration data onto another drive in the even I need to reset it. This doesn't seem to be working despite me overriding the files filesystem. 

If it's not a priority, I get it. Other than that, works great.

PR #3333, opened 2022-07-17
[WIP] Add uk.co.pcem_emulator.pcem

AsciiWolf:

- [ ] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://github.com/sarah-walker-pcem/pcem/issues/164
Desktop icon file is from [here](https://github.com/X-m7/flathub/blob/pcem/uk.co.pcem_emulator.pcem.svg). I have already contacted the repo owner for permission to ship the icon with this Flatpak (the Flatpak itself is however not based on that outdated repo version).
The upstream project does not have any icon/desktop files and is most likely not interested in shipping them.


uk.co.pcem_emulator.pcem.yaml
```diff
@@ -0,0 +1,62 @@
...
+      - type: git
+        url: https://github.com/sarah-walker-pcem/pcem
+        commit: 13f53a25687be71ee3ee8482b1a90b1b0aa64fb6
+      - type: script
+        commands:
+          - autoreconf -fiv
+        dest-filename: autogen.sh
+      - type: script
+        commands:
+          - "[ -d ~/.pcem/logs ] || mkdir ~/.pcem/logs && [ -d ~/.pcem/roms ] || mkdir ~/.pcem/roms"
```
AsciiWolf, line 49: This is a workaround for a bug causing PCem to crash when the `~/.pcem/logs` directory does not exist. The `~/.pcem/roms` directory does not seem to get created automatically and since the directory is mentioned in AppStream metadata, let's create it automatically as well.

AsciiWolf:
bot, build uk.co.pcem_emulator.pcem

AsciiWolf:
bot, build uk.co.pcem_emulator.pcem

AsciiWolf:
Tested and seems to work fine. :-)

uk.co.pcem_emulator.pcem.yaml
JakobDev: You could just use `mkdir -p`. This creates a directory when it's not exists and does nothing when it exists, so you don't need to check if the directory exists before running `mkdir`.

This Bug also should be fixed Upstream, so that is Workaround is only needed until the next release.

AsciiWolf: Thanks! I did not report it yet because it seems to happen only when the `--persist=.pcem` is used and I am not quite sure why it happens.

uk.co.pcem_emulator.pcem.yaml
```diff
@@ -0,0 +1,63 @@
...
+  - name: pcem
+    rm-configure: true
+    config-opts:
+      - --enable-release-build
+      - --enable-networking
+      - --enable-alsa
+    sources:
+      - type: git
+        url: https://github.com/sarah-walker-pcem/pcem
+        commit: 13f53a25687be71ee3ee8482b1a90b1b0aa64fb6
```
gasinvein, line 43: This commit has a tag, `v17`. If you add it, shallow clone will work.

AsciiWolf:
To be honest, I thought that 86Box had slower development and was less stable that PCem. But it looks like that I was wrong. Thanks for pointing this out, I will probably rewrite this Flatpak to use 86Box instead. :-)

gasinvein:
Tbf I'm not very familiar with either and used them only briefly, but at least 86Box works relatively well with native Wayland, and supports `XDG_DATA_DIRS` for finding ROMs (i.e. more flatpak-friendly overall).

AsciiWolf:
I am trying to rewrite the Flatpak to use 86Box, however I am not sure whether it is a good idea after all. The GUI is better, it has more hardware support and more features, but on the other hand it has its own problems. And its XDG_DATA_DIRS support is actually a mess.

gasinvein:
I've built 86Box flatpak previously ([manifest](https://github.com/gasinvein/flatpak-apps-staging/tree/net._86box._86Box/master)), it seemed to work rather well; the only problem I recall was that only software renderer (the default) worked on native Wayland.

uk.co.pcem_emulator.pcem.yaml
AsciiWolf: Done, thanks!

AsciiWolf:
Thanks! It looks better than my manifest. :-) I am however not sure about redistributing the roms.

gasinvein:
Well, you can just drop that module, or turn it into extra-data.

AsciiWolf:
Ok, 86Box Flatpak is ready and seems to be much better than PCem! Thanks for the tip! :-)

AsciiWolf:
Replaced by #3337. @gasinvein Feel free to review. :-)

PR #3337, opened 2022-07-19
Add net._86box._86Box

AsciiWolf:

- [ ] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://github.com/86Box/86Box/issues/2483


AsciiWolf:
bot, build net._86box._86Box

net._86box._86Box.yaml
```diff
@@ -0,0 +1,103 @@
...
+        commands:
+          - mkdir -p ${XDG_DATA_HOME}/86Box/roms
+          - mkdir -p ${XDG_CONFIG_HOME}/86Box
+          - exec 86Box --vmpath ${XDG_CONFIG_HOME}/86Box "$@"
+    post-install:
+      - install -Dm755 86Box.sh ${FLATPAK_DEST}/bin/86Box.sh
+      - sed -i 's/net.86box.86Box/net._86box._86Box/g' src/unix/assets/net.86box.86Box.metainfo.xml
+      - install -Dm644 src/unix/assets/net.86box.86Box.metainfo.xml ${FLATPAK_DEST}/share/metainfo/net._86box._86Box.metainfo.xml
+      - sed -i 's/Exec=86Box/Exec=86Box.sh/g' src/unix/assets/net.86box.86Box.desktop
+      - sed -i 's/net.86box.86Box/net._86box._86Box/g' src/unix/assets/net.86box.86Box.desktop
```
gasinvein, line 76: Use `desktop-file-edit` instead.

net._86box._86Box.yaml
```diff
@@ -0,0 +1,103 @@
+app-id: net._86box._86Box
+runtime: org.kde.Platform
+runtime-version: '5.15-21.08'
+sdk: org.kde.Sdk
+command: 86Box.sh
+finish-args:
+  - --share=network
+  - --share=ipc
+  # 86Box has issues on Wayland
```
gasinvein, line 9: IMO these issues are pretty minor. Are you sure it worth sacrificing wayland's security advantages and having that red "unsecure" label in GNOME Software?

net._86box._86Box.yaml
```diff
@@ -0,0 +1,103 @@
...
+          type: git
+      - type: script
+        dest-filename: 86Box.sh
+        commands:
+          - mkdir -p ${XDG_DATA_HOME}/86Box/roms
+          - mkdir -p ${XDG_CONFIG_HOME}/86Box
+          - exec 86Box --vmpath ${XDG_CONFIG_HOME}/86Box "$@"
+    post-install:
+      - install -Dm755 86Box.sh ${FLATPAK_DEST}/bin/86Box.sh
+      - sed -i 's/net.86box.86Box/net._86box._86Box/g' src/unix/assets/net.86box.86Box.metainfo.xml
```
hfiguiere, line 73: if you install it with the original name and then `rename-appdata-file`, I do believe it will do this too.

hfiguiere (changes requested):

hfiguiere:
```
CMake Error at src/codegen/CMakeLists.txt:26 (message):
  Dynarec is incompatible with target platform arm64
```

Also lack of screenshot in the AppStream file.

AsciiWolf:
> Also lack of screenshot in the AppStream file.

I was not able to find any good 86Box screenshot. I will make one myself.

net._86box._86Box.yaml
AsciiWolf: You're right, but I think that this is a cleaner solution.

net._86box._86Box.yaml
AsciiWolf: That's a good idea, thanks!

net._86box._86Box.yaml
AsciiWolf: I think it is worth it. The 86Box window had some weird glitches and other issues (for example when moving the window or switching to fullscreen) when using native Wayland on my GNOME Wayland session. And if only a software renderer works on Wayland, it is quite a big issue.

AsciiWolf:
> Another idea regarding the ROMs: we can put them into extension.

That's not a bad idea, but I think that I will keep them as extra-data for now.

gasinvein:
> > Another idea regarding the ROMs: we can put them into extension.
> 
> That's not a bad idea, but I think that I will keep them as extra-data for now.

It's your call, but bear in mind that if you decide to move ROMs to an extension later, there will be a transition period during which the app will be without ROMs (after the extension point is added and before the extension is accepted to flathub).

net._86box._86Box.yaml
gasinvein: Well, the software renderer is the default anyway (I suppose it's the most stable one?). This should be fixed in 86Box in the end.

AsciiWolf:
> It's your call, but bear in mind that if you decide to move ROMs to an extension later, there will be a transition period during which the app will be without ROMs (after the extension point is added and before the extension is accepted to flathub).

That's a good point. Any chance you could write the extension (and suggest changes needed in the 86Box manifest)? I sadly don't have much experience with app extensions.

net._86box._86Box.yaml
AsciiWolf: Should be, but I think that it is a good idea to disable the Wayland socket until it actually gets fixed.

net._86box._86Box.yaml
AsciiWolf: Done, thanks!

net._86box._86Box.yaml
```diff
@@ -0,0 +1,103 @@
...
+  - name: rtmidi
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DRTMIDI_API_JACK=OFF
+    sources:
+      - type: archive
+        url: https://github.com/thestk/rtmidi/archive/refs/tags/5.0.0.tar.gz
+        sha256: c7923e4eee82b06c007435892cb2c3212d9007fa482c6b718943bda71c02c5a7
+
+  - name: fluidsynth
```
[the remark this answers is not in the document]
AsciiWolf: Done, thanks!

gasinvein:
Since the ROMs are just a single directory, this should be relatively simple. Just add something like this to the extension:
```yaml
add-extensions:
  net._86box._86Box.ROMs:
    directory: share/86Box/roms
```
And then you can create a `net._86box._86Box.ROMs` flatpak that will have its contents mounted at `/app/share/86Box/roms` in the app sandbox. So, to "build" the extension, you'll just need to put to `$FLATPAK_DEST/` of the extension. The extension manifest can look something like this:
```yaml
id: net._86box._86Box.ROMs
build-extension: true
runtime: net._86box._86Box
sdk: org.kde.Sdk//5.15-21.08
...
modules:
  - name: 86Box-roms
    buildsystem: simple
    build-commands:
     - |
       for dir in floppy hdd icons machines network printer scsi sound video; do
         ln -s extra/roms/$dir $dir
       done
    sources:
      ...
```

AsciiWolf:
Thanks! I will look into it in the upcoming days. Are extra-data also supported for extensions? I would like to keep the roms as extra-data if possible.

AsciiWolf:
> > Also lack of screenshot in the AppStream file.
> 
> I was not able to find any good 86Box screenshot. I will make one myself.

Done. It will be available when https://github.com/86Box/86Box/pull/2486 gets merged.

gasinvein:
> Are extra-data also supported for extensions?

Yes, it's how e.g. Nvidia GL extensions work.

AsciiWolf:
bot, build net._86box._86Box

net._86box._86Box.yaml
AsciiWolf: It looks like that the Vulkan renderer is not working also on x11.

gasinvein: Hmm, it works if ran with `--devel`, i.e. on Sdk instead of Platform. Looks like the app improperly tries to open (`dlopen()`?) some development library.

gasinvein:
Hmm, do we really need to rename the upstream ID `net.86box.86Box` to `net._86box._86Box`?

X-m7:
> Hmm, do we really need to rename the upstream ID `net.86box.86Box` to `net._86box._86Box`?

As per the [D-Bus naming conventions](https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names) (which the [Flatpak documentation](https://docs.flatpak.org/en/latest/conventions.html) points to):

> Each element must only contain the ASCII characters "[A-Z][a-z][0-9]_" and must not begin with a digit.

But then the Flathub [app requirements](https://github.com/flathub/flathub/wiki/App-Requirements#application-id) (which points to the above as well) also says:

> The Application ID should be a real URL of a domain that the app author has control over or where their app is hosted.

So I guess it's a question of whether "real URL" or "must not begin with a digit" should take precedence.

net._86box._86Box.yaml
```diff
@@ -64,6 +68,7 @@ modules:
           - mkdir -p ${XDG_CONFIG_HOME}/86Box
           - exec 86Box --vmpath ${XDG_CONFIG_HOME}/86Box "$@"
     post-install:
+      - mkdir -p ${FLATPAK_DEST}/share/86Box
```
gasinvein, line 71: Extension mount directory must exist in the app.
```suggestion
      - mkdir -p ${FLATPAK_DEST}/share/86Box/roms
```

AsciiWolf: Ah, I did not realize this. Thanks!

AsciiWolf:
bot, build net._86box._86Box

gasinvein:
Other then some nitpicks, LGTM. Note that it won't build until the app itself is on Flathub.

net._86box._86Box.yaml
Cacodemon345: 86Box's OpenGL and Vulkan renderers works just fine here with LabWC and with `wayland` platform. This is strictly a runtime issue as the application dlopens neither `libGL.so` nor `libOpenGL.so` on its own and explicitly avoids dynamically linking against Vulkan.

PR #3348, opened 2022-07-21
Add com.github.d4nj1.tlpui

d4nj1:


nedrichards:
bot, build com.github.d4nj1.tlpui

com.github.d4nj1.tlpui.yml
```diff
@@ -0,0 +1,34 @@
+app-id: com.github.d4nj1.tlpui
+command: tlpui
+runtime: org.gnome.Platform
+runtime-version: '3.38'
+sdk: org.gnome.Sdk
+finish-args:
+  - --socket=fallback-x11
```
nedrichards, line 7: you need --share=ipc as well if you're going to allow x11.

also - fallback-x11 is designed for use in a scenario where --socket=wayland is also used. Does this not support wayland?

nedrichards:
Is there a reason why the desktop file isn't part of the upstream project yet?

com.github.d4nj1.tlpui.yml
```diff
@@ -0,0 +1,34 @@
+app-id: com.github.d4nj1.tlpui
+command: tlpui
+runtime: org.gnome.Platform
+runtime-version: '3.38'
```
hfiguiere, line 4:
```suggestion
runtime-version: '42'
```

3.38 has been EOL for quite a while.

com.github.d4nj1.tlpui.yml
```diff
@@ -0,0 +1,34 @@
...
+sdk: org.gnome.Sdk
+finish-args:
+  - --socket=fallback-x11
+  - --talk-name=org.freedesktop.Flatpak
+  - --filesystem=host
```
hfiguiere, line 9: this is a very broad filesystem permission.

com.github.d4nj1.tlpui.desktop
```diff
@@ -0,0 +1,14 @@
+[Desktop Entry]
```
hfiguiere, line 1: This file is at https://github.com/d4nj1/TLPUI/blob/master/tlpui.desktop and doesn't need to be here.

hfiguiere (changes requested):
There is no appstream file

hfiguiere:
> Is there a reason why the desktop file isn't part of the upstream project yet?

It is https://github.com/d4nj1/TLPUI/blob/master/tlpui.desktop (12 month ago)

com.github.d4nj1.tlpui.desktop
hfiguiere: Also since you are upstream the desktop file is invalid. `Icon` MUST be an icon name not a path. `Exec` should be an executable name, possible with arguments, not a path.

hfiguiere:
About the appstream file I expect this one https://github.com/d4nj1/TLPUI/blob/master/build/AppImage/tlpui.appdata.xml to be used. It will need patches though as it needs `releases` and `content_rating` at least.

hfiguiere:
And the app-id in it is invalid.

com.github.d4nj1.tlpui.yml
d4nj1: Hi,
I created this config more than a year ago and only adopted to yaml notation. I guess some stuff has changed since then. I do not even find the option for x11-fallback in the docs anymore. https://docs.flatpak.org/en/latest/sandbox-permissions-reference.html

What do I have to add if would like to support both X11 and Wayland? Will it be auto discovered if I avoid adding `--socket=wayland` and `--socket=x11`?

com.github.d4nj1.tlpui.yml
d4nj1: I will test and upgrade to latest version.

com.github.d4nj1.tlpui.yml
d4nj1: Sorry to bother, it's actually documented here: https://docs.flatpak.org/en/latest/sandbox-permissions.html

com.github.d4nj1.tlpui.yml
d4nj1: Back in early 2021 there was no option to allow permissions on individual host files. Tlpui only needs read and write access to either `/var/run/host/etc/default/tlp` or `/var/run/host/etc/tlp.conf`depending on the TLP version.

Should the following work?
`- --filesystem=/var/run/host/etc/default/tlp:rw`
`- --filesystem=/var/run/host/etc/tlp.conf:rw`

Or is it still only folder paths like this then?
`- --filesystem=/var/run/host/etc:rw`

com.github.d4nj1.tlpui.desktop
d4nj1: If I remember correctly there were some issues when building Debian packages when not defined like this. I check if this is still the case.

* Regarding Exec:
It is allowed to have the full path but I will try to only mention the executable directly as you desire.
See https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-1.1.html#exec-variables 

* Regarding Icon:
Same as for Exec the full path can be used. I used it because I was unsure if the icons that are defined and applied to the current theme engine in Tlpui get recognized by Flatpak.
See https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-1.1.html#recognized-keys

I try to adopt to one `.desktop` file and update the PR accordingly.

d4nj1:
I will fix the mentioned issues and also try to apply a proper `appdata.xml` to match the specs.

Thanks for looking on the PR and providing instant feedback. That's awesome. :smiley:

d4nj1:
bot, build com.github.d4nj1.tlpui

d4nj1:
bot, build com.github.d4nj1.tlpui

d4nj1:
bot, build com.github.d4nj1.tlpui

d4nj1:
bot, build com.github.d4nj1.tlpui

d4nj1:
bot, build com.github.d4nj1.tlpui

com.github.d4nj1.tlpui.yml
```diff
@@ -0,0 +1,53 @@
...
+      - install -D tlpui/icons/themeable/hicolor/16x16/apps/tlpui.png /app/share/icons/hicolor/16x16/apps/com.github.d4nj1.tlpui.png
+      - install -D tlpui/icons/themeable/hicolor/32x32/apps/tlpui.png /app/share/icons/hicolor/32x32/apps/com.github.d4nj1.tlpui.png
+      - install -D tlpui/icons/themeable/hicolor/48x48/apps/tlpui.png /app/share/icons/hicolor/48x48/apps/com.github.d4nj1.tlpui.png
+      - install -D tlpui/icons/themeable/hicolor/64x64/apps/tlpui.png /app/share/icons/hicolor/64x64/apps/com.github.d4nj1.tlpui.png
+      - install -D tlpui/icons/themeable/hicolor/96x96/apps/tlpui.png /app/share/icons/hicolor/96x96/apps/com.github.d4nj1.tlpui.png
+      - install -D tlpui/icons/themeable/hicolor/128x128/apps/tlpui.png /app/share/icons/hicolor/128x128/apps/com.github.d4nj1.tlpui.png
+      - install -D tlpui/icons/themeable/hicolor/256x256/apps/tlpui.png /app/share/icons/hicolor/256x256/apps/com.github.d4nj1.tlpui.png
+      - install -D AppImage/com.github.d4nj1.tlpui.appdata.xml /app/share/metainfo/com.github.d4nj1.tlpui.metainfo.xml
+      - sed -i 's/Icon=tlpui/Icon=com.github.d4nj1.tlpui/g' tlpui.desktop
```
hfiguiere, line 36:
```suggestion
      - desktop-file-edit --set-icon=com.github.d4nj1.tlpui tlpui.desktop
```

But you could as well had all properly named upstream.

com.github.d4nj1.tlpui.yml
```diff
@@ -0,0 +1,53 @@
...
+      - install -D tlpui/icons/themeable/hicolor/16x16/apps/tlpui.png /app/share/icons/hicolor/16x16/apps/com.github.d4nj1.tlpui.png
+      - install -D tlpui/icons/themeable/hicolor/32x32/apps/tlpui.png /app/share/icons/hicolor/32x32/apps/com.github.d4nj1.tlpui.png
+      - install -D tlpui/icons/themeable/hicolor/48x48/apps/tlpui.png /app/share/icons/hicolor/48x48/apps/com.github.d4nj1.tlpui.png
+      - install -D tlpui/icons/themeable/hicolor/64x64/apps/tlpui.png /app/share/icons/hicolor/64x64/apps/com.github.d4nj1.tlpui.png
+      - install -D tlpui/icons/themeable/hicolor/96x96/apps/tlpui.png /app/share/icons/hicolor/96x96/apps/com.github.d4nj1.tlpui.png
+      - install -D tlpui/icons/themeable/hicolor/128x128/apps/tlpui.png /app/share/icons/hicolor/128x128/apps/com.github.d4nj1.tlpui.png
+      - install -D tlpui/icons/themeable/hicolor/256x256/apps/tlpui.png /app/share/icons/hicolor/256x256/apps/com.github.d4nj1.tlpui.png
+      - install -D AppImage/com.github.d4nj1.tlpui.appdata.xml /app/share/metainfo/com.github.d4nj1.tlpui.metainfo.xml
```
hfiguiere, line 35: the launchable is wrong in the appdata.

One way to do it clean is to install the .destkop file without renaming and adding a `rename-desktop-file`

d4nj1:
bot, build com.github.d4nj1.tlpui

d4nj1:
bot, build com.github.d4nj1.tlpui

com.github.d4nj1.tlpui.yml
barthalion: None will. Even with `host`, you will have no write access to `/var/run/host/etc`.

barthalion: That being said, you could limit it to `host-etc`.

barthalion:
```
~ 9.8s ❱ flatpak run com.github.d4nj1.tlpui
Portal call failed: Failed to start command: Failed to execute child process “tlp-stat” (No such file or directory)
  File "/app/bin/tlpui", line 5, in <module>
    from tlpui.__main__ import main
  File "/app/lib/python3.9/site-packages/tlpui/__main__.py", line 8, in <module>
    from . import settings
  File "/app/lib/python3.9/site-packages/tlpui/settings.py", line 21, in <module>
    tlpbaseversion = settingshelper.get_installed_major_minor_version()
  File "/app/lib/python3.9/site-packages/tlpui/settingshelper.py", line 29, in get_installed_major_minor_version
    return get_installed_tlp_version()[0:3]
  File "/app/lib/python3.9/site-packages/tlpui/settingshelper.py", line 21, in get_installed_tlp_version
    currentconfig = check_output(["tlp-stat", "-c"]).decode(sys.stdout.encoding)
  File "/usr/lib/python3.9/subprocess.py", line 424, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['tlp-stat', '-c']' returned non-zero exit status 1.
```

It should fail gracefully if TLP is not installed on the host.

d4nj1:
bot, build com.github.d4nj1.tlpui

barthalion:
bot, build com.github.d4nj1.tlpui

barthalion:
Still doesn't seem to be working for me.

com.github.d4nj1.tlpui.yml
d4nj1: Ok, then `host-etc` should be sufficient for me. The write to `/var/run/host/etc/default/tlp` or `/var/run/host/etc/tlp.conf`  happens anyways through a priviledged host copy-over-command from flatpak tmp-folder. At least that worked on my machine during development.

d4nj1:
@barthalion Do you expect a UI window to pop up if TLP dependecy is missing on system?

barthalion:
Well, it's a desktop app. From a perspective of a desktop user, I think it's a reasonable expectation to have an actually visible error without using a terminal.

com.github.d4nj1.tlpui.yml
d4nj1: There does not seem to exist a `host-usr`. How can this be achieved for `/usr` access?

d4nj1:
Yes, definitely.
I guess there is no default way to present those for flatpaks? Some desktop environments support this kind of behaviour to show  an error trace when a UI application fails.

d4nj1:
bot, build com.github.d4nj1.tlpui

barthalion:
> Yes, definitely. I guess there is no default way to present those for flatpaks? Some desktop environments support this kind of failure handling to show an error trace when a UI application fails.

Flatpak doesn't handle anything related to UIs apart from exposing X11 or Wayland socket. Checking host requisites is something the app should be doing, even if some DEs would do it automatically.

d4nj1:
bot, build com.github.d4nj1.tlpui

com.github.d4nj1.tlpui.yml
d4nj1: And also vor `/var` access?

d4nj1:
I think the `host:ro` is the least priviledge I need since there is no `host-usr` and `host-var` similar to the `host-etc`.

d4nj1:
bot, build com.github.d4nj1.tlpui

d4nj1:
I think the `host:ro` is the least priviledge I need since there is no `host-usr` similar to the `host-etc`.

barthalion:
Does it include a prompt for missing TLP on the host now?

d4nj1:
No, it does **not**.
I found a bug in the flatpak config in regards to the host existence for tlp and tlp-stat. Please be patient. I will report back once I think it's done.

d4nj1:
bot, build com.github.d4nj1.tlpui

d4nj1:
I tested on a newly installed system without TLP installed and the prompt appears as desired.

PR #3349, opened 2022-07-23
Adds Contour Terminal Emulator.

christianparpart:

### Dear Reviewers

One question I am having, so I had to manually install some nvidia platform flatpak package. This is a terminal emulator that is utilizing OpenGL 3.3 Core profile, and I wonder if I might need to specify anything else on top that might be required to avoid users to know that they might need to install some nvidia flatpak just to get Contour running.


christianparpart:
After some forth and back, I decided against using the existing domain name as App ID, because dashes are not allowed. So I do not own `contour.org` (but `contour-terminal.org`), but looking at the former's website, that will most likely never conflict with flathub packages anyways, and secondly, the app id `org.contour.Terminal` sounds very reasonable to me.

Is the appdata filename correct though?

Contour.appdata.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="runtime">
+  <id>org.contour.Terminal</id>
+  <project_license>Apache-2</project_license>
+  <name>Contour Terminal Emulator</name>
+  <summary>Contour is a modern and actually fast, modal, virtual terminal emulator, for everyday use. It is aiming for power users with a modern feature mindset.</summary>
+  <url type="homepage">https://github.com/contour-terminal/contour/</url>
```
hfiguiere, line 7: this is completely insuficient.

missing:
- releases
- screenshot
- content rating
- description

Contour.appdata.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="runtime">
```
hfiguiere, line 2: this is NOT a runtime.

org.contour.Terminal.yml
```diff
@@ -0,0 +1,92 @@
...
+runtime: org.kde.Platform
+runtime-version: '5.15'
+sdk: org.kde.Sdk
+command: contour
+finish-args:
+  - --filesystem=host
+  - --device=all          # we need access to at least DRI (for OpenGL) & PTY
+  - --share=ipc
+  - --socket=x11
+  - --socket=wayland
```
hfiguiere, line 12:
```suggestion
  - --socket=fallback-x11
  - --socket=wayland
```

assuming proper wayland support

org.contour.Terminal.yml
```diff
@@ -0,0 +1,92 @@
...
+  - --socket=pulseaudio
+  - --own-name=org.contour.Terminal
```
hfiguiere, line 14:
```suggestion
```
it's automatically granted

org.contour.Terminal.yml
```diff
@@ -0,0 +1,92 @@
...
+  - --talk-name=org.freedesktop.Flatpak
```
hfiguiere, line 15: why do you need this?

org.contour.Terminal.yml
```diff
@@ -0,0 +1,92 @@
+app-id: org.contour.Terminal
```
hfiguiere, line 1: `contour.org` redirect to forums about the Ford Contour. Is this a really a valid domain the developers control?

It doesn't look like it.

org.contour.Terminal.yml
```diff
@@ -0,0 +1,92 @@
...
+name: Contour Terminal Emulator
+runtime: org.kde.Platform
+runtime-version: '5.15'
```
hfiguiere, line 4: this runtime version is END OF LIFE
```suggestion
runtime-version: '5.15-21.08'
```

hfiguiere (changes requested):

hfiguiere:
bot, build org.contour.Terminal

hfiguiere:
also the appdata file is never installed. (has it been submitted upstream?)

org.contour.Terminal.yml
christianparpart: I prefer it to remain fallback-x11 for the time being, because I know that works more well. In fact, it's an OpenGL based app that is fully utilizing Qt APIs, and yet I sometimes have struggles and I even files one bug report to official Qt bug tracker due to some infinite recursion that I detected. 

So with this in mind, I think it should favor X11 until I am more confident.

org.contour.Terminal.yml
christianparpart: Because of sandbox privilege escalation. It's a terminal and I need to [escape from the sandbox](https://github.com/contour-terminal/contour/blob/4f565357720915402aca5ca63ca1fa812ed4741f/src/terminal/Process_unix.cpp#L170-L191).

org.contour.Terminal.yml
christianparpart: This is what I was trying to argue about in https://github.com/flathub/flathub/pull/3349#issuecomment-1193103621

I own the domain `contour-terminal.org`, which is forwarding to github pages (not yet set up => 404).

However, `contour-terminal` being part of the app id seems invalid, flatpak-builder was at least bailing out saying that. So I have to chose something I do not own, because it does not allow me to specify of what I own. When reading the docs, it did not seem like a hard requirement, so this is why i was giving this explanation above. :)

If you have better proposals for the App ID, I am all open :-)

Contour.appdata.xml
christianparpart: I was apparently looking at the wrong example. I now found https://freedesktop.org/software/appstream/docs/sect-Metadata-Application.html and will adapt that tonight. Thanks.

org.contour.Terminal.yml
hfiguiere: the thing it is currently NOT `fallback-x11`

hfiguiere:
> After some forth and back, I decided against using the existing domain name as App ID, because dashes are not allowed. So I do not own `contour.org` (but `contour-terminal.org`), but looking at the former's website, that will most likely never conflict with flathub packages anyways, and secondly, the app id `org.contour.Terminal` sounds very reasonable to me.
> 
> Is the appdata filename correct though?

You should re-read the documentation.

1. the domain must under the control of the developer. Which obviously it's not since this is not a Ford Contour. Seems obvious. It's not about whether it will conlfict or not. IT IS NOT YOURS.
2. If you re-read the document you'll see that `_` gets replaced by `-` to form a valid D-Bus name.

So no you chosen app-id is NOT reasonable.

hfiguiere:
And for the last question: the appdata filename only matter when installed, but it's not. There is no magic trick done it is simply ignored if you don't install it.

Now my take is that:
1. it should be upstream for a future release
2. it should be name properly, which it is not at the moment. While this is not a requirement on the source it's better.

org.contour.Terminal.yml
christianparpart: Okay, I've removed the wayland-socket then, and I assume that it'll then take the X11 protocol when on Wayland desktop, right?

christianparpart:
> also the appdata file is never installed. (has it been submitted upstream?)

Actually, I am the upstream. I was looking at other PRs and realized they're providing an appdata.xml file, so to me it did look like that's the way it should be done on GUI apps. 

Thanks for helping me completing the appdata.xml file regardless of that. And I'll be including it in future releases. I guess I've to add another patch such that it'll also install the appdata.xml into the app's image at `/usr/share/metainfo`, right? (I'll do so ASAP).

christianparpart:
> it should be name properly, which it is not at the moment. While this is not a requirement on the source it's better.

@hfiguiere I am in control of `contour-terminal.org`  but that does contain a dash, so I cannot use that as app-id. What do you recommend me as name here? I'm open for suggestions, if `org.contour.Terminal` isn't sufficient. :)

org.contour.Terminal.yml
hfiguiere: absolutely not. This is not what I was suggesting.

hfiguiere:
> @hfiguiere I am in control of `contour-terminal.org` but that does contain a dash, so I cannot use that as app-id. What do you recommend me as name here? I'm open for suggestions, if `org.contour.Terminal` isn't sufficient. :)

As I wrote:
https://github.com/flathub/flathub/pull/3349#issuecomment-1193168053

hfiguiere:
> > also the appdata file is never installed. (has it been submitted upstream?)
> 
> Actually, I am the upstream. I was looking at other PRs and realized they're providing an appdata.xml file,

If you looked at other PR you would have also seen I often ask the same thing. Also you would see how they actually install it.

An ideal submission only contain a manifest.

Also from the documentation you DID acknowledge have read: https://github.com/flathub/flathub/wiki/App-Requirements#appstream
You definitey CAN NOT install in `/usr` as it is read only.

Contour.appdata.xml
```diff
@@ -1,9 +1,99 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<component type="runtime">
+
+<component type="desktop">
   <id>org.contour.Terminal</id>
+  <metadata_license>MIT</metadata_license>
   <project_license>Apache-2</project_license>
+  <content_rating>none</content_rating>
```
hfiguiere, line 7: this is an invalid value

Contour.appdata.xml
```diff
@@ -1,9 +1,99 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<component type="runtime">
+
+<component type="desktop">
   <id>org.contour.Terminal</id>
+  <metadata_license>MIT</metadata_license>
```
hfiguiere, line 5: the documentation recommend `CC0-1.0`

org.contour.Terminal.yml
```diff
@@ -1,18 +1,16 @@
...
-runtime-version: '5.15'
+runtime-version: '5.15-21.08'
 sdk: org.kde.Sdk
 command: contour
 finish-args:
   - --filesystem=host
   - --device=all          # we need access to at least DRI (for OpenGL) & PTY
   - --share=ipc
   - --socket=x11
-  - --socket=wayland
```
hfiguiere: this is not was I was suggesting.

org.contour.Terminal.yml
hfiguiere: that's what it's supposed to do. Unfortunate but it's Qt.

Contour.appdata.xml
```diff
@@ -0,0 +1,99 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<component type="desktop">
+  <id>org.contour.Terminal</id>
+  <metadata_license>MIT</metadata_license>
+  <project_license>Apache-2</project_license>
```
hfiguiere, line 6: This is an invalid SPDX license identifier

https://spdx.dev/licenses/

hfiguiere (changes requested):

Contour.appdata.xml
christianparpart: I've fixed that as I realized it won't let me finish flatpak-builder even locally. Thanks.

christianparpart:
I think I've applied all recommendations now @hfiguiere except maybe open still could be to also rename the file names (of the appdata.xml and .yml) with respect to the AppStream's App ID. 

That should then be something like: `org.contour_terminal.Terminal.appdata.xml` and `org.contour_terminal.Terminal.yml`?

hfiguiere:
The manifest filename has to match the app-id.

org.contour.Terminal.yml
```diff
@@ -0,0 +1,100 @@
...
+      #   path: /home/christianparpart/projects/contour
+      #   skip:
+      #     - target
+      #     - _deps
+      # }}}
+
+      # {{{ Add and make sure appdata.xml is installed.
+      # NOTE: These two items should be removed with the next release of Contour.
+      - type: patch
+        path: install-appdata.xml.patch
```
hfiguiere, line 93: you can just add a `post-install` rule to install it.

flatpak-tweaks.patch
```diff
@@ -0,0 +1,13 @@
...
+index acef652..1bd3a15 100644
+--- a/cmake/FilesystemResolver.cmake
++++ b/cmake/FilesystemResolver.cmake
+@@ -12,7 +12,7 @@ else()
+     set(USING_BOOST_FILESYSTEM FALSE PARENT_SCOPE)
+     set(USING_BOOST_FILESYSTEM FALSE)
+     message(STATUS "[FilesystemResolver]: Using standard C++ filesystem API")
+-    if(APPLE OR "${CMAKE_SYSTEM}" MATCHES "Linux")
++    if(APPLE)
+         set(FILESYSTEM_LIBS "stdc++fs")
```
hfiguiere, line 11: BTW this is a bug upstream whereas it assumes a specific version of ubuntu with an old version of gcc.

christianparpart:
> Also from the documentation you DID acknowledge have read: https://github.com/flathub/flathub/wiki/App-Requirements#appstream
> You definitey CAN NOT install in `/usr` as it is read only.

I meant it as an example installation path -- in the `CMakeLists.txt` I am not hard-coding any install paths but use correct install prefixes. It just happens to end up in /usr/ on system installs (not so much when customize nor via flatpak, I know), and therefore I was taking that as well known example when asking for the path structure (regardless of the install root) without trying to give overly-long but 100% correct statements, especially since I was asking from a smartphone, so I was trying to keep my sentence concise (again, regardless of installation prefix). 

> The manifest filename has to match the app-id.

The manifest file-name I will be changing on next push.

> An ideal submission only contain a manifest.

This is the initial package and I want to use an existing (latest stable) release for that, possibly including one or two patches. The next release on flatpak shall not contain any patches as everything is done on the upstream side already (which I am 100% in control of, so I don't see any brocker here). :)

christianparpart:
> you can just add a post-install rule to install it.

I'm not sure what the cleaner solution might have been, but I've changed that now, too, to be installed via post-install rather than CMake.

christianparpart:
While content-wise, being totally ready with that PR and I think I did not accidentally skip anything in the review process of this PR, what bugs me the most is that `org.contour_terminal.Terminal` looks really bad to me in terms of esthetics and I do not want to give users the impression of repeating themselves.

The concern I am having over the documentation and the review-process here is in fact, that the documentation states:

> Ownership of {vendor}.{tld} in the domain name system guarantees uniqueness of IDs. 

This (to my understanding) sounds not like its mandatory to "own" that domain (but that it would simply guarantee uniqueness), in case it does exist and is owned by the App-Id creator. While also not liking that I have to actually go out and buy a domain just to publish an FOSS product that is hosted on Github to flathub feels wrong to me, too.

Now, that domain (contour.org) to the rDNS inspired original App-Id is owned by someone hosting a forum to a car type. I am fine with that and certainly, how high are the chances that there will be a FOSS app by that guy related to that forum such that it would justify not to use that rDNS notation as App Id for the Contour terminal.

Since that chance is highly unlikely, and given the fact that when looking at my own sys installs ...metainfo/ dir, where I see plenty of clashes with respect to ownership but also to the use of hyphens rather than underscores, I wonder why to mandate that now.

So I either stick with `org.contour.Terminal` as App-Id (which is unique relatively speaking but w/o mapped ownership of a real domain behind) or I could see myself going for `org.contourterminal.SOMETHING) (with SOMETHING maybe being App or so?).

What would be best fitting (without including any underscores)?

org.contour_terminal.Terminal.yml
```diff
@@ -0,0 +1,100 @@
...
+  - name: contour
+    buildsystem: cmake-ninja
+    config-opts:
+      - "-DCONTOUR_INSTALL_TOOLS=ON"
+      - "-DCONTOUR_BLUR_PLATFORM_KWIN=ON"
+      - "-DCONTOUR_TESTING=OFF"
+      - "-DLIBTERMINAL_TESTING=OFF"
+    # NOTE: This post-install should be removed with the next release of Contour, when appdata.xml is auto-installed.
+    post-install:
+      - install -p -m644 org.contour.Terminal.appdata.xml /app/share/metainfo
```
hfiguiere, line 26: as I said befor **it must be installed with the right name**. The simple course of action would be for the original file to have the right name.
BTW this also install the file as `/app/share/metainfo` because the destination is a file name unless `-t/--target` is used.

christianparpart: > **must be installed with the right name**

The "right name" is what [I am trying to figure out](https://github.com/flathub/flathub/pull/3349#issuecomment-1193308461) now. When that's figured out, I'll go and do the appropriate Id & file renames.

hfiguiere: `${FLATPAK_ID}.metainfo.xml`. That pretty clear in the documentation all over

(.appdata.xml works too)

`$FLATPAK_ID` is the shell environment that susbtitute with the app-id. Since you look at other repositories you would have noticed that too.

hfiguiere:
Still need a `rename-destkop-file` for the destkop file.

org.contourterminal.App.appdata.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <component type="desktop">
-  <id>org.contour_terminal.Terminal</id>
+  <id>org.contourterminal.App</id>
```
hfiguiere, line 4: NO,
The app-id was settled.

hfiguiere: To reiterate, it doesn't fit the requirements.

`.App` is just bad taste. 


`io.github.contour_terminal.contour` would be acceptable too.

org.contourterminal.App.yml
```diff
@@ -23,7 +23,7 @@ modules:
       - "-DLIBTERMINAL_TESTING=OFF"
     # NOTE: This post-install should be removed with the next release of Contour, when appdata.xml is auto-installed.
     post-install:
-      - install -p -m644 org.contour.Terminal.appdata.xml /app/share/metainfo
+      - install -p -m644 org.contourterminal.App.appdata.xml /app/share/metainfo
```
hfiguiere, line 26: you also ignored my comment about this line.

org.contourterminal.App.appdata.xml
christianparpart: I don't like the underscore, so I am trying to get rid of it before it is too late. This was my whole point of my rather biggish post this morning. :)

org.contourterminal.App.yml
christianparpart: I did not "ignore" it. I was reading it, paused and pondered what exactly you meant. Are you referring to the "right name" argument (which I reasoned about already) or the statement " BTW this also install the file as /app/share/metainfo because the destination is a file name unless -t/--target is used."?
I don't see the problem in that install command as it's going to install the source (the file) into that directory. I was re-reading the man-page's description and I don't see any problem. I am aware of `-t` but I do not see the need for using the 3rd syntax form of install when the 2nd one suffices, too. In a local install call  it all went well. Or am I misunderstanding something here?

christianparpart:
bot, build org.contourterminal.Contour

christianparpart:
bot, build org.contourterminal.Contour

org.contourterminal.Contour.desktop
```diff
@@ -0,0 +1,196 @@
+# TODO
+# - [ ] make Exec prefix configurable via @CMAKE_INSTALL_PREFIX@/bin
+# - [ ] include it as install target
+
+[Desktop Entry]
```
hfiguiere, line 5: there is a .desktop file upstream it should be used instead.

org.contourterminal.Contour.yml
```diff
@@ -0,0 +1,106 @@
...
+      - "-DCONTOUR_INSTALL_TOOLS=ON"
+      - "-DCONTOUR_BLUR_PLATFORM_KWIN=ON"
+      - "-DCONTOUR_TESTING=OFF"
+      - "-DLIBTERMINAL_TESTING=OFF"
+    # NOTE: This post-install should be removed with the next release of Contour, when metainfo.xml is auto-installed.
+    post-install:
+      - install -p -m755 -d /app/share/metainfo
+      - install -p -m644 org.contourterminal.Contour.metainfo.xml /app/share/metainfo
+      - install -p -m644 org.contourterminal.Contour.desktop /app/share/applications
+      - rm /app/share/applications/contour.desktop
```
hfiguiere, line 29: we have `rename-desktop-file` for that, You should use this file and not include one here.

org.contourterminal.Contour.yml
```diff
@@ -0,0 +1,106 @@
...
+      - mv /app/share/pixmaps/{contour,org.contourterminal.Contour}.png
```
hfiguiere, line 30: this icon location is invalid. And also we have `rename-icon` for that (but given it is in the wrong place, it's moot)

hfiguiere:
The app-id is still unacceptable.

org.contourterminal.Contour.yml
christianparpart: I used `rename-desktop-file` initially, but then I realized that `flatpak-builder` locally is generating a warning during desktop file validation stage, so I fixed it by providing a fixed one and having the right name in place right away. I could insist in using rename-desktop-file (as used in a previous commit) but then apply a patch. But this solution was more clean to me.

org.contourterminal.Contour.desktop
christianparpart: I am upstream. It's a copy from the upstream repo with the problems as raised by the validation step fixed. If it passes here (including the AppStream XML file), i.e. if flathub CI won't complain anymore, I'll gladly move it over and create a new release out of it.

christianparpart:
bot, build org.contourterminal.Contour

christianparpart:
bot, build org.contourterminal.Contour

christianparpart:
> The app-id is still unacceptable.

With all respect, let's be as clear as possible, even if you might repeat yourself, where I am apologizing ahead of time, but maybe I am skipping here something by accident. What exactly is "unacceptable" in that App Id?

The previous suggestion you said is bad taste, which IMHO is subjective. What is it now?

org.contourterminal.Contour.yml
```diff
@@ -0,0 +1,91 @@
+app-id: org.contourterminal.Contour
```
barthalion, line 1: Just clarify, contourterminal.org is an actual domain you own?

flatpak-tweaks.patch
```diff
@@ -0,0 +1,176 @@
+commit a50acc67521dd1a3f28ad523c05d5328835a496f
```
barthalion, line 1: Bear with me – this will be part of the next release?

org.contourterminal.Contour.yml
```diff
@@ -0,0 +1,91 @@
...
+name: Contour Terminal Emulator
+runtime: org.kde.Platform
+runtime-version: '5.15-21.08'
+sdk: org.kde.Sdk
+command: contour
+finish-args:
+  - --filesystem=host
```
barthalion, line 8: Why does it need host access if it launches shell on the host with flatpak-spawn?

barthalion (commented):
```
~ 4s ❱ flatpak run org.contourterminal.Contour
[error] Superfluous config key found: profiles.main.bold_is_bright
QSocketNotifier: Can only be used with threads started with QThread
QWaylandGLContext: failed to create EGLContext, error=3009
QOpenGLWidget: Failed to create context
QWaylandGLContext: failed to create EGLContext, error=3009
QOpenGLWidget: Failed to create context
QWaylandGLContext: failed to create EGLContext, error=3009
qt.qpa.backingstore: composeAndFlush: QOpenGLContext creation failed
```

It doesn't start for me on Fedora 36 Silverblue.

org.contourterminal.Contour.yml
christianparpart: I bought it this weekend exclusively for the App-Id, yes.

org.contourterminal.Contour.yml
christianparpart: This is a left-over from back in the day, when I didn't know I will be needing flatpak-spawn to create a non-sandboxed process. I've dropped that in a new commit (but didn't test, as I'm on mobile / commute) just yet. :)

christianparpart:
> `~ 4s ❱ flatpak run org.contourterminal.Contour [error] Superfluous config key found: profiles.main.bold_is_bright QSocketNotifier: Can only be used with threads started with QThread QWaylandGLContext: failed to create EGLContext, error=3009 QOpenGLWidget: Failed to create context QWaylandGLContext: failed to create EGLContext, error=3009 QOpenGLWidget: Failed to create context QWaylandGLContext: failed to create EGLContext, error=3009 qt.qpa.backingstore: composeAndFlush: QOpenGLContext creation failed`

I had the same issue after switching to NVIDIA graphics card (AMD previously) and resuming my work on getting flatpak support it. I had to manually install nvidia flatpak packages. That fixed it. But I did not figure out if I should express that in the flatpak yml file. Is there a more straight forward way to avoid users from that?

flatpak-tweaks.patch
christianparpart: Here it is: https://github.com/contour-terminal/contour/pull/767

barthalion:
In that case, I think it's about Wayland. The terminal launches fine if I pass `--nosocket=wayland`.

christianparpart:
> In that case, I think it's about Wayland. The terminal launches fine if I pass `--nosocket=wayland`.

I have been stripping it down to X11 communication exclusively. Since this works on Wayland (using X11 protocol), too, that should be fine for now.

christianparpart:
bot, build org.contourterminal.Contour

christianparpart:
@barthalion Hey. I know this ticket is closed already and it's all fine so far. Sorry for pinging you directly but I think that's the most straight way to progress in this particular case, as Google search didn't really help much so far.

So. I have received a steam deck last week and realized their /usr/include directory is completely wiped (possibly more, but that is actually the main breaker). So using the Steam Deck to develop is kind of hard to impossible without messing with the readonly filesystem and forgetting every manual modification right after a system update.

My idea now is to provide a Terminal Emulator version (Contour) that does NOT escape the sandbox, intentionally. Because for flatpak actually there do exist Sdk runtimes that might be the actual savior to all those that want a native (e.g. C/C++) dev environment on their Steam Deck. With Contour and Flatpak I could provide that.

I realized that I might now have two options. Either submit a new package to flathub with the minor changes to it (i.e. not escaping sandbox via flatpak-spawn, sharing home, and using org.kde.Sdk as runtime rather than org.kde.Platform);
or I somehow create a second flatpak branch. that can be installed via `flatpak install flathub org.contourterminal.Contour//SteamDeck` (or similar branch name). 

I am not sure what solution would be best (1st or 2nd). For the branch-solution however it seems I found zero living examples I could look at to see how I could provide two different versions of my App, utilizing the flatpak's branch feature.

Would you mind giving me some hints or advice on that matter? I think that would be a great addition to the Deck from a dev's user point of view.

PR #3359, opened 2022-07-28
add  org.sil.Bloom

StephenMcConnel:

- [x] I am using only the minimal set of permissions. *(If not, please explain each non-standard permission.)*  I think I explained in comments in the manifest file.
- [x] Any additional patches or files have been submitted to the upstream projects concerned. *(If not, explain why.)*  Patches to projects owned by sil.org are pending, and may or may not be made to the current development work.


libpalaso.yml
```diff
@@ -0,0 +1,128 @@
...
+        install output/Release/net461/SIL.Windows.Forms.Keyboarding.dll /app/libpalaso
+        install output/Release/net461/SIL.Windows.Forms.Keyboarding.pdb /app/libpalaso
+        install output/Release/net461/SIL.Windows.Forms.WritingSystems.dll /app/libpalaso
+        install output/Release/net461/SIL.Windows.Forms.WritingSystems.pdb /app/libpalaso
+        install output/Release/net461/SIL.WritingSystems.dll /app/libpalaso
+        install output/Release/net461/SIL.WritingSystems.pdb /app/libpalaso
+sources:
+    - type: git
+      url: https://github.com/sillsdev/libpalaso.git
+      branch: Bloom-5.2
```
hfiguiere, line 56:
```suggestion
```

Never a branch. this will break the build.

libpalaso.yml
```diff
@@ -0,0 +1,128 @@
...
+      commit: c4fd324c4024640ba6652e5184d66bf7962672d2
+      disable-shallow-clone: true
+    - type: file
+      url: https://dist.nuget.org/win-x86-commandline/v6.2.1/nuget.exe
```
hfiguiere, line 60: so this is x86 only?
(we don't support x86)

org.sil.Bloom.yml
```diff
@@ -0,0 +1,647 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+app-id: org.sil.Bloom
+
+runtime: org.gnome.Platform
+runtime-version: "41"
```
hfiguiere, line 26:
```suggestion
runtime-version: "42"
```

org.sil.Bloom.yml
```diff
@@ -0,0 +1,647 @@
...
+    # via a mechanism like DropBox).  Users need to load pictures, sound files,
+    # and videos from arbitrary locations in the filesystem. Users also need to
+    # be able to export books in various formats to arbitrary locations.
+    - --filesystem=home
+    - --filesystem=host
+    - --filesystem=xdg-data
+    # Bloom writes temporary folders and files to /tmp, to open in a browser to
+    # show the user, such as creating PDFs, ePUBs, or BloomPub ebooks.  Use the
+    # same /tmp as the host, so the host programs can find the files.
+    - --filesystem=/tmp
```
hfiguiere, line 55: there is a lot of redundancy in these very very very broad filesystem permissions.

org.sil.Bloom.yml
```diff
@@ -0,0 +1,647 @@
...
+    # libGL needs this
+    - --device=all
```
hfiguiere, line 57: actually it only needs
```suggestion
    # libGL needs this
    - --device=dri
```

org.sil.Bloom.yml
```diff
@@ -0,0 +1,647 @@
...
+    # debugging verbosely.  I need all the help I can get!
+    - -vv
```
hfiguiere, line 59: not sure what you think this does, but I doubt it does what you think it does.

org.sil.Bloom.yml
```diff
@@ -0,0 +1,647 @@
...
+command: bloom
+
+modules:
+    - name: BloomDesktop
+      buildsystem: simple
+      build-options:
+          prepend-path: /app/bin
```
hfiguiere, line 66:
```suggestion
```

this is already set.

org.sil.Bloom.yml
```diff
@@ -0,0 +1,647 @@
...
+              # These are needed to compile libpalaso on Windows, but aren't really used even there I suspect.
+              find ${BLOOMLIB} -name 'AtkSharp.dll' -print0 | xargs -0 rm -f
+              find ${BLOOMLIB} -name 'CairoSharp.dll' -print0 | xargs -0 rm -f
+              find ${BLOOMLIB} -name 'GdkSharp.dll' -print0 | xargs -0 rm -f
+              find ${BLOOMLIB} -name 'GioSharp.dll' -print0 | xargs -0 rm -f
+              find ${BLOOMLIB} -name 'GLibSharp.dll' -print0 | xargs -0 rm -f
+              find ${BLOOMLIB} -name 'GtkSharp.dll' -print0 | xargs -0 rm -f
+              find ${BLOOMLIB} -name 'PangoSharp.dll' -print0 | xargs -0 rm -f
+              # make up for deficient mono mapping
+              ln -s /usr/lib/x86_64-linux-gnu/libdl.so.2 ${INSTALLATION_PREFIX}/lib/libdl.so
```
hfiguiere, line 160: so you are x86_64 only ?

What about aarch64?

org.sil.Bloom.yml
```diff
@@ -0,0 +1,647 @@
...
+              install -d ${INSTALLATION_PREFIX}/lib/ffmpeg
+      cleanup:
+          - "/include"
+          - "/lib/*/include"
+          - "*.a"
+      sources:
+          - type: git
+            dest: BloomDesktop
+            url: https://github.com/BloomBooks/BloomDesktop.git
+            branch: Version5.2
```
hfiguiere, line 171: you need a commit or a tag, not a branch. NEVER a branch.

org.sil.Bloom.yml
```diff
@@ -0,0 +1,647 @@
...
+                # absent.
+                - "/lib/mono/msbuild"
+                - "/lib/mono/xbuild"
+            sources:
+                - type: file
+                  url: http://linux.lsdev.sil.org/ubuntu/pool/main/m/mono5-sil-msbuild/mono5-sil-msbuild_15.8+xamarinxplat.2018.07.31.22.43-0xamarin5+ubuntu1804b1_all.deb
+                  sha256: 758274d8bc438f325fc5fc808c0482ab50eb2bdaf02f61747888e190760753de
+          - name: libgdiplus
+            build-options:
+                prefix: /app
```
hfiguiere, line 348:
```suggestion
```

this is already set.

org.sil.Bloom.yml
```diff
@@ -0,0 +1,647 @@
...
+                  commit: b9c581ba87da89651d5ec56c5da0960ab739a63a
+            modules:
+                - name: giflib
+                  buildsystem: simple
+                  build-commands:
+                      # Skip building documentation
+                      - perl -pi -e 's#doc/Makefile##' configure.ac
+                      - ./autogen.sh --prefix=/app
+                      - make
+                      - make install
```
hfiguiere, line 369: you should use `buildsystem: autotools` and just have a `shell` source for the perl incantation. (or sed)

org.sil.Bloom.yml
```diff
@@ -0,0 +1,647 @@
...
+                - name: chmlib
+                  cleanup:
+                      - "/include"
+                      - "*.a"
+                  sources:
+                      - type: git
+                        url: https://github.com/jedwing/CHMLib
+                        # Latest commit, from 2009-07-03.
+                        commit: 2bef8d063ec7d88a8de6fd9f0513ea42ac0fa21f
+                      - type: script
```
hfiguiere, line 407: I think you meant 
```suggestion
                      - type: shell
```

otherwise you are missing the dest-filename.

org.sil.Bloom.yml
```diff
@@ -0,0 +1,647 @@
...
+          - name: wget
+            buildsystem: autotools
+            cleanup:
+                # Don't ship.
+                - "*"
+            sources:
+                - type: archive
+                  url: https://ftp.gnu.org/gnu/wget/wget-1.20.3.tar.gz
+                  sha512: e8b82b40e270296228094a78d47f81580bdbdea9e6b93fd61b37dccb39430aeb9bda5397dc53a31c952a61629383c7e2a8c8abf414c8a4dd369af6ecf2717e6c
+          - name: dbus-glib
```
hfiguiere, line 433: this module is part of the shared-modules. you should use that instead.

org.sil.Bloom.yml
```diff
@@ -0,0 +1,647 @@
...
+          - name: libtidy-sil
+            buildsystem: cmake
+            config-opts:
+                - -DBUILD_SHARED_LIB:BOOL=TRUE
+            cleanup:
+                - "*.a"
+            sources:
+                - type: git
+                  url: https://github.com/BloomBooks/tidy-html5.git
+                  branch: debian
```
hfiguiere, line 479:
```suggestion
```

hfiguiere (changes requested):

hfiguiere:
bot, build org.sil.Bloom

libpalaso.yml
StephenMcConnel: nuget.exe is a C# assembly that runs just fine on a x64 Linux box using mono.  I suspect it was built for AnyCPU despite the win-x86-commandline in the path.

libpalaso.yml
StephenMcConnel: line removed.  the commit is the more important (and permanent) value.

hfiguiere:
something failed to download.

However the log output is garbbled.

libpalaso.yml
hfiguiere: ah ok.

org.sil.Bloom.yml
StephenMcConnel: I admit I don't fully understand these settings.  I know that users need access to their home directories, the /tmp directory, and various other places such as mounted USB drives and network shared folders.

org.sil.Bloom.yml
StephenMcConnel: removed it

org.sil.Bloom.yml
StephenMcConnel: removed

org.sil.Bloom.yml
StephenMcConnel: We've never tried targeting aarch64.  99% of our users are on Windows x86_64 machines and the rest on Linux x86_64 machines.  I'd rather not try anything fancy with this submission when I have no way of testing the result.

org.sil.Bloom.yml
StephenMcConnel: changed to use the commit.

org.sil.Bloom.yml
StephenMcConnel: removed line

org.sil.Bloom.yml
StephenMcConnel: done.

org.sil.Bloom.yml
StephenMcConnel: removed.  the commit was there already.

org.sil.Bloom.yml
hfiguiere: things is `aarch64` is built by default. Also I thought the idea beihnd .Net was to be arch independent?

org.sil.Bloom.yml
StephenMcConnel: changed this as suggested

org.sil.Bloom.yml
hfiguiere: Also something that is mean to improve access not runing on one of the least expensive Linux computer (RaspberryPi 400) is a bit sad.

org.sil.Bloom.yml
StephenMcConnel: changed to shell

org.sil.Bloom.yml
StephenMcConnel: done.

barthalion:
At least this part isn't:

```
Failed to download sources: module geckofx: The requested URL returned error: 502
```

Let's see if it's a hiccup. bot, build org.sil.Bloom

org.sil.Bloom.yml
StephenMcConnel: done.

org.sil.Bloom.yml
StephenMcConnel: I've added a comment to that effect in the manifest file.

StephenMcConnel:
I've tried to respond to  all of the initial review comments, which are much appreciated.  The URL link to geckofx is rather flaky.  I've had 502 failures locally, but I've also had success downloading as well.

org.sil.Bloom.yml
barthalion: `host` includes  `/tmp` and `home`, which includes `xdg-data`. The app is unlikely to need so many permissions in the first place. Probably `home` and `/tmp` are enough.

barthalion:
bot, build org.sil.Bloom

org.sil.Bloom.yml
StephenMcConnel: I'm pretty sure that users will need access outside of home to load pictures, access shared team folders on a local network, save books and publications to USB drives, etc.

StephenMcConnel:
The source of "geckofx" which has failed repeatedly, and which is known to be unreliable, has been replaced by a new project on github which should work much better.

barthalion:
bot, build org.sil.Bloom

StephenMcConnel:
I don't understand the most recent build failure at all.  It appears that the build system is building BloomDesktop before it has built the modules listed under that entry. The submodule in libpalaso.yml needs to build before BloomDesktop because it creates artifacts used in the build process for BloomDesktop.  This always happens on my local system, but seems not to have happened on the flathub build system.  Several other submodules also need to be built before BloomDesktop.  Is there something obvious I'm missing because of my inexperience with flatpak?

org.sil.Bloom.yml
```diff
@@ -0,0 +1,581 @@
...
+    - --filesystem=host
+    # Bloom writes temporary folders and files to /tmp, to open in a browser to
+    # show the user, such as creating PDFs, ePUBs, or BloomPub ebooks.  Use the
+    # same /tmp as the host, so the host programs can find the files.
+    - --filesystem=/tmp
+    # libGL needs this
+    - --device=dri
+command: bloom
+only-arches:
+    - x86_64
```
hfiguiere, line 58: this doesn't do what you think it does. I'm pretty sure flatpak-builder witll print a warning about this.

org.sil.Bloom.yml
```diff
@@ -0,0 +1,581 @@
...
+                - |
+                    set -xueo pipefail
+                    pwd
+                    mkdir -p "BloomDesktop/DistFiles/pdf" && unzip -uqo BloomDesktop/Downloads/pdfjs-viewer.zip -d "BloomDesktop/DistFiles/pdf"
+          # python3 flatpak-builder-tools/node/flatpak-node-generator.py -o bloom-sources-browser-node.json yarn ../src/BloomBrowserUI/yarn.lock
+          - bloom-sources-browser-node.json
+          - type: patch
+            path: patches/BloomDesktop/BuildFileFixesForFlatpak.patch
+            options:
+                - --verbose
```
hfiguiere, line 239:
```suggestion
```

hfiguiere:
There are three warning by flaptak-builder

```
(flatpak-builder:2): flatpak-builder-WARNING **: 06:36:33.654: Unknown property only-arches for type BuilderManifest
(flatpak-builder:2): flatpak-builder-WARNING **: 06:36:33.657: Unknown property cleanup for type BuilderSourceShell
(flatpak-builder:2): Json-WARNING **: 06:36:33.706: Failed to deserialize "modules" property of type "gpointer" for an object of type "BuilderModule"
```

org.sil.Bloom.yml
```diff
@@ -0,0 +1,581 @@
...
+                    cp --verbose --force /app/libpalaso/SIL.WritingSystems.dll "${NUGET_DEST}"/SIL.WritingSystems.9.0.0-beta0119/lib/net461/
+                    cp --verbose --force /app/libpalaso/SIL.WritingSystems.pdb "${NUGET_DEST}"/SIL.WritingSystems.9.0.0-beta0119/lib/net461/
+                    rm "${NUGET_DEST}"/Geckofx60.64.Linux.60.0.51/lib/net40/Geckofx-*.dll.mdb
+                    cp --verbose --force /app/geckofx/Geckofx-Core.dll "${NUGET_DEST}"/Geckofx60.64.Linux.60.0.51/lib/net40/
+                    cp --verbose --force /app/geckofx/Geckofx-Core.dll.config "${NUGET_DEST}"/Geckofx60.64.Linux.60.0.51/lib/net40/
+                    cp --verbose --force /app/geckofx/Geckofx-Core.pdb "${NUGET_DEST}"/Geckofx60.64.Linux.60.0.51/lib/net40/
+                    cp --verbose --force /app/geckofx/Geckofx-Winforms.dll "${NUGET_DEST}"/Geckofx60.64.Linux.60.0.51/lib/net40/
+                    cp --verbose --force /app/geckofx/Geckofx-Winforms.dll.config "${NUGET_DEST}"/Geckofx60.64.Linux.60.0.51/lib/net40/
+                    cp --verbose --force /app/geckofx/Geckofx-Winforms.pdb "${NUGET_DEST}"/Geckofx60.64.Linux.60.0.51/lib/net40/
+            cleanup:
```
hfiguiere, line 216: this is misplaced. It doesn't belong to a `source`.

org.sil.Bloom.yml
```diff
@@ -0,0 +1,581 @@
...
+                    install Geckofx-Winforms/bin/Release_Linux/Geckofx-Winforms.dll.config /app/geckofx
+                    install Geckofx-Winforms/bin/Release_Linux/Geckofx-Winforms.pdb /app/geckofx
+            sources:
+                - type: git
+                  url: https://github.com/StephenMcConnel/Geckofx60.git
+                  commit: a77e5ba72728630ee02b95892d5fd0c219f87c98
+                - type: patch
+                  path: patches/geckofx/FixesForFlatpak.patch
+                  options:
+                      - --verbose
```
hfiguiere, line 564:
```suggestion
```

libpalaso.yml
```diff
@@ -0,0 +1,129 @@
...
+    - palaso/SIL.Windows.Forms.WritingSystems-nuget.json
+    - palaso/SIL.Windows.Forms.WritingSystems.Tests-nuget.json
+    - palaso/SIL.WritingSystems-nuget.json
+    - palaso/SIL.WritingSystems.Tests-nuget.json
+    - palaso/TestApps-nuget.json
+    - palaso/TestAppKeyboard-nuget.json
+    - type: patch
+      path: patches/libpalaso/RestrictLibpalasoToNet461.patch
+      options:
+          - --verbose
```
hfiguiere, line 113:
```suggestion
```

libpalaso.yml
```diff
@@ -0,0 +1,129 @@
...
+    - type: patch
+      path: patches/libpalaso/FixesForFlatpak.patch
+      options:
+          - --verbose
```
hfiguiere, line 117:
```suggestion
```

libpalaso.yml
```diff
@@ -0,0 +1,129 @@
...
+    - name: msbuild
+      sources:
+          - type: file
+            url: https://download.mono-project.com/repo/ubuntu/pool/main/m/msbuild/msbuild_16.10.1+xamarinxplat.2021.05.26.14.00-0xamarin2+ubuntu2004b1_all.deb
+            sha256: 32f1eba1c2df67bc55920cd45ce3689c32a5c705ae01fd7ead52ca180221b469
+      buildsystem: simple
+      build-commands:
+          - ar x msbuild_*.deb
+          - tar xf data.tar.xz
+          - sed -e 's|/usr/|/usr/lib/sdk/mono6/|g' -i usr/bin/msbuild
```
hfiguiere, line 128: instead you should install this in `/app`

hfiguiere (changes requested):
manifest would be easier to review if it had a more flat structure. (like less nested modules)

org.sil.Bloom.yml
hfiguiere: You are probably thinking of this:

Which remain disappointing give that this is all C#

StephenMcConnel: Unfortunately, the Geckofx60 package which provides access to the Mozilla/Firefox engine is only available for x86_64.

libpalaso.yml
StephenMcConnel: removed.

libpalaso.yml
StephenMcConnel: There's an older version of msbuild (mono5-sil-msbuild) already installed in /app.  I'm pretty sure the version of msbuild has to be pretty close to the version of the mono compilers and libraries, so I haven't ever tried mixing the msbuild versions and the mono versions.  Why the mono6 Sdk doesn't include msbuild is beyond me since it's the standard build tool for C# development.

org.sil.Bloom.yml
StephenMcConnel: removed.

org.sil.Bloom.yml
```diff
@@ -0,0 +1,581 @@
...
+          # python3 flatpak-builder-tools/node/flatpak-node-generator.py -o bloom-sources-browser-node.json yarn ../src/BloomBrowserUI/yarn.lock
+          - bloom-sources-browser-node.json
+          - type: patch
+            path: patches/BloomDesktop/BuildFileFixesForFlatpak.patch
+            options:
+                - --verbose
+          - type: patch
+            path: patches/BloomDesktop/CodeFixesForFlatpak.patch
+            options:
+                - --verbose
```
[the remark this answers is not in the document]
StephenMcConnel: removed.

org.sil.Bloom.yml
```diff
@@ -0,0 +1,581 @@
...
+                - type: archive
+                  # Update URL by doing: cd .../path/to/mono-repo && .../determine-monolite-url
+                  url: https://download.mono-project.com/monolite/monolite-linux-1051600011-latest.tar.gz
+                  sha512: b48bb273f5a9107f9bb9795d47277aeaaeb458588a6e011676bb9a152b83759f305cf1031a5e74a62145854ef17026cb1aa20ea9adb5e6b15f275b9d6af585db
+                  dest: mcs/class/lib/monolite-linux/1051600011
+                - type: patch
+                  # mono uses a hard-coded path to ping. Help it find ping in /app.
+                  path: patches/mono/mono-ping-path.patch
+            modules:
+                - name: python-2.7
```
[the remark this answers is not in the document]
StephenMcConnel: changed to use the shared-modules, and moved before mono5-sil instead of using a modules section.

org.sil.Bloom.yml
StephenMcConnel: removed the cleanup and added a "rm -rf" command following all the "cp" commands.

org.sil.Bloom.yml
StephenMcConnel: removed.

libpalaso.yml
StephenMcConnel: removed.

libpalaso.yml
StephenMcConnel: removed.

StephenMcConnel:
I tried removing the modules that were two levels deep by moving the content to before the items they are/were modules of.  A clean build succeeded locally without any of the WARNING messages in the previous flathub build.  I hope these changes will get things over the hump.  I really appreciate your feedback and patience.
Where I haven't taken your advice, I've tried to explain why.

hfiguiere:
Why doesn't this flatpak use the .desktop and .appdata.xml, or even the icons that are in the upstream repository?

StephenMcConnel:
I thought it was using/installing the icons provided by github.com/BloomBooks/BloomDesktop project in the debian folder.  (I assume that's what you mean by the "upstream repository".)
I could update the .appdata.xml file in the debian folder to match what I'm providing explicitly since that upstream file hasn't been updated for years apart from version number and release date.  That would require a .deb repository to be able to share exactly the same .appdata.xml file, which I wasn't sure was possible, but I suppose there's no reason why this file can't be shared.
The .desktop file for flatpak seemed to indicate that the program executable should be "org.sil.Bloom", while the .desktop file for the older .deb files had just "bloom".  I suppose I could make this adjustment with sed or a patch file if it's really needed and use the existing file as a base.

hfiguiere:
Discard was I said about the icon, I misspoke.

> I could update the .appdata.xml file in the debian folder to match what I'm providing explicitly since that upstream file hasn't been updated for years apart from version number and release date. That would require a .deb repository to be able to share exactly the same .appdata.xml file, which I wasn't sure was possible, but I suppose there's no reason why this file can't be shared.

Many projects have one single appdata file and desktop file.

You could patch it in the first place, but definitely should be updated upstream for a future releae.

> The .desktop file for flatpak seemed to indicate that the program executable should be "org.sil.Bloom",

No. It expect the executable name as installed in the install process, the same one mentionned in the manifest at the top.

> while the .desktop file for the older .deb files had just "bloom". I suppose I could make this adjustment with sed or a patch file if it's really needed and use the existing file as a base.

1. don't use `sed` on .destkop but `desktop-file-edit` because it's much clearer.
2. we have `rename-desktop-file` to deal with improper name (but this should really be fixed upstream in the future)
3. `rename-icons` handle the icon name in the .desktop file when you actually rename icons, otherwise see `desktop-file-edit`

extra/org.sil.Bloom.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
+Version=1.0
+Name=Bloom Alpha
+Comment=Literacy materials development for language communities
+Exec=org.sil.Bloom %u
```
hfiguiere, line 5: manifest say ` command: bloom`

So this is incorrect.

org.sil.Bloom.yml
```diff
@@ -0,0 +1,563 @@
...
+    - --filesystem=/tmp
+    # libGL needs this
+    - --device=dri
+command: bloom
+
+modules:
+    - name: BloomDesktop
+      buildsystem: simple
+      only-arches:
+          - x86_64
```
hfiguiere, line 62: this still make an `aarch64` flatpak to be built. I linked you to the specific bit in the documentation about that.

org.sil.Bloom.yml
```diff
@@ -0,0 +1,563 @@
...
+    - name: ArtOfReading
+      sources:
+          # This .deb contains only PNG image files with a few text files such
+          # as an index and changelog.
+          - type: file
+            url: https://packages.sil.org/ubuntu/pool/main/a/art-of-reading3/art-of-reading3_3.3.1+bionic1_all.deb
+            sha256: 96d99d60c1838aaac8461e3568185cc287206d2852bd3e11d94eef5d9d6a3f15
+      buildsystem: simple
+      only-arches:
+          - x86_64
```
hfiguiere, line 554:
```suggestion
```

org.sil.Bloom.yml
StephenMcConnel: My bad.  I didn't scroll up from the exact location your link pointed to, so I didn't realize it was supposed to be in a separate file.

org.sil.Bloom.yml
StephenMcConnel: removed in both places.  added a flathub.json file

extra/org.sil.Bloom.desktop
StephenMcConnel: I changed to using the .desktop and .appdata.xml from the BloomDesktop/debian folder.  I updated the .appdata.xml and updated the commit value to get the updated file.

StephenMcConnel:
I think I've responded to all the comments.  Thanks for clarifying the .desktop and .appdata.xml situation.

nedrichards:
bot, build org.sil.Bloom

hfiguiere:
```
 extracting: BloomDesktop/packages/TechTalk.JiraRestClient.RestSharp105.2.3.0.1/.signature.p7s  
+ mv BloomDesktop/packages/MarkdownDeep.NET.1.5.0/lib/.NetFramework%203.5 'BloomDesktop/packages/MarkdownDeep.NET.1.5.0/lib/.NetFramework 3.5'
+ chmod 755 'BloomDesktop/packages/sqlite-net-pcl.1.7.335/[Content_Types].xml' BloomDesktop/packages/sqlite-net-pcl.1.7.335/LICENSE.txt BloomDesktop/packages/sqlite-net-pcl.1.7.335/sqlite-net-pcl.nuspec
+ chmod 755 BloomDesktop/packages/sqlite-net-pcl.1.7.335/lib/netstandard2.0/SQLite-net.dll BloomDesktop/packages/sqlite-net-pcl.1.7.335/lib/netstandard2.0/SQLite-net.xml
+ cp --verbose --force /app/libpalaso/SIL.Core.dll BloomDesktop/packages/SIL.Core.9.0.0-beta0119/lib/net461/
cp: cannot stat '/app/libpalaso/SIL.Core.dll': No such file or directory
FB: host_command_exited_cb 2129278 256
```

hfiguiere:
There is something about not having installation scripts upstream. like a `make install` or whatever equivalent.

StephenMcConnel:
I don't understand why this failed.   This configuration built locally without any problem.  It did take 90 minutes on my old Linux box.  The command I've been using is
```
flatpak-builder --user --repo=../flatpak/local-repo --keep-build-dirs --force-clean ./build-dir org.sil.Bloom.yml
```

And I don't understand your comment at all.

hfiguiere:
> And I don't understand your comment at all.

If you mean

> There is something about not having installation scripts upstream. like a make install or whatever equivalent.

Then what I mean is that of lot of this ought to be part of the upstream build system. Probably because I don't  understand how .net works, but most flatpak can build by doing `make install` (or equivalent) instead of a very long streak of commands, very error prone, like it is in this manifest.

Given that is one of these lines that fail...

BTW doing `rm -rf .flatpak-builder` just make you build longer for no reason, and `--keep-build-dirs` is only usefull to analyze because they are never reused anyway.

What I imagine here is that the file you are trying to copy hasn't been installed yet.

StephenMcConnel:
I don't understand your comment about installation scripts upstream.  With a "simple" buildsystem, all the install commands can be and are given inline.

StephenMcConnel:
I'm more concerned with why there's still a 
(flatpak-builder:2): Json-WARNING **:0:33:34.809: Failed to deserialize "modules" property of type "gpointer" for an object of type "BuilderModule"
error logged at line 28 of output at https://buildbot.flathub.org/#/builders/10/builds/12451/steps/6/logs/stdio.  This explains why libpalaso didn't build and place its files where they can be found.  But that error isn't occurring locally, so everything builds okay.

hfiguiere:
are you sure all the files are checked in? you have so many...

StephenMcConnel:
git clean -ndx shows only the build-dir and .flatpak-builder folders, which indicates that everything is checked in for a successful build since I've been building off the branch in my fork since creating it.

hfiguiere:
I have the error here `(flatpak-builder:227308): Json-WARNING **: 18:30:09.738: Failed to deserialize "modules" property of type "gpointer" for an object of type "BuilderModule"`

Looks like it is missing the `shared-modules` submodule. it was probably not added correctly.

Doing
```
git submodule add https://github.com/flathub/shared-modules.git
```
git status shows
```
	new file:   shared-modules
```

fixes it.

hfiguiere:
> and the bulk of these instructions are almost directly from the debian rules file, which is essentially a makefile.

Maybe this should be extracted. After all where I come from, this is part of the build system. Which is exactly what I was talking about.

StephenMcConnel:
That is strange.  I did the git submodule add command here, and checked in the resulting .gitmodules file.  I'm not used to working with submodules.  Should I have done something more?  Do I need to do something more?

I may try to refactor things to extract a makefile for BloomDesktop.  I've never really felt the need for one before because just building the program was only a simple "msbuild" command, and installing the program only went through the debian packaging system.  But I can see that could simplify some things for local testing and certainly for the manifest file.  I lived in the C/C++ Makefile/autotools world for years before .Net so I'm certainly familiar with the concepts.

hfiguiere:
If you pass `--ccache` to `flatpak-buidler` it will use ccache to build C/C++ code. Off course it's moot if you delete the `.flatpak-builder` directory, but it might help a lot when you have to build more than once.

barthalion:
Looks like I cannot push to your branch. Please add `shared-modules` properly, that should fix the build issue.

StephenMcConnel:
I think I've added shared-modules properly now.

hfiguiere:
bot, build org.sil.Bloom

StephenMcConnel:
Build 103640 works when I installed it from the testing repository for the quick smoke test of some of Bloom's features.

StephenMcConnel:
Is there anything else I need to do to get this approved and merged?

PR #3368, opened 2022-08-03
Add io.itch.doctorm64.skippybot

Miepee:

  I do not, but DoctorM64 does, whose permission I have to use the domain.
  I submitted them to the lead developer/project maintainer. They may or may not appear in the next upstream version.

Additional notes: 
- I asked on whether the project could be open sourced. Currently he wishes to not do so for personal private reasons.
- As explained in a comment in the desktop file, the download source is being fetched from a third-party URL rather than from itch.io themselves. This is because (to my knowledge) itch.io does not provide a way to directly download of assets without the use of an itch account and API keys. However, since the sha512 hash is checked against the file from itch.io, this should not cause any malicious hijacks to go through.


io.itch.doctorm64.skippybot.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
+Type=Application
+Categories=Game
+Encoding=UTF-8
+Name=Skippy The Bot
+Comment=Skippy is on an epic quest to unravel its true destiny.
+Exec=/app/bin/skippybot
```
hfiguiere, line 7:
```suggestion
Exec=skippybot
```

io.itch.doctorm64.skippybot.yaml
```diff
@@ -0,0 +1,44 @@
+app-id: io.itch.doctorm64.skippybot
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+tags:
+  - proprietary
+finish-args:
+- --socket=wayland
```
hfiguiere, line 8: does it really support wayland?

io.itch.doctorm64.skippybot.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - install -Dm644 UnityPlayer.so /app/bin/UnityPlayer.so
+      - install -Dm755 skippy_${SKIPPY_VERSION}.x86_64 /app/bin/skippybot
+      - cp -r skippy_${SKIPPY_VERSION}_Data/ /app/bin/skippybot_Data
+      # meta files
+      - install -Dm644 $FLATPAK_ID.appdata.xml /app/share/metainfo/$FLATPAK_ID.appdata.xml
+      - install -Dm644 $FLATPAK_ID.png /app/share/icons/hicolor/128x128/apps/$FLATPAK_ID.png
+      - install -Dm644 $FLATPAK_ID.desktop /app/share/applications/$FLATPAK_ID.desktop
+    sources:
+      - type: archive
+        # redirected URL because itch.io does not allow direct downloads of assets. We are checking the hash tho, so file is the same.
```
hfiguiere, line 34: this basically will break on a whim.

also is redistribution authorized?

hfiguiere (changes requested):

hfiguiere:
bot, build io.itch.doctorm64.skippybot

io.itch.doctorm64.skippybot.desktop
Miepee: That won't work. The Unity engine (which this game uses) is a bit special in how it wants games to be launched. It needs to be launched via the absolute path, as otherwise the game will not find it's assets.
I meant to add a comment to this before pushing, but forgot. Added it now locally.

io.itch.doctorm64.skippybot.yaml
Miepee: Wayland is indeed supported.

io.itch.doctorm64.skippybot.yaml
Miepee: Redistribution is authorized.

Can you go a bit more detail with how it'll break? 
I do find it quite sad that itch.io does not give a way to directly download assets, which leaves us with two options (to my knowledge at least):
- Use the redirected URL and update that on new versions and hope that the service itself never breaks
- Push the zip directly into the repository, and use it as a "file" source.

I personally opted for the former as having a 30MB binary commit on each new version is very ugly and bloats the repo too much IMO. However, if you think that that's a better option than relying on the third party, I'll use that instead.
And if you've got any other ideas on how to get around the issue, I'm all ears for that as well.

hfiguiere: if the sha256 is different that mean the build will be broken. 

A .zip in the flathub repo would be REJECTED.

io.itch.doctorm64.skippybot.yaml
```diff
@@ -0,0 +1,44 @@
...
+        url: https://cdn.matix-media.net/dd/595efc05
+        dest-filename: skippy_linux64_$SKIPPY_VERSION.zip
```
hfiguiere, line 36: This make the file saved as `skippy_linux64_$SKIPPY_VERSION.zip`.

I don't even see the reason to do so.

io.itch.doctorm64.skippybot.yaml
Miepee: > if the sha256 is different that mean the build will be broken.

*If* the hash is different the build will be broken, yes, but the hash should never be different (and it never was in my few months of using it).

io.itch.doctorm64.skippybot.yaml
Miepee: Ah, thanks for the notice, didn't catch that.
Changed it locally to just use the actual version number instead.

io.itch.doctorm64.skippybot.yaml
Miepee: Since redistribution of the assets are allowed, one could technically repost the asset to some other place (gdrive, owncloud etc.) and then use that as a source.
Although that probably wouldn't be that different from the current approach.

hfiguiere: if the url provide the same content in a stable way (and get changed on a new release) then to shouldn't be a problem. I initially understood that it wasn't like that.

Miepee: I'm sorry then if I made that unclear initially.
But yeah, just for clarification again, the current matix link is just a redirect to the actual asset on itch.io in order to directly download the asset there.
I'll try to make the comment a bit more clear.

Miepee:
I noticed the game didn't launch when ran via non-desktop file, so now I've introduced a wrapper for it instead.

Miepee:
bot, build io.itch.doctorm64.skippybot

PR #3391, opened 2022-08-05
Add org.kotlinlang.Kotlin

Doomsdayrs:

  - home: To allow compilation of files 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+build/
+.flatpak-builder/
--- a/org.kotlinlang.Kotlin.metainfo.xml
+++ b/org.kotlinlang.Kotlin.metainfo.xml
@@ -11,4 +11,7 @@
   <metadata_license>CC0-1.0</metadata_license>
   <project_license>Apache-2.0</project_license>
   <content_rating type="oars-1.1"/>
+  <releases>
+    <release version="1.8.22" date="2023-06-08"/>
+  </releases>
 </component>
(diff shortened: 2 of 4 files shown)


barthalion:
Is it going to be used by some application or what's the rationale behind this PR?

Doomsdayrs:
> Is it going to be used by some application or what's the rationale behind this PR?

Currently, its to make it easier to use the kotlin compiler on Linux easier.

SDKMAN is janky, and the only other option is snap.

I rather have it function with flatpak so I do not need to handle the trouble of upgrading libs and the so forth.

Regarding other programs, I believe it is possible for InteliJ to benefit from this flatpak, but I would have to research into it.

Currently this is a draft since I want to verify a few more things in regards to it before I continue, especially regarding an extension for kotlin native compilation.

barthalion:
We don't accept non-apps like this. It also would need to be a proper SDK extension to be useful for other apps or building.

Doomsdayrs:
> We don't accept non-apps like this. It also would need to be a proper SDK extension to be useful for other apps or building.

But this is an application? Its job is to compile Kotlin?

I feel like it is in the same class of applications as FEDC, which just checks a file and verifies ED, this can check a file and compiles it.

Doomsdayrs:
> I don't think there is anything not Android built with Kotlin. Maybe for AndroidStudio? (isn't there a flatpak for that?)

Kotlin is used for more then android. It has compilation to native and JavaScript.

It can also be run as a script, in which this program can run said scripts.

Doomsdayrs:
bot, build org.kotlinlang.Kotlin

Doomsdayrs:
A console application needs screenshots?

Doomsdayrs:
bot, build org.kotlinlang.Kotlin

PR #3404, opened 2022-08-10
New pr for com.github.scrivano.scrivano

scrivanolabs:


com.github.scrivano.scrivano.yml
```diff
@@ -0,0 +1,31 @@
+app-id: com.github.scrivano.scrivano
```
hfiguiere, line 1: the app-id doesn't meet the requirement. There is no `scrivano` user or organization on Github.

com.github.scrivano.scrivano.yml
```diff
@@ -0,0 +1,31 @@
...
+runtime: org.kde.Platform
+runtime-version: '5.15-21.08'
+sdk: org.kde.Sdk
+command: Scrivano
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --socket=wayland
+  - --filesystem=host
```
hfiguiere, line 10: this is a very very broad file system permission

com.github.scrivano.scrivano.yml
```diff
@@ -0,0 +1,31 @@
+app-id: com.github.scrivano.scrivano
+runtime: org.kde.Platform
+runtime-version: '5.15-21.08'
+sdk: org.kde.Sdk
+command: Scrivano
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --socket=wayland
```
hfiguiere, line 9:
```suggestion
  - --socket=fallback-x11
  - --socket=wayland
```

if it supports wayand, x11 should be obly fallback.

com.github.scrivano.scrivano.yml
```diff
@@ -0,0 +1,31 @@
...
+    build-commands:
+    - install -D Scrivano /app/bin/Scrivano
+    - install -D libskia.so /app/lib/libskia.so
+    - install -D libpdfium.so /app/lib/libpdfium.so
+    - install -Dm644 com.github.scrivano.scrivano.appdata.xml -t /app/share/metainfo
+    - install -Dm644 com.github.scrivano.scrivano.desktop -t /app/share/applications
+    - install -Dm644 app_icon.svg /app/share/icons/hicolor/scalable/apps/com.github.scrivano.scrivano.svg
+    sources:
+    - type: archive
+      url: https://github.com/scrivanolabs/ScrivanoForLinux/releases/download/0.13.4/Scrivano_flatpak_files_0.13.4.zip
```
hfiguiere, line 24: this contains x86_64 binaries, but you still build a aarch64 flatpak.

com.github.scrivano.scrivano.yml
```diff
@@ -0,0 +1,31 @@
...
+      sha256: 48841ba44d95b6d7b625266c9d4622b0492465f29fec2491292139054b27cd09
+    - type: file
+      path: com.github.scrivano.scrivano.appdata.xml
+    - type: file
+      path: com.github.scrivano.scrivano.desktop
+    - type: file
+      path: app_icon.svg
```
hfiguiere, line 31: neitiher of these 3 files is available.

hfiguiere (changes requested):

com.github.scrivano.scrivano.yml
scrivanolabs: Ok I changed the id to meet the requirements

com.github.scrivano.scrivano.yml
scrivanolabs: I've now set up more specific file system and sandbox permissions

com.github.scrivano.scrivano.yml
scrivanolabs: Thanks for letting me know, done.

com.github.scrivano.scrivano.yml
scrivanolabs: Thanks for pointing this out, indeed those files are contained in my Github repo. Done.

hfiguiere:
bot, build com.github.scrivanolabs.scrivano

com.github.scrivanolabs.scrivano.yml
```diff
@@ -0,0 +1,32 @@
...
+    - install -D libpdfium.so /app/lib/libpdfium.so
+    - install -Dm644 com.github.scrivanolabs.scrivano.appdata.xml -t /app/share/metainfo
+    - install -Dm644 com.github.scrivanolabs.scrivano.desktop -t /app/share/applications
+    - install -Dm644 app_icon.svg /app/share/icons/hicolor/scalable/apps/com.github.scrivanolabs.scrivano.svg
+    sources:
+    - type: archive
+      url: https://github.com/scrivanolabs/ScrivanoForLinux/releases/download/0.13.4/Scrivano_flatpak_files_0.13.4.zip
+      sha256: b1d5730851b2bfdc5f69c6914c6ce1c21a48ac8e1d75a81e09b8ea9383c77959
+      only-arches:
+        - 'x86_64'
```
hfiguiere, line 32: not this doesn't do what you think it does.

It remain disappointing though.

com.github.scrivanolabs.scrivano.yml
```diff
@@ -0,0 +1,32 @@
+app-id: com.github.scrivanolabs.scrivano
+runtime: org.kde.Platform
+runtime-version: '5.15-21.08'
+sdk: org.kde.Sdk
+command: Scrivano
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --filesystem=xdg-config
```
hfiguiere, line 10:
```suggestion
```

Why would you need that?

com.github.scrivanolabs.scrivano.yml
```diff
@@ -0,0 +1,32 @@
...
+  - --filesystem=xdg-desktop
+  - --filesystem=xdg-documents
+  - --filesystem=xdg-download
+  - --filesystem=xdg-pictures
+  - --filesystem=xdg-public-share
```
hfiguiere, line 15: Not sure this is wise either.

hfiguiere (changes requested):

com.github.scrivanolabs.scrivano.yml
scrivanolabs: My reasoning was that if the user is using both a Flatpak version and a non Flatpak version of Scrivano, they would be able to use access the same configuration files. However, this is probably not wise and I'll commit your suggestion.

com.github.scrivanolabs.scrivano.yml
scrivanolabs: Ah I see, thank you for pointing this out. I wasn't aware of the need of a separate file to specify build architectures.

hfiguiere:
```
builddir/files/share/appdata/com.github.scrivanolabs.scrivano.appdata.xml: FAILED:
• tag-invalid           : <project_license> is not valid [Proprietary]SPDX ID 'Proprietary' unknown
Validation of files failed
```

com.github.scrivanolabs.scrivano.yml
scrivanolabs: Could you please explain why this is not wise? Are you referring to the xdg-public-share permission? What do you think would be the most appropriate set of permissions for a notetaking application?

hfiguiere: you should explain why you want to read-write all these location including `xdg-public-share`.

com.github.scrivanolabs.scrivano.yml
hfiguiere: This is not the standard practice at all. You also grant access to just other app preferences.

It's not like we can verify the source code at any point.

com.github.scrivanolabs.scrivano.yml
scrivanolabs: Scrivano is a notetaking application and allows users to save their notes in a folder of their choice on their machine (personally, as a user I would like to have the freedom to save my notes anywhere, at least in my home directory, but I understand that Flatpak users might not want this).

To do this, Scrivano uses a custom file dialog built-in within the app (this choice was made of several reasons, primarily because of consistency across operating systems and easily display thumbnails for .scrvn files and also keeping in mind that in future an integrated sidebar will likely be displayed inside the application to allow users to easily switch between their notes). I've read on the Flatpak website that an app should first consider using portals to let users use a a native file dialog to access this, but this wouldn't be possible with Scrivano (also because the app requires the ability to auto-save files based on a time interval without the user directly intervening).

So I think granting access to xdg-documents is quite important. Scrivano also needs the ability to import/export PDF files and pictures thus it needs to access xdg-downloads and xdg-pictures folders. I think one could argue that only read-only access should be given to these folders, however users often want to save their notes in these folders and I think it would be quite limiting to not be able to do so. For example, a scenario I run into often is this: I have a PDF I would like to annotate so I import it in Scrivano and then save it next to the file until I finish annotating it and then eventually export it as a PDF in a folder of my choice.
For the xdg-public-share folder, I think this is not strictly necessary (I personally do use this as a way to share notes across my local devices but most users probably wouldn't need this).
Although, I should probably add xdg-desktop in the permission list as it's very common for users to save their notes in their desktop folder.

hfiguiere: so by design you don't use the FileChooser portal. You should.

scrivanolabs: Yes, that's (somewhat) correct. Scrivano doesn't use native file dialogs (although this may be implemented for *some* things, e.g. importing pictures, in the future). As I mentioned, the choice isn't just by design but dictated by some features that are needed by the app (for example, as I mentioned above, the ability to save files using an interval based timer which needs to happen without user intervention and this, as far as I know, wouldn't be possible using the portal api).

scrivanolabs:
bot, build com.github.scrivanolabs.scrivano

scrivanolabs:
bot, build com.github.scrivanolabs.scrivano

com.github.scrivanolabs.scrivano.yml
```diff
@@ -0,0 +1,40 @@
...
+    - install -Dm644 scrivano_64.png /app/share/icons/hicolor/64x64/apps/com.github.scrivanolabs.scrivano.png
+    - install -Dm644 scrivano_128.png /app/share/icons/hicolor/128x128/apps/com.github.scrivanolabs.scrivano.png
+    - install -Dm644 scrivano_256.png /app/share/icons/hicolor/256x256/apps/com.github.scrivanolabs.scrivano.png
+    - install -Dm644 scrivano_512.png /app/share/icons/hicolor/512x512/apps/com.github.scrivanolabs.scrivano.png
+    sources:
+    - type: archive
+      url: https://github.com/scrivanolabs/ScrivanoForLinux/releases/download/0.13.4/Scrivano_flatpak_files_0.13.4.zip
+      sha256: 46a79db8cb96eb9ad9a089d487bc7d97d1c585b17fb08b35ecd8e3cc9460708e
+    - type: file
+      path: icons/scrivano_64.png
```
hfiguiere, line 33: why aren't the icons in the downloaded archive?

hfiguiere: also why is the included svg never installed?

hfiguiere: The build fail because the appdata file has:
```
  <launchable type="desktop-id">com.github.scrivano.scrivano.desktop</launchable>
```
which is incorrect.

scrivanolabs: For some reason, when I tested locally I was getting a strange `error: syncfs: Input/output error` which disappeared when I moved the icons in the repo. Anyway, I tested again and moved the icons to the installation source and it seems to work correctly.

scrivanolabs: > also why is the included svg never installed?

Fixed.

scrivanolabs: > The build fail because the appdata file has:
> 
> ```
>   <launchable type="desktop-id">com.github.scrivano.scrivano.desktop</launchable>
> ```
> 
> which is incorrect.

Thanks for pointing this out. I fixed this.

scrivanolabs:
bot, build com.github.scrivanolabs.scrivano

PR #3406, opened 2022-08-11
add foot manifest

microo8:

- [x] I am using only the minimal set of permissions. (It just needs to run the shell of the user and access it's config files in `xdg-config/foot`)
- [x] I contacted upstream developers about submitting their software to Flathub. **Link: https://codeberg.org/dnkl/foot/issues/1129#issuecomment-598905**


org.codeberg.dnkl.foot.yml
```diff
@@ -0,0 +1,45 @@
+app-id: org.codeberg.dnkl.foot
+runtime: org.freedesktop.Platform
+runtime-version: "21.08"
+sdk: org.freedesktop.Sdk
+command: /app/bin/foot.sh
```
hfiguiere, line 5:
```suggestion
command: foot.sh
```

org.codeberg.dnkl.foot.yml
```diff
@@ -0,0 +1,45 @@
...
+
+finish-args:
+  - --socket=wayland
```
hfiguiere, line 8: what happens if the user doesn't have Wayland?

org.codeberg.dnkl.foot.yml
```diff
@@ -0,0 +1,45 @@
...
+  - --filesystem=xdg-config/foot:ro
+  - --talk-name=org.freedesktop.Flatpak
+
+modules:
+  - name: libutf8proc
+    buildsystem: cmake
```
hfiguiere, line 14:
```suggestion
    buildsystem: cmake-ninja
```

unless it doesn't work.

org.codeberg.dnkl.foot.yml
```diff
@@ -0,0 +1,45 @@
...
+        url: https://codeberg.org/dnkl/foot/archive/1.13.0.zip
+        sha256: d0af98e61ea25e121549ded561ee1fe1b11c901b81a7c11844bd6fb32139f256
+  - name: foot.sh
+    buildsystem: simple
+    sources:
+      - type: file
+        path: ./foot.sh
+    build-commands:
+      - cp ./foot.sh /app/bin/
+      - chmod +x /app/bin/foot.sh
```
hfiguiere, line 45:
```suggestion
      - install -Dm755 ./foot.sh -t /app/bin/
```

hfiguiere (changes requested):

hfiguiere:
bot, build org.codeberg.dnkl.foot

hfiguiere:
there is no appstream file found

org.codeberg.dnkl.foot.yml
JakobDev: If you take a lookj at the [linked Codeberg repo](https://codeberg.org/dnkl/foot) you will find this description:

>  A fast, lightweight and minimalistic Wayland terminal emulator

so it looks like it is Wayland only

microo8: yes, it's wayland only

org.codeberg.dnkl.foot.yml
```diff
@@ -0,0 +1,53 @@
...
+      - type: file
+        path: ./foot.sh
+    build-commands:
+      - install -Dm755 ./foot.sh -t /app/bin/
+  - name: host-spawn
+    buildsystem: simple
+    sources:
+      - type: file
+        url: https://github.com/1player/host-spawn/releases/download/1.2.0/host-spawn-x86_64
+        sha256: ab31ad0d980194e4c29d1e82715b37e7a02201504612d08521b413d1a7f74651
```
barthalion, line 50: Please switch it to build from source or just provide a simple shell wrapper for `flatpak-spawn --host` that is already available in all runtimes.

org.codeberg.dnkl.foot.yml
barthalion: This should be mentioned in the appdata.

org.codeberg.dnkl.foot.yml
```diff
@@ -0,0 +1,53 @@
+app-id: org.codeberg.dnkl.foot
+runtime: org.freedesktop.Platform
+runtime-version: "21.08"
+sdk: org.freedesktop.Sdk
+command: foot.sh
+
+finish-args:
+  - --socket=wayland
+  - --filesystem=xdg-config/foot:ro
```
barthalion, line 9: What exactly is this one for?

org.codeberg.dnkl.foot.metainfo.xml
```diff
@@ -0,0 +1,27 @@
...
+  <id>org.codeberg.dnkl.foot</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>MIT</project_license>
+
+  <developer_name>dnkl</developer_name>
+  <name>foot</name>
+  <summary>The fast, lightweight and minimalistic Wayland terminal emulator.</summary>
+  <description>
+    <p>The fast, lightweight and minimalistic Wayland terminal emulator.</p>
+  </description>
```
barthalion, line 13: Repeating the summary in the description doesn't make it a great app store entry.

org.codeberg.dnkl.foot.metainfo.xml
```diff
@@ -0,0 +1,27 @@
...
+    <release version="1.13.0" date="2022-08-07">
+    </release>
+  </releases>
+
+  <launchable type="desktop-id">org.codeberg.dnkl.foot.desktop</launchable>
+
+  <url type="homepage">https://codeberg.org/dnkl/foot</url>
+  <url type="bugtracker">https://codeberg.org/dnkl/foot/issues</url>
+
+  <content_rating type="oars-1.0"/>
```
barthalion, line 25:
```suggestion
  <content_rating type="oars-1.1"/>
```

foot.sh
```diff
@@ -0,0 +1,2 @@
+export SHELL="host-spawn $(flatpak-spawn --host getent passwd $USER | awk -F : '{ print $7 }')"
+/app/bin/foot --term=xterm-256color
```
barthalion, line 2: As it is, it will just ignore any arguments that can be potentially passed by user.

org.codeberg.dnkl.foot.metainfo.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
barthalion, line 1: This file should be submitted and maintained upstream.

org.codeberg.dnkl.foot.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
```
barthalion, line 1: This file should be submitted and maintained upstream.

microo8: Sorry for the noob question, but what does it mean "maintained upstream"?

org.codeberg.dnkl.foot.yml
microo8: foot reads it's configuration from `~/.config/foot/foot.ini`

JakobDev: Flatpak has it's own config directory, so this is not needed

org.codeberg.dnkl.foot.desktop
JakobDev: It means that it should be put in the Codebrg repo, not here in the Flathub repo

k6Q28k5w5eWuRuuzTg9j7GrKCbjt7: He has no affiliation with the domain. The manifest is blindly spawning a terminal outside the sandbox (this seems to be for packaging only). This manifest should never reach flathub as it doesnt use flatpak internals in any sense other than packaging .

Please see responses from upstream: https://codeberg.org/dnkl/foot/issues/1129#issuecomment-598905

hfiguiere: A Linux desktop appication without a .desktop, it is a bug. Upstream is the original developers, they should be the one maintaining both the .desktop, icon and appdata (neither is flatpak specific) and we try to make this a policy.

org.codeberg.dnkl.foot.yml
hfiguiere: > foot reads it's configuration from `~/.config/foot/foot.ini`

Did you actually check it was needed? this is the standar xdg path outside flatpak

One way to check is not have that permission and see if the settings are retained across launches.

org.codeberg.dnkl.foot.desktop
microo8: > He has no affiliation with the domain.

Yes I wanted to package it on my own (maybe if there was someone in whom it would spark an interest in flatpak, there is a `help wanted` label on the [issue](https://codeberg.org/dnkl/foot/issues/1129) after all.

>The manifest is blindly spawning a terminal outside the sandbox (this seems to be for packaging only). This manifest should never reach flathub as it doesnt use flatpak internals in any sense other than packaging .

It doesn't spawn a terminal outside the sandbox. It just spawns a shell on the host, using [host-spawn](https://github.com/1player/host-spawn) because it solves some things with the pty that `flatpak-spawn --host` does not.
I don't know what's wrong on spawning the host shell, [BlackBox](https://flathub.org/apps/details/com.raggesilver.BlackBox) does it too [here](https://gitlab.gnome.org/raggesilver/blackbox/-/blob/main/src/widgets/Terminal.vala#L302).

As for the "maitained upstream" issue, there is a [foot.desktop](https://codeberg.org/dnkl/foot/src/branch/master/foot.desktop) in the repo, but there is no appstream and the main dev has said, [this](https://codeberg.org/dnkl/foot/issues/1129#issuecomment-597445):

> FWIW, I really really really do not want to add flatpak specific code to foot. If that means foot cannot be packaged as a flatpak, then so be it.

So I tried to overcome it, without changing the code, with just rewriting the `$SHELL` var and prepend `host-spawn`. 

I could try to convince the author to add appstream file to the repo. After all it's not flatpak specific.

microo8:
I've removed `xdg-config/foot` permission and it is now building `host-spawn` from source. The only thing would now be just the appstream file in the repo.

org.codeberg.dnkl.foot.yml
```diff
@@ -0,0 +1,60 @@
...
+    build-commands:
+      - install -Dm755 ./foot.sh -t /app/bin/
+  - name: host-spawn
+    buildsystem: simple
+    build-options:
+      append-path: /usr/lib/sdk/golang/bin
+      env:
+        GOBIN: /app/bin
+      build-args:
+        - --share=network
```
hfiguiere, line 53: build-args are NOT allowed - it should have failed if you tried locally. You need to list the dependencies for go. See https://github.com/flatpak/flatpak-builder-tools for helpers.

org.codeberg.dnkl.foot.desktop
hfiguiere: appstream file ARE NOT FLATPAK SPECIFIC.

Only the manifest is.

microo8:
Yes, that's what I'm saying.

> After all it's not flatpak specific.

English is not my native language, maybe there needs to be a comma?

org.codeberg.dnkl.foot.desktop
barthalion: > The manifest is blindly spawning a terminal outside the sandbox (this seems to be for packaging only). This manifest should never reach flathub as it doesnt use flatpak internals in any sense other than packaging .

To clarify, there's no other way to ship a terminal with flatpak. The shell inside the sandbox isn't exactly as useful as the one in the host, and that's what multiple projects already do.

k6Q28k5w5eWuRuuzTg9j7GrKCbjt7: > To clarify, there's no other way to ship a terminal with flatpak. The shell inside the sandbox isn't exactly as **useful** as the one in the host, and that's what multiple projects already do.

There are ways to ship the terminal. but as you say, "useful" is the keyword here. And that is kind of my point. The feature request on upstream was created by @micro8 with the argument of **wanting sandboxing features**. Totally bypassing upstream on the very issue one created to begin with, then submitting pr claiming critierias are met. The problem isnt that the application is on flathub or how one end up with defining the manifest. -Going rogue is. Hope that clarifies.

org.codeberg.dnkl.foot.yml
microo8: Ok I removed the network build arg. Now it builds `host-spawn` from source and without network. I hope now it will be ok.

And the appstream file is on it's way https://codeberg.org/dnkl/foot/pulls/1149

tinywrkb:
Foot terminfo was only added to ncurses in the 6.2 release. Users with older version will end up with issues, maybe only input related, and blame the app.  
I highly recommend documenting the host ncurses version requirement, and its workarounds:  
1. Set TERM as `xterm-256color`. This likely need to be in the form of Flatpak override.
2. Copy the terminfo file into `~/.terminfo`. For other locations, see `man:terminfo(5)`.  
I suggest that these notes will be added to the metainfo file.

microo8:
So I've added `--env=TERM=xterm-256color` to the manifest.

Where can I find a terminfo file and how do I copy it to `~`?

tinywrkb:
> Where can I find a terminfo file and how do I copy it to `~`?

https://codeberg.org/dnkl/foot/src/tag/1.9.2/meson_options.txt#L10

tinywrkb:
> how do I copy it to ~?

This should be done by the user. If the host has ncurses 6.2 or newer, the user doesn't need to do anything.  
If it's an older release, then the user can choose to override the TERM value and set it to `xterm-256color`, or copy the terminfo file that comes with the app to `~/.terminfo`.  
In any case, this should be left to the user, which is why I suggested documenting this and not forcing it.

org.codeberg.dnkl.foot.yml
```diff
@@ -0,0 +1,67 @@
+app-id: org.codeberg.dnkl.foot
+runtime: org.freedesktop.Platform
+runtime-version: "21.08"
```
hfiguiere, line 3:
```suggestion
runtime-version: "22.08"
```

it's out now. might as well.

JakobDev: The used Sdk extension didn't provide this version currently

hfiguiere: as I said, it's out now, might as well.

I know it wasn't available before.

JakobDev: I mean `org.freedesktop.Sdk.Extension.golang` not the Freedesktop Sdk.

hfiguiere: the goland sdk is now build for 22.08 (will be available in a few hours)

barthalion:
I'm generally hesitant to accept this submission given it doesn't look like upstream is convinced to the idea.

tinywrkb:
> I'm generally hesitant to accept this submission given it doesn't look like upstream is convinced to the idea.

I see no objection in the upstream issue for the Flatpak from an actual developer who contribute code to foot.  

Main developer did not object to packaging as a Flatpak and distributing through Flathub, only to
1. Packaging recipes inside foot repo.
2. Flatpak specific code in foot.

And there's another developer there who seem to be a maintainer and his responses are neutral about this.  

So with all due respect, vocal users don't count. If upstream doesn't object, and this is fully functioning, useful to someone, and will be maintained by a passionate user of the app, then why not?  
Seeing that this is @microo8 first contribution here, my only question is the willing to maintain this packaging for the foreseen future, as there are enough abandoned apps already.

I personally don't see the point, foot is very light on dependencies, it doesn't even need GL, which is why I can easily build it statically linked, so if I wanted something portable, that's what I would do, instead of working around the sandbox.

microo8:
I will be glad to maintain foot on flathub. Now I'm playing with the new `org.freedesktop.Sdk` and updating foot and the dependencies (it doesn't want to cooperate).
Also waiting for the PR of the appstream file to be approved.

JakobDev:
>  Also waiting for the PR of the appstream file to be approved.

The AppStream file from your PR is missing screenshots. Also use `<content_rating type="oars-1.1"/>` instead of `<content_rating type="oars-1.0"/>`

microo8:
> The AppStream file from your PR is missing screenshots. Also use `<content_rating type="oars-1.1"/>` instead of `<content_rating type="oars-1.0"/>`

Thanks, I'll get to it.

orowith2os:
1.13.1 is out, so this manifest is slightly behind; additionally, the needed files are all upstream, so those need to have their metadata updated as well. 

I'm looking forward to helping improve the package and metadata and the like when possible.

microo8:
Yes, there is a new version. I've tried to update that, but is doesn't want to work. I can't figure it out on my own, so I've asked for [help](https://codeberg.org/dnkl/foot/issues/1129#issuecomment-614857). Maybe someone will know what to do.

refi64:
https://github.com/microo8/flathub/pull/1 should cover updating to runtime 22.08.

One thing I did notice is that, if I set the shell in the config file, I need to add the `host-spawn` command in front of it manually. Is this something actually desired in this case? (If not, the fix isn't too hard, it would just involve a patch to prefix everything with `host-spawn` inside the foot code, when it goes to run the shell.)

refi64:
(On that note, `foot-server` and `footclient` are also not exported atm, it's not clear to me how those are supposed to behave here.)

anthr76:
Are there any updates here?

PR #3409, opened 2022-08-11
Com.GitHub.mtkennerly.ludusavi

LeSnake04:
Its done from my perpective.

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. mtkennerly/ludusavi/issues/96


LeSnake04:
bot, build com.github.mtkennerly.ludusavi

com.github.mtkennerly.ludusavi.yaml
```diff
@@ -0,0 +1,52 @@
...
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+build-options:
+  append-path: /usr/lib/sdk/rust-stable/bin
+  env:
+    CARGO_HOME: /run/build/ludusavi/cargo
+finish-args:
+  - --socket=fallback-x11
+  - --socket=wayland
```
hfiguiere, line 13: does it really support wayland?

either way
```suggestion
  - --socket=fallback-x11
  - --socket=wayland
  - --share=ipc
```

IPC is still necessary.

com.github.mtkennerly.ludusavi.yaml
```diff
@@ -0,0 +1,52 @@
...
+  - --filesystem=host
+  ## GPU acceration
+  - --device=dri
+  ## Fix for GUI being to large on the steam deck
+  - --env=WINIT_X11_SCALE_FACTOR=1
+command: ludusavi
+modules:
+  - name: ludusavi
+    buildsystem: simple
+    sources:
```
hfiguiere, line 25: sources are usually last

hfiguiere (changes requested):
You should submit the metainfo and desktop file upstream.

hfiguiere:
bot, build com.github.mtkennerly.ludusavi

com.github.mtkennerly.ludusavi.metainfo.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+	<id>com.github.mtkennerly.ludusavi</id>
+	<metadata_license>CC0-1.0</metadata_license>
+	<project_license>MIT</project_license>
+	<name>ludusavi</name>
+	<developer_name>mtkennerly</developer_name>
+	<url type="homepage">https://rambox.app/#ce</url>
```
hfiguiere, line 8: are you sure?

ludusavi.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
+Type=Application
+Encoding=UTF-8
+Name=Ludusavi
+Comment=Tool for backing up your PC video game save data.
+Exec=ludusavi %U
+Icon=ludusavi
```
hfiguiere, line 7: You could as well set it to the right, and name the file properly. And have it that way upstream.

com.github.mtkennerly.ludusavi.metainfo.xml
LeSnake04: oh, this slipped trough when converting an old metainfo

com.github.mtkennerly.ludusavi.yaml
```diff
@@ -0,0 +1,52 @@
...
+    build-commands:
+      - cargo --offline fetch --verbose
+      - cargo build --offline --release
+      - install -D target/release/ludusavi /app/bin/ludusavi
+    post-install:
+      - install -Dm644 assets/icon.png /app/share/icons/hicolor/64x64/${FLATPAK_ID}.png
+      - install -Dm644 assets/icon.svg /app/share/icons/hicolor/scalable/${FLATPAK_ID}.svg
+      - install -Dm644 ${FLATPAK_ID}.metainfo.xml /app/share/metainfo/${FLATPAK_ID}.metainfo.xml
+      - install -Dm644 ${APPNAME}.desktop /app/share/applications/${FLATPAK_ID}.desktop
+      - desktop-file-edit --set-key=Icon --set-value=${FLATPAK_ID} /app/share/applications/${FLATPAK_ID}.desktop
```
hfiguiere, line 52:
```suggestion
      - desktop-file-edit --set-icon${FLATPAK_ID} /app/share/applications/${FLATPAK_ID}.desktop
```

But as I mentionned, since you have the .desktop here, it might as well do it right directly.

com.github.mtkennerly.ludusavi.yaml
LeSnake04: I havent tested it (use x11 bc nvidia), but it uses iced.rs, so it should work.

ludusavi.desktop
LeSnake04: outside of flatpaks its common to name  the icon APPNAME.png and my idea would be put it upstream (might only be possible once the next release comes)

LeSnake04:
I think its bugged again. 

But it builds locally just fine

com.github.mtkennerly.ludusavi.yaml
mtkennerly: It does work on Wayland. For example, some users have verified that it works using the Steam Deck's game mode, which runs Wayland.

LeSnake04:
bot, build com.github.mtkennerly.ludusavi

ludusavi.desktop
LeSnake04: I prefer to keep the desktop in sync with upstream (We now have a plan for moving the files soon, all but the yaml will become part of upstream)
when @mtkennerly is willing to change the upstream naming scheme for the future, I will gladly change the icon name in the desktop.

hfiguiere: if it is upstream then why is it here?

LeSnake04:
bot, build com.github.mtkennerly.ludusavi

LeSnake04:
I would consider this done and ready to be published. @hfiguiere

com.github.mtkennerly.ludusavi.yaml
```diff
@@ -0,0 +1,48 @@
...
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+build-options:
+  append-path: /usr/lib/sdk/rust-stable/bin
+  env:
+    CARGO_HOME: /run/build/ludusavi/cargo
+finish-args:
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --share=network
```
barthalion, line 14: What is network access needed for?

com.github.mtkennerly.ludusavi.yaml
```diff
@@ -0,0 +1,48 @@
...
+  ## Needed for custom paths and restoring saves
+  - --filesystem=host
```
barthalion, line 16: Given that it uses [specific locations](https://github.com/mtkennerly/ludusavi/blob/743e524d4337b44851f2567a56f01c04ccd90b49/src/config.rs#L273) to find saves, I don't think host access is warranted here, especially combined with network access.

LeSnake04: This program also restores gamesaves, which could be located anywere, including external drives, so i dont see how host permission can be avoided without breaking this core functianality.

com.github.mtkennerly.ludusavi.yaml
LeSnake04: The app automaticly syncs manifests of where the saves are located on launch.

com.github.mtkennerly.ludusavi.yaml
LeSnake04: reduced the permissions.

LeSnake04:
bot, build com.github.mtkennerly.ludusavi

com.github.mtkennerly.ludusavi.yaml
mtkennerly: @barthalion That function (`find_missing_roots`) is only a set of default launcher install directories, so that Ludusavi can find your installation of the Steam/Epic/etc store clients. But these are just some defaults, and you can add any other folders you like, including custom Wine prefix directories, which could be anywhere.

When it comes to finding the actual game saves, Ludusavi downloads and uses the data from this manifest file, which is updated every 24 hours: https://github.com/mtkennerly/ludusavi-manifest/blob/master/data/manifest.yaml

The saves could technically be anywhere. For example, any save file that depends on `$XDG_DATA_HOME` (`<xdgData>` in the manifest) could be in a surprising location if the user has customized that environment variable. When restoring the backed up saves, users are also able to redirect the saves to another location (e.g., the file was backed up from `/foo/...` but can be restored to `/bar/foo/...`).

mtkennerly: One example of a surprising save location: [According to PCGamingWiki](https://www.pcgamingwiki.com/wiki/NetHack#Save_game_data_location), Nethack stores its saves in `/var/games/nethack/save/`. Outliers like this are why Ludusavi needs broad file system access.

LeSnake04:
Are there some more things you would like me to change or can be pushed now?

PR #3425, opened 2022-08-19
Add Syntalos

ximion:
Hi!

- [ ] I am using only the minimal set of permissions. (See explanation below)
- [ ] All assets referenced in the manifest are redistributable by any party.  If not, the unredistributable parts are using an extra-data source type. (See comment below)
- [x] I own the domain used in the application ID or the domain has a policy for delegating subdomains.

## Notes on permissions

This software is very much IDE-like in that users can program custom Python code and use it to perform arbitrary actions that suit the scientific experiment they want to run (including calling out to scripts on the host that they may have written). The software also interacts very closely with hardware (cameras, FPGAs, Arduinos, etc.), therefore limiting its access by making full use of the sandbox will mostly annoy users and will have no benefit to the users. In addition to that, since the app was written without sandbox constraints in mind, it will also not run very well once e.g. DBus access to the system bus is restricted (some work is ongoing to fix this particular issue though).
Due to all that, Syntalos will currently request a lot of permissions in order to run.

## Notes on assets

All code shipped with this project is fully open source, except for one piece, the OpalKelly Front Panel library, which is used by one of Syntalos' modules to interface with a particular piece of hardware (and is otherwise not loaded).
The FrontPanel SDK license prohibits distribution, but according to https://docs.opalkelly.com/fpsdk/licensing/, distribution of the individual library is actually allowed and encouraged. Therefore I do not think this is an issue, but it would be nice if someone else could have a look at this situation as well.

Any feedback is very welcome, this is an extensive manifest and quite a large application (that is currently still being adjusted to play nicer when run within Flatpak).
Cheers,
    Matthias


flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64"]
```

io.github.bothlab.syntalos.yaml
```diff
@@ -0,0 +1,262 @@
...
+runtime-version: 5.15-21.08
+sdk: org.kde.Sdk
+command: syntalos
+rename-icon: syntalos
+rename-desktop-file: syntalos.desktop
+
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=x11
```
hfiguiere, line 12: it's `fallback-x11` if wayland is supported

io.github.bothlab.syntalos.yaml
```diff
@@ -0,0 +1,262 @@
...
+  - --socket=wayland
+  - --device=all
+  - --filesystem=host
+  - --socket=system-bus
+  - --socket=session-bus
```
hfiguiere, line 17: allowing specific names a much prefered

io.github.bothlab.syntalos.yaml
```diff
@@ -0,0 +1,262 @@
...
+rename-icon: syntalos
+rename-desktop-file: syntalos.desktop
+
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=x11
+  - --socket=wayland
+  - --device=all
+  - --filesystem=host
```
hfiguiere, line 15: this is a wide filesystem permission.

io.github.bothlab.syntalos.yaml
```diff
@@ -0,0 +1,262 @@
...
+    - type: archive
+      url: https://github.com/opencv/opencv/archive/refs/tags/4.6.0.tar.gz
+      sha256: 1ec1cba65f9f20fe5a41fda1586e01c70ea0c9a6d7b67c9e13edf0cfe2239277
+    - type: archive
+      url: https://github.com/opencv/opencv_contrib/archive/refs/tags/4.6.0.tar.gz
+      sha256: 1777d5fd2b59029cf537e5fd6f8aa68d707075822f90bde683fcde086f85f7a7
+    - type: patch
+      path: patches/ffmpeg_01_Add-missing-header-for-LIBAVCODEC_VERSION_INT.patch
+
+- name: libusb
```
hfiguiere, line 146: we prefer the use of shared-modules for libusb.

io.github.bothlab.syntalos.yaml
```diff
@@ -0,0 +1,262 @@
...
+
+- name: pomidaq
+  buildsystem: cmake-ninja
+  builddir: true
+  config-opts:
+    - -DPYTHON=OFF
+    - -DGUI=OFF
+  sources:
+    - type: git
+      url: https://github.com/bothlab/pomidaq.git
```
hfiguiere, line 250: you need a commit or a tag here.

modules/pyqt5.yaml
```diff
@@ -0,0 +1,51 @@
+name: pyqt5
```
hfiguiere, line 1: we have a base app for pyqt5 that allow not having to build it.

hfiguiere (changes requested):

flathub.json
ximion: It's because on ARM, Qt is only built with support for GLES, not the complete OpenGL. So either that would need to be changed, or, probably easier and better, Syntalos would need to be ported to GLES.
The bulk of direct GL calls is in https://github.com/bothlab/syntalos/blob/master/modules/canvas/imageviewwidget.cpp, but it's still annoying porting work :-/

io.github.bothlab.syntalos.yaml
ximion: The app currently works quite terribly on Wayland, as it relies on multiple windows being able to open (so the researcher can move for example the microscope display to another screen and place the control panel at a convenient spot). Think like GIMP with its original multi-window mode.
Wayland has no way for Syntalos to tell the WM where to place windows relative to each other, which results in them appearing in random places when a project is loaded (not great), and in addition to that, all windows have the same icons, making them a bit harder to distinguish.
So, while it is possible to run on Wayland, the experience is not that great yet...

hfiguiere: them disable wayland.

io.github.bothlab.syntalos.yaml
ximion: The user may want to write code which makes DBus calls to other applications on the system (like custom-built external applications) that we have no control over. So we don't know in advance which DBus interfaces are needed.
To filter interfaces, Syntalos also would need to be changed to e.g. not speak to RtKit directly but through a portal, which is also quite some work that needs to be done.

io.github.bothlab.syntalos.yaml
ximion: Yes, this is required because Syntalos will actually need to probe the host system for available external applications, available udev rules for new hardware to notify the user in case anything is missing on the host etc.
The user may likewise want to store generated data in a number of places. For our lab it's usually some drive mounted in `/mnt`, but I have also seen other labs which had a dedicated `/scratch` root directory.

modules/appstream.yaml
```diff
@@ -0,0 +1,34 @@
+name: appstream
```
hfiguiere, line 1: why do you need to build appstream?

hfiguiere:
bot, build io.github.bothlab.syntalos

io.github.bothlab.syntalos.yaml
hfiguiere: and how is it gonna launch these external applications? etc.

modules/appstream.yaml
ximion: Because `appstreamcli` is not yet included in the runtime, and we need it to compose the final metadata. It is very much a build-time-only dependency :-)

hfiguiere: this should be done for a `make dist` or equivalent.

hfiguiere: or done when you write the release note and check them in prior to the release, if you don't make tarballs.

ximion: No, it is using `appstreamcli news-to-metainfo`, which generates the final metadata from a NEWS.md file and the metainfo file. It's quite a common practice, actually, and allows you to only write your changelog once, as well as not writing it in XML which can become quite hard to read after a while. Some `make dist` like thing would only work if we would publish tarballs, not when building from a Git tag.

io.github.bothlab.syntalos.yaml
ximion: There's `flatpak-spawn --host` for that :-)

modules/appstream.yaml
hfiguiere: "quite common" doesn't make it right. This is NOT a build time thing. This is a release time thing.

Why not using tarballs?

io.github.bothlab.syntalos.yaml
hfiguiere: ah yeah with the black hole size dbus hole you put above.

io.github.bothlab.syntalos.yaml
ximion: That's an option... But that way, testing on Wayland will be much harder for people who *do* want to try it. In the long run we need some way to get this to play nice with Wayland anyway (maybe by addressing https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/72 in some way that makes everyone happy).
I need to think about this some more...

io.github.bothlab.syntalos.yaml
ximion: Exactly. Not great, but workable. This is software run in specialized environments anyway, where people need to tinker with it. That's why in this particular instance I don't see it as much of an issue as it would be with a less developer-focused tool like a mediaplayer.
So far, the software was only deployed in source or as a local deb package.

modules/pyqt5.yaml
ximion: Hmm, that would lock me into whatever runtime the pyqt5 app uses, right? That would be a bit annoying as it would hinder the Qt6 porting effort (the Python stuff is separate from the C++ stuff, so the latter can be ported to Qt6 first, while people can still make use of Qt5 in their custom Python modules).
Not having to build PyQt would be very nice though, especially since it currently is only used by one of the modules that we ship with by default.

io.github.bothlab.syntalos.yaml
hfiguiere: Flathub is meant to distribute apps that work. If wayland doesn't wayland don't activate wayland. Developers can still chose to enable it. But otherwise not. You don't even force the QPA to be xcb so basically the flatpak is configure to use a setting that doesn't work.

io.github.bothlab.syntalos.yaml
hfiguiere: you can explicitely allow RTKit. Also RTKit doesn't work through the sandbox you have to use the portal. Same DBus function, but on a different interface, and on the session bus.

modules/pyqt5.yaml
hfiguiere: not it wouldn't. there is a version for Qt6 as well.

hfiguiere:
```
builddir/files/share/appdata/io.github.bothlab.syntalos.appdata.xml: FAILED:
Validation of files failed
```

io.github.bothlab.syntalos.yaml
ximion: Fair enough. Wayland does work, but the experience is not really great while using it. I can always enable it again once we have tested this properly with Wayland and maybe compositors support multi-window apps better.

io.github.bothlab.syntalos.yaml
ximion: Exactly, it requires changes on existing code to work in the sandbox. RtKit is also not the only interface we use, and I have no idea what code other researchers are running as modules of the application - if it does DBus calls (which is actually very likely), limiting DBus access will break their custom code and result in a harder time to migrate to the Flatpak'ed Syntalos build, which is not what I want.
This application can run full, complex Python code to talk to other programs on the same machine, do online analysis of incoming data, control external hardware devices, etc.
I would very much prefer to at least initially have very few restrictions of the sandbox and then later see if we can tighten them at all (we'd still need to be able to run `flatpak-spawn` though for the "runcmd" module, and at that point limiting DBus access feels quite futile).

modules/pyqt5.yaml
ximion: That's really nice, thanks for the hint!

modules/appstream.yaml
ximion: Simple, because it's extra effort to implement a release procedure. And duplicating the same data in two places is bad too. Allowing this kind of implementation was actually one of the number-one feature requests for AppStream for a while, and is even documented there (but I explained it much better in a [blogpost](https://blog.tenstral.net/2020/03/maintain-release-info-easily-in-metainfo-files.html) once, the documentation really needs an overhaul).

For a project like this, the more stuff that is automated, the better. In natural science, it is really hard to keep a project maintained, because people finish their PhDs and take their knowledge with them, the project looses funding, the project was specific to only one department, etc. So, the less things people need to know to maintain a thing, the better (and creating a release tarball in addition to Git-tagging it is an extra step).
Also, it allows me to just use a commit-hash here until the final release with some more Flatpak support code is tagged and this build can use a Git tag instead of the hash :-)

ximion:
> ```
> builddir/files/share/appdata/io.github.bothlab.syntalos.appdata.xml: FAILED:
> • tag-missing           : <content_rating> required [use https://odrs.gnome.org/oars]
> • tag-missing           : <url> is not present
> Validation of files failed
> ```

Fixed that, along with a bunch of other recommendations that you had. I'm currently building it locally, so see if the PyQt baseapp thing worked and the modules that need it actually run :-)

hfiguiere:
bot, build io.github.bothlab.syntalos

ximion:
Looks like the autobuilder somehow has hit an ultra-rare ordering issue in the build process where ninja was trying to make an executable before a particular header was auto-generated. I've never seen it once here, locally or on our CI, but there was definitely an explicit relation missing in the respective meson file to tighten the build order. So I'm sure that this is fixed now (updated the manifest to use a later commit that includes this fix).

barthalion:
bot, build io.github.bothlab.syntalos

ximion:
Thank you very much, @barthalion ! :-)

PR #3447, opened 2022-08-29
Import Konsole 22.08.0

travier:

- [X] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub: Member of the @flathub/kde team


travier:
bot, build org.kde.konsole

travier:
On Fedora 36 KDE Wayland, the app opens for a spit second and then terminates. Nothing in the command line log.

travier:
(CC @flathub/kde for the general app submission)

travier:
With `flatpak run --user org.kde.konsole --noclose`:
```
/bin/bash: /bin/bash: cannot execute binary file
```

travier:
CC @hsitter

Mailaender:
I assume you have to bundle `bash`.

hsitter:
@travier if you could try the patch from https://invent.kde.org/utilities/konsole/-/merge_requests/728 that would be amazing.

travier:
> I assume you have to bundle `bash`.

We already do: you can try with `flatpak run --command=bash foo`

travier:
> @travier if you could try the patch from [invent.kde.org/utilities/konsole/-/merge_requests/728](https://invent.kde.org/utilities/konsole/-/merge_requests/728) that would be amazing.

Thanks, will push it here.

travier:
bot, build org.kde.konsole

travier:
OK, this fixes the issue and now I get an interactive shell. Thanks @hsitter!
However, I can not run true terminal commands such as `htop`:
```
$ htop
Error opening terminal: unknown.
```

Outside:
```
$ echo $TTY
/dev/pts/2
```

Inside:
```
$ echo $TTY

```

Maybe we need something like what's done in https://github.com/1player/host-spawn instead of using flatpak-spawn.

travier:
We might need a way to change https://invent.kde.org/frameworks/kcoreaddons/-/blob/master/src/lib/util/ksandbox.cpp#L49 without having to recompile the whole Runtime.

hsitter:
For testing purposes we could just build kcoreaddons as part of the konsole manifest and see where that takes us. If host-spawn works better for us I can look into giving us more flexibility in ksandbox

TheEvilSkeleton:
Interestingly, when launching, I get this error when using fish by default:
```
Warning: Could not find '/bin/fish', starting '/bin/sh' instead.  Please check your profile settings.
```

However, `/bin/fish` does indeed exist:
```
sh-5.1$ ls /bin/fish
/bin/fish
```

Running `fish` outputs this:
```
sh-5.1$ /bin/fish
warning: Could not set up terminal.
warning: TERM environment variable not set.
```

hfiguiere:
Shouldn't the terminal just open a pty that will actually run everything on the host outside the sandbox?

hsitter:
Curiously when I open konsole from the menu htop works as expected, meanwhile when started from a host tty it is broken. Maybe this problem actually has to do with the fact that we can't have a controlling tty (ctty) inside the sandbox and so the ctty ends up being the host konsole and thus things are getting confused?

travier:
bot, build org.kde.konsole

tsdgeos:
I still get nothing with
 flatpak run org.kde.konsole

travier:
Sorry, the last push is broken. I've try with host-spawn and I get better results. I'll work on a patch for ksandbox to use that as an option.

travier:
I'm currently fighting to build the Go binary from source in the Flatpak.

This one from @hsitter has good results: https://github.com/hsitter/flathub/tree/org.kde.konsole.
Weirdly, running `host-spawn` directly from `flatpak run --command=bash org.kde.konsole` works well with resizes, but running it in Konsole does not work well.

travier:
We could figure out some scheme to pass env variables (https://github.com/1player/host-spawn/blob/master/main.go#L74).

travier:
bot, build org.kde.konsole

travier:
Latest push does not work for resizes but spawns a shell at least.

1bcb:
The newest build works for me in Mint 21 Cinnamon.

Newbytee:
bot, build org.kde.konsole

1bcb:
This one also works for me on Mint 21 Cinnamon.

Newbytee:
> This one also works for me on Mint 21 Cinnamon.

Yeah, it should be identical. I just rebuilt it because the previous one had expired and I wanted to test it.

travier:
Things should be a little bit better now. We have two remaining known issues:
* crashes on right click
* wrong check for binary for non custom profiles

travier:
https://invent.kde.org/utilities/konsole/-/merge_requests/747

travier:
bot, build org.kde.konsole

TheEvilSkeleton:
bot, build org.kde.konsole

travier:
See https://github.com/flathub/flathub/pull/3447#issuecomment-1264652401 & https://github.com/flathub/flathub/pull/3447#issuecomment-1264693778 which should be resolved before me merge this one.

Waqar144:
Merge requests to fix two issues the one mentioned above:

- https://invent.kde.org/utilities/konsole/-/merge_requests/829
- https://invent.kde.org/utilities/konsole/-/merge_requests/829

Hoping to get both of these into 23.04

travier:
Thanks!

travier:
bot, build org.kde.konsole

travier:
Looks great to me! Thanks @Waqar144 !

travier:
There is a one remaining issue but it's not critical: Closing a tab will ask you to confirm as Konsole thinks that there is still a program running.

org.kde.konsole.json
```diff
@@ -0,0 +1,46 @@
...
+    "runtime-version": "5.15-22.08",
+    "sdk": "org.kde.Sdk",
+    "command": "konsole",
+    "rename-icon": "utilities-terminal",
+    "finish-args": [
+        "--device=dri",
+        "--share=ipc",
+        "--socket=wayland",
+        "--socket=fallback-x11",
+        "--talk-name=org.a11y.Bus",
```
Erick555, line 13: This should be removed, see https://github.com/flathub/org.kde.ark/commit/ec438fb43c7ec1f71b0c946f5c54232723a7f8ba

Waqar144:
> There is a one remaining issue but it's not critical: Closing a tab will pop a confirmation dialog as Konsole thinks that there is still a program running.

Simply avoiding the popup is easy but that doesn't fix anything and can lead to the user accidentally closing  a tab when a process was running.

Konsole needs to know about any foreground running processes. This doesn't seem possible in a direct way right now.

org.kde.konsole.json
travier: Good catch

travier:
bot, build org.kde.konsole

Waqar144:
Also, not sure yet as I haven't tested but we might need `filesystem=host:ro` `filesystem=/usr/include:ro`. Konsole has a feature where it scans the text for possible links to files and then if you Ctrl+Click on a file it can open it for you using the default program.

hfiguiere:
`filesystem=/usr/include:ro` this won't work

Erick555:
The `:ro` would also make everything unwritable unless sandbox is escaped through flatpak-spawn.

Waqar144:
> The :ro would also make everything unwritable ,,

Write access is not needed as far as I understand / know.

The reason it needs access is because it regex matches the scrollback history and looks for files that match the current directory's files and allows you to access them. For relative paths this is completely broken atm.

For absolute it does highlight the links, but isn't able to do much. Probably a bug somewhere else in the KDE frameworks. Konsole doesn't do much checking for absolute paths.

Erick555:
> Write access is not needed as far as I understand / know.
For this feature or at all? There won't be write access at all.

travier:
We're already running everything outside the sandbox via flatpak-spawn. Konsole should check if files exists that way. I don't think we should add more holes in the sandbox as using two different mechanisms to look for things will definitely be confusing.

While I agree that this is a bug, I don't think it should stop including it in Flathub for now.

Waqar144:
Yes, its not a blocker.

travier:
> Konsole has file backed scrollback history which is stored somewhere (not sure where atm). That might need write access to a specific directory. Maybe its stored in `/tmp`, will have to check

That should stay in the sandbox.

travier:
bot, build org.kde.konsole

travier:
This is working well for me. Should be ready now.

tsdgeos:
are we aiming beta or "normal"?

travier:
I think we still have https://github.com/flathub/flathub/pull/3447#issuecomment-1469803713 (not re-checked), so maybe Beta?

tsdgeos:
I'm good with beta :)

TheEvilSkeleton:
> I think we still have [#3447 (comment)](https://github.com/flathub/flathub/pull/3447#issuecomment-1469803713) (not re-checked), so maybe Beta?

I tested on my end and I can reproduce it

Waqar144:
Keeping it in beta is okay, but that generally means close to 0 real users. IMO, konsole is fairly usable now so we should aim for stable so that more people are able to use it and test it.

travier:
The remaining issue is not critical I agree. I does not break the experience, just makes it less fluid. I'm OK with stable or beta, both will be better than keeping it here

travier:
Let's merge it in stable and track the issue? @tsdgeos

travier:
Thanks!

PR #3449, opened 2022-08-30
Add io.sourceforge.atanks

Mailaender:
Deprecate https://github.com/flathub/net.sourceforge.atanks/pull/8 as upstream changed the application ID as SourceForge changed their URLs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+build-dir/
+.flatpak-builder/
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "shared-modules"]
+	path = shared-modules
+	url = https://github.com/flathub/shared-modules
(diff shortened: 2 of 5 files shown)


Mailaender:
bot, build io.sourceforge.atanks

Mailaender:
bot, build io.sourceforge.atanks

barthalion:
bot, build io.sourceforge.atanks

barthalion:
I don't think it makes sense to rename the app only because Sourceforge changed its domain. Renaming is generally pretty disruptive process so we tend to avoid it unless the app is maintained by upstream.

Mailaender:
Upstream changed the name in their metainfo file and as per policy this is maintained by them it leaves you no choice.

hfiguiere:
`rename-appdata-file` will actually take care of it.

hfiguiere:
it leave all the choice, including patching.

Mailaender:
I think that should be disallowed via @flathub guidelines.

PR #3452, opened 2022-08-31
Add Hidamari

jeffshee:


io.github.jeffshee.Hidamari.json
```diff
@@ -0,0 +1,240 @@
...
+        "--talk-name=org.gnome.Shell",
+        "--talk-name=org.gnome.Shell.Extensions",
+        "--talk-name=org.gnome.SessionManager",
+        "--talk-name=org.gnome.ScreenSaver",
+        "--talk-name=org.cinnamon.ScreenSaver",
+        "--talk-name=org.kde.screensaver",
+        "--talk-name=org.freedesktop.ScreenSaver",
+        "--talk-name=org.kde.StatusNotifierWatcher",
+        "--talk-name=org.freedesktop.Flatpak",
+        "--system-talk-name=org.freedesktop.login1"
```
hfiguiere, line 27: please why you need all of these D-Bus permissions.

io.github.jeffshee.Hidamari.json
```diff
@@ -0,0 +1,240 @@
...
+                    "type": "archive",
+                    "url": "https://github.com/jeffshee/hidamari/archive/refs/tags/v3.0-beta1-2.tar.gz",
+                    "sha256": "9050096f0a01de4a949259d8a40d3f2bc3f905604114fb608552b85bcf7ec0b8"
+                }
+            ]
+        }
+    ],
+    "build-options": {
+        "env": {}
+    }
```
hfiguiere, line 239:
```suggestion
    ]
```

io.github.jeffshee.Hidamari.json
```diff
@@ -0,0 +1,240 @@
...
+    "runtime-version": "42",
+    "sdk": "org.gnome.Sdk",
+    "command": "hidamari",
+    "finish-args": [
+        "--share=network",
+        "--share=ipc",
+        "--device=dri",
+        "--socket=x11",
+        "--socket=pulseaudio",
+        "--filesystem=home:ro",
```
JakobDev, line 13: Why is this needed?

io.github.jeffshee.Hidamari.json
```diff
@@ -0,0 +1,240 @@
...
+        "--filesystem=xdg-config/hidamari",
+        "--filesystem=xdg-config/autostart",
+        "--own-name=io.github.jeffshee.Hidamari.player",
+        "--own-name=io.github.jeffshee.Hidamari.server",
```
hfiguiere, line 17:
```suggestion
```

`io.github.jeffshee.Hidamari.*` is granted by default

io.github.jeffshee.Hidamari.json
```diff
@@ -0,0 +1,240 @@
...
+    "sdk": "org.gnome.Sdk",
+    "command": "hidamari",
+    "finish-args": [
+        "--share=network",
+        "--share=ipc",
+        "--device=dri",
+        "--socket=x11",
+        "--socket=pulseaudio",
+        "--filesystem=home:ro",
+        "--filesystem=xdg-config/hidamari",
```
hfiguiere, line 14: why do you need this?

io.github.jeffshee.Hidamari.json
```diff
@@ -0,0 +1,240 @@
...
+        "--filesystem=xdg-config/autostart",
```
hfiguiere, line 15: or this?

io.github.jeffshee.Hidamari.json
JakobDev, line 14: Yous shoudl respect the [XDG Base Directry Spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). Flatpak sets some env vars to point to custom directories, so you should not save in .config, if the App is run as Flatpak.

io.github.jeffshee.Hidamari.json
```diff
@@ -0,0 +1,240 @@
...
+        "--own-name=io.github.jeffshee.Hidamari.server",
+        "--talk-name=org.gnome.Shell",
+        "--talk-name=org.gnome.Shell.Extensions",
+        "--talk-name=org.gnome.SessionManager",
+        "--talk-name=org.gnome.ScreenSaver",
+        "--talk-name=org.cinnamon.ScreenSaver",
+        "--talk-name=org.kde.screensaver",
+        "--talk-name=org.freedesktop.ScreenSaver",
+        "--talk-name=org.kde.StatusNotifierWatcher",
+        "--talk-name=org.freedesktop.Flatpak",
```
JakobDev, line 26: Please also explain this very broad permission

io.github.jeffshee.Hidamari.json
jeffshee: @hfiguiere Hi, thanks for the review.
Hidamari is a live wallpaper application.
 `org.*.ScreenSaver` is for monitoring the event of screen lock. When a screen lock happens, the app will pause the live wallpaper playback.
 `org.kde.StatusNotifierWatcher` Apparently it is a requirement for displaying the systray icon, which is an option of the app.
 `org.freedesktop.Flatpak` is required to change the system wallpaper by calling `gsettings` on the host. Currently, this is the only option due to the limitation of Flatpak I believe.
 `org.freedesktop.login1` is for monitoring the event of the session ending (power-off, restart, etc). The app has to restore the system wallpaper to the user's original one before exiting.

io.github.jeffshee.Hidamari.json
jeffshee: Thanks!

io.github.jeffshee.Hidamari.json
jeffshee: `xdg-config/hidamari` is the location the app writes its config file, and some temporary files it needs.

io.github.jeffshee.Hidamari.json
hfiguiere: xdg-picture and maybe xdg-videos should be enough.

jeffshee: Hi @JakobDev , thanks for the review.
Which part are you referring to? (it seems normal to me)
Anyway, `socket=x11` is required as the app needs to run in x11. It needs to show a window with a `WINDOW_TYPE_HINT_DESKTOP`, which is not supported in Wayland.
`--filesystem=home:ro` is required to (1) read user video files, and (2) read an HTML file that could be anywhere within the user directory.
`--share=network` is needed for the video streaming function.

io.github.jeffshee.Hidamari.json
jeffshee: The app needs to write a `.desktop` file into that directory if the user wants the app to autostart after login.

io.github.jeffshee.Hidamari.json
jeffshee: @hfiguiere Yeah for (1) `xdg-videos` should be enough, but for (2) it isn't. That's why it needs to be `home:ro`.

io.github.jeffshee.Hidamari.json
jeffshee: Answered here. https://github.com/flathub/flathub/pull/3452#discussion_r959700213

io.github.jeffshee.Hidamari.json
JakobDev: >  Which part are you referring to? (it seems normal to me)

Only the `--filesystem=home` part

> (1) read user video files, and (2) read an HTML file that could be anywhere within the user directory.

That's sounds like the User can pick a File. On Gtk (which looks like you are using) you get the access to the file when the User picks it, so you don't need this.

io.github.jeffshee.Hidamari.json
jeffshee: Hi, I'm not sure about this. Aren't apps writing their config data in `~/.config/foobar` directory? What is the alternative you would suggest?

hfiguiere:
also the app-id in that appdata file is wrong. `.desktop` isn't part of the app-id. (it's still tolerated by the validation due to legacy)

io.github.jeffshee.Hidamari.json
jeffshee: Isn't that `xdg-config` = `~/.config`? The app writes its config data into `~/.config/hidamari/`

hfiguiere: if you write to `~/.config/foobar` you are doing it wrong. If you write to what XDG gives you (which CAN be that), then you don't need this.

I do suggest your remove it.

io.github.jeffshee.Hidamari.json
hfiguiere: > @hfiguiere Yeah for (1) `xdg-videos` should be enough, but for (2) it isn't. That's why it needs to be `home:ro`.

then xdg-documents.

io.github.jeffshee.Hidamari.json
jeffshee: @hfiguiere Solved in the last commit.

io.github.jeffshee.Hidamari.json
jeffshee: Solved. Now it only use `xdg-videos/Hidamari:create`

io.github.jeffshee.Hidamari.json
jeffshee: Solved. Now it is using `$XDG_CONFIG_HOME/hidamari`

io.github.jeffshee.Hidamari.json
```diff
@@ -0,0 +1,232 @@
...
+            "sources": [
+                {
+                    "type": "file",
+                    "url": "https://files.pythonhosted.org/packages/98/79/85f33652bb51b3b161d609966cb894f4b3104b70a3fba00f7df09bdf8411/python_vlc-3.0.16120-py3-none-any.whl",
+                    "sha256": "c409afb38fe9f788a663b4302ca583f31289ef0860ab2b1668da96bbe8f14bfc"
+                }
+            ]
+        },
+        {
+            "name": "python3-yt-dlp",
```
hfiguiere, line 52: if these python dependencies have been generated by the `flatpak-builder-tools`, then it's better to have them in the separat manifest so that it's easier to update.

hfiguiere:
bot, build io.github.jeffshee.Hidamari

io.github.jeffshee.Hidamari.json
jeffshee: Thanks for the suggestion. I will try that next time! 👍

hfiguiere:
build failure because it can't find the icon.

jeffshee:
@hfiguiere Is it a requirement to have PNG icons as well if there is already an SVG icon?

hfiguiere:
SVG ought to be enough.

hfiguiere:
are you sure the value is correct in the appdata ?
```
  <launchable type="desktop-id">@APPLICATION_ID@.desktop</launchable>
```

Is `@APPLICATION_ID@` the same as the app-id at the top which is verbatim...

JakobDev:
The metainfo file ends with .in. That's a sign that this file goes through a compiler step first. @APPLICATION_ID@ should be replaced there with the ID. The AppStream validation didn't see a problem.

jeffshee:
Thanks. It could be some mistake in the meson file. I'll replace that `@APPLICATION_ID@` with the actual id and try again...

jeffshee:
The test build works fine.

The screenshots are broken in gnome-software though.
Screenshot from 2022-09-02 08-14-00
Also, the detail of the release doesn't show.
Screenshot from 2022-09-02 08-15-58

I suppose gnome-software pulls these info from the AppData, isn't it? If that is the case, the AppData seems right to me.
https://github.com/jeffshee/hidamari/blob/flatpak/data/io.github.jeffshee.Hidamari.appdata.xml.in
Not sure if this is normal for the test repo.

barthalion:
> org.freedesktop.Flatpak is required to change the system wallpaper by calling gsettings on the host. Currently, this is the only option due to the limitation of Flatpak I believe.

This is not true; see https://beta.flathub.org/apps/details/im.bernard.Nostalgia for an example how to use wallpaper portal.

jeffshee:
> > org.freedesktop.Flatpak is required to change the system wallpaper by calling gsettings on the host. Currently, this is the only option due to the limitation of Flatpak I believe.
> 
> This is not true; see https://beta.flathub.org/apps/details/im.bernard.Nostalgia for an example how to use wallpaper portal.

@barthalion It has to be without user interaction. AFAIK wallpaper portal summon a set wallpaper dialog which is not acceptable in my use case.

barthalion:
It doesn't seem to work correctly either way. Applying web page ends up with this trace:

```
  File "/app/share/hidamari/hidamari/gui/control.py", line 171, in on_local_web_page_apply
    file_path = choose.get_path()
AttributeError: 'NoneType' object has no attribute 'get_path'
```

jeffshee:
> It doesn't seem to work correctly either way. Applying web page ends up with this trace:
> 
> ```
> Traceback (most recent call last):
>   File "/app/share/hidamari/hidamari/gui/control.py", line 171, in on_local_web_page_apply
>     file_path = choose.get_path()
> AttributeError: 'NoneType' object has no attribute 'get_path'
> ```

Thanks! I will look into this later. 🐞

jeffshee:
@barthalion I checked and the local web page function is working fine. You got that error probably because you didn't select any HTML file before clicking Apply. You could, and have to select an HTML file by clicking on the FileChooser widget, which is next to the Apply button.

@hfiguiere @JakobDev So I think I had solved most of the issues here, except for the gnome-software issue mentioned above. https://github.com/flathub/flathub/pull/3452#issuecomment-1234905950 Is there anything I need to do?

jeffshee:
bot, build io.github.jeffshee.Hidamari

io.github.jeffshee.Hidamari.json
jeffshee: Done!

PR #3483, opened 2022-09-08
Add com.github.marhkb.Pods

marhkb:

  - `--filesystem=/run/systemd/journal` for writing logs
  - `--filesystem=xdg-run/podman:ro` to use the podman unix socket for the RESTful API

I only want to use the beta branch for now, until I have a first stable release sometime in the future. I think releasing on Flathub Beta, that's a good way to gather user feedback.


marhkb:
bot, build com.github.marhkb.Pods.Devel

com.github.marhkb.Pods.Devel.json
```diff
@@ -0,0 +1,98 @@
...
+                        }
+                    ]
+                }
+            ]
+        },
+        {
+            "name": "pods",
+            "buildsystem": "meson",
+            "config-opts": [
+                "-Dprofile=development"
```
hfiguiere, line 86: this should be release.

com.github.marhkb.Pods.Devel.json
```diff
@@ -0,0 +1,98 @@
...
+    "runtime-version": "42",
+    "sdk": "org.gnome.Sdk",
+    "command": "pods",
+    "finish-args": [
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--share=network",
+        "--share=ipc",
+        "--device=dri",
+        "--env=RUST_LOG=pods=debug",
```
hfiguiere, line 13: are you sure it should be that verbose?

com.github.marhkb.Pods.Devel.json
```diff
@@ -0,0 +1,98 @@
+{
+    "app-id": "com.github.marhkb.Pods.Devel",
```
hfiguiere, line 2: it should be the "release" app-id.

hfiguiere:
not sure why it doesn't find rustc on aarch64.

com.github.marhkb.Pods.Devel.json
marhkb: Done. Omitting this is enough for producing a release build.

com.github.marhkb.Pods.Devel.json
marhkb: This is a leftover from early days. The application actually has a cli argument for this and in case it is omitted, it uses `info` for release builds and `debug` for debug builds. So, I've just removed that line.

com.github.marhkb.Pods.Devel.json
marhkb: Done.
One thing that is unclear to me:
I had assumed that I need to change it later back to `*.Devel` on the beta branch to make beta releases distinguishable from production builds. But other manifests I found in beta branches don't do this ([Example](https://github.com/flathub/org.gnome.Fractal/tree/beta)). Does the build bot take care of this?

marhkb:
bot, build com.github.marhkb.Pods

com.github.marhkb.Pods.Devel.json
hfiguiere: definitely not. It won't work. The app-id is the app-id.

Also .Devel is not for beta, Devel is for development. Flathub isn't for development.

hfiguiere:
this build will fail as it won't be able to find the manifest that has the wrong name.

com.github.marhkb.Pods.Devel.json
marhkb: So I guess it's not possible for users to install a stable and a beta version at the same time, because the app-id is not different?

hfiguiere: yes it is. Different branches. They'll share the config directory too.

com.github.marhkb.Pods.json
```diff
@@ -0,0 +1,94 @@
...
+                "-Dtests=false",
+                "-Dexamples=false",
+                "-Dvapi=false",
+                "-Dintrospection=disabled"
+            ],
+            "sources": [
+                {
+                    "type": "git",
+                    "url": "https://gitlab.gnome.org/GNOME/libadwaita.git",
+                    "tag": "1.2.beta",
```
hfiguiere, line 46: I'll assume you really need this one instead of the version in the SDK.

marhkb: Yes, I am using stuff that was introduced with 1.2

marhkb:
Just noticed that I should update the release meta data and content_rating could be set to `oars-1.1`

https://github.com/marhkb/pods/blob/8129761acc082a00d28be0b8e85de81be19a0b47/data/com.github.marhkb.Pods.metainfo.xml.in.in#L61

marhkb:
bot, build com.github.marhkb.Pods

PR #3490, opened 2022-09-09
Add com.github.Inochi2D.Inochi-Creator

orowith2os:

1. `system-bus` may not be needed, but the app complains if it isn't given that permission. Better safe than sorry, I figure.
2. `host` also is not REQUIRED, but the app doesn't yet follow portals, so I've given it full access until then.

1. The logos and branding are not redistributable without permission, but they can be toggled on and off with a build option. 

- [ ] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** Not a link, but it's been worked on by separate package maintainers for a different build, such as maintainer of the COPR build. The project founder also is fine with it.

1. Not really any, but there isn't a metadata file yet. When the kinks have been worked out, I'll submit that and should be good to go.

### DIFF
--- a/com.github.Inochi2D.Inochi-Creator.yml
+++ b/com.github.Inochi2D.Inochi-Creator.yml
@@ -11,12 +11,8 @@ finish-args:
     - "--socket=x11"
     - "--share=ipc"
     - "--socket=wayland"
-    - "--filesystem=host" #This is a dev tool, it should probably have as much access to the host as it can.
-    - "--filesystem=home" #So that it can access the project files, should they be located in the home folder
-    - "--filesystem=~/inochi-creator-crashdump.txt:rw" # Error logs
-#    - "--filesystem=~/.config/inochi-creator:rw" #configuration files, not needed, as Inochi follows the appropriate environment variables.
-    # - "--socket=session-bus"
-    - "--socket=system-bus"
+    - "--filesystem=host" #This is a dev tool, it should probably have as much access to the host as it can. It also doesn't yet utilize portals.
+#    - "--socket=system-bus"
 modules:
   # --- Tools ---
   - name: xprop
@@ -37,6 +33,7 @@ modules:
     sources:
       - type: git
         url: https://github.com/KDE/kdialog.git
+        tag: v22.08.1
 
 
   # --- Inochi Creator ---
--- a/com.github.Inochi2D.Inochi-Creator.yml
+++ b/com.github.Inochi2D.Inochi-Creator.yml
@@ -1,0 +1,164 @@
+id: com.github.Inochi2D.Inochi-Creator
+runtime: org.kde.Platform
+runtime-version: "5.15-21.08"
+sdk: org.kde.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.ldc
+  - org.freedesktop.Sdk.Extension.dmd
+command: "inochi-creator"
+finish-args:
+    - "--device=all" #OpenGL rendering, webcams (may be required? not sure)
+    - "--socket=x11"
+    - "--share=ipc"
+    - "--socket=wayland"
+    - "--filesystem=host" #This is a dev tool, it should probably have as much access to the host as it can.
+    - "--filesystem=home" #So that it can access the project files, should they be located in the home folder
+    - "--filesystem=~/inochi-creator-crashdump.txt:rw" # Error logs
+#    - "--filesystem=~/.config/inochi-creator:rw" #configuration files, not needed, as Inochi follows the appropriate environment variables.
+    # - "--socket=session-bus"
+    - "--socket=system-bus"
+modules:
+  # --- Tools ---
+  - name: xprop
+    sources: 
+      - type: archive
+        url: https://xorg.freedesktop.org/archive/individual/app/xprop-1.2.4.tar.bz2
+        sha256: "8c77fb096e46c60032b7e2bde9492c3ffcc18734f50b395085a5f10bfd3cf753"
+
+  - name: zenity
+    buildsystem: meson
+    sources:
+      - type: archive
+        url: https://download.gnome.org/sources/zenity/3.41/zenity-3.41.0.tar.xz
+        sha256: 19b676c3510e22badfcc3204062d432ba537402f5e0ae26128c0d90c954037e1
+
+  - name: kdialog
+    buildsystem: cmake
+    sources:
+      - type: git
+        url: https://github.com/KDE/kdialog.git
+
+
+  # --- Inochi Creator ---
+  - name: inochi-creator
+    buildsystem: simple
+    build-options:
+      append-path: /usr/lib/sdk/ldc/bin
+    build-commands:
+      - dub add-local ./deps/bindbc-imgui/ 0.7.0+build.20.g237a209;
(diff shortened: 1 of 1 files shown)


orowith2os:
The main issue currently is that it doesn't pull up kdialog when you go to save or open a file. It used to work, but I don't know what caused it to break.

com.github.Inochi2D.Inochi-Creator.yml
```diff
@@ -0,0 +1,164 @@
...
+    - "--device=all" #OpenGL rendering, webcams (may be required? not sure)
+    - "--socket=x11"
+    - "--share=ipc"
+    - "--socket=wayland"
+    - "--filesystem=host" #This is a dev tool, it should probably have as much access to the host as it can.
+    - "--filesystem=home" #So that it can access the project files, should they be located in the home folder
+    - "--filesystem=~/inochi-creator-crashdump.txt:rw" # Error logs
+#    - "--filesystem=~/.config/inochi-creator:rw" #configuration files, not needed, as Inochi follows the appropriate environment variables.
+    # - "--socket=session-bus"
+    - "--socket=system-bus"
```
hfiguiere, line 19: you should determine what bus name is needed instead.

com.github.Inochi2D.Inochi-Creator.yml
```diff
@@ -0,0 +1,164 @@
...
+  - org.freedesktop.Sdk.Extension.ldc
+  - org.freedesktop.Sdk.Extension.dmd
+command: "inochi-creator"
+finish-args:
+    - "--device=all" #OpenGL rendering, webcams (may be required? not sure)
+    - "--socket=x11"
+    - "--share=ipc"
+    - "--socket=wayland"
+    - "--filesystem=host" #This is a dev tool, it should probably have as much access to the host as it can.
+    - "--filesystem=home" #So that it can access the project files, should they be located in the home folder
```
hfiguiere, line 15: 1. this is broad
2. host include home

com.github.Inochi2D.Inochi-Creator.yml
```diff
@@ -0,0 +1,164 @@
...
+sdk: org.kde.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.ldc
+  - org.freedesktop.Sdk.Extension.dmd
+command: "inochi-creator"
+finish-args:
+    - "--device=all" #OpenGL rendering, webcams (may be required? not sure)
+    - "--socket=x11"
+    - "--share=ipc"
+    - "--socket=wayland"
```
hfiguiere, line 13: does it support wayland?

com.github.Inochi2D.Inochi-Creator.yml
```diff
@@ -0,0 +1,164 @@
...
+    sources:
+      - type: archive
+        url: https://download.gnome.org/sources/zenity/3.41/zenity-3.41.0.tar.xz
+        sha256: 19b676c3510e22badfcc3204062d432ba537402f5e0ae26128c0d90c954037e1
+
+  - name: kdialog
+    buildsystem: cmake
+    sources:
+      - type: git
+        url: https://github.com/KDE/kdialog.git
```
hfiguiere, line 39: you need a commit or tag here.

com.github.Inochi2D.Inochi-Creator.yml
```diff
@@ -0,0 +1,164 @@
...
+        dub add-local ./deps/bindbc-opengl/ 1.0.2;
+        dub add-local ./deps/bindbc-sdl/ 1.1.3;
+        dub add-local ./deps/imagefmt/ 2.1.2;
+        dub add-local ./deps/mir-algorithm/ 3.14.11;
+        dub add-local ./deps/mir-core/ 1.1.109;
+        dub add-local ./deps/semver/ 0.3.4;
+        dub add-local ./deps/silly/ 1.1.1;
+        dub add-local ./deps/tinyfiledialogs/ 0.10.1;
+        dub build --skip-registry=all --compiler=ldc2 --build=release --config=full;
+      - install -m 755 ./out/inochi-creator /app/bin/inochi-creator
```
hfiguiere, line 67: so I guess you don't install mandatory icons, .desktop, appdata files.

orowith2os: Indeed, I have not yet sorted those out. I want to get the app in a functioning state without them first.

com.github.Inochi2D.Inochi-Creator.yml
```diff
@@ -0,0 +1,164 @@
...
+  - org.freedesktop.Sdk.Extension.dmd
+command: "inochi-creator"
+finish-args:
+    - "--device=all" #OpenGL rendering, webcams (may be required? not sure)
+    - "--socket=x11"
+    - "--share=ipc"
+    - "--socket=wayland"
+    - "--filesystem=host" #This is a dev tool, it should probably have as much access to the host as it can.
+    - "--filesystem=home" #So that it can access the project files, should they be located in the home folder
+    - "--filesystem=~/inochi-creator-crashdump.txt:rw" # Error logs
```
hfiguiere, line 16: this doesn't work. It can be a file. Also since you already grant `home`... it pointless.

com.github.Inochi2D.Inochi-Creator.yml
```diff
@@ -0,0 +1,164 @@
...
+#    - "--filesystem=~/.config/inochi-creator:rw" #configuration files, not needed, as Inochi follows the appropriate environment variables.
+    # - "--socket=session-bus"
```
hfiguiere, line 18:
```suggestion
```

com.github.Inochi2D.Inochi-Creator.yml
orowith2os: It does, and it does some checks to see what it's running on: https://github.com/Inochi2D/inochi-creator/commit/62099f3c6228a34c1d0127abd51a78802d5bddde

I have X and Wayland as permissions since KDialog seems to take longer to start if one or the other is blocked.

com.github.Inochi2D.Inochi-Creator.yml
orowith2os: This is addressed by the fact that the app does not yet utilize portals, so I give it access to everything to cover where the project files could be located. As for having both of them set, there's not really a specific reason.

com.github.Inochi2D.Inochi-Creator.yml
orowith2os: I have the permission set, specifically, just in case the user revokes home/host access when portal use is implemented. Easier to set it specifically and let the user come to the conclusion that it's more or less required for logs, so that they don't remove the permission.


Also, why would it not work? Would I need to use :create instead?

hfiguiere: I meant "It can NOT be a file". Again, no point having it.

orowith2os:
Turns out that it doesn't complain anymore if I remove the dbus permissions. Maybe it was kdialog that was complaining?

com.github.Inochi2D.Inochi-Creator.yml
hfiguiere: it's required for a build

orowith2os:
Interesting, it seems that the issue is still there even on earlier commits. I'll close this and wait until everything's implemented to ensure it works properly in the sandbox, then.

PR #3510, opened 2022-09-18
Add io.github.Foldex.AdwSteamGtk

Foldex:


Foldex:
bot, build io.github.Foldex.AdwSteamGtk

hfiguiere:
bot, build o.github.Foldex.AdwSteamGtk

Foldex:
Had a typo in Appdata's `desktop-id` which broke icon gen. Should be good now.

Foldex:
bot, build io.github.Foldex.AdwSteamGtk

Foldex:
Build looks good to me

barthalion:
It doesn't work on my end with flatpaked Steam:

```
Directory /var/home/bpiotrowski/.var/app/com.valvesoftware.Steam/.steam/steam does not exist
Directory /var/home/bpiotrowski/.steam/steam does not exist
```

This is probably because `skins` directory doesn't exist.

Foldex:
Could you check where the `~/.var/app/com.valvesoftware.Steam/.steam/steam` symlink is pointing to?
It worked fine on my flatpak install, but I did dig up an older install and test with that.
My newer install points to `~/.var/app/com.valvesoftware.Steam/.local/share/Steam` while the older install points to `~/.local/share/Steam`.
Deleting it and relaunching Steam did seem to regenerate the symlink so that the install worked. but it's probable that the upstream install should target the flatpak dir directly instead of using a symlink since the sandbox complicates resolving those correctly.

Foldex:
I widened the filesystem permissions so the install can correctly check the parent exists and create the skins dir when necessary.

hfiguiere:
You should just have added `:create` to have it created automatically.

Foldex:
To my understanding `:create` would always create the directories, which would mean the installer would check that those dirs exist and always install to both the native steam and the flatpak paths. It'd work, but I'd prefer to not litter the user's home dir if they're only using the flatpak version of steam and vice versa.

PR #3529, opened 2022-09-26
Net.hovancik.stretchly

hovancik:


jwillikers:
bot, build net.hovancik.stretchly

JakobDev:
The App ID is case sensitive

bot, build net.hovancik.Stretchly

.github/workflows/auto-update.yaml
```diff
@@ -0,0 +1,65 @@
+name: Auto update
```
hfiguiere, line 1: you shouldn't have these in the flathub repo

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+	"automerge-flathubbot-prs": true
```
hfiguiere, line 2: nor this.

net.hovancik.Stretchly.yaml
```diff
@@ -0,0 +1,57 @@
...
+      append-path: /usr/lib/sdk/node14/bin
+      env:
+        XDG_CACHE_HOME: /run/build/stretchly/flatpak-node/cache
+        npm_config_cache: /run/build/stretchly/flatpak-node/npm-cache
+        npm_config_nodedir: /usr/lib/sdk/node14
+        npm_config_offline: 'true'
+    subdir: main
+    sources:
+      - type: dir
+        path: .
```
hfiguiere, line 37: not sure what you are trying to do here, but surely this doesn't look right at all.

stretchly-sources.yaml
```diff
@@ -0,0 +1,10 @@
+  - type: git
+    url: https://github.com/hovancik/stretchly.git
```
hfiguiere, line 2: this shouldn't be in a separate manifest since it looks like it's the main source.

net.hovancik.Stretchly.yaml
```diff
@@ -0,0 +1,57 @@
+app-id: net.hovancik.Stretchly
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
```
hfiguiere, line 3: 22.08 maybe?

hfiguiere (changes requested):

net.hovancik.Stretchly.yaml
jwillikers: There is no `org.freedesktop.Sdk.Extension.node14/x86_64/22.08` yet, otherwise I would update it to `22.08`.

net.hovancik.Stretchly.yaml
hfiguiere: this doesn't do what you think it does. `mkdir -p` can be used to create the directory.

.github/workflows/auto-update.yaml
jwillikers: Removed.

flathub.json
jwillikers: Removed.

stretchly-sources.yaml
jwillikers: Removed.

net.hovancik.Stretchly.yaml
jwillikers: I can't get this to work using `mkdir -p`. I've tried both of the following and get the same error above.

```yaml
- mkdir -p main
- npm install --offline
```

```yaml
- mkdir -p "$FLATPAK_BUILDER_BUILDDIR/main"
- npm install --offline
```

jwillikers: I figured this out. Fixed.

net.hovancik.Stretchly.yaml
jwillikers: For some reason, when I attempt to update to node16, I can no longer build due to npm trying to install `electron-20.2.0.tgz`, which is not in the cached sources. Maybe a bug in the `flatpak-node-generator`?

barthalion:
Keep in mind the 21.08 runtime will be supported only for a year more.

bot, build net.hovancik.stretchly

hfiguiere:
> Keep in mind the 21.08 runtime will be supported only for a year more.

The node14 maintainer stance was to use node16.

jwillikers:
> > Keep in mind the 21.08 runtime will be supported only for a year more.
> 
> The node14 maintainer stance was to use node16.

I'm not having issues related to node16. I can build the application with node16 locally. The weird thing is that when I build locally, electron version 19.1.0 is used as expected. For some reason, when I attempt to update to the `node16` Flatpak extension, `electron-builder` tries to install electron 20.2.0 from the internet instead of using the cached version of electron, version 19.1.0. I've tried a bunch of different ways to get around the problem without success. I'm not sure why this is happening or how to fix it.

jwillikers:
bot, build net.hovancik.Stretchly

net.hovancik.Stretchly.yaml
jwillikers: I figured this out. I needed to pass `--recursive` to `flatpak-node-generator` to pick up the dependencies in `app/package-lock.json`. That's why the version was missing. This allowed me to update to node16 and the 22.08 platform.

jwillikers:
> > > Keep in mind the 21.08 runtime will be supported only for a year more.
> > 
> > 
> > The node14 maintainer stance was to use node16.
> 
> I'm not having issues related to node16. I can build the application with node16 locally. The weird thing is that when I build locally, electron version 19.1.0 is used as expected. For some reason, when I attempt to update to the `node16` Flatpak extension, `electron-builder` tries to install electron 20.2.0 from the internet instead of using the cached version of electron, version 19.1.0. I've tried a bunch of different ways to get around the problem without success. I'm not sure why this is happening or how to fix it.

I figured this out. I needed to pass `--recursive` to `flatpak-node-generator` to pick up the dependencies in `app/package-lock.json`. That's why the version was missing. This allowed me to update to node16 and the 22.08 platform.

jwillikers (approved):

PR #3535, opened 2022-09-28
Add Freerdp

akallabeth:

- [ ] Any additional patches or files have been submitted to the upstream projects concerned. *(If not, explain why.)* `.desktop` and `appdata.manifest` are only required for the flatpak build, so keep it here.


barthalion:
bot, build com.freerdp.FreeRDP

com.freerdp.FreeRDP.json
```diff
@@ -0,0 +1,177 @@
...
+        {
+            "type": "git",
+            "url": "https://github.com/FreeRDP/FreeRDP.git",
+            "tag": "2.8.0",
+            "commit": "e3fc97feb512053189e276b2ca79762990bb8c4c"
+        }
+        ]
+    },
+    {
+        "name": "freerdp-icons",
```
hfiguiere, line 140: there is no reason to have this in a separate module since it's the same source as above.

com.freerdp.FreeRDP.json
```diff
@@ -0,0 +1,177 @@
...
+        ]
+    },
+
+		{
+        "name": "wrapper",
+        "buildsystem": "simple",
+        "build-commands": [
+                "install -D freerdp.sh /app/bin/freerdp.sh",
+                "install -D com.freerdp.FreeRDP.appdata.xml /app/share/appdata/com.freerdp.FreeRDP.appdata.xml",
+                "install -D com.freerdp.FreeRDP.desktop /app/share/applications/com.freerdp.FreeRDP.desktop",
```
hfiguiere, line 165: it surely won't find these two files.

also this should be part of the main module too.

com.freerdp.FreeRDP.json
```diff
@@ -0,0 +1,177 @@
...
+			]
+		},
+    {
+        "name": "freerdp",
+        "buildsystem": "cmake-ninja",
+        "cleanup": [
+        ],
+        "config-opts": [
+					"-DCMAKE_VERBOSE_MAKEFILE=ON", 
+					"-DCMAKE_BUILD_TYPE:STRING=Release",
```
hfiguiere, line 112:
```suggestion
					"-DCMAKE_BUILD_TYPE:STRING=RelWithDebInfo",
```

com.freerdp.FreeRDP.json
```diff
@@ -0,0 +1,177 @@
...
+        /* X11 + XShm access */
+        "--share=ipc",
+        "--socket=x11",
+        /* Needs to talk to the network */
+        "--share=network",
+        /* Play sounds redirected from guests */
+        "--socket=pulseaudio",
+        "--socket=cups",
+        /* Wayland access */
+        "--socket=wayland",
```
hfiguiere, line 27: you have either x11 or wayland with fallback-x11.

com.freerdp.FreeRDP.json
```diff
@@ -0,0 +1,177 @@
+{
+    "app-id": "com.freerdp.FreeRDP",
+    "runtime": "org.freedesktop.Platform",
+    "runtime-version": "22.08",
+    "sdk": "org.freedesktop.Sdk",
+    "build-options" : {
+        "cflags": "-O3",
+        "cxxflags": "-O3"
```
hfiguiere, line 8: why do you need to override the -O flags?

hfiguiere (changes requested):
the json indentation is off in many places making it harder to read.

com.freerdp.FreeRDP.json
hfiguiere: that's the build failure btw.

hfiguiere:
> .desktop and appdata.manifest are only required for the flatpak build, so keep it here.

This is actually not true. A desktop application without a `.desktop` it is a bug. It has been around for 2 decades.
Neither is the appstream file.

BOTH should be submitted upstream and installed in the standard location by upstream build. (for a future release)

com.freerdp.FreeRDP.json
akallabeth: since we process a lot of bitmap data having better optimization flags set really improves performance. (specifically the simdi intrinsics that are automatically used with higher optimization flags)

com.freerdp.FreeRDP.json
akallabeth: @hfiguiere ok, can I use different build systems on the same module? (sorry, not my first flatpak but not done too many others)

akallabeth:
> > .desktop and appdata.manifest are only required for the flatpak build, so keep it here.
> 
> This is actually not true. A desktop application without a `.desktop` it is a bug. It has been around for 2 decades. Neither is the appstream file.
> 
> BOTH should be submitted upstream and installed in the standard location by upstream build. (for a future release)

we may consider this. (`xfreerdp` and `wlfreerdp` are mostly command line, but they open up a graphical window, so we treated it as command line app so far)

com.freerdp.FreeRDP.json
akallabeth: ok, found the `post-install` array

barthalion:
bot, build com.freerdp.FreeRDP

com.freerdp.FreeRDP.appdata.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
```
barthalion, line 2: This shouldn't be desktop app given it has to be launched from CLI.

akallabeth: ok, changed to `conosle-application`

hfiguiere:
> @hfiguiere ok, I´ve tested `flatpak run --command=xfreerdp com.freerdp.FreeRDP /buildconfig` which is failing now. worked before with `--socket=x11`

that's because you run Wayland and therefor should launch the wayland version like the wrapper script is trying to do.

hfiguiere:
If your flatpak require to know the command here, then maybe it's not suited for flatpak at all.

com.freerdp.FreeRDP.json
```diff
@@ -132,7 +132,6 @@
       "name": "wrapper",
       "buildsystem": "simple",
       "build-commands": [
-        "install -D freerdp.sh /app/bin/freerdp.sh",
```
hfiguiere: you need to change 'command' too.

akallabeth: good catch

barthalion:
bot, build com.freerdp.FreeRDP

PR #3574, opened 2022-10-10
Add org.nickvision.tubeconverter

nlogozzo:

### DIFF
--- a/org.nickvision.tubeconverter.json
+++ b/org.nickvision.tubeconverter.json
@@ -34,7 +34,7 @@
                 {
                     "type": "git",
                     "url": "https://github.com/nlogozzo/NickvisionTubeConverter.git",
-                    "tag": "2022.10.0-f1"
+                    "tag": "2022.10.1"
                 }
             ]
         }
--- a/org.nickvision.tubeconverter.json
+++ b/org.nickvision.tubeconverter.json
@@ -34,7 +34,7 @@
                 {
                     "type": "git",
                     "url": "https://github.com/nlogozzo/NickvisionTubeConverter.git",
-                    "tag": "2022.10.0"
+                    "tag": "2022.10.0-f1"
                 }
             ]
         }
--- a/org.nickvision.tubeconverter.json
+++ b/org.nickvision.tubeconverter.json
@@ -1,0 +1,42 @@
+{
+    "app-id" : "org.nickvision.tubeconverter",
+    "runtime" : "org.gnome.Platform",
+    "runtime-version" : "43",
+    "sdk" : "org.gnome.Sdk",
+    "command" : "org.nickvision.tubeconverter",
+    "finish-args":[
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--device=dri",
+        "--share=ipc",
+        "--share=network",
+        "--filesystem=host"
+    ],
+    "modules" : [
+        {
+            "name": "jsoncpp",
+            "buildsystem": "meson",
+            "sources": [
+                {
+                    "type": "git",
+                    "url": "https://github.com/open-source-parsers/jsoncpp.git",
+                    "tag": "1.9.5"
+                }
+            ]
+        },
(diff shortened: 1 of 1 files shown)


org.nickvision.tubeconverter.json
```diff
@@ -0,0 +1,42 @@
...
+    "runtime-version" : "43",
+    "sdk" : "org.gnome.Sdk",
+    "command" : "org.nickvision.tubeconverter",
+    "finish-args":[
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--device=dri",
+        "--share=ipc",
+        "--share=network",
+        "--filesystem=host"
```
hfiguiere, line 13: why do you need host permissions?

hfiguiere:
bot, build org.nickvision.tubeconverter

hfiguiere (changes requested):

org.nickvision.tubeconverter.json
nlogozzo: Tube Conveter can download a file to wherever a user chooses. Which won't necessarily be through a GtkFileChooserNative. 
If a user selects a folder to save files too, Tube Coverter will remember that location on disk and automatically download to that location when the application runs again without the user necessarily choosing that folder through a dialog.

nlogozzo:
Guess I need to fix the icon...

nlogozzo:
bot, build org.nickvision.tubeconverter

org.nickvision.tubeconverter.json
JakobDev: If you use GtkFileChooserNative, you don't need host access. The File chooser will give you access to the directory the User had selected.

nlogozzo: Yes but are those permissions reset when the application is closed? As i said before, the application remembers the last folder selected and may automatically write files in that directory without it being chosen in a dialog, thus requiring host.

barthalion: Yes, the permission to that directory should be stored in the permission store. You can verify that with `flatpak permission-list`.

barthalion: Please give it a try and if it misbehaves, I'll merge it as-is.

nlogozzo: It's not working for me. The folder path gets stored as /run/user/1000/... instead of the actual folder path and I get a bunch of permission denied errors when trying to write the file to it.

nlogozzo: Forgot to mention, the application uses yt-dlp inside the flatpak to download a file as well. And yt-dlp needs access to the filesystem to download the file as it doesn't use xdg portals.

nlogozzo:
bot, build org.nickvision.tubeconverter

org.nickvision.tubeconverter.json
hfiguiere: host is still very broad. You could limit to `xdg-download` and `xdg-videos`

hfiguiere (changes requested):

org.nickvision.tubeconverter.json
nlogozzo: The problem is in the application I don't force the user to a specific directory, they can pick wherever. And in turn, the application will remember that location for next time the application is run. And the yt-dlp exe needs access to whatever directory the user chooses at any time.

nlogozzo: I could change it to home...that's probably as best as it'll get.

nlogozzo: I had a similar problems with my other app Tagger, we had to publish with `filesystem=host` since for some reason C++ API and some bundled exes (i.e., ffmpeg, yt-dlp) don't work well with the "sandboxed" paths provided by the application using the xdg-portals without have full filesystem access.

nlogozzo:
Thanks! @barthalion

PR #3584, opened 2022-10-14
added arx Data Manipulation tool

muelli:
It's a simple self-contained Java app.
The app starts.
I haven't checked whether it does useful things, though.

https://arx.deidentifier.org/

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. [**Link:**](https://github.com/arx-deidentifier/arx/issues/416#issuecomment-1278631438) 

@prasser here is the submission for flathub. I would be glad if you can commit to co-maintaining the app to make sure the latest release makes it onto Flathub. It detects new releases automatically, so it should be a matter of reviewing the changes in the download URL.


JakobDev:
bot, build org.deidentifier.arx

org.deidentifier.arx.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+  - name: arx-binary
+    buildsystem: simple
+    build-commands:
+        - install -D -ma+rx arx.sh /app/bin/arx
+        - install -D -ma+r arx-gtk-64.jar /app/bin/arx-gtk-64.jar
+        - install -D -ma+r org.deidentifier.arx.desktop /app/share/applications/org.deidentifier.arx.desktop
+        - install -D -ma+r org.deidentifier.arx.appdata.xml /app/share/metainfo/org.deidentifier.arx.appdata.xml
+        - install -D -ma+r org.deidentifier.arx.svg /app/share/icons/hicolor/scalable/apps/org.deidentifier.arx.svg
+        - install -D -ma+r org.deidentifier.arx.png /app/share/icons/hicolor/128x128/apps/org.deidentifier.arx.png
```
JakobDev, line 35:
```suggestion
        - install -D -ma+r org.deidentifier.arx.svg /app/share/icons/hicolor/scalable/apps/org.deidentifier.arx.svg
```
You don't need the PNG if you have a SVG

org.deidentifier.arx.desktop
```diff
@@ -0,0 +1,13 @@
+[Desktop Entry]
+Type=Application
+Version=1.0
+Name=ARX
+GenericName=Arx Data Anonymisation Tool
+Comment=anonymizing sensitive personal data
```
JakobDev, line 6:
```suggestion
Comment=Anonymizing sensitive personal data
```

org.deidentifier.arx.yaml
```diff
@@ -0,0 +1,62 @@
...
+  - name: openjdk
+    buildsystem: simple
+    build-commands:
+      - /usr/lib/sdk/openjdk17/install.sh
+
+  - name: arx-binary
+    buildsystem: simple
+    build-commands:
+        - install -D -ma+rx arx.sh /app/bin/arx
+        - install -D -ma+r arx-gtk-64.jar /app/bin/arx-gtk-64.jar
```
JakobDev, line 31: The JAr is executed by the script and not directly, So you may want to put it in another folder than bin

muelli: yeah, good point. I don't care where that file lives. Do you have suggestions? lib?

org.deidentifier.arx.yaml
hfiguiere: the SVG look off-center.

org.deidentifier.arx.yaml
```diff
@@ -0,0 +1,62 @@
...
+      - type: file
+        path: org.deidentifier.arx.png
+
+      - type: file
+        path: org.deidentifier.arx.svg
+
+      - type: file
+        url: https://github.com/arx-deidentifier/arx/releases/download/v3.9.0/arx-3.9.0-gtk-64.jar
+        sha256: 2e5527de2dcdebadc31f72bc6cbec385aa1cea6aa9ed65fd73193b325ac99a5c
+        dest-filename: arx-gtk-64.jar
```
hfiguiere, line 56: `gtk-64` does that mean will only work on x86_64 despite being in Java?

org.deidentifier.arx.yaml
```diff
@@ -0,0 +1,62 @@
+app-id: org.deidentifier.arx
+runtime: org.gnome.Platform
```
hfiguiere, line 2: do you really need the gnome platform given this is Java

org.deidentifier.arx.yaml
muelli: that's a question for @prasser. I don't have any non x86_64 hardware to test.
Once we can build from source, it should also run on other platforms.

org.deidentifier.arx.yaml
muelli: yeah, the quality of the file is not good. That's why I prefer the rasterised version. Maybe @prasser can look at the SVG file and make it look nice.

org.deidentifier.arx.yaml
muelli: yes. It doesn't run otherwise. I've tested it and complained on their bug tracker: https://github.com/arx-deidentifier/arx/issues/416
It's a typical Java stacktrace: super long but no information as to what's going on.

org.deidentifier.arx.yaml
JakobDev: The JAR contains some .so files, but I don't know if they are actually required or just for some Optional feature

org.deidentifier.arx.yaml
JakobDev: Maybe /app/arx-gtk

org.deidentifier.arx.yaml
JakobDev: The filename contains Gtk. So I guess it uses things from Gtk at one point.

hfiguiere: Gtk3 is in the freedestkop runtime. Has always been.

hfiguiere: > yes. It doesn't run otherwise. I've tested it and complained on their bug tracker: [arx-deidentifier/arx#416](https://github.com/arx-deidentifier/arx/issues/416) It's a typical Java stacktrace: super long but no information as to what's going on.

yuk.

ok

org.deidentifier.arx.yaml
hfiguiere: because currently there is another 64-bits architecture built for this flatpak, but I doubt the "64" here mean "any 64-bit architecture".

prasser: Yes, x86_64 only.

hfiguiere: that's disappointing.

prasser: ARX can be built for PPCLE as well, but we currently don't do that. Would that be needed for ARX to be added?

hfiguiere: aarch64?

I advocate building from source in general, but because it's java, I just don't bother.

barthalion:
bot, build org.deidentifier.arx

PR #3603, opened 2022-10-22
wxBriscola app

numerunix:
This app is the trump game at two players, one with pc without multiplayer.
It needs cards that must be taken apart.


hfiguiere:
why is there the source code here?

Also submission have to be made against the `new-pr` branch.

PR #3616, opened 2022-10-31
Add MediathekView

lunaryorn:

- [ ] I am using only the minimal set of permissions.  **MediathekView needs to talk to Flatpak over DBus in order to start VLC**.
- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:**  https://github.com/mediathekview/MediathekView/issues/679
- [x] I own the domain used in the application ID or the domain has a policy for delegating subdomains (e.g. GitHub, SourceForge). **The domain is owned by MediathekView upstream**.
- [ ] Any additional patches or files have been submitted to the upstream projects concerned. *The patches are specific to the flatpak build.*

### DIFF
--- a/de.mediathekview.MediathekView.appdata.xml
+++ b/de.mediathekview.MediathekView.appdata.xml
@@ -1,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+    <id>de.mediathekview.MediathekView</id>
+    <launchable type="desktop-id">de.mediathekview.MediathekView.desktop</launchable>
+    <name>MediathekView</name>
+    <summary>Access to public German TV Mediathek</summary>
+    <metadata_license>CC0-1.0</metadata_license>
+    <project_license>GPL-3.0</project_license>
+    <project_group>MediathekView</project_group>
+    <developer_name>MediathekView developers</developer_name>
+    <url type="homepage">https://mediathekview.de/</url>
+    <url type="help">https://mediathekview.de/anleitung/</url>
+    <url type="vcs-browser">https://github.com/mediathekview/MediathekView</url>
+    <url type="donation">https://mediathekview.de/spenden/</url>
+    <description>
+        <p>
+            MediathekView searches the online media libraries of German public broadcasting stations
+            (ARD, ZDF, Arte, 3Sat, SWR, BR, MDR, NDR, WDR, HR, RBB, ORF, SRF) and lists available shows.
+            It supports filtering this list for specific broadcasts, and allows viewing and downloading videos.
+            It also enables subscriptions and automatically downloads new episodes.
+        </p>
+        <p>
+            Viewing videos requires the <a href="https://flathub.org/apps/details/org.videolan.VLC">VLC</a> flatpak.
+        </p>
+    </description>
+    <screenshots>
+        <screenshot type="default">
+            <image>https://mediathekview.de/images/screenshots/screenshot1.png</image>
+        </screenshot>
+    </screenshots>
+    <update_contact>sebastian@swsnr.de</update_contact>
+    <releases>
+        <release version="13.9.1" date="2022-07-14" />
+    </releases>
+    <content_rating type="oars-1.1" />
+</component>
--- a/de.mediathekview.MediathekView.desktop
+++ b/de.mediathekview.MediathekView.desktop
@@ -1,0 +1,11 @@
+[Desktop Entry]
+Categories=AudioVideo;Audio;Video;Player;TV;
+Comment=Access to the Mediathek of different tv stations
+Comment[de]=Durchsucht die Online-Mediatheken verschiedener Sender
+Exec=de.mediathekview.MediathekView
+GenericName=Mediathek
+Icon=de.mediathekview.MediathekView
+Name=MediathekView
+StartupNotify=true
+Terminal=false
+Type=Application
--- a/de.mediathekview.MediathekView.yaml
+++ b/de.mediathekview.MediathekView.yaml
@@ -1,0 +1,104 @@
+app-id: de.mediathekview.MediathekView
+runtime: org.freedesktop.Platform
+runtime-version: '22.08'
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.openjdk17
+add-extensions:
+  org.freedesktop.Platform.ffmpeg-full:
+    directory: "lib/ffmpeg"
+    version: "22.08"
+    add-ld-path: "."
+command: de.mediathekview.MediathekView
+finish-args:
+  - '--share=network' # We need the network to download films
+  - '--socket=x11' # Show windows (Java doesn't support wayland yet)
+  - '--share=ipc' # Required for X11
+  - '--device=dri' # Render things with opengl
+  - '--socket=pulseaudio' # Sound
+  - '--talk-name=org.freedesktop.Notifications'   # For desktop notifications
+  - '--filesystem=xdg-download' # Standard directory for downloads
+  - '--talk-name=org.freedesktop.Flatpak' # to launch the VLC flatpak
+  # Persist Mediathekview config directories; it doesn't use XDG_CONFIG_HOME unfortunately, see
+  # https://github.com/mediathekview/MediathekView/issues/339
+  - '--persist=.mediathek3'
+  - '--persist=.mediathek.javafx.tool.JFXHiddenApplication/'
+  # See https://github.com/flathub/org.freedesktop.Sdk.Extension.openjdk#usage
+  - '--env=PATH=/app/bin/:/app/jre/bin:/usr/bin'
+cleanup-commands:
+  # Some cargo-culting for ffmpeg; copied this from elsewhere
+  - 'mkdir -p /app/lib/ffmpeg'
+modules:
+  # See https://github.com/flathub/org.freedesktop.Sdk.Extension.openjdk#usage
+  - name: openjdk
+    buildsystem: simple
+    build-commands: ['/usr/lib/sdk/openjdk17/install.sh']
+  # Libnotify for desktop notifications; we build it separately to avoid depending
+  # on the entire Gnome runtime.
+  - name: libnotify
+    buildsystem: meson
+    config-opts:
+      - -Dtests=false
+      - -Dintrospection=disabled
+      - -Dman=false
+      - -Dgtk_doc=false
+      - -Ddocbook_docs=disabled
+    sources:
+      - type: archive
+        url: https://download.gnome.org/sources/libnotify/0.8/libnotify-0.8.1.tar.xz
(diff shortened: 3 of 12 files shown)


lunaryorn:
Should you accept this submission please do give additional access to @derreisende77 and @alex1702 which are MediathekView upstream maintainers.

lunaryorn:
@hfiguiere I'm not sure what this build failure means 🤔   The manifest [successfully builds on GH actions](https://github.com/lunaryorn/de.mediathekview.MediathekView/actions/runs/3362626850/jobs/5574598329).

It looks as if flatpak spawn is not allowed, but MediathekView really needs to launch VLC.  What alternatives do I have?

open-vlc.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+#
+# This wrapper makes a lot of assumptions on how openscad is called
+# from within FreeCAD
```
JakobDev, line 5: You should probably change this comment

open-vlc.sh
```diff
@@ -0,0 +1,8 @@
...
+#
+
+exec flatpak-spawn --directory="$(pwd)" --host flatpak run --filesystem="$XDG_CACHE_HOME/tmp" org.videolan.VLC "$@"
```
JakobDev, line 8: So it does not work with a native installed VLC

JakobDev (commented):
You should probably upstream your patches

JakobDev:
> It looks as if flatpak spawn is not allowed, but MediathekView really needs to launch VLC. What alternatives do I have?

`talk-name=org.freedesktop.Flatpak` causes this linter error. Don't worry, it is allowed, but you need a reason for it, which you have. You should open a PR and add MediathekView to [this file](https://github.com/flathub/flatpak-builder-lint/blob/master/flatpak_builder_lint/staticfiles/exceptions.json).

open-vlc.sh
hfiguiere: why would you do that?

open-vlc.sh
hfiguiere: as the one who wrote it, I agree.

lunaryorn: Sorry 😬 I've removed this comment.

open-vlc.sh
lunaryorn: I'm sorry, but I'm not sure what I should do here… should I call VLC in a different way?  If so, what is the proper way to call an external flatpak/tool outside the current flatpak?

I came across https://github.com/flatpak/flatpak/issues/4046#issuecomment-1177381023 and figured that `flatpak-spawn` was the (semi-?)official way to do this.

I'm happy to use a different way, but I don't know how to do this differently.  As far as I'm concerned it's fine if this only works with a flatpaked VLC; someone who doesn't use a flatpaked VLC would likely also not use a flatpaked MediathekView.  MediathakView upstream provides an AppImage download for those who don't use flatpak so I don't think this flatpak needs to care for semi-flatpaked installations 🤔

hfiguiere: I was replying to "does not work with a native installed VLC".

Calling the flaptak version is the right approach.

lunaryorn: Ah, sorry, then I misunderstood your remark 😊 So I can consider this conversation resolved?

lunaryorn:
@JakobDev I've opened https://github.com/flathub/flatpak-builder-lint/pull/78 to add an exception for mediathekview.

open-vlc.sh
Erick555: Why `--filesystem="$XDG_CACHE_HOME/tmp"` is needed here?

lunaryorn: I don't know. I copied this from the above link.

I'll try to remove it, and see if it still works 😇

hfiguiere: > I don't know. I copied this from the above link.

please don't copy stuff you don't understand.

lunaryorn: @Erick555 I've removed the redundant argument.

> please don't copy stuff you don't understand.

I appreciate the lecture but come on, if I only copied what I fully understand I'd give you an empty pull request.  I don't even understand YAML in all its weird weirdness 😉

I wanted to get this working before adding the polish, and that this use of spawn to start other flatpaks doesn't seem to be documented anywhere didn't exactly help with understanding 😉   I'm sorry that I forgot to cleanup this script before submitting, tho.

barthalion:
bot, build de.mediathekview.MediathekView

default-vlc-path.patch
```diff
@@ -0,0 +1,13 @@
+diff --git i/src/main/java/mediathek/tool/GuiFunktionenProgramme.java w/src/main/java/mediathek/tool/GuiFunktionenProgramme.java
+index 5f25443..6a1d737 100644
+--- i/src/main/java/mediathek/tool/GuiFunktionenProgramme.java
++++ w/src/main/java/mediathek/tool/GuiFunktionenProgramme.java
+@@ -77,7 +77,7 @@ public static String getPathToApplicationJar() {
+         return s;
+     }
+ 
+-    private static final String PFAD_LINUX_VLC = "/usr/bin/vlc";
++    private static final String PFAD_LINUX_VLC = "/app/bin/vlc";
```
barthalion, line 10: This somehow feels wrong, as you will need to instruct user they need to install VLC from Flathub. There's no way to express cross-apps dependencies with flatpak.

de.mediathekview.MediathekView.yaml
```diff
@@ -0,0 +1,104 @@
+app-id: de.mediathekview.MediathekView
```
barthalion, line 1: Given it's strongly coupled with VLC, I wonder if it wouldn't be better to bundle VLC with the app. This can be easily achieved by using it as BaseApp.

de.mediathekview.MediathekView.appdata.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
lunaryorn: I can try to upstream these files, but given that MediathekView is a cross-platform application and has most of its users on Windows I think it's unlikely to get upstream to care for metadata of a (from their point of view) somewhat niche packaging format 😬  Same goes for the desktop file.

So if these are not blockers, I'd prefer to address this later, to be honest.

default-vlc-path.patch
lunaryorn: I think I asked this before, but… what else should I do?

de.mediathekview.MediathekView.yaml
lunaryorn: @barthalion You can perfectly use MediathekView without VLC, if you just use it to record live streams (this requires ffmpeg which is used as an extension) or download shows (this doesn't require anything at all).

I can make VLC a base app of Mediathek, but wouldn't this install VLC twice for everyone who also installs VLC explicitly?

lunaryorn:
@barthalion Can I somehow disable the aarch64 build?  MediathekView uses openjfx which has platform-specific JARs.  Extracting a list of Java sources from Maven is already hard enough because maven literally has no interface for this, so we're down to parsing its download logs, and doing that for two platforms (one of which I don't even have access to) is maintenance work I don't really feel like doing 😬  I'd prefer to create an issue for the aarch64 build in the Mediathek flatpak repo and wait for a contributor interested in aarch64 😇

de.mediathekview.MediathekView.yaml
barthalion: If it's usable without VLC then it's fine as it is.

barthalion:
Yup: https://github.com/flathub/flathub/wiki/App-Maintenance#only-build-on-x86_64

de.mediathekview.MediathekView.yaml
lunaryorn: I think a base app would be surprising in so far as it'd duplicate VLC settings: If a user has customized their regular flatpak VLC these settings wouldn't propagate into the MediathekView sandbox, would they?

lunaryorn:
@barthalion Ah, I've seen this already but I meant to asked whether I'm permitted to do that on flathub 😇   I take your answer as "yes" 😊 

I'll disable the aarch build, and push again.

lunaryorn:
@barthalion I've pushed a `flathub.json` to disable aarch64.

de.mediathekview.MediathekView.appdata.xml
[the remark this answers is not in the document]
hfiguiere: this is not niche. This is actually an xdg standard and used by regular distros. Flatpak/flathub just requires it.

de.mediathekview.MediathekView.desktop
```diff
@@ -0,0 +1,11 @@
+[Desktop Entry]
```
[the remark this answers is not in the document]
hfiguiere: It's a bug to support a Linux desktop and not ship a .desktop.

de.mediathekview.MediathekView.appdata.xml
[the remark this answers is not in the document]
lunaryorn: Perhaps, but from a cross-platform (consumer) application point of view the entire Linux desktop and all its distros is niche.

MediathekView ships an AppImage already, and I presume that's as far as they'll go for now.  If upstream had interest in flatpak they'd have added it [four years ago](https://github.com/mediathekview/MediathekView/issues/382).

So again, if these files are not blockers, then I'd very much prefer to address this later 🙂

de.mediathekview.MediathekView.desktop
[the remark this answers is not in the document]
lunaryorn: I'm sorry, @hfiguiere, but this doesn't help me.  I have only so much time for this, so as I said I'd very much prefer to get the flatpak working and only aftwards start upstreaming changes, and as far as I understood @barthalion having these files in here is not a blocker.

Upstream may have a quite different point of view, or may simply not care, so it may take some time to get these things upstream, and I may not even succeed.

Hence I'd prefer if the review could focus on what's actually blocking this package.  And, on a personal note, I'd also appreciate a bit less condescending and passive-aggressive comments 🙂

de.mediathekview.MediathekView.yaml
barthalion: It might be true, depending on whether VLC respects `$XDG_CONFIG_HOME` env var. But yeah, feel free to ignore it.

barthalion:
bot, build de.mediathekview.MediathekView

de.mediathekview.MediathekView.yaml
lunaryorn: I think it does. I understand flathub has the possibility of beta builds into a separate repository?  If so couldn't I use this to test a separate MediathekView manifest with VLC as base app?  

And then, if it works well, use that as new regular manifest?

lunaryorn:
@barthalion Thank you for merging 🙏 It's a bit of a surprise; I wasn't aware that  this was so close to being merged 😅

@hfiguiere @JakobDev @Erick555 Thank you for your review and your feedback about the manifest 🙏

PR #3631, opened 2022-11-08
Add bino

marlam:

Explanations of permissions:
- `--device=all`: webcam capturing
- `--filesystem=host:ro`: read any video file, but write access is not necessary

### DIFF
--- a/org.bino3d.bino.yml
+++ b/org.bino3d.bino.yml
@@ -1,0 +1,70 @@
+app-id: org.bino3d.bino
+runtime: org.kde.Platform
+runtime-version: '6.3'
+sdk: org.kde.Sdk
+command: bino
+finish-args:
+  # X11 + XShm access
+  - --share=ipc
+  - --socket=x11
+  # Wayland access
+  - --socket=wayland
+  # Play videos from the net
+  - --share=network
+  # OpenGL rendering
+  - --device=dri 
+  # Webcam capturing
+  - --device=all
+  # Audio output
+  - --socket=pulseaudio
+  # Play video files, but only read access required
+  - --filesystem=host:ro
+cleanup:
+  - '/include'
+  - '/lib/*.a'
+  - '/lib/cmake'
+  - '/lib/debug'
+modules:
+  - name: openvr
+    buildsystem: cmake
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+    sources:
+      - type: archive
+        url: https://github.com/ValveSoftware/openvr/archive/refs/tags/v1.14.15.tar.gz
+        sha256: 40644e569f057e86c733d818a22d821c640307e5058d95958414422134d79584
+    cleanup:
+      - '/share/pkgconfig'
+  - name: vrpn
+    buildsystem: simple
+    build-commands:
+      - cmake -B build --install-prefix=/app -DCMAKE_BUILD_TYPE=Release
+      - cmake --build build
+      - cmake --install build
+    sources:
+      - type: archive
+        url: https://github.com/vrpn/vrpn/archive/refs/tags/version_07.35.tar.gz
+        sha256: faa98bd6f81be2ddaaf438fc3a89cfcc929a67627ddb72a3d4fe795fcf51de29
+    cleanup:
+      - '/bin'
+      - '/etc'
+      - '/lib/python*'
+      - '/share/vrpn-07.35'
+  - name: qvr
+    buildsystem: simple
+    build-commands:
+      - cmake -S libqvr -B build --install-prefix=/app -DCMAKE_BUILD_TYPE=Release
+      - cmake --build build
+      - cmake --install build
+    sources:
+      - type: archive
+        url: https://marlam.de/qvr/releases/qvr-4.0.1.tar.gz
+        sha256: eebca73cab2799bd48d800f606896e68128cc04faff72a374349c0b5901f1ac0
+  - name: bino
+    buildsystem: cmake
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+    sources:
+      - type: archive
+        url: https://bino3d.org/releases/bino-2.0.tar.gz
+        sha256: 7c62014c8c2fcdff8d78a66f65b5c52ff83aac6ea21521be20b19a0ae2395439
--- a/org.bino3d.bino.yml
+++ b/org.bino3d.bino.yml
@@ -63,3 +63,5 @@ modules:
       - type: archive
         url: https://bino3d.org/releases/bino-2.0.tar.gz
         sha256: 7c62014c8c2fcdff8d78a66f65b5c52ff83aac6ea21521be20b19a0ae2395439
+      - type: patch
+        path: update-metainfo.patch
--- a/org.bino3d.bino.yml
+++ b/org.bino3d.bino.yml
@@ -1,19 +1,17 @@
 app-id: org.bino3d.bino
 runtime: org.kde.Platform
-runtime-version: '6.3'
+runtime-version: '6.4'
 sdk: org.kde.Sdk
 command: bino
 finish-args:
   # X11 + XShm access
   - --share=ipc
-  - --socket=x11
+  - --socket=fallback-x11
   # Wayland access
   - --socket=wayland
   # Play videos from the net
   - --share=network
-  # OpenGL rendering
-  - --device=dri 
-  # Webcam capturing
+  # OpenGL rendering (dri) and Webcam capturing (all)
   - --device=all
   # Audio output
   - --socket=pulseaudio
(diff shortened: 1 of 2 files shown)


org.bino3d.bino.yml
```diff
@@ -0,0 +1,70 @@
...
+  # Play videos from the net
+  - --share=network
+  # OpenGL rendering
+  - --device=dri 
+  # Webcam capturing
+  - --device=all
+  # Audio output
+  - --socket=pulseaudio
+  # Play video files, but only read access required
+  - --filesystem=host:ro
```
hfiguiere, line 21: how about `xdg-videos` instead?

org.bino3d.bino.yml
```diff
@@ -0,0 +1,70 @@
+app-id: org.bino3d.bino
+runtime: org.kde.Platform
+runtime-version: '6.3'
```
hfiguiere, line 3: there is a 6.4 runtime.

org.bino3d.bino.yml
```diff
@@ -0,0 +1,70 @@
...
+  # Play video files, but only read access required
+  - --filesystem=host:ro
+cleanup:
+  - '/include'
+  - '/lib/*.a'
+  - '/lib/cmake'
+  - '/lib/debug'
+modules:
+  - name: openvr
+    buildsystem: cmake
```
hfiguiere, line 29:
```suggestion
    buildsystem: cmake-ninja
```

org.bino3d.bino.yml
```diff
@@ -0,0 +1,70 @@
...
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+    sources:
+      - type: archive
+        url: https://github.com/ValveSoftware/openvr/archive/refs/tags/v1.14.15.tar.gz
+        sha256: 40644e569f057e86c733d818a22d821c640307e5058d95958414422134d79584
+    cleanup:
+      - '/share/pkgconfig'
+  - name: vrpn
+    buildsystem: simple
```
hfiguiere, line 39: why no using the `cmake-ninja` buildsystem?

org.bino3d.bino.yml
```diff
@@ -0,0 +1,70 @@
...
+        sha256: faa98bd6f81be2ddaaf438fc3a89cfcc929a67627ddb72a3d4fe795fcf51de29
+    cleanup:
+      - '/bin'
+      - '/etc'
+      - '/lib/python*'
+      - '/share/vrpn-07.35'
+  - name: qvr
+    buildsystem: simple
+    build-commands:
+      - cmake -S libqvr -B build --install-prefix=/app -DCMAKE_BUILD_TYPE=Release
```
hfiguiere, line 56: why not using `cmake-ninja` buildsystem?

org.bino3d.bino.yml
```diff
@@ -0,0 +1,70 @@
...
+      - cmake --build build
+      - cmake --install build
+    sources:
+      - type: archive
+        url: https://marlam.de/qvr/releases/qvr-4.0.1.tar.gz
+        sha256: eebca73cab2799bd48d800f606896e68128cc04faff72a374349c0b5901f1ac0
+  - name: bino
+    buildsystem: cmake
```
hfiguiere, line 64:
```suggestion
    buildsystem: cmake-ninja
```

org.bino3d.bino.yml
```diff
@@ -0,0 +1,70 @@
...
+  - --device=all
+  # Audio output
+  - --socket=pulseaudio
+  # Play video files, but only read access required
+  - --filesystem=host:ro
+cleanup:
+  - '/include'
+  - '/lib/*.a'
+  - '/lib/cmake'
+  - '/lib/debug'
```
hfiguiere, line 26:
```suggestion
```

org.bino3d.bino.yml
```diff
@@ -0,0 +1,70 @@
...
+runtime: org.kde.Platform
+runtime-version: '6.3'
+sdk: org.kde.Sdk
+command: bino
+finish-args:
+  # X11 + XShm access
+  - --share=ipc
+  - --socket=x11
+  # Wayland access
+  - --socket=wayland
```
hfiguiere, line 11: does it really support wayland? if yes then it is the `fallback-x11` socket above instead of `x11`

org.bino3d.bino.yml
```diff
@@ -0,0 +1,70 @@
...
+  # Play videos from the net
+  - --share=network
+  # OpenGL rendering
+  - --device=dri
```
hfiguiere, line 15: this is redundant with `all`

org.bino3d.bino.yml
```diff
@@ -0,0 +1,70 @@
...
+  # X11 + XShm access
+  - --share=ipc
+  - --socket=x11
+  # Wayland access
+  - --socket=wayland
+  # Play videos from the net
+  - --share=network
+  # OpenGL rendering
+  - --device=dri 
+  # Webcam capturing
```
muelli, line 16: just for completeness: A camera portal exists: https://github.com/flatpak/xdg-desktop-portal/issues/414 it could be worth raising this with upstream so that a more fine-grained access can be provided rather than the wholesale access to all devices.

org.bino3d.bino.yml
marlam: That would probably be fine in GUI mode, but in Virtual Reality mode for large-scale systems with a render cluster, there is typically a dedicated cluster filesystem involved that users are encouraged to use for data that cluster nodes need to access. Those use cases are not typically covered by the xdg specs, but it's one of the focus areas of Bino.

org.bino3d.bino.yml
marlam: Updated, thanks.

org.bino3d.bino.yml
marlam: Changed, thanks.

org.bino3d.bino.yml
marlam: I could not get it to work. VRPN's use of CMake is ... hard to deal with. They require an extra build directory, and passing the -B option does not work. I found no better way to build VRPN.

marlam: But at least I switched the generator to ninja.

org.bino3d.bino.yml
marlam: With QVR the problem is that the CMakeLists.txt file is in the libqvr subdirectory, not the root directory, and simply passing the `-S libqvr` option in `config-opts` does not work for some reason; I had to specify the full command.
But I was able to change the generator to ninja.

org.bino3d.bino.yml
marlam: Changed, thanks.

org.bino3d.bino.yml
marlam: OK, I removed this, although I don't know why an empty /lib/debug should remain in the flatpak.

org.bino3d.bino.yml
marlam: Yes, it really supports Wayland. I changed to `fallback-x11`, thanks for the hint.

org.bino3d.bino.yml
marlam: I condensed this to one line, thanks.

org.bino3d.bino.yml
hfiguiere: `subdir` is here for that.

org.bino3d.bino.yml
hfiguiere: because it's actually a separate package for the debug symbols. (`.Debug` suffix)

org.bino3d.bino.yml
marlam: Speaking as upstream, I see two problems with this approach for Bino:

1. The portal mechanisms involve dialogs to get user feedback if I understood correctly, which would work in GUI mode, but not in Virtual Reality mode where you either wear a head mounted display or stand in a large scale VR system.
2. Bino uses Qt Multimedia for camera access, so there is no place in the code to integrate calls to portal APIs -- and to be completely honest, I rewrote Bino based on Qt precisely so that I don't have to deal with platform-specific code such as this...

org.bino3d.bino.yml
marlam: Ah - thanks, I did not know about that, I only looked in the manifest documentation and not the flatpak-builder documentation. Fixed now and works with `cmake-ninja`.

org.bino3d.bino.yml
marlam: Again, thanks for that. Fixed now and works with `cmake-ninja`.

org.bino3d.bino.yml
marlam: OK, thanks for the explanation.

hfiguiere:
bot, build org.bino3d.bino

hfiguiere:
```
builddir/files/share/appdata/org.bino3d.bino.appdata.xml: FAILED:
• style-invalid         : Not enough <screenshot> tags, minimum is 1
Validation of files failed
```

marlam:
OK, I updated the metainfo so that it passes local validation (and should pass bot validation) via a patch. This patch is already included in the main git repository of Bino, so this is not a deviation from upstream.
Is that acceptable?

hfiguiere:
bot, build org.bino3d.bino

PR #3633, opened 2022-11-09
io.github.Lunarequest.NightPDF

Lunarequest:

Adding the metainfo.xml & desktop file to upstream would add a set of files that are rarely updated and not reusable anywhere else in the project. Hence I would prefer if these where to remain with the flatpak and be updated with the flatpak rather then then downstream.

### DIFF
--- a/io.github.lunarequest.NightPDF.desktop
+++ b/io.github.lunarequest.NightPDF.desktop
@@ -1,6 +1,6 @@
 [Desktop Entry]
 Name=NightPDF
-Exec=/app/bin/nightpdf %U
+Exec=nightpdf %U
 Terminal=false
 Type=Application
 Icon=io.github.lunarequest.NightPDF
--- a/io.github.lunarequest.NightPDF.desktop
+++ b/io.github.lunarequest.NightPDF.desktop
@@ -1,0 +1,9 @@
+[Desktop Entry]
+Name=NightPDF
+Exec=/app/bin/nightpdf %U
+Terminal=false
+Type=Application
+Icon=io.github.lunarequest.NightPDF
+StartupWMClass=NightPDF
+Comment=Dark Mode for PDFs
+Categories=Utility;
--- a/io.github.lunarequest.NightPDF.metainfo.xml
+++ b/io.github.lunarequest.NightPDF.metainfo.xml
@@ -1,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>io.github.lunarequest.NightPDF</id>
+  
+  <name>NightPDF</name>
+  <summary>Dark Mode PDF reader</summary>
+  <url type="homepage">https://github.com/Lunarequest/NightPDF</url>
+  <content_rating type="oars-1.1" />
+
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>GPL-2.0-only</project_license>
+  
+  <supports>
+    <control>pointing</control>
+    <control>keyboard</control>
+    <control>touch</control>
+  </supports>
+  
+  <description>
+    <p>
+      NightPDF is a electron based application that allows one to read PDF files with dark mode. It has 3 default filters(monochrome, sepia &amp; redeye) along with a option to generate a custom filter along with no filter. NightPDF has a simplistic design that makes it intuitive to use.
+    </p>
+  </description>
+  
+  <launchable type="desktop-id">io.github.Lunarequest.NightPDF.desktop</launchable>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://github.com/Lunarequest/NightPDF/blob/mistress/docs/nightpdf.gif?raw=true</image>
+    </screenshot>
+  </screenshots>
+  <releases>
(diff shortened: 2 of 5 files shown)


Lunarequest:
bot, build io.github.lunarequest.NightPDF

io.github.lunarequest.NightPDF.yaml
```diff
@@ -0,0 +1,59 @@
+app-id: io.github.lunarequest.NightPDF
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
```
muelli, line 3: can 22.08 be used?

io.github.lunarequest.NightPDF.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
+Name=NightPDF
+Exec=/app/bin/nightpdf %U
```
hfiguiere, line 3:
```suggestion
Exec=nightpdf %U
```

io.github.lunarequest.NightPDF.yaml
```diff
@@ -0,0 +1,59 @@
...
+    sources:
+     - type: git
+       url: https://github.com/Lunarequest/NightPDF.git
+       tag: v0.4.3
+       commit: 9d5e77603284b5070966269ce04056aae5a9448d
+     - generated-sources.json
+     - type: file
+       path: io.github.lunarequest.NightPDF.metainfo.xml
+     - type: file
+       path: io.github.lunarequest.NightPDF.desktop
```
hfiguiere, line 53: these files should be hosted upstream.

io.github.lunarequest.NightPDF.yaml
hfiguiere: it should

hfiguiere (changes requested):

io.github.lunarequest.NightPDF.yaml
Lunarequest: I've already mentioned in the pr description I would prefer that these remain with the flatpak
> Adding the metainfo.xml & desktop file to upstream would add a set of files that are rarely updated and not reusable anywhere else in the project. Hence I would prefer if these where to remain with the flatpak and be updated with the flatpak rather then then downstream.

Lunarequest:
bot, build io.github.lunarequest.NightPDF

io.github.lunarequest.NightPDF.yaml
hfiguiere: a desktop app without a .desktop is a bug. you already seem to special case FLATPAK.

hfiguiere: and the AppStream isn't flatpak specific either.

Lunarequest:
why does the 128x128 icon fail? Its very specifically  put into place here
https://github.com/Lunarequest/flathub/blob/8e8a90d10cac777d055fc0dca2c76d89a0379de8/io.github.lunarequest.NightPDF.yaml#L41

io.github.lunarequest.NightPDF.yaml
hfiguiere: this is not addresser here. Version shoud match in both places.

io.github.lunarequest.NightPDF.yaml
Lunarequest: > a desktop app without a .desktop is a bug. you already seem to special case FLATPAK.

We do have a desktop file. Its auto generated by electron builder. It is not possible to make electron builder reuse .desktop files. Flatpak is a special case. It is not packaged by electron builder.

io.github.lunarequest.NightPDF.metainfo.xml
```diff
@@ -0,0 +1,38 @@
...
+    <control>touch</control>
+  </supports>
+  
+  <description>
+    <p>
+      NightPDF is a electron based application that allows one to read PDF files with dark mode. It has 3 default filters(monochrome, sepia &amp; redeye) along with a option to generate a custom filter along with no filter. NightPDF has a simplistic design that makes it intuitive to use.
+    </p>
+  </description>
+  
+  <launchable type="desktop-id">io.github.Lunarequest.NightPDF.desktop</launchable>
```
hfiguiere, line 25: this doesn't match the app-id. It **case sensitive**

Lunarequest:
bot, build io.github.lunarequest.NightPDF

io.github.lunarequest.NightPDF.yaml
hfiguiere, line 3:
```suggestion
runtime-version: '22.08'
```

io.github.lunarequest.NightPDF.yaml
hfiguiere: but you pull from git. so whether electron package it or not doesn't matter. Not that since it's generated in the build process, why not using the generated .desktop ?

Lunarequest: its generated for the following targets

- AppImage
- deb
- rpm

As per the electron packaging docs(flatpaks) the only target that should be used is `dir` as otherwise more files need to be downloaded and it gets messy very fast.

PR #3670, opened 2022-11-23
Moe.karaokes.mugen

AxelTerizaki:

I hope this will work out. Thanks for looking into this submission! :)


joshua-stone:
bot, build moe.karaokes.mugen.app

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "skip-arches": ["i386", "aarch64"]
```
hfiguiere, line 2:
```suggestion
  "skip-arches": ["aarch64"]
```

But still it's disappointing to ignore aarch64.

moe.karaokes.mugen.app.yml
```diff
@@ -0,0 +1,93 @@
+app-id: moe.karaokes.mugen.app
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
```
hfiguiere, line 3:
```suggestion
runtime-version: '22.08'
```

moe.karaokes.mugen.app.yml
```diff
@@ -0,0 +1,93 @@
...
+sdk: org.freedesktop.Sdk
+base: org.electronjs.Electron2.BaseApp
+base-version: '21.08'
```
hfiguiere, line 6:
```suggestion
base-version: '22.08'
```

moe.karaokes.mugen.app.yml
```diff
@@ -0,0 +1,93 @@
...
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node16
+command: km-wrapper
+separate-locales: false
+finish-args:
+  - '--filesystem=host'
+  - '--share=ipc'
+  - '--socket=x11'
+  - '--socket=pulseaudio'
+  - '--socket=wayland'
```
hfiguiere, line 16: does it support wayland?

moe.karaokes.mugen.app.yml
```diff
@@ -0,0 +1,93 @@
...
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+base: org.electronjs.Electron2.BaseApp
+base-version: '21.08'
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node16
+command: km-wrapper
+separate-locales: false
+finish-args:
+  - '--filesystem=host'
```
hfiguiere, line 12: that's a very broad file system permission.

moe.karaokes.mugen.app.yml
```diff
@@ -0,0 +1,93 @@
+app-id: moe.karaokes.mugen.app
```
hfiguiere, line 1: I have already expressed my concern about the app-id. (the fact that it ends in `.app` just reinforce it)

moe.karaokes.mugen.app.yml
```diff
@@ -0,0 +1,93 @@
...
+        commands:
+          - |
+            echo Stub sandbox ignoring command: $@
+            exit 1
+      - type: file
+        path: km-wrapper
+      - type: file
+        path: moe.karaokes.mugen.app.metainfo.xml
+      - type: file
+        path: moe.karaokes.mugen.app.png
```
hfiguiere, line 84: why including the icon here when there is one at https://gitlab.com/karaokemugen/karaokemugen-app/-/blob/next/build/icon512.png ?

moe.karaokes.mugen.app.yml
```diff
@@ -0,0 +1,93 @@
...
+      - type: file
+        path: moe.karaokes.mugen.app.desktop
```
hfiguiere, line 86: doesn't electron builder generate one?

hfiguiere (changes requested):

flathub.json
AxelTerizaki: We can try to look into it for future releases. :)

moe.karaokes.mugen.app.yml
AxelTerizaki: We debated with that internally. The app stores its data by default in `home` because it needs to make some files available for streamers (text files containing info they can display on their stream's layout via OBS for example). Since it can download and manipulate big files, we also have an option to let users store their video folder on something other than `home`, hence the `host` permission.

If there's a better way to do this permission-wise (allow home as well as any removeable media like an external HDD) I'd like to know to modify the yaml accordingly.

moe.karaokes.mugen.app.yml
AxelTerizaki: Regarding the 22.08 runtime version instead of 21.08, we tested with 22 but it threw a weird node exception we didn't know how to fix, We can look into this further in later releases, unless 22 is mandatory?

moe.karaokes.mugen.app.yml
AxelTerizaki: That's because there's an app and a server for KM. We don't plan on releaseing the server via flatpak but we wanted to differentiate it. If that's a big issue, we can remove the `.app`

We're discovering flathub and how to build for it, sorry for the mistakes.

moe.karaokes.mugen.app.yml
AxelTerizaki: Good question, must be a leftover from before. We will try to fix that.

moe.karaokes.mugen.app.yml
AxelTerizaki: Electron-builder does generate one when you specify a `deb` or other packed targets. When building into an unpacked folder it doesn't generate a `.desktop` file. 

I've seen other apps like codium build a `.deb` with electron-builder and extract its contents to get the `.desktop` file. Should we do that as well instead of providing it here? See [here](https://github.com/flathub/com.vscodium.codium/blob/c9cfa14dd1c71d5dce6e42bf47e975dbce84f2f0/com.vscodium.codium.yaml#L70)

Some other repos like [Visual Studio Code](https://github.com/flathub/com.visualstudio.code/blob/master/com.visualstudio.code.desktop) provide their own, so we thought it was okay.

AxelTerizaki: I removed the png from the repo but left the .desktop file, is that okay?

moe.karaokes.mugen.app.yml
AxelTerizaki: That actually causes no issues, so I'll have this commited, thank you. 

(I thought the base and runtime had to be aligned version wise)

moe.karaokes.mugen.app.yml
AxelTerizaki: Resolved with [548c838](https://github.com/flathub/flathub/pull/3670/commits/548c838eb3030a80e3456f766d379005fd7591e3) and [28f1506](https://github.com/flathub/flathub/pull/3670/commits/28f1506c1f258d62bc2ad87d8dd3d7274e7e8bc4)

moe.karaokes.mugen.app.yml
AxelTerizaki: This actually does build but doesn't run if base is 22.08 but runtime is still 21.08

moe.karaokes.mugen.app.yml
AxelTerizaki: Now it does, I just had it tested by a friend. Thanks for letting us know it could be an issue.

moe.karaokes.mugen.app.yml
hfiguiere: > That actually causes no issues, so I'll have this commited, thank you.
> 
> (I thought the base and runtime had to be aligned version wise)

suggestion was to fix both. of course they have to match.

AxelTerizaki: Yes, my bad. We're working on it right now to update to the 22.08 runtime AND base.

moe.karaokes.mugen.app.yml
AxelTerizaki: This should be working now, we upgraded to 22.08 and it builds/runs on our end.

moe.karaokes.mugen.app.yml
AxelTerizaki: We've found a way to limit the filesystem permissions to `~KaraokeMugen:create`, we need to release a new version though to fix that. I'll resolve this conversation once this is done.

AxelTerizaki: Thanks to @Dragicafit this should be better now. See [this commit](https://github.com/flathub/flathub/pull/3670/commits/288e9ffe66482eaafc415f09c592f3d1c43f657f)

barthalion:
bot, build moe.karaokes.mugen.app

hfiguiere:
bot, build moe.karaokes.mugen

moe.karaokes.mugen.yml
```diff
@@ -0,0 +1,87 @@
...
+runtime-version: '22.08'
+sdk: org.freedesktop.Sdk
+base: org.electronjs.Electron2.BaseApp
+base-version: '22.08'
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node18
+command: km-wrapper
+separate-locales: false
+finish-args:
+  - '--filesystem=~/KaraokeMugen:create'
```
barthalion, line 12: I'm not a fan of this, and it generally goes against XDG standard. Any reason it doesn't use $XDG_DATA_HOME and $XDG_CONFIG_HOME to store data?

AxelTerizaki: Historically speaking we store data created by the app in this folder on *all* systems, since the app is available on Windows, Linux and macOS. I saw in the documentation this wasn't advised, but there are a few reasons this is done like that : 

- The app generates text files for streamers using OBS, who can display information about their ongoing karaoke session by reading those files. They're stored in that folder (among other things).
- Other things are customizable by the user, like player backgrounds, jingles, and so on, so the user needs an easy way to access them.
- It's like an IDE in the sense that users can edit/create their own karaoke files and videos and as such need an easy access to their files when editing them with a third party tool like Aegisub or a video editor of their choice. (to be more precise, the app sotres its data files in that folder, and users can modify them outside of the app as needs be)

We believe `~/KaraokeMugen:create` is about right on how the app works and how to allow users to customize it or work with their data files inside that folder.

PR #3671, opened 2022-11-24
Add Feeel

12people:


dginovker:
bot, build com.enjoyingfoss.Feeel

com.enjoyingfoss.Feeel.yml
```diff
@@ -0,0 +1,47 @@
...
+sdk: org.freedesktop.Sdk
+command: feeel
+separate-locales: false
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --socket=pulseaudio
+  - --own-name=com.enjoyingfoss.feeel
```
hfiguiere, line 13: You should just use that d-bus name as the app-id as it is granted byb default. (yes it's case sensitive)

12people: @hfiguiere OK, should be fixed.

com.enjoyingfoss.Feeel.yml
```diff
@@ -0,0 +1,46 @@
+app-id: com.enjoyingfoss.feeel
```
hfiguiere, line 1: this file is a duplicate.

com.enjoyingfoss.feeel.yml
```diff
@@ -0,0 +1,46 @@
...
+    buildsystem: simple
+    build-commands:
+      - cp -R feeel/bin/ /app/feeel
+      - chmod +x /app/feeel/feeel
+      - mkdir /app/bin/
+      - mkdir /app/lib/
+      - ln -s /app/feeel/feeel /app/bin/feeel
+      - ln -s /usr/lib/x86_64-linux-gnu/libsqlite3.so.0 /app/lib/libsqlite3.so
+      - if [ ! -e '/app/lib/libsqlite3.so' ]; then ln -s -f /usr/lib/aarch64-linux-gnu/libsqlite3.so.0 /app/lib/libsqlite3.so; fi
+      - if [ $(arch)="aarch64"  ]; then sed -si "/Exec=/cExec=env LIBGL_ALWAYS_SOFTWARE=1 feeel" feeel/com.enjoyingfoss.feeel.desktop; fi
```
hfiguiere, line 24: 1. you should use desktop-file-edit. Also I'm not sure the sed command is right
2. It's better to use a wrapper script to the the environment appropriately. You could even skip the `ln -s` above with such a script.

hfiguiere:
bot, build com.enjoyingfoss.feeel

com.enjoyingfoss.Feeel.yml
12people: Whoops — thanks for letting me know!

It seems that macOS has ignorecase set to true by default, and even when disabling it, renaming the file makes git act really strange. Come to think of it, I think the strange behavior might have come from my cloud storage syncing.

Anyway, hopefully it's alright now.

com.enjoyingfoss.feeel.yml
12people: any guidance on how to write such a script, specifically how to refer to the executable? I could use the absolute path, but that doesn't seem like a clean solution; not sure how to get the relative path, though

(that said, this code is specifically for 64-bit ARM, so it shouldn't affect many people and I've tested that it works at the very least via OS launchers that refer to desktop files; I would like to write a wrapper script, but I hope such a thing isn't a blocker before release)

barthalion: You can do something along https://github.com/flathub/com.irccloud.desktop/blob/master/com.irccloud.desktop.json#L99 but this is indeed not a blocker.

PR #3672, opened 2022-11-24
org.blackfennec.app.extensions.core

SimonKindhauser:

### DIFF
--- a/org.blackfennec.app.extensions.core.yml
+++ b/org.blackfennec.app.extensions.core.yml
@@ -1,18 +1,18 @@
-id: org.blackfennec.extensions.core
+id: org.blackfennec.app.extensions.core
 branch: '0.10'
 runtime: org.blackfennec.app
 runtime-version: 'stable'
 sdk: org.gnome.Sdk//43
 build-extension: true
 modules:
-  - name: blackfennec.extensions.core
+  - name: blackfennec.app.extensions.core
     buildsystem: simple
     build-commands:
       # will be enabled with version 0.11
       #- pip install --prefix=${FLATPAK_DEST} .
-      - install -Dm644 ./org.blackfennec.extensions.core.metainfo.xml ${FLATPAK_DEST}/share/metainfo/org.blackfennec.extensions.core.metainfo.xml
+      - install -Dm644 ./org.blackfennec.app.extensions.core.metainfo.xml ${FLATPAK_DEST}/share/metainfo/org.blackfennec.app.extensions.core.metainfo.xml
     sources:
       - type: git
         url: https://gitlab.ost.ch/blackfennec/extensions/core.git
         tag: v0.10.0
-        commit: 910ac424a531347abc10cd7e6d9109a4c06fcbc0
+        commit: f90f755166c2f3bc8d9fa2312a0cce8d6c116977
--- a/org.blackfennec.app.extensions.core.yml
+++ b/org.blackfennec.app.extensions.core.yml
@@ -1,5 +1,4 @@
 id: org.blackfennec.app.extensions.core
-branch: '0.10'
 runtime: org.blackfennec.app
 runtime-version: 'stable'
 sdk: org.gnome.Sdk//43
--- a/org.blackfennec.extensions.core.yml
+++ b/org.blackfennec.extensions.core.yml
@@ -10,7 +10,7 @@ modules:
     build-commands:
       # will be enabled with version 0.11
       #- pip install --prefix=${FLATPAK_DEST} .
-      - install -Dm644 ./org.blackfennec.extensions.core.metainfo.xml ${FLATPAK_DEST}/share/metainfo/org.blackfennec.app.metainfo.xml
+      - install -Dm644 ./org.blackfennec.extensions.core.metainfo.xml ${FLATPAK_DEST}/share/metainfo/org.blackfennec.extensions.core.metainfo.xml
     sources:
       - type: git
         url: https://gitlab.ost.ch/blackfennec/extensions/core.git
--- a/org.blackfennec.extensions.core.yml
+++ b/org.blackfennec.extensions.core.yml
@@ -1,17 +1,18 @@
 id: org.blackfennec.extensions.core
 branch: '0.10'
 runtime: org.blackfennec.app
-runtime-version: '0.10'
+runtime-version: 'stable'
 sdk: org.gnome.Sdk//43
 build-extension: true
 modules:
   - name: blackfennec.extensions.core
     buildsystem: simple
     build-commands:
-      - pip install --prefix=${FLATPAK_DEST} .
+      # will be enabled with version 0.11
+      #- pip install --prefix=${FLATPAK_DEST} .
       - install -Dm644 ./org.blackfennec.extensions.core.metainfo.xml ${FLATPAK_DEST}/share/metainfo/org.blackfennec.app.metainfo.xml
     sources:
       - type: git
         url: https://gitlab.ost.ch/blackfennec/extensions/core.git
         tag: v0.10.0
-        commit: 47b62d3b9a0179d277fa52472f2789149d4bb5b8
+        commit: e05daa78c1388223485bed53ab6071cf1c338771
--- a/org.blackfennec.extensions.core.yml
+++ b/org.blackfennec.extensions.core.yml
@@ -15,4 +15,4 @@ modules:
       - type: git
         url: https://gitlab.ost.ch/blackfennec/extensions/core.git
         tag: v0.10.0
-        commit: 8887a180cf944d4cb239d9f665bb792e5542423f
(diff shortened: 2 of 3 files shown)


SimonKindhauser:
bot, build org.blackfennec.extensions.core

SimonKindhauser:
bot, build org.blackfennec.extensions.core

SimonKindhauser:
bot, build org.blackfennec.extensions.core

SimonKindhauser:
The pipeline fails with the following error:  
`** (process:2893135): ERROR **: 17:26:32.680: Cannot get the default GSettingsSchemaSource - is the gsettings-desktop-schemas package installed?`

I don't know what could trigger it...

dginovker:
You have to manually provide the pip dependencies like this

https://github.com/flathub/org.flathub.flatpak-external-data-checker/blob/master/org.flathub.flatpak-external-data-checker.yml#L271

https://github.com/flathub/org.flathub.flatpak-external-data-checker/blob/master/python3-jsonschema.json

You can read why and find a handy link here: https://github.com/flathub/flathub/wiki/App-Requirements#stable-releases-reproducible-builds. In particular:

> Accessing the network is not allowed during the build process, to download extra files, everything that an app requires must be included within the manifest. There is a [community provided set of tools](https://github.com/flatpak/flatpak-builder-tools) to make this easier for some projects (such as npm) that would traditionally want to access the network.

SimonKindhauser:
bot, build org.blackfennec.extensions.core

SimonKindhauser:
bot, build org.blackfennec.extensions.core

SimonKindhauser:
bot, build org.blackfennec.extensions.core

SimonKindhauser:
bot, build org.blackfennec.extensions.core

SimonKindhauser:
bot, build org.blackfennec.extensions.core

org.blackfennec.extensions.core.yml
```diff
@@ -0,0 +1,18 @@
...
+runtime-version: 'stable'
+sdk: org.gnome.Sdk//43
+build-extension: true
+modules:
+  - name: blackfennec.extensions.core
+    buildsystem: simple
+    build-commands:
+      # will be enabled with version 0.11
+      #- pip install --prefix=${FLATPAK_DEST} .
+      - install -Dm644 ./org.blackfennec.extensions.core.metainfo.xml ${FLATPAK_DEST}/share/metainfo/org.blackfennec.extensions.core.metainfo.xml
```
hfiguiere, line 13: so right now this package just install the appstream file in `/app/share/metainfo`.

It doesn't look right.

SimonKindhauser: we will uncomment line 12 in approx. 3 weeks for release 0.11. At the same time we will also remove line 29 and 30 of [org.balckfennec.app](https://github.com/flathub/org.blackfennec.app/blob/master/org.blackfennec.app.yml). The two lines are in conflict with each other.

org.blackfennec.extensions.core.yml
```diff
@@ -0,0 +1,18 @@
+id: org.blackfennec.extensions.core
```
barthalion, line 1:
```suggestion
id: org.blackfennec.app.extensions.core
```

Extensions targeting single app should use that app's ID.

barthalion (commented):
I don't think I really follow the idea here. Is the core extension expected to be released on a different cadence than the main app?

SimonKindhauser:
> 

Not sure what you're asking..
The main app has been first released about 6 weeks ago (two sprints) and has so far included the core and base extensions.
With the next release in ~two weeks we would like to (finally) ship the extensions separate from the app. Generally speaking, the extensions will see updates once every sprint as does the main app.

The reason we have started early with the preparations (e.g. requesting the extensions to be available on flathub) is to ensure a seamless upgrade to our users. We would hate to ship a broken product. It would be a shame to remove the extensions from the app without them being available on flathub for a few days!

org.blackfennec.extensions.core.yml
SimonKindhauser: we would have named the app `org.blackfennec` if we could have :grinning:  
i personally prefer the version without the `app`. 

Is it a SHOULD or a COULD?
Consider it done iff its the former :smile:

hfiguiere:
TBF I don't understand why you want to make these **flatpak** extensions. Even more since all is released at the same time and they seem to be required for the app to do anything.

SimonKindhauser:
Duly noted. We hope for more useful extensions in the future too.

barthalion:
> With the next release in ~two weeks we would like to (finally) ship the extensions separate from the app. Generally speaking, the extensions will see updates once every sprint as does the main app.

If the core extension is required to make the app work at all, and it's released at the same time when the app is updated, it doesn't make any sense to split it. The app has to work out of the box. The extensions should be used only for optional features.

org.blackfennec.extensions.core.yml
barthalion: Let me rephrase it: they need to match.

org.blackfennec.extensions.core.yml
```diff
@@ -0,0 +1,18 @@
...
+branch: '0.10'
+runtime: org.blackfennec.app
+runtime-version: 'stable'
+sdk: org.gnome.Sdk//43
+build-extension: true
+modules:
+  - name: blackfennec.extensions.core
```
SimonKindhauser, line 8:
```suggestion
  - name: blackfennec.app.extensions.core
```

SimonKindhauser:
The app works without it. However, our users expect it to be available.

In the future the release cadence of the main application and its extensions will differ.

Respectfully, the architecture of an application and thus its use of extensions should be up to the developers and maintainers. We do value your opinions but remain confident in our decisions.

SimonKindhauser:
bot, build org.blackfennec.app.extensions.core

SimonKindhauser:
bot, build org.blackfennec.app.extensions.core

SimonKindhauser:
bot, build org.blackfennec.app.extensions.core

SimonKindhauser:
bot, build org.blackfennec.app.extensions.core

hfiguiere:
> The app works without it. However, our users expect it to be available.

Then it should come by default.

> In the future the release cadence of the main application and its extensions will differ.

That's not a problem. Just update that part in the manifest and roll a new build. It's like updating a dependency. And if your fear is that "user will have to redownload everything" I can tell you that you just missed the part were flatpak deduplicate files (via ostree) and thus the update would only download the changed files.
 
> Respectfully, the architecture of an application and thus its use of extensions should be up to the developers and maintainers. We do value your opinions but remain confident in our decisions.

Respectfully, when it come to the context of Flatpak, we are pretty confident this is the wrong approach. We are not arguing about your architecture, just that Flatpak extensions are definitely the wrong solution here. You'll get a lot of complaints about missing functionality, and if you default to *auto download* the extension, then there is no point in making it separate.

org.blackfennec.extensions.core.yml
hfiguiere: > we would have named the app `org.blackfennec` if we could have 😀
> i personally prefer the version without the `app`.

Thing is usually I outline that `app` isn't proper (I would have suggested `blackfennec` or anything that isn't `app` or `desktop`) but it appear that I didn't this time.

hfiguiere: And just to clarify `org.blackfennec` would be rejected flat out.

SimonKindhauser:
> Then it should come by default.

We will auto download it. The platform is designed such that even the core extension can be replaced by a user provided implementation.

> That's not a problem. [...]

We agree. This was in response to an earlier question by [barthalion](https://github.com/barthalion).

> You'll get a lot of complaints about missing functionality

Some things we will have to learn the hard way. We can always revert back to the old ways for the core module if it turns out to be a mistake.

> Respectfully, when it come to the context of Flatpak, we are pretty confident this is the wrong approach.

Hopefully, together we will find a solution we can all stand behind.

> and if you default to auto download the extension, then there is no point in making it separate.

This feature seems tailor made for our use case. It provides a sane default but allows advanced user to customize their experience if they choose to.

org.blackfennec.app.extensions.core.yml
```diff
@@ -0,0 +1,18 @@
+id: org.blackfennec.app.extensions.core
+branch: '0.10'
```
barthalion, line 2: We discussed this briefly on Matrix, but I will repeat it here: you shouldn't use versioned branches for extensions targetting only apps. Once it's removed, I'm inclined to merge it, even though you will be switching back to the single manifest sooner than later.

SimonKindhauser:
```suggestion
```

SimonKindhauser:
bot, build org.blackfennec.app.extensions.core

PR #3682, opened 2022-11-27
io.github.TTimo.GtkRadiant

TTimo:


io.github.TTimo.GtkRadiant.json
```diff
@@ -0,0 +1,220 @@
+{
+  "app-id": "io.github.TTimo.GtkRadiant",
+  "runtime": "org.gnome.Platform",
+  "runtime-version": "3.28",
```
hfiguiere, line 4: this version of the runtime as been **EOL for a few years**.

io.github.TTimo.GtkRadiant.json
```diff
@@ -0,0 +1,220 @@
...
+  "sdk": "org.gnome.Sdk",
+  "command": "gtkradiant",
+  "finish-args": [
+    "--share=network",
+    "--share=ipc",
+    "--socket=x11",
+    "--socket=pulseaudio",
+    "--device=dri",
+    "--filesystem=host"
```
hfiguiere, line 13: why host filesystem permissions?

io.github.TTimo.GtkRadiant.json
```diff
@@ -0,0 +1,220 @@
...
+            },
+            {
+                "type": "file",
+                "path": "scons-Makefile",
+                "dest-filename": "Makefile"
+            }
+        ]
+    },
+    {
+      "name": "libGLU",
```
hfiguiere, line 49: you should use the shared-module for this

io.github.TTimo.GtkRadiant.json
```diff
@@ -0,0 +1,220 @@
...
+        }
+      ],
+      "post-install": [
+        "mv install /app/gtkradiant",
+        "ln -s /var/data/games /app/gtkradiant/games",
+        "echo '<?xml version=\"1.0\" encoding=\"iso-8859-1\" standalone=\"yes\"?><links></links>' > /app/gtkradiant/global.xlink"
+      ]
+    },
+    {
+      "name": "pkg-bits",
```
hfiguiere, line 190: this is  better if it is part of the "main" module.

io.github.TTimo.GtkRadiant.json
```diff
@@ -0,0 +1,220 @@
...
+        {
+          "type": "file",
+          "path": "io.github.TTimo.GtkRadiant.appdata.xml"
+        }
+      ],
+      "build-commands": [
+        "install -m755 gtkradiant.sh /app/bin/gtkradiant",
+        "install -Dm644 io.github.TTimo.GtkRadiant.desktop /app/share/applications/io.github.TTimo.GtkRadiant.desktop",
+        "install -Dm644 io.github.TTimo.GtkRadiant.appdata.xml /app/share/appdata/io.github.TTimo.GtkRadiant.appdata.xml",
+        "install -Dm644 io.github.TTimo.GtkRadiant.png /app/share/icons/io.github.TTimo.GtkRadiant.png"
```
hfiguiere, line 216: this is the wrong path for the icon.

io.github.TTimo.GtkRadiant.appdata.xml
```diff
@@ -0,0 +1,76 @@
...
+    <category>X-GameLevelEditor</category>
+  </categories>
+  <description>
+    <p>
+      GtkRadiant is the official level design toolchain for games powered by id Tech engines from id Software, and is maintained by a community of volunteers. GtkRadiant is powered by the GTK+ Project and released under a GPL license.
+    </p>
+  </description>
+  <project_license>GPL-2.0</project_license>
+  <metadata_license>CC0-1.0</metadata_license>
+  <icon type="remote" height="128" width="128">http://icculus.org/gtkradiant/branding/radiant_legacy_gtkr.png</icon>
```
barthalion, line 22: Please don't use remote icons.

hfiguiere (changes requested):

io.github.TTimo.GtkRadiant.desktop
```diff
@@ -0,0 +1,12 @@
+[Desktop Entry]
```
barthalion, line 1: This file is not specific to Flathub; you should maintain it in your repo instead.

io.github.TTimo.GtkRadiant.appdata.xml
```diff
@@ -0,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
barthalion, line 1: This file is not specific to Flathub; you should maintain it in your repo instead.

io.github.TTimo.GtkRadiant.appdata.xml
```diff
@@ -0,0 +1,76 @@
...
+          <li>Added a toolbar button to toggle creating detail brushes or structural brushes by default.</li>
+          <li>Added a preference option to create all brushes with a caulk texture.</li>
+          <li>Numerous fixes to the 4-view layout, in particular to the floating dialog window that works with it.</li>
+          <li>idTech3 surface inspector was redone and simplified.</li>
+          <li>Improved GL font rendering on Windows.</li>
+          <li>A lot of code cleanups.</li>
+        </ul>
+      </description>
+    </release>
+  </releases>
```
hfiguiere, line 75: `content_rating` is missing

io.github.TTimo.GtkRadiant.json
```diff
@@ -0,0 +1,220 @@
...
+        "cleanup": [ "*" ],
+        "no-autogen": true,
+        "sources": [
+            {
+                "type": "archive",
+                "url": "https://netix.dl.sourceforge.net/project/scons/scons/3.0.1/scons-3.0.1.tar.gz",
+                "sha256": "24475e38d39c19683bc88054524df018fe6949d70fbd4c69e298d39a0269f173"
+            },
+            {
+                "type": "file",
```
barthalion, line 42: Simply switch to `buildsystem: simple` and define commands in `build-commands`. It's been years since hacking around Makefile was needed.

io.github.TTimo.GtkRadiant.json
barthalion, line 4: Any reason you're using so old runtime?

io.github.TTimo.GtkRadiant.json
```diff
@@ -0,0 +1,220 @@
...
+        },
+        {
+            "type": "file",
+            "path": "gtkradiant-Makefile",
+            "dest-filename": "Makefile"
+        }
+      ],
+      "post-install": [
+        "mv install /app/gtkradiant",
+        "ln -s /var/data/games /app/gtkradiant/games",
```
hfiguiere, line 185: I doubt this will work. `/var` isn't part of the package.

io.github.TTimo.GtkRadiant.json
hfiguiere:
```suggestion
        "install -Dm644 io.github.TTimo.GtkRadiant.png /app/share/icons/hicolor/128x128/apps/io.github.TTimo.GtkRadiant.png"
```

io.github.TTimo.GtkRadiant.json
illwieckz: This kind of application requires to read files from installation folder and read and write files from and to user folder of games that are not installed with Flatpak and that existed before Flatpak was even an idea. The purpose is to load the data of a game, to create new content for the game, to write the new content in the place the game expects it, then to load it from the game.

io.github.TTimo.GtkRadiant.json
TTimo: The current source doesn't build out of the box against newer SDKs. A bit of investigation suggests it would take some non trivial refactoring and developer time to modernize.

GtkRadiant was written 20+ years ago against the Gtk 2 libraries. It uses an obsolete module (gtkglext) that depends on more obsolete pieces (pangox).

It has been increasingly difficult to build GtkR on modern distributions; we are looking to leverage flatpak's older SDKs and Platforms to help with this problem.

TTimo: The current source doesn't build out of the box against newer SDKs. A bit of investigation suggests it would take some non trivial refactoring and developer time to modernize.

GtkRadiant was written 20+ years ago against the Gtk 2 libraries. It uses an obsolete module (gtkglext) that depends on more obsolete pieces (pangox).

It has been increasingly difficult to build GtkR on modern distributions; we are looking to leverage flatpak's older SDKs and Platforms to help with this problem.

hfiguiere: OK, I'll rephrase it. Use a supported runtime if you want this to be accepted. You probably will need to bring some dependencies like Python2 (it's a one line addition to the manifest with shared-modules).

io.github.TTimo.GtkRadiant.json
hfiguiere: and? `host` is VERY BROAD. And might not even do what you think it does.

io.github.TTimo.GtkRadiant.json
hfiguiere: > A bit of investigation suggests it would take some non trivial refactoring and developer time to modernize.

You mean just a few patches?

TTimo: Interesting. Thanks for the feedback. We'll have to see internally if it's even worth it for us being on flathub then.

TTimo: Interesting. Thanks for the feedback. We'll have to see internally if it's even worth it for us being on flathub then.

io.github.TTimo.GtkRadiant.json
illwieckz: > and? `host` is VERY BROAD. And might not even do what you think it does.

I know enough of flatpak to even know `host` is not broad enough for this application.

Let's give an example. Let's pick the [Quake 3 Arena](https://en.wikipedia.org/wiki/Quake_III_Arena) game which was released in 1999, and for which the file layout and file path conventions for reading and writing were written in stone in 1999 and could not be modified without breaking compatibility.

With some luck Quake 3 would install in `/opt` and `host` would be broad enough. With bad luck it would install in `/usr` and then you would not be able to use this GtkRadiant flatpak to read content for this game because `host` is not broad enough.

Then, once you managed to read the game data with the GtkRadiant editor, you start creating content for the game (basically, game levels). The Quake 3 game expects the new content files produced by GtkRadiant to be written by GtkRadiant in `${HOME}/.q3a/baseq3` which is the user home folder of the game. There is no other way to do it, this is defined by the game and the game cannot read files from other places.

GtkRadiant is the official in-house id Software editor for Quake 3 and Quake Live, ported to GTK2 by them and made open source by them decades ago.

GtkRadiant is a level editor for many games developed in the last 20 years, starting with Quake 3 which was released in 1999 to games like [Unvanquished](https://unvanquished.net/) which is still in active development in 2022 (14 games are actually listed in GtkRadiant, including well known franchises like Wolfenstein, Star Trek, Jedi Knight and others, and this number becomes way larger when you start counting the countless mods based on them).

Just to give a glance on how much old the history of this software is, and how written in stone some design may be, GtkRadiant inherits some design from QuakeEd which was a NextStep application for ancestors of Quake 3 that was first released in 1996 and existed internally at id Software before that. . Changing some of those design requires to take a time machine and to jump 26 or 27 years ago in time, or if you can't afford a time machine, to rewrite all the binaries that were published and installed in millions of computers since more than 20 years, we talk about some of the most popular games in history.

So yes some of those design really don't feel like 2022 ones, but this is history and legacy.

So, this kind of editor should read and write files in places that are dictated by other applications, GtkRadiant is not the application deciding where to read and write. Reading and writing elsewhere would be a GtkRadiant bug making the application unusable.

Then, a big part of the design this application has to comply with is that the games it edits files for are implementing an user space file system and this application has to load the whole game file system when editing one file. So for example loading a file with a file picker that would open a portal to just the file being opened cannot work: you would need access to the whole game user file system to be able to edit the file, or  even create a brand new file without even opening it with a file picker first.

This application doesn't read and write files, it read and write an other application's user file system. This user file system isn't an image file, it's a whole tree of folders and files, and GtkRadiant requires arbitrary access to all of those folders and files to be able to click the “create” button (not even the “open” and “save” one).

Another thing that is worth mentioning is that not only GtkRadiant ships a graphical application, but GtkRadiant ships also some dedicated compilers, though those compilers don't compile C++ code, they compile 3D levels. Such compilers are also meant to be called from outside the flatpak when building game source directories. I know no one who would question the requirement for a compiler to have a broad access to a filesystem when called by a build system. Though, while one can work around that issue with some `flatpak run` option when calling the  compiler, the application needs to run the compiler with broad access.

On that compiler point of view, GtkRadiant is a bit like an IDE that would require it's compiler to access system's `/usr/include`, (that said, that just reminds me some [flatpak problems](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1504) users cannot even work around with `--filesystem=host` because of some flatpak design shortcoming).

Note: I'm not a maintainer of GtkRadiant itself though I'm a contributor to GtkRadiant. I'm myself a maintainer of a fork named [NetRadiant](https://netradiant.gitlab.io/) (exchanging code back and forth with GtkRadiant) and I also plan to make a flatpak for NetRadiant. So not only I know well the needs of GtkRadiant, but I also scrutinize with much attention and interest any answer you will do because like GtkRadiant, NetRadiant inherits the same legacy design no one has any power to change.

Outside of the topic of “old libraries and old SDK being used”, if GtkRadiant fails to be included into flathub, what would block it for inclusion would probably block NetRadiant as well. This would be the same for other applications like [DarkRadiant](https://www.darkradiant.net/), [UFO Radiant](https://ufoai.org/wiki/UFORadiant_installation), [NRC](https://github.com/Garux/netradiant-custom), [worldspawn](https://github.com/VeraVisions/worldspawn), and [others], plus some other editors not sharing GtkRadiant code like [J.A.C.K](https://jack.hlfx.ru/en/main.html) or [TrenchBroom](https://trenchbroom.github.io/). If the usage of the `--filesystem=host` option would block GtkRadiant for inclusion this would mean flathub/flatpak would not be compatible with more-than 25 years old design and method of working of the industry.

Wait… TrenchBoom is actually [included in flathub as a flatpak](https://flathub.org/apps/details/com.kristianduske.TrenchBroom) and… [it sets `--filesystem=host`](https://github.com/flathub/com.kristianduske.TrenchBroom/blob/4c17169f220f323ee5910f27ca010dac9af352d3/com.kristianduske.TrenchBroom.json#L14).

io.github.TTimo.GtkRadiant.json
illwieckz: > > A bit of investigation suggests it would take some non trivial refactoring and developer time to modernize.
> 
> You mean just a few patches?

@TTimo is a bit humble but he is the best people on earth to evaluate the amount of work required to implement this or that. He was actually part of the id Software team (the company that initially developed this software and the games this software is meant to produce content for) and was in charge of the Linux ports of those games and applications. I assume he is the one who actually ported Quake 3 to Linux, the one who ported this application from Windows MFC to GTK back in the days, ported this application to more recent games like Quake Live, etc. After that he worked for Valve to port games to Linux ([example of interview on the topic](https://www.gamingonlinux.com/2016/09/interview-with-timothee-besset-formerly-of-id-software-who-helped-port-rocket-league-to-linux/)). TTimo is likely doing the maintenance of this software since two decades. So not only he knows very well about porting software to this  or that technology, but he also knows very well how to evaluate the cost of such effort in both term of time, effort or money. If he says his investigation suggests it would take some non trivial refactoring and developer time to modernize, he fully means it and this is not underrated in any way.

Now, GtkRadiant is a fully volunteer project, including TTimo's maintenance effort. The summed available volunteer time of all GtkRadiant contributors is very small (maybe less than a couple of hours per year). This software is still the reference on the topic, though.

Let's elaborate on this sentence:

> GtkRadiant was written 20+ years ago against the Gtk 2 libraries. It uses an obsolete module (gtkglext) that depends on more obsolete pieces (pangox).

I don't know exactly what port to what new version of what libraries requires the newer SDKs, but let's just pick an obvious example: GTK.

What if someone comes and ask “why you would not port it to GTK3, it's just a few patch”? Here is the answer:

This software uses GTK 2 with GtkGLExt with OpenGL 1. GTK3 replaced GtkGLExt with GtkGLArea that don't support OpenGL 1. Rewriting the renderer from OpenGL 1 to OpenGL core (what supports GTK3's GtkGlArea) would require a similar amount of work that would be required to port a game from OpenGL to Vulkan (to make a comparison with things from today). This is huge. As a comparison, the [Dæmon Engine](https://github.com/DaemonEngine/Daemon) who has an OpenGL core reimplementation of the OpenGL 1 Quake 3 renderer haven't reached  100% compatibility after 17 years of volunteer work. Luckily the GtkRadiant renderer only implement a very small subset of those features, but really, doing such port would not be easy. Instead of a few patch it would be rewriting some core components from scratch.

Hopefully GTK3 GtkGLArea has a undocumented feature to run OpenGL 1, but this undocumented feature only exists on Linux and Windows, not on macOS. So to keep GtkRadiant buildable on both Linux, Windows and macOS, GtkRadiant would be able to be built with both GTK2 and GTK3 with the same source tree. As an example, [NetRadiant](https://netradiant.gitlab.io/) implemented a wrapper over both GTK2 and GTK3 to allow this, and after years of work of volunteer time, the GTK3 implementation is still not fully working and can still not share an OpenGL context across windows, meaning the application has to run every OpenGL view in a single window instead of allowing to use multiple window (eventually on multiple screens) and other bugs (power consumption is much higher because it still renders the frame multiple time per second instead of only redrawing it when something changes). That means for NetRadiant, the GTK3 build was never released and is only available for those building it themselves. After months or years of work starting from today to do something similar, GtkRadiant would be in the same half-baked situation.

I don't know what is required to use a more recent runtime, but if more recent runtime requires GTK3, a port may require several months of full time work of a seasoned developer.

Other changes in other components may also require huge rewrite in similar ways.

GtkRadiant hasn't seen a Linux build since 2013 while there are more recent versions for Windows and macOS, one of the reason seems to be that it is hard to rely on old libraries on Linux. You'll also notice this flatpak work started5 years ago, that also gives a glance on how few volunteer time is available to make this software move forward.

Since flatpak can host multiple applications running multiple runtimes from various ages, I can understand that flatpak appears as a good candidate for running applications using obsolete technologies. If flathub refuses to include applications using obsolete technologies just on principle of refusing applications using old technologies despite being technically compatible with that, flathub would be no better than other packages managers. Such limitation being a technical limitation (like traditional package repositories have) or a moral principle would have the exact same effect.

barthalion: Yeah, I think it's a valid point to make. If the software is old, and the only sensible way to make it work is to use an old runtime, I'd say it's acceptable.

barthalion:
The build will likely fail because of some issues pointed out above, but feel free to trigger new builds when you push new commits.

bot, build io.github.TTimo.GtkRadiant

barthalion:
Hm, weird failure. Have you managed to build it locally?

TTimo:
It did build and run locally for x86_64, but I had to fiddle with the flatpak commands to get the 3.28 Sdk and Platform installed.

io.github.TTimo.GtkRadiant.json
TTimo: I see a Gtk2 folder in the shared-modules repo. We're not going to upgrade to Gtk 3, but we may still be able to move to a newer Sdk reasonably easily if we leverage the shared-modules backports.

hfiguiere: it starts too, but then this is a workstation.

hfiguiere: oh and this is using freedesktop sdk 22.08

io.github.TTimo.GtkRadiant.appdata.xml
```diff
@@ -0,0 +1,76 @@
...
+      <caption>GtkRadiant. Level design tools</caption>
+      <image type="source" width="897" height="432">https://icculus.org/gtkradiant/branding/radiant_v3.svg</image>
+    </screenshot>
+  </screenshots>
+  <releases>
+    <release version="1.6.7" date="2022-11-27">
+      <description>
+        <ul>
+          <li>Flathub availability.</li>
+        <ul>
```
hfiguiere, line 34:
```suggestion
        </ul>
```

this doesn't even valid xml otherwise

numpy-Makefile
```diff
@@ -0,0 +1,5 @@
+all:
```
hfiguiere, line 1: this file is never referenced, ie you can remove it.

io.github.TTimo.GtkRadiant.json
```diff
@@ -0,0 +1,220 @@
...
+          "url": "https://github.com/TTimo/GtkRadiant.git",
+          "commit": "6be32bdeecbe1332936a4c43422844691743d75f"
+        },
+        {
+          "type": "patch",
+          "path": "patches/gtkradiant/revert-uintptr.patch"
+        },
+        {
+            "type": "file",
+            "path": "gtkradiant-Makefile",
```
hfiguiere, line 179: same as above, use buildsystem symple. also `-j8` isn't nice. there is `$FLATPAK_BUILDER_N_JOBS` for that (so that it build faster on builder, but doesn't hammer the poor laptop)

io.github.TTimo.GtkRadiant.appdata.xml
hfiguiere: this one is required. if it wasn't clear.

io.github.TTimo.GtkRadiant.json
```diff
@@ -0,0 +1,220 @@
...
+          "revision": "144",
+          "dest": "install/installs/UrTPack"
+        },
+        {
+          "type": "svn",
+          "url": "svn://svn.icculus.org/gtkradiant-gamepacks/WolfPack/trunk",
+          "revision": "144",
+          "dest": "install/installs/WolfPack"
+        },
+        {
```
hfiguiere, line 168: I'd move this source at the top of list since it's the main one. It doesn't change the build, but that's a good practice.

io.github.TTimo.GtkRadiant.json
hfiguiere: It's all there https://github.com/hfiguiere/flathub/tree/gtkradiant

barthalion:
bot, build io.github.TTimo.GtkRadiant

barthalion:
```
builddir/files/share/appdata/io.github.TTimo.GtkRadiant.appdata.xml: FAILED:
```

io.github.TTimo.GtkRadiant.json
TTimo: Thanks for the edits! We're on the new runtime now.

illwieckz:
@TTimo Is there something to keep from https://github.com/TTimo/flathub-GtkRadiant/pull/1 or everything was reimplemented and I can close that PR?

io.github.TTimo.GtkRadiant.json
```diff
@@ -0,0 +1,214 @@
...
+          "path": "patches/gtkglext-1.2.0/libGL.so.1.diff"
+        },
+        {
+          "type": "patch",
+          "path": "patches/gtkglext-1.2.0/03_gdkglext-config-h-installation-path.diff"
+        },
+        {
+          "type": "patch",
+          "path": "patches/gtkglext-1.2.0/glibc2.27-ftbfs.diff"
+        }
```
hfiguiere, line 87: The config.sub and config.guess in the tarball are too old. This is the fix (untested):

```suggestion
        },
        {
            "type": "shell",
            "commands": [
                "cp -p /usr/share/automake-*/config.{sub,guess} ."
            ]
        }
```

TTimo:
bot, build io.github.TTimo.GtkRadiant

barthalion:
LGTM. @TTimo @illwieckz, can you please confirm it works as expected?

illwieckz:
I got this error message, it means that “The file system is read only”:

20221205-225207-000 gtkradiant-flatpak-error-read-only-filesytem

hfiguiere:
yes it is. Not sure what is expected, but /app, like /usr is immutable.

TTimo:
Oh, right. This old code assumes it can just read/write it's settings and other misc. stuff right next to it's executable. We probably need to do a pass through the code and fix it up to something like $XDG_CONFIG_HOME

barthalion:
Yeah, $XDG_CONFIG_HOME and $XDG_DATA_HOME are writable by default for the apps.

TTimo:
bot, build io.github.TTimo.GtkRadiant

TTimo:
https://github.com/TTimo/GtkRadiant/issues/687 should be mostly fixed now. Looks like it fails to initialize OpenGL now.

barthalion:
Starts fine on my end, but I have no working game installations to fully test it.

TTimo:
bot, build io.github.TTimo.GtkRadiant

barthalion:
The latest build doesn't start for me.

```
Stack trace of thread 2:
                #0  0x0000557cd48cb147 n/a (/app/gtkradiant/radiant.bin + 0x4f147)
                #1  0x000000010001c0b0 n/a (n/a + 0x0)
                #2  0x0000000000000000 n/a (n/a + 0x0)
                #3  0x2f746e6169646172 n/a (n/a + 0x0)
```

TTimo:
Thanks I see it. Crashing in release configs only. A few quick checks but I couldn't narrow it down yet.

barthalion:
bot, build io.github.TTimo.GtkRadiant

TTimo:
bot, build io.github.TTimo.GtkRadiant

barthalion:
bot, build io.github.TTimo.GtkRadiant

barthalion:
I don't have any game to test it with, but the manifest et al. look good to me. @TTimo should we merge it?

TTimo:
yes please, this is ready to merge

PR #3714, opened 2022-12-14
Add DatMan

SjoerdB93:
DatMan is a simple but powerful data manipulation tool developed with Libadwaita using Gnome Builder. It opens and plots two-column data files which then easily can be manipulated and transformed. Alternatively, data can be inserted and plotted using an equation. The transformed data can also be saved as .txt file for further use.

### DIFF
--- a/python3-matplotlib.json
+++ b/python3-matplotlib.json
@@ -1,0 +1,90 @@
+{
+    "name": "python3-matplotlib_modules",
+    "buildsystem": "simple",
+    "build-commands": [],
+    "modules": [
+{
+    "name": "python3-matplotlib",
+    "buildsystem": "simple",
+    "build-commands": [
+        "pip3 install --verbose --exists-action=i --no-index --find-links=\"file://${PWD}\" --prefix=${FLATPAK_DEST} \"matplotlib\" --no-build-isolation"
+    ],
+    "sources": [
+        {
+            "type": "file",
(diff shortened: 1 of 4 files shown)


SjoerdB93:
Just to comment on the permissions, it does have access to the home directory. This is because I want it to be able to save edited files to arbitrary places. I save data using np.savetxt(), which doesn't work without access to home. 

I could limit access to be in the document folder only if that's preferred, but that may be confusing for the end user. Opening works great without that permission as the Filechooser uses the xdg-desktop-portal if I recall correctly. But I don't think Python otherwise has access to saving files.

SjoerdB93:
bot, build se.sjoerd.DatMan

barthalion:
I think you would need to wrap np.savetxt() somehow so it properly uses a portal. Unfortunately, it's not like Flatpak achieved world domination and can rig `open()` to make that happen automatically.

/merge

SjoerdB93:
> I think you would need to wrap np.savetxt() somehow so it properly uses a portal. Unfortunately, it's not like Flatpak achieved world domination and can rig `open()` to make that happen automatically.
> 
> /merge

I've asked for help online, and if you treat the filepath as immutable and don't cut parts of them, then it actually works without home permissions. I removed the extension in the end and appended ` _edited.txt ` to the file path string. This breaks whatever kind of abstraction this portal uses for some reason. Now I no longer add this _edited part in the file and now it works without permissions. 

They'll be removed in the next build :)

PR #3724, opened 2022-12-18
add rocks.koreader.KOReader

proletarius101:

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://github.com/koreader/koreader/issues/6349
- [x] ~I~ The Upstream owns the domain used in the application ID or the domain has a policy for delegating subdomains (e.g. GitHub, SourceForge).


org.koreader.KOReader.yml
```diff
@@ -0,0 +1,58 @@
+app-id: org.koreader.KOReader
```
hfiguiere, line 1: `koreader.org` doesn't exist. Also no mention of that domain in other reference. Is this app-id valid?

org.koreader.KOReader.yml
```diff
@@ -0,0 +1,58 @@
...
+  - --socket=fallback-x11
+  - --talk-name=org.freedesktop.Notifications
+  - --filesystem=host
+modules:
+  - name: koreader
+    buildsystem: simple
+    sources:
+      - type: file
+        only-arches: [x86_64]
+        url: https://github.com/koreader/koreader/releases/download/v2022.11/koreader-appimage-x86_64-linux-gnu-v2022.11.AppImage
```
hfiguiere, line 22: why no building from source?

flathub.json
```diff
@@ -0,0 +1,4 @@
+{
+  "automerge-flathubbot-prs": true,
+  "only-arches": ["x86_64"]
```
hfiguiere, line 3: this is disappointing given that the license is FLOSS.

org.koreader.KOReader.yml
pazos: it is `koreader.rocks`

hfiguiere: then why did you use `koreader.org` ?

flathub.json
pazos: Hi, upstream contributor. Using prebuilt debian packages would give you armhf and aarch64 too.

Same building them from source. All targets are intended to be built on a x86_64 host with standard gcc cross toolchains.

hfiguiere: building from source is prefered. Also flathub build native. not cross.

org.koreader.KOReader.yml
pazos: I don't know :p

I came from https://github.com/koreader/koreader/issues/6349#issuecomment-1356721925 and have no relation with the person who's PR'ing this :)

hfiguiere: oh sorry I thought you were the submitter.

Apologies.

Anyway. app-id as submitted doesn't meet the requirements.

flathub.json
hfiguiere: just to clarify aarch64 only is supported beide x86_64.

pazos: Native builds won't work without minimal patching. If OP is interested in that route we would be able to make it work.

org.koreader.KOReader.yml
hfiguiere: so why is there even an appdata file here then?

org.koreader.KOReader.yml
```diff
@@ -0,0 +1,58 @@
...
+      - install -Dm644 org.koreader.KOReader.metainfo.xml ${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.metainfo.xml
+
+      - chmod +x koreader.AppImage
+      - ./koreader.AppImage --appimage-extract
+
+      - desktop-file-edit --set-icon ${FLATPAK_ID} --set-key Exec --set-value 'koreader %u' squashfs-root/koreader.desktop
+      - install -Dm644 squashfs-root/koreader.desktop ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop
+      - install -Dm644 squashfs-root/resources/koreader.svg ${FLATPAK_DEST}/share/icons/hicolor/scalable/apps/${FLATPAK_ID}.svg
+
+      - sed -i "s,!./luajit,!${FLATPAK_DEST}/koreader/luajit,g" squashfs-root/reader.lua
```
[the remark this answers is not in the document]
Frenzie: We do, but the link isn't quite relevant. ^_^

https://github.com/koreader/koreader-base/blob/0d8f7076c9904d2894c03b0a1db7c3f5a92e3267/thirdparty/luajit/koreader-luajit-enable-table_pack.patch

org.koreader.KOReader.yml
TingPing: The project looks very buildable, its documented here: https://github.com/koreader/koreader/blob/master/doc/Building.md

TingPing: In general it works by luck, the CI happens to use an ABI similar to the Flatpak runtime. All it takes is Flatpak or the CI system to have a mismatch and everything breaks, sometimes in subtle ways.

Frenzie: The AppImage is purposefully built on Ubuntu 18.04 (perhaps not quite CentOS 7, but it's not, say, Ubuntu 22.04) so presumably much of the luck is manufactured, even if it's still luck. ;-)

flathub.json
pazos: > I managed to utilize the debs: [2a7d3021869b821b85bbad04bdf8b0caba1a17a7](https://github.com/flathub/flathub/pull/3724/commits/2a7d3021869b821b85bbad04bdf8b0caba1a17a7). Arm64 will be supported

Keep in mind the debian version isn't self contained as we try to repurpose host deps when it is possible: https://github.com/koreader/koreader/blob/master/platform/debian/do_debian_package.sh#L51

So fonts will be missing, unless you add them yourself to your build.

To be honest it would be easier for us to include a new target that does the same as the debian build without calling https://github.com/koreader/koreader/blob/master/platform/debian/do_debian_package.sh and then you should be able to build from source using a couple of commands.

org.koreader.KOReader.yml
pazos: > While I understand your frustration, it's good enough for me as it's built by CI (not flathub.org, of course). It's actually more maintainable from my perspective, considering the effort of syncing the complicated build script from upstream. Said so, I may revisit it to make it build from source on flathub.org and I always welcome such contribution.

Please give it a try with https://github.com/koreader/koreader-base/pull/1504 and https://github.com/koreader/koreader/pull/9351 merged.

It should be as easy as:

1. clone repo and jump there
2. fetch 3rdparty: `./kodev fetch-thirdparty`
3. build native `./kodev release linux`
4. then you'll end with a tar.gz with everything *except* sdl2
5. use that as a base for flathub recipe

TingPing: @pazos Just to note, you can't "fetch" things during a build, however if they are just submodules they can cloned, or listed in the sources.

org.koreader.KOReader.yml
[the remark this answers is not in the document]
barthalion: Thanks for explanation.

org.koreader.KOReader.yml
```diff
@@ -13,49 +13,32 @@ finish-args:
   - --talk-name=org.freedesktop.Notifications
   - --filesystem=host
 modules:
-  - name: koreader
+  - shared-modules/SDL2/SDL2-2.26.1-with-libdecor.json
+  - shared-modules/lua5.1/lua-5.1.5.json
```
barthalion, line 17: Hasn't it used luajit before?

org.koreader.KOReader.yml
barthalion: To clarify further, it is a requirement for the domain to exist, as it is tied to upcoming verification and payments/donations features.

org.koreader.KOReader.yml
barthalion: Does `./kodev release linux` do it? That's what is unclear to me.

org.koreader.KOReader.yml
TingPing: If you *only* depend on libc ABI, and thats a big if unless you explicitly test this, then you will be OK.

pazos: >  you can't "fetch" things during a build, however if they are just submodules they can cloned, or listed in the sources.

Well, that's unfortunate.

We rely on CMake scripts to pull different packages from specific git branches/tags, from our own forked git repos or from plain `tar.gz` releases from upstream.

The same set of scripts is in charge of checking, downloading, patching and building each one of the dependencies.

> can you provide a target without git submodule update or other network access? I can patch them out but it's better to include it in the upstream

See 3 from https://github.com/koreader/koreader-base/issues/1486#issuecomment-1160305536

We're not interested in doing more work upstream. That work belongs to package maintainers.

> to my understanding we'll need to git clone each and every packages under koreader-base/thirdparty/ to build the base during build time?

A few more things, including translations, ` lfs` and a few other submodules spreaded across the codebase.

> If that's the case, I don't think it makes sense to replicate that in this PR, unless we can make those packages as submodules under koreader-base.

I'm beginning to agree with you. Is not possible to build from source without *somebody* **not upstream** willing to put the effort to adapt KO build script to the specific requirements of the build environment.

The best thing upstream can do without too much effort is publish a generic linux `tar.gz`  release and flathub can use it as the input for the recipe until somebody steps in and makes possible to build from source on this environment (see https://github.com/NixOS/nixpkgs/pull/178557 for an attempt to build KO from source on a networkless host).

org.koreader.KOReader.yml
TingPing: So `lua` can be removed?

rocks.koreader.KOReader.yml
```diff
@@ -0,0 +1,94 @@
...
+          url: https://github.com/koreader/koreader/raw/master/platform/appimage/koreader.appdata.xml
+          pattern: https://github.com/koreader/koreader/raw/master/platform/appimage/koreader.appdata.xml
+
+      - type: file
+        url: https://github.com/koreader/koreader/raw/master/resources/koreader.svg
+        sha256: b285123a86736a5c317de1e9362c6a84fcebd7075c1f947586541d81f99ef29d
+        dest-filename: koreader.svg
+        x-checker-data:
+          type: rotating-url
+          url: https://github.com/koreader/koreader/raw/master/resources/koreader.svg
```
barthalion, line 84: In the future, these should come from the deb file.

rocks.koreader.KOReader.yml
```diff
@@ -0,0 +1,94 @@
...
+          pattern: https://github.com/koreader/koreader/raw/master/resources/koreader.svg
+
+      - type: archive
```
barthalion, line 87: Possibly same about this?

barthalion:
LGTM. @pazos @Frenzie Any chance the SVG and appdata files can be included in the deb files? They are not specific to Flathub, and should be generally considered part of the release. Also, whom from KOReader team should be invited to the Flathub repo?

Frenzie:
Sorry, which SVG files are you referring to?

pazos (changes requested):
Using deb packages to build the flatpak image is cool, but you will need to pick fonts from https://github.com/koreader/koreader-fonts.

In particular you need to pull both the `droid` and `noto` dirs and push  them to the relevant places, since they're not there in debian.

Droid is a single truetype font. On noto you need to pick all ttf fonts and avoid `NotoSansCJKsc-Regular.otf`

See https://github.com/koreader/koreader/blob/master/platform/debian/do_debian_package.sh#L97-L105

PR #3737, opened 2022-12-21
Add RSS Guard Lite

guihkx:

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://github.com/martinrotter/rssguard

---

Hello. This is the exact same app as of PR #3736, except with no built-in web browser support.

If this gets accepted, please add @martinrotter as a collaborator as well.

Thank you!

### DIFF
--- a/io.github.martinrotter.rssguardlite.yml
+++ b/io.github.martinrotter.rssguardlite.yml
@@ -1,0 +1,60 @@
+id: io.github.martinrotter.rssguardlite
+runtime: org.kde.Platform
+runtime-version: '6.3'
+sdk: org.kde.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node16
+command: rssguard
+finish-args:
+  - --device=dri
+  - --share=ipc
+  - --share=network
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --socket=wayland
+  - --filesystem=xdg-download
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=org.freedesktop.Notifications
+modules:
+  # TODO: Remove the following module once we update to the 6.4 SDK.
+  # See also: https://invent.kde.org/packaging/flatpak-kde-runtime/-/issues/26
+  - name: qt6-core5compat
+    buildsystem: cmake-ninja
+    builddir: true
+    sources:
+      - type: archive
+        url: https://download.qt.io/official_releases/qt/6.3/6.3.2/submodules/qt5compat-everywhere-src-6.3.2.tar.xz
+        sha256: 91a3ac0f3f87e4103afa5834ccbecff9374daf716fab362d0c5e2d6ab98560cc
+    post-install:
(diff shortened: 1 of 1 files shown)


guihkx:
bot, build io.github.martinrotter.rssguardlite

barthalion:
Can you explain why would someone use Lite variant instead of the main one? The difference size is marginal (saving 100MB when runtimes take 2-3GB doesn't seem to matter), and it sounds like users who don't want to use the built-in browser can just not do it?

martinrotter:
> Can you explain why would someone use Lite variant instead of the main one? The difference size is marginal (saving 100MB when runtimes take 2-3GB doesn't seem to matter), and it sounds like users who don't want to use the built-in browser can just not do it?

I do not honestly understand why would you even ask this question? You could instead do 1 minute research and you would find "lite" (also known as "nowebengine" in RSS Guard community) variant of our program is actually used bit MORE than full-blown variant. Both variants of RSS Guard are used frequently by many many users for various reasons (philosophical and technical) and it really makes no sense dissecting those reasons here.

guihkx:
Let's calm ourselves. 👼 

>Can you explain why would someone use Lite variant instead of the main one?

Personal opinion (might not reflect @martinrotter's): QtWebEngine isn't updated as frequently as the web browser you and I use daily, which is not particularly good for security. At the same time, I believe that there are users who enjoy reading their news within RSS Guard itself, without having to rely on an external browser for that (I'm not one of them 😄).

>The difference size is marginal (saving 100MB when runtimes take 2-3GB doesn't seem to matter)

I disagree on this point, every MB we can save is definitely worth it. Additionally, memory footprint on the Lite version is lower as well... And I imagine that if Flatpak didn't have file deduplication to help save disk space, it wouldn't be as popular, so disk space usage seems like an important topic.

>and it sounds like users who don't want to use the built-in browser can just not do it?

That's true to *some* extent: There is a `--no-web-engine` option on the "regular" build that forces the usage of a simpler "webview reader", but even then, Qt will still launch `QtWebEngineProcess` and, unfortunately, it seems [we can't prevent it](https://github.com/martinrotter/rssguard/issues/713#issuecomment-1114662018).

martinrotter:
Moreover, there are people who hate Google and their code and do not want it on their PCs. QtWebEngine module is de-facto Chromium core.

barthalion:
@martinrotter It would be more than sufficient to just reply that these are considered separate applications. No need to be hostile because a reviewer is reviewing your submission.

martinrotter:
> @martinrotter It would be more than sufficient to just reply that these are considered separate applications. No need to be hostile because a reviewer is reviewing your submission.

I do not feel being hostile, but OK.

Yes, they are considered two distinctive flavors of one application.

guihkx:
(Another example is how RSS Guard has two different packages on Arch Linux: [rssguard](https://archlinux.org/packages/community/x86_64/rssguard/) and [rssguard-nowebengine](https://archlinux.org/packages/community/x86_64/rssguard-nowebengine/))

PR #3753, opened 2022-12-26
add media.emby.EmbyTheater

alucryd:

### DIFF
--- a/media.emby.EmbyTheater.json
+++ b/media.emby.EmbyTheater.json
@@ -1,0 +1,35 @@
+{
+    "app-id": "media.emby.EmbyTheater",
+    "runtime": "org.freedesktop.Platform",
+    "runtime-version": "21.08",
+    "sdk": "org.freedesktop.Sdk",
+    "command": "media.emby.EmbyTheater",
+  "finish-args": [
+    "--device=dri",
+    "--share=ipc",
+    "--share=network",
+    "--socket=fallback-x11",
+    "--socket=pulseaudio"
+  ],
+    "modules": [
+        {
+            "name": "media.emby.EmbyTheater",
+            "buildsystem": "simple",
+            "build-commands": ["tar -xf emby-theater-flatpak_3.0.19_${FLATPAK_ARCH}.tar.xz -C ${FLATPAK_DEST}"],
+            "sources": [
+                {
+                    "type": "file",
+                    "url": "https://github.com/MediaBrowser/emby-theater-electron/releases/download/3.0.19/emby-theater-flatpak_3.0.19_aarch64.tar.xz",
+                    "sha256": "031390a91b67b63e9745d4639fd50ff879b9293a597faa984fb4670a969e5c29",
+                    "only-arches": ["aarch64"]
+                },
+                {
+                    "type": "file",
+                    "url": "https://github.com/MediaBrowser/emby-theater-electron/releases/download/3.0.19/emby-theater-flatpak_3.0.19_x86_64.tar.xz",
+                    "sha256": "f21464402e6bf3c17203dd03b2ae60d915f00bcaa41df889df4b12cee15e5b3d",
+                    "only-arches": ["x86_64"]
+                }
+            ]
+        }
+    ]
+}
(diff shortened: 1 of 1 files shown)


media.emby.EmbyTheater.json
```diff
@@ -0,0 +1,35 @@
+{
+    "app-id": "media.emby.EmbyTheater",
+    "runtime": "org.freedesktop.Platform",
+    "runtime-version": "21.08",
```
hfiguiere, line 4: 22.08?

media.emby.EmbyTheater.json
```diff
@@ -0,0 +1,35 @@
...
+    "--share=ipc",
+    "--share=network",
+    "--socket=fallback-x11",
+    "--socket=pulseaudio"
+  ],
+    "modules": [
+        {
+            "name": "media.emby.EmbyTheater",
+            "buildsystem": "simple",
+            "build-commands": ["tar -xf emby-theater-flatpak_3.0.19_${FLATPAK_ARCH}.tar.xz -C ${FLATPAK_DEST}"],
```
hfiguiere, line 18: it would  be better to use `archive` type for the source and copy things as needed.

hfiguiere:
bot, build media.emby.EmbyTheater

media.emby.EmbyTheater.json
alucryd: Will bump, I always assumed the latest runtime was unstable for some reason, but it seems I was wrong.

media.emby.EmbyTheater.json
alucryd: Actually tried when I first pushed the server, but I got a `file format not recognized` error and didn't think much about it. Turns out my buildbot recipe had a typo, lowercase `j` instead of uppercase in the tar command, resulting in a bz2 archive instead of xz :/ Will fix the theater archive, and will do the same with the server package with the next update.

alucryd: @hfiguiere When I use the archive type, flatpak-builder doesn't extract it correctly for some reason (haven't fixed the archive type yet, but I forced `tar-bzip2`). Files and directories appear to be flattened, and don't respect the archive directory tree, if that makes sense.
Screenshot from 2022-12-28 11-40-00
Is that a known issue?

barthalion: It might be that `tar-*` stuff is buggy as I've never seen it used in the wild.

alucryd: Finally replaced the archives with proper `tar.xz` files, however I'm still seeing the same issue, even without `archive-type` :/ Are you able to reproduce?

```
{
    "app-id": "media.emby.EmbyTheater",
    "runtime": "org.freedesktop.Platform",
    "runtime-version": "22.08",
    "sdk": "org.freedesktop.Sdk",
    "command": "media.emby.EmbyTheater",
  "finish-args": [
    "--device=dri",
    "--share=ipc",
    "--share=network",
    "--socket=fallback-x11",
    "--socket=pulseaudio"
  ],
    "modules": [
        {
            "name": "media.emby.EmbyTheater",
            "buildsystem": "simple",
            "build-commands": ["cp -dr --no-preserve=ownership bin electron etc lib licenses share ${FLATPAK_DEST}"],
            "sources": [
                {
                    "type": "archive",
                    "url": "https://github.com/MediaBrowser/emby-theater-electron/releases/download/3.0.19/emby-theater-flatpak_3.0.19_aarch64.tar.xz",
                    "sha256": "133e402e54915faeb2b171b9e63bf6d12687538f8215eff43abe082b9aeee1b2",
                    "only-arches": ["aarch64"]
                },
                {
                    "type": "archive",
                    "url": "https://github.com/MediaBrowser/emby-theater-electron/releases/download/3.0.19/emby-theater-flatpak_3.0.19_x86_64.tar.xz",
                    "sha256": "2c5969be51052f91d89f5128da856d127d9338b1cd0e0449a61d1b82a862c6cb",
                    "only-arches": ["x86_64"]
                }
            ]
        }
    ]
}
```

barthalion: flatpak-builder passes `--strip-components=1` by default, and these tarballs don't have a "top level" directory.

barthalion:
bot, build media.emby.EmbyTheater

media.emby.EmbyTheater.json
alucryd: Ah, that explains it, thanks for the insight and fix. Just tried the resulting build and it looks good to me. Thanks!

PR #3777, opened 2023-01-03
org.freedesktop.Sdk.Extension.flutter

carterbox:
Waiting on:

https://github.com/flutter/flutter/issues/91318
https://github.com/flutter/flutter/issues/22429

- [ ] My pull request follows the instructions at [App Submission][submission].

The Flutter SDK is built with the assumption that it is installed in a user-writable directory as a git repository. It uses git to auto-manage itself and its dependencies (It installs its own dart compiler and builds itself). It will raise and error and refuse to do anything if the git repository information is deleted or if its directory is not writable. Writes some config files to the user's home directory, but these are non-critical (I think). All of this makes the Flutter SDK hard to package as an SDK extension because, by default, Flatpak SDK extensions are installed in a non-writeable location.

There is something going on that I don't understand because when I set `writable-sdk: true`, and try to use flutter located in `/usr/lib/sdk/flutter` it gets stuck on a progress bar while trying to update. BUT if I copy the flutter SDK into the build directory, then run the same commands, they complete in 3 seconds.

Are there some secret manifest parameters that will allow the Flutter SDK to easily/quickly write to it's own directory? Should I install it in a different place? Or is there a different mechanism for shipping a reusable build tool for flatpak. Otherwise, it might be better for me to contribute a guide to building with flutter in the docs (like the Python,QT,Electron guides).

Related to https://github.com/flathub/flathub/pull/3758.


barthalion:
I think moving cache elsewhere makes sense, probably to the working directory. Flatpak generally doesn't try to deal with caching external dependencies as part of the build as they're typically done offline.

carterbox:
FLATPAK_BUILDER_BUILDDIR probably

carterbox:
I spent a lot of time creating a list of all of the things that the Flutter SDK downloads and adding them manually to this recipe. However, in the end that didn't matter because I wasn't able to completely patch away all of the instances where the SDK tries to edit itself.

I finally stopped trying to add patches when I learned that flutter_tools and the other Dart packages that make up part of the SDK try to rebuild themselves whenever the Dart cache is updated? Preventing this behavior may or may not have required patching Dart too.

PR #3778, opened 2023-01-03
supersedes org.racket_lang.DrRacket

capfredf:
Based on feedback from the Racket community, the entry point is preferred to be `racket`, which is consistent with other Racket distributions. However, using the new entry point with the old application id means that `flatpak run org.racket_lang.DrRacket` opens a repl, which is very likely to throw users off.

Therefore, the new application is created. The old application will later be declared EOL.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+.flatpak-builder
+build/
+rebuild/
--- a/org.racket_lang.Racket.appdata.xml
+++ b/org.racket_lang.Racket.appdata.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<application>
+  <id type="desktop">org.racket_lang.Racket</id>
+  <name>Racket</name>
+  <summary>The Language-Oriented Programming Language</summary>
+  <metadata_license>CC0-1.0</metadata_license>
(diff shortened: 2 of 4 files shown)


capfredf:
Based on the feedback from the Racket community, the entry point is preferred to
be `racket`, which is consistent with other Racket distributions. However, using
the new entry point with the old application id means that `flatpak run
org.racket_lang.DrRacket` opens a repl, which is very likely to throw users off.

Therefore, the new application is created. The old application will later be
declared EOL.

capfredf:
bot, build org.racket_lang.Racket

barthalion:
I don't think I get the point, and even more so I'm disappointed this needs renaming a day after it has been merged. Renames are not something to be done on a whim, and it sounds like the community could be asked for opinion before this has been published.

Make sure to EOL the old ref as described in the docs, and please open a separate issue about archiving the old repo.

/merge

capfredf:
@barthalion my bad. I didn't know the id was not changeable. I thought I could change everything later after the app was published. I am sorry for causing any trouble

barthalion:
This is not a trouble per se, it's just inconvenient to users when ID changes, and also means I need to archive the repo.

PR #3780, opened 2023-01-04
com.helix_editor.Helix

blt-r:

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://github.com/helix-editor/helix/issues/2076
(Its actually [helix-editor.com](https://helix-editor.com), but dashes are supposed to be replaced with underscores)


flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64", "aarch64"]
```
barthalion, line 2: No need to define that, these are the only supported architectures already.

barthalion:
bot, build com.helix_editor.Helix

helix.appdata.xml
```diff
@@ -0,0 +1,57 @@
...
+  <screenshots>
+    <screenshot type="default">
+      <caption>Helix with default theme</caption>
+      <image>https://github.com/helix-editor/helix/raw/d4565b4404cabc522bd60822abd374755581d751/screenshot.png</image>
+    </screenshot>
+  </screenshots>
+
+  <url type="homepage">https://helix-editor.com/</url>
+
+  <content_rating>none</content_rating>
```
barthalion, line 33: This is wrong. It should be `  <content_rating type="oars-1.1" />`.

flathub.json
blt-r: Ok, removed it

helix.appdata.xml
blt-r: I see. Weird that `org.freedesktop.appstream-glib validate` didn't complain. Anyway, its fixed

com.helix_editor.Helix.yml
```diff
@@ -0,0 +1,86 @@
...
+        sources:
+          - type: archive
+            url: https://www.x.org/releases/individual/lib/libXmu-1.1.3.tar.bz2
+            sha256: 9c343225e7c3dc0904f2122b562278da5fed639b1b5e880d25111561bac5b731
+        cleanup:
+          - /include
+          - /lib/pkgconfig
+          - '*.la'
+
+finish-args:
```
hfiguiere, line 77: finish args are usually above modules.

com.helix_editor.Helix.yml
```diff
@@ -0,0 +1,86 @@
...
+  - --env=HELIX_RUNTIME=/app/lib/helix/runtime
+  # Filesystem acess
+  - --filesystem=host
```
hfiguiere, line 80: this is a very broad filesystem access.

com.helix_editor.Helix.yml
```diff
@@ -0,0 +1,86 @@
...
+      - install -D hx /app/bin/hx
+
+      - mkdir -p /app/lib/helix
+      - rm -rf /app/lib/helix/runtime
+      - cp -r runtime /app/lib/helix/runtime
+
+      - install -D Helix.desktop /app/share/applications/com.helix_editor.Helix.desktop
+      - install -D helix.png /app/share/icons/hicolor/256x256/apps/com.helix_editor.Helix.png
+
+      - install -D helix.appdata.xml /app/share/metainfo/com.helix_editor.Helix.metainfo.xml
```
hfiguiere, line 22: these files should be submitted upstream. A desktop app without icon or .desktop is just a bug.

com.helix_editor.Helix.yml
```diff
@@ -0,0 +1,86 @@
+app-id: com.helix_editor.Helix
+
+runtime: org.freedesktop.Sdk
+runtime-version: '22.08'
+sdk: org.freedesktop.Sdk
+
+command: hx
+
+modules:
+  - name: hx
```
hfiguiere, line 10: this being the main module. it should be last.

also any chance to build from source?

blt-r: Might as well build from source, no problem with that. I don't know what are the best practices here, if building from source is better then I'll do that, but I think it's easier to just use prebuilt things

com.helix_editor.Helix.yml
blt-r: Icon and .desktop are from upstream (https://github.com/helix-editor/helix/tree/master/contrib), but they don't put them in release archive. The appdata manifest however I wrote myself

com.helix_editor.Helix.yml
blt-r: Fixed it

blt-r:
I actually just realized we also need `--talk-name=org.freedesktop.Flatpak` so we can `flatpak-spawn` language servers installed on the host system. Is it ok if I add that?

com.helix_editor.Helix.yml
hfiguiere: building from source is always preferred.

barthalion:
It's allowed for IDEs/text editors, but you need to submit a PR to https://github.com/flathub/flatpak-builder-lint/blob/master/flatpak_builder_lint/staticfiles/exceptions.json.

com.helix_editor.Helix.yml
blt-r: Ok so I tried to make it build from source, but I don't think it's worth it. We need to build tree-sitter grammars. And for that we need to fetch sources for them. So we will have to fetch pkgs.tspm.io for sources and generate manifest module source definitions from them, and then compile them. It just better to use prebuilt grammars. But if we use prebuilt grammars, we might as well use the prebuilt binary

barthalion:
bot, build com.helix_editor.Helix

PR #3781, opened 2023-01-04
Zoho Notebook (com.zoho.Notebook)

DenysMb:

- [X] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Contacted via e-mail:**
- [ ] I own the domain used in the application ID or the domain has a policy for delegating subdomains (e.g. GitHub, SourceForge). **No, owned by Zoho**.

### DIFF
--- a/com.zoho.Notebook.metainfo.xml
+++ b/com.zoho.Notebook.metainfo.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2018-2022 Flathub maintainers -->
+<component type="desktop">
+  <id>com.zoho.Notebook</id>
+  <launchable type="desktop-id">com.zoho.Notebook.desktop</launchable>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>GPL-3.0</project_license>
+  <name>Notebook</name>
+  <summary>The most beautiful note-taking app across devices</summary>
+  <description>
+    <p>
+      Different types of notes should be treated differently. Each Note Card type is designed to complement the type of note you're taking.
+    </p>
+    <p>
+      Safely and securely share your notes with friends and co-workers. Share notes and collaborate securely with the people you trust—and only them.
+    </p>
+    <p>
+      This is the Notebook you'll never lose because it syncs to the cloud and across your devices. It's always backed-up and always up to date.
+    </p>
+  </description>
+  <categories>
+    <category>Utility</category>
+  </categories>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://www.zohowebstatic.com/sites/zweb/images/notebook/notebook-ban-laptop.webp</image>
+    </screenshot>
+  </screenshots>
+  <url type="homepage">https://www.zoho.com/notebook/</url>
+  <url type="help">https://help.zoho.com/portal/en/kb/notebook</url>
+  <developer_name>Zoho Corporation Pvt. Ltd.</developer_name>
+  <update_contact>support@zohonotebook.com</update_contact>
+  <releases>
+    <release version="2.0.4" date="2022-01-03"/>
+  </releases>
+  <content_rating type="oars-1.1"/>
+</component>
--- a/com.zoho.Notebook.metainfo.xml
+++ b/com.zoho.Notebook.metainfo.xml
@@ -23,7 +23,7 @@
   </categories>
   <screenshots>
     <screenshot type="default">
-      <image>https://www.zohowebstatic.com/sites/zweb/images/notebook/notebook-ban-laptop.webp</image>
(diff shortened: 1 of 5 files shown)


flathub.json
```diff
@@ -0,0 +1,4 @@
+{
+    "only-arches": ["x86_64"],
```

com.zoho.Notebook.yaml
```diff
@@ -0,0 +1,38 @@
...
+runtime: org.freedesktop.Platform
+runtime-version: '22.08'
+sdk: org.freedesktop.Sdk
+base: org.electronjs.Electron2.BaseApp
+base-version: '22.08'
+command: notebook
+separate-locales: false
+finish-args:
+  - --share=ipc
+  - --socket=wayland
```
hfiguiere, line 11: does it support wayland? if yes then it should be `fallback-x11` on the next line. Otherwise this line should be removed.

hfiguiere:
bot, build com.zoho.Notebook

hfiguiere:
The appstream file is never installed.

flathub.json
DenysMb: Actually, I think it works on others arches but I was not able to try it.
We can just remove it from here and check.

com.zoho.Notebook.yaml
DenysMb: It support. I will change.

barthalion:
bot, build com.zoho.Notebook

com.zoho.Notebook.metainfo.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
barthalion, line 1: This file is not referenced and installed in the manifest.

DenysMb: Ops! I didn't realize that. Referenced now.

com.zoho.Notebook.yaml
DenysMb: Ok, so before was working on Wayland, but after change to this way it stopped to work, so I guess it not works on Wayland. I changed to only X11.

barthalion:
bot, build com.zoho.Notebook

barthalion:
bot, build com.zoho.Notebook

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+    "automerge-flathubbot-prs": true
```
hfiguiere, line 2: there is no automation that warrant this.

com.zoho.Notebook.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - tar -xf data.tar.xz
+      - rm -f control.tar.gz data.tar.xz debian-binary
+      - cp -r opt/* usr/* $FLATPAK_DEST
+      - rm -rf opt usr
+      - install notebook.sh $FLATPAK_DEST/bin/notebook
+      - install -D -m 644 -t $FLATPAK_DEST/share/metainfo $FLATPAK_ID.metainfo.xml
+      - desktop-file-edit --set-key=Exec --set-value="notebook %U" $FLATPAK_DEST/share/applications/notebook.desktop
+    sources:
+      - type: file
+        url: https://downloads.zohocdn.com/notebooklinux-desktop/Notebook-2.0.4.deb
```
hfiguiere, line 33: are you sure this is architecture independent? because it is contains an x86_64 binary, this will be broken on aarch64.

DenysMb: Actually, I don't. At first it had the arches rules but I removed because I wasn't sure. I will just add it again, since I don't have any way to test in another arch.

barthalion:
It doesn't seem to work well on my laptop. 

image

DenysMb:
> It doesn't seem to work well on my laptop.

It was working on my openSUSE Tumbleweed system but not on my openSUSE MicroOS (same problem as yours).
After went to Flatseal and enable `D-Bus session bus` it worked.
I will change the .yaml file to add it.

com.zoho.Notebook.yaml
```diff
@@ -0,0 +1,42 @@
...
+runtime: org.freedesktop.Platform
+runtime-version: '22.08'
+sdk: org.freedesktop.Sdk
+base: org.electronjs.Electron2.BaseApp
+base-version: '22.08'
+command: notebook
+separate-locales: false
+finish-args:
+  - --share=ipc
+  - --socket=session-bus
```
hfiguiere, line 11: this is not allowed. You need to determine which dbus name the app needs and add them individually.

DenysMb: Alright!

hfiguiere:
bot, build com.zoho.Notebook

PR #3922, opened 2023-02-21
Add app.drey.Elastic

Exalm:

The patch is temporary. The app was written against nightly GNOME SDK, and needs newer libadwaita (which is important here, otherwise there's no way to draw the animation graph), which in turn needs newer GTK.

Now, GTK deprecated `GtkStyleContext` this cycle and instead added `gtk_widget_get_color()`, and the app is using that. However, GTK also doesn't ship its own vapi and I would need to bundle Vala as well to get that func. So instead it's easier to bundle a patch reverting to the older func for now.

Once 44 SDK is available, both the patch and all of the deps except `template-glib` can go away, for comparison here's the upstream manifest: https://gitlab.gnome.org/exalm/elastic/-/blob/main/app.drey.Elastic.json

### DIFF
--- a/app.drey.Elastic.json
+++ b/app.drey.Elastic.json
@@ -1,0 +1,134 @@
+{
+    "app-id" : "app.drey.Elastic",
+    "runtime" : "org.gnome.Platform",
+    "runtime-version" : "43",
+    "sdk" : "org.gnome.Sdk",
+    "sdk-extensions" : [
+        "org.freedesktop.Sdk.Extension.vala"
+    ],
+    "command" : "app.drey.Elastic",
+    "finish-args" : [
+        "--share=network",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--device=dri"
+    ],
+    "build-options" : {
+        "append-path" : "/usr/lib/sdk/vala/bin",
+        "prepend-ld-library-path" : "/usr/lib/sdk/vala/lib"
+    },
+    "cleanup" : [
+        "/include",
+        "/lib/pkgconfig",
+        "/man",
+        "/share/doc",
+        "/share/gtk-doc",
+        "/share/man",
+        "/share/pkgconfig",
+        "/share/vala",
+        "*.la",
+        "*.a"
+    ],
+    "modules" : [
+        {
+            "name" : "libsass",
+            "buildsystem" : "meson",
+            "cleanup" : [
+                "*"
+            ],
+            "sources" : [
+                {
+                    "type" : "git",
+                    "url" : "https://github.com/lazka/libsass.git",
+                    "branch" : "meson",
+                    "commit" : "aac79dccd3c8f7e8f22125f87a119f3b1ee9d487"
+                }
+            ]
+        },
+        {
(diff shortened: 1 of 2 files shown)


Exalm:
bot, build

Exalm:
bot, build app.drey.Elastic

app.drey.Elastic.json
```diff
@@ -0,0 +1,134 @@
...
+    "app-id" : "app.drey.Elastic",
+    "runtime" : "org.gnome.Platform",
+    "runtime-version" : "43",
+    "sdk" : "org.gnome.Sdk",
+    "sdk-extensions" : [
+        "org.freedesktop.Sdk.Extension.vala"
+    ],
+    "command" : "app.drey.Elastic",
+    "finish-args" : [
+        "--share=network",
```
Exalm, line 11: Whoops, missed this one. It's not needed.

app.drey.Elastic.json
```diff
@@ -0,0 +1,133 @@
+{
+    "app-id" : "app.drey.Elastic",
+    "runtime" : "org.gnome.Platform",
+    "runtime-version" : "43",
+    "sdk" : "org.gnome.Sdk",
+    "sdk-extensions" : [
+        "org.freedesktop.Sdk.Extension.vala"
```
hfiguiere, line 7: Vala is in the GNOME SDK. Why do you need this?

app.drey.Elastic.json
```diff
@@ -0,0 +1,133 @@
...
+            "name" : "libsass",
+            "buildsystem" : "meson",
+            "cleanup" : [
+                "*"
+            ],
+            "sources" : [
+                {
+                    "type" : "git",
+                    "url" : "https://github.com/lazka/libsass.git",
+                    "branch" : "meson",
```
hfiguiere, line 43:
```suggestion
```

NEVER branch

(this should be a linter error)

app.drey.Elastic.json
```diff
@@ -0,0 +1,133 @@
...
+            "name" : "sassc",
+            "buildsystem" : "meson",
+            "cleanup" : [
+                "*"
+            ],
+            "sources" : [
+                {
+                    "type" : "git",
+                    "url" : "https://github.com/lazka/sassc.git",
+                    "branch" : "meson",
```
hfiguiere, line 58:
```suggestion
```

NEVER branch.

app.drey.Elastic.json
Exalm: The language server is in that extension. But true, not needed for flathub, only upstream.

app.drey.Elastic.json
Exalm: https://github.com/search?q=org%3Aflathub+%22branch%22%3A+%22meson%22&type=code well :) But ok.

Exalm:
bot, build app.drey.Elastic

app.drey.Elastic.json
hfiguiere: it is an error.

Exalm: I'm sure it is, but also looks rather widespread.

PR #3925, opened 2023-02-23
add io.github.rocket.editor

kusma:

### DIFF
--- a/io.github.rocket.editor.appdata.xml
+++ b/io.github.rocket.editor.appdata.xml
@@ -1,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2023 Erik Faye-Lund <kusmabite@gmail.com> -->
+<component type="desktop-application">
+  <id>io.github.rocket.editor</id>
+  <metadata_license>FSFAP</metadata_license>
+  <project_license>Zlib</project_license>
+  <name>Rocket</name>
+  <summary>Rocket Editor</summary>
+
+  <description>
+    <p>
+      Rocket is an intuitive new way of... bah, whatever. It's a
+      sync-tracker, a tool for synchronizing music and visuals in
+      demoscene productions. It consists of a GUI editor (using Qt5), and
+      an ANSI C library that can either communicate with the editor over
+      a network socket, or play back an exported data-set.
+    </p>
+  </description>
+
+  <launchable type="desktop-id">io.github.rocket.editor.desktop</launchable>
+
+  <screenshots>
+    <screenshot type="default">
+      <image>https://user-images.githubusercontent.com/47073/220860003-4feef034-a937-4fdc-9e27-7cb4d710516c.png</image>
+    </screenshot>
+  </screenshots>
+
+  <url type="homepage">https://rocket.github.io/</url>
+
+  <content_rating type="oars-1.1" />
+
+  <releases>
+    <release version="0.11.0" date="2018-08-01" />
+  </releases>
+</component>
--- a/io.github.rocket.editor.yml
+++ b/io.github.rocket.editor.yml
@@ -1,0 +1,29 @@
+app-id: io.github.rocket.editor
+runtime: org.kde.Platform
+runtime-version: '5.15-22.08'
+sdk: org.kde.Sdk
+command: editor
+finish-args:
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=network
+  - --device=dri
+modules:
+  - name: Rocket
+    buildsystem: qmake
+    no-make-install: true
+    subdir: editor
+    post-install:
+      - install -Dm644 ../contrib/logo.svg /app/share/icons/hicolor/scalable/apps/io.github.rocket.editor.svg
+      - install -Dm644 -t /app/share/applications/ ../io.github.rocket.editor.desktop
+      - install -Dm644 -t /app/share/metainfo/ ../io.github.rocket.editor.appdata.xml
+      - install -D -t /app/bin/ ./editor
+    sources:
+      - type: git
+        url: https://github.com/rocket/rocket.git
+        tag: v0.11.0
+      - type: file
+        path: io.github.rocket.editor.desktop
+      - type: file
+        path: io.github.rocket.editor.appdata.xml
--- a/io.github.rocket.editor.desktop
+++ b/io.github.rocket.editor.desktop
@@ -1,0 +1,8 @@
+[Desktop Entry]
+Type=Application
+Name=Rocket
+Comment=Rocket Editor
+Exec=editor
+Icon=io.github.rocket.editor
+Categories=Development;Qt;X-Demoscene-Tools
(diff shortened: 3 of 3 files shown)


io.github.rocket.editor.appdata.xml
```diff
@@ -0,0 +1,35 @@
...
+    </screenshot>
+  </screenshots>
+
+  <url type="homepage">https://rocket.github.io/</url>
+
+  <content_rating type="oars-1.1" />
+
+  <releases>
+    <release version="0.11.0" date="2018-08-01" />
+  </releases>
```
kusma, line 34: It's a bit unclear to me if the releases should be *all* releases ever made (including source-only releases), or just the ones for this flatpak. If it's the former, I guess I'm missing a few here. If it's the latter, the question becomes if the date should be the date of the source-release or the flatpak release. Those differ quite widely in this case...

JakobDev: It only requires the latest release. You can add former if you want, so Software Centers can show a full changelog, but it is not required.

JakobDev:
You should upstream the AppStream, the Desktop Entry and the Icon.

JakobDev:
bot, build io.github.rocket.editor

io.github.rocket.editor.yml
```diff
@@ -0,0 +1,33 @@
+app-id: io.github.rocket.editor
+runtime: org.kde.Platform
+runtime-version: '5.15-22.08'
+sdk: org.kde.Sdk
+command: editor
+finish-args:
+  - --socket=wayland
```
JakobDev, line 7:
```suggestion
  - --socket=wayland
  - --socket=fallback-x11
```
Or else it would not work under X11

kusma:
> Error: Command 'editor' not found

Hmm, that's what I got when I didn't patch the install target. With that patch, it works fine locally. I wonder what's going on...

kusma:
> You should upstream the AppStream, the Desktop Entry and the Icon

Yes, I intend to. The upstream project isn't in a great state right now, with our CI being broken. I'm working on that right now.

io.github.rocket.editor.appdata.xml
hfiguiere: it's always when upstream released. I usually take the tag date if I have to guess, or the tarball date.

io.github.rocket.editor.yml
```diff
@@ -0,0 +1,34 @@
...
+  - --socket=fallback-x11
+  - --share=network
+  - --device=dri
```
hfiguiere, line 10: `- --share=ipc` is also needed.

io.github.rocket.editor.yml
```diff
@@ -0,0 +1,34 @@
...
+      - type: file
+        path: io.github.rocket.editor.svg
+      - type: file
+        path: io.github.rocket.editor.desktop
+      - type: file
+        path: io.github.rocket.editor.appdata.xml
+    build-commands:
+      - install -Dm644 -t /app/share/icons/hicolor/scalable/apps io.github.rocket.editor.svg
+      - install -Dm644 -t /app/share/applications/ io.github.rocket.editor.desktop
+      - install -Dm644 -t /app/share/metainfo io.github.rocket.editor.appdata.xml
```
hfiguiere, line 34:
```suggestion
    post-install:
      - install -Dm644 -t /app/share/icons/hicolor/scalable/apps ../io.github.rocket.editor.svg
      - install -Dm644 -t /app/share/applications/ ../io.github.rocket.editor.desktop
      - install -Dm644 -t /app/share/metainfo ../io.github.rocket.editor.appdata.xml
    sources:
      - type: git
        url: https://github.com/rocket/rocket.git
        tag: v0.11.0
      - type: patch
        path: 0001-install-editor.patch
        use-git: true
      - type: file
        path: io.github.rocket.editor.svg
      - type: file
        path: io.github.rocket.editor.desktop
      - type: file
        path: io.github.rocket.editor.appdata.xml
```

kusma: Yeah, that seems nicer!

io.github.rocket.editor.yml
kusma: Applied that change, and squashed it all! Thanks!

kusma:
> > Error: Command 'editor' not found
> 
> Hmm, that's what I got when I didn't patch the install target. With that patch, it works fine locally. I wonder what's going on...

Just tried again, and I can't reproduce it. Anyone able to trigger another build and see if it still happens?

kusma:
It's really strange.

```
make: Nothing to be done for 'install'.
```

This seems to indicate that the patching didn't work; there's nothing in the install target.

But I see the patching work:

```
Applying patch 0001-install-editor.patch
FB: Running: git apply -v -p1 /srv/buildbot/worker/build-x86_64-2/build/0001-install-editor.patch
Checking patch editor/editor.pro...
Applied patch editor/editor.pro cleanly.
```

I'm really lost as to what's going wrong here...

hfiguiere:
the previous failed build say `make: Nothing to be done for 'install'.`

You can still install it in the post-install. That should be enough.

kusma:
bot, build io.github.rocket.editor

io.github.rocket.editor.yml
```diff
@@ -0,0 +1,32 @@
...
+      - install -Dm644 -t /app/share/metainfo ../io.github.rocket.editor.appdata.xml
+    sources:
+      - type: git
+        url: https://github.com/rocket/rocket.git
+        tag: v0.11.0
+      - type: patch
+        path: 0001-install-editor.patch
+        use-git: true
+      - type: file
+        path: io.github.rocket.editor.svg
```
hfiguiere, line 28: you should be able to just use the file in `contrib/logo.svg` instead.

kusma:
@hfiguiere: Yeah, that's a fair point. Maybe that's a bit less error-prone in the first place.

io.github.rocket.editor.yml
kusma: Yeah, that's the same file, good point.

hfiguiere:
also is it supposed to play sounds? I see no permission for it if that's the case.

kusma:
> also is it supposed to play sounds? I see no permission for it if that's the case.

No, it doesn't play sound by itself. That's up to the "remote end", which is connected over a socket, and something that I don't control.

kusma:
bot, build io.github.rocket.editor

io.github.rocket.editor.yml
kusma: OK, done.

kusma:
bot, build io.github.rocket.editor

kusma:
OK, seems close. The built version works fine here, but: the screenshot doesn't load for me any more. It did before, so I'm not quite sure what's up with that. Is this maybe some optimization for non-production builds that's expected?

kusma:
With the local build, the screenshot works fine...

kusma:
Tested the aarch64 build as well, and it seems to work just fine!

PR #3945, opened 2023-03-01
Add Black Chocobo

sithlord48:

- [X] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub.     Link:  www.github.com/sithlord48/blackchocobo
  Stable version needs the patches,  future releases have them [MR#57](https://github.com/sithlord48/blackchocobo/pull/57)  & [MR#58](https://github.com/sithlord48/blackchocobo/pull/58)

### DIFF
--- a/org.sithlord48.BlackChocobo.yml
+++ b/org.sithlord48.BlackChocobo.yml
@@ -1,0 +1,58 @@
+app-id: org.sithlord48.BlackChocobo
+default-branch: stable
+runtime: org.kde.Platform
+runtime: org.kde.Sdk.Locale
+sdk: org.kde.Sdk
+runtime-version: 6.4
+command: blackchocobo
+finish-args:
+  - --filesystem=host
+  - --socket=fallback-x11
+  - --socket=wayland
+
+cleanup:
+  - /include
+  - /lib/*.a
+  - /lib/*.la
+  - /lib/cmake
+  - /lib/debug
+  - /lib/pkgconfig
+  - /share/cmake
+  - /share/man
+
+modules:
+  - name: ff7tk
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+      - -DDEMOS=OFF
+      - -DFORMATS=OFF
+      - -DOCS=OFF
+      - -DUTILS=OFF
+    sources:
+      - type: archive
+        url: https://github.com/sithlord48/ff7tk/archive/refs/tags/v0.83.1.tar.gz
+        sha256: 94a33f0b9fa77c43d0a3bf47939b97e5bc43bc6e2ec5d8b732a59d00e2be4298
+        x-checker-data:
+          type: html
+          url: https://api.github.com/repos/sithlord48/ff7tk/tags
+          version-query: map(.name | select(test("^release-[\\d\\.]+$"))) | first
+            | sub("^release-"; "")
+          url-query: '"https://github.com/sithlord48/ff7tk/archive/refs/tags/release-"
+            + $version + ".tar.gz"'
+
+  - name: blackchocobo
+    buildsystem: cmake-ninja
+    builddir: true
+    sources:
+      - type: archive
+        url: https://github.com/sithlord48/blackchocobo/archive/refs/tags/v1.13.0.tar.gz
+        sha256: a046e23d8eb31d1c3f3e16c4e24ecd1ddfaf5127f67217ae63d370ba25960a5d
+        x-checker-data:
+          type: json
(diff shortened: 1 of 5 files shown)


org.sithlord48.BlackChocobo.yml
```diff
@@ -0,0 +1,58 @@
+app-id: org.sithlord48.BlackChocobo
```
hfiguiere, line 1: `sithlord48.org` doesn't exist. the app-id doesn't meet the requirements.

org.sithlord48.BlackChocobo.yml
```diff
@@ -0,0 +1,58 @@
...
+default-branch: stable
```
hfiguiere, line 2:
```suggestion
```

org.sithlord48.BlackChocobo.yml
```diff
@@ -0,0 +1,58 @@
...
+runtime: org.kde.Platform
+runtime: org.kde.Sdk.Locale
```
hfiguiere, line 4:
```suggestion
```

org.sithlord48.BlackChocobo.yml
```diff
@@ -0,0 +1,58 @@
...
+sdk: org.kde.Sdk
+runtime-version: 6.4
+command: blackchocobo
+finish-args:
+  - --filesystem=host
```
hfiguiere, line 9: that's a very broad permission

org.sithlord48.BlackChocobo.yml
```diff
@@ -0,0 +1,58 @@
...
+  - --socket=fallback-x11
+  - --socket=wayland
```
hfiguiere, line 11: missing `--share=ipc`

org.sithlord48.BlackChocobo.yml
sithlord48: Is there a more limited way to allow r/w of user files

sithlord48: Changed to home

com.github.sithlord48.blackchocobo.yml
```diff
@@ -0,0 +1,59 @@
+app-id: com.github.sithlord48.blackchocobo
```
hfiguiere, line 1: it should be `io.github` as per the documentation.

com.github.sithlord48.blackchocobo.yml
```diff
@@ -0,0 +1,59 @@
...
+runtime: org.kde.Platform
+sdk: org.kde.Sdk
+runtime-version: '6.4'
+command: blackchocobo
+finish-args:
+  - --filesystem=home
+  - --filesystem=xdg-data
```
hfiguiere, line 8:
```suggestion
```
no.

org.sithlord48.BlackChocobo.yml
sithlord48: will use `com.github.sithlord48.blackchocobo`

sithlord48:
I have to use `--filesystem=host` 

There is no way to override the portal dialog(s) to allow a save name to be suggested when the current save as type changes. Todo this I am forced to use the Qt's Non Native Dialog. I'm not talking about just adding ".ext" to a file either just about every format that is save has some rules around what file names are valid so for each type a suggestion is provided. For example the psv format requires the output filename to be parts of the file contents in the form of shift-jis hex char values. Its just not reasonable to expect my users to know these things[[1]](https://github.com/sithlord48/blackchocobo/wiki/PSX-Formats#psv-format). That is one reason they are using this app and not hexediting by hand, While There is an option in the application to use force the use of the os' native dialog when doing so users do lose all name suggestions. For this reason I had to switch back using host its the only way I can give users both suggestions and the ability to save / load from mounted usb / network drives . This app works with console saves so users often are reading from a mounted device outside of their home folder.

barthalion:
bot, build io.github.sithlord48.blackchocobo

barthalion:
f-e-d-c stanza fails validation

sithlord48:
> f-e-d-c stanza fails validation

Fixed

barthalion:
bot, build io.github.sithlord48.blackchocobo

sithlord48:
Use correct name `metainfo`  not `metadata`

barthalion:
bot, build io.github.sithlord48.blackchocobo

sithlord48:
Fixed: / instead of . for path

sithlord48:
bot, build io.github.sithlord48.blackchocobo

sithlord48:
bot, build io.github.sithlord48.blackchocobo

sithlord48:
bot, build io.github.sithlord48.blackchocobo

PR #3958, opened 2023-03-06
dev.deedles.Trayscale

DeedleFake:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,11 @@
+Trayscale
+=========
+
+Trayscale is an unofficial GUI wrapper around the Tailscale CLI client, particularly for use on Linux, as no official Linux GUI client exists. Despite the name, it does _not_ provide a tray icon, as support for them has been removed in Gtk4. If support can ever be brought back, however, a tray icon would be nice.
+
+![image](https://user-images.githubusercontent.com/326750/188052311-2267af08-82a1-422f-b6ad-bc2cd4de0ac5.png)
+
+Tailscale Config
+----------------
+
+Trayscale makes calls to the Tailscale CLI for some operations. In order for this to work, the `tailscale` command must be in your `$PATH`. Additionally, the daemon must have been configured with the current user as the "operator". To do this, run `sudo tailscale up --operator=$USER` from the command-line at least once manually.
--- a/README.md
+++ b/README.md
@@ -1,11 +1,1 @@
-Trayscale
-=========
-
-Trayscale is an unofficial GUI wrapper around the Tailscale CLI client, particularly for use on Linux, as no official Linux GUI client exists. Despite the name, it does _not_ provide a tray icon, as support for them has been removed in Gtk4. If support can ever be brought back, however, a tray icon would be nice.
-
-![image](https://user-images.githubusercontent.com/326750/188052311-2267af08-82a1-422f-b6ad-bc2cd4de0ac5.png)
-
-Tailscale Config
-----------------
-
-Trayscale makes calls to the Tailscale CLI for some operations. In order for this to work, the `tailscale` command must be in your `$PATH`. Additionally, the daemon must have been configured with the current user as the "operator". To do this, run `sudo tailscale up --operator=$USER` from the command-line at least once manually.
+Source repo: https://github.com/DeedleFake/trayscale
--- a/dev.deedles.Trayscale.yml
+++ b/dev.deedles.Trayscale.yml
@@ -25,7 +25,7 @@ modules:
   - name: trayscale
     buildsystem: simple
     build-commands:
-      - $GOROOT/bin/go build -v -trimpath -ldflags="-X 'deedles.dev/trayscale/internal/version.version=v0.7.12'" ./cmd/trayscale
+      - $GOROOT/bin/go build -v -trimpath -ldflags="-X 'deedles.dev/trayscale/internal/version.version=v0.7.13'" ./cmd/trayscale
       - install -Dm0755 trayscale $FLATPAK_DEST/bin/trayscale
       - install -Dm0644 dev.deedles.Trayscale.png $FLATPAK_DEST/share/icons/hicolor/256x256/apps/dev.deedles.Trayscale.png
       - install -Dm0644 dev.deedles.Trayscale.desktop $FLATPAK_DEST/share/applications/dev.deedles.Trayscale.desktop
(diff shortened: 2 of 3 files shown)


dev.deedles.Trayscale.yml
```diff
@@ -0,0 +1,49 @@
...
+runtime-version: '43'
+sdk: org.gnome.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.golang
+
+command: trayscale
+
+finish-args:
+  - --socket=wayland
+  - --socket=x11
```
hfiguiere, line 13: if it support wayland then this should be 'fallback-x11'

dev.deedles.Trayscale.yml
```diff
@@ -0,0 +1,49 @@
...
+  - --device=dri
+  - --share=network
+  - --share=ipc
+  - --filesystem=/run/tailscale:ro
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.freedesktop.Flatpak
+
+build-options:
+  build-args:
+    - --filesystem=home # To access the Go build cache.
```
hfiguiere, line 23: which cache?

README.md
```diff
@@ -0,0 +1,11 @@
...
+=========
+
+Trayscale is an unofficial GUI wrapper around the Tailscale CLI client, particularly for use on Linux, as no official Linux GUI client exists. Despite the name, it does _not_ provide a tray icon, as support for them has been removed in Gtk4. If support can ever be brought back, however, a tray icon would be nice.
+
+![image](https://user-images.githubusercontent.com/326750/188052311-2267af08-82a1-422f-b6ad-bc2cd4de0ac5.png)
+
+Tailscale Config
+----------------
+
+Trayscale makes calls to the Tailscale CLI for some operations. In order for this to work, the `tailscale` command must be in your `$PATH`. Additionally, the daemon must have been configured with the current user as the "operator". To do this, run `sudo tailscale up --operator=$USER` from the command-line at least once manually.
```
hfiguiere, line 11: no user will ever see this.

DeedleFake: Ah, whoops. I thought it would be used for a description on Flathub.

dev.deedles.Trayscale.yml
DeedleFake: The Go toolchain caches previously compiled packages. The default location is `$XDG_CACHE_HOME/go-build`. It was just to make repeat builds on the local system faster. If it's an issue, I can remove it.

README.md
dginovker: Flathub description comes from Appdata

dev.deedles.Trayscale.yml
```diff
@@ -0,0 +1,49 @@
...
+
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --device=dri
+  - --share=network
+  - --share=ipc
+  - --filesystem=/run/tailscale:ro
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.freedesktop.Flatpak
```
barthalion, line 19: Wouldn't it make more sense to bundle tailscale CLI with the app?

DeedleFake: That's an option. I _had_ looked into bundling the whole of Tailscale, including `tailscaled`, with the app, but that didn't work because it needed root permissions to create a bridge network device. It might be possible to bundle just the `tailscale` binary, but it might require the version to match between `tailscaled` and `tailscale`, in which case that'll make usage rather awkward. I'll try bundling it with a version purposefully mismatched from my system's and see what happens.

dev.deedles.Trayscale.yml
DeedleFake: Is there any way to pass them to `flatpak-builder` on the command line so that I can replicate the effect locally?

DeedleFake: On a related note, is there a way to specify the version for entries in `sdk-extensions`? I couldn't find anything in the documentation. The current one seems to be defaulting to `22.08` which is a version of Go behind.

hfiguiere: 22.08 is the latest SDK, and matches GNOME 43. There is no way to specify a specific go version. 

The SDK should be updated soon. (it should have)

dev.deedles.Trayscale.yml
DeedleFake: Bundling it worked, though it does add architecture-specific sources. I was able to remove the `--talk-name=org.freedesktop.Flatpak` permission because of it, though.

dev.deedles.Trayscale.yml
```diff
@@ -0,0 +1,55 @@
...
+      - install -Dm0755 tailscale $FLATPAK_DEST/bin/tailscale
+    sources:
+      - type: archive
+        only-arches: [x86_64]
+        url: https://pkgs.tailscale.com/stable/tailscale_1.36.2_amd64.tgz
+        sha256: 696d08ec8b9a73fe8f3cb537eaec93b45ee621d8bf0b7f5b954359a17ad185df
+      - type: archive
+        only-arches: [i386]
+        url: https://pkgs.tailscale.com/stable/tailscale_1.36.2_386.tgz
+        sha256: f874aef63a6e101b77e49c9e1a133f67f0b8849979f14561ae3dfe6912a41f64
```
hfiguiere, line 55:
```suggestion
```

Won't support i386

However aarch64 is.....

hfiguiere:
bot, build dev.deedles.Trayscale

dev.deedles.Trayscale.yml
hfiguiere: why not building it from source?

DeedleFake: I didn't want to have to bundle a `vendor` directory for their dependencies, but I guess it's probably less of a problem than trying to support every architecture of pre-built binaries. I've switched it to build it.

hfiguiere: there is that tool to generate a manifest for go dependencies.
https://github.com/flatpak/flatpak-builder-tools/tree/master/go-get

DeedleFake: I tried that originally but it didn't work because it requires that the initial build be done in GOPATH mode, and many newer modules don't work correctly in GOPATH mode, resulting in build failures. Vendoring makes it very easy to deal with, though, as I can just produce a bundle of all of the dependencies in a single archive. The only really awkward thing is trying to figure out where to host it.

hfiguiere: what works the best.

barthalion:
bot, build dev.deedles.Trayscale

PR #3977, opened 2023-03-12
Add the mouse-tracker extension for pro.vpup.vpuppr

orowith2os:

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link**: https://github.com/virtual-puppet-project/mouse-tracker/issues/2

Relies on https://github.com/flathub/pro.vpup.vpuppr/pull/2

Feedback very much appreciated.

### DIFF
--- a/pro.vpup.vpuppr.Extension.mouse-tracker.yml
+++ b/pro.vpup.vpuppr.Extension.mouse-tracker.yml
@@ -10,6 +10,8 @@ modules:
     buildsystem: simple
     build-commands:
       - ls -a
+
+      - mv * /app/share/pro.vpup.vpuppr/extensions/mouse-tracker/
         #post-install:
         #- install -Dm644 -t ${FLATPAK_DEST}/share/metainfo data/${FLATPAK_ID}.metainfo.xml
         #- appstream-compose --basename=${FLATPAK_ID} --prefix=${FLATPAK_DEST} --origin=flatpak ${FLATPAK_ID}
--- a/pro.vpup.vpuppr.Extension.mouse-tracker.yml
+++ b/pro.vpup.vpuppr.Extension.mouse-tracker.yml
@@ -5,17 +5,21 @@ runtime-version: stable
 sdk: org.freedesktop.Sdk//22.08
 build-extension: true
 appstream-compose: false
+build-options:
+  prefix: /app/extensions/mouse-tracker
 modules:
   - name: mouse-tracker
     buildsystem: simple
     build-commands:
-      - ls -a
-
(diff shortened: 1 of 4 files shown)


orowith2os:
Builds currently fail with 
```sh
Running: ls -a
bwrap: Can't mkdir /app/utils: Read-only file system
Error: module mouse-tracker: Child process exited with code 1
```

pro.vpup.vpuppr.Extension.mouse-tracker.yml
```diff
@@ -0,0 +1,19 @@
...
+build-extension: true
+appstream-compose: false
+modules:
+  - name: mouse-tracker
+    buildsystem: simple
+    build-commands:
+      - ls -a
+        #post-install:
+        #- install -Dm644 -t ${FLATPAK_DEST}/share/metainfo data/${FLATPAK_ID}.metainfo.xml
+        #- appstream-compose --basename=${FLATPAK_ID} --prefix=${FLATPAK_DEST} --origin=flatpak ${FLATPAK_ID}
```
hfiguiere, line 15: so the package is empty.

orowith2os: For now, yes. I want to figure out why the build fails first. I should've marked this as draft, since not everything is in place yet. After I figure out why builds fail, everything else should be easy to do.

orowith2os:
I figured out what was going on. I was misunderstanding how add-extensions worked. I've got some local changes I'll push tomorrow that fix it up, for the most part. Upstream needs a new changes too.

orowith2os:
bot, build pro.vpup.vpuppr.Extension.mouse-tracker

orowith2os:
bot, build pro.vpup.vpuppr.Extension.mouse-tracker

orowith2os:
I can confirm that this is mounted inside of the environment properly, all that's left is for upstream to search /app/extensions for extensions. Don't merge. I'll link the upstream issue when I make it.

barthalion:
Given a relatively low development activity of this project, does it make sense to split this into an extension instead of just bundling with the main app? It's not a third party plugin and uses a binary version anyway.

orowith2os:
I think it should be separate, at least for the fact that it would be used as a reference for how extensions for vpuppr would be built inside of Flatpak. It also looks like I might be able to build this extension from source, since it doesn't seem to require Godot itself to build.

As for the low development: right now, the focus is more on the Godot 4 port, so the original vpuppr won't get too many updates. The latest stable was more of one that mostly works to satisfy users

orowith2os:
This shouldn't have been merged, upstream still needs to work around the plugins system not working. Can you delete this from flathub please?

PR #3986, opened 2023-03-14
Add com.protonvpn.www

proletarius101:

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://github.com/ProtonVPN/linux-app/issues/7#issuecomment-1468638156

---

~~To reviewers: we will need to allow `--talk-name=org.freedesktop.Flatpak`~~ This permission is no longer required.

```yaml
  # To store credentials
  - --talk-name=org.freedesktop.secrets
  # To check the Network Manager status on the host system
  - --system-talk-name=org.freedesktop.NetworkManager
```

---


com.protonvpn.ProtonVPN.yml
```diff
@@ -0,0 +1,501 @@
...
+  - org.freedesktop.Sdk.Extension.rust-stable
+command: protonvpn
+separate-locales: false
+finish-args:
+  - --device=dri
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --talk-name=org.freedesktop.Notifications
```
A6GibKm, line 17: I could not find anywhere in the app's code a direct call to dbus or a use of libnotify, what justifies this?

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+    "automerge-flathubbot-prs": true
```
A6GibKm, line 2: I would advice against this.

patches/protonvpn-gui/application-id.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/protonvpn_gui/main.py b/protonvpn_gui/main.py
+index 96f1a11..312de6a 100644
+--- a/protonvpn_gui/main.py
++++ b/protonvpn_gui/main.py
+@@ -36,7 +36,7 @@ class ProtonVPNGUI(Gtk.Application):
+ 
+     def __init__(self):
+         super().__init__(
+-            application_id='com.protonvpn.www',
```
A6GibKm, line 9: Why this change? If the app id is com.protonvpn.www then so it is, this should be brought up upstreams instead. App ids in theory are not a user facing thing.

Note that changing the app id can unforeseen issues.

patches/protonvpn-gui/webkit-version.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/protonvpn_gui/view/dialog.py b/protonvpn_gui/view/dialog.py
+index e26f815..f8adf3a 100644
+--- a/protonvpn_gui/view/dialog.py
++++ b/protonvpn_gui/view/dialog.py
+@@ -3,7 +3,7 @@ import os
+ import gi
+ 
+ gi.require_version("Gtk", "3.0")
+-gi.require_version("WebKit2", "4.0")
++gi.require_version("WebKit2", "4.1")
```
A6GibKm, line 10: If possible this should be upstreamed with the explanation of why it is good (uses libsoup 3). Note that one can load 4.1 or 4.0 dynamically depending on what the system has, see https://github.com/lutris/lutris/pull/4554

com.protonvpn.ProtonVPN.metainfo.xml
```diff
@@ -0,0 +1,68 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
A6GibKm, line 1: This should be submitted upstream, all apps should install a metainfo file

com.protonvpn.ProtonVPN.metainfo.xml
```diff
@@ -0,0 +1,68 @@
...
+            <li>MULTIPLE PLATFORM SUPPORT - Easy to use on Android, iOS, Windows, macOS, Linux, and
+                more.</li>
+        </ul>
+        <p>Note: Free users can only connect one device to Free servers in Japan, the Netherlands,
+            and the US.</p>
+    </description>
+    <screenshots>
+        <screenshot type="default">
+            <image>
+                https://images.prismic.io/gatsby-landing-pages/e6a71a70-741e-44ed-a947-439c0ea92801_secure-core.png</image>
```
A6GibKm, line 54: Is it possible to use a screenshot from an official source? If not possible screenshots should not include desktop backgrounds (but should include window shadows).

com.protonvpn.ProtonVPN.metainfo.xml
```diff
@@ -0,0 +1,68 @@
...
+    <url type="homepage">https://protonvpn.com/</url>
+    <url type="bugtracker">https://github.com/ProtonVPN/linux-app/issues</url>
+    <url type="vcs-browser">https://github.com/ProtonVPN/linux-app</url>
+
+    <releases>
+        <release version="1.12.0" date="2022-10-02" />
+    </releases>
+
+    <content_rating type="oars-1.1">
+        <content_attribute id="money-purchasing">intense</content_attribute>
```
A6GibKm, line 66: I am not sure about this one it looks weird and might be referring to something else, maybe someone more knowledgeable could review this.

patches/protonvpn-gui/application-id.patch
hfiguiere: this also probably require adding a `own-name` if that's needed as it wouldn't be granted by default.

A6GibKm:
Also the app is not running on my machine

```
flatpak run com.protonvpn.ProtonVPN
Portal call failed: org.freedesktop.DBus.Error.ServiceUnknown
Portal call failed: org.freedesktop.DBus.Error.ServiceUnknown
Portal call failed: org.freedesktop.DBus.Error.ServiceUnknown
Portal call failed: org.freedesktop.DBus.Error.ServiceUnknown
Portal call failed: org.freedesktop.DBus.Error.ServiceUnknown
  File "/app/bin/protonvpn", line 5, in <module>
    from protonvpn_gui.main import main
  File "/app/lib/python3.10/site-packages/protonvpn_gui/main.py", line 16, in <module>
    from protonvpn_nm_lib.api import protonvpn
  File "/app/lib/python3.10/site-packages/protonvpn_nm_lib/api.py", line 6, in <module>
    from .core.report import BugReport
  File "/app/lib/python3.10/site-packages/protonvpn_nm_lib/core/report/__init__.py", line 1, in <module>
    from .bug import BugReport
  File "/app/lib/python3.10/site-packages/protonvpn_nm_lib/core/report/bug.py", line 7, in <module>
    from ..subprocess_wrapper import subprocess
  File "/app/lib/python3.10/site-packages/protonvpn_nm_lib/core/subprocess_wrapper.py", line 147, in <module>
    subprocess = SubprocessWrapper() # noqa
  File "/app/lib/python3.10/site-packages/protonvpn_nm_lib/core/subprocess_wrapper.py", line 66, in __init__
    self.__search_for_matching_executables()
  File "/app/lib/python3.10/site-packages/protonvpn_nm_lib/core/subprocess_wrapper.py", line 86, in __search_for_matching_executables
    if not self.is_root_owned(binary_path_candidate):
  File "/app/lib/python3.10/site-packages/protonvpn_nm_lib/core/subprocess_wrapper.py", line 31, in is_root_owned
    owner_uid = int(_subprocess.check_output(
  File "/usr/lib/python3.10/subprocess.py", line 420, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.10/subprocess.py", line 501, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib/python3.10/subprocess.py", line 969, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.10/subprocess.py", line 1778, in _execute_child
    self.pid = _posixsubprocess.fork_exec(
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

hfiguiere:
It should fail graceful. Not crash.

A6GibKm:
If the app cant run without it, then it has to be in the manifest. There has to be a way to ignore the linter until it is in the allow list

A6GibKm:
No idea, please ask in irc/matrix https://matrix.to/#/#flatpak:matrix.org, but maybe https://matrix.to/#/#flathub:matrix.org is more appropiate.

com.protonvpn.ProtonVPN.yml
```diff
@@ -14,6 +14,8 @@ finish-args:
   - --share=network
   - --socket=wayland
   - --socket=fallback-x11
+  # https://github.com/ProtonVPN/linux-app/blob/23350af7d5117784141786ef88fa58c47a9dbfdc/protonvpn_gui/main.py#L39
+  - --own-name=com.protonvpn.www
```
A6GibKm, line 18: Please just name the json file and everything to the correct id.

A6GibKm:
The app bundles the app icon https://github.com/ProtonVPN/linux-app/blob/master/protonvpn_gui/assets/img/vpn-logo.svg somehow, it should be possible to use the bundled one with `rename-icon` or something, see https://github.com/ProtonVPN/linux-app/blob/master/protonvpn.desktop#L6 .

com.protonvpn.ProtonVPN.yml
A6GibKm: I do, but it seems the app id was choosen incorrectly by upstreams. It might be a good idea to point out that it should be com.protonvpn.protonvpn or com.protonvpn.ProtonVPN or similar.

com.protonvpn.www.yml
```diff
@@ -0,0 +1,513 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/flatpak/flatpak-builder/main/data/flatpak-manifest.schema.json
+
+app-id: com.protonvpn.www
+runtime: org.gnome.Platform
+runtime-version: '43'
+sdk: org.gnome.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+command: protonvpn
+separate-locales: false
```
A6GibKm, line 10: Is there any rationale for this? If not, its better to not add this line.

com.protonvpn.www.yml
```diff
@@ -0,0 +1,513 @@
...
+finish-args:
+  - --device=dri
```
A6GibKm, line 12: Why was this added? This would be absolutely needed if for example the app was GTK 4, but last I checked it used GTK 3.

A6GibKm:
It crashes now with
```
flatpak run --talk-name=org.freedesktop.Flatpak com.protonvpn.ProtonVPN
  File "/app/bin/protonvpn", line 5, in <module>
    from protonvpn_gui.main import main
  File "/app/lib/python3.10/site-packages/protonvpn_gui/main.py", line 16, in <module>
    from protonvpn_nm_lib.api import protonvpn
  File "/app/lib/python3.10/site-packages/protonvpn_nm_lib/api.py", line 6, in <module>
    from .core.report import BugReport
  File "/app/lib/python3.10/site-packages/protonvpn_nm_lib/core/report/__init__.py", line 1, in <module>
    from .bug import BugReport
  File "/app/lib/python3.10/site-packages/protonvpn_nm_lib/core/report/bug.py", line 7, in <module>
    from ..subprocess_wrapper import subprocess
  File "/app/lib/python3.10/site-packages/protonvpn_nm_lib/core/subprocess_wrapper.py", line 147, in <module>
    subprocess = SubprocessWrapper() # noqa
  File "/app/lib/python3.10/site-packages/protonvpn_nm_lib/core/subprocess_wrapper.py", line 67, in __init__
    self.__ensure_executables_exist()
  File "/app/lib/python3.10/site-packages/protonvpn_nm_lib/core/subprocess_wrapper.py", line 101, in __ensure_executables_exist
    raise RuntimeError(
RuntimeError: Couldn't find acceptable executables for {'xdg-open'}
```

A6GibKm:
xdg-open in the sandbox, it should just work. Flatpaks have to work out of the box, having xdg-open in the system is a weird requirement, extra so if I happen to have one variation of xdg-open installed...

hfiguiere:
> > RuntimeError: Couldn't find acceptable executables for {'xdg-open'}
> 
> You have to make `xdg-open` installed in your host system. I'll mention this in the application description.

not on the host, no.

com.protonvpn.www.yml
TingPing: An empty extension does no harm. If a year from now somebody adds a library that happens to have translations it should be a separate locale.

patches/protonvpn-nm-lib/make-subprocess-wrapper-flatpak-aware.patch
```diff
@@ -0,0 +1,105 @@
...
+     def is_root_owned(path):
++        if SubprocessWrapper.is_flatpak():
++            try:
++                owner_uid = int(_subprocess.check_output(
++                    ["flatpak-spawn", "--host", "stat", "-c", "%u", path]).decode().strip())
++                owner_gid = int(_subprocess.check_output(
++                    ["flatpak-spawn", "--host", "stat", "-c", "%g", path]).decode().strip())
++                return owner_uid == owner_gid == 0
++            except _subprocess.CalledProcessError:
++                pass
```
TingPing, line 24: Adding a giant sandbox escape to check if you are root makes no sense at all.

Anyway you should just assume that you are never root in flatpak. Flatpak is not meant to run as root and will not under normal usage.

patches/protonvpn-nm-lib/make-subprocess-wrapper-flatpak-aware.patch
```diff
@@ -0,0 +1,105 @@
...
++
++    @staticmethod
++    def which(binary):
++        """Returns the path to an executable which would be run if the given cmd was called.
++
++        If we are in a Flatpak, get the corresponding host command.
++        """
++        if SubprocessWrapper.is_flatpak():
++            try:
++                return _subprocess.check_output(["flatpak-spawn", "--host", "which", binary]).decode().strip()
```
TingPing, line 42: What are you trying to run on the host and why?

TingPing: I understand you aren't upstream but it should be talking to `org.freedesktop.NetworkManager` and systemd's equivalent dbus. This is a terrible permission.

TingPing: They may limit their own usage but its still a sandbox escape.

I would at least start a discussion with upstream so they know that the DBus APIs are a better way to do this.

hfiguiere: xdg-open should not be called outside the sandbox anyway.

com.protonvpn.www.yml
```diff
@@ -13,6 +13,9 @@ finish-args:
   - --share=network
   - --socket=wayland
   - --socket=fallback-x11
+  # AT-SPI accessibility bus
+  - --talk-name=org.a11y.Bus
+  - --filesystem=xdg-run/at-spi
```
TingPing, line 18: Why are you doing this?

Flatpak sets the `AT_SPI_BUS_ADDRESS` env var and all applications/libraries should be respecting it.

patches/protonvpn-nm-lib/make-subprocess-wrapper-flatpak-aware.patch
TingPing: Hmm, I wonder if `nmcli` actually can use DBus itself. Have you tried building it inside of the package? Looking through the codebase its full of DBus usage.

com.protonvpn.www.yml
A6GibKm: And this is a gtk3 app, it *should* just work.

com.protonvpn.www.yml
A6GibKm: Unless they do some custom gl thing, a general gtk3 wouldn't need this.

com.protonvpn.www.yml
```diff
@@ -15,7 +15,6 @@ finish-args:
   - --socket=fallback-x11
   # AT-SPI accessibility bus
   - --talk-name=org.a11y.Bus
```
A6GibKm, line 17: Please also remove this line, it is not needed

flathub.json
A6GibKm: Except I have seen packages break more often than not with autogenerated updates. Proton-bridge for example breaks around 1/5 of the updates.

com.protonvpn.ProtonVPN.metainfo.xml
A6GibKm: Most apps have to deal with this manually. Some apps just ignore the releases and one just patches in the last one here. Not ideal but works.

com.protonvpn.www.yml
A6GibKm: Good question, I didn't consider that. I asked around and it indeed it is probably a good idea to have it enabled.

com.protonvpn.www.yml
```diff
@@ -0,0 +1,513 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/flatpak/flatpak-builder/main/data/flatpak-manifest.schema.json
+
+app-id: com.protonvpn.www
```
A6GibKm, line 3: Please use `id` instead of `app-id`, the later is deprecated.

kevinlekiller:
Any idea what this means:

    $ flatpak run com.protonvpn.www connect

    Executable not found: can't find the executable for the dialog-like program

kevinlekiller:
Thanks that solved it.

Couple of new issues:

(OpenSuse Tumbleweed KDE Plasma, NetworkManager)

When trying to connect to a server or trying to disconnect, KDE Policy Agent asks for the root password:

Screenshot_20230322_075453

When connecting, the connection dialog hangs, but never connects and network is unavailable (until `flatpak run com.protonvpn.www d` is ran and root password is provided):

Screenshot_20230322_080445

Here's journald output when connecting: 
[connect.log](https://github.com/flathub/flathub/files/11039811/connect.log)

journald output when disconnecting:
[disconnect.log](https://github.com/flathub/flathub/files/11039814/disconnect.log)

kevinlekiller:
Gotcha, wasn't sure if it was something system specific or program specific.

The other issue I think is down to NetworkManager-OpenVPN not being installed. Will try installing that and see if it works.

kevinlekiller:
Installed NetworkManager-OpenVPN (plasma5-nm-openvpn also) package, that got things a bit further, now this message appears when connecting:

Screenshot_20230322_090947

kevinlekiller:
> I can't reproduce this. Can you test whether the official packages have this issue too?

Edited the other post -> I think it's some kind of issue with how it communicates to the keyring.

Unfortunately there's no official package for OpenSuse, I do have protonvpn-cli on Arch which works fine.

I'll try installing this flatpak to the Arch install and see if the same issue exists, but I'm guessing the issue is there's missing some dependencies on OpenSuse.

kevinlekiller:
> Have you installed it from #3986 (comment)?

    $ flatpak --user info com.protonvpn.www
    
    Proton VPN - High-speed Swiss VPN that safeguards your privacy.
    
              ID: com.protonvpn.www
             Ref: app/com.protonvpn.www/x86_64/test
            Arch: x86_64
          Branch: test
         Version: 1.12.0
         License: GPL-3.0
          Origin: www-origin
      Collection: org.flathub.Stable.Build12436
    Installation: user
       Installed: 45.9 MB
         Runtime: org.gnome.Platform/x86_64/43
             Sdk: org.gnome.Sdk/x86_64/43
    
          Commit: fb57eb69e601a9fd30a1336ba7b7ee67cd0c8fc5439623046a3c9acd060cefb2
         Subject: fix `can't find the executable for the dialog-like program` (72fe9223)
            Date: 2023-03-22 11:31:56 +0000

> It uses the binaries inside the sandbox. So it doesn't matter whether your host system has that binary.

Odd, I wonder why it's missing in the sandbox then.

The system does have it however:

    $ which clear
    /usr/bin/clear

A6GibKm:
`clear` Should be used from the runtime, the flatpak has to be as self contained as possible.

kevinlekiller:
> I think it's due to that the user on archlinux has a UID other than 1000. I'm going to remove the UID check.

Unfortunately the UID is 1000 for me, so it might not resolve the issue:

$ echo $UID && flatpak --user run --command=bash com.protonvpn.www -c 'echo $UID'; cat /etc/os-release | grep ^NAME
1000
1000
NAME="Arch Linux"

kevinlekiller:
> I'm sorry, but I don't think I'm able to debug on an issue that is not generally reproducible. Nor should it block this PR from being merged.

No worries, already using protonvpn-cli from the aur on that system so not a big deal for me.

com.protonvpn.www.yml
```diff
@@ -0,0 +1,541 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/flatpak/flatpak-builder/main/data/flatpak-manifest.schema.json
+
+id: com.protonvpn.www
+runtime: org.gnome.Platform
+runtime-version: '43'
```
hfiguiere, line 5: 44 is out.

storopoli:
Just a comment here. This is a great addition! Thanks for the effort, looking forward to having it as a flatpak.

A6GibKm:
App starts now but errors when trying to log in

kevinlekiller:
> GNOME, this is what I get
> 
> image
> 
> besides requiring the secret service (again, where is this invoked?) I don't think there is anything that would be problematic with KDE.

On my Arch / KDE system with the protonvpn-cli, it uses the KDE Wallet:

Screenshot_20230323_102931

If we can figure out what they're doing on that aur package maybe we can implement it here.

On my OpenSuse Tumbleweed / KDE system, I've made the kdewallet package taboo since it pulls it too many things I don't want.

On Fedora 37 Gnome, com.protonvpn.www works out of the box without installing any additional system packages.

A6GibKm:
There is nothing special about libsecret, any secret service provider should work, hell even keepassXC can act as one. If some KDE installs don't come with any, thats a tangential issue.

A6GibKm:
> Firstly, the stock protonvpn doesn't support the kde one

What does that mean? the secret service is a dbus api, its agnostic.

> Second, kde doesn't provide a complete implementation of secret service. Keepassxc seems does.

Thats tangential, users can still install one

PR #4009, opened 2023-03-25
Add com.spacestation14.Launcher

dylanmccall:

  - `--socket=x11` is required because Wayland is not yet properly supported. That work is ongoing in space-wizards/RobustToolbox#2468.
  - `--device=all` is required for MIDI device support, which is necessary for some parts of the game.
   - I am not an upstream contributor, but I contacted upstream developers in space-wizards/SS14.Launcher#66.
   - The domain is owned by the upstream project.
   - The appdata.xml file is unlikely to be maintained upstream, so I added it here.

### DIFF
--- a/com.spacestation14.Launcher.yaml
+++ b/com.spacestation14.Launcher.yaml
@@ -1,0 +1,76 @@
+---
+app-id: com.spacestation14.Launcher
+runtime: org.freedesktop.Platform
+runtime-version: '22.08'
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.dotnet7
+build-options:
+  append-path: "/usr/lib/sdk/dotnet7/bin"
+  append-ld-library-path: "/usr/lib/sdk/dotnet7/lib"
+  env:
+    PKG_CONFIG_PATH: "/app/lib/pkgconfig:/app/share/pkgconfig:/usr/lib/pkgconfig:/usr/share/pkgconfig:/usr/lib/sdk/dotnet7/lib/pkgconfig"
+
+command: SS14.Launcher
+
+finish-args:
+  - --device=dri
+  # Allow all-device access for MIDI device support
+  - --device=all
+  - --share=ipc
+  - --share=network
+  - --socket=pulseaudio
+  # TODO: Replace this with wayland and fallback-x11 once Wayland support
+  #       becomes available:
+  #       <https://github.com/AvaloniaUI/Avalonia/pull/8003>
+  #       <https://github.com/space-wizards/RobustToolbox/issues/2468>
+  - --socket=x11
+  # Allow access to the Music directory for MIDI files
+  - --filesystem=xdg-music
+  - --env=DOTNET_ROOT=/app/lib/dotnet
+
+modules:
+  - shared-modules/linux-audio/fluidsynth2.json
+
+  - name: space-station-14-launcher
+    buildsystem: simple
+    sources:
+      - sources/ss14-launcher-git.yaml
+      - sources/space-station-14-launcher-nuget-sources.json
+    build-commands:
+      - dotnet publish SS14.Launcher/SS14.Launcher.csproj /p:FullRelease=True -c Release --no-self-contained --source ./nuget-sources
+      - mkdir -p ${FLATPAK_DEST}/bin
+      - cp -r ${FLATPAK_BUILDER_BUILDDIR}/SS14.Launcher/bin/Release/net7.0/publish/* ${FLATPAK_DEST}/bin
+
+  - name: space-station-14-loader
+    buildsystem: simple
+    sources:
+      - sources/ss14-launcher-git.yaml
+      - sources/space-station-14-loader-nuget-sources.json
+    build-commands:
+      - dotnet publish SS14.Loader/SS14.Loader.csproj -c Release --no-self-contained --source ./nuget-sources
+      - mkdir -p ${FLATPAK_DEST}/bin/loader
+      - cp -r ${FLATPAK_BUILDER_BUILDDIR}/SS14.Loader/bin/Release/net7.0/publish/* ${FLATPAK_DEST}/bin/loader
+
+  - name: extra-data
+    buildsystem: simple
+    sources:
+      - type: dir
+        path: data
+        dest: data
+    build-commands:
+      # The upstream project does not include a desktop file. Instead, we
+      # write one from scratch.
+      - install -Dm644 data/com.spacestation14.Launcher.desktop -t /app/share/applications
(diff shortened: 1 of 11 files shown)


JakobDev:
bot, build com.spacestation14.Launcher

com.spacestation14.Launcher.yaml
```diff
@@ -0,0 +1,76 @@
...
+      # The upstream project does not include a desktop file. Instead, we
+      # write one from scratch.
+      - install -Dm644 data/com.spacestation14.Launcher.desktop -t /app/share/applications
+      # The upstream project does not include an appdata.xml file. Instead, we
+      # write one from scratch.
+      - install -Dm644 data/com.spacestation14.Launcher.appdata.xml -t /app/share/appdata
+      # The upstream project does not include an SVG icon file, but one is
+      # available as part of the Space Station 14 website:
+      # <https://github.com/space-wizards/website-content/blob/master/static/images/main/icon.svg>
+      - install -Dm644 -T data/icons/icon-scalable.svg /app/share/icons/hicolor/scalable/apps/com.spacestation14.Launcher.svg
```
hfiguiere, line 71: have these files submitted upstream?

com.spacestation14.Launcher.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - sources/space-station-14-loader-nuget-sources.json
+    build-commands:
+      - dotnet publish SS14.Loader/SS14.Loader.csproj -c Release --no-self-contained --source ./nuget-sources
+      - mkdir -p ${FLATPAK_DEST}/bin/loader
+      - cp -r ${FLATPAK_BUILDER_BUILDDIR}/SS14.Loader/bin/Release/net7.0/publish/* ${FLATPAK_DEST}/bin/loader
+
+  - name: extra-data
+    buildsystem: simple
+    sources:
+      - type: dir
```
hfiguiere, line 58: it's better to not use `type: dir`

com.spacestation14.Launcher.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - install -Dm644 data/com.spacestation14.Launcher.desktop -t /app/share/applications
+      # The upstream project does not include an appdata.xml file. Instead, we
+      # write one from scratch.
+      - install -Dm644 data/com.spacestation14.Launcher.appdata.xml -t /app/share/appdata
+      # The upstream project does not include an SVG icon file, but one is
+      # available as part of the Space Station 14 website:
+      # <https://github.com/space-wizards/website-content/blob/master/static/images/main/icon.svg>
+      - install -Dm644 -T data/icons/icon-scalable.svg /app/share/icons/hicolor/scalable/apps/com.spacestation14.Launcher.svg
+
+  - name: dotnet
```
hfiguiere, line 73: usually this is done first

sources/ss14-launcher-git.yaml
```diff
@@ -0,0 +1,3 @@
+type: git
+url: https://github.com/space-wizards/SS14.Launcher.git
```
hfiguiere, line 2: this has no reason to exist in a separate file.

com.spacestation14.Launcher.yaml
```diff
@@ -0,0 +1,76 @@
+---
+app-id: com.spacestation14.Launcher
```
hfiguiere, line 2: it is `com` or is it `io` ?

com.spacestation14.Launcher.yaml
dylanmccall: You got me to look at this again to double check why I had it like this. This does a fun thing: for the `dotnet` command to resolve, we have…

```yaml
build-options:
  append-path: "/usr/lib/sdk/dotnet7/bin"
```

(As suggested in the README for [org.freedesktop.Sdk.Extension.dotnet7](https://github.com/flathub/org.freedesktop.Sdk.Extension.dotnet7)).

… but after we run the install script, `dotnet` resolves to `/app/bin/dotnet`, which is the non-SDK version of dotnet (and we get an error complaining that the SDK isn't installed). Not what we want here, and that seems like a good bug report. I can fix it here by using `prepend-path`, or using the full path in the build-commands. I'll go with prepend-path for now :)

sources/ss14-launcher-git.yaml
dylanmccall: Yeah, I see what you mean. Makes it hard to figure out what version this is pointing to without inside knowledge :) I stuck together the two modules that were building from this source, and I changed the `update-ss14-version.py` script to simply read the current version from the flatpak manifest instead of trying to update it. (That's a good change anyway since it'll be easy to add an action to automate it, so I'm glad you pointed this out!).

com.spacestation14.Launcher.yaml
dylanmccall: It's … a bit odd? My understanding from https://github.com/space-wizards/SS14.Launcher/issues/66#issuecomment-1140082966 is .com will be canonical … eventually. The email domain lives under .com, for instance. Do you think it would be better to assume the current TLD will be around for a long time regardless, given it's pretty well stuck at this point?

com.spacestation14.Launcher.yaml
dylanmccall: I changed this so we're fetching the icon SVG from the official git repo instead of copying it here.

For the desktop file, I'll take a look at how this is being packaged elsewhere (I think the package for Arch includes its own version as well), and see if I can submit _something_ upstream. It has two main issues (the Exec line uses a relative path to the application, and there is no Icon for a similar reason) and the rest is so trivial that it seems silly to be patching it, but obviously if everyone who packages it has that same thought, it gets differently silly, so I'll do some poking there next week and send a patch :)

com.spacestation14.Launcher.yaml
dylanmccall: That makes the comments a bit nicer, too. Sure! :)

hfiguiere:
bot, build com.spacestation14.Launcher

barthalion:
/merge @PJB3005 

Hi @PJB3005, if you would consider the Flathub package official, and it's not too much to ask, can you log in at https://beta.flathub.org/ in a few hours and verify it as such? Thanks!

PJB3005:
@barthalion I accepted the repo invitation and I'll deal with the verification tonight.

com.spacestation14.Launcher.yaml
PJB3005: For future reference, the proper git repo that includes these the assets is over here: https://github.com/space-wizards/asset-dump/

com.spacestation14.Launcher.yaml
PJB3005: I should *really* move everything over some time huh...

PR #4018, opened 2023-03-27
pwall submit

ltiber:


org.pwall.Pwall.json
```diff
@@ -0,0 +1,39 @@
+{
+    "app-id": "org.pwall.Pwall",
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "40",
```
hfiguiere, line 4: GOME 40 has bee EOL for over 18month.

44 is the current one.

org.pwall.Pwall.json
```diff
@@ -0,0 +1,39 @@
...
+		    "install -D res/pwall/videoIcon.png /app/share/videoIcon.png",
+			"install -D res/pwall/clearbackup.sh /app/share/clearbackup.sh",
+		    "make -f makefile-flatpak",
+			"install -D pwall /app/bin/pwall",
+			"install -D flatpak/jhead /app/bin/jhead"
+            ],
+          "sources": [
+              {
+              "type": "git",
+              "url": "https://github.com/ltiber/pwall.git"
```
hfiguiere, line 27: a commit or tag is required,

org.pwall.Pwall.json
```diff
@@ -0,0 +1,39 @@
...
+             }    
+          ]
+      }
+  ],
+
+    "finish-args": [
```
hfiguiere, line 33: finish args is usually found above `modules`.

org.pwall.Pwall.json
```diff
@@ -0,0 +1,39 @@
...
+     "--socket=x11",
+     "--share=network",
+     "--filesystem=host",
```
hfiguiere, line 36: this is a very broad filesystem permission

org.pwall.Pwall.json
```diff
@@ -0,0 +1,39 @@
...
+              {
+              "type": "git",
+              "url": "https://github.com/ltiber/pwall.git"
+             }    
+          ]
+      }
+  ],
+
+    "finish-args": [
+     "--socket=x11",
```
hfiguiere, line 34: `--share=ipc` is necessary for x11

org.pwall.Pwall.json
```diff
@@ -0,0 +1,39 @@
...
+      "install -D flatpak/org.pwall.Pwall.appdata.xml /app/share/metainfo/org.pwall.Pwall.appdata.xml",
+			"install -D res/pwall/pwall.png /app/share/icons/hicolor/128x128/apps/org.pwall.Pwall.png",
+		    "install -D res/pwall/help.html /app/share/help.html",
+		    "install -D res/pwall/noimage.png /app/share/noimage.png",
+			"install -D res/pwall/novideo.png /app/share/novideo.png",
+		    "install -D res/pwall/videoIcon.png /app/share/videoIcon.png",
+			"install -D res/pwall/clearbackup.sh /app/share/clearbackup.sh",
+		    "make -f makefile-flatpak",
+			"install -D pwall /app/bin/pwall",
+			"install -D flatpak/jhead /app/bin/jhead"
```
hfiguiere, line 22: this looks like a binary. How does that work to support both x86_64 an aarch64?

org.pwall.Pwall.json
```diff
@@ -0,0 +1,39 @@
+{
+    "app-id": "org.pwall.Pwall",
```
hfiguiere, line 2: is `pwall.org` under the control of the developers? It doesn't look like it has anything.

org.pwall.Pwall.json
```diff
@@ -0,0 +1,39 @@
...
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "40",
+    "sdk": "org.gnome.Sdk",
+    "command": "pwall",
+    "modules": [
+      {
+          "name": "pwall",
+          "buildsystem": "simple",
+          "build-commands": [
+			"install -D flatpak/pwall-flatpak.desktop /app/share/applications/org.pwall.Pwall.desktop",
```
hfiguiere, line 12: why would this file be flatpak specific?

hfiguiere (changes requested):
Some change are required to build.

Also indentation is all over the place making it harder to read.

org.pwall.Pwall.json
ltiber: I've no domain registered for this software. The idea was to make a correct id in 3 part. For the moment the domain is available...

hfiguiere: Actually it's a requirement. You can use github if you don't have a domain.

https://docs.flatpak.org/en/latest/conventions.html?highlight=app-id#application-ids

org.pwall.Pwall.json
hfiguiere: you can use `destkop-file-edit` at install to change that.

org.pwall.Pwall.json
ltiber: I'll change the version.

org.pwall.Pwall.json
ltiber: the main function of pwall is to be a photos and videos library. I need to have access to this file wherever they are.

org.pwall.Pwall.json
ltiber: ok I'll add this property.

org.pwall.Pwall.json
ltiber: As you said, I can use github, I would think to use com.github.ltiber.Pwall 
But the documentation says it's better to use something like io.github.pwall.Pwall
but I'm not registered in github.io
any suggestion ?

ltiber:
I've just commit a new manifest. Let me know if it's ok for you.

io.github.pwall.Pwall.json
```diff
@@ -0,0 +1,54 @@
+{
+    "app-id":"io.github.pwall.Pwall",
+    "runtime":"org.gnome.Platform",
```
hfiguiere, line 3: do you really need the gnome platform?

ltiber: Yes my application uses gtk library.

hfiguiere: gtk 3 is in freedesktop-sdk.

io.github.pwall.Pwall.json
```diff
@@ -0,0 +1,54 @@
+{
+    "app-id":"io.github.pwall.Pwall",
```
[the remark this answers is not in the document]
hfiguiere: it should be `io.github.ltiber.Pwall` otherwise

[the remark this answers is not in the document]
ltiber: There was probably a misunderstanding.
I've suggested to use com.github.ltiber.Pwall as my url is https://github.com/ltiber/pwall 
but the doc said to use io pattern so I suggested to use io.github.pwall.Pwall 
But actually I think the app id should be io.github.ltiber.Pwall.

ltiber:
I've just commit a new manifest with the appId io.github.ltiber.Pwall and using only freedesktop. I didn't know it came with gtk3 support. Thanks for the review.

hfiguiere:
bot, build io.github.ltiber.Pwall

ltiber:
I've just installed and test pwall from the testing repository and everything works fine.

PR #4038, opened 2023-04-02
Add io.github.slgobinath.SafeEyes

otuva:

resolves slgobinath/SafeEyes#432


com.github.slgobinath.SafeEyes.metainfo.xml
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+    <id>com.github.slgobinath.SafeEyes</id>
```
hfiguiere, line 3: it should `io.github` as per the documentation.

com.github.slgobinath.SafeEyes.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+  - name: wlrctl # required for 'wayland do not disturb' plugin
+    buildsystem: meson
+    sources:
+      - type: archive
+        url: "https://git.sr.ht/~brocellous/wlrctl/archive/v0.2.1.tar.gz"
+        sha256: de8848de0a81749fe7ee7316bead0261bd978e26c2f899d02897609972a6e98c
+
+  - name: alsa-utils # required for 'audible alert' plugin
+    buildsystem: simple
```
hfiguiere, line 62: this should be `autotools` buildsystem.

com.github.slgobinath.SafeEyes.yaml
```diff
@@ -0,0 +1,99 @@
...
+  - /lib/pkgconfig
+  - /share/gtk-doc
+  - /share/man
+  - /lib/*.a
+  - /lib/*.la
+
+modules:
+  - shared-modules/libappindicator/libappindicator-gtk3-introspection-12.10.json # app indicator support
+  - shared-modules/libappindicator/libappindicator-gtk3-12.10.json
+  - libcanberra/libcanberra.json # don't clean /lib/gtk-3.0
```
hfiguiere, line 39: why do you not use the shared-module?

com.github.slgobinath.SafeEyes.yaml
```diff
@@ -0,0 +1,99 @@
...
+command: safeeyes
+
+finish-args:
+  - "--share=network"
+  - "--share=ipc"
+  - "--socket=x11"
+  - "--socket=pulseaudio"
+  - "--device=dri"
+  - "--system-talk-name=org.freedesktop.DBus" # for tray icon
+  - "--talk-name=org.freedesktop.Notifications" # for 'notification before breaks' plugin
```
hfiguiere, line 14: are you sure? libnotification 0.8.2 uses the portal and doesn't need any permission

hfiguiere (changes requested):
have the desktop appstream and icon submitted upstream?

com.github.slgobinath.SafeEyes.metainfo.xml
```diff
@@ -0,0 +1,55 @@
...
+
+    <name>Safe Eyes</name>
+    <developer_name>Gobinath</developer_name>
+    <summary>A Free and Open Source tool for Linux users to reduce and prevent repetitive strain
+        injury (RSI).</summary>
+
+    <metadata_license>CC-BY-4.0</metadata_license>
```
hfiguiere, line 10: `CC0-1.0` is prefered.

com.github.slgobinath.SafeEyes.yaml
otuva: autotools without build-commands right?
Without prefix, build will fail 

```
/usr/bin/install: cannot create regular file '/lib/udev/rules.d/90-alsa-restore.rules': Read-only file system
```

com.github.slgobinath.SafeEyes.yaml
```diff
@@ -0,0 +1,99 @@
...
+runtime-version: "22.08"
+command: safeeyes
+
+finish-args:
+  - "--share=network"
+  - "--share=ipc"
+  - "--socket=x11"
+  - "--socket=pulseaudio"
+  - "--device=dri"
+  - "--system-talk-name=org.freedesktop.DBus" # for tray icon
```
hfiguiere, line 13: I don't think this VERY BROAD permission is neceary.

com.github.slgobinath.SafeEyes.yaml
otuva: After this [commit](https://github.com/flathub/shared-modules/commit/096ab6b0be271f07a42a647b5ffc83808a160e59), installation deletes `/lib/gtk-3.0` too so instead of forking it I'm just using the same without the gtk-3.0 line

Without  `/lib/gtk-3.0`, app will give a warning `Failed to load module "canberra-gtk-module"` probably it's not that important but I didn't check the whole codebase and upstream says it's a requirement

com.github.slgobinath.SafeEyes.yaml
otuva: Tried a couple of times and it didn't work without it. Let me try again though

otuva:
> have the desktop appstream and icon submitted upstream?

Icon is from the upstream, but I don't see why the XML file should be submitted. I can though if it's necessary

hfiguiere:
> > have the desktop appstream and icon submitted upstream?
> 
> Icon is from the upstream, but I don't see why the XML file should be submitted. I can though if it's necessary

the the icon should NOT be in this repository if it is upstream. How about the .desktop?

Yes the appstream file too should be submitted upstream. It is not flatpak specific either.

com.github.slgobinath.SafeEyes.yaml
hfiguiere: this is how all the packages that use it use it.

hfiguiere:
the files need to be renamed too.

com.github.slgobinath.SafeEyes.yaml
hfiguiere: `config-opts`

otuva:
> the files need to be renamed too.

Should I open new PR with io one to resolve discrepancy between file reverse dns and repo? I saw a bunch of packages following com.github so that's why I did that

hfiguiere:
just rename the files.

hfiguiere:
>  I saw a bunch of packages following com.github so that's why I did that

This rule was not enforced before. But it's in the documentation.

com.github.slgobinath.SafeEyes.yaml
otuva: Is it ok to use `org.freedesktop.StatusNotifierItem.Menu` ?

com.github.slgobinath.SafeEyes.yaml
```diff
@@ -0,0 +1,99 @@
+id: com.github.slgobinath.SafeEyes
```
hfiguiere, line 1: this also doesn't match. The app id has to be changed everywhere.

otuva: Yeah, I'm going to push the commit that changes them. But currently I couldn't figure out one thing. 

So, the app, after setup, installs icons and desktop file under `/app/lib/python3.10/site-packages/safeeyes/platform/` I couldn't find any docs or different packages related to this. How do you think I should proceed? `install`, mv, cp or symlinks don't feel 'clean'

io.github.slgobinath.SafeEyes.metainfo.xml
```diff
@@ -0,0 +1,55 @@
...
+        </p>
+        <p>
+            Remind you to take breaks with exercises to reduce RSI, Disable keyboard during breaks,
+            Notification before and after breaks, Smart pause if system is idle, Multi-screen
+            support, Customizable user interface, RPC API to control externally, Command-line
+            arguments to control the running instance, Customizable using plug-ins
+        </p>
+    </description>
+
+    <launchable type="desktop-id">com.github.slgobinath.SafeEyes.desktop</launchable>
```
hfiguiere, line 34: this doesn't match.

io.github.slgobinath.SafeEyes.yaml
```diff
@@ -0,0 +1,95 @@
...
+          project-id: 13149
+          type: anitya
+          url-template: https://download.gnome.org/sources/libnotify/$major.$minor/libnotify-$version.tar.xz
+
+  - name: safeeyes # shortcut
+    buildsystem: simple
+    build-commands:
+      - install -Dm644 /app/lib/python3.10/site-packages/safeeyes/platform/icons/hicolor/128x128/apps/safeeyes.png /app/share/icons/hicolor/128x128/apps/io.github.slgobinath.SafeEyes.png
+      - install -Dm644 /app/lib/python3.10/site-packages/safeeyes/platform/safeeyes.desktop /app/share/applications/io.github.slgobinath.SafeEyes.desktop
+      - desktop-file-edit --rebuild-mime-info-cache --set-icon=io.github.slgobinath.SafeEyes --set-key=Exec --set-value=safeeyes /app/share/applications/io.github.slgobinath.SafeEyes.desktop
```
hfiguiere, line 91:
```suggestion
      - desktop-file-edit --set-icon=io.github.slgobinath.SafeEyes --set-key=Exec --set-value=safeeyes /app/share/applications/io.github.slgobinath.SafeEyes.desktop
```

io.github.slgobinath.SafeEyes.yaml
```diff
@@ -0,0 +1,95 @@
...
+  - pypi-dependencies.json
+
+  - name: dbus-python # dbus - for tray
+    buildsystem: autotools
+    sources:
+      - type: archive
+        url: "https://dbus.freedesktop.org/releases/dbus-python/dbus-python-1.3.2.tar.gz"
+        sha256: ad67819308618b5069537be237f8e68ca1c7fcc95ee4a121fe6845b1418248f8
+
+  - name: wlrctl # required for 'wayland do not disturb' plugin
```
hfiguiere, line 53: why do you need this if you don't have the wayland socket.

io.github.slgobinath.SafeEyes.yaml
```diff
@@ -0,0 +1,95 @@
...
+    buildsystem: meson
+    sources:
+      - type: archive
+        url: "https://git.sr.ht/~brocellous/wlrctl/archive/v0.2.1.tar.gz"
+        sha256: de8848de0a81749fe7ee7316bead0261bd978e26c2f899d02897609972a6e98c
+
+  - name: alsa-utils # required for 'audible alert' plugin
+    buildsystem: autotools
+    config-opts:
+      - "--prefix=/app"
```
hfiguiere, line 63:
```suggestion
```

io.github.slgobinath.SafeEyes.yaml
otuva: Even though the app only runs through x11, it can still detect wayland and `wlrctl` is required to detect full screen

hfiguiere: how does it detect wayland?

otuva: with [this function](https://github.com/slgobinath/SafeEyes/blob/cae19b0f28f14036d1a2a4c840bf3e6f68c02b6f/safeeyes/utility.py#L299) 

without wlrctl you can see [the warning](https://github.com/slgobinath/SafeEyes/blob/cae19b0f28f14036d1a2a4c840bf3e6f68c02b6f/safeeyes/plugins/donotdisturb/plugin.py#L52)

hfiguiere: do you test is inside the flatpak sandbox? Because this detection is broken.

hfiguiere:
bot, build io.github.slgobinath.SafeEyes

io.github.slgobinath.SafeEyes.yaml
otuva: Yeah, you're right. I'm going to keep to fiddle with it. It probably acts like that because of missing wayland socket

flathub.json
```diff
@@ -0,0 +1,5 @@
+{
+    "only-arches": [
+        "x86_64"
```
hfiguiere, line 3: you could just fix the broken code instead.

otuva: Code should be 

```
bool is_ascii(const char str[])
{
    for (int i = 0; str[i] != '\0'; i++) {
        if (str[i] > 127) {
            return false;
        }
    }
    return true;
}
```

right? How do I patch that though, can you supply patch even if it's meson

otuva: Apparently author of the package marked it as x86_64 arch only

[aur pkgbuild](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=wlrctl)

hfiguiere: that will not build on x86_64.

hfiguiere: > Apparently author of the package marked it as x86_64 arch only
> 
> [aur pkgbuild](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=wlrctl)

then don't build this on aarch64.

hfiguiere: and by this I mean wlrctl.

otuva:
Hi, @hfiguiere, sorry to bother you mate, but what is the mergeability status of this?

otuva:
> Shouldn't it create some default config on start?

Yeah, it should, it stems from [this part](https://github.com/slgobinath/SafeEyes/blob/master/safeeyes/utility.py#L377) and author is aware of the problem. I initially wanted to fix it in manifest, but evidently in docs there's no way to create config files for the user.

I added to the README for the app's repo, but probably no one is going to read that.  

I'm going to submit a patch to the upstream, until then if you think it's a big problem I can write a wrapper for initial launch.

PR #4052, opened 2023-04-06
add net.mullvad.MullvadBrowser

proletarius101:

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://gitlab.torproject.org/tpo/applications/mullvad-browser/-/issues/162

### DIFF
--- a/net.mullvad.MullvadBrowser.yml
+++ b/net.mullvad.MullvadBrowser.yml
@@ -17,8 +17,10 @@ finish-args:
   - --device=all
   - --share=ipc
   - --share=network
+  # Use `--socket=wayland` + `--socket=x11` like firefox does. `--socket=fallback-x11`` should be removed when app prefers x11.
+  # https://github.com/flathub/flathub/pull/4052#discussion_r1164379207
   - --socket=wayland
-  - --socket=fallback-x11
+  - --socket=x11
   - --socket=pulseaudio
   - --filesystem=xdg-run/pipewire-0
   # profiles
--- a/net.mullvad.MullvadBrowser.yml
+++ b/net.mullvad.MullvadBrowser.yml
@@ -68,6 +68,7 @@ modules:
       - desktop-file-edit "${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop"
         --set-key "Exec" --set-value "${FLATPAK_DEST}/bin/mullvadbrowser --class \"Mullvad Browser\" --name \"Mullvad Browser\" %u"
         --set-key "Icon" --set-value "${FLATPAK_ID}"
+        --set-key "Name" --set-value "Mullvad Browser"
     sources:
       - type: archive
         url: https://cdn.mullvad.net/browser/12.0.4/mullvad-browser-linux64-12.0.4_ALL.tar.xz
--- a/net.mullvad.MullvadBrowser.yml
+++ b/net.mullvad.MullvadBrowser.yml
@@ -63,12 +63,12 @@ modules:
         --set-key "Icon" --set-value "${FLATPAK_ID}"
     sources:
       - type: archive
-        url: https://github.com/mullvad/mullvad-browser/releases/download/mullvad-browser-102.9.0esr-12.0-2-build1/mullvad-browser-linux64-12.0.4_ALL.tar.xz
+        url: https://cdn.mullvad.net/browser/12.0.4/mullvad-browser-linux64-12.0.4_ALL.tar.xz
         sha256: ab875328d42472a691c22176e6254e412bf054b037b494654380f30b7b6e1b90
         x-checker-data:
           type: rotating-url
-          url: https://download.mozilla.org/?product=firefox-nightly-latest-ssl&os=linux64&lang=en-US
-          pattern: https://download-installer.cdn.mozilla.net/pub/firefox/nightly/latest-mozilla-central/firefox-([0-9.]+)a([0-9.]+).en-US.linux-x86_64.tar.bz2
+          url: https://mullvad.net/en/download/browser/linux64/latest
+          pattern: https://cdn.mullvad.net/browser/([0-9.]+)/mullvad-browser-linux64-([0-9.]+)_ALL.tar.xz
 
       - type: file
         url: https://gitlab.torproject.org/tpo/applications/mullvad-browser/-/raw/e712cb197b83778cbda85e96525dd56ada25945e/browser/branding/mb-release/firefox.svg
@@ -83,7 +83,7 @@ modules:
 
       - type: file
         path: default-preferences.js
-      
+
       - type: file
         path: net.mullvad.MullvadBrowser.metainfo.xml
 
--- a/net.mullvad.MullvadBrowser.yml
+++ b/net.mullvad.MullvadBrowser.yml
@@ -13,7 +13,8 @@ add-extensions:
     version: *sdk-version
 
 finish-args:
-  - --device=dri
+  # For webcam
+  - --device=all
   - --share=ipc
   - --share=network
   - --socket=wayland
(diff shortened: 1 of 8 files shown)


pospeselr:
I'm not quite sure how the flathub ecosystem works, but you can get the current browser version and donwload url from the downloads.json file here:

- https://cdn.mullvad.net/browser/update_responses/update_1/release/downloads.json

This file will stay up to date with each release, though if there is a watershed update (eg a version of the browser all users must update to directly, before updating further) then this may move to update_2/release/downloads.json (etc) in the future.

pospeselr:
@proletarius101 So iirc we have a way to specify this at runtime to handle these types of system-installs. The difficulty though will be the updater, since it expects the browser to be living somewhere the user has write access. This is a holdover from Tor Browser but it is on the roadmap to handle proper packages more elegantly.

PieroV:
@ proletarius101 you have two choices:

1. you specify `-profile` (see [Firefox documentation](https://wiki.mozilla.org/Firefox/CommandLineOptions#-profile_.22profile_path.22))
2. you can add a `system-install` file next to `mullvadbrowser`. It will fallback to using usual Firefox ways to decide where to store the profile directory (IIRC, `~/.mullvad/mullvadbrowser`; in any case it shouldn't conflict with Firefox, as we customize a few variables that influence the profile directory)

2 should be seen as a sort of feature preview, we have not tested it much, yet :slightly_smiling_face:.

PieroV:
> Thanks for the hint. 2 works well.

Nice :smile:. Firefox has a few `if`s for when it's a flatpak.
It might be necessary to double check them and round some rough edges.

ruihildt:
Tried this build and the system shortcut is called "Mullvad Browser Setup".

ruihildt:
I tried WebRTC with this link, and it seems to fail. Is it the case for you also?
Is there some permissions to set regarding audio/video maybe?

https://webrtc.github.io/test-pages/src/audio-and-video/

net.mullvad.MullvadBrowser.yml
```diff
@@ -0,0 +1,111 @@
...
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.gtk.vfs.*
+  - --system-talk-name=org.freedesktop.RealtimeKit1
+  - --system-talk-name=org.freedesktop.login1
+  - --system-talk-name=org.freedesktop.UPower
+  # Wi-Fi scanner
+  - --system-talk-name=org.freedesktop.NetworkManager
+  - --filesystem=xdg-download:create
+  # fix wayland
+  - --env=MOZ_ENABLE_WAYLAND=1
```
Erick555, line 43: Why this is set? FF default is still to use X11.

net.mullvad.MullvadBrowser.yml
```diff
@@ -0,0 +1,108 @@
...
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.gtk.vfs.*
+  - --system-talk-name=org.freedesktop.RealtimeKit1
+  - --system-talk-name=org.freedesktop.login1
+  - --system-talk-name=org.freedesktop.UPower
+  # Wi-Fi scanner
+  - --system-talk-name=org.freedesktop.NetworkManager
+  - --filesystem=xdg-download:create
+  # the author doesn't control the domain mozilla.org
+  - --own-name=org.mozilla.mullvadbrowser.*
```
hfiguiere, line 43: instead they should control their own dbus, ie change this upstream.

net.mullvad.MullvadBrowser.yml
```diff
@@ -0,0 +1,108 @@
...
+    no-autodownload: true
+    version: *sdk-version
+
+finish-args:
+  # For webcam
+  - --device=all
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --socket=fallback-x11
```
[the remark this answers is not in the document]
Erick555: Use `--socket=wayland` + `--socket=x11` like firefox does.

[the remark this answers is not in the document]
Erick555: This is only warning in linter, it doesn't need exception. Having --socket=wayland + --socket=x11 won't alter app behavior seen outside flatpak, anything else does.

net.mullvad.MullvadBrowser.yml
```diff
@@ -0,0 +1,108 @@
...
+  - --persist=.mullvad
+  # Smartcard access - Requires flatpak >= 1.3.2
+  - --socket=pcsc
+  - --socket=cups
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=org.freedesktop.secrets
+  - --talk-name=org.freedesktop.FileManager1
+  - --talk-name=org.gnome.SessionManager
+  - --talk-name=org.freedesktop.ScreenSaver
+  - --talk-name=org.freedesktop.Notifications
```
Erick555, line 34: This isn't needed with recent libnotify (I know ff still has it)

Thorin-Oakenpants:
@pospeselr  @fxbrit is hopefully onto it: will compare TZP, TZP 2.0 (not live), and creepyjs

Thorin-Oakenpants:
fonts & glyphs & system fonts differ. Using TZP 2.0 (which correctly uses the same lists for TB and MB), we test 154 fonts, 134 are whitelisted, and 134 are bundled (these are the lists at the top of the section)

for starters the flatpack leaks the following
- see https://github.com/arkenfox/TZP/blob/4c3947bddd1bac44c16bd31a7ea8e1d3a6d047b3/js/fonts.js#L76-L89
- @PieroV , so glad I built this in - send cake
```
{
  "unexpected": [
    "Arial",
    "Cantarell",
    "Courier New",
    "Droid Sans",
    "Liberation Mono",
    "Noto Color Emoji",
    "Noto Sans",
    "STIX"
  ]
}
```

glyphs differ because, my instinct says, is that some of those glyphs are mapping to the "unexpected" fonts

The other one seems to be some difference in system fonts, and again I think this is because the wrong font is being used, or it could be the sizes - @fabrizo would need to click the details and do a diff

I won't post the pics from @fabrizio - fix the leaky system fonts and then we can retest - @pospeselr

Erick555:
Does it mean Mullvad browser is somewhat leaking system fonts (perhaps Tor browser as well)?

Thorin-Oakenpants:
also RLBox doesn't seem to be implemented - there is no `--without-wasm-sandboxed-libraries` in `about:buildconfig`

Thorin-Oakenpants:
@Erick555 TB is fine - that's what the TZP 2.0 test detects - it outputs notations in red/green with checks or X's. Obviously detecting extra fonts affects the FP, but they were added to detect the patch breaking

PS: I fixed the red notation to say MB if isMullvad = true :)
TZP2

fxbrit:
@Thorin-Oakenpants RLBox should be in TB soon-ish tho --> https://forum.torproject.net/t/tor-project-pierovs-monthly-status-report-march-2023/7145

sorry everyone for the OT.

Erick555:
Yeah, lost of this settings is relevant. I think shipping simplified script in flatpak until mullvad makes their own flatpak-friendly is good idea.

Thorin-Oakenpants:
@proletarius101 
- https://arkenfox.github.io/TZP2/tzp.html#fonts
- this is just a temp repo and a WIP
- no need to compare anything (fxbrit can do that later) - just check that `font names` = 134 and has a green tick
   - if not then click the red notation and the leaked or missing fonts are listed in the console log

PieroV:
> edit: nvm I was confused. It seems TB stable is the one with RLBox - @PieroV

TB stable does not, IIRC. But MB is based on the main/alpha branch of the build repository, so it should have RLBox indeed.

For the fonts, I think it might be time to customize `font.system.whitelist` also for Linux, at least as a defense-in-depth.

ruihildt:
@barthalion 

I have been invited to join the Flathub repo, but we (Mullvad Browser team) haven't yet decided how we deal with third party packaging efforts. We're looking into it, and I'll get back to you when we have reached a decision.

Erick555:
@proletarius101

> Seems to be fixed now

Only partially I think. The first hash is still different in mullvad flatpak than non-flatpak and TB. Also the` [system] fonts` is hash different and shown as a failure in test.

@Thorin-Oakenpants is it supposed to look like this?

Thorin-Oakenpants:
system fonts hash applies to the next release - the TZP2 WIP demo is early for you guys

Thorin-Oakenpants:
It is best to compare on the same machine since we're comparing builds, because TZP exploits more than just lists of enumerating fonts (it's trying to get entropy between machines) - for example the fonts sizes are measured with high precision and could exploit aliasing/subpixels. Obviously the same hash for a list of 134 expected fonts is different. For your amusement, the entire TZP hash will change if you have the console (ctrl-k) open vs console closed - it will also change if you have loaded specific tabs in the web console like storage - you can see it if you rerun the `misc` section

> Yes, they match on my machine.

Cool

PR #4069, opened 2023-04-13
At.vintagestory.vintage story.unstable

agithubuserlol:

- [X ] I have read the [App Requirements][reqs] and [App Maintenance][maint] pages.

Additional notes:
Confirmed with VintageStory developer Tyron that Flathub packages ("all Linux packages") are community maintained. jascrain maintains the stable build Flathub (https://github.com/flathub/at.vintagestory.VintageStory). This package is just a clone of their work, with the underlying packages and descriptions altered to point at the unstable branch so that Steam Deck users can more easily access Vintage Story release candidates.

This is my first attempt at FlatHub deployment. I hope I got everything right.


at.vintagestory.VintageStory.yaml
```diff
@@ -0,0 +1,62 @@
+app-id: at.vintagestory.VintageStory.unstable
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
```
hfiguiere, line 3: 22.08 is the latest.

at.vintagestory.VintageStory.metainfo.xml
```diff
@@ -0,0 +1,57 @@
...
+            inspired by lovecraftian horror themes. Find yourself in a ruined
+            world reclaimed by nature and permeated by unnerving temporal
+            disturbances. Relive the advent of human civilization, or take your
+            own path.</p>
+    </description>
+    <developer_name>Anego Studios</developer_name>
+    <categories>
+        <category>Game</category>
+    </categories>
+    <content_rating type="oars-1.0">
```
hfiguiere, line 22: we use oars 1.1

at.vintagestory.VintageStory.metainfo.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+    <id>at.vintagestory.VintageStory.unstable</id>
```
hfiguiere, line 3: this should match the app-id. It doesn't

agithubuserlol: I just double checked and it does appear to match the app-id. Is it possible for you to look again? Apologies if I am missing the reference somewhere.

I've compared 
https://github.com/flathub/flathub/blob/d117c67117bfe14ea0cd28b44aef681a13bc9714/at.vintagestory.VintageStory.metainfo.xml#L3

to
https://github.com/flathub/flathub/blob/d117c67117bfe14ea0cd28b44aef681a13bc9714/at.vintagestory.VintageStory.metainfo.xml#L3

hfiguiere: also why the `unstable`?

agithubuserlol: I'll resolve the filenames now. Thank you.

The name is unstable because it is pointed at the unstable branch for VintageStory. The stable version is already on Flathub (see link in the PR template above).

Would you prefer I follow a different naming convention? Happy to alter however desired.

hfiguiere: just use the beta branch of the existing package to be on flathub-beta.

in general flathub is not for "unstable" software.

PR #4092, opened 2023-04-22
Add org.sabnzbd.sabnzbd

francoism90:

- [ ] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **[Link](https://github.com/sabnzbd/sabnzbd/issues/2469):**


org.sabnzbd.SABnzbd.yaml
```diff
@@ -0,0 +1,114 @@
...
+    make-install-args:
+      - DESTDIR=$(FLATPAK_DEST)
+    sources:
+      - type: archive
+        url: https://www.rarlab.com/rar/unrarsrc-6.2.5.tar.gz
+        sha256: 9a3974410d1d340e3998dd2a6f98faefbe838cad556266e714adfb0e8cf9377c
+  - name: par2cmdline
+    buildsystem: simple
+    build-commands:
+      - ./configure --prefix=/app
```
hfiguiere, line 96: this should be `buildsystem: autotools` instead.

org.sabnzbd.SABnzbd.yaml
```diff
@@ -0,0 +1,114 @@
...
+        sha256: 7b2fcc19b54c7080939fc2cbaca33ae40ac33759a526292933c94a85ba850d11
+  - name: sabnzbdplus
+    buildsystem: simple
+    build-commands:
+      - cp -a * /app/bin/
+      - install -Dm644 ${PWD}/linux/org.sabnzbd.sabnzbd.appdata.xml /app/share/appdata/$FLATPAK_ID.xml
+      - install -Dm644 ${PWD}/icons/logo-arrow.svg /app/share/icons/hicolor/scalable/apps/$FLATPAK_ID.svg
+      - install -Dm644 ${PWD}/linux/sabnzbd.desktop /app/share/applications/$FLATPAK_ID.desktop
+      - desktop-file-edit --set-icon=${FLATPAK_ID} /app/share/applications/${FLATPAK_ID}.desktop
+      - sed -i 's|Exec=/opt/sabnzbd/SABnzbd.py --browser 1 %F|Exec=SABnzbd.py|g' /app/share/applications/${FLATPAK_ID}.desktop
```
hfiguiere, line 110: you should use desktop-file-edit here as well.

org.sabnzbd.SABnzbd.yaml
```diff
@@ -0,0 +1,114 @@
...
+      - make install
+    sources:
+      - type: archive
+        url: https://github.com/Parchive/par2cmdline/releases/download/v0.8.1/par2cmdline-0.8.1.tar.gz
+        sha256: 7b2fcc19b54c7080939fc2cbaca33ae40ac33759a526292933c94a85ba850d11
+  - name: sabnzbdplus
+    buildsystem: simple
+    build-commands:
+      - cp -a * /app/bin/
+      - install -Dm644 ${PWD}/linux/org.sabnzbd.sabnzbd.appdata.xml /app/share/appdata/$FLATPAK_ID.xml
```
hfiguiere, line 106: this problem is that the app-id in upstream is different than the one used here. You should use whatever upstream uses. *it's case sensitive*

hfiguiere: also upstream file misses `releases` that is mandatory.

org.sabnzbd.SABnzbd.yaml
```diff
@@ -0,0 +1,114 @@
...
+      - type: archive
+        url: https://github.com/p7zip-project/p7zip/archive/v17.05/p7zip-v17.05.tar.gz
+        sha256: 9473e324de6a87d89cb7ff65b0fec4ae3f147f03ffc138189c336a4650d74804
+      - type: shell
+        commands:
+          - sed -i 's|/usr/local|/app|g' makefile.common
+  - name: unrar
+    no-autogen: true
+    build-options:
+      strip: true
```
hfiguiere, line 86: why?

```suggestion
```

francoism90:
@hfiguiere Thanks for the review, I'll resolve the issues as best as I can, kinda new to Flatpak. :)

How to handle things like `README.md`? Is it Ok to include 'build from source' instructions?
See https://github.com/francoism90/flatpak-org.sabnzbd.SABnzbd/blob/main/README.md

Thanks

org.sabnzbd.SABnzbd.yaml
francoism90: @hfiguiere I've contacted upstream, hopefully it can be resolved soon.

francoism90:
https://github.com/sabnzbd/sabnzbd/issues/2469#issuecomment-1519026019

Could you guys look at this as well? :)
It would indeed be great to generate this and/or by using a GitHub hook.

Maybe this is already possible, I don't know sorry.

org.sabnzbd.SABnzbd.yaml
```diff
@@ -0,0 +1,112 @@
+app-id: org.sabnzbd.SABnzbd
```
hfiguiere, line 1: as I mentionned previously upstream as an all lower case app-id, and you should use that here.

This is not an option.

org.sabnzbd.SABnzbd.yaml
hfiguiere: you create a patch to fix is without them.

francoism90: @hfiguiere Could you please give me details? Is it possible to link to something?

I could add the `releases` tag, no problem, just wanted to see if upstream has some ideas.

org.sabnzbd.SABnzbd.yaml
francoism90: Sorry, now I understand, fixed. :)

org.sabnzbd.sabnzbd.yaml
```diff
@@ -0,0 +1,112 @@
...
+      - type: archive
+        url: https://github.com/Parchive/par2cmdline/releases/download/v0.8.1/par2cmdline-0.8.1.tar.gz
+        sha256: 7b2fcc19b54c7080939fc2cbaca33ae40ac33759a526292933c94a85ba850d11
+  - name: sabnzbdplus
+    buildsystem: simple
+    build-commands:
+      - cp -a * /app/bin/
+      - install -Dm644 ${PWD}/linux/org.sabnzbd.sabnzbd.appdata.xml /app/share/appdata/org.sabnzbd.sabnzbd.appdata.xml
+      - install -Dm644 ${PWD}/icons/logo-arrow.svg /app/share/icons/hicolor/scalable/apps/$FLATPAK_ID.svg
+      - install -Dm644 ${PWD}/linux/sabnzbd.desktop /app/share/applications/$FLATPAK_ID.desktop
```
hfiguiere, line 106:
```suggestion
      - install -Dm644 linux/org.sabnzbd.sabnzbd.appdata.xml /app/share/appdata/org.sabnzbd.sabnzbd.appdata.xml
      - install -Dm644 icons/logo-arrow.svg /app/share/icons/hicolor/scalable/apps/$FLATPAK_ID.svg
      - install -Dm644 linux/sabnzbd.desktop /app/share/applications/$FLATPAK_ID.desktop
```

org.sabnzbd.sabnzbd.yaml
```diff
@@ -0,0 +1,112 @@
...
+      - DESTDIR=$(FLATPAK_DEST)
+    sources:
+      - type: archive
+        url: https://www.rarlab.com/rar/unrarsrc-6.2.5.tar.gz
+        sha256: 9a3974410d1d340e3998dd2a6f98faefbe838cad556266e714adfb0e8cf9377c
+  - name: par2cmdline
+    buildsystem: autotools
+    build-commands:
+      - ./configure --prefix=/app
+      - make install
```
hfiguiere, line 95:
```suggestion
```

francoism90:
bot, build org.sabnzbd.sabnzbd

francoism90:
bot, build org.sabnzbd.sabnzbd

francoism90:
bot, build org.sabnzbd.sabnzbd

francoism90:
@hfiguiere Upstream merged the version tag, however the build still fails.

Is the icon required? Also is it possible to validate the appstream.xml before I contact upstream?
I do not want to spam them with requests.

Thanks

hfiguiere:
The appstream file has `  <launchable type="desktop-id">sabnzbd.desktop</launchable>`.

The best way to go about is to NOT rename the desktop file on install and use `rename-desktop-file` in the manifest.

francoism90:
bot, build org.sabnzbd.sabnzbd

francoism90:
@hfiguiere Could you please give more details? :)

I've tried adding `rename-desktop-file`, but it results in build failure.

hfiguiere:
> @hfiguiere Could you please give more details? :)
> 
> I've tried adding `rename-desktop-file`, but it results in build failure.

I wrote:

> The best way to go about is to NOT rename the desktop file on install and use rename-desktop-file in the manifest.

Not to do half of it.

Also for the icon, there is none with the name you give. So it's not surprising it all fails.

francoism90:
@hfiguiere I'm really sorry, but I'm not a Linux developer. So some things are a bit difficult to understand.

Could I also use `desktop-file-edit` and set the  `<launchable type="desktop-id">sabnzbd.desktop</launchable>`?

barthalion:
No, desktop-file-edit is only for desktop files. You will need to patch the appdata file.

hfiguiere:
or use `rename-desktop-file` which is way easier. As I previously said the attempt failed because in that case the desktop file should NOT be renamed on install.

francoism90:
bot, build org.sabnzbd.sabnzbd

francoism90:
@barthalion @hfiguiere Is this what I need to do?

I still don't understand, I need to copy the desktop file, as I need to alter the executable path.

hfiguiere:
> @barthalion @hfiguiere Is this what I need to do?
> 
> I still don't understand, I need to copy the desktop file, as I need to alter the executable path.

there is copy, and there is copy with renaming. Nothing Linux specific.

org.sabnzbd.sabnzbd.yaml
```diff
@@ -0,0 +1,111 @@
...
+        url: https://github.com/Parchive/par2cmdline/releases/download/v0.8.1/par2cmdline-0.8.1.tar.gz
+        sha256: 7b2fcc19b54c7080939fc2cbaca33ae40ac33759a526292933c94a85ba850d11
+  - name: sabnzbdplus
+    buildsystem: simple
+    build-commands:
+      - cp -a * /app/bin/
+      - install -Dm644 linux/org.sabnzbd.sabnzbd.appdata.xml /app/share/appdata/org.sabnzbd.sabnzbd.appdata.xml
+      - install -Dm644 icons/logo-arrow.svg /app/share/icons/hicolor/scalable/apps/sabnzbd.svg
+      - install -Dm644 linux/sabnzbd.desktop /app/share/applications/sabnzbd.desktop
+      - desktop-file-edit --set-icon=sabnzbd /app/share/applications/sabnzbd.desktop
```
hfiguiere, line 106: you don't need this with rename-icon

```suggestion
```

francoism90:
bot, build org.sabnzbd.sabnzbd

francoism90:
It seems to build successfully!

Thanks for the help, what happens now? :)

barthalion:
I'm curious, what's the primary use case for something typically running on a server as a flatpak? Are users expected to use a systemd unit to start it? There's also no tray icon to indicate it's actually running in the background.

francoism90:
You don't have to run this on a server. It can be run on a desktop as well. I think it will be 50/50, as this program also sometimes works together with other apps.

It's the same with BitTorrent clients, they can also be run on a server, but they don't have to.

francoism90:
I don't understand why this hasn't been merged yet. I'm seeing packages being merged left-right, but this package isn't?
Furthermore, I had to work hard with upstream to get it to work, which isn't fair to them either.

So what's missing? Is it ready? Any status update would be appreciated.

PR #4096, opened 2023-04-23
Updated pull request for trustedqsl

k1mu:

### DIFF
--- a/org.arrl.trustedqsl.json
+++ b/org.arrl.trustedqsl.json
@@ -145,11 +145,7 @@
 	  	}
 	    ],
 	    "cleanup" : [
-		"/include/adif.h",
-		"/include/cabrillo.h",
-		"/include/tqslconvert.h",
-		"/include/tqslerrno.h",
-		"/include/tqsllib.h"
+		"/include"
 	    ]
 	}
     ]
--- a/org.arrl.trustedqsl.json
+++ b/org.arrl.trustedqsl.json
@@ -1,0 +1,158 @@
+{
+    "app-id" : "org.arrl.trustedqsl",
+    "runtime" : "org.freedesktop.Platform",
+    "runtime-version" : "22.08",
+    "sdk" : "org.freedesktop.Sdk",
+    "sdk-extensions" : [ "org.freedesktop.Sdk.Locale" ],
+    "command" : "tqsl",
+    "finish-args" : [
+	"--share=network",
+	"--socket=x11",
+	"--filesystem=home"
+     ],
+    "modules" : [
+        {
+            "name" : "openssl",
+            "buildsystem" : "simple",
+            "build-commands" : [
+                "./Configure --prefix=/app && make depend && make build_libs && make install"
+            ],
+            "sources" : [
+                {
+                    "type" : "archive",
+                    "url" : "https://www.openssl.org/source/openssl-3.1.0.tar.gz",
+                    "sha256" : "aaa925ad9828745c4cad9d9efeb273deca820f2cdcf2c3ac7d7c1212b7c497b4"
+                }
+            ],
+	    "cleanup" : [
+		"/bin/c_rehash",
+		"/bin/openssl",
+		"/lib64/libcrypto.a",
+		"/lib64/libssl.a",
+		"/share/doc/openssl/*",
+		"/share/doc/openssl",
+		"/share/man/man1/*",
+		"/share/man/man3/*",
+		"/share/man/man5/*",
+		"/share/man/man7/*",
+		"/share/man/man1",
+		"/share/man/man3",
+		"/share/man/man5",
+		"/share/man/man7",
+		"/include/openssl/*",
+		"/include/openssl",
+		"/lib64/pkgconfig/*",
+		"/lib64/pkgconfig"
+	    ]
+        },
+        {
+            "name" : "lmdb",
+            "buildsystem" : "simple",
+            "build-commands" : [
+                "cd libraries/liblmdb/ && make prefix=/app install"
+            ],
+            "sources" : [
+                {
+                    "type" : "archive",
+                    "url" : "https://github.com/LMDB/lmdb/archive/LMDB_0.9.23.tar.gz",
+                    "sha256" : "abf42e91f046787ed642d9eb21812a5c473f3ba5854124484d16eadbe0aa9c81"
+                }
+            ],
+	    "cleanup" : [
+		"/include/lmdb.h",
+		"/lib/liblmdb.a",
+		"/share/man/man1/*",
+		"/share/man/man1",
+		"/bin/mdb_copy",
+		"/bin/mdb_dump",
+		"/bin/mdb_load",
+		"/bin/mdb_stat"
+	    ]
+        },
+        {
+            "name" : "curl",
+            "buildsystem" : "simple",
+            "build-commands" : [
+                "./configure prefix=/app --with-openssl && make install"
+            ],
+            "sources" : [
+                {
+                    "type" : "archive",
+                    "url" : "https://curl.se/download/curl-8.0.1.tar.gz",
+                    "sha256" : "5fd29000a4089934f121eff456101f0a5d09e2a3e89da1d714adf06c4be887cb"
+                }
+            ],
+	    "cleanup" : [
+		"/bin/curl",
+		"/include/curl/*",
+		"/include/curl",
+		"/lib/libcurl.a",
+		"/bin/curl-config",
+		"/share/man/man1/*",
+		"/share/man/man3/*",
+		"/share/man/man7/*",
+		"/share/man/man1",
+		"/share/man/man3",
+		"/share/man/man7",
+		"/share/aclocal/libcurl.m4",
+		"/share/aclocal",
+		"/lib/pkgconfig/libcurl.pc",
+		"/lib/pkgconfig"
+	    ]
+        },
+	{
+	    "name" : "wxWidgets",
+	    "buildsystem" : "cmake",
+    	    "config-opts" : [
+		    "-DwxUSE_OPENGL=No",
+		    "-DwxUSE_HTML=Yes",
+		    "-DwxUSE_STREAMS=Yes",
+		    "-DwxUSE_WXHTML_HELP=Yes",
+		    "-DCMAKE_BUILD_TYPE=Release",
+		    "-DwxUSE_UNICODE=Yes"
+	    ],
+	    "sources" : [
+		{
+		    "type" : "archive",
+        	    "url" : "https://github.com/wxWidgets/wxWidgets/releases/download/v3.2.2.1/wxWidgets-3.2.2.1.tar.bz2",
+        	    "sha256" : "dffcb6be71296fff4b7f8840eb1b510178f57aa2eb236b20da41182009242c02"
+	  	}
+	    ],
+	    "cleanup" : [
+			"/lib/wx",
+			"/bin/wx-config",
+			"/bin/wxrc",
+			"/bin/wxrc-3.2",
+			"/share/bakefile",
+			"/include/wx-3.2/wx/*",
+			"/include/wx-3.2/wx",
+			"/include/wx-3.2",
+			"/lib/cmake/wxWidgets/*",
+			"/lib/cmake/wxWidgets",
+			"/lib/cmmake"
+	    ]
+	},
+	{
+	    "name" : "tqsl",
+	    "buildsystem" : "cmake",
+    	    "config-opts" : [
+	        "-DTQSL_FLATPAK=Yes",
+		"-DCMAKE_BUILD_TYPE=Release"
+	    ],
+	    "sources" : [
+		{
+		    "type" : "archive",
+        	    "url" : "https://www.rickmurphy.net/lotw/tqsl-2.7.tar.gz",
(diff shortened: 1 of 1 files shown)


k1mu:
Corrective actions: removed unnecessary cleanup actions, added required configuration (icons, screenshot). Changed wx build to cmake-ninja.

org.arrl.trustedqsl.json
```diff
@@ -0,0 +1,148 @@
+{
+    "app-id" : "org.arrl.trustedqsl",
+    "runtime" : "org.freedesktop.Platform",
+    "runtime-version" : "22.08",
+    "sdk" : "org.freedesktop.Sdk",
+    "command" : "tqsl",
+    "finish-args" : [
+	"--share=network",
+	"--socket=x11",
```
hfiguiere, line 9: `share=ipc` is  needed for X11.

org.arrl.trustedqsl.json
```diff
@@ -0,0 +1,148 @@
...
+	"--filesystem=home"
+     ],
+    "modules" : [
+        {
+            "name" : "openssl",
```
hfiguiere, line 14: I will repeat what I said on the PR you closed: you do NOT need to build either openssl or curl.

org.arrl.trustedqsl.json
```diff
@@ -0,0 +1,148 @@
...
+		"*.a",
+		"/share/man",
+		"/bin/mdb_copy",
+		"/bin/mdb_dump",
+		"/bin/mdb_load",
+		"/bin/mdb_stat"
+	    ]
+        },
+        {
+            "name" : "curl",
```
hfiguiere, line 71: repeating once again: curl is in the runtime.

org.arrl.trustedqsl.json
```diff
@@ -0,0 +1,148 @@
...
+	    "buildsystem" : "cmake-ninja",
+    	    "config-opts" : [
+		    "-DwxUSE_OPENGL=No",
+		    "-DwxUSE_HTML=Yes",
+		    "-DwxUSE_STREAMS=Yes",
+		    "-DwxUSE_WXHTML_HELP=Yes",
+		    "-DCMAKE_BUILD_TYPE=Release",
+		    "-DwxUSE_UNICODE=Yes"
+	    ],
+	    "sources" : [
```
hfiguiere, line 112: sources is usually last in the module by good practice. This applies everywhere.

org.arrl.trustedqsl.json
```diff
@@ -0,0 +1,148 @@
...
+			"/bin/wxrc",
+			"/bin/wxrc-3.2",
+			"/share/bakefile",
+			"/include",
+			"/lib/cmake"
+	    ]
+	},
+	{
+	    "name" : "tqsl",
+	    "buildsystem" : "cmake",
```
hfiguiere, line 131: it doesn't build with cmake-ninja ?

org.arrl.trustedqsl.json
```diff
@@ -0,0 +1,148 @@
+{
+    "app-id" : "org.arrl.trustedqsl",
+    "runtime" : "org.freedesktop.Platform",
+    "runtime-version" : "22.08",
+    "sdk" : "org.freedesktop.Sdk",
+    "command" : "tqsl",
+    "finish-args" : [
+	"--share=network",
+	"--socket=x11",
+	"--filesystem=home"
```
hfiguiere, line 10: this is a broad filesystem permission.

org.arrl.trustedqsl.json
k1mu: Added - I missed this in the last round.

org.arrl.trustedqsl.json
k1mu: Well, duh. You're absolutely correct - a build without those works just fine. 
I'm removing this.

org.arrl.trustedqsl.json
k1mu: Removed.

org.arrl.trustedqsl.json
k1mu: Corrected - moved the sources blocks to the end.

org.arrl.trustedqsl.json
k1mu: Yes, it does. Changed.

org.arrl.trustedqsl.json
k1mu: TrustedQSL stores configuration data in $HOME/.tqsl and can be used to sign log files from other logging applications.  I don't think full filesystem access is needed, but the way this is used (primarily as a utility for loggers to sign and upload), home access is needed.

k1mu:
I've pushed an update based on the comments so far.

org.arrl.trustedqsl.json
hfiguiere: then you can use `--persist` instead.

ideally it would use xdg standard locations.

hfiguiere:
bot, build org.arrl.trustedqsl

hfiguiere:
tha app-id in the appstream file is incorrect.

org.arrl.trustedqsl.json
k1mu: I'll try with --persist.

k1mu:
I see the error, will correct and resubmit.

k1mu:
Trying with --persist versus --filesystem=home effectively makes this unusable. None of the configuration data (certificates, locations, etc.) is visible from the home folder and there's no way to import any of the necessary data. Opening files from outside of the container also fails, which is the major purpose of the app - to take a file exported by another application, read that for contact data, sign it, then save or upload the resulting container. 

I understand the desire to restrict what data applications have access to, but this can't run with container-only access.

org.arrl.trustedqsl.json
AsciiWolf: `--persist=.tqsl` should work fine.

k1mu:
I've tried a clean build to verify the behavior with --persist=.tqsl and that's not going to work.
Let me explain this application and what it does, with hopes of making the case for --filesystem=home.

This is an application that has two facets - a GUI tool and a command line tool.
The GUI manages a PKI for the user - generating a certificate request and either saving it for later upload, or directly uploading that request.  The central system performs an identity verification step (which can include using the mail to validate that the requester is indeed the license holder) and issues a signed PKCS6 file back to the user once they're verified. There's some unique certificate extensions to indicate what "country" someone is  from (the country that issued their ham license) as well as a separate set of validity dates. (The certificate might expire in 3 years, but that ham license has a separate validity period that could be just a small number of days.) This is why there's a specialized PKI.
In addition, the GUI allows managing the data about the physical location of the operator - State, County, etc.

That gui management is relatively infrequent, but the data needs to be persistent and portable between computer systems. There's a backup/restore mechanism to allow populating a new computer with the operator's details.

Then, given that setup, a user submits a file of radio contacts in a standardized form (ADIF - Amateur Data Interchange Format).  those contacts are augmented with a digital signature verifying that the operator made that radio contact. The resulting signed file is either saved for later upload, or directly uploaded if there's an Internet connection.

The centralized server, Logbook of the World, matches contacts - station "A" says they contacted station "B" at a given date, time, and other details. Matches are generated and can be used for awards - operating a certain number of countries.

The majority of the time, radio contacts are managed by completely separate logging programs. Those emit ADIF-format logs and use the application's command line interface to digitally sign and optionally upload contacts. It's fairly rare for the users to manually upload themselves (but there's the ability to do that, as well as a GUI menu for creating and editing ADIF files.)

All of that means that an arbitrary program has to be able to pass "tqsl random-file.adif" and have it read, process, and upload the result. So TQSL has to be able to open files wherever that application put them.  I've tried with the mentioned build with "-persist=.tqsl" and I'm not able to open any files outside of that directory.  In addition, the existing ".tqsl" folder isn't being mapped, so  the existing configuration is lost. I can't even copy a backup package to that folder as it's not visible outside of the container.

This application is ancient - initially started in 2001.

tl;dr - think of this as a filter that takes input files and transforms them. I've got to be able to get input from almost anywhere; While I can state that there's a limitation for the flatpak build that any input files have to be somewhere under $HOME I can't expect several logging programs to adapt to a highly restrictive folder structure.

That all being said, I've been wildly off base already here. I am not trying to be argumentative and would be happy to discuss or be proven wrong yet again.

hfiguiere:
> Let me explain this application and what it does, with hopes of making the case for --filesystem=home.

Then `--filesystem=~/.tqsl`

k1mu:
> > Let me explain this application and what it does, with hopes of making the case for --filesystem=home.
> 
> Then `--filesystem=~/.tqsl`

That does get me to the existing configuration data, so that's a start.  However, wxWidgets stores config data in a dotfile off of home, so "--filesystem=~/.tqslapp" is needed as well. But that starts a slippery slope. 

Then how is the application supposed to open file ~/cqrlog/k1mu.adif when that application saves the file?
If the user saves a certificate as ~/callsign.tq6, how can they import it?

I can't dictate this big of a change on the dependent programs. And  having this be so restrictive on Linux versus MacOS/BSD/Windows would be a big downgrade - as well as being a significant regression from the packages that some distributions already have in place (many of which aren't being maintained, thus the idea of a Flatpak build so I can manage this.)
-

k1mu:
Another option here that might be worth trying is to add the xdg-documents and xdg-downloads folders. I haven't tried that to see how it works, but that may be worth considering.

hfiguiere:
AppStream fail to validate.
```
builddir/files/share/appdata/org.arrl.trustedqsl.appdata.xml: FAILED:
Validation of files failed
```

hfiguiere:
bot, build org.arrl.trustedqsl

hfiguiere:
the icon is not found.

hfiguiere:
```
-- Installing: /app/share/app-info/hicolor/icons/128x128/org.arrl.trustedqsl.png
-- Installing: /app/share/app-info/hicolor/icons/64x64/org.arrl.trustedqsl.png
-- Installing: /app/share/app-info/hicolor/icons/48x48/org.arrl.trustedqsl.png
-- Installing: /app/share/app-info/hicolor/icons/32x32/org.arrl.trustedqsl.png
-- Installing: /app/share/app-info/hicolor/icons/16x16/org.arrl.trustedqsl.png
```

This is  very wrong as icon path.

barthalion:
bot, build org.arrl.trustedqsl

barthalion:
Failed `Check for 128x128 icon` typically means a mismatch somewhere between the appdata, desktop and icon files. The compose command first reads the appdata file, uses ID from it (alternatively, refers to `<launchable>`) to look up the desktop file, from which it reads the `Icon` field. Everything has to match the app ID used in the manifest.

hfiguiere:
I don't see a .desktop file installed.

k1mu:
Thanks - that's easily fixed.
I've run the desktop validator against the desktop file, removed the icon spec from the metadata XML file, and added to the desktop. The icon is now being installed to the "right" directory.  Will update the package once I check the build output.

hfiguiere:
bot, build org.arrl.trustedqsl

barthalion:
It segfaults on my end:

```
[bpiotrowski@purei ~]$ coredumpctl info
           PID: 37484 (tqsl)
           UID: 1000 (bpiotrowski)
           GID: 1000 (bpiotrowski)
        Signal: 11 (SEGV)
     Timestamp: Fri 2023-05-12 16:14:51 CEST (30s ago)
  Command Line: tqsl
    Executable: /app/bin/tqsl
 Control Group: /user.slice/user-1000.slice/user@1000.service/app.slice/app-flatpak-org.arrl.trustedqsl-37467.scope
          Unit: user@1000.service
     User Unit: app-flatpak-org.arrl.trustedqsl-37467.scope
         Slice: user-1000.slice
     Owner UID: 1000 (bpiotrowski)
       Boot ID: c75385931ced49a58ed473a5a22c2a09
    Machine ID: b245857f888a412b840acce4c4cc3577
      Hostname: purei
       Storage: /var/lib/systemd/coredump/core.tqsl.1000.c75385931ced49a58ed473a5a22c2a09.37484.1683900891000000.zst (present)
  Size on Disk: 1.4M
       Message: Process 37484 (tqsl) of user 1000 dumped core.
                
                Stack trace of thread 2:
                #0  0x00007f9a57e7bc35 n/a (/usr/lib/x86_64-linux-gnu/libc.so.6 + 0x7bc35)
                #1  0x0000555958df95c0 n/a (/app/bin/tqsl + 0x14f5c0)
                #2  0x00007f9a58790538 n/a (/app/lib/libwx_baseu-3.2.so.0.2.1 + 0x190538)
                #3  0x0000555958d8ad34 n/a (/app/bin/tqsl + 0xe0d34)
                #4  0x00007f9a57e2954a n/a (/usr/lib/x86_64-linux-gnu/libc.so.6 + 0x2954a)
                #5  0x00007f9a57e2960b n/a (/usr/lib/x86_64-linux-gnu/libc.so.6 + 0x2960b)
                #6  0x0000555958d8fb75 n/a (/app/bin/tqsl + 0xe5b75)
                ELF object binary architecture: AMD x86-64
```

k1mu:
The segfault is because there was insufficient checking of the code that creates the $HOME/.tqsl folder - when unable to do that, it'd eventually cause a null pointer dereference. A fix has been committed to the repo, will be pushing an update shortly.

barthalion:
bot, build org.arrl.trustedqsl

PR #4114, opened 2023-05-01
Add org.tryton.Tryton-68

cedk:
This is the same as org.tryton.Tryton-66 but for the series 6.8


hfiguiere:
I would have expected that by now the .desktop, icon and appsteream would come directly from upstream.

org.tryton.Tryton-68.yml
```diff
@@ -0,0 +1,37 @@
...
+runtime: org.gnome.Platform
+runtime-version: '44'
+command: tryton
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=all
+  - --share=network
+  - --talk-name=org.freedesktop.Notifications
```
hfiguiere, line 12: if you use GNotification, you shouldn't need this dbus permission

hfiguiere:
bot, build org.tryton.Tryton-68

cedk:
> I would have expected that by now the .desktop, icon and appstream would come directly from upstream.

The desktop and icon files are different because the application name is not the same.
For appstream also the application name is different but also there is no point to manage it upstream.

hfiguiere:
> > I would have expected that by now the .desktop, icon and appstream would come directly from upstream.
> 
> The desktop and icon files are different because the application name is not the same. 

There is `desktop-file-edit` to fix that. Short of actually making it the same everywhere (nothing is flathub specific, just that we mandate the use of the updated convetions)

> For appstream also the application name is different but also there is no point to manage it upstream.

Yes there is. It's not flatpak specific either.

org.tryton.Tryton-68.yml
cedk: It is using https://docs.gtk.org/gio/method.Application.send_notification.html

cedk:
> like other python app by installing the app from the repo. Or just getting said files from there.

Releases are published because the repository does not contain the generated files.
Also the repository is 100 times bigger than the release archive.

hfiguiere:
bot, build org.tryton.Tryton-68

barthalion:
If the package is considered official, please verify it at https://flathub.org/login when it becomes visible on the website in a few hours.

PR #4140, opened 2023-05-10
Add Obsidian Level Maker Flatpak

dashodanger:

- [x] I am NOT using only the minimal set of permissions.
  - Wayland, Fallback-X11, and X11 sockets are required for the FLTK interface
  - home:rw permissions needed for file creation in a directory of the user's choosing (this program is a map generator, and the created files are its output)


io.github.obsidian-level-maker.yml
```diff
@@ -0,0 +1,44 @@
+app-id: io.github.obsidian-level-maker
```
hfiguiere, line 1: this app-id doesn't meet the requirements. github based id always include the user/org name AND the app name (repository). They have 4 components.

io.github.obsidian-level-maker.yml
```diff
@@ -0,0 +1,44 @@
...
+runtime: org.freedesktop.Platform
+runtime-version: '22.08'
+sdk: org.freedesktop.Sdk
+command: obsidian.sh
+finish-args:
+- --socket=wayland
+- --socket=fallback-x11
+- --socket=x11
```
hfiguiere, line 9: it's either or. of the too.

You mention fltk, but it's not built either.

io.github.obsidian-level-maker.yml
```diff
@@ -0,0 +1,44 @@
...
+- --filesystem=home:rw
```
hfiguiere, line 10: this is a broad permission.

io.github.obsidian-level-maker.yml
```diff
@@ -0,0 +1,44 @@
...
+modules:
+  - name: obsidian
+    buildsystem: cmake
```
hfiguiere, line 13: it doesn't build with cmake-ninja? It's prefered.

io.github.obsidian-level-maker.yml
```diff
@@ -0,0 +1,44 @@
...
+    config-opts:
+    - -DCMAKE_BUILD_TYPE=Release
+    sources:
+      - type: git
+        url: http://github.com/obsidian-level-maker/Obsidian
+        tag: Obsidian-v20-20230509
+        commit: fb2c9f6269ced76643fa4c362738784ebae26cf7
+    build-commands:
+      - install -D obsidian /app/bin/obsidian
+  - name: startscript
```
hfiguiere, line 23: this should be merged with the main module.

io.github.obsidian-level-maker.yml
```diff
@@ -0,0 +1,44 @@
...
+      - type: script
+        commands:
+          - obsidian --install /app
+        dest-filename: obsidian.sh
+      - type: file
+        path: io.github.obsidian-level-maker.desktop
+      - type: file
+        path: io.github.obsidian-level-maker.metainfo.xml
+      - type: file
+        path: io.github.obsidian-level-maker.64.png
```
hfiguiere, line 35: you don't need a 64 pixel icon.

io.github.obsidian-level-maker.yml
dashodanger: A vendored version of FLTK from our repository is both included and built.

io.github.obsidian-level-maker.yml
```diff
@@ -0,0 +1,44 @@
...
+        tag: Obsidian-v20-20230509
+        commit: fb2c9f6269ced76643fa4c362738784ebae26cf7
+    build-commands:
+      - install -D obsidian /app/bin/obsidian
+  - name: startscript
+    buildsystem: simple
+    sources:
+      - type: script
+        commands:
+          - obsidian --install /app
```
hfiguiere, line 28: not sure what `--install` does, but if it is supposed to perform an install, then this will fail as `/app` is read only.

io.github.obsidian-level-maker.yml
dashodanger: From the Flatpak documentation

"io.github.Foo

This is problematic because while foo.github.io may be unique to your project, it does not include a project-specific identifier. This may cause issues if another project creates io.github.Foo-Bar which should be its own namespace but areas of flatpak may treat them similar. A better ID would be io.github.foo.Foo even if it is redundant."

This language is not definitive. If there is a hard requirement for what you suggest, I recommend updating the documentation to reflect this.

io.github.obsidian-level-maker.metainfo.xml
```diff
@@ -0,0 +1,18 @@
...
+  <project_license>GPL-2.0-only</project_license>
+  
+  <description>
+    <p>
+      Obsidian Level Maker is a community continuation of OBLIGE Level Maker, originally by Andrew Apted. It generates random levels that are (hopefully) fun to play. The following games are currently supported: - Doom 1/The Ultimate Doom/Freedoom Phase 1 - Doom 2/Freedoom Phase 2 - Final Doom - Heretic - Hexen - Chex Quest 3 - HacX 1.2 - Harmony - Strife - Wolfenstein 3-D - Spear of Destiny - Super Noah&apos;s Ark 3-D
+    </p>
+  </description>
+  
+  <launchable type="desktop-id">io.github.obsidian-level-maker.desktop</launchable>
+</component>
```
hfiguiere, line 18: missing mandatoy `url`, `content_rating`, `screenshots`, `releases`.

io.github.obsidian-level-maker.yml
dashodanger: That is a subject term. Please explain what you mean.

io.github.obsidian-level-maker.yml
dashodanger: Why is it preferred?

io.github.obsidian-level-maker.yml
dashodanger: Why?

io.github.obsidian-level-maker.yml
dashodanger: It designates the /app directory as the location of various files that it reads on startup. It does not "install" anything, despite the parameter name.

hfiguiere (changes requested):
Given the content of https://github.com/obsidian-level-maker/Obsidian/tree/Obsidian-v20-20230509/flatpak_stuff there is no reason to have more than the manifest + eventual patches.

io.github.obsidian-level-maker.metainfo.xml
dashodanger: I used the recommended metadata generator. Where is it documented that these are mandatory?

io.github.obsidian-level-maker.yml
dashodanger: > Given the content of https://github.com/obsidian-level-maker/Obsidian/tree/Obsidian-v20-20230509/flatpak_stuff there is no reason to have more than the manifest + eventual patches.

I do not understand what this means.

dashodanger:
> Given the content of https://github.com/obsidian-level-maker/Obsidian/tree/Obsidian-v20-20230509/flatpak_stuff there is no reason to have more than the manifest + eventual patches.

flatpak_stuff is a transient location that I used for convenience between virtual machines. It is not planned to remain.

hfiguiere:
> > Given the content of https://github.com/obsidian-level-maker/Obsidian/tree/Obsidian-v20-20230509/flatpak_stuff there is no reason to have more than the manifest + eventual patches.
> 
> flatpak_stuff is a transient location that I used for convenience between virtual machines. It is not planned to remain.

None of this is flatpak specific anyway, so it belongs in the upstream repository.

io.github.obsidian-level-maker.yml
hfiguiere: because that's a good practice.

io.github.obsidian-level-maker.yml
hfiguiere: because it's faster with ninja.

io.github.obsidian-level-maker.yml
hfiguiere: that granting permission to the whole HOME directory is considered broad.

io.github.obsidian-level-maker.yml
hfiguiere: then `socket=x11` should be removed if it supports wayland *out of the box*.

dashodanger: Done

io.github.obsidian-level-maker.yml
dashodanger: This program generates maps for a variety of Doom engine games. We cannot predict how a user organizes the custom content for their games nor where they are stored within the user's home directory.

io.github.obsidian-level-maker.yml
hfiguiere: so you want to question the reviewers? Feel free, but the linter will not let it through. The flatpak documentation (that you are quoting) is not flathub's. But clearly say it is problematic, so it is only ambiguous if you want to ignore it.

io.github.obsidian-level-maker.yml
dashodanger: It is now removed

io.github.obsidian-level-maker.yml
dashodanger: The documentation may not be flathub's, but is linked from https://github.com/flathub/flathub/wiki/App-Requirements. I read the "problematic" language, but also ran into an issue where we cannot have dashes in the 'middle' of the app-id. Since we do not have a site like "obsidian_level_maker.github.io", I decided to err on using a name correlating to a site that is under our purview.

io.github.obsidian-level-maker.yml
dashodanger: To me, this is a logical separation of building the application itself and 'building' the extra shell script and files associated with being a flatpak.

dashodanger:
> > > Given the content of https://github.com/obsidian-level-maker/Obsidian/tree/Obsidian-v20-20230509/flatpak_stuff there is no reason to have more than the manifest + eventual patches.
> > 
> > 
> > flatpak_stuff is a transient location that I used for convenience between virtual machines. It is not planned to remain.
> 
> None of this is flatpak specific anyway, so it belongs in the upstream repository.

What belongs in the upstream repository? Why should the files whose only reason to exist is to support a flatpak not be in the same repository as the manifest?

hfiguiere:
As I said, none of these files in specific to flatpak. Any desktop application on Linux (and *BSD) should have an icon and a .desktop. It's a 20 yr old spec now, shared by all the desktop environments.
As for the AppStream file, it's much newer, but again not flatpak specific (a predates it)

They are part of the requirements, because if they weren't they'd be missing most of the time.

io.github.obsidian-level-maker.yml
dashodanger: Updated to use cmake-ninja

dashodanger:
> As I said, none of these files in specific to flatpak. Any desktop application on Linux (and *BSD) should have an icon and a .desktop. It's a 20 yr old spec now, shared by all the desktop environments. As for the AppStream file, it's much newer, but again not flatpak specific (a predates it)
> 
> They are part of the requirements, because if they weren't they'd be missing most of the time.

We do not install Linux as a desktop application when it is built for Linux, and as such we have never had the files you mention before. The icon for the window itself is embedded in the source as an XPM file. Is it MANDATORY to have only the manifest, or are you simply expressing an opinion?

dashodanger:
Closing in favor of working on a distribution method with less hoop-jumping.

PR #4144, opened 2023-05-10
Add Minion, gg.minion.Minion

zastrixarundell:

- ~~Minion is a mod manager for WoW and Elder Scrolls Online, it can be installed from everywhere on the system. So the host filesystem is required.~~

- I can't give a direct link to the conversation as it's a private conversation but: 
image. Nothing came out of forking the repo so I'm posting it as so.

- I'm not sure exactly what this implies, but I made the application ID same as the website for the client and after analyzing the code after it was decompiled, the main class was `gg.minion.Minion`.

- I did add a JRE, and removed MacOS items but that shouldn't be added to the upstream (also it's closed source). 


flathub.json
hfiguiere, line 1: you don't need this file. These are the only architecture supported.

icons/512.png
hfiguiere, line 1: since the icons don't come from the upstream package, you should only have one here.

gg.minion.Minion.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id>gg.minion.Minion</id>
+  <name>Minion</name>
+  <developer_name>Good Game Mods, LLC</developer_name>
+  <summary>Premier Addon Management</summary>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>LicenseRef-proprietary</project_license>
+  <url type="homepage">https://minion.mmoui.com/</url>
```
hfiguiere, line 9: why does this differ from the app-id

gg.minion.Minion.yml
```diff
@@ -0,0 +1,78 @@
...
+        path: icons
+      - type: file
+        path: gg.minion.Minion.desktop
+      - type: script
+        dest-filename: minion.sh
+        commands:
+          # If someone does press udpate it won't create a null folder + it won't hang the application
+          - cd /tmp
+          - java -jar /app/extra/Minion-jfx.jar
+finish-args:
```
hfiguiere, line 71: finish-args is usually above modules.

gg.minion.Minion.yml
```diff
@@ -0,0 +1,78 @@
...
+  - --env=PATH=$PATH:/app/java/jre/bin:/app/bin
+  - --socket=x11
+  - --socket=fallback-x11
```
hfiguiere, line 74:
```suggestion
  - --socket=x11
```

gg.minion.Minion.yml
```diff
@@ -0,0 +1,78 @@
...
+  - --share=network
+  - --device=dri
+  - --filesystem=host
```
hfiguiere, line 77: this is a very broad filesystem permission

gg.minion.Minion.xml
zastrixarundell: Do you mean the URL? Another valid url for it is `https://minion.gg/`, I can place it there instead.

hfiguiere: that's, just asking. sometime people fell that a random domain if fine as an app-id

gg.minion.Minion.xml
hfiguiere, line 1: this file has an incorrect name. and is never installed either.

zastrixarundell:
bot, build gg.minion.Minion

icons/512.png
hfiguiere, line 1: I wasn't joking about the unecessary icon files.

zastrixarundell: I thought that adding multiple of the same images can save on performance / looks when the correct size exists. Which one should I keep then?

hfiguiere: the biggest.

zastrixarundell:
bot, build gg.minion.Minion

gg.minion.Minion.yml
zastrixarundell: This program is primarily made to be ran with wine games and the files can change. I can try and default it to allow steam and lutris, then add additional documentation for using flatseal or something.

zastrixarundell: Actually, after editing to not use the host, Minion doesn't want to work properly. It completely ignores folder selection to manage addons in there. 

Also without `host` it won't work well for installations on custom paths (i.e. installing addons for a game which is on another drive or system linking it).

gg.minion.Minion.yml
```diff
@@ -0,0 +1,75 @@
+app-id: gg.minion.Minion
+runtime: org.freedesktop.Platform
+runtime-version: '22.08'
+command: minion.sh
+sdk: org.freedesktop.Sdk
+tags:
+  - proprietary
+finish-args:
+  - --env=PATH=$PATH:/app/java/jre/bin:/app/bin
```
hfiguiere, line 9:
```suggestion
  - --env=PATH=/app/java/jre/bin:/app/bin
```

There is no env substitution.

gg.minion.Minion.yml
```diff
@@ -0,0 +1,75 @@
...
+  - --socket=x11
+  - --share=network
+  - --filesystem=host
+  - --filesystem=~/.var/app/com.valvesoftware.Steam
+  - --filesystem=~/.var/app/net.lutris.Lutris
```
hfiguiere, line 14: this are redundant with the broad "host"

zastrixarundell: I'm guessing that flatpaks don't allow editing of files between themselves unless explicitly allowed so?

zastrixarundell:
bot, build gg.minion.Minion

README.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Disclaimer (important to read)
+
+### Regarding Hardware Acceleration
+
+As hardware acceleration doesn't work well with XWayland under Wayland with the closed-source Nvidia drivers, hardware acceleration was disabled to have higher coverage for GPUs.
+
+One may enable hardware acceleration by running `flatpak override --user --device=dri gg.minion.Minion`. Be wary, this will cause flickering for Minion if you're using the proprietary Nvidia driver while running Wayland.
+
+### First start
```
hfiguiere, line 12: just to clarify expectations, no user will read this.

zastrixarundell: I know, but just put it there just in case so that potentially I don't get issue reports and PRs in the future for dri being enabled.

gg.minion.Minion.yml
zastrixarundell: Using the `home` filesystem option should actually work in most scenarios as per [this commit](https://github.com/flathub/flathub/pull/4144/commits/4b5c6d36da55a2edcc90b785f2ccabd3932e0e65). I tried manually adding `~/.minion` but it doesn't want to create the folder (it works if it already exists). Usually people also syslink to the Documents folder so they don't need to go through ~10 different subdirectories so I think only home is fine for this app considering its' usage.

gg.minion.Minion.yml
zastrixarundell: Same issue is happening when using `filesystem=home`, without these two lines it makes Minion essentially unusuable.

gg.minion.Minion.metainfo.xml
```diff
@@ -0,0 +1,65 @@
...
+    <content_attribute id="language-profanity">none</content_attribute>
+    <content_attribute id="language-humor">none</content_attribute>
+    <content_attribute id="language-discrimination">none</content_attribute>
+    <content_attribute id="social-chat">none</content_attribute>
+    <content_attribute id="social-info">none</content_attribute>
+    <content_attribute id="social-audio">none</content_attribute>
+    <content_attribute id="social-location">none</content_attribute>
+    <content_attribute id="social-contacts">none</content_attribute>
+    <content_attribute id="money-purchasing">none</content_attribute>
+    <content_attribute id="money-gambling">none</content_attribute>
```
barthalion, line 61: Attributes set to `none` can be dropped.

gg.minion.Minion.yml
```diff
@@ -0,0 +1,75 @@
...
+runtime-version: '22.08'
+command: minion.sh
+sdk: org.freedesktop.Sdk
+tags:
+  - proprietary
+finish-args:
+  - --env=PATH=/app/java/jre/bin:/app/bin
+  - --socket=x11
+  - --share=network
+  - --filesystem=home
```
barthalion, line 12: Have you checked if `--persist=.minion` helps here?

gg.minion.Minion.yml
```diff
@@ -0,0 +1,75 @@
...
+  - name: minion
+    buildsystem: simple
+    build-commands:
+      - install -Dm755 apply-extra                ${FLATPAK_DEST}/bin/apply_extra
+      - install -Dm755 minion.sh                  ${FLATPAK_DEST}/bin/minion.sh
+      - install -Dm644 ${FLATPAK_ID}.png          ${FLATPAK_DEST}/share/icons/hicolor/256x256/apps/${FLATPAK_ID}.png
+      - install -Dm644 ${FLATPAK_ID}.desktop      ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop
+      - install -Dm644 ${FLATPAK_ID}.metainfo.xml ${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.metainfo.xml
+    sources:
+      - type: extra-data
```
barthalion, line 42: Have you asked upstream if we could redistribute it using our CDN? This way users get delta upgrades instead of redownloading this file from scratch every time.

barthalion:
Also can you ask upstream for their GitHub handles? We will invite them to the Flathub repo once this is accepted.

gg.minion.Minion.yml
zastrixarundell: That seems to have fixed the issue I was having, thanks!

gg.minion.Minion.yml
zastrixarundell: Using `--persist=.minion` fixed the issue.

zastrixarundell:
@barthalion haven't asked for their private handles, but this is the Organization account for Minion which will host Minion4 https://github.com/minion-gg.

gg.minion.Minion.metainfo.xml
zastrixarundell: Having all contents set to none, as Minion is just an addon manager, can I drop `content_rating` attribute altogether, or just keep i empty?

gg.minion.Minion.yml
zastrixarundell: There shouldn't be any delta upgrades at all as this is the final version of Minion 3, hasn't been for more than a year. When Minion 4 comes out I do think that it'll need essentially a new builder file altogether.

zastrixarundell:
bot, build gg.minion.Minion

zastrixarundell:
bot, build gg.minion.Minion

barthalion:
bot, build gg.minion.Minion

gg.minion.Minion.yml
zastrixarundell: I've also asked upstream for their github handles to add them when this get gets published. Usually I get a response within a couple of days but can't give an ETA. 

When they respond whether they want to have this on the flathub CDN I can easily add it is a dependency to be hosted on the flathub CDN. The download in `apply-extra` is only 7.7 MiB.

barthalion: > There shouldn't be any delta upgrades at all as this is the final version of Minion 3, hasn't been for more than a year. When Minion 4 comes out I do think that it'll need essentially a new builder file altogether.

Is the plan to update this ID to Minion 4 once it's released, or is this going to be a new submission?

barthalion:
bot, build gg.minion.Minion

PR #4161, opened 2023-05-18
Add Easy Audio Sync

complexlogic:
Easy Audio Sync: an audio library syncing and conversion utility

  - Access to `/media` and `/run/media` are needed for writing to removable storage drives. This program is designed to write to an SD card for a portable media device.
  - N/A

### DIFF
--- a/io.github.complexlogic.EasyAudioSync.yml
+++ b/io.github.complexlogic.EasyAudioSync.yml
@@ -1,0 +1,29 @@
+app-id: io.github.complexlogic.EasyAudioSync
+runtime: org.kde.Platform
+runtime-version: 5.15-22.08
+sdk: org.kde.Sdk
+command: easyaudiosync
+finish-args:
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --filesystem=home
+  - --filesystem=/media
+  - --filesystem=/run/media
+  - --filesystem=xdg-run/gvfs
+modules:
+  - name: EasyAudioSync
+    buildsystem: cmake-ninja
+    config-opts:
+    - -DCMAKE_BUILD_TYPE=Release
+    - -DCMAKE_SYSTEM_NAME=Linux
+    - -DSTRIP_BINARY=ON
+    - -DQT_VERSION=5
+    - -DVCPKG=ON
+    sources:
+      - sources.yml
+    build-options:
+        build-args:
+        - --share=network
+        env:
+          VCPKG_FORCE_SYSTEM_BINARIES: '1'
+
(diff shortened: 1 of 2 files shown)


io.github.complexlogic.EasyAudioSync.yml
```diff
@@ -0,0 +1,29 @@
...
+  - name: EasyAudioSync
+    buildsystem: cmake-ninja
+    config-opts:
+    - -DCMAKE_BUILD_TYPE=Release
+    - -DCMAKE_SYSTEM_NAME=Linux
+    - -DSTRIP_BINARY=ON
+    - -DQT_VERSION=5
+    - -DVCPKG=ON
+    sources:
+      - sources.yml
```
barthalion, line 23: This should be part of the manifest, instead of being split.

io.github.complexlogic.EasyAudioSync.yml
```diff
@@ -0,0 +1,29 @@
...
+    build-options:
+        build-args:
+        - --share=network
```
barthalion, line 26: We don't allow network access during build.

complexlogic: The reason I've asked for network access is because my build script relies on the CMake `FetchContent` module which requires an external connection.

Are there any exceptions allowed to this rule?

hfiguiere: no. You need to download them by declaring them in the manifest.

io.github.complexlogic.EasyAudioSync.yml
```diff
@@ -0,0 +1,29 @@
+app-id: io.github.complexlogic.EasyAudioSync
+runtime: org.kde.Platform
+runtime-version: 5.15-22.08
+sdk: org.kde.Sdk
+command: easyaudiosync
+finish-args:
+  - --socket=fallback-x11
```
hfiguiere, line 7: `--share=ipc` is needed too.

io.github.complexlogic.EasyAudioSync.yml
```diff
@@ -0,0 +1,29 @@
...
+  - --filesystem=/media
+  - --filesystem=/run/media
+  - --filesystem=xdg-run/gvfs
+modules:
+  - name: EasyAudioSync
+    buildsystem: cmake-ninja
+    config-opts:
+    - -DCMAKE_BUILD_TYPE=Release
+    - -DCMAKE_SYSTEM_NAME=Linux
+    - -DSTRIP_BINARY=ON
```
hfiguiere, line 19:
```suggestion
```

flatpak will call strip to split out the debug symbols.

io.github.complexlogic.EasyAudioSync.yml
complexlogic: OK. In that case, I have to close this and figure out an alternative solution.

PR #4163, opened 2023-05-18
Add app.drey.PaperPlane

marhkb:

We only want to use the beta branch for now, until we have a first stable release sometime in the future. We think that's a good way to keep users up to date to gather more user feedback.

We use a patch on top on gtk to change the `GtkListView`'s direction from bottom to top for our messages view. There's a MR for that (https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/4447). But until now it's unclear whether this will be the final solution that will land.

Additionally, we us a specific libadwaita commit that has `AdwToolBar` and is still compatible with GTK 4.10.

### DIFF
--- a/app.drey.PaperPlane.json
+++ b/app.drey.PaperPlane.json
@@ -71,6 +71,34 @@
             ]
         },
         {
+            "name": "libsass",
+            "buildsystem": "meson",
+            "cleanup": [
+                "*"
+            ],
+            "sources": [
+                {
+                    "type": "git",
+                    "url": "https://github.com/lazka/libsass.git",
+                    "commit": "302397c0c8ae2d7ab02f45ea461c2c3d768f248e"
+                }
+            ]
+        },
+        {
+            "name": "sassc",
+            "buildsystem": "meson",
+            "cleanup": [
+                "*"
+            ],
+            "sources": [
+                {
+                    "type": "git",
+                    "url": "https://github.com/lazka/sassc.git",
+                    "commit": "82803377c33247265d779af034eceb5949e78354"
+                }
+            ]
+        },
+        {
             "name": "gtk",
             "buildsystem": "meson",
             "config-opts": [
@@ -95,36 +123,6 @@
                 {
                     "type": "patch",
                     "path": "gtk-reversed-list.patch"
-                }
-            ],
-            "modules": [
-                {
-                    "name": "libsass",
-                    "buildsystem": "meson",
-                    "cleanup": [
-                        "*"
-                    ],
-                    "sources": [
-                        {
-                            "type": "git",
-                            "url": "https://github.com/lazka/libsass.git",
-                            "commit": "302397c0c8ae2d7ab02f45ea461c2c3d768f248e"
-                        }
-                    ]
-                },
-                {
-                    "name": "sassc",
-                    "buildsystem": "meson",
-                    "cleanup": [
-                        "*"
-                    ],
-                    "sources": [
-                        {
-                            "type": "git",
-                            "url": "https://github.com/lazka/sassc.git",
-                            "commit": "82803377c33247265d779af034eceb5949e78354"
-                        }
-                    ]
                 }
             ]
         },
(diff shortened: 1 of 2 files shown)


marhkb:
bot, build app.drey.PaperPlane

marhkb:
bot, build app.drey.PaperPlane

app.drey.PaperPlane.json
```diff
@@ -0,0 +1,264 @@
...
+            ],
+            "sources": [
+                {
+                    "type": "git",
+                    "url": "https://gitlab.gnome.org/GNOME/libadwaita.git",
+                    "commit": "57bc21b4c51aa361609fe6f57031630589391b0b"
+                }
+            ],
+            "modules": [
+                {
```
hfiguiere, line 202: these are already built above.

app.drey.PaperPlane.json
```diff
@@ -0,0 +1,264 @@
...
+                    "commit": "8517026415e75a8eec567774072cbbbbb52376c1"
+                }
+            ]
+        },
+        {
+            "name": "paper-plane",
+            "buildsystem": "meson",
+            "config-opts": [
+                "-Dtg_api_id=22303002",
+                "-Dtg_api_hash=3cc0969992690f032197e6609b296599"
```
hfiguiere, line 252: if these are supposed to be secrets then they no longer are.

app.drey.PaperPlane.json
marhkb: Fixed

marhkb:
bot, build app.drey.PaperPlane

app.drey.PaperPlane.json
```diff
@@ -0,0 +1,232 @@
...
+                    "tag": "4.10.3"
+                },
+                {
+                    "type": "patch",
+                    "path": "gtk-reversed-list.patch"
+                }
+            ]
+        },
+        {
+            "name": "appstream",
```
hfiguiere, line 130: btw if you use this to generate the release in the appstream file, you should make that part of the `dist` process that generate your tarball instead. Because that's were it belongs.

Not a blocker for flathub though.

app.drey.PaperPlane.json
```diff
@@ -0,0 +1,232 @@
...
+            "buildsystem": "meson",
+            "config-opts": [
+                "-Dbuildtype=release",
+                "-Dstemming=false",
+                "-Dgir=false",
+                "-Dapidocs=false"
+            ],
+            "cleanup": [
+                "/bin/appstreamcli"
```
hfiguiere, line 139: why not "*" ?

app.drey.PaperPlane.json
```diff
@@ -0,0 +1,232 @@
...
+                {
+                    "type": "git",
+                    "url": "https://github.com/ximion/appstream.git",
+                    "commit": "1ee6af91310f9deb47f3863a7a7ffb22c46bcce6",
+                    "tag": "v0.16.2"
+                }
+            ],
+            "modules": [
+                {
+                    "name": "libyaml",
```
hfiguiere, line 151: if these are only for appstream, then they should both have a "*" cleanup rule.

app.drey.PaperPlane.json
marhkb: Appstream is used here rather as a runtime dependency for libadwaita (see https://gitlab.gnome.org/GNOME/libadwaita/-/commit/876672aac1191e4ac26d835cb75761ce07ef69ec).

app.drey.PaperPlane.json
marhkb: We can't, because libs are used at runtime.

app.drey.PaperPlane.json
marhkb: Same here. We'll end with
```
paper-plane: error while loading shared libraries: libyaml-0.so.2: cannot open shared object file: No such file or directory
```

app.drey.PaperPlane.json
hfiguiere: why is it needed at runtime? Isn't this a telegram client?

marhkb: Starting fails with
```
paper-plane: error while loading shared libraries: libappstream.so.4: cannot open shared object file: No such file or directory
```

AFAIK, libadwaita introduced `AdwAboutWindow::new_from_appdata` recently where it uses appstream. But `appstream` still isn't part of the GNOME runtime 44, but will is going to be in 45, AFAIK.

See https://gitlab.gnome.org/GNOME/libadwaita/-/commit/876672aac1191e4ac26d835cb75761ce07ef69ec

PR #4252, opened 2023-06-16
me.timschneeberger.jdsp4linux

ThePBone:

~This application has two flavors: one compiled for PulseAudio (`me.timschneeberger.jdsp4linux.pulse`) and one for PipeWire (`me.timschneeberger.jdsp4linux.pipewire`). Because of that, they need to have ownership of the shared D-Bus service name `--own-name me.timschneeberger.jdsp4linux`.
It would be impractical for 3rd-party developers having to deal with two different service names depending on which flavor is installed.~

~Both of the flavors clarify this in their metainfo XMLs like this:~
```xml
<provides>
    <binary>jamesdsp</binary>
    <dbus type="user">me.timschneeberger.jdsp4linux</dbus>
</provides>
```
~This submission contains the PipeWire flavor. The submission of the PulseAudio flavor can be found in PR #4251.~

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://github.com/Audio4Linux/JDSP4Linux


ThePBone:
bot build, me.timschneeberger.jdsp4linux.pipewire

ThePBone:
bot, build me.timschneeberger.jdsp4linux.pipewire

ThePBone:
bot, build me.timschneeberger.jdsp4linux.pulse

hfiguiere:
but do you really need two apps? How is one chosen over the over. Also pipewire somewhat unifies everything so is a pulse variant necessary?

ThePBone:
> but do you really need two apps? How is one chosen over the over. Also pipewire somewhat unifies everything so is a pulse variant necessary?

Yes. This is a desktop audio effects processor (like EasyEffects) that needs to target advanced Pulse/PipeWire APIs directly. 
There are no unified APIs for this particular use case. 
The user chooses a flavor based on whether they installed PipeWire or PulseAudio on their system.

Pipewire does have a Pulseaudio compatibility mode, however, it doesn't re-implement the particular APIs that I need.

ThePBone:
bot, build me.timschneeberger.jdsp4linux.pipewire

hfiguiere:
> The user chooses a flavor based on whether they installed PipeWire or PulseAudio on their system.

If you run this as a flatpak, there is pipewire, and I don't think anything else matter. Otherwise it's pretty poor UX to have to select a different app due to an implementation detail.

ThePBone:
>If you run this as a flatpak, there is pipewire, and I don't think anything else matter. Otherwise it's pretty poor UX to have to select a different app due to an implementation detail.

No, it needs to access the host installation of Pulse or Pipewire. It can't do that part in the sandbox, because it is supposed to intercept and manipulate the audio stream of other native applications on the host system.

I could drop PulseAudio support altogether as EasyEffects did for their app; however, I want to avoid that because some users depend on Pulse support.

ThePBone:
Change of plans: Since the PulseAudio version is legacy stuff, I'll only submit the new PipeWire build (under a new app id), which is the recommended version anyway.

This is a good opportunity to push users that are still sticking to the legacy pulseaudio implementation to the new pipewire-focused implementation that has much better performance.

hfiguiere:
you didn't need to close this.

ThePBone:
bot, build me.timschneeberger.jdsp4linux

me.timschneeberger.jdsp4linux.yaml
```diff
@@ -0,0 +1,48 @@
...
+  - "*.m4"
+  - "*.pm"
+  - /include
+  - /lib/pkgconfig
+  - /share/pkgconfig
+modules:
+  - name: jamesdsp
+    buildsystem: qmake
+    modules:
+      - jamesdsp-modules.yaml
```
hfiguiere, line 28: the *content* of this should be inserted above line 25.

It's more maintainable.

ThePBone:
bot, build me.timschneeberger.jdsp4linux

PR #4276, opened 2023-06-24
Add page.codeberg.libre_menu_editor.LibreMenuEditor

libre-menu-editor:


org.codeberg.LibreMenuEditor.yml
```diff
@@ -0,0 +1,42 @@
+app-id: org.codeberg.LibreMenuEditor
```
hfiguiere, line 1: this app-id is not valid:

- code hosting always require 4 component. 3 being the user/org and 4 being the app
- it must use the domain for hosting webpage.

org.codeberg.LibreMenuEditor.yml
```diff
@@ -0,0 +1,42 @@
...
+  - --filesystem=/var/lib/flatpak:ro
+    # for reading .desktop files in /var/lib/flatpak/exports/share/applications
+    # to avoid unreachable symlink targets in /var/lib/flatpak
+  - --filesystem=/var/lib/snapd:ro
+    # for reading .desktop files in /var/lib/snapd/desktop/applications
+    # to avoid unreachable symlink targets in /var/lib/snapd
+  - --filesystem=host:ro
+    # for reading .desktop files in /usr/local/share/applications
+    # for reading .desktop files in /usr/share/applications
+    # to avoid unreachable symlink targets
```
hfiguiere, line 32: this filesystem permissions are very excessive, and most of them not allowed by default.

org.codeberg.LibreMenuEditor.yml
libre-menu-editor: so one of these would be ok?

org.codeberg.LibreMenuEditor.LibreMenuEditor
org.codeberg.libremenueditor.LibreMenuEditor
org.codeberg.libre-menu-editor.LibreMenuEditor

hfiguiere: neither.

`page.codeberg.libre_menu_editor.LibreMenuEditor`

org.codeberg.LibreMenuEditor.yml
libre-menu-editor: --filesystem=host:ro is necessary for two reasons:
  1. it can not reach /usr/share/applications, so I have to get it through /run/host
  2. if there is a symlink in any of the directories and it points to a directory that is not accessible, my program can not load the .desktop file

I had to include --filesystem=/var/lib/flatpak:ro because it wouldn't find /run/host/var/lib/flatpak:ro 
I had to include --filesystem=/var/lib/snapd:ro because it wouldn't find /run/host/var/lib/snapd:ro 
--filesystem=xdg-data/flatpak:ro is for loading .desktop files of user-installed flatpaks

--filesystem=xdg-data/applications needs write access, because it needs to write the users changes

org.codeberg.LibreMenuEditor.yml
libre-menu-editor: Why "page.codeberg.libre_menu_editor.LibreMenuEditor" and not "org.codeberg.libre_menu_editor.LibreMenuEditor"?

libre-menu-editor: Does it need to be "page".codeberg.libre_menu_editor.LibreMenuEditor? I would like it much better if it started with "org"

libre-menu-editor: Never mind, I think I misunderstood when you said "it must use the domain for hosting webpage". I will create a webpage with codeberg.page and update the app-id to page.codeberg.{username}.{app-name}

hfiguiere: you don't need to create anything here. Just to set the app-id as needed.

libre-menu-editor: My username on codeberg is "libre-menu-editor" and not "libre_menu_editor". So https://libre-menu-editor.codeberg.page works and https://libre_menu_editor.codeberg.page does not work. If someone created a codeberg-account with the username "libre_menu_editor", there could be confusion.

Therefore I assume that I should use the app-id "page.codeberg.libre-menu-editor.LibreMenuEditor" instead of "page.codeberg.libre_menu_editor.LibreMenuEditor". Please tell me which one is correct.

It would be nice if the all the names were as consistent as possible.

page.codeberg.libre_menu_editor.LibreMenuEditor.yml
```diff
@@ -0,0 +1,42 @@
...
+  - --filesystem=host:ro
+    # for reading .desktop files in /usr/local/share/applications
+    # for reading .desktop files in /usr/share/applications
+    # to avoid unreachable symlink targets
+modules:
+  - name: git
+    buildsystem: simple
+    build-commands:
+      - cd libre-menu-editor; for file in $(find .); do install -D $file /app/libre-menu-editor/$file; done
+      - cd export; for file in $(find .); do install -D $file /app/$file; done
```
hfiguiere, line 38: I am of the opinion that this should be upstream, part of the `make install`. (all it needs it to honour prefix)

libre-menu-editor: I am sorry, I don't understand. I don't use "make install" anywhere

hfiguiere: and that's what I am saying. The install procedure should be in the upstream repository.

It's not blocking but it make the maintenance harder. 

(also it's very cryptic)

libre-menu-editor: If there is no problem, I would like to leave it this way

org.codeberg.LibreMenuEditor.yml
hfiguiere, line 1: the app-id still doesn't match, nor the filename.

libre-menu-editor: I uploaded a second file that should be correct. Do I need to delete the old one?

libre-menu-editor: I had uploaded a file that should be correct. The old one is deleted now

hfiguiere:
bot, build page.codeberg.libre_menu_editor.LibreMenuEditor

page.codeberg.libre_menu_editor.LibreMenuEditor.yml
```diff
@@ -0,0 +1,42 @@
+app-id: page.codeberg.libre_menu_editor.LibreMenuEditor
+runtime: org.gnome.Platform
+runtime-version: "44"
+sdk: org.gnome.Sdk
+command: libre-menu-editor
+finish-args:
+  - --env=FLATPAK_APP_DIR=/app/libre-menu-editor
```
barthalion, line 7: This can be overwritten. I recommend adding a shell wrapper that will export this variable unconditionally.

page.codeberg.libre_menu_editor.LibreMenuEditor.yml
```diff
@@ -0,0 +1,42 @@
+app-id: page.codeberg.libre_menu_editor.LibreMenuEditor
```
barthalion, line 1: Can you please verify that `libre_menu_editor.codeberg.page` is actually the website codeberg allows for your username?

libre-menu-editor: My username on codeberg.org is "libre-menu-editor" and not "libre_menu_editor" and the url that I can use on codeberg (if I enable the codeberg.page feature) is "https://libre-menu-editor.codeberg.page" and not "https://libre_menu_editor.codeberg.page".

Since hfiguiere told me that "page.codeberg.libre_menu_editor.LibreMenuEditor" was "needed", I assumed that all "-" would always be changed into "_" in the flatpak/flathub world.

If this was a mistake, I would be happy to correct it. Having all the names and links consistent is much nicer anyway.

page.codeberg.libre_menu_editor.LibreMenuEditor.yml
libre-menu-editor: Thank you, I will take it into consideration. Since you said "recommend", I assume that this is not a requirement. Please correct me if I am wrong.

TingPing: I think it is strange that it is an env var at all?

This is a path that will never change and isn't configurable.

hfiguiere: I'm of the opinion this should be part of what I'm suggesting here https://github.com/flathub/flathub/pull/4276#discussion_r1240963797

libre-menu-editor:
bot, build page.codeberg.libre_menu_editor.LibreMenuEditor

libre-menu-editor:
Please tell me if the manifest is acceptable now.

Please also tell me if I should change the app-id to "page.codeberg.libre-menu-editor.LibreMenuEditor" or if I should leave it as "page.codeberg.libre_menu_editor.LibreMenuEditor".

page.codeberg.libre_menu_editor.LibreMenuEditor.yml
libre-menu-editor: I updated the manifest. It should be acceptable now

PR #4315, opened 2023-07-08
Add net.longturn.freeciv21

jwrober:


flatpak.yaml
```diff
@@ -0,0 +1,70 @@
...
+#   - flatpak-builder --install-deps-from=flathub --force-clean build/flatpak dist/flatpak.yaml
+#   - flatpak-builder --user --install --force-clean build/flatpak dist/flatpak.yaml
+#   - flatpak run net.longturn.freeciv21
+#   - flatpak remove net.longturn.freeciv21
+#
+##########
+
+app-id: net.longturn.freeciv21
+runtime: org.kde.Platform
+runtime-version: '5.15'
```
hfiguiere, line 19: this runtime is EOL

flatpak.yaml
```diff
@@ -0,0 +1,70 @@
...
+        dest: include/lua5.3
+
+      # Get Libertinus font because the build system does not have access to internet.
+      - type: archive
+        url: https://github.com/alerque/libertinus/releases/download/v7.040/Libertinus-7.040.zip
+        sha256: 2cce08507441d8ae7b835cfe51fb643ad5d9f6b44db4360c4e244f0e474a72f6
+        dest: fonts/Libertinus
+
+      - type: git
+        branch: stable
```
hfiguiere, line 39: never a branch. you need a tag or commit

flatpak.yaml
```diff
@@ -0,0 +1,70 @@
...
+        -DLUA_INCLUDE_DIR=include/lua5.3/src
+        -DLUA_LIBRARY=include/lua5.3/src/liblua.a
+        -DFREECIV_DOWNLOAD_FONTS=OFF
+      # CMake build with Install
+      - cmake --build build-flatpak --target install
+      # Incorporate the Libertinus Font
+      - mkdir -p /app/share/freeciv21/fonts/Libertinus
+      - cp -R fonts/Libertinus/* /app/share/freeciv21/fonts/Libertinus
+
+finish-args:
```
hfiguiere, line 59: finish-args is usually above modules

flatpak.yaml
```diff
@@ -0,0 +1,70 @@
...
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc
+  - --share=network
+  - --socket=pulseaudio
+  - --filesystem=home
+  - --filesystem=~/.local/share/freeciv21:create
```
hfiguiere, line 66: instead if should just persist instead of granted broad filesystem permissions.

flatpak.yaml
```diff
@@ -0,0 +1,70 @@
...
+      - type: git
+        branch: stable
+        url: https://github.com/longturn/freeciv21
+
+    buildsystem: simple
+    build-commands:
+      # Compile Lua
+      - cd include/lua5.3 && make linux
+      # Configure
+      - cmake -S $PWD -B build-flatpak -G Ninja
```
hfiguiere, line 47: in general using `buildsystem: cmake-ninja` is preferred. And you'd build lua prior to that, separately.

flatpak.yaml
```diff
@@ -0,0 +1,70 @@
...
+runtime: org.kde.Platform
+runtime-version: '5.15'
+sdk: org.kde.Sdk
+
+command: freeciv21-client
+modules:
+  - name: freeciv21
+    sources:
+      # Get Lua 5.3 as a dependency not included in the org.kde.Platform package.
+      - type: archive
```
hfiguiere, line 27: this should be a separate module

flatpak.yaml
jwrober: Thanks so much for the quick review. I'll have a look at these. Is it OK to push more commits to the branch?

jwrober: Updated to same runtime as the KDE apps use.

flatpak.yaml
jwrober: Set the proper git tag

flatpak.yaml
jwrober: Moved to top similar to how KDE apps are done

flatpak.yaml
jwrober: Updated

flatpak.yaml
jwrober: Moved Lua and Libertinus to their own modules

flatpak.yaml
jwrober: changed to cmake-ninja

hfiguiere:
bot, build net.longturn.freeciv21

net.longturn.freeciv21.yaml
```diff
@@ -0,0 +1,74 @@
...
+runtime-version: '5.15-22.08'
+sdk: org.kde.Sdk
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc
+  - --share=network
+  - --socket=pulseaudio
+  - --filesystem=home
+  - --filesystem=~/.local/share/freeciv21
```
hfiguiere, line 28: these broad filesystem permssions are redundant.

Also are you sure `~/.local/share/freeciv21` is used inside the sandbox? Standard Qt API do respect XDG.

net.longturn.freeciv21.yaml
```diff
@@ -0,0 +1,74 @@
...
+  # Get Lua 5.3 as a dependency not included in the org.kde.Platform package.
+  - name: Lua
+    sources:
+      - type: archive
+        url: https://www.lua.org/ftp/lua-5.3.6.tar.gz
+        sha256: fc5fd69bb8736323f026672b1b7235da613d7177e72558893a0bdcd320466d60
+        dest: include/lua5.3
+    buildsystem: simple
+    build-commands:
+      - cd include/lua5.3 && make linux
```
hfiguiere, line 41:
```suggestion
    buildsystem: simple
    build-commands:
      - make linux
```

net.longturn.freeciv21.yaml
```diff
@@ -0,0 +1,74 @@
...
+  - name: Libertinus
+    sources:
+      - type: archive
+        url: https://github.com/alerque/libertinus/releases/download/v7.040/Libertinus-7.040.zip
+        sha256: 2cce08507441d8ae7b835cfe51fb643ad5d9f6b44db4360c4e244f0e474a72f6
+        dest: fonts/Libertinus
+    buildsystem: simple
+    build-commands:
+      - mkdir -p /app/share/freeciv21/fonts/Libertinus
+      - cp -r fonts/Libertinus/* /app/share/freeciv21/fonts/Libertinus
```
hfiguiere, line 54:
```suggestion
    buildsystem: simple
    build-commands:
      - mkdir -p /app/share/freeciv21/fonts/Libertinus
      - * /app/share/freeciv21/fonts/Libertinus
```

net.longturn.freeciv21.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+  # Now build and package freeciv21
+  - name: freeciv21
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_INSTALL_PREFIX=/app
```
hfiguiere, line 60:
```suggestion
```

This is set automatically

net.longturn.freeciv21.yaml
```diff
@@ -0,0 +1,74 @@
...
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DFREECIV_DOWNLOAD_FONTS=OFF
+      - -DLUA_INCLUDE_DIR=/app/include/lua5.3/src
```
hfiguiere, line 63: are you sure this exist?

net.longturn.freeciv21.yaml
```diff
@@ -0,0 +1,74 @@
...
+      - -DLUA_LIBRARY=/app/include/lua5.3/src/liblua.a
+    cleanup:
+      - /app/include
```
hfiguiere, line 66:
```suggestion
      - /include
```

hfiguiere: but I don't think it'll have any effect on the lua headers.

hfiguiere:
build failed because it can't find the appstream file.

net.longturn.freeciv21.yaml
hfiguiere: I see, this is fine.

Quite unusual.

net.longturn.freeciv21.yaml
```diff
@@ -0,0 +1,74 @@
...
+  - --share=ipc
+  - --share=network
+  - --socket=pulseaudio
+  - --filesystem=home
+  - --filesystem=~/.local/share/freeciv21
+
+command: freeciv21-client
+modules:
+  # Get Lua 5.3 as a dependency not included in the org.kde.Platform package.
+  - name: Lua
```
hfiguiere, line 33:
```suggestion
  - name: Lua
  - cleanup:
      - '*'
```

I think yu want to remove it all.

net.longturn.freeciv21.yaml
hfiguiere: this will copy also some PDF and other uneeded files. Either copy less or add a cleanup rule.

hfiguiere:
None have the right name
```
-- Installing: /app/share/metainfo/net.longturn.freeciv21.client.metainfo.xml
-- Installing: /app/share/applications/net.longturn.freeciv21.client.desktop
-- Installing: /app/share/metainfo/net.longturn.freeciv21.server.metainfo.xml
-- Installing: /app/share/applications/net.longturn.freeciv21.server.desktop
-- Installing: /app/share/metainfo/net.longturn.freeciv21.modpack.metainfo.xml
-- Installing: /app/share/applications/net.longturn.freeciv21.modpack-qt.desktop
-- Installing: /app/share/metainfo/net.longturn.freeciv21.ruledit.metainfo.xml
-- Installing: /app/share/applications/net.longturn.freeciv21.ruledit.desktop
```

hfiguiere:
desktop file as well and I can't find icons

jwrober:
> build failed because it can't find the appstream file.

Well, some progress! This is my first flatpak, so appreciate the updates. The desktop and appstream files get installed during `cmake-ninja`
```
-- Installing: /app/share/metainfo/net.longturn.freeciv21.client.metainfo.xml
-- Installing: /app/share/applications/net.longturn.freeciv21.client.desktop
-- Installing: /app/share/metainfo/net.longturn.freeciv21.server.metainfo.xml
-- Installing: /app/share/applications/net.longturn.freeciv21.server.desktop
-- Installing: /app/share/metainfo/net.longturn.freeciv21.modpack.metainfo.xml
-- Installing: /app/share/applications/net.longturn.freeciv21.modpack-qt.desktop
-- Installing: /app/share/metainfo/net.longturn.freeciv21.ruledit.metainfo.xml
-- Installing: /app/share/applications/net.longturn.freeciv21.ruledit.desktop
```

and the desktop files are processed
```
Finishing app
Exporting share/applications/net.longturn.freeciv21.modpack-qt.desktop
Exporting share/applications/net.longturn.freeciv21.server.desktop
Exporting share/applications/net.longturn.freeciv21.client.desktop
Exporting share/applications/net.longturn.freeciv21.ruledit.desktop
Please review the exported files and the metadata
```

and it installs and runs great on my local. Any thoughts on what I need to change?

net.longturn.freeciv21.yaml
jwrober: Let me try on my local. I may not need it.

net.longturn.freeciv21.yaml
jwrober: I was really having a hard time getting our build system to find Lua so this what the work around to make it work. If you have better suggestion(s), I am all ears.  We use standard cmake `find_package()`
```
find_package(Lua 5.3 REQUIRED)
```

net.longturn.freeciv21.yaml
jwrober: As you have seen, I copy the headers to `/app` so they persist between module builds. Then I was thinking this would remove it after freeciv21 module is finished as we don't need them any longer once Lua is statically linked to the game.

jwrober:
> desktop file as well and I can't find icons

in net.longturn.freeciv21.client.desktop... `Icon=/app/share/freeciv21/misc/civicon.png` and the file in in there.

lmoureaux:
It's doing `stat builddir/*/share/app-info/xmls/net.longturn.freeciv21.xml.gz` and failing because the file doesn't exist. Looks like we want to rename the client file (is there a way to package multiple apps in a single flatpak? we have 3 tightly coupled user-facing applications and a console app...)

net.longturn.freeciv21.yaml
jwrober: ok works fine on my local without. will remove

net.longturn.freeciv21.yaml
jwrober: yea IDK why, but you have to do the combo command to compile it. Thanks

net.longturn.freeciv21.yaml
jwrober: In all of our other packages, we include the entirety of the font package. Can look at reducing what we keep for flathub.

net.longturn.freeciv21.yaml
jwrober: Ah, cool. Will remove.

net.longturn.freeciv21.yaml
jwrober: Thanks. This does remove it.

net.longturn.freeciv21.yaml
hfiguiere: cleanup rules are always "from FLATPAK_DEST" which is in that case `/app` and also only applies to whatever the module installs. Except the top-level one.

net.longturn.freeciv21.yaml
hfiguiere: Upstream lua doesn't allow that. Most distro patch it, and we have a shared-module to build it that way

It's the recommended way.

net.longturn.freeciv21.yaml
hfiguiere: nobody will ever read the PDF. It's wasted space.

hfiguiere:
> > desktop file as well and I can't find icons
> 
> in `net.longturn.freeciv21.client.desktop`... `Icon=/app/share/freeciv21/misc/civicon.png` and the file in in there.

1. icon should be  a name. That's not specific to flatpak. Using a full path is not a practice of this decade. Nor the previous.
2. only icons whose base name is the resolution of `$FLATPAK_ID` are exported by flatpak. If it's not exported to won't be found.

hfiguiere:
> It's doing `stat builddir/*/share/app-info/xmls/net.longturn.freeciv21.xml.gz` and failing because the file doesn't exist. Looks like we want to rename the client file (is there a way to package multiple apps in a single flatpak? we have 3 tightly coupled user-facing applications and a console app...)

Ignore that path, it's the produce of appstream-compose, that failed. 

Just install an appstream file with the basename that is the resolution of `$FLATPAK_ID`. (it should have the right id in the XML too). `/app/share/metainfo/net.longturn.freeciv21.metainfo.xml`

hfiguiere:
same for the desktop file and the icons.

hfiguiere:
> and it installs and runs great on my local. Any thoughts on what I need to change?

Buildbot does more checks. Notably it validate the appstream file and check the appstram-compose output because all of this make sure apps are installed properly.

net.longturn.freeciv21.yaml
barthalion: Yeah, this won't work unless changed as Hubert suggested.

barthalion:
That cryptic error means some mismatch between icon, appdata/metainfo and desktop files. Appdata/metainfo refer to the desktop file with `launchable`, then the desktop file refers the icon file. All need to exist and follow the app ID naming scheme; also some take extensions as value (launchable in appdata) and others don't (icon filename in the desktop file). The best is to start from the appdata file and go down the path.

hfiguiere:
> That cryptic error means some mismatch between icon, appdata/metainfo and desktop files. Appdata/metainfo refer to the desktop file with `launchable`, then the desktop file refers the icon file. All need to exist and follow the app ID naming scheme; also some take extensions as value (launchable in appdata) and others don't (icon filename in the desktop file). The best is to start from the appdata file and go down the path.

As I mentionned is because there is no `net.longturn.freeciv21.metainfo.xml` installed. not a matching .desktop

also I foresee that the icon check will fail next if not tackled.

net.longturn.freeciv21.yaml
```diff
@@ -0,0 +1,89 @@
...
+      - -DLUA_LIBRARY=/app/include/lua5.3/src/liblua.a
+    cleanup:
+      - /include
+    sources:
+      - type: git
+        tag: v3.0-patch.1
+        url: https://github.com/longturn/freeciv21
+
+  # Finalize
+  - name: Finalize
```
hfiguiere, line 73: this doesn't need to be a separate module. it's usually part of the main one.

net.longturn.freeciv21.yaml
```diff
@@ -0,0 +1,89 @@
...
+    buildsystem: simple
+    build-commands:
+      - mkdir -p ${FLATPAK_DEST}/share/icons/hicolor/64x64/apps
+      - cp ${FLATPAK_DEST}/share/freeciv21/misc/civicon.png ${FLATPAK_DEST}/share/icons/hicolor/64x64/apps/${FLATPAK_ID}.png
+      - cp ${FLATPAK_DEST}/share/freeciv21/misc/mpicon.png ${FLATPAK_DEST}/share/icons/hicolor/64x64/apps/net.longturn.freeciv21.modpack-qt.png
+      - mv ${FLATPAK_DEST}/share/applications/net.longturn.freeciv21.client.desktop ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop
+      - sed -i 's/\/app\/share\/freeciv21\/misc\/civicon/net.longturn.freeciv21/' ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop
```
hfiguiere, line 80: please use `destkop-file-edit` to modify desktop file.

net.longturn.freeciv21.yaml
```diff
@@ -0,0 +1,89 @@
...
+        tag: v3.0-patch.1
+        url: https://github.com/longturn/freeciv21
+
+  # Finalize
+  - name: Finalize
+    buildsystem: simple
+    build-commands:
+      - mkdir -p ${FLATPAK_DEST}/share/icons/hicolor/64x64/apps
+      - cp ${FLATPAK_DEST}/share/freeciv21/misc/civicon.png ${FLATPAK_DEST}/share/icons/hicolor/64x64/apps/${FLATPAK_ID}.png
+      - cp ${FLATPAK_DEST}/share/freeciv21/misc/mpicon.png ${FLATPAK_DEST}/share/icons/hicolor/64x64/apps/net.longturn.freeciv21.modpack-qt.png
```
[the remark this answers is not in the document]
hfiguiere: btw the minimum size for icons is 128x128. So this might fail.

net.longturn.freeciv21.yaml
```diff
@@ -0,0 +1,89 @@
...
+      - mv ${FLATPAK_DEST}/share/applications/net.longturn.freeciv21.client.desktop ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop
```
hfiguiere, line 79: you should just use `rename-desktop-file` at the top-level of the manifest.

net.longturn.freeciv21.yaml
```diff
@@ -0,0 +1,89 @@
...
+      - sed -i 's/\/app\/share\/freeciv21\/misc\/civicon/net.longturn.freeciv21/' ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop
+      - sed -i 's/\/app\/share\/freeciv21\/misc\/mpicon/net.longturn.freeciv21.modpack-qt/' ${FLATPAK_DEST}/share/applications/net.longturn.freeciv21.modpack-qt.desktop
+      - sed -i 's/\/app\/share\/freeciv21\/misc\/server/net.longturn.freeciv21/' ${FLATPAK_DEST}/share/applications/net.longturn.freeciv21.server.desktop
+      - sed -i 's/\/app\/share\/freeciv21\/misc\/civicon/net.longturn.freeciv21/' ${FLATPAK_DEST}/share/applications/net.longturn.freeciv21.ruledit.desktop
+      - mv ${FLATPAK_DEST}/share/metainfo/net.longturn.freeciv21.client.metainfo.xml ${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.metainfo.xml
```
hfiguiere, line 84: you should just use `rename-appdata-file` at the top level of the manifest

net.longturn.freeciv21.yaml
```diff
@@ -0,0 +1,89 @@
...
+      - sed -i 's/net.longturn.freeciv21.client/net.longturn.freeciv21/' ${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.metainfo.xml
```
hfiguiere, line 85: please 
```suggestion
      - appstream-util modify $FLATPAK_DEST/share/metainfo/net.longturn.freeciv21.client.metainfo.xml id ${FLATPAK_ID}",
```

this granted you use `rename-appdata-file` as mentionned above

net.longturn.freeciv21.yaml
hfiguiere: same for the lines below.

hfiguiere: btw sharing the same icon for the different desktop entries is bad UX (it's not blocking though)

hfiguiere:
bot, build net.longturn.freeciv21

hfiguiere:
The appstream file fail to validate:

```
```

hfiguiere:
also a warning down the way: 
```
WARNING: Icon referenced in desktop file but not exported: net.longturn.freeciv21.png
```

but my suggestions make all of this more readable and easier to read.

lmoureaux:
> The appstream file fail to validate:
> 
> ```
> • tag-missing           : <release> required
> ```

The release tag is optional in the spec and using `appstream-util validate-relax` [as documented](https://docs.flatpak.org/en/latest/freedesktop-quick-reference.html?highlight=appstream#appdata-files) doesn't complain about this. Are the requirements added by Flathub on top of the appstream spec documented anywhere?

hfiguiere:
From the documentation that has been acknowledged:
    https://docs.flathub.org/docs/for-app-authors/requirements#appstream

image

it also link to another page.
   https://docs.flathub.org/docs/for-app-authors/appdata-guidelines/

As you are trying to argue the spec say it's optional, but the spec is NOT flathub requirements. The spec is not even flatpak specific.

lmoureaux:
Thanks for pointing me to your these links, I stand corrected: there is some documentation in the code of the checker tool.

jwrober:
I read the docs when I opened the PR. I must have missed those particulars. I'll keep poking along and hopefully will get this through validation. Thanks for all the suggestions @hfiguiere

net.longturn.freeciv21.yaml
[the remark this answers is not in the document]
jwrober: getting 128x128 from our repo

net.longturn.freeciv21.yaml
jwrober: That is a cool utility. Yes agree bad UX using same icon. Maybe someone nice will make me a nice icon someday.

net.longturn.freeciv21.yaml
jwrober: `appstream-util` is also a cool tool.

net.longturn.freeciv21.yaml
jwrober: thanks, that works a lot better

hfiguiere:
bot, build net.longturn.freeciv21

net.longturn.freeciv21.yaml
```diff
@@ -0,0 +1,93 @@
...
+  - --filesystem=home
+  - --persist=.local/share/freeciv21
+  - --persist=.config/freeciv21
+
+rename-desktop-file: net.longturn.freeciv21.client.desktop
+rename-appdata-file: net.longturn.freeciv21.client.metainfo.xml
+
+command: freeciv21-client
+branch: stable
+default-branch: stable
```
hfiguiere, line 36:
```suggestion
```

net.longturn.freeciv21.yaml
```diff
@@ -0,0 +1,93 @@
...
+    post-install:
+      - install -D freeciv21-client.png ${FLATPAK_DEST}/share/icons/hicolor/128x128/apps/${FLATPAK_ID}.png
+      - install freeciv21-server.png ${FLATPAK_DEST}/share/icons/hicolor/128x128/apps/${FLATPAK_ID}.server.png
+      - install freeciv21-modpack.png ${FLATPAK_DEST}/share/icons/hicolor/128x128/apps/${FLATPAK_ID}.modpack.png
+      - desktop-file-edit --set-icon=${FLATPAK_ID} ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.client.desktop
+      - desktop-file-edit --set-icon=${FLATPAK_ID}.server ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.server.desktop
+      - desktop-file-edit --set-icon=${FLATPAK_ID}.modpack ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.modpack-qt.desktop
+      - desktop-file-edit --set-icon=${FLATPAK_ID} ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.ruledit.desktop
+      - appstream-util modify ${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.client.metainfo.xml id ${FLATPAK_ID}
+      - appstream-util modify ${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.client.metainfo.xml launchable ${FLATPAK_ID}.desktop
```
hfiguiere, line 72: `rename-destkop-file` takes care of this.

```suggestion
```

net.longturn.freeciv21.yaml
```diff
@@ -0,0 +1,93 @@
...
+      - appstream-util modify ${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.client.metainfo.xml release "3.0-stable"
```
hfiguiere, line 73: as I suspected this doesn't work.

jwrober: hm, there is a valid release tag in the file. What would that not validate?
```
<content_rating type="oars-1.0">
    <content_attribute id="social-chat">mild</content_attribute>
  </content_rating>
  <release>3.0-stable</release>
</component>
```

jwrober: Ok, I read the spec in more detail. I see what the issue is now.

net.longturn.freeciv21.yaml
jwrober: removed

net.longturn.freeciv21.yaml
jwrober: fixed

jwrober:
Ok I think its ready. Hopefully bot will like it this time :)

hfiguiere:
you can aslo:

bot, build net.longturn.freeciv21

jwrober:
Thanks @hfiguiere. I wasn't sure if I could call the bot. Good to know for the future. Looks like the test package works well for me. What are the next steps?

PR #4333, opened 2023-07-17
Write comment when not targeting the `new-pr` branch

JakobDev:
As there a lot of of PR's accidental targeting the wrong branch. I think it's a good idea to 
automatically add a comment, if that happens.


bbhtt:
I don't think this is needed, we still merge even on the wrong branch. The only problem is that they inherit the git history and infra files from master.

PR #4364, opened 2023-07-26
Add io.kapsa.drive

tomekit:
Hi Flathub,

I would like to publish S3Drive: https://s3drive.app on Flathub. It's the first time we publish app, so please forgive us if our submission isn't 100% complete at this stage.

~~Currently we're having an issue with dependency being `libmpv` library, which despite being included in the `modules` for some reason is not picked by the app which fails with: `kapsa: error while loading shared libraries: libmpv.so.1: cannot open shared object file: No such file or directory`.
By starting the submission and preparing this public repo we can get some help and hopefully resolve it in the meantime.~~ 
UPDATE: This is now resolved

Thanks


io.kapsa.drive.metainfo.xml
```diff
@@ -0,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+    <id>io.kapsa.drive</id>
+    <name>S3Drive</name>
+    <summary>Zero Knowledge E2E encrypted storage compatible with any S3 provider</summary>
+    <developer_name>S3Drive</developer_name>
+    <url type="homepage">https://s3drive.app</url>
```
hfiguiere, line 7: why isn't the app id `app.s3drive.drive` ?

There is nothing at `kapsa.io`

S3Drive-Linux-Portable.tar.gz
hfiguiere, line 1: this file DO NOT BELONG to the repository.

io.kapsa.drive.json
```diff
@@ -0,0 +1,338 @@
...
+        "mkdir -p S3Drive",
+        "tar -xf S3Drive-Linux-Portable.tar.gz -C S3Drive",
+        "cp -r S3Drive /app/",
+        "chmod +x /app/S3Drive/kapsa",
+        "mkdir -p /app/bin",
+        "ln -s /app/S3Drive/kapsa /app/bin/kapsa",
+        "mkdir -p /app/share/metainfo",
+        "cp -r io.kapsa.drive.metainfo.xml /app/share/metainfo/",
+        "mkdir -p /app/share/icons/hicolor/scalable/apps",
+        "cp -r icon_flatpak.png /app/share/icons/hicolor/scalable/apps/io.kapsa.drive.png",
```
hfiguiere, line 312: since when a PNG is scalable?

io.kapsa.drive.json
```diff
@@ -0,0 +1,338 @@
...
+      ],
+      "post-install": [
+        "install -Dm644 LICENSE ${FLATPAK_DEST}/share/licenses/libsodium/LICENSE"
+      ]
+    },
+    {
+      "name": "s3drive",
+      "buildsystem": "simple",
+      "only-arches": [
+        "x86_64"
```
hfiguiere, line 300: this doesn't prevent an aarch64 flatpak to be built.

io.kapsa.drive.json
```diff
@@ -0,0 +1,338 @@
+{
+  "app-id": "io.kapsa.drive",
+  "runtime": "org.freedesktop.Platform",
+  "runtime-version": "22.08",
+  "sdk": "org.freedesktop.Sdk",
+  "sdk-extensions": [
+    "org.freedesktop.Sdk.Extension.vala"
```
hfiguiere, line 7: why do you need vala?

icon_flatpak.svg
hfiguiere, line 1: this file is not a SVG. It's a PNG. Changing the extension doesn't change anything.

io.kapsa.drive.flatpak
hfiguiere, line 1: What is that file and why is it here?

io.kapsa.drive.metainfo.xml
hfiguiere: Please change the app id to comply with the requirements. This is not a suggestion.

io.kapsa.drive.json
```diff
@@ -0,0 +1,341 @@
...
+          "path": "io.kapsa.drive.desktop"
+        },
+        {
+          "type": "file",
+          "path": "icon_flatpak.svg"
+        },
+        {
+          "type": "archive",
+          "only-arches": [
+            "x86_64"
```
hfiguiere, line 332: this still doesn't prevent from building an aarch64 package.

io.kapsa.drive.metainfo.xml
```diff
@@ -0,0 +1,71 @@
...
+    <metadata_license>CC0</metadata_license>
+    <project_license>Proprietary</project_license>
```
hfiguiere, line 9: is redistribution allowed?

io.kapsa.drive.metainfo.xml
tomekit: If you try https://drive.kapsa.io (which is the exact inverse of package ID) there is a redirection to: https://s3drive.app.
It's a future proofing, as at some point we'll be changing the main domain.
Would additional redirection from `kapsa.io` (as a proof of ownership) be sufficient?

S3Drive-Linux-Portable.tar.gz
tomekit: It's been moved as external archive: https://github.com/s3drive/flathub/blob/s3drive/io.kapsa.drive.json#L330

io.kapsa.drive.json
tomekit: It's been replaced with SVG: https://github.com/s3drive/flathub/blob/s3drive/io.kapsa.drive.json#L310

io.kapsa.drive.json
tomekit: Sorry, I am not exactly experienced with that.
What would be your suggestion? 
As such we haven't tested it on ARM and we would like to limit platforms to standard 32 / 64-bit PC.

io.kapsa.drive.json
tomekit: If we don't specify that the: `ayatana-ido` build fails with:
```
CMake Error at /usr/share/cmake-3.26/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Vala (missing: VALA_COMPILER)
Call Stack (most recent call first):
  /usr/share/cmake-3.26/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindVala.cmake:52 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  src/CMakeLists.txt:164 (find_package)
```

I am not sure if there are better ways to overcome this issue.

io.kapsa.drive.json
hfiguiere: This is a lie:
```
icon_flatpak.svg: PNG image data, 512 x 512, 8-bit/color RGBA, non-interlaced
```

io.kapsa.drive.json
hfiguiere: from the documentation you ackowledge having read:
https://docs.flathub.org/docs/for-app-authors/maintanance/#limiting-the-set-of-architectures-to-build-on

io.kapsa.drive.json
hfiguiere: a CMake option to disable it. I don't know the package so I can't tell. Other have the option.

io.kapsa.drive.metainfo.xml
hfiguiere: > If you try https://drive.kapsa.io (which is the exact inverse of package ID) there is a redirection to: https://s3drive.app.

The last component is the name of the app, not a hostname, so that's why it is confusing.

io.kapsa.drive.desktop
hfiguiere, line 1: so you mean you ship a desktop application without a desktop file nor icon? This is a bug.

all three files should be part of the upstream distribution assets.

icon_flatpak.svg
hfiguiere, line 1: there is a `./data/flutter_assets/assets/logos/logo.svg` in the tarball. why not using it instead?

tomekit:
Hi @hfiguiere,
Thank you for your prompt response reviewing my submission. I am going to address all of the remaining issues and let you know once I am done.

icon_flatpak.svg
tomekit: It's auto-generated copy over from Flutter and it's not something we would like to rely on.

io.kapsa.drive.json
hfiguiere: apparently there is no choice.

icon_flatpak.svg
hfiguiere: but policies about flathub are to minimize what's in the repository. Since you are upstream, you can fix this.

Icon and desktop file are part of a 20+ year old standard for desktop applications on Linux (et al.) so when they are missing upstream it's a bug.

io.kapsa.drive.metainfo.xml
tomekit: > Please change the app id to comply with the requirements. This is not a suggestion.

Can you please provide these requirements?

Based on: https://github.com/flathub/flathub/wiki/App-Requirements

```
Each application should have a unique application ID, following the standard reverse-DNS schema. See [the Flatpak documentation](http://docs.flatpak.org/en/latest/conventions.html#application-ids) for more information on this. The Application ID should be a real URL of a domain that the app author has control over or where their app is hosted.
```
Our Application ID is following reverse-DNS schema and is a  real URL of a domain that the app author has control over.

Are there any other requirements that we're not aware?

icon_flatpak.svg
tomekit: Can you please link to relevant Flathub policy?

Are you going to reject this based on additional 4.83KB? ... which will then likely be compressed down to almost zero, since the additional file (which location is not guaranteed) that's added by the Flutter will be the 1:1 copy.

tomekit: Well, look. I will have it fixed if we can. I am not sure about the consequences, after we apply the `extra-data` to archive which contain that icon.

hfiguiere: if you use extra-data it's a different story.

hfiguiere: > Can you please link to relevant Flathub policy?

https://docs.flathub.org/docs/for-app-authors/requirements#acceptable-but-should-be-submitted-upstream

 
> Are you going to reject this based on additional 4.83KB? ... 

I have no power to reject anything.

But if you use `extra-data` then the point is moot.

barthalion (commented):
You need to add a separate file named `flathub.json` limiting the build to x86_64.

```
{
  "only-arches": ["x86_64"]
}
```

tomekit:
I would appreciate if you could review our submission again.

Changes: 
- Switched to `extra-data` to inject .deb package during installation,
- Added the: `flathub.json` file to limit architecture,
- Added "anityta" URL checker,
- Fixed app logo,
- Resolved issue with MPV dependency,

Please let me know if you're happy with the: current `io.kapsa.drive` ID or would prefer to change it to either uppercase: `io.kapsa.Drive` or `io.kapsa.S3Drive`.

io.kapsa.drive.json
```diff
@@ -0,0 +1,365 @@
+{
+  "app-id": "io.kapsa.drive",
+  "runtime": "org.freedesktop.Platform",
+  "runtime-version": "22.08",
```
hfiguiere, line 4: current runtime is now 23.08

tomekit: We've tried using `23.08`, but we can't manage to compile the `ffmpeg` dependency due to: `./libavcodec/x86/mathops.h:125: Error: operand type mismatch for `shr'` likely related: https://www.linuxquestions.org/questions/slackware-14/regression-on-current-with-ffmpeg-4175727691/

Once we've solved this issue and potentially any others that may arise after upgrading to fresh 23.08 we would be using 22.08.

hfiguiere:
> > Build [64831 failed](https://buildbot.flathub.org/#/builders/6/builds/64831)
> 
> We're loading: ` "shared-modules/intltool/intltool-0.51.json",` in our manifest, in order to satisfy this dependency we've had to had to include shared modules, on local:
> 
> ```
> git submodule add https://github.com/flathub/shared-modules.git
> ```
> 
> Not exactly sure how to add them to pipeline.
> 
> Shared modules: https://docs.flathub.org/docs/for-app-authors/shared-modules

Like any other git submodule. 

https://git-scm.com/book/en/v2/Git-Tools-Submodules

io.kapsa.drive.json
```diff
@@ -0,0 +1,365 @@
...
+          "type": "file",
+          "path": "s3drive_logo.png"
+        },
+        {
+          "type": "extra-data",
+          "filename": "s3drive_amd64.deb",
+          "only-arches": [
+            "x86_64"
+          ],
+          "url": "https://github.com/s3drive/app/releases/latest/download/s3drive_amd64.deb",
```
hfiguiere, line 353: this unversioned URL for downloading the software will cause a lot of grief.

io.kapsa.drive.json
```diff
@@ -0,0 +1,365 @@
...
+          "sha256": "d0314a54d277c2dfa8ab11c0d9dd63aaab4769d6fff6028842ebf1a817ba9782",
+          "size": 10851828,
+          "x-checker-data": {
+            "type": "anitya",
+            "project-id": 369108,
+            "url-template": "https://github.com/s3drive/app/releases/download/$version/s3drive_amd64.deb"
```
hfiguiere, line 359: so it's $version here and `latest` above?

io.kapsa.drive.metainfo.xml
```diff
@@ -0,0 +1,71 @@
...
+        <p>Our roadmap: https://s3drive.canny.io/</p>
+        <p>Order by (name, size, modified),</p>
+        <p>Support streaming of encrypted videos,</p>
+        <p>Cross-client online sync,</p>
+        <p>Listing cache,</p>
+        <p>... and many many more, vote in our roadmap and stay tuned !</p>
+
+        <p>We would love to hear from you !</p>
+        <p>Missing some features? App not working as intended?</p>
+        <p>Please visit our Discord: https://discord.com/invite/YvxvypRqCu</p>
```
hfiguiere, line 58: don't put URL in descriptions. You can use the `<url>` tags for that.

io.kapsa.drive.json
tomekit: Sorry, I think I don't exactly understand it all exactly. The latest versioned URL is: https://github.com/s3drive/app/releases/download/1.5.1/s3drive_amd64.deb

Does it mean I will have to updated it each new release?

tomekit: Changed to:  https://github.com/s3drive/app/releases/download/1.5.1/s3drive_amd64.deb

io.kapsa.drive.metainfo.xml
tomekit: Moved out to: `<url type="contact">https://discord.com/invite/YvxvypRqCu</url>`

io.kapsa.drive.metainfo.xml
hfiguiere: this is an invalid SPDX license. It at least `LicenseRef-proprietary`

tomekit:
I've added `shared-modules` as Git submodule which shall hopefully resolve the pipeline issue.

io.kapsa.drive.json
tomekit: Hi @barthalion,
I've noticed that you've changed the `runtime-version` to `23.08`: https://github.com/s3drive/flathub/commit/57c087867ffedd821cc934fca771247b957786b4
I've failed so far to compile dependencies, we're having issue at least with `ffmpeg`, can we keep it: `22.08` and aim to update to `23.08` over the next couple months?

https://github.com/flathub/flathub/pull/4364#discussion_r1321336267

hfiguiere: There is that upstream patch you can find by Googlling:

https://github.com/flathub/org.ardour.Ardour/blob/master/patches/ffmpeg-as-fix.patch

tomekit: Thanks for linking the fix. I've applied it and it resolved the `ffmpeg` compile issue at least, but now I am getting: 
```
ar: error while loading shared libraries: libsframe.so.1: cannot open shared object file: No such file or directory
```
during installation.
It seems it fails on: `ar x s3drive_amd64.deb` however when I shell into:
```
flatpak-builder --build-shell=s3drive build-dir io.kapsa.drive.json
```
and run the: `ar x s3drive_amd64.deb` after fetching the .deb first, it seem to work.

These issues does not appear on: `22.08`.

tomekit:
bot, build io.kapsa.drive

tomekit:
bot, build io.kapsa.drive

tomekit:
Even though we've resolved ffmpeg compile issues (https://github.com/flathub/flathub/pull/4364#discussion_r1323753731)
I've reverted back from `23.08` and `22.08`, since we're facing another issue on the installation level: https://github.com/flathub/flathub/pull/4364#discussion_r1330949164

That's a long way before even app starts.

We don't want to rush: `23.08` and are perfectly fine with `22.08`, so unless you have some objections, let's please stick to it.

I am sure that after a while when the dust settle, issues will resolve themselves or there will be reasonable knowledge base to resolve issues promptly.

Please find the recent successful build.

tomekit:
Just wanted to say thank you @barthalion for picking this and thank you @hfiguiere for your input on preparing this submission.

PR #4368, opened 2023-07-27
io.github.giantpinkrobots.flatsweep.json

giantpinkrobots:

  My app needs to be able to run a "flatpak list" command in order to get a list of all Flatpaks installed on the user's system.   Therefore there is a need of the org.freedesktop.Flatpak DBus interface.


io.github.giantpinkrobots.flatsweep.json
```diff
@@ -0,0 +1,44 @@
...
+                    "type" : "git",
+                    "url" : "https://github.com/giantpinkrobots/flatsweep.git",
+                    "commit" : "05136b675ffae7a67ff3221f90fdb8ac5d664351"
+                }
+            ]
+        }
+    ],
+    "build-options" : {
+        "env" : {        }
+    }
```
hfiguiere, line 43:
```suggestion
    ]
```

io.github.giantpinkrobots.flatsweep.json
```diff
@@ -0,0 +1,44 @@
...
+    "runtime-version" : "44",
+    "sdk" : "org.gnome.Sdk",
+    "command" : "flatsweep",
+    "finish-args" : [
+        "--share=network",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--device=dri",
+        "--socket=wayland",
+        "--talk-name=org.freedesktop.Flatpak",
```
hfiguiere, line 13: this is not allowed by default

https://docs.flathub.org/docs/for-app-authors/linter

giantpinkrobots: I added this line because I needed to get the list of installed Flatpaks on the system and the way to do that was to run a flatpak list command using flatpak-spawn. If that is not allowed I will try to get the list using some other way. But that is the reason why.

JakobDev: If you have a good reason, you can add a exception

barthalion: You could add RO access to /var/lib/flatpak/app and xdg-data/flatpak/app. The directories there are the IDs of currently installed apps.

JakobDev: This is not enough. You can add a [custom installation path](https://docs.flatpak.org/en/latest/tips-and-tricks.html#adding-a-custom-installation). This won't be done by 99% of the users, but who those who do it, it's better to have it working.

barthalion: I wouldn't particularly care about less than 1% of users using custom installation paths.

giantpinkrobots: Is there a way to check if the user has chosen an alternative install path so I could show an error screen? Without using commands.

giantpinkrobots:
I have altered the app so it requires less permissions. It should be good to go now.

giantpinkrobots:
bot, build io.github.giantpinkrobots.flatsweep

giantpinkrobots:
bot, build io.github.giantpinkrobots.flatsweep

PR #4401, opened 2023-08-08
Add org.kde.kommit

Mailaender:

- [X] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://invent.kde.org/teams/flathub/issues/-/issues/14

### DIFF
--- a/org.kde.kommit.json
+++ b/org.kde.kommit.json
@@ -1,0 +1,62 @@
+{
+    "id": "org.kde.kommit",
+    "runtime": "org.kde.Platform",
+    "runtime-version": "5.15-22.08",
+    "sdk": "org.kde.Sdk",
+    "command": "kommit",
+    "rename-icon": "kommit",
+    "finish-args": [
+        "--share=ipc",
+        "--share=network",
+        "--device=dri",
+        "--socket=ssh-auth",
+        "--socket=gpg-agent",
+        "--socket=wayland",
+        "--socket=fallback-x11",
+        "--filesystem=home",
+        "--system-talk-name=org.freedesktop.UDisks2"
+    ],
+    "modules": [
+        {
+            "name": "git",
+            "make-args": [
+                "INSTALL_SYMLINKS=1"
+            ],
+            "make-install-args": [
+                "INSTALL_SYMLINKS=1"
+            ],
+            "cleanup": [
+                "/lib/pkgconfig",
+                "/man"
+            ],
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://www.kernel.org/pub/software/scm/git/git-2.41.0.tar.xz",
+                    "sha256": "e748bafd424cfe80b212cbc6f1bbccc3a47d4862fb1eb7988877750478568040",
+                    "x-checker-data": {
+                        "type": "anitya",
+                        "project-id": 5350,
+                        "stable-only": true,
+                        "url-template": "https://www.kernel.org/pub/software/scm/git/git-$version.tar.xz"
+                    }
+                }
+            ]
+        },
+        {
+            "name": "kommit",
+            "buildsystem": "cmake-ninja",
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://download.kde.org/stable/kommit/kommit-1.0.2.tar.xz",
+                    "sha256": "1e8ecad23aa10508d0685a2e9169fc562e12fad5a6adaacd4542d9007a68e96e"
+                },
+                {
+                    "type": "patch",
+                    "path": "appdata.patch"
+                }
+            ]
+        }
+    ]
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+build-dir/
+.flatpak-builder/
--- a/appdata.patch
+++ b/appdata.patch
@@ -1,0 +1,65 @@
+From f2e07ce35d3302e5f9f7c2f0798df656407edd08 Mon Sep 17 00:00:00 2001
+From: Jonathan Esk-Riddell <jr@jriddell.org>
+Date: Thu, 4 May 2023 17:10:43 +0100
+Subject: [PATCH] add 1.0.2 release and tidy description as the app can be used
+ by anyone not just kde people
+
+---
+ src/data/org.kde.kommit.appdata.xml | 15 +++++++++++++--
+ src/data/org.kde.kommit.desktop     |  2 +-
+ 2 files changed, 14 insertions(+), 3 deletions(-)
+
+diff --git a/src/data/org.kde.kommit.appdata.xml b/src/data/org.kde.kommit.appdata.xml
+index 62617a66..7f81b72c 100644
+--- a/src/data/org.kde.kommit.appdata.xml
++++ b/src/data/org.kde.kommit.appdata.xml
+@@ -20,7 +20,7 @@
+   <name xml:lang="tr">Kommit</name>
+   <name xml:lang="uk">Kommit</name>
+   <name xml:lang="x-test">xxKommitxx</name>
+-  <summary> Git gui client for KDE </summary>
++  <summary>Git GUI client</summary>
+   <summary xml:lang="ca">Client gràfic del Git per al KDE</summary>
+   <summary xml:lang="ca-valencia">Client gràfic de Git per a KDE</summary>
+   <summary xml:lang="cs">GUI git klient od KDE</summary>
+@@ -44,7 +44,7 @@
+   <summary xml:lang="zh-CN">KDE git 图形界面客户端</summary>
+   <summary xml:lang="zh-TW">KDE 的 Git 用戶端介面</summary>
+   <description>
+-    <p> Git gui client for KDE </p>
++    <p>Git GUI client</p>
+     <p xml:lang="ca">Client gràfic del Git per al KDE</p>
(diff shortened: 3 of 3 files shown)


Mailaender:
Can't build right now, as https://release-monitoring.org/ is down.

hfiguiere:
cc @flathub/kde

org.kde.kommit.json
```diff
@@ -0,0 +1,129 @@
...
+                        "type": "anitya",
+                        "project-id": 4941,
+                        "stable-only": true,
+                        "url-template": "https://prdownloads.sourceforge.net/tcl/tcl$version-src.tar.gz"
+                    }
+                }
+            ]
+        },
+        {
+            "name": "tk",
```
hfiguiere, line 54: is a tk app intended to be run from inside this flatpak?

org.kde.kommit.json
```diff
@@ -0,0 +1,79 @@
...
+        "--share=ipc",
+        "--share=network",
+        "--device=dri",
+        "--socket=wayland",
+        "--socket=fallback-x11",
+        "--system-talk-name=org.freedesktop.UDisks2"
+    ],
+    "modules": [
+        {
+            "name": "kommit",
```
hfiguiere, line 18: this should be last in the modules list.

org.kde.kommit.json
```diff
@@ -0,0 +1,79 @@
...
+                    "url": "https://github.com/git-lfs/git-lfs/releases/download/v3.4.0/git-lfs-linux-amd64-v3.4.0.tar.gz",
+                    "sha256": "60b7e9b9b4bca04405af58a2cd5dff3e68a5607c5bc39ee88a5256dd7a07f58c",
+                    "x-checker-data": {
+                        "type": "json",
+                        "url": "https://api.github.com/repos/git-lfs/git-lfs/releases/latest",
+                        "url-query": ".assets[] | select(.name==\"git-lfs-linux-amd64-\" + $version + \".tar.gz\") | .browser_download_url",
+                        "version-query": ".tag_name"
+                    },
+                    "only-arches": [
+                        "x86_64"
```
hfiguiere, line 73: why not on aarch64?

org.kde.kommit.json
Mailaender: No, git is not a compile time dependency.

hfiguiere:
bot, build org.kde.kommit

hfiguiere:
```
builddir/files/share/appdata/org.kde.kommit.appdata.xml: FAILED:
Validation of files failed
```

Mailaender:
bot, build org.kde.kommit

org.kde.kommit.json
```diff
@@ -0,0 +1,97 @@
...
+                    ]
+                },
+                {
+                    "type": "archive",
+                    "url": "https://github.com/git-lfs/git-lfs/releases/download/v3.4.0/git-lfs-linux-arm64-v3.4.0.tar.gz",
+                    "sha256": "e97c477981a9b6a40026cadc1bf005541d973fc32df2de2f398643b15df6b5c6",
```
hfiguiere, line 83: apparently this checksum is incorrect

org.kde.kommit.json
hfiguiere: doesn't matter. it's the convention.

Mailaender:
bot, build org.kde.kommit

tsdgeos:
This is not very useful since it doesn't seem to be able to access my ssh keys

So any operation i try fails with 

git@gitlab.freedesktop.org: Permission denied (publickey).

Justinzobel:
Yeah I think it's because it can't access the ssh folder which I think is a good/bad thing. I saw the ssh error in konsole when running it and it stopped until I forced it to close.

hfiguiere:
Shouldn't this work with the secrets service DBus API?

Mailaender:
Added the SSH auth socket.

Mailaender:
bot, build org.kde.kommit

org.kde.kommit.json
travier: We still need this

travier:
We probably need to give it GPG access too as some people sign their commits that way.

Mailaender:
Added that as well.

Mailaender:
bot, build org.kde.kommit

travier:
https://github.com/flathub/flathub/pull/4401/files#r1291086950

Mailaender:
bot, build org.kde.kommit

tsdgeos:
> Shouldn't this work with the secrets service DBus API?

No such thing on my system.

tsdgeos:
I woud really prefer if we put this on beta.

For example when trying to clone from an unkonwn host, it will ask

The authenticity of host 'gitlab.freedesktop.org (147.75.198.156)' can't be established.
ED25519 key fingerprint is SHA256:i4LwjlmsgSrc3aWq3aVaDa8VQs6Fqq4nRBuFdZ6A1tE.
This key is not known by any other names.
Are you sure you want to continue connecting (yes/no/[fingerprint])?

on the command line, that-s a bug that exists in kommit upstream too, but this can be solved in kommit upstream by doing things from the git command line first and it will learn about the host, but that's something that you can't do in flatpak i think? 

Or is there a way to share the known_hosts file from the host to the flatpak app?

travier:
We could explicitly share the known host file in the Flatpak (as RW?).

Mailaender:
bot, build org.kde.kommit

hfiguiere:
did you test it? AFAIK `--filesystem` should be a *directory*

Mailaender:
No, sorry I switched computers from Arch to an Ubuntu LTS derivative where I get
> error: Unknown socket type gpg-agent, valid types are: x11, wayland, pulseaudio, session-bus, system-bus, fallback-x11, ssh-auth, pcsc, cups

when I try to build.

Mailaender:
bot, build org.kde.kommit

tsdgeos:
There's still one more small wrinkle (sorry, testing as we go)

If i do file->clone

and then on folder manually type /home/tsdgeos/lololo  (where /home/tsdgeos is my home) and then press OK, it succeeds, BUT the clone is nowhere to be found ^_^

If i use the file selector button then i get a path like /run/user/1000/doc/416f60f/tsdgeos/lololo and then when pressing OK it works.

So it's not totally broken but "half" broken in the sense that:
 * manually typing the path doesn't work
 * selecting the path in the folder gives you a "weird" path that the user may be scared of

hfiguiere:
maybe the only way is `--filesystem=home`...

Mailaender:
We did the same for https://github.com/flathub/com.github.git_cola.git-cola/blob/master/com.github.git_cola.git-cola.yml#L17

org.kde.kommit.json
```diff
@@ -0,0 +1,100 @@
...
+            "name": "git-lfs",
+            "buildsystem": "simple",
+            "build-commands": [
+                "PREFIX=${FLATPAK_DEST} ./install.sh"
+            ],
+            "sources": [
+                {
+                    "type": "archive",
+                    "strip-components": 1,
+                    "url": "https://github.com/git-lfs/git-lfs/releases/download/v3.4.0/git-lfs-linux-amd64-v3.4.0.tar.gz",
```
tsdgeos, line 56: Why are we downloading a binary instead of compiling it?

Mailaender: Ask the authors of https://github.com/flathub/org.flatpak.Builder/blob/master/org.flatpak.Builder.json#L238 I assume packaging Go is not well supported yet.

tsdgeos: I don't know why I would ask people unrelated to this MR, they are not proposing anything.

https://github.com/fyne-io/flatpak_demo/blob/main/io.fyne.flatpak_demo.yml seems to say building a go app just works.

Mailaender: I mean, they probably had their reasons to choose that path in an official @flathub package.

tsdgeos: What do you mean "an official flathub package"?

hfiguiere: My understanding is that git-lfs had botched vendoring the Go packages, so build kept failing in the pursuit of always having the latest package. (This is not an approval)

`org.flatpak.Builder` is maintained by the Flathub maintainers as it is the one used on the builders.

tsdgeos: > org.flatpak.Builder is maintained by the Flathub maintainers as it is the one used on the builders.

This is starting to discussing the smell of clours, but that doesn't really make it a more official package than say Okular.

> My understanding is that git-lfs had botched vendoring the Go packages, so build kept failing in the pursuit of always having the latest package. (This is not an approval)

I would really like if there was an attempt at making this work and someone could show me the error before we accept defeat and include a "random binary" in this package.

Mailaender:
bot, build org.kde.kommit

tsdgeos:
So you are not even try to get it to build?

Ok for me.

Mailaender:
I understand your objection of avoiding binary blobs even though it is allowed on Flathub. I am not skilled enough to build it from source. My experience with Go is very limited.

hfiguiere:
ok to merge, stable or beta?

git-lfs can still be fixed afterwards

@Mailaender it's a good way to learn.

travier:
Let's go for stable and we'll create an issue for Git LFS support

PR #4403, opened 2023-08-09
Add info.puredata.Pd

umlaeute:

### non-standard permissions

- `--device=all`
Pd is typically used to interact with MIDI-devices, but also HID-controllers and the like.
we also want to access the webcam (for live video processing)

- `--filesystem=xdg-run/pipewire`

info.puredata.Pd is a realtime audio environment. as such we need full access to the audio subsystem (typically Pd is run with the JACK audio system: for flatpak we use the pipewire emulation of the same)

- `--filesystem=home`

we need to be able to access both our own data (programs created with Pd) and additional footage (sounds, videos,...)

### DIFF
--- a/info.puredata.Pd.yml
+++ b/info.puredata.Pd.yml
@@ -56,10 +56,6 @@ modules:
     config-opts:
       - --enable-jack
     post-install:
-      - echo "<releases>" > releases.xml
-      - git log --tags --simplify-by-decoration --pretty="format:%ai %d" | sed -e '/^\([0-9-]*\) .*tag. \(0\.[0-9][0-9]-[0-9]\)[,)].*/!d' -e 's|^\([0-9-]*\) .*tag. \(0\.[0-9][0-9]-[0-9]\)[,)].*|<release version="\2" date="\1"/>|g' | sort -r >> releases.xml
-      - echo "</releases>" >> releases.xml
-      - sed -e "/<content_rating.*\/>/r releases.xml" -i "${FLATPAK_DEST}/share/metainfo/org.puredata.pd-gui.metainfo.xml"
       - appstream-util modify "${FLATPAK_DEST}/share/metainfo/org.puredata.pd-gui.metainfo.xml" id "${FLATPAK_ID}"
     sources:
       - type: git
@@ -70,3 +66,5 @@ modules:
           type: git
           tag-pattern: "^([\\d.]+-[\\d]+)$"
           is-main-source: true
+      - type: patch
+        path: patches/metainfo.patch
--- a/info.puredata.Pd.yml
+++ b/info.puredata.Pd.yml
@@ -1,0 +1,74 @@
+app-id: info.puredata.Pd
+runtime: org.freedesktop.Platform
+runtime-version: '22.08'
+sdk: org.freedesktop.Sdk
+command: pd-gui
+finish-args:
+# Display
+  - --share=ipc
+  - --socket=x11
+  - --device=dri
+# Audio
+  - --socket=pulseaudio
+  - --share=network
+  - --filesystem=xdg-run/pipewire-0
+# Filesystem access
+  - --filesystem=home
+# MIDI, webcams, controllers,...
+  - --device=all
+modules:
+  - shared-modules/linux-audio/jack2.json
+  - name: tcl
+    buildsystem: autotools
+    subdir: unix
+    build-options:
+      no-debuginfo: true
+    sources:
+      - type: archive
+        url: https://prdownloads.sourceforge.net/tcl/tcl8.6.13-src.tar.gz
+        sha256: 43a1fae7412f61ff11de2cfd05d28cfc3a73762f354a417c62370a54e2caf066
+        x-checker-data:
+          type: anitya
+          project-id: 4941
+          stable-only: true
+          url-template: https://prdownloads.sourceforge.net/tcl/tcl$version-src.tar.gz
+  - name: tk
+    buildsystem: autotools
+    subdir: unix
+    build-options:
+      no-debuginfo: true
+    post-install:
+      - test -e /app/bin/wish || find /app/bin/ -name "wish*.*" -printf "%f\0/app/bin/wish" -quit | xargs -0 ln -s
+    sources:
+      - type: archive
+        url: https://prdownloads.sourceforge.net/tcl/tk8.6.13-src.tar.gz
+        sha256: 2e65fa069a23365440a3c56c556b8673b5e32a283800d8d9b257e3f584ce0675
+        x-checker-data:
+          type: anitya
+          project-id: 11426
+          stable-only: true
+          url-template: https://prdownloads.sourceforge.net/tcl/tk$version-src.tar.gz
+
+  - name: puredata
+    buildsystem: autotools
+    config-opts:
+      - --enable-jack
+    sources:
+      - type: git
+        url: https://github.com/pure-data/pure-data.git
+        tag: 0.54-0
+        commit: c0a7c16c63e86ccc8ce05c8273ccd8aec7ae044c
+        x-checker-data:
+          type: git
+          tag-pattern: "^([\\d.]+-[\\d]+)$"
+          is-main-source: true
+  - name: integration
+    buildsystem: simple
+    build-commands:
+      - mv "/app/share/icons/hicolor/512x512/apps/puredata.png"    "/app/share/icons/hicolor/512x512/apps/${FLATPAK_ID}.png"
+      - mv "/app/share/icons/hicolor/48x48/apps/puredata.png"      "/app/share/icons/hicolor/48x48/apps/${FLATPAK_ID}.png"
+      - mv "/app/share/icons/hicolor/scalable/apps/puredata.svg"   "/app/share/icons/hicolor/scalable/apps/${FLATPAK_ID}.svg"
+      - mv "/app/share/applications/org.puredata.pd-gui.desktop"   "/app/share/applications/${FLATPAK_ID}.desktop"
+      - sed -e "s|^Icon=.*|Icon=${FLATPAK_ID}|"                 -i "/app/share/applications/${FLATPAK_ID}.desktop"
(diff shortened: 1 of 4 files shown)


info.puredata.Pd.yml
```diff
@@ -0,0 +1,74 @@
...
+# Audio
+  - --socket=pulseaudio
+  - --share=network
+  - --filesystem=xdg-run/pipewire-0
+# Filesystem access
+  - --filesystem=home
+# MIDI, webcams, controllers,...
+  - --device=all
+modules:
+  - shared-modules/linux-audio/jack2.json
```
hfiguiere, line 20:
```suggestion
```

You don't need jack here. The headers are in the SDK.

info.puredata.Pd.yml
```diff
@@ -0,0 +1,74 @@
...
+        x-checker-data:
+          type: git
+          tag-pattern: "^([\\d.]+-[\\d]+)$"
+          is-main-source: true
+  - name: integration
+    buildsystem: simple
+    build-commands:
+      - mv "/app/share/icons/hicolor/512x512/apps/puredata.png"    "/app/share/icons/hicolor/512x512/apps/${FLATPAK_ID}.png"
+      - mv "/app/share/icons/hicolor/48x48/apps/puredata.png"      "/app/share/icons/hicolor/48x48/apps/${FLATPAK_ID}.png"
+      - mv "/app/share/icons/hicolor/scalable/apps/puredata.svg"   "/app/share/icons/hicolor/scalable/apps/${FLATPAK_ID}.svg"
```
hfiguiere, line 70: for the icons, `rename-icon` in the manifest is the best way to do this.

info.puredata.Pd.yml
```diff
@@ -0,0 +1,74 @@
...
+      - mv "/app/share/applications/org.puredata.pd-gui.desktop"   "/app/share/applications/${FLATPAK_ID}.desktop"
```
hfiguiere, line 71: use `rename-desktop-file`

info.puredata.Pd.yml
```diff
@@ -0,0 +1,74 @@
...
+      - sed -e "s|^Icon=.*|Icon=${FLATPAK_ID}|"                 -i "/app/share/applications/${FLATPAK_ID}.desktop"
```
hfiguiere, line 72: `rename-icon` would als take care of this.

info.puredata.Pd.yml
```diff
@@ -0,0 +1,74 @@
+app-id: info.puredata.Pd
+runtime: org.freedesktop.Platform
+runtime-version: '22.08'
+sdk: org.freedesktop.Sdk
+command: pd-gui
+finish-args:
+# Display
+  - --share=ipc
+  - --socket=x11
+  - --device=dri
```
hfiguiere, line 10: this is part of `device-all`

info.puredata.Pd.yml
```diff
@@ -0,0 +1,74 @@
...
+        url: https://github.com/pure-data/pure-data.git
+        tag: 0.54-0
+        commit: c0a7c16c63e86ccc8ce05c8273ccd8aec7ae044c
+        x-checker-data:
+          type: git
+          tag-pattern: "^([\\d.]+-[\\d]+)$"
+          is-main-source: true
+  - name: integration
+    buildsystem: simple
+    build-commands:
```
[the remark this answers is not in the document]
hfiguiere: but with all the suggestion below this might just need to be removed.

info.puredata.Pd.yml
```diff
@@ -0,0 +1,74 @@
...
+      - mv "/app/share/icons/hicolor/512x512/apps/puredata.png"    "/app/share/icons/hicolor/512x512/apps/${FLATPAK_ID}.png"
+      - mv "/app/share/icons/hicolor/48x48/apps/puredata.png"      "/app/share/icons/hicolor/48x48/apps/${FLATPAK_ID}.png"
+      - mv "/app/share/icons/hicolor/scalable/apps/puredata.svg"   "/app/share/icons/hicolor/scalable/apps/${FLATPAK_ID}.svg"
+      - mv "/app/share/applications/org.puredata.pd-gui.desktop"   "/app/share/applications/${FLATPAK_ID}.desktop"
+      - sed -e "s|^Icon=.*|Icon=${FLATPAK_ID}|"                 -i "/app/share/applications/${FLATPAK_ID}.desktop"
+      - mv "/app/share/metainfo/org.puredata.pd-gui.metainfo.xml"  "/app/share/metainfo/${FLATPAK_ID}.metainfo.xml"
+      - sed -e "s|org\.puredata\.pd-gui|${FLATPAK_ID}|g"        -i "/app/share/metainfo/${FLATPAK_ID}.metainfo.xml"
```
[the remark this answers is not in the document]
hfiguiere: actually changing the id which this also does should be done with `appstream-util`
```
appstream-util modify $FLATPAK_DEST/share/metainfo/${FLATPAK_ID}.metainfo.xml id ${FLATPAK_ID}
```

hfiguiere:
bot, build info.puredata.Pd

info.puredata.Pd.yml
umlaeute: thanks.fixing

info.puredata.Pd.yml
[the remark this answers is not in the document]
umlaeute: indeed.

info.puredata.Pd.yml
umlaeute: thanks.fixing.

info.puredata.Pd.yml
umlaeute: thanks.fixing.

info.puredata.Pd.yml
umlaeute: thanks.fixing.

info.puredata.Pd.yml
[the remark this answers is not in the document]
umlaeute: i tend to disagree.
internally the application consists of two components: Pd-core and Pd-gui, which are separate programs that communicate via a network connection.
the current metainfo.xml comes from the Debian packaging, where the gui and the core are available in separate packages, and the metainfo.xml is shipped alongside the gui - hence the name.

for flatpak, the two components are shipped in a single flatpak application, which (imho) should be made available under a name that people will understand.
this could be either
- `info.puredata.Pd`
- `info.puredata.PureData`

info.puredata.Pd.yml
umlaeute: thanks.fixing.

info.puredata.Pd.yml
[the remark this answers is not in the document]
hfiguiere: no. whatever is used upstream should be used here. Unless it's invalid (it's not).

umlaeute:
bot, build info.puredata.Pd

umlaeute:
Please excuse my ignorance, but what is the `rename-appdata-file` for them?

hfiguiere:
unecessary if properly named.

info.puredata.Pd.yml
```diff
@@ -0,0 +1,72 @@
...
+
+  - name: puredata
+    buildsystem: autotools
+    config-opts:
+      - --enable-jack
+    post-install:
+      - echo "<releases>" > releases.xml
+      - git log --tags --simplify-by-decoration --pretty="format:%ai %d" | sed -e '/^\([0-9-]*\) .*tag. \(0\.[0-9][0-9]-[0-9]\)[,)].*/!d' -e 's|^\([0-9-]*\) .*tag. \(0\.[0-9][0-9]-[0-9]\)[,)].*|<release version="\2" date="\1"/>|g' | sort -r >> releases.xml
+      - echo "</releases>" >> releases.xml
+      - sed -e "/<content_rating.*\/>/r releases.xml" -i "${FLATPAK_DEST}/share/metainfo/org.puredata.pd-gui.metainfo.xml"
```
hfiguiere, line 62: really a patch is preferred.

info.puredata.Pd.yml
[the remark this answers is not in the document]
umlaeute: so i think i'm going to change this upstream.

am i right in assuming that changing the app-id will basically mean a new application, both from the user perspective (as in: there will be two distinct applications (with similar descriptions) available on flathub, in software centers,...) and from a dev perspective (i would need to re-issue an add request and a separate repository will be created)?

so if we plan to change the app id, we should do it *before* entering flathub.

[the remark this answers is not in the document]
umlaeute: otoh, i'm sure we won't be able to get a release very soon (but of course: you never know...)

I'm about to push a  commit to our (that is: upstream's) `develop` branch. would that be enough to accept local (that is: flatpak) workarounds that preempt the new upstream ID?

or do we have to stall this until there's a new upstream release with the fixed IDs?

info.puredata.Pd.yml
umlaeute: thanks.done

info.puredata.Pd.yml
```diff
@@ -0,0 +1,70 @@
...
+      - type: git
+        url: https://github.com/pure-data/pure-data.git
+        tag: 0.54-0
+        commit: c0a7c16c63e86ccc8ce05c8273ccd8aec7ae044c
+        x-checker-data:
+          type: git
+          tag-pattern: "^([\\d.]+-[\\d]+)$"
+          is-main-source: true
+      - type: patch
+        path: patches/metainfo.patch
```
hfiguiere, line 70: patch is missing.

hfiguiere:
does this support LADSPA or other kind of audio plugins?

umlaeute:
not directly. (only via 3rd party addons)

i'm aware of `--env=LADSPA_PATH=/app/extensions/Plugins/ladspa:/app/lib/ladspa`


just to be sure: once this gets accepted, I will still be able to update it, right?

umlaeute:
bot, build info.puredata.Pd

PR #4423, opened 2023-08-15
add io.github.NickKarpowicz.LightwaveExplorer

NickKarpowicz:
This is an open source nonlinear optics simulation that is both easy to use and powerful. It has already been used for several publications, and has an emphasis on visual feedback.

PR #4425, opened 2023-08-16
add io.github.NickKarpowicz.LightwaveExplorer

NickKarpowicz:

    - Extra permission requested is --filesystem=home. This is due to the nature of how the simulation code should run: The user enters the path where all of the results should be saved and can leave a queue that runs for a long time (some students have had them go for a full day). In these cases it's better if the results are saved immediately and automatically without further interaction. 
- [X] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:[I am the sole developer of the project, repo here](https://github.com/NickKarpowicz/LightwaveExplorer) **


NickKarpowicz:
Thank you very much @hfiguiere for the improvements made on my [last PR](https://github.com/flathub/flathub/pull/4423). Since I made a number of errors in the submission process, I created a new PR with a new version of the manifest which addresses your comments:
 - I now use fallback-X11 socket
 - hashes have been changed to sha256 from md5
 - buildsystems have been changed to cmake-ninja, including that of the main repo, which used a shell script earlier

io.github.NickKarpowicz.LightwaveExplorer.json
```diff
@@ -0,0 +1,145 @@
...
+                }
+            ]
+        },
+        {
+            "name": "LightwaveExplorer",
+            "buildsystem": "cmake-ninja",
+            "config-opts": ["-DMAKECUDAFLATPAK=TRUE", "-DCMAKE_CUDA_ARCHITECTURES=75", "-DCUDAToolkit_ROOT=/app/opt/cuda","-DCMAKE_CUDA_COMPILER=/app/opt/cuda/bin/nvcc"],
+            "build-options": {
+                "cxxflags": "-O3 -g -pipe -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong -grecord-gcc-switches",
+                "cxxflags-override": true},
```
hfiguiere, line 134: if the code triggers the assertion it should be fixed. disabling the assertions is just hiding the potential crashes.

io.github.NickKarpowicz.LightwaveExplorer.json
```diff
@@ -0,0 +1,145 @@
...
+                    "type": "git", 
+                    "url": "https://github.com/davisking/dlib", 
+                    "tag": "v19.24.2",
+                    "commit": "39749fba5163d4e25030358f6aa8d91617bfdca2" } 
+                ]
+        },
+        {
+            "name": "cuda",
+            "buildsystem": "simple",
+            "build-commands":["mkdir /app/tmp","sh cuda_12.2.1_535.86.10_linux.run --silent --toolkitpath=/app/opt/cuda --toolkit --override --tmpdir=/app/tmp"],
```
hfiguiere, line 119: you still haven't explained how the aarch64 flatpak that is built will work (or even build)

hfiguiere:
bot, build io.github.NickKarpowicz.LightwaveExplorer

io.github.NickKarpowicz.LightwaveExplorer.json
```diff
@@ -0,0 +1,145 @@
...
+                    "type": "file",
+                    "url": "https://developer.download.nvidia.com/compute/cuda/12.2.1/local_installers/cuda_12.2.1_535.86.10_linux.run",
+                    "sha256": "c51e814eaa756160a73289b1f35880cc38c18e1ddedc87f46f7ff0d1696bdb95"
+                }
+            ]
+        },
+        {
+            "name": "LightwaveExplorer",
+            "buildsystem": "cmake-ninja",
+            "config-opts": ["-DMAKECUDAFLATPAK=TRUE", "-DCMAKE_CUDA_ARCHITECTURES=75", "-DCUDAToolkit_ROOT=/app/opt/cuda","-DCMAKE_CUDA_COMPILER=/app/opt/cuda/bin/nvcc"],
```
hfiguiere, line 131: please format the JSON properly. its hard to read,

io.github.NickKarpowicz.LightwaveExplorer.json
NickKarpowicz: Is there a way to disable aarch64 entirely? The other dependency OneAPI is not compatible with it.

NickKarpowicz: I added an only-arches command to this effect, commit a46b776fa28f8f4821df3d1a02c37a681c206daf

hfiguiere: this won't do it

hfiguiere: https://docs.flathub.org/docs/for-app-authors/maintanance/#flathubjson

hfiguiere:
bot, build io.github.NickKarpowicz.LightwaveExplorer

io.github.NickKarpowicz.LightwaveExplorer.json
NickKarpowicz: I modified the code and point the manifest to the new tagged version in commit 40eac10c4e2c8ed439b978be97cbfd37708f786e. Earlier, I would check if the result of a call to std::exp() was NaN, which would cause the assertion to fail for a bad argument; now I check the argument first. Simulation results are the same, but now I don't override the build options.

NickKarpowicz:
I noticed that when launched the flatpak built by the buildbot has the default x11 icon in the KDE taskbar, while the one that I built with flatpak-builder locally has the application's icon. Is this normal behavior?

io.github.NickKarpowicz.LightwaveExplorer.json
NickKarpowicz: Build succeeds after making flathub.json; thanks

NickKarpowicz:
bot, build io.github.NickKarpowicz.LightwaveExplorer

NickKarpowicz:
I tested the flatpak on a few machines and I think it works well now. Is there anything else I should change in order to proceed?

NickKarpowicz:
bot, build io.github.NickKarpowicz.LightwaveExplorer

io.github.NickKarpowicz.LightwaveExplorer.json
```diff
@@ -0,0 +1,149 @@
...
+            "sources": [
+                {
+                    "type": "file",
+                    "url": "https://registrationcenter-download.intel.com/akdlm/IRC_NAS/992857b9-624c-45de-9701-f6445d845359/l_BaseKit_p_2023.2.0.49397_offline.sh",
+                    "sha256": "3dd04d2cc9ba2a8785580ba09068dbaf9aff9aa930b5fa7f10688210ea9be4a0"
+        
+                }
+            ],
+            "cleanup":[
+                "/app/opt/intel/oneapi/mkl/latest/lib/ia32",
```
hfiguiere, line 78: this doesn't do what you think it does.

>        cleanup (array of strings)
>           An array of file patterns that should be removed at the end. Patterns starting with / are taken to be full pathnames (without the /app prefix),
>           otherwise they just match the basename.

tl;dr the `/` of from the `$FLATPAK_DEST`.

NickKarpowicz: thank you; I removed /app from the paths in commit 083be8aa04c0f8951a5521a53eab4f4c0024ac5b

NickKarpowicz:
The patched version to work better in Flatpak form (config/database search path starts in home rather than ../share relative to binary) passed my tests.

PR #4435, opened 2023-08-22
Add com.howlingmoonsoftware.CrayonBall

slembcke:
This is an old Mac shareware game we made back in 2007-2009. We decided to open source it, and I figured it could find a new life on Flathub. [CrayonBall website](http://howlingmoonsoftware.com/crayonball.php)

Flathub setup Co-authored-by: Dallas Strouse <dastrouses@gmail.com>


orowith2os:
bot, build com.howlingmoonsoftware.CrayonBall

orowith2os:
Oh, the manifest file needs a rename to the proper capitalization.

slembcke:
Do I need to use the magic phrase again? (/me tries)

bot, build com.howlingmoonsoftware.CrayonBall

orowith2os:
Going to suggest a fix for the git tag soon

slembcke:
Apparently backslashes don't need to be escaped in yaml. Take 3.

bot, build com.howlingmoonsoftware.CrayonBall

slembcke:
Well, bummer. I was pretty sure something would fail on aarch64, but digging into why a 15 year old version of Ruby doesn't build for it is probably more effort than I'm willing to take on right now. I guess I can just skip it for now.

bot, build com.howlingmoonsoftware.CrayonBall

orowith2os:
You can enable aarch64 again, you just need to override the file that Ruby ships with a newer config.guess

com.howlingmoonsoftware.CrayonBall.yml
```diff
@@ -0,0 +1,59 @@
...
+    - /lib/cmake
+    - /lib/pkgconfig
+    - /share
+- name: Ruby
+  sources:
+  - type: archive
+    url: https://cache.ruby-lang.org/pub/ruby/1.8/ruby-1.8.7.tar.gz
+    sha256: 600dccf13bca3e4179fa6ff554220ce4ba67ffc72bce1ac3bf74c2599c03a0ca
+  - type: patch
+    path: ruby-no-openssl.patch
```
orowith2os, line 40:
```suggestion
  - type: patch
    path: ruby-no-openssl.patch
  # Our old Ruby (naturally) uses an outdated version of the
  # config.guess/config.sub script, so we override it here to fix aarch64 builds.
  - type: shell
    commands:
      - cp -p /usr/share/automake-*/config.{sub,guess} config-aux/
```

slembcke:
Well, once I saw that I remembered that the header search path ended up being `/app/lib/ruby/1.8/x86_64-linux`. So I also need to figure out how to get that value from within meson. Surely solveable by reading some documentation, just not tonight. ;)

com.howlingmoonsoftware.CrayonBall.yml
```diff
@@ -0,0 +1,59 @@
...
+- --share=ipc
+- --socket=fallback-x11
+- --socket=wayland
+- --socket=pulseaudio
+
+modules:
+- shared-modules/SDL2/SDL2-with-libdecor.json
+- shared-modules/glu/glu-9.json
+- name: openal-soft
+  buildsystem: cmake
```
hfiguiere, line 18: cmake-ninja?

com.howlingmoonsoftware.CrayonBall.yml
hfiguiere: if a 19KLOC patch really needed?

hfiguiere:
> For some reason it's looking for a Json, not yaml. @hfiguiere ?

because YAML suck.

But this was not the problem you are looking for.

crayonball-ruby.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/meson.build b/meson.build
+index 9b423f3..50e5247 100755
+--- a/meson.build
++++ b/meson.build
+@@ -68,7 +68,7 @@ exe = executable('aerosol',
+ 		'Aerosol/ext/Photon/ruby/ext/rb_photon_chipmunk.c',
+ 	],
+ 	include_directories: [
+-		include_directories(ruby_dir),
++		include_directories('/app/lib/ruby/1.8/x86_64-linux'),
```
hfiguiere, line 10: is that why you excluded aarch64?

orowith2os: Indeed it was, it should be fixable with a few modifications to how CrayonBall finds the path, or just hardcoding the path and switching based on arch. It'll also need to fix https://github.com/slembcke/CrayonBall/blob/9bcab497c890e96e12788734bf0fdaa0c2a97050/Aerosol/ext/Chipmunk/CMakeLists.txt#L33

slembcke: Yeah. I have more experience with CMake, but wanted to try Meson for once. A bit out of my element making it do anything other than basic builds, but I can figure that out later. I want to make it control better with touchscreens too.

hfiguiere:
bot, build com.howlingmoonsoftware.CrayonBall

orowith2os:
Just a few minor bugs to work out, and I'd like to postpone this temporarily so that some quick fixes for aarch64 can land.

flathub.json
```diff
@@ -0,0 +1,4 @@
+{
+  "automerge-flathubbot-prs": true,
```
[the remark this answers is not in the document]
orowith2os: When was that decided? There are a ton of people using it, myself included, for non-extra-data purposes.

PR #4444, opened 2023-08-25
Add com.ghostery.browser

milouse:

  + gtk3-werror.patch is required as part of cups support (also used by librewolf or firefox flatpak. See https://github.com/flathub/org.mozilla.firefox.BaseApp/pull/21 https://github.com/flathub/org.mozilla.firefox.BaseApp/pull/22 and https://phabricator.services.mozilla.com/D138908

Hi there,

We would like to distribute our mozilla-based browser on flathub. This is our first interaction with flathub or flatpak in general, thus do not hesitate to come back to us if we did something wrong. We tried hard to follow all readmes/guidelines, but we might have missed something. This manifest is heavily inspired by the one of librewolf, if it can help review.

Thank you very much!


com.ghostery.browser.yml
```diff
@@ -0,0 +1,102 @@
...
+  - install -D -m644 /app/lib/ghostery/browser/chrome/icons/default/default128.png
+    /app/usr/share/icons/hicolor/128x128/apps/com.ghostery.browser.png
+  - sed -i 's|"app.update.channel", "release"|"app.update.channel", "flatpak"|' /app/lib/ghostery/defaults/pref/channel-prefs.js
+  - install -D -m644 com.ghostery.browser.desktop -t /app/share/applications
+  - install -D -m644 com.ghostery.browser.metainfo.xml -t /app/share/metainfo
+  - install -D -m755 ghostery.sh /app/bin/com.ghostery.browser
+  - install -d /app/lib/ffmpeg
+  sources:
+  - type: archive
+    url: https://github.com/ghostery/user-agent-desktop/releases/download/2023-07-21/Ghostery-2023.6.en-US.linux.tar.gz
```
JakobDev, line 55: Why you don't build from source?

com.ghostery.browser.yml
```diff
@@ -0,0 +1,102 @@
+---
+app-id: com.ghostery.browser
+default-branch: stable
```
hfiguiere, line 3:
```suggestion
```

com.ghostery.browser.yml
```diff
@@ -0,0 +1,102 @@
...
+- "--talk-name=org.freedesktop.FileManager1"
+- "--system-talk-name=org.freedesktop.NetworkManager"
+- "--talk-name=org.a11y.Bus"
+- "--talk-name=org.gnome.SessionManager"
+- "--talk-name=org.freedesktop.ScreenSaver"
+- "--talk-name=org.gtk.vfs.*"
+- "--talk-name=org.freedesktop.Notifications"
+- "--env=MOZ_ENABLE_WAYLAND=1"
+- "--env=MOZ_USE_XINPUT2=1"
+- "--own-name=com.ghostery.browser.*"
```
hfiguiere, line 35:
```suggestion
```

this is granted based on the app-id

com.ghostery.browser.yml
```diff
@@ -0,0 +1,102 @@
...
+- "--own-name=org.mpris.MediaPlayer2.firefox.*"
```
hfiguiere, line 36: you mean that it doesn't use it's own mpris name and "borrow" the one from Firefox?

com.ghostery.browser.yml
```diff
@@ -0,0 +1,102 @@
...
+modules:
+- shared-modules/dbus-glib/dbus-glib-0.110.json
+- name: ghostery_browser
```
hfiguiere, line 39: it's customary to have the main module last because it's dependency order.

com.ghostery.browser.yml
hfiguiere, line 55: how does this work with aarch64 since an aarch64 package is build here as well?

hfiguiere:
bot, build com.ghostery.browser

com.ghostery.browser.yml
```diff
@@ -0,0 +1,102 @@
...
+  buildsystem: simple
+  build-commands:
+  - mkdir -p /app/lib
+  - mv ghostery_app/Ghostery /app/lib/ghostery
+  - install -D -m644 /app/lib/ghostery/browser/chrome/icons/default/default128.png
+    /app/share/icons/hicolor/128x128/apps/com.ghostery.browser.png
+  - install -D -m644 /app/lib/ghostery/browser/chrome/icons/default/default128.png
+    /app/usr/share/icons/hicolor/128x128/apps/com.ghostery.browser.png
```
hfiguiere, line 47: why do you install an icon in `/app/usr` ?

com.ghostery.browser.yml
hfiguiere: this is where the linter errors.

milouse: Oh, I didn’t know. Thanks for the tip!

com.ghostery.browser.yml
milouse: Can you explain why aarch64 will be built too? I added the `only-arches: [x86_64]` attribute to this source to only build for this target. Did I misunderstood something?

hfiguiere: this just exclude the *module*

https://docs.flathub.org/docs/for-app-authors/maintanance/#limiting-the-set-of-architectures-to-build-on

But then why not building from source to ship an aarch64 package.

com.ghostery.browser.yml
milouse: Yes, [same as for LibreWolf](https://github.com/flathub/io.gitlab.librewolf-community/blob/master/io.gitlab.librewolf-community.json#L37). It’s a limitation in the mozilla build chain. As we don’t want to patch it too deeper, we keep this naming.

com.ghostery.browser.yml
milouse: It looks like another misunderstanding. I removed it.

com.ghostery.browser.yml
milouse: Please see my general comment below

milouse: Got it.

About the build step, please see my general comment below.

milouse:
Thank you very much for your quick and detailed review. I hope my various answer/action will be ok for you.

Both of you ask about why not actually build the browser when packaging it. The mozilla build chain is a bit complicated and requires both big resources and lot of various customization. From our point of view it’s way more simpler to only build our browser once on our side, controlling the whole process, then distributing it on various channel from the same tarball. It looks a bit as a waste of resources to compile it several time, just for various packaging purpose. This also simplify a lot our support process, knowing that normally our end-users will have the same binary under the hood. So for now, if it’s not a problem we would like to keep things like this. However we do not completely close the door to build a specific version of our browser while creating flatpak package. But currently we do not have the resources to do it.

JakobDev:
Please add a flathub.json file with the following content:
```json
{
    "only-arches": ["x86_64"]
} 
```
This  disables the ARM build.

A problem I can see is that you have different binaries for different languages (en,de,fr). So if a User installs Ghostery from Flathub, he will only get the english version. Could you pack all languages into one binary?

milouse:
Thank you for your feedback!

> Please add a flathub.json file with the following content:
> 
> ```json
> {
>     "only-arches": ["x86_64"]
> } 
> ```
> 
> This disables the ARM build.

Done in my last commit.

> A problem I can see is that you have different binaries for different languages (en,de,fr). So if a User installs Ghostery from Flathub, he will only get the english version. Could you pack all languages into one binary?

Yes, we know about it and it’s on our roadmap. For various reason we did not succeed to do it for now (nor took time to fix it). So yes our first flatpak release(s) will be english only. It will be fixed on a future release.

hfiguiere:
bot, build com.ghostery.browser

com.ghostery.browser.yml
```diff
@@ -0,0 +1,97 @@
...
+  - install -D -m755 ghostery.sh /app/bin/com.ghostery.browser
+  - install -d /app/lib/ffmpeg
+  sources:
+  - type: archive
+    url: https://github.com/ghostery/user-agent-desktop/releases/download/2023-07-21/Ghostery-2023.6.en-US.linux.tar.gz
+    sha256: 84bbd2221172e4b78fc88c8f9c46af0a43c8cc9e08d33ec32e7e35334b9e8609
+    dest: ghostery_app
+    strip-components: 0
+  - type: file
+    path: "../data/ghostery-private-browser.desktop"
```
hfiguiere, line 92:
```
flatpak-builder: Downloading ghostery_browserFailed to download sources: module ghostery_browser: Unable to get source file '../data/ghostery-private-browser.desktop': File '/srv/buildbot/worker/build-x86_64-2/data' not found
```

com.ghostery.browser.yml
```diff
@@ -0,0 +1,97 @@
...
+    dest-filename: com.ghostery.browser.desktop
```
hfiguiere, line 93:
```suggestion
    path: com.ghostery.browser.desktop
```

milouse: 🤦 exactly. Very sorry we are using another repo to test flatpak, in which the desktop file is shared in another folder. When I moved the ghostery target from the second to the last step, I forget to fix the file path :/ My bad. I’ll fix that.

milouse:
Done. Sorry for the back and forth :/

hfiguiere:
bot, build com.ghostery.browser

PR #4449, opened 2023-08-27
Add com.librumreader.librum

DavidLazarescu:


DavidLazarescu:
Hello, I got the manifest file to build the project correctly and made sure to move the .desktop and icon file to the positions they need to be.

I am running into errors with a dependency of my project (that I am building from source). This dependency internally tries downloading `libclang` and `swig` via `pip install` which causes errors, since you are not supposed to (from what I know) download other things from the internet during your build steps. I tried using the `flatpak-pip-generator` to prevent the download which worked fine for `libclang`, but I got multiple `ModuleNotFoundError`s for `swig` which lead me to run `flatpak-pip-generator` for the two other pip modules as well. One of them tries downloading something itself though, which again causes my build to fail.

I hope someone here might have an idea on how to resolve this. I have tried fixing this for ~2 days now but I can not get a solution anywhere. Is it possible to enable downloading from within a build step? If no, what else can I do?

Thanks in advance.

DavidLazarescu:
> The manifest needs some fixes too.

What would I need to fix in the manifest? It's my first submissions, so I do not know exactly what needs to be provided.

DavidLazarescu:
> > One of them tries downloading something itself though, which again causes my build to fail.
> 
> If downloading a binary is hardcoded into the build process, you need to patch it out.
> 
> bot, build com.librumreader.librum

It seems like while using the script generated by `flatpak-pip-generator` to download and build `swig`, it doesn't find `wheel` so it tries to download it, which leads to the error "error: downloading 'https://github.com/PCRE2Project/pcre2/releases/download/pcre2-10.40/pcre2-10.40.zip' failed". I have tried using `flatpak-pip-generator` to generate a config file for `wheel`, but `swig` still tries to download it, resulting in the same error.

Do you have any how I could approach this? Is there by any chance a way to enable downloads during the build steps?

DavidLazarescu:
Ok, this should fix the manifest issues. At least the ones detected by `flatpak run --command=flatpak-builder-lint org.flatpak.Builder --exceptions com.librumreader.librum.yaml`, but the previous problem still remains and I do not know how to resolve it.

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+    "only-arches": ["x86_64"]
```
hfiguiere, line 2: an ebook reader that doesn't run on aarch64?

com.librumreader.librum.yaml
```diff
@@ -0,0 +1,41 @@
+app-id: com.librumreader.librum
+runtime: org.kde.Platform
+runtime-version: '6.5'
+sdk: org.kde.Sdk
+command: librum
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=fallback-x11
+  - --filesystem=host
```
hfiguiere, line 10: this is a very broad filesystem permission for an ebook reader.

com.librumreader.librum.yaml
```diff
@@ -0,0 +1,41 @@
+app-id: com.librumreader.librum
+runtime: org.kde.Platform
+runtime-version: '6.5'
+sdk: org.kde.Sdk
+command: librum
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=fallback-x11
```
hfiguiere, line 9:
```suggestion
  - --socket=x11
```

com.librumreader.librum.yaml
```diff
@@ -0,0 +1,41 @@
...
+    buildsystem: cmake-ninja
+    build-options:
+      - build-args:
+        - --share-network
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release 
+      - -DBUILD_TESTS=Off
+    sources:
+        - type: git
+          branch: main
```
hfiguiere, line 28: NEVER a git branch. this is why the linter is unhappy.
```
        "module-librum-bin-source-git-no-commit-or-tag",
        "module-librum-bin-source-git-branch"
```

flathub.json
DavidLazarescu: Right, I forgot I had this file. Should I completely remove this, or should I just add aarch64 to the list?

com.librumreader.librum.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - python3-wheel.yaml
+  - python3-setuptools_scm.yaml
+  - python3-scikit-build.yaml
+  - python3-swig.yaml
+  - python3-libclang.yaml
+  - name: librum
+    buildsystem: cmake-ninja
+    build-options:
+      - build-args:
+        - --share-network
```
hfiguiere, line 22: this is not allowed. The build will fail because of that.

com.librumreader.librum.metainfo.xml
```diff
@@ -0,0 +1,29 @@
...
+      <image>https://user-images.githubusercontent.com/69865187/244929153-ea94fc68-1bf0-4933-8d80-43a57c6590c5.png</image>
+    </screenshot>
+    <screenshot>
+      <image>https://user-images.githubusercontent.com/69865187/261866073-bf1d0401-62bd-4f4e-b008-523fb2efd275.png</image>
+    </screenshot>
+    <screenshot>
+      <image>https://user-images.githubusercontent.com/69865187/262316821-b8995cf1-a0e6-4993-8c8b-92f7f8e79ebd.png</image>
+    </screenshot>
+  </screenshots>
+</component>
```
hfiguiere, line 29: missing mandatory `url`, `releases` and `content_rating`

python3-wheel.yaml
hfiguiere, line 1: this file does nothing.

com.librumreader.librum.yaml
hfiguiere: this assume it doesn't support wayland as you signaled by removing the wayland socket.

com.librumreader.librum.yaml
DavidLazarescu: From the docs I got that "--share=ipc - share IPC namespace with the host (necessary for X11)". Since my application is supposed to run on X11, I supposed that I need to have it in my permissions. I need `--share=network` since my application is syncing books and other kind of data to the server, thus needs internet connection for http requests.
`--filesystem=host` is needed since my application will load selected files from the user's file system and export the files from the application if wanted by the user. From what I understood the `--socket=...` is for showing windows? Since my application is a GUI application I supposed that I need that permission as well.

Please correct me if I misunderstood any permission. Also would my application run on both X11 and wayland the way my permissions are set up? I did not get if `--socket=fallback-x11` tries wayland and then falls back to x11 and thus covers both wayland and x11, or if I need also `--socket=x11` and `--socket=wayland`.

com.librumreader.librum.yaml
DavidLazarescu: When I do not specify a branch, it tries getting it from master, which fails. Is there another fix for that then?

com.librumreader.librum.yaml
DavidLazarescu: I got that recommendation from another person, while I was trying to fix my network errors. I will remove it.

python3-wheel.yaml
DavidLazarescu: Ah, I just saw that comment from `flatpak-pip-generator` saying: "wheel is in system_packages. Skipping.". This is throwing me off a little though, since my main build error comes from `swig` not finding wheel, thus needing to download it.

com.librumreader.librum.yaml
DavidLazarescu: Never mind, I have tried removing the `branch: main` and it seems to work and not complain about not finding master now.

com.librumreader.librum.yaml
hfiguiere: > very broad filesystem permission

there is only ONE line that is a filesystem permission here. Its line 10.

DavidLazarescu: > > very broad filesystem permission
> 
> there is only ONE line that is a filesystem permission here. Its line 10.

My bad, I didn't read the "filesystem". I chose `host` since my application is not supposed to care where the user wants their file installed to or imported from, it can be from a home directory or from a privileged location.

DavidLazarescu:
Are the last commits fixing all the problems you mentioned, or is there still something?

hfiguiere:
bot, build com.librumreader.librum

com.librumreader.librum.yaml
hfiguiere: this is too broad. it's an ebook reader. Since it's Qt based, in theory it should use the portal by default, and not need filesystem performission. Otherwise `host` is very broad.

flathub.json
hfiguiere, line 1: just remove the file.

com.librumreader.librum.metainfo.xml
hfiguiere: this wasn't addressed.

hfiguiere (changes requested):
the appstream file is never installed. icons and desktop files have the wrong name. see `rename-desktop-file` and `rename-icon`

hfiguiere:
swig still tried to download stuff from the internet when building.

com.librumreader.librum.metainfo.xml
DavidLazarescu: I will add this once the build succeeds

DavidLazarescu:
> swig still tried to download stuff from the internet when building.

I realize that, I do not know how I would prevent that. Do you have an idea?

com.librumreader.librum.metainfo.xml
hfiguiere: the build will not. The appstream validation will also fail the build for that.

DavidLazarescu: I understand, I will fix it now

DavidLazarescu:
> 

With appstream file you mean the .metainfo.xml file, right?

DavidLazarescu:
> 



> the appstream file is never installed. icons and desktop files have the wrong name. see `rename-desktop-file` and `rename-icon`

My .desktop file is located in my github's root directory, and my icon file is in the src/ directory. Am I supposed to move `librum.desktop` to /app/share/applications and then use `rename-desktop-file: librum.desktop` and do the same for the icon: moving it to /app/share/icons/hicolor/scalable/apps/ and then add `rename-icon: librum.svg`?

DavidLazarescu:
I have fixed the installation of the appdata, icon and desktop file in my last commit. Please tell me if I did it correctly.

com.librumreader.librum.yaml
```diff
@@ -0,0 +1,37 @@
...
+          commit: 13386c7185f37e1f9525f86ba959d5fcf1e0f7ef
+  - name: install-files
+    buildsystem: simple
+    build-commands:
+      - install -Dm644 librum.desktop -t /app/share/applications
+      - install -Dm644 com.librumreader.librum.metainfo.xml -t /app/share/metainfo
+      - install -Dm644 src/logo.svg -t /app/share/icons/hicolor/scalable/apps
+    rename-desktop-file: librum.desktop
+    rename-icon: logo.svg
+    rename-appdata-file: com.librumreader.librum.metainfo.xml
```
hfiguiere, line 35: these are invalid here.

See documentation https://docs.flatpak.org/en/latest/flatpak-builder-command-reference.html

hfiguiere: line 34 is incorrect (no extension it's the icon **name**)
line 35 is unecessary

com.librumreader.librum.yaml
```diff
@@ -0,0 +1,37 @@
...
+    sources:
+        - type: git
+          url: https://github.com/Librum-Reader/Librum.git
+          commit: 13386c7185f37e1f9525f86ba959d5fcf1e0f7ef
+  - name: install-files
+    buildsystem: simple
+    build-commands:
+      - install -Dm644 librum.desktop -t /app/share/applications
+      - install -Dm644 com.librumreader.librum.metainfo.xml -t /app/share/metainfo
+      - install -Dm644 src/logo.svg -t /app/share/icons/hicolor/scalable/apps
```
hfiguiere, line 32: this is not the icon name specified in the .desktop

com.librumreader.librum.yaml
```diff
@@ -0,0 +1,37 @@
...
+      - -DCMAKE_BUILD_TYPE=Release 
+      - -DBUILD_TESTS=Off
+    sources:
+        - type: git
+          url: https://github.com/Librum-Reader/Librum.git
+          commit: 13386c7185f37e1f9525f86ba959d5fcf1e0f7ef
+  - name: install-files
+    buildsystem: simple
+    build-commands:
+      - install -Dm644 librum.desktop -t /app/share/applications
```
hfiguiere, line 30: the .desktop file should not use an absolute path for the `Excec`. You need to change it with `desktop-file-edit` until it's fixed upstream.

com.librumreader.librum.yaml
DavidLazarescu: Moved them to the top scope in my latest commit

com.librumreader.librum.yaml
```diff
@@ -0,0 +1,37 @@
...
+      - install -Dm644 ../com.librumreader.librum.metainfo.xml -t /app/share/metainfo
+      - install -Dm644 src/logo.svg -t /app/share/icons/hicolor/scalable/apps
+rename-desktop-file: librum.desktop
+rename-icon: logo.svg
+rename-appdata-file: com.librumreader.librum.metainfo.xml
```
hfiguiere, line 35: these three lines should be **above** modules by convention.

actually as I previously said, line 35 should be removed.

hfiguiere: and line 34 still incorrect.

com.librumreader.librum.yaml
DavidLazarescu: Right, the icon name in my .desktop file is `librum.svg`. Am I expected to change this, to whatever the `rename-icon` operation will call it, or can I simply change the name of the logo to `librum.svg` and leave the rest as it is?

hfiguiere:
> > swig still tried to download stuff from the internet when building.
> 
> I realize that, I do not know how I would prevent that. Do you have an idea?

try installing the package it tries to download. That would be my troubleshooting step.

com.librumreader.librum.yaml
DavidLazarescu: Did this.

com.librumreader.librum.yaml
hfiguiere: it's `librum` in the .desktop file. which is **consistent**. `rename-icon` will fix the desktop file which is why I suggested it.

I both case the icon name **exclude** the extension because you can also have bitmap icons allong the svg.

com.librumreader.librum.yaml
DavidLazarescu: Fixed upstream

com.librumreader.librum.yaml
```diff
@@ -0,0 +1,35 @@
...
+command: librum
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --filesystem=home
+  - --device=dri
+rename-desktop-file: librum.desktop
+rename-icon: librum.svg
```
hfiguiere, line 14: one more time: there should not be the extension. it's the icon **name**.

com.librumreader.librum.yaml
```diff
@@ -0,0 +1,35 @@
...
+        - type: git
+          url: https://github.com/Librum-Reader/Librum.git
+          commit: e6c69697fc0dfc2a9e625dd849dcbb25a66258ce
+  - name: install-files
+    buildsystem: simple
+    build-commands:
+      - install -Dm644 librum.desktop -t /app/share/applications
+      - install -Dm644 ../com.librumreader.librum.metainfo.xml -t /app/share/metainfo
+      - mv src/logo.svg src/librum.svg
+      - install -Dm644 src/librum.svg -t /app/share/icons/hicolor/scalable/app
```
hfiguiere, line 35:
```suggestion
      - install -Dm644 src/logo.svg /app/share/icons/hicolor/scalable/app/librum.svg
```

com.librumreader.librum.yaml
hfiguiere:
```suggestion
rename-icon: librum
```

com.librumreader.librum.yaml
DavidLazarescu: > I both case the icon name exclude the extension because you can also have bitmap icons allong the svg.

I am not sure if I understand this. Am I supposed to remove the .svg only for the `rename-icon` command?

I have just made a commit about it. Is that everything that I needed to do, or is there still something wrong with the icon?

com.librumreader.librum.yaml
DavidLazarescu: Did this in the last commit

com.librumreader.librum.metainfo.xml
hfiguiere, line 1: this file should be manged upstream in a future release.

(it's not flathub specific)

DavidLazarescu: I understand. Do I need it to publish something on flathub at all, or can I remove it completely?

hfiguiere: it's required.

DavidLazarescu:
> > > swig still tried to download stuff from the internet when building.
> > 
> > 
> > I realize that, I do not know how I would prevent that. Do you have an idea?
> 
> try installing the package it tries to download. That would be my troubleshooting step.

Swig itself seems to have a dependency on another c project (PCRE2). Would this mean that I need to get both of these big cmake repositories and compile them from source in order to get swig? Am I not able to somehow get a precompiled binary of it (wheel)?

com.librumreader.librum.metainfo.xml
DavidLazarescu: > it's required.

Ok, you told me to not install it though ("Remove line 35"). Am I just supposed to leave it in the folder of the repository?

hfiguiere: no. it was the `rename-appdata-file` line that was unecessary since it is installed with the proper name.

hfiguiere: image

DavidLazarescu: > no. it was the `rename-appdata-file` line that was unecessary since it is installed with the proper name.

I understand. Thanks for elaborating.

DavidLazarescu:
@hfiguiere I have continued resolving the dependencies of swig in my last commit. I saw that it is failing to download pcre2, so I generated the config file for it using `flatpak-pip-generator` (See my last commit). Non the less, swig still tries to download it which fails with `error: downloading 'https://github.com/PCRE2Project/pcre2/releases/download/pcre2-10.40/pcre2-10.40.zip`

hfiguiere:
bot, build com.librumreader.librum

DavidLazarescu:
> @hfiguiere I have continued resolving the dependencies of swig in my last commit. I saw that it is failing to download pcre2, so I generated the config file for it using `flatpak-pip-generator` (See my last commit). Non the less, swig still tries to download it which fails with `error: downloading 'https://github.com/PCRE2Project/pcre2/releases/download/pcre2-10.40/pcre2-10.40.zip`

Does someone have an idea on how I could do this properly?

DavidLazarescu:
> E.g. Kodi disabled PCRE support in SWIG: https://github.com/flathub/tv.kodi.Kodi/blob/bf52825a4848e1d73a6e1a26cfc7436f8a289907/modules/swig/swig.json#L2
> 
> So that's probably some way to fix it.

Thanks, I will look into that

DavidLazarescu:
@barthalion I have managed to get swig to work, but the script of the third-party library I am using is running `python -m pip install --upgrade pip && python -m pip install --upgrade libclang swig`. I suppose this is causing the network errors, since `--upgrade` tries to access the internet, right?

Is it possible to package up a venv or something similar? What options do I have here?

DavidLazarescu:
> There's no way to get it right without patching the third-party library. It just shouldn't be doing that.

Ok so I managed to remove the --venv option from my script invocation which means that it doesn't try to upgrade any packages anymore. Also I have removed `swig` completely since it is not needed when not using --venv.
Now, even though I have installed libclang (see my python3-libclang.yaml file above), pip is not able to find it, telling me:

"clang.cindex.LibclangError: libclang.so: cannot open shared object file: No such file or directory. To provide a path to libclang use Config.set_library_path() or Config.set_library_file()."

Where is the library installed to in the runtime?

DavidLazarescu:
Does anyone have an idea?

DavidLazarescu:
> You probably need to install the LLVM extension as part of your manifest: https://github.com/flathub/org.freedesktop.Sdk.Extension.llvm16

Can I use the freedesktop extension while I am using org.kde.Sdk?

DavidLazarescu:
> Yes, KDE and GNOME runtimes are built on top of freedesktop and are compatible with it.

I still get this error

```
FAILED: libs/mupdf/build/shared-release/libmupdfcpp.so /run/build/librum/libs/mupdf/build/shared-release/libmupdfcpp.so
cd /run/build/librum/libs/mupdf && ./scripts/mupdfwrap.py -d build/shared-release -b -j 0 m01
(+0.0s): parse.py:14:<module>: Warning, could not import clang: No module named 'clang'
(+0.1s): state.py:18:<module>: Warning: failed to import clang.cindex: e=ModuleNotFoundError("No module named 'clang'")
(+0.1s): state.py:18:<module>: We need Clang Python to build MuPDF python.
(+0.1s): state.py:18:<module>: Install with `pip install libclang` (typically inside a Python venv),
(+0.1s): state.py:18:<module>: or (OpenBSD only) `pkg_add py3-llvm.
```

This is my manifest:
```
app-id: com.librumreader.librum
runtime: org.kde.Platform
runtime-version: '6.5'
sdk-extensions:
- org.freedesktop.Sdk.Extension.llvm16
sdk: org.kde.Sdk
command: librum
finish-args:
- --share=ipc
- --share=network
- --socket=wayland
- --socket=fallback-x11
- --filesystem=home
- --device=dri
rename-desktop-file: librum.desktop
rename-icon: librum
modules:
- name: libclangInstall
buildsystem: simple
build-commands:
- mkdir -p $FLATPAK_DEST/lib
- cp $(readlink -f /usr/lib/sdk/llvm16/lib/libclang.so) $FLATPAK_DEST/lib/libclang.so
- name: librum
buildsystem: cmake-ninja
config-opts:
- -DCMAKE_BUILD_TYPE=Release
- -DBUILD_TESTS=Off
- -DNO_VENV=On
sources:
- type: git
url: https://github.com/Librum-Reader/Librum.git
commit: d62ee446dc3ad0b79c2598cc2132316902b46316
- name: install-files
buildsystem: simple
build-commands:
- install -Dm644 librum.desktop -t /app/share/applications
- install -Dm644 ../com.librumreader.librum.metainfo.xml -t /app/share/metainfo
- install -Dm644 src/logo.svg -t /app/share/icons/hicolor/scalable/apps/librum.svg
```
(The indentation got messed up by copy pasting)

DavidLazarescu:
bot, build com.librumreader.librum

DavidLazarescu:
Ok the build works now, the only thing left is the installation of the .desktop file from what I see. What is wrong there? I thought the position is in the root of my cloned github repo (where librum.desktop) is located, but it seems like its in `/run/build/install-files`. How would I do this correctly?

com.librumreader.librum.yaml
```diff
@@ -0,0 +1,34 @@
...
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release 
+      - -DBUILD_TESTS=Off
+      - -DNO_VENV=On
+    sources:
+        - type: git
+          url: https://github.com/Librum-Reader/Librum.git
+          commit: d62ee446dc3ad0b79c2598cc2132316902b46316
+  - name: install-files
```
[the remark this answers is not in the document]
hfiguiere: you don't need a separate module.

DavidLazarescu:
@hfiguiere fixed the install paths, now on the last renaming step, I get:
"Error: icon librum not found below /home/creapermann/Programming/Etovex/Deploy/flatpak-Librum/.flatpak-builder/rofiles/rofiles-M4iFx2/files/share/icons"

com.librumreader.librum.yaml
```diff
@@ -0,0 +1,33 @@
...
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release 
+      - -DBUILD_TESTS=Off
+      - -DNO_VENV=On
+    sources:
+        - type: git
+          url: https://github.com/Librum-Reader/Librum.git
+          commit: 0e9c8449d99b82721994e8e151064e8b59559247
+    build-commands:
+      - ls -R .
```
hfiguiere, line 30:
```suggestion
    post-install:
```

com.librumreader.librum.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - install -Dm644 librum.desktop -t /app/share/applications
+      - install -Dm644 com.librumreader.librum.metainfo.xml -t /app/share/metainfo
+      - install -Dm644 src/logo.svg -t /app/share/icons/hicolor/scalable/apps/librum.svg
```
hfiguiere, line 33:
```suggestion
      - install -Dm644 src/logo.svg -t /app/share/icons/hicolor/scalable/apps/
```

DavidLazarescu:
bot, build com.librumreader.librum

DavidLazarescu:
It works perfectly fine @hfiguiere, will it automatically be merged into flathub or do I still need to do something?

DavidLazarescu:
Never mind, I still need to move one folder to the correct location

DavidLazarescu:
My application expects the /resources/fonts folder from my git repo in $HOME/.var/app/com.librumreader.librum/data/Librum-Reader/Librum. How exactly can I reach that folder, I dont just want to `mv ... $HOME/.var/app/com.librumreader.librum/data/Librum-Reader/Librum` since I suppose that this will create problems on system wide installs?

DavidLazarescu:
> If your app expects something in $XDG_DATA_HOME, it should be putting that data there itself. If these are "assets", you should probably add make it look in `$PREFIX/share/Librum-Reader/Librum `, with `$PREFIX` being configurable build time.

I have instead embedded the fonts into my application to avoid any complications with copying the directory.
The app seems to work just fine, how exactly should I proceed from here? What do I now need to do to bring the application to flathub?
Also, what will the update process look like, can I just continue pushing commits to this fork to e.g. update the version, or how am I supposed to do that?

PR #4479, opened 2023-09-03
Add com.pojtinger.felicitas.Multiplex

pojntfx:

Aside from the permissions typical for GNOME applications, this app uses the `--talk-name=org.freedesktop.Flatpak` and `--filesystem=/tmp:rw` permissions. This is a requirement due to it launching MPV (from Flatpak or from a different package manager) in order to play back videos from [hTorrent](https://github.com/pojntfx/htorrent) and controlling it over an IPC socket. The code using this permission can be found here: [GitHub](https://github.com/pojntfx/multiplex/blob/675bd358da7bff96aac9435f658c81a54e1df3f1/pkg/client/mpv.go#L16-L61).

The app handles MPV not being available gracefully:

 
image

It also allows overwriting the command for starting MPV:

image

I decided against vendoring MPV inside the Flatpak due to two reasons:

1) The maintenance overhead of having to vendor a C project as part of what is currently a pure Go project
2) Playback-specific permissions, codecs being handled by the MPV Flatpak or the native package version instead of them having to be vendored as part of this package

Portals for playing video can not be used because MPV's socket and ability to stream video over HTTP via range requests are a requirement.
 

pojntfx:
bot, build com.pojtinger.felicitas.Multiplex

pojntfx:
I'm assuming these build errors relate to the use of the aforementioned permissions:

image

hfiguiere:
why not building mpv in the flatpak instead>

com.pojtinger.felicitas.Multiplex.json
```diff
@@ -0,0 +1,648 @@
+{
+  "app-id": "com.pojtinger.felicitas.Multiplex",
+  "runtime": "org.gnome.Platform",
+  "runtime-version": "44",
+  "sdk": "org.gnome.Sdk",
+  "sdk-extensions": ["org.freedesktop.Sdk.Extension.golang"],
+  "command": "multiplex",
+  "finish-args": [
+    "--share=network",
+    "--socket=fallback-x11",
```
hfiguiere, line 10: --share=ipc is necessary (see the screenshot you posted)

com.pojtinger.felicitas.Multiplex.json
```diff
@@ -0,0 +1,648 @@
...
+    "--socket=wayland",
+    "--device=dri",
+    "--filesystem=xdg-download:rw",
+    "--filesystem=xdg-videos:rw",
+    "--filesystem=/tmp:rw",
+    "--talk-name=org.freedesktop.Flatpak"
+  ],
+  "modules": [
+    {
+      "name": "app",
```
hfiguiere, line 20: a more descriptive name is nicer.

com.pojtinger.felicitas.Multiplex.json
```diff
@@ -0,0 +1,648 @@
...
+      "sources": [
+        {
+          "type": "git",
+          "url": "https://github.com/pojntfx/multiplex.git",
+          "tag": "v0.1.0",
+          "commit": "09b08c412c150bfcc13ddc6d82caf3e10fe8a7d6"
+        },
+        {
+          "type": "archive",
+          "url": "https://proxy.golang.org/github.com/anacrolix/torrent/@v/v1.52.5.zip",
```
hfiguiere, line 38: usually the generated manifest like for go are "included". This make it easier for everyone.

pojntfx: Do you have docs on how this is done? I include the vendored modules list (see `modules.txt`), but I don't seem to be able to figure out how to include the sources from a separate file 👀

pojntfx:
> why not building mpv in the flatpak instead?

I added a few of the reasons in the comment on why the spawn is necessary. In summary: I probably wouldn't be able to do the maintenance overhead of vendoring a whole other C app in addition to what is currently a pure Go app, plus making sure that codecs and HA etc. work. On a normal package manager I'd just declare a dependency on `mpv` 🤷‍♀️

com.pojtinger.felicitas.Multiplex.json
pojntfx: Oh, I see - so we need `--share=ipc` in order to use spawn? I tried running it with this manifest, but it seemed to be able to launch IPC w/o issues since the Flatpak DBus talk is allowlisted.

hfiguiere: not. It's for the line quoted above. ^^^^^^^^

image

com.pojtinger.felicitas.Multiplex.json
pojntfx: Switched to "gui" and "dependencies" for the two modules in 94a3be1f512d76bcd40deb8708fe3c38658a4bcc; that should make them a bit clearer hopefully.

pojntfx:
bot, build com.pojtinger.felicitas.Multiplex

com.pojtinger.felicitas.Multiplex.json
pojntfx: nvm, looks like its possible to just provide a file path instead of a `type: include`! See 94a3be1f512d76bcd40deb8708fe3c38658a4bcc

com.pojtinger.felicitas.Multiplex.json
pojntfx: Oh, thanks, now I see! That makes more sense. See ea4adc79ecb01ca57816b45ac5f1950c4e917228

pojntfx:
For the remaining build error, see the explanation on why the permission is required above:

Screenshot of buildbot output

com.pojtinger.felicitas.Multiplex.json
hfiguiere: where is `type: include` documented?

dependencies.json
```diff
@@ -0,0 +1,616 @@
+{
+  "name": "dependencies",
+  "buildsystem": "simple",
+  "build-commands": ["mv vendor /app"],
```
hfiguiere, line 4: that's not how it works.

hfiguiere: remove lines 1 to 5, included, line 5 shuod just be `[`. And remove the last line.

Actually did you use https://github.com/flatpak/flatpak-builder-tools/tree/master/go-get ? If you didn't you should.

This should not be a `module`, but a list of sources.

com.pojtinger.felicitas.Multiplex.json
hfiguiere: really? It's not better at all. A proper name would simply be `multiplex` since that the name of the app.

hfiguiere:
> For the remaining build error, see the explanation on why the permission is required above:
> 
> Screenshot of buildbot output

So https://docs.flathub.org/docs/for-app-authors/linter/

dependencies.json
pojntfx: I tried both that tool and a few others, but from the docs:

image

The tool you mentioned doesn't support Go modules, plus its written in Python, so I used this instead which uses vendor: https://github.com/dennwc/flatpak-go-mod

Both of them seem to recommend just pasting the sources list into the main manifest though. I've checked other Go Flathub apps - they all do it like this, see for example:

- https://github.com/flathub/xyz.diamondb.gtkcord4
- https://github.com/flathub/io.github.noxworld_dev.OpenNox/blob/master/io.github.noxworld_dev.OpenNox.yml

pojntfx:
> > For the remaining build error, see the explanation on why the permission is required above:
> > Screenshot of buildbot output
> 
> So https://docs.flathub.org/docs/for-app-authors/linter/

Oh thanks, that's a good point. I was going by https://github.com/flathub/flathub/wiki/App-Requirements - doesn't mention that there is a linter aside from the one for the metadata.

com.pojtinger.felicitas.Multiplex.json
pojntfx: Why would that be necessary? The module name is relative to the manifest's ID as in `com.pojtinger.felicitas.Multiplex/gui`, `com.pojtinger.felicitas.Multiplex/dependencies`. Doing `com.pojtinger.felicitas.Multiplex/multiplex` feels like its an unnecessary duplication (being that we can use `$FLATPAK_ID` everywhere but here so as to not have to copy-and-paste it around), and it also makes adapting it for another up unnecessarily hard.

If its a convention to do that duplication though then sure I can change it since this is probably close to bikeshedding :)

com.pojtinger.felicitas.Multiplex.json
pojntfx: It isn't. There isn't any docs on how to "include modules", thus the question: https://www.google.com/search?q=include+sources+flatpak+manifest

hfiguiere:
the wiki is obsolete, the new doc is linked from the PR template where you acknowledged to have read it.

dependencies.json
pojntfx: Nvm, figured it out - fixed in ee7a56fd23bf685107ae0fe6b048ba682190dc63

com.pojtinger.felicitas.Multiplex.json
hfiguiere: > Why would that be necessary? 

Becaus when maintainer go awol it's people like myself that end up picking up the pieces. So better make this less unpleasant by having things nicely done from the get go. Call that bikeshedding like you want, I build car garages.

pojntfx:
Hmm, I see. Might make sense to put a disclaimer in the wiki that its archived etc. - showed up in a higher Google ranking than the docs when I was researching requirements before submitting the PR here.

com.pojtinger.felicitas.Multiplex.json
pojntfx: Sure, fixed in ee7a56fd23bf685107ae0fe6b048ba682190dc63

com.pojtinger.felicitas.Multiplex.json
hfiguiere: Like here: 

https://docs.flatpak.org/en/latest/manifests.html#shared-modules

pojntfx: Oh thanks, that's neat! Guess `s/include/shared/g` is what I was missing.

PR #4498, opened 2023-09-09
Add io.github.sunsided.k8sfwd

sunsided:
This adds [k8swfd](https://github.com/sunsided/k8sfwd), a port-forwarding tool and manager for Kubernetes.

- [ ] ~~I am using only the minimal set of permissions.~~ **see below**
- [x] I am an upstream contributor to the project ([sunsided/k8sfwd](https://github.com/sunsided/k8sfwd).

## Permissions

I do not use the minimal set of permissions and instead request `--filesystem=host` and `--allow=devel`.

The main reason is that this application is a wrapper around `kubectl` which is installed as part of the [Google Cloud SDK](https://cloud.google.com/sdk). In addition, my application utilizes a hierarchical configuration file system where configuration files are discovered in the config and home directory, as well as up along the folder hierarchy.

I would prefer to have less permissions but was not sure how to achieve my goals otherwise.


com.github.sunsided.k8sfwd.yml
```diff
@@ -0,0 +1,40 @@
+---
+app-id: com.github.sunsided.k8sfwd
```
hfiguiere, line 2: app should start with `io.github`

com.github.sunsided.k8sfwd.yml
```diff
@@ -0,0 +1,40 @@
...
+runtime: org.freedesktop.Platform
+runtime-version: '22.08'
```
hfiguiere, line 4: current runtime is 23.08

com.github.sunsided.k8sfwd.yml
```diff
@@ -0,0 +1,40 @@
...
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+command: k8sfwd
+finish-args:
+  - "--share=network"
+  # - "--filesystem=xdg-config"
+  # - "--filesystem=home"
+  - "--filesystem=host"
```
hfiguiere, line 13: this is a very broad filesystem permission. Your explanation is not convincing. 

If you need some SDK then it should be in the flatpak.

com.github.sunsided.k8sfwd.yml
```diff
@@ -0,0 +1,40 @@
...
+  - "--allow=devel"
```
hfiguiere, line 14: also why 'devel'. you don't even explain why.

com.github.sunsided.k8sfwd.metainfo.xml
hfiguiere, line 1: since you are upstream, icon, desktop and appstream file should be upstream. There is nothing flatpak specific.

com.github.sunsided.k8sfwd.yml
```diff
@@ -0,0 +1,40 @@
...
+    build-commands:
+      - cargo --offline fetch --manifest-path Cargo.toml --verbose
+      - cargo --offline build --release --verbose
+      - install -Dm755 ./target/release/k8sfwd -t /app/bin/
+      - install -Dm644 ${FLATPAK_ID}.metainfo.xml -t /app/share/metainfo/
+      - install -Dm644 ${FLATPAK_ID}.desktop -t /app/share/applications/
+      - install -Dm644 ${FLATPAK_ID}.svg -t /app/share/icons/hicolor/scalable/apps/
+    sources:
+      - type: archive
+        dest: .
```
hfiguiere, line 31:
```suggestion
```

com.github.sunsided.k8sfwd.svg
hfiguiere, line 1: svg icon is not square. this will not pass validation.

com.github.sunsided.k8sfwd.metainfo.xml
```diff
@@ -0,0 +1,46 @@
...
+            <image>https://raw.githubusercontent.com/sunsided/k8sfwd/6791a5e65c5fcb42357b69104ee0a2d981f62272/readme/screenshot-0.3.0.png</image>
+            <caption>Terminal Output</caption>
+        </screenshot>
+    </screenshots>
+    <url type="homepage">https://github.com/sunsided/k8sfwd</url>
+    <!--
+    Open Age Rating Service
+    https://hughsie.github.io/oars/index.html
+    -->
+    <content_rating type="oars-1.0" />
```
hfiguiere, line 22:
```suggestion
    <content_rating type="oars-1.1" />
```

com.github.sunsided.k8sfwd.metainfo.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+    <id>com.github.sunsided.k8sfwd</id>
+    <metadata_license>CC0</metadata_license>
```
hfiguiere, line 4: this is not a valid SPDX identifier.

com.github.sunsided.k8sfwd.yml
sunsided: It's less about the SDK than about the fact that the tool searches the current directory hierarchy for configuration files. It's a development tool and is supposed to access files outside its sandbox.

`kubectl` can be installed either as part of the Google Cloud SDK (which isn't flatpak'd as far as I know) or separately, depending on what the developer chooses. Specific versions are required for the Kubernetes version; I arguably do not know whether this would work if it could be bundled, but `kubectl` itself isn't flatpak'd either.

com.github.sunsided.k8sfwd.yml
sunsided: Devel is not on purpose. I am trying to sort out which requirements are needed to get the file access I needed and `devel` sounded a lot like "developer tools" - removed it. :)

com.github.sunsided.k8sfwd.yml
sunsided: Fixed.

com.github.sunsided.k8sfwd.yml
sunsided: Fixed.

com.github.sunsided.k8sfwd.yml
sunsided: Fixed.

com.github.sunsided.k8sfwd.metainfo.xml
sunsided: I didn't get that - these files should be in the [sunsided/k8sfwd](https://github.com/sunsided/k8sfwd) repo itself and pulled via the source archive when building?

com.github.sunsided.k8sfwd.svg
sunsided: Content-wise? The viewbox is 500×500.

com.github.sunsided.k8sfwd.metainfo.xml
sunsided: Fixed.

com.github.sunsided.k8sfwd.metainfo.xml
sunsided: Sneaky! Fixed it.

com.github.sunsided.k8sfwd.yml
sunsided: I understand the need for restricted permissions by the way, I simply don't know a better way of doing this. If I can at least clearly announce that this tool accesses the local file system, I would feel better about it.

com.github.sunsided.k8sfwd.svg
hfiguiere: nevermind. Sounds like once again github priorities are not to fix UX (it shows a rectangle canvas).

com.github.sunsided.k8sfwd.yml
hfiguiere: at that point installing your tool as `cargo install` would be better.

Command line tool are not the use case for flatpak, so the ergonomics are far from optimum.

com.github.sunsided.k8sfwd.metainfo.xml
hfiguiere: exactly.

com.github.sunsided.k8sfwd.yml
sunsided: Yeah, I agree. The `cargo install` part requires the toolchain to be installed, making it harder to build (let alone trustworthy if I distribute my own builds). [neovim](https://github.com/flathub/io.neovim.nvim) made a compelling case (including `--filesystem=host`).

com.github.sunsided.k8sfwd.metainfo.xml
sunsided: Can do. I noticed other (e.g. [this one](https://github.com/flathub/io.github.vinegarhq.Vinegar)) bundled it too but I do see your point. Will update.

hfiguiere: two wrongs don't make a right. Also when you are not the upstream developer and upstream doesn't care/want there is no choice.

io.github.sunsided.k8sfwd.svg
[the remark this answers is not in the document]
sunsided: Previously @hfiguiere also mentioned it should be square (which it is, according to my tools), so something does seem to be off indeed. It does look good for me - what is off for you?

io.github.sunsided.k8sfwd.metainfo.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
```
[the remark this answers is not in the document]
sunsided: Updated it to `console-application` but need to figure out now what to do with the `<provides><binary>` section.

io.github.sunsided.k8sfwd.svg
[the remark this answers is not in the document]
sunsided: @barthalion Could you check the icon again? As mentioned earlier I'll move it upstream later but I want to make sure the issues are gone.

PR #4523, opened 2023-09-17
Add io.github.loot.loot

Ortham:

  - The manifest contains comments for all non-standard permissions. They are all due to needing access to find and modify game installs, as LOOT is a game modding utility.


io.github.loot.loot.yml
```diff
@@ -0,0 +1,51 @@
...
+    buildsystem: simple
+    build-commands:
+      - install -D lib/intel64/gcc4.8/libtbb.so.2 /app/lib/
+      - install -D LICENSE /app/share/doc/libtbb2/copyright
+    sources:
+      - type: archive
+        url: https://github.com/oneapi-src/oneTBB/releases/download/v2020.3/tbb-2020.3-lin.tgz
+        sha256: bb8cddd0277605d3ee7f4e19b138c983f298d69fcbb585385b59ef7239d5ef83
+
+finish-args:
```
hfiguiere, line 27: finish-args should be above modules per convention

io.github.loot.loot.yml
```diff
@@ -0,0 +1,51 @@
...
+  - --device=dri
+  - --share=ipc
+  - --share=network
+  - --socket=x11
+  - --socket=wayland
```
hfiguiere, line 32: it either X11 only or wayland + fallback-x11 depending on the situation. (ie wayland support)

io.github.loot.loot.yml
```diff
@@ -0,0 +1,51 @@
...
+modules:
+  - name: application
+    buildsystem: simple
+    build-commands:
+      - install -D bin/LOOT /app/bin/LOOT
+      - install -D lib/libloot.so /app/lib/libloot.so
+      - cp -r share /app
+    sources:
+      - type: archive
+        url: https://github.com/loot/loot/releases/download/0.21.0/loot_0.21.0-Linux.tar.xz
```
hfiguiere, line 15: can't you build from source?

hfiguiere (changes requested):
I see none of the required icon, desktop and appstream file installed.

io.github.loot.loot.yml
Ortham: I'm fairly new to Flatpak and have never successfully used Wayland, but that's the combination that's suggested in Flatpak's documentation: https://docs.flatpak.org/en/latest/qt.html

io.github.loot.loot.yml
```diff
@@ -0,0 +1,51 @@
...
+        sha256: 05e6a4eb1cbb237f507dc39957fe6ac485b7f5f402d351caa1a5fe7dc179c869
+  - name: libtbb
+    buildsystem: simple
+    build-commands:
+      - install -D lib/intel64/gcc4.8/libtbb.so.2 /app/lib/
```
hfiguiere, line 20: really you should build tbb from source.

io.github.loot.loot.yml
Ortham: It takes 10 minutes and a lot of setup. Flatpak is not the primary target for LOOT, it's been a Windows-only application for a decade, and I'm not about to start maintaining a separate copy of my CI setup just to build flatpaks for Flathub, that's just not worth the effort.

io.github.loot.loot.yml
Ortham: Why?

Ortham:
> I see none of the required icon, desktop and appstream file installed.

They're included in the archive that gets extracted.

io.github.loot.loot.yml
Ortham: OK, I've swapped them.

io.github.loot.loot.yml
hfiguiere: This is a question for you to answer. This bit of the documentation need update, it was written when "fallback-x11" didn't exist.

Ortham: OK, I've updated it to use wayland + fallback-x11.

io.github.loot.loot.yml
hfiguiere: then you can build these in the manifest.

io.github.loot.loot.yml
hfiguiere: because it's better.

io.github.loot.loot.yml
Ortham: I would strongly prefer not to, as doing so would make this more of a maintenance burden for me, and would mean that non-Flathub and Flathub Linux builds wouldn't necessarily be the same, which might cause headaches when testing and when dealing with issue reports. I'm also concerned about how I also have no knowledge of or (AFAIK) control over the build environment that Flathub builds run in.

While I was learning how to build Flatpaks I saw many examples of existing packages that don't build from source, so it's clearly not a problem to take the approach I have.

io.github.loot.loot.yml
Ortham: In what way?

This approach integrates nicely with my existing CI pipeline, and I've been producing Flatpak packages as single-file bundles as part of CI and testing them over the past month or so, so I know that this libtbb binary works as expected in a Flatpak package. Building from source just seems like it would increase the complexity and running time of Flathub builds and my own CI pipelines for no obvious benefit.

io.github.loot.loot.yml
hfiguiere: > While I was learning how to build Flatpaks I saw many examples of existing packages that don't build from source,

Two wrongs don't make a right.

hfiguiere:
bot, build io.github.loot.loot

io.github.loot.loot.yml
Ortham: > > While I was learning how to build Flatpaks I saw many examples of existing packages that don't build from source,
> 
> Two wrongs don't make a right.

I'm trying to explain to you why I've taken the approach I have, and this isn't a constructive response, it makes it seem like my PR is not being reviewed in good faith. I've put in a lot of work to get my application working as a Flatpak package, and you're asking me to do more work that will introduce ongoing inconvenience, without any explanation on your part.

This manifest file is a copy of one I've been using to produce Flatpak packages as single-file bundles in CI for the past month or so (and I'll be folding the edits made in response to your review back into the original manifest), so I know the current approach works, and (in addition to the reasons I've already given) I'd rather not introduce discrepancies between Flathub builds and my CI builds.

Ortham:
It looks like the build failed because it was expecting at least one screenshot in the appdata (which I wasn't expecting, I've run it through the validator before with no complaints, but that must be with different settings). I've pushed an update that adds a screenshot.

bot, build io.github.loot.loot

Ortham:
This time failure was due to a trailing period. I've run the same validation command locally and it's now passing for me, so hopefully this will work.

bot, build io.github.loot.loot

Ortham:
I've tried out that build and it seems to work fine.

Ortham:
> /merge

Thanks @barthalion!

PR #4536, opened 2023-09-23
Add so.libdb.gtkcord4

milkknife:

  - https://github.com/flathub/flathub/pull/3857
  - https://github.com/diamondburned/gtkcord4/issues/156
  - https://github.com/flathub/xyz.diamondb.gtkcord4/pull/11


so.libdb.gtkcord4.yml
```diff
@@ -0,0 +1,244 @@
+id: so.libdb.gtkcord4
+runtime: org.gnome.Platform
+runtime-version: '44'
```
hfiguiere, line 3: current is 45.

milkknife: fixed

hfiguiere:
bot, build so.libdb.gtkcord4

so.libdb.gtkcord4.yml
```diff
@@ -0,0 +1,244 @@
...
+modules:
+  - name: gtkcord4
+    buildsystem: simple
+    build-options:
+      append-path: /usr/lib/sdk/golang/bin
+      env:
+        GOBIN: /app/bin
+    build-commands:
+      - go install -v -mod=vendor ./...
+      - desktop-file-edit --set-icon=$FLATPAK_ID --remove-key=DBusActivatable nix/$FLATPAK_ID.desktop
```
bbhtt, line 32: `nix/$FLATPAK_ID.desktop` -> `xyz.diamondb.gtkcord4.desktop`

milkknife:
bot, build so.libdb.gtkcord4

so.libdb.gtkcord4.yml
hfiguiere: the proper fix is to NOT use `$FLATPAK_ID` here. Because inherently the name in he repository is NOT dependent on that value. Only the installed one.

appid.patch
bbhtt, line 1: Just change the name in the manifest. Don't maintain yet another patch.

metainfo.patch
```diff
@@ -0,0 +1,53 @@
...
+@@ -14,10 +14,10 @@
+     </p>
+   </description>
+ 
+-  <launchable type="desktop-id">xyz.diamondb.gtkcord4.desktop</launchable>
++  <launchable type="desktop-id">so.libdb.gtkcord4.desktop</launchable>
+   <screenshots>
+     <screenshot type="default">
+-      <image>https://raw.githubusercontent.com/diamondburned/gtkcord4/main/.github/screenshot2.png</image>
++      <image>https://raw.githubusercontent.com/diamondburned/gtkcord4/main/.github/screenshot4.png</image>
```
bbhtt, line 25: Pull the screenshot from a tag or a commit.

metainfo.patch
bbhtt, line 1: Did you submit these changes upstream?

appid.patch
milkknife: >Just change the name in the manifest. Don't maintain yet another patch. For changing the app-id sed can be used before building.

`sed` would have been so much simpler, I'm not sure why I haven't thought of that

>Also the old id is being used here too

thanks!

>Has the golang part of this patch been upstreamed?

Not yet, which is why the patch exists as a temporary fix until the developer changes the app-id. As soon as the patch is no longer needed, I will remove it.

metainfo.patch
milkknife: Just now, yes: https://github.com/diamondburned/gtkcord4/pull/158

milkknife:
I appreciate the reviews, and that the PR was noticed very quickly. I submitted a PR upstream (https://github.com/diamondburned/gtkcord4/pull/158) which just adds the new metainfo.xml. Once that is merged, I will adjust things here accordingly.

appid.patch
hfiguiere: To change the app-id without patch or `sed` (sed is never a good idea it doesn't have a failure mode):

`appstream-util modify $FLATPAK_DEST/share/metainfo/so.libdb.gtkcord4.metainfo.xml id ${FLATPAK_ID}`

(this assume the file has been renamed, which is currently the case, but if you use `rename-appdata-file` instead then this need to be adapted. the good news it fail in file not found)

bbhtt: > Not yet, which is why the patch exists as a temporary fix

You should submit that also, upstream needs to agree on the app-id and merge the changes so that it is stable and won't change again. 

Preferably, then you can pull from a release or a commit without any need to patch.

> ed is never a good idea it doesn't have a failure mode

That's a good point, thanks. But in `main.go`, the app-id pattern will always exist.

hfiguiere: For the source code a patch is good.

hfiguiere:
bot, build so.libdb.gtkcord4

milkknife:
bot, build so.libdb.gtkcord4

milkknife:
bot, build so.libdb.gtkcord4

milkknife:
bot, build so.libdb.gtkcord4

milkknife:
Right now it's a bit messy, I'm half asleep after work today. I made another PR for upstream which includes another file, and I'll fix things here again. Still not ready yet.

milkknife:
bot, build so.libdb.gtkcord4

milkknife:
bot, build so.libdb.gtkcord4

milkknife:
I tested it on my machine and the build works. I believe it's finally ready. @hfiguiere @bbhtt

I also apologize for spamming with flathubbot, I should have tested it locally first to make sure everything works.

PR #4558, opened 2023-09-28
Add com.budgetwithbuckets.Buckets

iffy:
I've tried to read and understand all the docs, but this is my first experience with flatpaks, so let me know if I did anything wrong.


com.budgetwithbuckets.Buckets.appdata.xml
bbhtt, line 1: This isn't getting installed

bbhtt, line 1: Also it should be upstreamed

com.budgetwithbuckets.Buckets.desktop
```diff
@@ -0,0 +1,6 @@
+[Desktop Entry]
+Name=Buckets
+Exec=com.budgetwithbuckets.Buckets
```
bbhtt, line 3: Should be buckets.sh I think

com.budgetwithbuckets.Buckets.desktop
bbhtt, line 1: Upstream doesn't have a desktop file?

com.budgetwithbuckets.Buckets.yaml
```diff
@@ -0,0 +1,58 @@
...
+        for dir in icons/*; do
+          size=$(basename $dir)
+          install -Dm 644 $dir/apps/com.budgetwithbuckets.Buckets.png -t /app/share/icons/hicolor/$size/apps
+        done
+      - install -Dm 755 buckets.sh /app/bin/buckets.sh
+      - cp -r buckets /app/buckets
+    sources:
+      - type: file
+        path: com.budgetwithbuckets.Buckets.desktop
+      - type: dir
```
bbhtt, line 52: Use file

com.budgetwithbuckets.Buckets.yaml
```diff
@@ -0,0 +1,58 @@
+app-id: com.budgetwithbuckets.Buckets
+runtime: org.freedesktop.Platform
+runtime-version: '22.08'
```
hfiguiere, line 3: 23.08 is the current runtime.

com.budgetwithbuckets.Buckets.yaml
```diff
@@ -0,0 +1,58 @@
...
+    build-commands:
+      - install apply_extra /app/bin/apply_extra
+
+  - name: buckets
+    buildsystem: simple
+    build-commands:
+      - install -Dm 644 com.budgetwithbuckets.Buckets.desktop -t /app/share/applications
+      - |
+        set -x
+        for dir in icons/*; do
```
bbhtt, line 43: You can just install the 512x512 one, appstream-compose will create 64px and 128px icons from it

com.budgetwithbuckets.Buckets.yaml
```diff
@@ -0,0 +1,58 @@
...
+          install -Dm 644 $dir/apps/com.budgetwithbuckets.Buckets.png -t /app/share/icons/hicolor/$size/apps
+        done
+      - install -Dm 755 buckets.sh /app/bin/buckets.sh
+      - cp -r buckets /app/buckets
+    sources:
+      - type: file
+        path: com.budgetwithbuckets.Buckets.desktop
+      - type: dir
+        path: icons
+        dest: icons
```
hfiguiere, line 54: don't use dir source.

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64"]
```

icons/128x128/apps/com.budgetwithbuckets.Buckets.png
hfiguiere, line 1: you don't need more than one icon since they don't come from upstream. Remove this and keep the bigger one.

com.budgetwithbuckets.Buckets.yaml
```diff
@@ -0,0 +1,58 @@
...
+sdk: org.freedesktop.Sdk
+base: org.electronjs.Electron2.BaseApp
+base-version: '22.08'
+command: buckets.sh
+separate-locales: false
+tags:
+  - proprietary
+finish-args:
+  - --device=dri
+  - --filesystem=home
```
hfiguiere, line 13: this is a very broad filesystem permission.

com.budgetwithbuckets.Buckets.yaml
```diff
@@ -0,0 +1,58 @@
...
+  - --socket=x11
+modules:
+  - name: bucketstar
+    buildsystem: simple
+    sources:
+      - type: extra-data
+        filename: buckets.tar.gz
+        only-arches:
+          - x86_64
+        url: https://github.com/buckets/application/releases/download/v0.71.1/Buckets-0.71.1.tar.gz
```
hfiguiere, line 25: does this Electron app contain the vulnerable WebP that was recently exploited?

com.budgetwithbuckets.Buckets.desktop
hfiguiere: to be clearer: this should match what's in `command` in the manifest (maybe with arguments)

com.budgetwithbuckets.Buckets.yaml
bbhtt: `Chrome/76.0.3809.146 Electron/6.1.2` :(

com.budgetwithbuckets.Buckets.desktop
hfiguiere: it's an `extra-data` so it's kind of necessary to have one here.

bbhtt: Ah yea noticed it now

flathub.json
iffy: More archs are coming (in the next release, in fact)

com.budgetwithbuckets.Buckets.yaml
iffy: This is fixed in the upcoming v0.72.0 of the app. Shall I wait for that one to be out?

And while yes, it's vulnerable, it's unlikely to be exploited due to how the app works. It does not load remote content for most users, and for those that do elect to load remote content, they are only navigating to their bank websites. It would be very bad for users, independent of using Buckets, if their bank is hosting malicious webp images.

iffy:
Thanks for reviewing this so quickly! I'll make changes and bring up a Linux VM to actually test everything before I submit again. I tried testing in docker, but it didn't completely work.

flathub.json
refi64: (flathub does have a beta branch where you could put beta manifests prior to their stable release as well)

com.budgetwithbuckets.Buckets.desktop
refi64: ooc is there a particular reason this is extra-data? Usually that's done when flathub doesn't have permission to redistribute the binaries; extra-data will download them on the client's system. But iiuc the submitter in this *is* upstream? So with permission, this could just embed the source directly and avoid duplicating the desktop / metainfo / icons.

iffy:
I've repushed and I think I've addressed everything y'all have mentioned except upstreaming `com.budgetwithbuckets.Buckets.appdata.xml`. I'm not sure where that would live upstream.

com.budgetwithbuckets.Buckets.desktop
```diff
@@ -1,6 +1,6 @@
 [Desktop Entry]
 Name=Buckets
-Exec=com.budgetwithbuckets.Buckets
+Exec=buckets.sh
 Type=Application
-Icon=com.budgetwithbuckets.Buckets
+Icon=../icons/hicolor/512x512/apps/com.budgetwithbuckets.Buckets.png
```
hfiguiere, line 5: this was correct. The change makes it INCORRECT.

hfiguiere: 1. it's never a path
2. it never has an extension.

com.budgetwithbuckets.Buckets.desktop
iffy: Yep, that's what I ended up doing. *Much* simpler :)

com.budgetwithbuckets.Buckets.desktop
iffy: While building with it as it was, it complained with this warning:
Screen Shot 2023-09-28 at 11 28 57 AM

But I can change it back.

iffy: Changed back

hfiguiere:
bot, build com.budgetwithbuckets.Buckets

com.budgetwithbuckets.Buckets.yaml
```diff
@@ -0,0 +1,37 @@
...
+  - --share=ipc
+  - --share=network
+  - --socket=x11
+modules:
+  - name: buckets
+    buildsystem: simple
+    build-commands:
+      - install -Dm 644 com.budgetwithbuckets.Buckets.desktop -t /app/share/applications
+      - install -Dm 644 com.budgetwithbuckets.Buckets.png -t /app/share/icons/hicolor/512x512/apps
+      - install -Dm 755 buckets.sh /app/bin/buckets.sh
```
bbhtt, line 23: appdata is not being installed here

iffy: Oh, I see what you mean now. Okay, pushed a change and now it should be installing in the right place.

PR #4566, opened 2023-09-29
Add site.someones.Lonewolf 

aichingm:

   - `filesystem=home`: Lonewolf references files relative to the project file. `:ro` would allow reading related files but it would not allow to get the "real"  (not/run/...) absolute path of the project file. 

**Note:** Runtime version 42 is the last one which ships with a webkitgtk version which is compatible with tauri. Once tauri updates to a newer webkitgtk version the runtime version can be changed.


site.someones.Lonewolf.yml
```diff
@@ -0,0 +1,62 @@
+app-id: site.someones.Lonewolf
+runtime: org.gnome.Platform
+# Stuck with GNOME 42 until tauri supports newer webkitgtk
+runtime-version: '42'
```
hfiguiere, line 4: thing is this runtime is EOL. You'll need to build webkit or fix tauri.

site.someones.Lonewolf.yml
```diff
@@ -0,0 +1,62 @@
...
+sdk: org.gnome.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node16
+  - org.freedesktop.Sdk.Extension.rust-stable
+command: /app/bin/lonewolf
```
hfiguiere, line 9:
```suggestion
command: lonewolf
```

site.someones.Lonewolf.yml
```diff
@@ -0,0 +1,62 @@
...
+      # build image
+      - (cd lonewolf-tauri; convert -background transparent src/assets/lonewolf.svg -resize 512x512 -format png public/lonewolf.png; convert -background transparent src/assets/icon.svg -resize 128x128 -format png public/icon.png)
+      # build tauri images
+      - (cd lonewolf-tauri; mkdir -p src-tauri/icons; convert -background transparent src/assets/icon.svg -resize 512x512 -format png src-tauri/icons/512x512-lonewolf.png; npm run tauri icon -- src-tauri/icons/512x512-lonewolf.png -o src-tauri/icons)
+      - (cd lonewolf-tauri; npm run --offline tauri build -- -b deb)
+      - install -Dm755 -t /app/bin/ lonewolf-tauri/src-tauri/target/release/bundle/deb/*/data/usr/bin/*
+      - install -Dm644 -t /app/share/applications/ lonewolf-tauri/site.someones.Lonewolf.desktop
+      - mkdir -p /app/share/icons/hicolor/256x256/apps
+      - convert -background transparent lonewolf/assets/icon.svg -resize 256x256 -format png icon.png
+      - mv icon.png /app/share/icons/hicolor/256x256/apps/site.someones.Lonewolf.png
```
hfiguiere, line 60: You really only need to install the SVG.
```suggestion
      - install -Dm644 lonewolf/assets/icon.svg /app/share/icons/hicolor/scalable/apps/site.someones.Lonewolf.svg
```

site.someones.Lonewolf.yml
```diff
@@ -0,0 +1,62 @@
...
+        url: https://github.com/aichingm/lonewolf.git
+        tag: v1.0.2
+        commit: b8f181047be72e5976c1c67e35ed6a314febf7f7
+      - cargo-sources.json
+      - node-sources.json
+    build-commands:
+      - (cd lonewolf-tauri; npm ci --no-audit --offline --cache=/run/build/lonewolf/flatpak-node/npm-cache)
+      - (cd lonewolf-tauri; cargo --offline fetch --manifest-path src-tauri/Cargo.toml --verbose)
+      # build image
+      - (cd lonewolf-tauri; convert -background transparent src/assets/lonewolf.svg -resize 512x512 -format png public/lonewolf.png; convert -background transparent src/assets/icon.svg -resize 128x128 -format png public/icon.png)
```
hfiguiere, line 52: why don't you just use rsvg-convert?

site.someones.Lonewolf.yml
```diff
@@ -0,0 +1,62 @@
+app-id: site.someones.Lonewolf
```
hfiguiere, line 1: `http://someones.site/` doesn't seem to answer.

site.someones.Lonewolf.appdata.xml
```diff
@@ -0,0 +1,69 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>site.someones.Lonewolf</id>
+
+  <name>Lonewolf</name>
+  <summary>Organize and track your tasks with ease and flexibility.</summary>
```
hfiguiere, line 6:
```suggestion
  <summary>Organize and track your tasks with ease and flexibility</summary>
```

site.someones.Lonewolf.appdata.xml
hfiguiere, line 1: this file should be managed upstream.

site.someones.Lonewolf.yml
hfiguiere: just to clarify. We don't accept new submissions with EOL runtime.

Build this with GNOME 45.

site.someones.Lonewolf.appdata.xml
```diff
@@ -0,0 +1,69 @@
...
+  </screenshots>
+
+  <releases>
+    <release type="stable" version="1.0.0" date="2023-09-25">
+      <description>
+        <p>Lonewolf's first release</p>
+      </description>
+    </release>
+  </releases>
+  <content_rating type="oars-1.0"/>
```
hfiguiere, line 68:
```suggestion
  <content_rating type="oars-1.1"/>
```

site.someones.Lonewolf.appdata.xml
```diff
@@ -0,0 +1,69 @@
...
+      <li>Split up larger tasks with check lists</li>
+      <li>Keep your thoughts by adding comments for your future self</li>
+    </ul>
+  </description>
+
+  <launchable type="desktop-id">site.someones.Lonewolf.desktop</launchable>
+
+  <screenshots>
+    <screenshot type="default">
+      <image>https://raw.githubusercontent.com/flathub/site.someones.Lonewolf/master/assets/screenshot_light_main.png</image>
```
hfiguiere, line 44: this URL doesn't exist. That mean the validation WILL fail.

site.someones.Lonewolf.yml
aichingm: Yeah, I know. I don't host anything there at the moment.

Ownership can be checked with:
```
nslookup -type=TXT lonewolf.someones.site 8.8.8.8
```
```
Server:         8.8.8.8
Address:        8.8.8.8#53

Non-authoritative answer:
Lonewolf.someones.site  text = "github.com/aichingm/lonewolf"
```

site.someones.Lonewolf.appdata.xml
aichingm: I see. I added the screen shots in anticipation of the repository being created. I will just comment them out.

site.someones.Lonewolf.appdata.xml
aichingm: Interesting, is there a reason why not 1.0? My understanding was that 1.1 is only used if  works with GOME software but not other clients.

site.someones.Lonewolf.appdata.xml
hfiguiere: that will fail too. at least ONE screenshot is required.

site.someones.Lonewolf.appdata.xml
aichingm: Once upstreamed do I need to let flatpak/flathub know somehow where to find it?

hfiguiere: it always need to be installed, so the build system should install it. Like everything else.

aichingm: Ah, I see. I missed that, thx!

site.someones.Lonewolf.appdata.xml
aichingm: Oh. Since the xml should be upstreamed I can also add the screenshots upstream.

assets/Showcase.lwp
hfiguiere, line 1: what is this for?

site.someones.Lonewolf.appdata.xml
hfiguiere: 1.1 is the latest. not sure where you get that idea

assets/Showcase.lwp
aichingm: Just the data file from which the screenshots are taken. I can also move the upstream.

site.someones.Lonewolf.appdata.xml
hfiguiere: I missed that it was not installed in the forest of install lines.

site.someones.Lonewolf.yml
```diff
@@ -0,0 +1,62 @@
...
+    build-commands:
+      - (cd lonewolf-tauri; npm ci --no-audit --offline --cache=/run/build/lonewolf/flatpak-node/npm-cache)
+      - (cd lonewolf-tauri; cargo --offline fetch --manifest-path src-tauri/Cargo.toml --verbose)
+      # build image
+      - (cd lonewolf-tauri; convert -background transparent src/assets/lonewolf.svg -resize 512x512 -format png public/lonewolf.png; convert -background transparent src/assets/icon.svg -resize 128x128 -format png public/icon.png)
+      # build tauri images
+      - (cd lonewolf-tauri; mkdir -p src-tauri/icons; convert -background transparent src/assets/icon.svg -resize 512x512 -format png src-tauri/icons/512x512-lonewolf.png; npm run tauri icon -- src-tauri/icons/512x512-lonewolf.png -o src-tauri/icons)
+      - (cd lonewolf-tauri; npm run --offline tauri build -- -b deb)
+      - install -Dm755 -t /app/bin/ lonewolf-tauri/src-tauri/target/release/bundle/deb/*/data/usr/bin/*
+      - install -Dm644 -t /app/share/applications/ lonewolf-tauri/site.someones.Lonewolf.desktop
```
hfiguiere, line 57: I bet the icon is wrongly set in there.

site.someones.Lonewolf.appdata.xml
hfiguiere: definitely. 

and the docker thing-a-mabob.

site.someones.Lonewolf.appdata.xml
aichingm: The second dropdown on the generator page: https://hughsie.github.io/oars/generate.html

site.someones.Lonewolf.yml
aichingm: What icon do you mean exactly?

site.someones.Lonewolf.appdata.xml
aichingm: Would it be a problem to keep it in this repo, since it is used by the Makefile to run `flatpak-builder` locally.?

site.someones.Lonewolf.yml
aichingm: I was not aware of rsvg-convert.

site.someones.Lonewolf.yml
hfiguiere: I checked and it's correct (the build would fail anyway)

site.someones.Lonewolf.appdata.xml
hfiguiere: it needs to be updated.

site.someones.Lonewolf.appdata.xml
hfiguiere: why do you need Docker?

aichingm: I don't actually need docker podman is fine too. I like to compartmentalize build environments so that I can easily run them on different machines without the need to replicate the build environment on all of them and keep them in sync.

site.someones.Lonewolf.yml
```diff
@@ -0,0 +1,57 @@
...
+  - --device=dri
+  - --filesystem=home
+
+modules:
+  - name: lonewolf
+    buildsystem: simple
+    build-options:
+      append-path: /usr/lib/sdk/node16/bin:/usr/lib/sdk/rust-stable/bin
+      env:
+        npm_config_nodedir: /usr/lib/sdk/node16
```
bbhtt, line 24: Node16 is EOL too and doesn't have a branch for 23.08/45, does this work with Node18?

site.someones.Lonewolf.yml
aichingm: is this an issue or can I mark this as resolved?

hfiguiere:
bot, build site.someones.Lonewolf

site.someones.Lonewolf.yml
hfiguiere: you should really put a page there. It will be necessary to serve a file for verification purposes. (it's optional though)

site.someones.Lonewolf.yml
```diff
@@ -0,0 +1,91 @@
...
+    #no-parallel-make: true
+    config-opts:
+      - -DPORT=GTK
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DENABLE_DOCUMENTATION=OFF
+      - -DENABLE_GAMEPAD=OFF
+      - -DUSE_SOUP2=ON
+      - -DENABLE_BUBBLEWRAP_SANDBOX=OFF
+    modules:
+      - name: libsoup-2.4
```
hfiguiere, line 47: you should use the one from shared-modules.

site.someones.Lonewolf.yml
aichingm: Ok, I think I will tackle this once I find time for the verification stuff.

bbhtt:
bot, build site.someones.Lonewolf

bbhtt:
> filesystem=home: Lonewolf references files relative to the project file. :ro would allow reading related files but it would not allow to get the "real" (not/run/...) absolute path of the project file.

I'm wondering whether you can restrict this to xdg-document or another folder by default instead of full home. Users can always add their own locations through overrides. And the filepicker seems to use portals already.

aichingm:
> I'm wondering whether you can restrict this to xdg-document or another folder by default instead of full home. Users can always add their own locations through overrides. And the filepicker seems to use portals already.

There are a few issues I found for which I don't have fixes at the moment but maybe there are solutions I don't know about yet. 

* With the permission is set to `--filesystem=xdg-documents`, if a user opens a file outside of `xdg-documents`, attachments can't be displayed (attachments like images are linked in markdown text via ``)
* Adding files as attachments via the portal would also yield only `/run` paths which the application could than not convert to working relative directories. 
* With the permission is set to `--filesystem=xdg-documents`, if a user opens a project via the portal the files path is `/run/...`, this does not allow the program to open the same file after a relaunch. 
* `--filesystem=home:ro` could work to read the attachments but then the program will only receive `/run/...` path from the portal file chooser and from this path I can't calculate the relative assets because I don't know the "original" path of the project. Or is there an api with which I can translate the /run paths back to "real" paths?
* If I restrict the permissions to something too limited there is to my understanding no api with which I could "request" permissions to a specific from flatpak/a user. Or am I just nor aware of it?
* I could check if the program receives a `/run` path and tell the user to add `rw` permissions for the path (which I don't know) to the flatpak
   * So I would need to display a dialog like this:
   > This path is not readable by lonewolf.  To allow lonewolf to work with this this project run this command `flatpak override site.someones.lonewolf --filesystem=<path_to_the_directory_of_the_project_here_sorry_cant_help_you_to_figure_this_one_out>` or add the directory of the project via your platforms flatpak management tool (if installed) could be flatseal, KDEs 'Flatpak Permission System' settings panel, and so on..., don't forget to restart the application to reload the permissions.
   * Alternatively I could just ask the user to set it to `home` on first run anyway which wouldn't differ from setting it to `home` by default...

Am I missing something or is this just the way it is?

bbhtt:
Yea, some of those cases will require more permission than a single folder.

aichingm:
@hfiguiere  @bbhtt Thanks for the guidance and bearing with me!

PR #4570, opened 2023-09-30
Add io.github.najepaliya.kleaner

najepaliya:

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://github.com/najepaliya/kleaner


com.github.najepaliya.kleaner.json
```diff
@@ -0,0 +1,110 @@
+{
+    "id": "com.github.najepaliya.kleaner",
```
hfiguiere, line 2: app id MUST start by `io.github`

com.github.najepaliya.kleaner.json
```diff
@@ -0,0 +1,110 @@
...
+    "runtime": "org.kde.Platform",
+    "runtime-version": "5.15-22.08",
+    "sdk": "org.kde.Sdk",
+    "command": "kleaner",
+    "finish-args": [
+        "--device=dri",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--filesystem=host"
```
hfiguiere, line 12: this is a very broad filesystem permission.

com.github.najepaliya.kleaner.json
```diff
@@ -0,0 +1,110 @@
...
+                        "project-id": 8763,
+                        "stable-only": true,
+                        "url-template": "https://download.kde.org/stable/release-service/$version/src/libkexiv2-$version.tar.xz"
+                    }
+                }
+            ]
+        },
+        {
+            "name": "kleaner",
+            "buildsystem": "cmake",
```
hfiguiere, line 98: cmake-ninja?

com.github.najepaliya.kleaner.json
```diff
@@ -0,0 +1,110 @@
...
+    "sdk": "org.kde.Sdk",
+    "command": "kleaner",
+    "finish-args": [
+        "--device=dri",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--filesystem=host"
+    ],
+    "modules": [
```
hfiguiere, line 14: probably missing cleanup rules to remove build only file like includes and pkgconfig files, and also .a

com.github.najepaliya.kleaner.json
najepaliya: What would be the suggested change? Only $XDG_PICTURES_DIR?

com.github.najepaliya.kleaner.json
najepaliya: Yes, will fix.

com.github.najepaliya.kleaner.json
najepaliya: Yes, will fix.

com.github.najepaliya.kleaner.json
najepaliya: Yes, will fix.

com.github.najepaliya.kleaner.json
hfiguiere: since it's a Qt app (KDE?) doesn't it use portals? (it's supposedly automatic)

najepaliya: I have gone ahead and made the suggested changes, please review again.

hfiguiere:
the manifest name must match the app-id.

hfiguiere:
bot, build io.github.najepaliya.kleaner

najepaliya:
Seems like I need to rename the appdata file as well?

hfiguiere:
of course. and icons. eveything should match.

najepaliya:
Another go?

hfiguiere:
you can also

bot, build io.github.najepaliya.kleaner

najepaliya:
Looks to be working as intended.

io.github.najepaliya.kleaner.json
```diff
@@ -0,0 +1,112 @@
...
+    "sdk": "org.kde.Sdk",
+    "command": "kleaner",
+    "finish-args": [
+        "--device=dri",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--filesystem=xdg-pictures"
+    ],
+    "modules": [
```
bbhtt, line 14: Add cleanups: `/lib/cmake`, `/lib/pkgconfig`, `/share/man`.

bbhtt:
You should probably add a better indication for opening the filepicker. It's invisible unless hovered and low-contrast when hovered :) Took me a while to test it.

<details>
<summary> Image </summary>

e7ec089d66
</details>

najepaliya:
> You should probably add a better indication for opening the filepicker. It's invisible unless hovered and low-contrast when hovered :) Took me a while to test it.
> 
> Image
> e7ec089d66

Woah. This is my first time seeing this. In the code, I don't actually set the color of anything as the QQC2 style should take over. I am not sure how to handle this.

io.github.najepaliya.kleaner.json
najepaliya: Done.

bbhtt:
> So either qqc2-desktop-style package wasn't build at all, or breeze style was somehow built without QML support, or some pieces of plasma-integration are missing. As a last resort, there always environment variables for that too, but you'd better figure it out without messing up with env.

It's on GNOME, and I don't think packages on host would interfere with flatpaks.

Anyway I see some of the same issues with the apps you mentioned. This app and the other ones work fine when forcing `--env=XDG_CURRENT_DESKTOP=KDE`

najepaliya:
@bbhtt would you like me to add the variable or dbus permission to finish-args?

bbhtt:
> @bbhtt do you prefer adding the environment variable or the dbus permission to finish-args?

There's no dbus permission I know that will affect this. The environment variable is your choice. But this shouldn't require anything to be done, ideally.

bbhtt:
Ok, I think I figured out the issue. In your application, you aren't setting a "default" style anywhere. Per the documentation if you want to use qqc2-desktop-style as the default one you should do something like this https://api.kde.org/frameworks/qqc2-desktop-style/html/index.html. You can set it to something else by looking at `QQuickStyle::availableStyles();`

When the desktop environment is set to KDE, Qt probably through some magic automatically picks up the qqc2-desktops-style but not on GNOME even if it is included in the KDE runtime.

So simply:

```patch

diff --git a/src/main.cpp b/src/main.cpp
index 2546df1..fa0bce6 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2,6 +2,7 @@
 #include <QQmlApplicationEngine>
 #include <QtQml>
 #include <QUrl>
+#include <QQuickStyle>
 #include <KLocalizedContext>
 #include <KLocalizedString>
 #include <Kleaner.hpp>
@@ -17,6 +18,8 @@ int main(int argc, char *argv[])
 
     QGuiApplication::setDesktopFileName("com.github.najepaliya.kleaner");
 
+    QQuickStyle::setStyle("org.kde.desktop");
+
     QQmlApplicationEngine engine;
 
     Kleaner kleaner;
@@ -30,4 +33,4 @@ int main(int argc, char *argv[])
     }
 
     return app.exec();
```

seems to fix it for me. This also probably explains why it defaults to a basic QtQuickControls style.

najepaliya:
If it does fix it for you then I definitely don't mind adding it. I think the hellworld example in the docs will also need to be updated in that case.

ratijas:
> Qt probably through some magic automatically picks up the qqc2-desktops-style

I believe this magic is called QPT — Qt Platform Theme, and our implementation in a form of plasma-integration project takes care of so much more than just [setting QtQuick.Controls style](https://invent.kde.org/plasma/plasma-integration/-/blob/a4794775b793fbc9005e352cb395f16c80e23266/qt6/src/platformtheme/kdeplatformtheme.cpp#L582)

hfiguiere:
> ```diff
>      QGuiApplication::setDesktopFileName("com.github.najepaliya.kleaner");
> ```

This is incorrect. The id doesn't match the one here.
Also 

> This is the file name, without the full path

So it's missing `.desktop`.

najepaliya:
This hopefully fixes all the issues discussed.

bbhtt:
bot, build io.github.najepaliya.kleaner

bbhtt:
> Build [70987 successful](https://buildbot.flathub.org/#/builders/6/builds/70987) To test this build, install it from the testing repository:
> 
> ```
> flatpak install --user https://dl.flathub.org/build-repo/53643/io.github.najepaliya.kleaner.flatpakref
> ```

Works now

<details>
<summary> Image </summary>

image

</details>

ratijas:
I skimmed though app's UI code. Some quick feedback:

- I don't see licenses mentioned anywhere;
- Buttons are not accessible. They need to have `text` property set.  If you don't want to see text at all times, use `display: T.AbstractButton.IconOnly` mode, and add an attached ToolTip object with relevant bindings
    ```qml
    QQC2.ToolTip.visible: Kirigami.Settings.tabletMode ? pressed : hovered
    QQC2.ToolTip.delay: Kirigami.Settings.tabletMode ? Qt.styleHints.mousePressAndHoldInterval : Kirigami.Units.toolTipDelay
    ```
- Hardcoded `activeBackgroundColor: "lightblue"` is not needed, and in fact makes list items unreadable with dark color themes
- Consider using `QQC2.RadioDelegate` for templates list;
- Set elide property on all labels, add tooltips when `thuncated` if needed;
- Consider ElideMiddle for file paths;
- > i18n(display)
  
  This won't work. i18n is not needed here;
- Wrap ListView in a ScrollView to get scrollbars and better desktop integration. Then you won't need GroupBox, and there won't be any weird paddings on the sides.
- > ```qml
  >     Kirigami.Action {
  >         displayComponent: Controls.Button {
  > ```

  Empty Action with a displayComponent that mimics ToolButton's flat style. Why?  Just set properties on action directly.

najepaliya:
Very helpful. I'll make sure to add these before the merge. I honestly can't remember why I used the displayComponent property

hfiguiere:
bot, build io.github.najepaliya.kleaner

najepaliya:
Still making some last second changes. It should be ready soon.

najepaliya:
Ok, I think these are all the changes I will be making for 1.0.

najepaliya:
bot, build io.github.najepaliya.kleaner

najepaliya:
This is the old build... and I pushed to the right branch.

najepaliya:
bot, build io.github.najepaliya.kleaner

najepaliya:
Yeah this seems correct now.

najepaliya:
bot, build io.github.najepaliya.kleaner

najepaliya:
Can I get a final test/review?

ratijas:
UI structure is fine now.

For the internationalization,

- `icon.name: i18n("list-add")` icon names are not translatable :)
- string interpolation should be used for messages with parameters, otherwise you won't have any static string to translate! Also, don't do [word puzzles](https://techbase.kde.org/Development/Tutorials/Localization/i18n_Mistakes#Pitfall_#2:_Word_Puzzles): they are not translatable in many locales. Don't:
   ```js
   resultMessage.text = i18n("Template \"" + radioGroup.checkedButton.name + "\" applied to " + Kleaner.processFiles(radioGroup.checkedButton.index))
   ```
   Do:
   ```js
   var templateName = radioGroup.checkedButton.name;
   var templateIndex = radioGroup.checkedButton.index;
   var result = Kleaner.processFiles(templateIndex);
   resultMessage.text = i18n("Template \"%1\" applied to %2 file(s) with %3 metadata clearing/saving failure(s) and %4 metadata loading failure(s).", templateName, result.count, result.metaSaveFailedCount, result.metaLoadFailedCount);
   ```
   with `result` being some ad-hoc data structure with a bunch of fields. `Q_GADGET` or even QVariantMap would be suitable for that.
   due to the lack of multi-pluralization-in-one-message, you'd have to deal with "file(s)" kind of suffixes, or if you are feeling creative — experiment with splitting the message into multiple sentences, and use i18np flavor. Just keep in mind that not all languages even use a dot and a whitespace as separators for sentences :)
- It is known that ListElement is hostile toward any non-const data —  even for translatable strings. To use i18n there are couple of options:
  * use plain javascript array — probably wouldn't work with property updates as you'd expect it to
    ```qml
    model: [
        { name: i18n("Clear all EXIF"), checked: true },
        { name: i18n("Clear all IPTC"), checked: false },
        { name: i18n("Clear all XMP"), checked: false },
        { name: i18n("Clear all comments"), checked: false },
    ]
    ```
  * Use [`qsTrNoOp`](https://doc.qt.io/qt-6/i18n-source-translation.html#qml-use-noop-functions) which magically works with ListModel — not sure it would work out with i18n, and I couldn't find any counterpart in KI18n (they wouldn't be supported by ListModel anyway even if they existed)
  * Populate list model with entries imperatively — this should just work™:
   ```qml
    model: ListModel {
        id: templateModel

        Component.onCompleted: {
            append({ name: i18n("Clear all EXIF"), checked: true });
            append({ name: i18n("Clear all IPTC"), checked: false });
            append({ name: i18n("Clear all XMP"), checked: false });
            append({ name: i18n("Clear all comments"), checked: false });
        }
    }
   ```

najepaliya:
Done.

najepaliya:
bot, build io.github.najepaliya.kleaner

io.github.najepaliya.kleaner.json
```diff
@@ -0,0 +1,115 @@
...
+    "runtime": "org.kde.Platform",
+    "runtime-version": "5.15-22.08",
+    "sdk": "org.kde.Sdk",
+    "command": "kleaner",
+    "finish-args": [
+        "--device=dri",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--filesystem=xdg-pictures"
```
bbhtt, line 12: If you can use [QFileDialog](https://doc.qt.io/qt-5/qfiledialog.html) for the filechooser, it'd benefit from the portal and allow users add files from directories other than this.

Can be post merge on flathub, just a note.

najepaliya: I have tried to keep all the UI elements in QML. This might be something I implement for the next release though. I'm probably going to need a  refactor for that anyway

ratijas: I don't exactly remember which is which but there's also another FileDialog type in `import Qt.labs.platform 1.1`, and also `QtQuick.Dialogs` were somewhat refactored in Qt 6 from what I heard.

PR #4602, opened 2023-10-12
Add io.github.pd_l2ork.Pd_L2Ork

pd-l2ork:

The original project (in development since 2009) exists at https://github.com/pd-l2ork/pd-l2ork

There is only one patch to the build script that removes rsync and instead uses install command, in order to build the core executable.

There are no patches that need to be submitted upstream. Whatever changes have been made to the build process are unique to the needs of building a flatpak package.

One thing I had noticed when trying to build on my own (outdated) laptop, is that the flatpak-builder would randomly exit with code 9 after having compiled for a while (this is a big project, so it will take a while to compile--on my laptop it took little over an hour). I had to replace make `-j ${FLATPAK_BUILDER_N_JOBS}` with only `make`, which fixed the problem. However, looking at other similar manifests, it seems that including -j for parallel compiling is advised. I am unsure why this fails on my laptop. It could be that my laptop is too old, and/or it may be because I am running lowlatency kernel (for audio needs). Either way, please let me know if you encounter trouble and I will promptly update the make script reverting these calls back to `make`.

Also, note since this project pulls from a number of submodules and combines them into one build, git needs to allow file call. Below is the command I run. Not sure if this is allowed/necessary to configure on the flathub side of things:

`git config --global --add protocol.file.allow always`

Please see README.md for additional info, and let me know if you have any questions. Thank you.


shared/glu-9.json
hfiguiere, line 1: what's wrong with the one in the shared-modules.

pd-wrapper.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+pd-l2ork -alsa -alsaadd pulse $@
```
hfiguiere, line 3:
```suggestion
exec pd-l2ork -alsa -alsaadd pulse "$@"
```

net.pdl2ork.PdL2Ork.yml
```diff
@@ -0,0 +1,199 @@
...
+  - install -m 644 net.pdl2ork.PdL2Ork.desktop /app/share/applications/
+  # AppStream metadata (for app store / package manager)
+  - install -d /app/share/metainfo/
+  - install -m 644 net.pdl2ork.PdL2Ork.metainfo.xml /app/share/metainfo
+  # icon (for everyone)
+  - install -d /app/share/icons/hicolor/scalable/apps
+  - install -m 644 net.pdl2ork.PdL2Ork.svg /app/share/icons/hicolor/scalable/apps/
+  # add mime type to .pd files
+  - install -d /app/share/mime/packages/
+  - install -m 644 packages/linux_make/pd-l2ork.xml /app/share/mime/packages/net.pdl2ork.PdL2Ork.xml
```
hfiguiere, line 179:
```suggestion
  - install -Dm755 pd-wrapper.sh /app/bin
  # .desktop file (so we appear in list of applications)
  - install -Dm644 net.pdl2ork.PdL2Ork.desktop -t /app/share/applications/
  # AppStream metadata (for app store / package manager)
  - install -Dm644 net.pdl2ork.PdL2Ork.metainfo.xml -t /app/share/metainfo
  # icon (for everyone)
  - install -Dm644 net.pdl2ork.PdL2Ork.svg -t /app/share/icons/hicolor/scalable/apps/
  # add mime type to .pd files
  - install -Dm644 packages/linux_make/pd-l2ork.xml /app/share/mime/packages/net.pdl2ork.PdL2Ork.xml
```

net.pdl2ork.PdL2Ork.yml
```diff
@@ -0,0 +1,199 @@
+id: net.pdl2ork.PdL2Ork
+runtime: org.freedesktop.Platform
+runtime-version: '22.08'
```
hfiguiere, line 3:
```suggestion
runtime-version: '23.08'
```

net.pdl2ork.PdL2Ork.yml
```diff
@@ -0,0 +1,199 @@
...
+# Filesystem access (pd doesn't yet work with portals)
+  - --filesystem=host
+# MIDI access (Flatpak doesn't yet support exposing only MIDI devices)
+  - --device=all
+# Linux Audio plugins
+  - --env=LADSPA_PATH=/app/extensions/Plugins/ladspa:/app/lib/ladspa
+add-extensions:
+  org.freedesktop.LinuxAudio.Plugins:
+    directory: extensions/Plugins
+    version: '22.08'
```
hfiguiere, line 26:
```suggestion
    version: '23.08'
```

net.pdl2ork.PdL2Ork.yml
```diff
@@ -0,0 +1,199 @@
...
+  - --share=ipc
+  - --socket=x11
+  - --socket=wayland
+  - --device=dri
+# Audio
+  - --socket=pulseaudio
+  - --system-talk-name=org.freedesktop.RealtimeKit1
+  - --filesystem=xdg-run/pipewire-0
+# Filesystem access (pd doesn't yet work with portals)
+  - --filesystem=host
```
hfiguiere, line 18: can we limit to home at least?

net.pdl2ork.PdL2Ork.yml
```diff
@@ -0,0 +1,199 @@
...
+command: pd-wrapper.sh
+finish-args:
+# Display
+  - --share=ipc
+  - --socket=x11
+  - --socket=wayland
+  - --device=dri
+# Audio
+  - --socket=pulseaudio
+  - --system-talk-name=org.freedesktop.RealtimeKit1
```
hfiguiere, line 15:
```suggestion
```

net.pdl2ork.PdL2Ork.yml
```diff
@@ -0,0 +1,199 @@
...
+runtime: org.freedesktop.Platform
+runtime-version: '22.08'
+sdk: org.freedesktop.Sdk
+default-branch: stable
+command: pd-wrapper.sh
+finish-args:
+# Display
+  - --share=ipc
+  - --socket=x11
+  - --socket=wayland
```
hfiguiere, line 11: does it support wayland? If not, this should be removed. Otherwise line 10 is `fallback-x11`

net.pdl2ork.PdL2Ork.metainfo.xml
```diff
@@ -0,0 +1,49 @@
...
+      easily work over local and remote networks to integrate wearable
+      technology, motor systems, lighting rigs, and other equipment. It
+      is suitable for learning basic multimedia processing and visual
+      programming methods as well as for realizing complex systems for
+      large-scale projects.
+    </p>
+  </description>
+  <screenshots>
+    <screenshot type="default">
+      <image type="source">screenshots/L2Ork-Tweeter.png</image>
```
hfiguiere, line 36: this won't work. It must be a URL. Reachable at the time of build.

net.pdl2ork.PdL2Ork.metainfo.xml
```diff
@@ -0,0 +1,49 @@
...
+      <image type="source">screenshots/K12-Demo.png</image>
+      <image type="source">screenshots/Image-Help-File.png</image>
+    </screenshot>
+  </screenshots>
+  <developer_name>Ivica Ico Bukvic</developer_name>
+  <launchable type="desktop-id">net.pdl2ork.PdL2Ork.desktop</launchable>
+  <url type="homepage">https://bit.ly/pd-l2ork</url>
```
hfiguiere, line 43: please don't use URL shortener here.

net.pdl2ork.PdL2Ork.metainfo.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>net.pdl2ork.PdL2Ork</id>
```
hfiguiere, line 3: `pdl2ork.net` doesn't seem to exist. This must be a domain under the control of the ustream developers.

net.pdl2ork.PdL2Ork.yml
```diff
@@ -0,0 +1,199 @@
...
+  # add mime type to .pd files
+  - install -d /app/share/mime/packages/
+  - install -m 644 packages/linux_make/pd-l2ork.xml /app/share/mime/packages/net.pdl2ork.PdL2Ork.xml
+  sources:
+  - type: git
+    url: https://github.com/pd-l2ork/pd-l2ork/
+    tag: '20231012'
+    commit: 27b66ed5efd4de7cb0fd7ad33cf46c60cb8201fc
+  - type: file
+    # x86_64 for now only. Still waiting for a comparable arm7 version for RPi.
```
hfiguiere, line 186:
```suggestion
    # x86_64 for now only. Still waiting for aarch64.
```

net.pdl2ork.PdL2Ork.yml
```diff
@@ -0,0 +1,199 @@
...
+  - type: file
+    path: ./pd-wrapper.sh
+  - type: file
+    path: net.pdl2ork.PdL2Ork.desktop
+  - type: file
+    path: net.pdl2ork.PdL2Ork.metainfo.xml
+  - type: file
+    path: net.pdl2ork.PdL2Ork.svg
+  - type: file
+    path: default.settings
```
hfiguiere, line 199: since these files are in the repo upstream, why having theme here?

net.pdl2ork.PdL2Ork.metainfo.xml
```diff
@@ -0,0 +1,49 @@
...
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>GPL-3.0+ and BSD-3-Clause</project_license>
```
hfiguiere, line 5: I don't think this will pass. It expects `AND` in uppercase.

hfiguiere: Also `GPL-3.0+` is deprecated and it should be `GPL-3.0-or-later`

(I don't think we fail on this yet)

shared/glu-9.json
pd-l2ork: Thanks for your feedback. It fails to build on certain older systems, like Ubuntu 22.04, likely due to the following bug: https://github.com/flatpak/flatpak-builder/issues/468

Given the complexity of the build process, I have tested in 22.08 and transitioning to 23.08 may require another month of testing, so we could focus on that transition as part of the next release. Hence, the aforesaid issue will remain an issue for the current cycle.

net.pdl2ork.PdL2Ork.yml
pd-l2ork: Please see my comment. I would prefer if we left it at 22.08, as testing 23.08 will require a good month, given the complexity of the build process and its reliance on a number of other submodules.

shared/glu-9.json
bbhtt: > Thanks for your feedback. It fails to build on certain older systems, like Ubuntu 22.04, likely due to the following bug: https://github.com/flatpak/flatpak-builder/issues/468

Download the archive from your browser, go to the `.flatpak-builder/downloads/<sha256 of glue>`, delete the leftover file and copy-paste it there.

hfiguiere: Use the flatpak-builder shipped as a flatpak.
```
flatpak install org.flatpak.Builder
```

pd-l2ork: > > Thanks for your feedback. It fails to build on certain older systems, like Ubuntu 22.04, likely due to the following bug: https://github.com/flatpak/flatpak-builder/issues/468
> 
> Download the archive from your browser, go to the `.flatpak-builder/downloads/<sha256 of glue>`, delete the leftover file and copy-paste it there.
>

Where do I paste it? Inside the yml file? Would you please provide full syntax? I am still new to this and already spent way too much time trying to learn ins and outs of building a flatpak.
 
> It will work in Flathub CI so you don't need this.

Do I have access to this, or is this Flathub does on its own once deployed?

bbhtt: > Given the complexity of the build process, I have tested on 22.08 for the past month and transitioning to 23.08 will likely require another month 

What exactly is taking so long? Building it should be one command and testing shouldn't require that much time.

pd-l2ork: > > Thanks for your feedback. It fails to build on certain older systems, like Ubuntu 22.04, likely due to the following bug: https://github.com/flatpak/flatpak-builder/issues/468
> 

Thanks. Please see my comments below.

> Download the archive from your browser, go to the `.flatpak-builder/downloads/<sha256 of glue>`, delete the leftover file and copy-paste it there.
> 

Would you please give me a clear step-by-step what you mean? I am new to flatpak and already spent 2 weeks wading through sparse documentation only to get this far.

> It will work in Flathub CI so you don't need this.

So, are you saying that when run via flathub, this should not be an issue? What about if I build it locally to test? I would prefer implementing something that works in both places.

Thanks.

net.pdl2ork.PdL2Ork.yml
pd-l2ork: How do I pull this commit into my branch once I merge it as part of this pull request?

net.pdl2ork.PdL2Ork.yml
pd-l2ork: How would I do that?

net.pdl2ork.PdL2Ork.yml
pd-l2ork: What would this do? I saw this requirement on flatpaks of other forks of this project, but am not entirely sure why this is needed, other than the fact that audio is central to this application.

pd-l2ork: Looks like this may give rt permission to the application, which, if so, is desirable to ensure low-latency operation audio-wise.

net.pdl2ork.PdL2Ork.yml
pd-l2ork: It uses node webkit which I think is in part reliant on the wayland.

net.pdl2ork.PdL2Ork.metainfo.xml
pd-l2ork: I will update this.

net.pdl2ork.PdL2Ork.yml
hfiguiere: I don't think so.

net.pdl2ork.PdL2Ork.yml
hfiguiere: It will NOT work. Remove it. Those that still have it, it's a mistake. Two wrong don't make a right.

net.pdl2ork.PdL2Ork.metainfo.xml
pd-l2ork: I will fix this shortly.

net.pdl2ork.PdL2Ork.yml
hfiguiere: you are mistaken. 

This is not for device access.

net.pdl2ork.PdL2Ork.metainfo.xml
pd-l2ork: I see. I was simply going by what another fork had on flathub. My URL is l2ork.music.vt.edu. Should I do edu.vt.music.PdL2Ork?

net.pdl2ork.PdL2Ork.yml
hfiguiere: I don't understand your question.

shared/glu-9.json
hfiguiere: then the PR can take another month to merge if that's what it takes.

net.pdl2ork.PdL2Ork.metainfo.xml
hfiguiere: So you didn't even read the questions you aknowledged:

> I own the domain used in the application ID or the domain has a policy for delegating subdomains (e.g. GitHub, SourceForge).

Nor the documentation.

Two wrongs don't make a right.

shared/glu-9.json
pd-l2ork: > > Given the complexity of the build process, I have tested on 22.08 for the past month and transitioning to 23.08 will likely require another month
> 
> What exactly is taking so long?
> 
> Note that, 23.08 is supported for longer (2 years) than 22.08 which is set to reach End of Life on August 2024.

This project has over 2,000 objects, many of which come from 3rd parties, and some that have been last touched over 20 years ago. I have a very frail and complex build environment that ensures these build ok. Since this is a programming language, it does everything from audio and networking, to openGL and beyond. And every error when it crops up, requires me to rebuild again, which can take up to 2 hours. I ran into a number of issues where ./configure scripts would fail because it thinks I am cross-compiling and need to redundantly specify --host parameter. Problems also changed depending on the version of SDK I used. At any rate, I can try to move things (and likely will). Just not sure if that is an urgent issue right now.

net.pdl2ork.PdL2Ork.metainfo.xml
pd-l2ork: I will upload updated version of this file shortly.

net.pdl2ork.PdL2Ork.yml
pd-l2ork: I am still trying to understand when you issue build/install commands to a module without a subfolder: option, what folder am I in? In the root of the cloned git folder where thing was checked out, or in the folder from which I ran the script? Some of the commands suggest former, and some latter. Yet all of them appear to work.

net.pdl2ork.PdL2Ork.yml
pd-l2ork: Thanks. Would this make the UI (e.g. mouse cursor, etc.) look different on Wayland desktops (e.g. Ubuntu) if we fell back to X11? For reference, other similar forks have this kind of setup, although they also did not have a release in a while.

net.pdl2ork.PdL2Ork.yml
pd-l2ork: Got it. Will remove it. Will this work perhaps at some point in the future, so that I leave it just commented out, as a reminder to revisit at a later date?

net.pdl2ork.PdL2Ork.yml
pd-l2ork: Got it. So, how do I specify just access to the home folder?

shared/glu-9.json
pd-l2ork: > then the PR can take another month to merge if that's what it takes.

So, are you saying that 22.08 is a deal-breaker for allowing this project to move forward?

hfiguiere: Each time we let this go it goes as follow:

- Package is accepted not on the latest runtime.
- Runtime is EOL (August 2024 in that case).
- Package is still using the runtime that is now EOL.
- Volunteers like me end up picking up the pieces.

net.pdl2ork.PdL2Ork.metainfo.xml
pd-l2ork: > So you didn't even read the questions you aknowledged:
> 
> > I own the domain used in the application ID or the domain has a policy for delegating subdomains (e.g. GitHub, SourceForge).
> 
> Nor the documentation.
> 
> Two wrongs don't make a right.

As a faculty at a university who owns this domain, I don't own the URL, but I do manage the server and am in charge of managing that specific subdomain. Should I register a separate, external domain and create a site (e.g. via github.io) just to be able to have a flatpak? If so, can I set it up as a redirect to our current site or github?

net.pdl2ork.PdL2Ork.yml
pd-l2ork: Should I also remove line 11?

hfiguiere: just change `x11` to `fallback-x11`.

net.pdl2ork.PdL2Ork.yml
hfiguiere: does you code call it directly? Then it should instead use the portal. If it doesn't call it, then there is no point.

net.pdl2ork.PdL2Ork.metainfo.xml
hfiguiere: you can simply use something associated to the github repository (`io.github.`) It's the low friction way.

shared/glu-9.json
bbhtt: > Would you please give me a clear step-by-step what you mean? 

Leave it, delete shared/glu-9.json, replace it with the one is shared-modules/glu-9.json in the manifest.

net.pdl2ork.PdL2Ork.yml
pd-l2ork: Thanks, unfortunately pd and it's forks doesn't work with portals yet, so access to the home dir a must at this point in time. I changed this to --filesystem=home.

net.pdl2ork.PdL2Ork.metainfo.xml
pd-l2ork: I think I've addressed all other issues and have pushed changes to the branch. I will look into creating a site github.io tomorrow. Then I only have to add one more file that I missed in the build process and we should be hopefully good to go, pending your approval. Thanks for your help.

pd-l2ork:
BTW, is there a way to break down a module from a single source into smaller individually cache-able parts? This would greatly expedite building. Right now, if I change anything in any one thing in the pd-externals module, it wants to rebuild the whole pd-externals, which is by far the largest one and the one that takes the longest to build. So, for instance, is there a way to cache these two calls, so that bsaylor is cached separately from fluid and without having to require each to pull its own copy of the same git snapshot (like it does when I separate these two into two distinct modules, that have to have source specified and then the build process looks to pull every submodule, resulting in a significantly longer build process)?

```
 # how to cache bsaylor separately from fluid below without separating them into two modules?
  - |
    export BSAYLOR_FLATPAK_LDFLAGS=-L/app/lib && \
         make prefix=/app \
         bsaylor
  - |
    make prefix=/app LDFLAGS=-L/app/lib \
         fluid
```

net.pdl2ork.PdL2Ork.metainfo.xml
pd-l2ork: So, I just created a site that is accessible at https://pd-l2ork.github.io/pd-l2ork/

Should therefore my ID be io.github.Pd-L2Ork, io.github.pdl2ork.Pd-L2Ork, io.github.pd-l2ork.Pd-L2Ork (last one has a dash in penultimate and last namespace, and the convention, unless I am misreading it, only allows dashes in the last one)? Thanks.

net.pdl2ork.PdL2Ork.yml
```diff
@@ -0,0 +1,197 @@
...
+  # K12 externals
+  - mkdir -p /app/lib/pd-l2ork/extra/K12
+  - cp -rf l2ork_addons/K12/* /app/lib/pd-l2ork/extra/K12/
+  # install nw
+  - mkdir -p /app/lib/pd-l2ork/bin/nw
+  - tar -x -f nwjs.tar.gz -C /app/lib/pd-l2ork/bin/nw --strip-components=1
+  # wrapper script
+  - install -Dm755 packages/linux_flatpak/pd-wrapper.sh /app/bin
+  # .desktop file (so we appear in list of applications)
+  - install -Dm644 packages/linux_flatpak/net.pdl2ork.PdL2Ork.desktop -t /app/share/applications/
```
bbhtt, line 173: You should fix these, it won't cause the build to fail but will generate warnings

```
$ desktop-file-validate net.pdl2ork.PdL2Ork.desktop

net.pdl2ork.PdL2Ork.desktop: hint: value "Audio;AudioVideo;Development;Midi;Music;3D" for key "Categories" in group "Desktop Entry" contains more than one main category; application might appear more than once in the application menu
net.pdl2ork.PdL2Ork.desktop: error: value "Audio;AudioVideo;Development;Midi;Music;3D" for key "Categories" in group "Desktop Entry" contains an unregistered value "3D"; values extending the format should start with "X-"
```

net.pdl2ork.PdL2Ork.yml
```diff
@@ -0,0 +1,197 @@
...
+add-extensions:
+  org.freedesktop.LinuxAudio.Plugins:
+    directory: extensions/Plugins
+    version: '23.08'
+    add-ld-path: lib
+    merge-dirs: ladspa
+    subdirectories: true
+    no-autodownload: true
+
+cleanup:
```
bbhtt, line 33: Add cleanup:


- /etc - libtirpc generates some config files here, are they needed?
- /lib/pkgconfig
- '*.a*
- '*.la'
- /share/info
- /share/man

net.pdl2ork.PdL2Ork.yml
```diff
@@ -0,0 +1,197 @@
...
+  - --socket=x11
+  #- --socket=wayland
+  - --device=dri
+# Audio
+  - --socket=pulseaudio
+  # currently disabled per a flathub maintainer's request
+  #- --system-talk-name=org.freedesktop.RealtimeKit1
+  - --filesystem=xdg-run/pipewire-0
+# Filesystem access (pd doesn't yet work with portals)
+  - --filesystem=home
```
bbhtt, line 19: Does this need to full home or you can restrict it to some specific folders like ~/Documents, Downloads etc.?

net.pdl2ork.PdL2Ork.yml
```diff
@@ -0,0 +1,197 @@
...
+
+  # original, local testing build uses simply
+  #make ...
+  # for flathub, consider replacing make with this commented out version
+  # this needs to be tested. if it proves to work, we will be able to build
+  # with multiple cores in parallel.
+  #make -j ${FLATPAK_BUILDER_N_JOBS} ...
+
+  - |
+    make prefix=/app \
```
bbhtt, line 90: Why did you remove parallel make here and below?

bbhtt:
> BTW, is there a way to break down a module from a single source into smaller individually cache-able parts?

You can make them separate locally to make it faster I guess. It won't rebuild the ones you haven't changed anything in the manifest but every module you changed and the ones after that will get rebuilt.

It doesn't matter much in the Flathub because it should be pretty fast. Fwiw, on my PC it takes around ~15 minutes for the whole thing.

net.pdl2ork.PdL2Ork.yml
bbhtt: only oscx seems to break on parallel

net.pdl2ork.PdL2Ork.yml
```diff
@@ -0,0 +1,197 @@
...
+
+cleanup:
+  - /include/*
+  - /lib/pd-l2ork/Makefile.am
+  - /lib/pd-l2ork/README.txt
+  - /lib/pd-l2ork/bin/todo.txt
+  - /lib/pd-l2ork/bin/nw/credits.html
+
+modules:
+- name: pd-l2ork
```
bbhtt, line 41: Seems like some libraries are missing?

image

net.pdl2ork.PdL2Ork.yml
bbhtt: Are you sure the it doesn't support portals? It seems to work with protals from what I can see?

net.pdl2ork.PdL2Ork.yml
```diff
@@ -0,0 +1,197 @@
+id: net.pdl2ork.PdL2Ork
+runtime: org.freedesktop.Platform
+runtime-version: '23.08'
+sdk: org.freedesktop.Sdk
+default-branch: stable
+command: pd-wrapper.sh
+finish-args:
```
bbhtt, line 7: It crashes immediately on opening Help> Show Devtools

It seems that it needs `--allow=devel` for ptrace but that doesn't seem to fix the crash.

net.pdl2ork.PdL2Ork.metainfo.xml
bbhtt: > io.github.pdl2ork.Pd-L2Ork

This is fine, you can replace the last dash with underscore https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic if you want

pd-l2ork: @bbhtt thank you. I will make this change shortly and upload the updated version asap.

hfiguiere: > I will look into creating a site github.io tomorrow.

this is not even necessary.

hfiguiere: > This is fine, you can replace the last dash with underscore

not can. should.

net.pdl2ork.PdL2Ork.yml
```diff
@@ -0,0 +1,197 @@
+id: net.pdl2ork.PdL2Ork
+runtime: org.freedesktop.Platform
+runtime-version: '23.08'
+sdk: org.freedesktop.Sdk
+default-branch: stable
```
hfiguiere, line 5:
```suggestion
```

net.pdl2ork.PdL2Ork.yml
```diff
@@ -0,0 +1,197 @@
...
+finish-args:
+# Display
+  - --share=ipc
+  - --socket=x11
+  #- --socket=wayland
+  - --device=dri
+# Audio
+  - --socket=pulseaudio
+  # currently disabled per a flathub maintainer's request
+  #- --system-talk-name=org.freedesktop.RealtimeKit1
```
hfiguiere, line 16:
```suggestion
```
Just remove it.

net.pdl2ork.PdL2Ork.metainfo.xml
pd-l2ork: Can I omit the underscore and just do io.github.pdl2ork.PdL2Ork?

net.pdl2ork.PdL2Ork.yml
pd-l2ork: I have added these to the cleanup part. I don't think I need any of them. Will update them with next commit.

net.pdl2ork.PdL2Ork.yml
pd-l2ork: I am not sure since this is my first time building a flatpak. I do know that that comment was copied from a manifest of a close fork, so I would tend to trust them over anything I can say. The patches can pull audio files, 3D objects, video files, text files, and bunch of other stuff, and often go beyond that, depending on the needs and imagination of its users, so if I understand the role of portals correctly, I would err on the side of full access. The users can also create folders and save their code/patches, as needed.

pd-l2ork: By default, access is to ~/.pd-l2ork/ ~/Pd-L2Ork (with its subfolders), and ~/pd-l2ork-externals. Users can also save their stuff wherever they like (and where they have permission to do so).

net.pdl2ork.PdL2Ork.yml
pd-l2ork: Those would be nw.js dev tools. Could be a combo of 23.08 and nw.js 0.67.1 which is the only one that consistently works across Linux, Mac, and Windows with minimal idiosyncrasies.

net.pdl2ork.PdL2Ork.yml
pd-l2ork: Will include it with the next push.

net.pdl2ork.PdL2Ork.yml
pd-l2ork: This is just a default build. This would work on a regular (e.g. deb deployment) if the user has NDI hardware, as in that case they would also have libraries installed system-wide. So, the error would be still there if they don't. I just build it for convenience. Here I could disable it but would rather wait and see if there is a request to bundle it (assuming that the library is FOSS-friendly).

net.pdl2ork.PdL2Ork.yml
pd-l2ork: On local builds it breaks anytime I use -j make option. The build process just randomly exits after 30-45 minutes of build time with a Code 9 (sometimes other code may show up). No other error reported.

net.pdl2ork.PdL2Ork.metainfo.xml
hfiguiere: no. it SHOULD match `io.github.pd_l2ork.*`

pd-l2ork: OK, thanks. Will retitle them...

net.pdl2ork.PdL2Ork.yml
pd-l2ork: I will try to rebuild wtih nwjs-sdk instead of the nwjs library to see if this fixes it.

pd-l2ork:
> > BTW, is there a way to break down a module from a single source into smaller individually cache-able parts?
> 
> You can make them separate locally to make it faster I guess. It won't rebuild the ones you haven't changed anything in the manifest but every module you changed and the ones after that will get rebuilt.
> 
> It doesn't matter much in the Flathub because it should be pretty fast. Fwiw, on my PC it takes around 15 minutes for the whole thing.

That's really nice. On my linux laptop, it's been going for 45 minutes now and still not done, maybe because I cannot use multithreaded build (see my Code 9 note in a previous comment).

net.pdl2ork.PdL2Ork.yml
pd-l2ork: So, it appears that using nwjs-sdk instead of nwjs has fixed things. I just pushed the update that should hopefully address all the questions you raised.

pd-l2ork:
OK, so this should address all of your concerns, plus any deficiencies in the build process I was able to identify so far. Please let me know if you have any additional requests. Thanks.

hfiguiere:
bot, build io.github.pd_l2ork.Pd_L2Ork

net.pdl2ork.PdL2Ork.yml
bbhtt: That's because oscx exists when using parallel, using parallel on the other ones makes it a lot faster.

net.pdl2ork.PdL2Ork.yml
bbhtt: No it seems to support portals, my guess would be it needs some kind of persistent access or it is just wrong.

It'd be better if it can be avoided or just restricted to something like 

```
--persist=.pd-l2ork
--filesystem=Pd-L2Ork
--filesystem=pd-l2ork-externals
```

Try running it with `flatpak run --nofilesystem=home io.github.pd_l2ork.Pd_L2Ork` and check the new and open/save dialogues.

hfiguiere: I think leaving it to `home` is good enough.

pd-l2ork:
I don't know if you want me to also check the test build, but I did so and it matches my local one. So, as far as I am concerned, I am good to go.

One last question, how often can I update the build script to match new releases? In other words, is there an accepted standard I should follow (e.g. once per month/quarter/etc.)? Thanks.

bbhtt:
> One last question, how often can I update the build script to match new releases? In other words, is there an accepted standard I should follow (e.g. once per month/quarter/etc.)? Thanks.

Whenever you need/there's a release/update

net.pdl2ork.PdL2Ork.yml
bbhtt: > Try running it with flatpak run --nofilesystem=home io.github.pd_l2ork.Pd_L2Ork and check the new and open/save dialogues.

@pd-l2ork Did you try this?

io.github.pd_l2ork.Pd_L2Ork.yml
```diff
@@ -0,0 +1,219 @@
+id: io.github.pd_l2ork.Pd_L2Ork
+runtime: org.freedesktop.Platform
+runtime-version: '23.08'
+sdk: org.freedesktop.Sdk
+default-branch: stable
```
hfiguiere, line 5:
```suggestion
```

pd-l2ork:
Would it be ok if I renamed my repo from where the pull request was made? Would that break the pull request and/or should I wait until the pull request was merged? Thank you.

hfiguiere:
Wait until this is merged, because then it will be irrelevant.

pd-l2ork:
> Wait until this is merged, because then it will be irrelevant.

Thank you. How long does it usually take for these to be merged?

hfiguiere:
as much as needed. This is all volunteer run.

pd-l2ork:
> as much as needed. This is all volunteer run.

I understand and respect the volunteer-based aspect--my effort is also laregly outside my work-related expectations and I do this because I find a great reward in seeing others being creative with the tools I've developed. I also do not mean to rush anyone. That said, given you have likely administered a large number of these pull requests, you may have noticed a trend on what is an average turnaround time for these. If so, I would appreciate a general idea (e.g. a day, days, weeks, months, etc.). Thanks for your consideration.

PR #4605, opened 2023-10-13
Add com.ranfdev.Notify

ranfdev:

### DIFF
--- a/com.ranfdev.Notify.json
+++ b/com.ranfdev.Notify.json
@@ -1,0 +1,83 @@
+{
+    "id": "com.ranfdev.Notify",
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "45",
+    "sdk": "org.gnome.Sdk",
+    "sdk-extensions": [
+        "org.freedesktop.Sdk.Extension.rust-stable",
+        "org.freedesktop.Sdk.Extension.llvm16"
+    ],
+    "command": "notify",
+    "finish-args": [
+        "--share=ipc",
+        "--share=network",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--device=dri",
+        "--env=RUST_LOG=notify=info,ntfy_daemon=info",
+        "--env=G_MESSAGES_DEBUG=none",
+        "--talk-name=org.freedesktop.Notifications"
+    ],
+    "build-options": {
+        "append-path": "/usr/lib/sdk/rust-stable/bin:/usr/lib/sdk/llvm16/bin",
+        "env": {
+            "CARGO_REGISTRIES_CRATES_IO_PROTOCOL": "sparse",
+            "CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER": "clang",
+            "CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUSTFLAGS": "-C link-arg=-fuse-ld=/usr/lib/sdk/rust-stable/bin/mold",
+            "CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER": "clang",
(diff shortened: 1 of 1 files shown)


com.ranfdev.Notify.json
```diff
@@ -0,0 +1,76 @@
...
+            "CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER": "clang",
+            "CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUSTFLAGS": "-C link-arg=-fuse-ld=/usr/lib/sdk/rust-stable/bin/mold",
+            "CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER": "clang",
+            "CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_RUSTFLAGS": "-C link-arg=-fuse-ld=/usr/lib/sdk/rust-stable/bin/mold"
+        }
+    },
+    "modules": [
+        {
+            "name": "capnp",
+            "buildsystem": "cmake",
```
hfiguiere, line 34: cmake-ninja?

hfiguiere:
bot, build com.ranfdev.Notify

com.ranfdev.Notify.json
```diff
@@ -0,0 +1,76 @@
...
+    "command": "notify",
+    "finish-args": [
+        "--share=ipc",
+        "--share=network",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--device=dri",
+        "--env=RUST_LOG=notify=info,ntfy_daemon=info",
+        "--env=G_MESSAGES_DEBUG=none",
+        "--talk-name=org.freedesktop.Notifications"
```
hfiguiere, line 19: is this necessary ? Gtk does the right thing.

ranfdev:
bot, build com.ranfdev.Notify

com.ranfdev.Notify.json
ranfdev: Why should I put cmake-ninja though? With cmake it compiles fine

hfiguiere: because it's faster and we'd prefer that. It's just `cmake -G Ninja`

hfiguiere: cmake is only used for case where the ninja backend fails.

bbhtt: Is this a build time dep only? Also might want to disable the test and turn shared-libs on.

com.ranfdev.Notify.json
```diff
@@ -0,0 +1,77 @@
...
+    ],
```
bbhtt, line 20: Cleanup:

- /include
- /lib/pkgconfig

com.ranfdev.Notify.json
ranfdev: Probably it's not necessary, but I'm already having a really hard time pushing notifications through the Notification portal, because of https://github.com/flatpak/xdg-dbus-proxy/issues/46. 

I prefer to keep "--talk-name=org.freedesktop.Notifications" just to have less possible things causing problems while sending a notification.

com.ranfdev.Notify.json
ranfdev: Yes, It's a build time dep only. I should've added the appropriate cleanups now.
I'm sorry but I can't find a way to skip the compilation of tests without patching the build process of capnproto.

ranfdev:
If the review is ok, could this app be published in the beta branch instead of master?

bbhtt:
Yes it can beta only

com.ranfdev.Notify.json
bbhtt: It is BUILD_TESTING I think https://github.com/capnproto/capnproto/blob/a91ec65323c2577732224d0cf6f5bf1e79d3a724/c%2B%2B/src/capnp/CMakeLists.txt#L227

bbhtt:
bot, build com.ranfdev.Notify

PR #4706, opened 2023-11-16
Add org.vinegarhq.Vinegar 

lunarlettuce:


lunarlettuce:
This is a migration from the previous name to the new flatpak. The previous one will be EOL'd once this flatpak is accepted.

lunarlettuce:
bot, build org.vinegarhq.VInegar

lunarlettuce:
bot, build org.vinegarhq.Vinegar

bbhtt (commented):
The desktop file, icons and appdata should be upstream, otherwise seems fine.

bbhtt:
bot, build org.vinegarhq.Vinegar

bbhtt:
Is this a cli application by default?

lunarlettuce:
@bbhtt Yes and no, I have the .desktop as a hack to get flathub to show the application on the website and so that it installs properly (hence why it is hidden). Vinegar itself is meant to run as a CLI program wrapping a GUI program, but also have two other auxiliary .desktops as shortcuts.

Additionally, we don't plan to use the metainfo upstream; it's just for compliance with flathub rules.

Noted-Jrelvas:
Vinegar is essentially a wrapper for two GUI programs. Whie it provides its functionality via CLI, in reality, the user isn't mean to be interfacing with it directly unless they're troubleshooting or actively tinkering. That's why it creates two desktop files. For the end user, Vinegar is meant to be invisible.

Noted-Jrelvas:
Your concerns are certainly true. That's why our main objective is to pick the best defaults possible, to ensure the user doesn't need to change the wrapper's configuration unless they're actually interested in tinkering with it and/or contributing.

For instance, Vinegar will automatically choose the dgpu for prime systems. The Wine binary we ship with our Flatpak has all of the best patches and modifications necessary to ensure the best out of the box functionality with the programs wrapped.

Our objective is to provide an experience as close to native as possible. Vinegar doesn't introduce a launcher UI or configuration UI because no equivalent exists in the native versions of the programs it wraps around.

Our Flatpak has a good track record of working out of the box, and is generally what we recommend to any users who complain about issues.

If we were to introduce a separate interface for configuring Vinegar, it'd realistically only expose a portion of actual options, as most of them are definitely not meant for the end-user. @lunarlettuce can comment more on that as a maintainer.

lunarlettuce:
Can confirm, the program is designed to have the best possible settings as a default. The user is not encouraged to change the config (unless Roblox breaks something upstream), and is intended to start from Roblox.com, mimicking the official launcher.

lunarlettuce:
@barthalion I have the Vinegar desktop file because Vinegar itself does not ship with a primary desktop file. Instead, it ships with 3 different .desktops (Roblox player, app, studio), all with different mime-type associations (due to how Roblox.com launching works). It would be inappropriate to designate one of these .desktops as the primary .desktop, so we use a hidden org.vinegarhq.Vinegar.desktop instead.

Noted-Jrelvas:
imagem

Well... I think the user's first instinct would be to use the two desktop entries Vinegar creates. You don't really need to mess around with the terminal.

lunarlettuce:
> If you happen to be able enough, you would drop to a terminal, and maybe figure out what to make of it:
> 
> ```
> ~ 6.7s ❱ flatpak run org.vinegarhq.Vinegar
> usage: vinegar [-config filepath] player|studio [args...]
> usage: vinegar [-config filepath] exec prog [args...]
>        vinegar [-config filepath] edit|kill|uninstall|delete|install-webview2|winetricks|sysinfo
> ```
> 
> But would you if you aren't a Linux user with some grasp of how things work under the hood? I know the answer.

The user would go to roblox.com to launch Player or Studio as they would on Windows. This isn't a design decision on our part.

PR #4717, opened 2023-11-20
Add me.jtheoof.swappy

zefr0x:

### DIFF
--- a/me.jtheoof.swappy.metainfo.xml
+++ b/me.jtheoof.swappy.metainfo.xml
@@ -30,6 +30,8 @@
         <mediatype>image/png</mediatype>
         <mediatype>image/jpeg</mediatype>
     </provides>
+    <!-- FIX: Deprecated tag. -->
+    <developer_name>Jeremy Attali</developer_name>
     <developer id="io.github.jtheoof">
         <name>Jeremy Attali</name>
         <url>https://github.com/jtheoof</url>
--- a/me.jtheoof.swappy.metainfo.xml
+++ b/me.jtheoof.swappy.metainfo.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<component type="desktop">
+<component type="console-application">
     <id>me.jtheoof.swappy</id>
     <metadata_license>CC0-1.0</metadata_license>
     <project_license>MIT</project_license>
--- a/me.jtheoof.swappy.metainfo.xml
+++ b/me.jtheoof.swappy.metainfo.xml
@@ -5,7 +5,6 @@
     <project_license>MIT</project_license>
     <name>swappy</name>
     <summary>Wayland native snapshot editing tool, inspired by Snappy on macOS</summary>
-    <icon type="remote" width="128" height="128">https://github.com/jtheoof/swappy/raw/master/res/icons/hicolor/scalable/apps/swappy.svg</icon>
     <description>
         <p>A Wayland native snapshot and image editing tool, inspired by Snappy on macOS. Works great with grim, slurp and sway. But can easily work with other screen copy tools that can output a final image to stdout.</p>
     </description>
@@ -37,7 +36,7 @@
     </developer>
     <screenshots>
         <screenshot type="default">
-            <image type="source" width="1347" height="672">https://raw.githubusercontent.com/jtheoof/swappy/master/docs/images/screenshot-1.0.0.png</image>
+            <image type="source" width="1347" height="672">https://raw.githubusercontent.com/jtheoof/swappy/596b9a8b17234303a9d4d2ea1d3107cb5442253e/docs/images/screenshot-1.0.0.png</image>
         </screenshot>
     </screenshots>
     <branding>
--- a/me.jtheoof.swappy.metainfo.xml
+++ b/me.jtheoof.swappy.metainfo.xml
@@ -1,0 +1,53 @@
+<?xml version="1.0" encoding="utf-8"?>
+<component type="desktop">
+    <id>me.jtheoof.swappy</id>
+    <metadata_license>CC0-1.0</metadata_license>
+    <project_license>MIT</project_license>
+    <name>swappy</name>
+    <summary>Wayland native snapshot editing tool, inspired by Snappy on macOS</summary>
+    <icon type="remote" width="128" height="128">https://github.com/jtheoof/swappy/raw/master/res/icons/hicolor/scalable/apps/swappy.svg</icon>
+    <description>
+        <p>A Wayland native snapshot and image editing tool, inspired by Snappy on macOS. Works great with grim, slurp and sway. But can easily work with other screen copy tools that can output a final image to stdout.</p>
+    </description>
+    <categories>
+        <category>Utility</category>
+        <category>Graphics</category>
+    </categories>
+    <keywords>
+        <keyword>wayland</keyword>
+        <keyword>snapshot</keyword>
+        <keyword>annotation</keyword>
+        <keyword>editing</keyword>
+    </keywords>
+    <content_rating type="oars-1.1" />
+    <url type="homepage">https://github.com/jtheoof/swappy</url>
+    <url type="help">https://github.com/jtheoof/swappy</url>
+    <url type="bugtracker">https://github.com/jtheoof/swappy/issues</url>
+    <url type="translate">https://github.com/jtheoof/swappy?tab=readme-ov-file#i18n</url>
+    <url type="vcs-browser">https://github.com/jtheoof/swappy</url>
+    <url type="contribute">https://github.com/jtheoof/swappy?tab=readme-ov-file#contributing</url>
+    <provides>
+        <binary>swappy</binary>
+        <mediatype>image/png</mediatype>
+        <mediatype>image/jpeg</mediatype>
+    </provides>
+    <developer id="io.github.jtheoof">
+        <name>Jeremy Attali</name>
+        <url>https://github.com/jtheoof</url>
+    </developer>
+    <screenshots>
+        <screenshot type="default">
+            <image type="source" width="1347" height="672">https://raw.githubusercontent.com/jtheoof/swappy/master/docs/images/screenshot-1.0.0.png</image>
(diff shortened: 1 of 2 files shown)


zefr0x:
bot, build me.jtheoof.swappy

zefr0x:
bot, build me.jtheoof.swappy

me.jtheoof.swappy.metainfo.xml
hfiguiere, line 1: this file is never installed.

also it should be upstreamed.

me.jtheoof.swappy.metainfo.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="utf-8"?>
+<component type="desktop">
+    <id>me.jtheoof.swappy</id>
+    <metadata_license>CC0-1.0</metadata_license>
+    <project_license>MIT</project_license>
+    <name>swappy</name>
+    <summary>Wayland native snapshot editing tool, inspired by Snappy on macOS</summary>
+    <icon type="remote" width="128" height="128">https://github.com/jtheoof/swappy/raw/master/res/icons/hicolor/scalable/apps/swappy.svg</icon>
```
hfiguiere, line 8: don't use remote icons.

me.jtheoof.swappy.yaml
```diff
@@ -0,0 +1,29 @@
...
+  - --device=dri
+  - --filesystem=home
+rename-desktop-file: swappy.desktop
+rename-icon: swappy
+modules:
+  - name: swappy
+    buildsystem: meson
+    config-opts:
+      - --buildtype=release
+      - --prefix=/app
```
hfiguiere, line 19:
```suggestion
```

me.jtheoof.swappy.yaml
```diff
@@ -0,0 +1,29 @@
+id: me.jtheoof.swappy
```
hfiguiere, line 1: `jtheoof.me` doesn't seem to have a website. it's hard to verify it's actually valid.

me.jtheoof.swappy.metainfo.xml
```diff
@@ -0,0 +1,53 @@
...
+        <mediatype>image/png</mediatype>
+        <mediatype>image/jpeg</mediatype>
+    </provides>
+    <developer id="io.github.jtheoof">
+        <name>Jeremy Attali</name>
+        <url>https://github.com/jtheoof</url>
+    </developer>
+    <screenshots>
+        <screenshot type="default">
+            <image type="source" width="1347" height="672">https://raw.githubusercontent.com/jtheoof/swappy/master/docs/images/screenshot-1.0.0.png</image>
```
hfiguiere, line 40: I can't wait to see this fail when the master branch changes and the 1.0.0 is removed.

This URL should point something more stable like a commit.

me.jtheoof.swappy.yaml
zefr0x: Its used in upstream: https://github.com/jtheoof/swappy/blob/596b9a8b17234303a9d4d2ea1d3107cb5442253e/src/application.c#L948

Should I use `io.github.jtheoof.swappy` instead?

hfiguiere: then you should ask them at the same time as you contact them.

me.jtheoof.swappy.metainfo.xml
zefr0x: I made a PR: https://github.com/jtheoof/swappy/pull/158

I hope the developer will response soon.

zefr0x:
bot, build me.jtheoof.swappy

me.jtheoof.swappy.metainfo.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="utf-8"?>
+<component type="desktop">
+    <id>me.jtheoof.swappy</id>
+    <metadata_license>CC0-1.0</metadata_license>
+    <project_license>MIT</project_license>
```
bbhtt, line 5: Add developer_name

me.jtheoof.swappy.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - --buildtype=release
+      - -Dman-pages=disabled
+    post-install:
+      - install -Dm644 ../$FLATPAK_ID.metainfo.xml $FLATPAK_DEST/share/metainfo/$FLATPAK_ID.metainfo.xml
+    sources:
+      - type: git
+        url: https://github.com/jtheoof/swappy
+        tag: v1.5.1
+        commit: a996b2cec65aadd04ded780f141b02c5bdfbeb32
+        disable-shallow-clone: true
```
bbhtt, line 27: Why?

me.jtheoof.swappy.yaml
```diff
@@ -0,0 +1,32 @@
...
+runtime: org.gnome.Platform
+runtime-version: "45"
+sdk: org.gnome.Sdk
+command: swappy
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --filesystem=home
```
bbhtt, line 11: Seems like the program prefers `$XDG_CONFIG_HOME/Desktop` first over `$HOME/Desktop` as the save location, `$XDG_CONFIG_HOME` is set inside the sandbox. Where does the screenshot end up without save_dir in config file?


Also, maybe restrict this to `$HOME/Desktop` or `$HOME/Pictures` by default. That's the usual and default save locations. If someone wants a custom location, they can add the permission.

me.jtheoof.swappy.yaml
bbhtt, line 27: Why?

me.jtheoof.swappy.metainfo.xml
zefr0x: Isn't it deprecated? I used the `<developer>` tag.

https://freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer_name

bbhtt: Flathub doesn't use libappstream. There's a check for missing developer name tag that's going to be turned on next year.

me.jtheoof.swappy.yaml
zefr0x: In the meson.build, `git` is used to create the `--version` string.

https://github.com/jtheoof/swappy/blob/596b9a8b17234303a9d4d2ea1d3107cb5442253e/meson.build#L17-L18

But after testing it, I can see that the commit hash is the same.

me.jtheoof.swappy.yaml
zefr0x: This is the code:
https://github.com/jtheoof/swappy/blob/596b9a8b17234303a9d4d2ea1d3107cb5442253e/src/config.c#L29-L32

Since `$XDG_DESKTOP_DIR` is not in the sandbox and `$XDG_CONFIG_HOME/Desktop` is I think a **bug** (it usually doesn't exist), it will save them in `$HOME/Desktop`. (tested behavior)

---

I didn't restrict it to `$HOME/Pictures` because it doesn't just save images, it also open images. It can be used by directly passing an image to stdin, or by specifying an image path (there is a .desktop file to open `png` and `jpen`).

zefr0x: Can I specify the `--filesystem=home` as `:ro` and `--filesystem=$HOME/Pictures` as `:rw`?

bbhtt: > it will save them in $HOME/Desktop. (tested behavior)

Yea, looks like, it saves to there. I didn't check there. xdg-desktop is a weird location to save the images, it should use xdg-pictures instead.

> Can I specify the --filesystem=home as :ro and --filesystem=$HOME/Pictures as :rw?

Yea, that seems fine: `--filesystem=xdg-desktop` and `--filesystem=home:ro`. $HOME/Pictures is entirely unused.

bbhtt:
bot, build me.jtheoof.swappy

bbhtt:
Please test the above build

bbhtt:
Looks like copying the image does not work. It depends on wl-clipboard https://github.com/jtheoof/swappy/blob/596b9a8b17234303a9d4d2ea1d3107cb5442253e/src/clipboard.c#L35 can you add these in the manifest and check if it works?

zefr0x:
It works without it actually, but when you close the app it will be lost from the clipboard. (The normal behavior without a clipboard manager)

With `wl-clipboard` it will be available even after closing the application.

This should only be possible in wlroots based Wayland compositors and KDE:
https://wayland.app/protocols/wlr-data-control-unstable-v1
Basically it's a hole in Wayland's security.

I think with a clipboard manager it will not make any different but I need to test it in a real desktop environment since I don't use any clipboard manager.

hfiguiere:
> https://wayland.app/protocols/wlr-data-control-unstable-v1

This is not availble with gnome-shell.

bbhtt:
If it expects to use wl-copy/paste to persistently copy-paste, I think it'd be better to give it that.

> It works without it actually, but when you close the app

Yes.

Another thing, the desktop file has NoDisplay=true, but this is a type=desktop-application in metainfo, not a console one. This would be confusing to people installing the application for the first time.

zefr0x:
> Another thing, the desktop file has NoDisplay=true, but this is a type=desktop-application in metainfo, not a console one. This would be confusing to people installing the application for the first time.

Should I change it to `console-application` in metainfo file?

It's a weird application, we can't open it without passing a file to it (so we mustn't show it in applications menu), but it has a GUI. It should be visible from context menu in file managers when opening an image.

zefr0x:
It seems to relay on FontAwesome v5: https://github.com/jtheoof/swappy/blob/596b9a8b17234303a9d4d2ea1d3107cb5442253e/res/style/swappy.css#L2

bbhtt:
> It's a weird application, we can't open it without passing a file to it (so we should not show it in applications menu), but it has a GUI. It should be visible from context menu in file managers when opening an image.

I was hoping there would be some way to spawn an empty GUI..

Also, the program doesn't have a close button.

PR #4722, opened 2023-11-22
add org.mfek.MFEKglif

Subject38:


org.mfek.glif.yml
```diff
@@ -0,0 +1,44 @@
...
+      - install -Dm755 ./target/release/MFEKglif -t /app/bin/
+      - install -Dm644 ./${FLATPAK_ID}.metainfo.xml -t /app/share/metainfo/
+      - install -Dm644 ./${FLATPAK_ID}.desktop -t /app/share/applications/
+      - install -Dm644 ./${FLATPAK_ID}.png -t /app/share/icons/hicolor/512x512/apps/
+    sources:
+      - type: git
+        url: https://github.com/MFEK/glif.git
+        commit: 8764b3e40a6dc219f3dc7db1429c58dff9795840
+      - type: file
+        url: https://github.com/rust-skia/skia-binaries/releases/download/0.62.0/skia-binaries-8cf4841aefdb295709e9-x86_64-unknown-linux-gnu-gl.tar.gz
```
hfiguiere, line 35: why x86_64 only?

org.mfek.glif.png
hfiguiere, line 1: this file exist upstream. no reason to have it here.

org.mfek.glif.metainfo.xml
hfiguiere, line 1: this should be submitted upstream. Also missing required `url`, `content_rating`, `releases`

org.mfek.glif.metainfo.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>org.mfek.glif</id>
+  
+  <name>MFEKglif</name>
+  <summary>Glyph editor for the Modular Font Editor K project.</summary>
```
hfiguiere, line 6:
```suggestion
  <summary>Glyph editor for the Modular Font Editor K project</summary>
```

org.mfek.glif.yml
```diff
@@ -0,0 +1,44 @@
+app-id: org.mfek.glif
+runtime: org.freedesktop.Platform
+runtime-version: '22.08'
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-nightly
```
hfiguiere, line 6: please, don't use rust-nightly

org.mfek.glif.yml
Subject38: Compiling for aarch64 on linux is untested. As far as I know, the only thing that would need to be resolved is skia-binaries but I'm unsure. Is there a way to condition the file used on the architecture? This is my first time working on a flatpak package so I'm very new to the process

org.mfek.glif.desktop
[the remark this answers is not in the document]
Subject38: Technically the file is slightly different since the icon image filename in this version is org.mfek.glif.png. When I was testing locally, the desktop file didn't work unless I used that name scheme. I imagine I could just use the upstream one and do a 1 line patch though?

org.mfek.glif.png
Subject38: Will fix

org.mfek.glif.desktop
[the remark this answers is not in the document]
hfiguiere: use `desktop-file-edit`

org.mfek.glif.yml
Subject38: Due to the features used by the program, rust-nightly is required. If it's possible to do so, we could freeze the rust-nightly version to today's (I don't know how), but I'm unsure of the amount of work required to use the stable branch instead... Paging @ctrlcctrlv

org.mfek.glif.metainfo.xml
```diff
@@ -0,0 +1,29 @@
...
+  <launchable type="desktop-id">org.mfek.glif.desktop</launchable>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://raw.githubusercontent.com/MFEK/glif/master/doc/screenshot_main.png</image>
+    </screenshot>
+    <screenshot>
+      <image>https://raw.githubusercontent.com/MFEK/glif/master/doc/screenshot.png</image>
+    </screenshot>
+    <screenshot>
+      <image>https://raw.githubusercontent.com/MFEK/glif/master/doc/screenshot2.png</image>
```
bbhtt, line 26: These screenshots should be from a tag or a commit not master branch.

org.mfek.glif.metainfo.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>org.mfek.glif</id>
+  
+  <name>MFEKglif</name>
+  <summary>Glyph editor for the Modular Font Editor K project.</summary>
+  
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>Apache-2.0</project_license>
+
```
bbhtt, line 10: Add developer_name

org.mfek.glif.yml
```diff
@@ -0,0 +1,44 @@
+app-id: org.mfek.glif
+runtime: org.freedesktop.Platform
+runtime-version: '22.08'
```
bbhtt, line 3:
```suggestion
runtime-version: '23.08'
```
23.08 is the latest runtime

org.mfek.glif.yml
```diff
@@ -0,0 +1,44 @@
...
+  env: 
+    CARGO_HOME: /run/build/MFEKglif/cargo
+    SKIA_BINARIES_URL: "file:///run/build/MFEKglif/skia-binaries.tar.gz"
+    FORCE_SKIA_BINARIES_DOWNLOAD: "1"
+finish-args: 
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --device=dri
+  - --filesystem=home
```
bbhtt, line 18: Why do you need this? Does it support filechoosers?

org.mfek.glif.yml
ctrlcctrlv: @hfiguiere Is this a Flathub policy, and what's the reason for it?

_cf._ [MFEK/glif@`53f4523`:/src/main.rs <sup>¶</sup> line № 6](https://github.com/MFEK/glif/blob/53f452363e97d402a5dbc9a02ce0d3177909945e/src/main.rs#L6) _et seq._:

```rust
#![feature(
    type_alias_impl_trait,
    local_key_cell_methods,
    trait_alias,
    const_trait_impl,
    generic_const_exprs,
    adt_const_params
)]
```

These are all in extensive use, especially `generic_const_exprs` (rust-lang/rust#76560) but also `adt_const_params` (rust-lang/rust#95174), neither of which are on track for _imminent_ stabilization (especially not the former, open since 2020).

If reproducibility is the concern as I suspect, can't @Subject38 just pick a date and use that nightly?

org.mfek.glif.yml
ctrlcctrlv: Yes, MFEKglif supports file choosers on all three platforms. On GNU/Linux, GTK3's one is used.

bbhtt: > Yes, MFEKglif supports file choosers on all three platforms. On GNU/Linux, GTK3's one is used.

Do they use the portal https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.FileChooser.html ? If so, you don't need this.

org.mfek.glif.yml
ctrlcctrlv: org.freedesktop.Sdk.Extension.rust-nightly.YYYY.MM.DD ?

hfiguiere: NO

Subject38: By bundle, do you mean we should be downloading that specific version using rustup in the build script?

ctrlcctrlv: yes

ctrlcctrlv: > NO

o_12

org.mfek.glif.yml
Subject38: The portal is not used. So having filesystem access of some sort is needed to be able to use the native file choosers. Given the fact that programs like Blender use host filesystem access, I don't see why it's a problem to use this permission

hfiguiere: > Given the fact that programs like Blender use host filesystem

1. it's hard to revert without breaking
2. two wrongs don't make a right
3. this is the problem with exceptions: everybody use them as the reason to ask for one.

ctrlcctrlv: If the UFO's guidelines change, MFEKglif changes the guidelines visible via Linux inotify, so the glif opened is never the only  file accessed. Layer metadata also accessed.

ctrlcctrlv: (&, Ctrl-O reopens “Open” dialog, it's also under “File”.)

Screenshot_20231128_230508

How is this “wrong” in any application?

bbhtt: > The portal is not used. 

You should switch to something that supports the portal. 

Right now seems like nfd-rs https://github.com/MFEK/glif/blob/53f452363e97d402a5dbc9a02ce0d3177909945e/src/filedialog.rs is used which provides bindings for https://github.com/mlabbe/nativefiledialog, and has stopped development/doesn't seem to support the portal.

ctrlcctrlv: > You should switch to something that supports the portal.

Why? NFD works fine, and if I do this, then I am giving up the ability to customize the file dialog on all platforms, because as things stand I can replace nfd.rlib with a future e.g mfek-filedialog.rlib which would provide, for example, UFO-aware file-dialogs which don't show them as directories, like FontForge does.

ctrlcctrlv: Am I right in thinking that “the rules” only apply to small projects?

hfiguiere:
bot, build org.mfek.glif

org.mfek.glif.yml
bbhtt: I can't force you to, hence "should". Portals are better than using static permissions. You have to decide the merits/demerits of it in the context of your app.

bbhtt: Eg. Apart from getting access to any files/folders without a static filesystem permission, you do get DE-native filechoosers.

Nativefiledialogue seems to support only GTK and an optional zenity backend. I guess it'd look weird on KDE/Qt.

org.mfek.glif.yml
Subject38: I distinctly remember there being an aarch64-gnu distribution. Looking again, I was wrong, so aarch64 can't be supported.
https://github.com/rust-skia/skia-binaries/releases/tag/0.62.0

org.mfek.glif.yml
```diff
@@ -0,0 +1,75 @@
...
+          - x86_64
+        url: https://github.com/rust-skia/skia-binaries/releases/download/0.62.0/skia-binaries-8cf4841aefdb295709e9-x86_64-unknown-linux-gnu-gl.tar.gz
+        dest-filename: skia-binaries.tar.gz
+        sha256: e4806ae8566b1cc1e7692d9c5f0e0a32754a07591db6de884c107adea1922b7a
+      - type: file
+        only-arches:
+          - aarch64
+        url: https://github.com/rust-skia/skia-binaries/releases/download/0.62.0/skia-binaries-8cf4841aefdb295709e9-aarch64-unknown-linux-gnu-gl.tar.gz
+        dest-filename: skia-binaries.tar.gz
+        sha256: e4806ae8566b1cc1e7692d9c5f0e0a32754a07591db6de884c107adea1922b7a
```
hfiguiere, line 50: the same sha256 for both arch? chances are this isn't right.

org.mfek.glif.yml
hfiguiere: why not building from source?

org.mfek.glif.yml
Subject38: Yes this was a mistake. I hastily copy pasted and then didn't realise until seeing the build fail that the file didn't even exist. There is no distribution for aarch64 gnu linux. The reason I was using this file in the first place was that building from source for this specific library proved infeasible offline

org.mfek.glif.yml
Subject38: I can look into building from source again, but I lost several hours trying to get it working offline. Likely a task for tomorrow though...

org.mfek.glif.yml
ctrlcctrlv: Yes, that wouldn't be right. I'm trying a real aarch64 build shortly as I have an aarch64 Arch VM.

First I am applying patches to make writing .desktop and icon unneeded, MFEKglif can now do it at runtime.

Screenshot_20231129_013700

ctrlcctrlv: MFEKglif opens /dev/shm/MFEKglif, copies into it the icon and .desktop.
It informs ipc.rlib of this, which creates symlinks from $XDG_SHARE_DIR/{applications,icons} to /dev/shm.
This way, MFEKglif can hopefully soon support showing the current glyph being edited as the icon, like FontForge does.
Invariably, this breaks some best practice, like the rest of the project. :o)

Subject38:
I decided to build the binary from source on my m1 macbook running linux. I have build succeeding fully locally on that machine. I had to use an older version of rust nightly due to a feature related to SIMD but I can now happily say that aarch64 should successfully build

org.mfek.MFEKglif.yml
```diff
@@ -0,0 +1,75 @@
...
+      - type: file
+        only-arches:
+          - x86_64
+        url: https://github.com/rust-skia/skia-binaries/releases/download/0.62.0/skia-binaries-8cf4841aefdb295709e9-x86_64-unknown-linux-gnu-gl.tar.gz
+        dest-filename: skia-binaries.tar.gz
+        sha256: e4806ae8566b1cc1e7692d9c5f0e0a32754a07591db6de884c107adea1922b7a
+      - type: file
+        only-arches:
+          - aarch64
+        path: skia-binaries-0.62.0-aarch64-unknown-linux-gnu-gl.tar.gz
```
bbhtt, line 48: This shouldn't be here. If they don't supply aarch64 builds, and you can't build from source, then build for x86_64 only.

Shipping packages built on your PC isn't a good idea. The trusted source of this is the rust-skia organisation.

Subject38: I spent some time trying to chase down building this lib from source offline but ultimately the process is too confusing for me to figure out. There is an open pull request https://github.com/rust-skia/rust-skia/pull/757 that will add these binaries to the distributed ones, and once it is merged, we can either hopefully get them to release binaries for the version we're using, or look into updating the library to the latest. Are there any legitimate blockers that prevent this from being merged as is? The things I saw discussed seemed to be more suggestions than anything. And anything else that wasn't I addressed.

hfiguiere:
bot, build org.mfek.glif

hfiguiere:
bot, build org.mfek.MFEKglif

barthalion:
bot, build org.mfek.MFEKglif

PR #4725, opened 2023-11-24
Add page.codeberg.JakobDev.jdProcessFileWatcher

JakobDev:

The file permissions are needed for reading all desktop files. `org.freedesktop.Flatpak` is needed to run strace on the Host.

### DIFF
--- a/page.codeberg.JakobDev.jdProcessFileWatcher.yaml
+++ b/page.codeberg.JakobDev.jdProcessFileWatcher.yaml
@@ -1,0 +1,72 @@
+id: page.codeberg.JakobDev.jdProcessFileWatcher
+runtime: org.kde.Platform
+runtime-version: '6.6'
+sdk: org.kde.Sdk
+base: com.riverbankcomputing.PyQt.BaseApp
+base-version: '6.6'
+command: jdprocessfilewatcher
+cleanup-commands:
+  - /app/cleanup-BaseApp.sh
+build-options:
+  env:
+    - BASEAPP_REMOVE_PYWEBENGINE=1
+finish-args:
+  - --filesystem=xdg-data/flatpak/exports/share/applications:ro
+  - --filesystem=/var/lib/flatpak/exports/share/applications:ro
+  - --filesystem=xdg-data/flatpak/exports/share/icons:ro
+  - --filesystem=xdg-data/flatpak/exports/share/icons:ro
+  - --filesystem=xdg-data/flatpak/runtime:ro
+  - --filesystem=/var/lib/flatpak/runtime:ro
+  - --filesystem=/var/lib/flatpak/runtime:ro
+  - --filesystem=xdg-data/applications:ro
+  - --filesystem=xdg-data/flatpak/app:ro
+  - --talk-name=org.freedesktop.Flatpak
+  - --filesystem=/var/lib/flatpak/app:ro
+  - --filesystem=host-etc:ro
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --share=ipc
+
+modules:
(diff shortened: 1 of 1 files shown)


page.codeberg.JakobDev.jdProcessFileWatcher.yaml
```diff
@@ -0,0 +1,72 @@
...
+  - --talk-name=org.freedesktop.Flatpak
+  - --filesystem=/var/lib/flatpak/app:ro
+  - --filesystem=host-etc:ro
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --share=ipc
+
+modules:
+  - name: strace
```
hfiguiere, line 32: strace can't run in the sandbox unless `--devel`

page.codeberg.JakobDev.jdProcessFileWatcher.yaml
```diff
@@ -0,0 +1,72 @@
...
+base-version: '6.6'
+command: jdprocessfilewatcher
+cleanup-commands:
+  - /app/cleanup-BaseApp.sh
+build-options:
+  env:
+    - BASEAPP_REMOVE_PYWEBENGINE=1
+finish-args:
+  - --filesystem=xdg-data/flatpak/exports/share/applications:ro
+  - --filesystem=/var/lib/flatpak/exports/share/applications:ro
```
hfiguiere, line 15: I wonder how that work if the path is non standard.

page.codeberg.JakobDev.jdProcessFileWatcher.yaml
JakobDev: strace isn't run in the sandbox. It is run on the Host. `--devel` don't help here.

page.codeberg.JakobDev.jdProcessFileWatcher.yaml
JakobDev: No, but there's nothing I can do about it. Luckily this can only be [done by editing a config file](https://docs.flatpak.org/en/latest/tips-and-tricks.html#adding-a-custom-installation), so majority will use the standard,

page.codeberg.JakobDev.jdProcessFileWatcher.yaml
hfiguiere: then you shouldn't be building it.

JakobDev: Why?

hfiguiere: how do you guarantee that this will run on the host and how do you run it on the host.

JakobDev: It is run using `ld-linux.so.2` with the `--libary-path` argument, so it will only use libs from the Runtime and the App. You can find the full logic [here](https://codeberg.org/JakobDev/jdProcessFileWatcher/src/commit/bad7b1b0ebb22ac926a6f3668d83764591556f37/jdProcessFileWatcher/StraceWrapper.py#L128). I have tested it on an older Ubuntu and it works.

JakobDev:
You need to merge the PR for flatpak-builder-lint first

JakobDev:
@hfiguiere Why is this blocked? It just needs the exception PR merged.

hfiguiere:
> @hfiguiere Why is this blocked? It just needs the exception PR merged.

because it hasn't been merged. so it's blocked. note that the label are just labels, and don't do anything else.

JakobDev:
bot, build page.codeberg.JakobDev.jdProcessFileWatcher

PR #4734, opened 2023-11-27
Add org.openscopeproject.TrguiNG

qu1ck:

The patches are:
1. work around hardcoded paths in tauri framework not compatible with flatpaks living in `/app` 
2. work around flatpak build system putting everything in the repo worktree before building, thus making repo appear dirty for version detection system


qu1ck:
Note, this submission is heavily based on another similar app build on top of tauri https://github.com/flathub/in.cinny.Cinny

The app is fully functional in my local testing but I did encounter issue with app icons not showing correctly in the taskbar and in window decorations. Other flatpak apps also have similar issues on kde plasma so I'm assuming it's not related to the way I built the flatpak bundle but if anyone suggests a way to correct this, I'd be grateful.

The app icon does show correctly in the app menu (i.e. .desktop file finds it fine).

Screenshots are taken from the flatpak version of the app.

org.openscopeproject.TrguiNG.yml
```diff
@@ -0,0 +1,121 @@
+id: org.openscopeproject.TrguiNG
+runtime: org.gnome.Platform
+runtime-version: '44'
```
hfiguiere, line 3: current runtime is 45.

org.openscopeproject.TrguiNG.desktop
hfiguiere, line 1: why isn't this upstream? A destkop app for Linux without a .desktop, it's a bug.

org.openscopeproject.TrguiNG.yml
```diff
@@ -0,0 +1,121 @@
...
+
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --share=ipc
+  - --share=network
+  - --filesystem=home
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.kde.StatusNotifierWatcher
```
hfiguiere, line 21: do you really need these dbus names?

org.openscopeproject.TrguiNG.yml
```diff
@@ -0,0 +1,121 @@
...
+  - org.freedesktop.Sdk.Extension.node18
+  - org.freedesktop.Sdk.Extension.rust-stable
+
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --share=ipc
+  - --share=network
+  - --filesystem=home
```
hfiguiere, line 19: home is a broad filesystem permission

org.openscopeproject.TrguiNG.desktop
qu1ck: There is a template for it that tauri fills in during build, resulting file is not checked in
https://github.com/openscopeproject/TrguiNG/blob/master/src-tauri/app.desktop

org.openscopeproject.TrguiNG.yml
qu1ck: The app has desktop notifications functionality.

hfiguiere: that doesn't answer the question.

these days toolkit properly use the portal.

so are they really needed?

org.openscopeproject.TrguiNG.desktop
hfiguiere: it should be upstream, really.

org.openscopeproject.TrguiNG.yml
qu1ck: The app needs to be able to read and write torrent files. Most commonly from the downloads location but user should be able to save their created torrent file anywhere.

org.openscopeproject.TrguiNG.yml
qu1ck: Let me test without them to double check.

org.openscopeproject.TrguiNG.yml
bbhtt: > The app needs to be able to read and write torrent files. Most commonly from the downloads location but user should be able to save their created torrent file anywhere.

Does it support filechoosers? Then you can open torrents from anywhere but save it to Downloads only.

org.openscopeproject.TrguiNG.yml
```diff
@@ -0,0 +1,121 @@
...
+        make-install-args:
+          - prefix=${FLATPAK_DEST}
+        sources:
+          - type: archive
+            url: https://dotat.at/prog/unifdef/unifdef-2.12.tar.xz
+            sha256: 43ce0f02ecdcdc723b2475575563ddb192e988c886d368260bc0a63aee3ac400
+        cleanup:
+          - '*'
+
+      - name: libjxl
```
bbhtt, line 64: If you bump it to 45, remove this. It's in the runtime.

resources/Screenshot1.png
bbhtt, line 1: Why are screenshots here? They should be part of your project repo or website.

org.openscopeproject.TrguiNG.appdata.xml
bbhtt, line 1: This file should be upstream too.

org.openscopeproject.TrguiNG.appdata.xml
```diff
@@ -0,0 +1,59 @@
...
+    </p>
+    <p>
+      Requires transmission server version 2.40 or later.
+    </p>
+  </description>
+
+  <launchable type="desktop-id">org.openscopeproject.TrguiNG.desktop</launchable>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://raw.githubusercontent.com/flathub/org.openscopeproject.TrguiNG/master/resources/Screenshot1.png</image>
```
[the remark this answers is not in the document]
qu1ck: Updated, these are temporary, I'll move screenshots to website for real submission.

org.openscopeproject.TrguiNG.yml
qu1ck: Testing shows that org.freedesktop.Notifications is needed, otherwise desktop notifications break.
But org.kde.StatusNotifierWatcher is less clear cut:
* Without it tray icon works on x11 and not on wayland
* With it tray icon is semi-broken on KDE on both x11 and wayland: it exists in the tray but has no icon, but works fine on Cinnamon (x11).

Since it's less broken without this permission I have removed it.

org.openscopeproject.TrguiNG.yml
qu1ck: It does use filechoosers but if program is invoked from file manager by clicking torrent files it will not be invoked. Does path remapping still happen in such cases?

org.openscopeproject.TrguiNG.yml
qu1ck: Removed.

org.openscopeproject.TrguiNG.yml
qu1ck: Tested this some more, opening files from file managers also works fine.

But problem is when trying to choose save location for saved torrent, if it's outside of permitted filesystem it turns into path that transmission will not understand (which is not running inside of flatpak). So this adds significant hurdle to trying to use one of the basic features of the app.

resources/Screenshot1.png
qu1ck: These are temporary, I will upload them to website for real submission.

>Also in the screenshots the window shows a default Wayland icon.

That seems to be KDEPlasma/Wayland related, at least on my bone stock debian 12. Other flatpak and non-flatpak apps have similar issues.
Once I switched to x11 session the problem went away.

qu1ck:
bot, build org.openscopeproject.TrguiNG

qu1ck:
bot, build org.openscopeproject.TrguiNG

org.openscopeproject.TrguiNG.yml
bbhtt: Can you add a default save location to `~/Downloads` and check if the location is writeable before saving? 

If someone needs extra locations they can add the permission. "home" isn't a catch-all, someone may want to save to an external drive. You can't cover every location.

org.openscopeproject.TrguiNG.appdata.xml
qu1ck: Where else is it used?

hfiguiere: any distro will

GNOME Software and KDE Discover use it.

qu1ck: Thanks, I learned a bit more about these files. 
Since I have 0 intention to go through the hoops of submitting a package to a major distro this still remains flatpak specific for this project.
If this changes at some point I can always move the file to project repo later.

PR #4741, opened 2023-11-28
Add com.microsoft.AzureStorageExplorer

mkfrey:

- [x] I am using only the minimal set of permissions. *filesystem=home for file upload from home directory*
- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link: https://github.com/microsoft/AzureStorageExplorer/issues/3606**


hfiguiere:
how does this work on aarch64?

hfiguiere:
bot, build com.microsoft.AzureStorageExplorer

mkfrey:
> how does this work on aarch64?

It doesn't. There are no binaries provided by microsoft.

hfiguiere:
> > how does this work on aarch64?
> 
> It doesn't. There are no binaries provided by microsoft.

an aarch64 is still built...

mkfrey:
> > > how does this work on aarch64?
> > 
> > 
> > It doesn't. There are no binaries provided by microsoft.
> 
> an aarch64 is still built...

How do I prevent that? extra-data already has the only-arches-Parameter set to `x86_64`.

hfiguiere:
that mean the extra-data is ignored on other arch. nothing else.

https://docs.flathub.org/docs/for-app-authors/maintenance#limiting-the-set-of-architectures-to-build-on

com.microsoft.AzureStorageExplorer.desktop
```diff
@@ -0,0 +1,11 @@
+[Desktop Entry]
+Type=Application
+Name=Azure Storage Explorer
+GenericName=Azure storage management
+Comment=Manage Azure cloud storage resources from your desktop
+Icon=com.microsoft.AzureStorageExplorer
+Exec=
```
bbhtt, line 7: Why missing?

com.microsoft.AzureStorageExplorer.yaml
```diff
@@ -0,0 +1,82 @@
...
+separate-locales: false
+tags:
+  - proprietary
+build-options:
+  append-path: "/usr/lib/sdk/dotnet6/bin"
+  append-ld-library-path: "/usr/lib/sdk/dotnet6/lib"
+  env:
+    PKG_CONFIG_PATH: "/app/lib/pkgconfig:/app/share/pkgconfig:/usr/lib/pkgconfig:/usr/share/pkgconfig:/usr/lib/sdk/dotnet6/lib/pkgconfig"
+finish-args:
+  - "--filesystem=home"
```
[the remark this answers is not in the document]
mkfrey: I've removed the permission. I tested with `--nofilesystem=home`, but this seems to be different from only not having the permission.

The app uses portals.

com.microsoft.AzureStorageExplorer.desktop
mkfrey: According to the docs (https://docs.flatpak.org/en/latest/freedesktop-quick-reference.html) the Exec can be commited, in which case the Exec in the generated .desktop file will just not contain the --command flag.

com.microsoft.AzureStorageExplorer.yaml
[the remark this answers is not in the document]
bbhtt: persist has no meaning with home access. Forgot to mention this.

com.microsoft.AzureStorageExplorer.desktop
bbhtt: I know it can be empty, just not sure if it's a good idea

com.microsoft.AzureStorageExplorer.yaml
```diff
@@ -0,0 +1,84 @@
...
+app-id: com.microsoft.AzureStorageExplorer
+runtime: org.freedesktop.Platform
+runtime-version: "23.08"
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.dotnet6
+command: StorageExplorer
+separate-locales: false
+only-arches:
+  - x86_64
```
bbhtt, line 11: This has no meaning here

```suggestion
```

com.microsoft.AzureStorageExplorer.yaml
```diff
@@ -0,0 +1,84 @@
...
+failure: test
```
bbhtt, line 12:
```suggestion
```

This does not exist

com.microsoft.AzureStorageExplorer.desktop
bbhtt: I think it's best to add it even if flatpak and the command makes it redundant.

com.microsoft.AzureStorageExplorer.yaml
```diff
@@ -0,0 +1,84 @@
...
+  - "/include"
+  - "/lib/pkgconfig"
+  - "/share/pkgconfig"
+  - "/share/aclocal"
+  - "/man"
+  - "/share/man"
+  - "/share/gtk-doc"
+  - "/share/doc"
+  - "*.la"
+  - "*.a"
```
bbhtt, line 43:
```suggestion
```

libsecret has it's own cleanup, you aren't building anything that needs to be cleaned up.

com.microsoft.AzureStorageExplorer.desktop
mkfrey: Ok, I added it.

flathub.json
```diff
@@ -0,0 +1,4 @@
+{
+  "only-arches": ["x86_64"],
+  "automerge-flathubbot-prs": true
```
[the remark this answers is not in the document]
mkfrey: I've reverted the change. I just assumed that Microsoft would test their releases and I could save some work automatically approving all PRs from external data checker.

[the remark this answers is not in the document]
bbhtt: It could break due to changes in the freedesktop sdk/platform.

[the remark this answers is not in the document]
mkfrey: Fair point.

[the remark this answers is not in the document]
hfiguiere: they don't maintain this, so they don't test it.

bbhtt:
Seems to run, please test to make sure it's ok.

hfiguiere:
bot, build com.microsoft.AzureStorageExplorer

barthalion:
bot, build com.microsoft.AzureStorageExplorer

PR #4775, opened 2023-12-12
Add io.neovim.nvim.tool.ripgrep

franute:

[Link to the discussion](https://github.com/BurntSushi/ripgrep/discussions/2685)
The only patch I've added to the repo is to enable building the application with pre-downloaded dependencies and not by letting cargo download them automatically as part of the build process. That is flatpak specific really. I'd like to create an automation in the future though because to achieve this, one has to use [flatpak-cargo-generator](https://github.com/flatpak/flatpak-builder-tools/tree/master/cargo) against `Cargo.lock` from the original repo, but I thought I could play with that later and for now update it manually when a new version of *ripgrep* is released.

### DIFF
--- a/io.neovim.nvim.tool.ripgrep.metainfo.xml
+++ b/io.neovim.nvim.tool.ripgrep.metainfo.xml
@@ -2,7 +2,7 @@
 <component type="addon">
   <id>io.neovim.nvim.tool.ripgrep</id>
   <metadata_license>CC0-1.0</metadata_license>
-  <project_license>MIT</project_license>
+  <project_license>MIT and Unlicense</project_license>
   <name>RipGrep</name>
   <summary>RipGrep search tool to be used by Neovim's plugins</summary>
   <url type="homepage">https://github.com/BurntSushi/ripgrep</url>
--- a/io.neovim.nvim.tool.ripgrep.metainfo.xml
+++ b/io.neovim.nvim.tool.ripgrep.metainfo.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="addon">
+  <id>io.neovim.nvim.tool.ripgrep</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>MIT</project_license>
+  <name>RipGrep</name>
+  <summary>RipGrep search tool to be used by Neovim's plugins</summary>
+  <url type="homepage">https://github.com/BurntSushi/ripgrep</url>
+  <releases>
+    <release version="14.0.3" date="2023-11-28"/>
+  </releases>
+  <extends>io.neovim.nvim</extends>
+</component>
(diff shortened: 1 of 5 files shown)


io.neovim.nvim.tool.ripgrep.metainfo.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="addon">
+  <id>io.neovim.nvim.tool.ripgrep</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>MIT</project_license>
```
bbhtt, line 5:
```suggestion
  <project_license>MIT and Unlicense</project_license>
```

https://github.com/BurntSushi/ripgrep/blob/master/COPYING

io.neovim.nvim.tool.ripgrep.yml
```diff
@@ -0,0 +1,43 @@
...
+          type: json
+          url: https://api.github.com/repos/BurntSushi/ripgrep/releases/latest
+          tag-query: .tag_name
+          version-query: $tag
+          timestamp-query: .published_at
+          is-main-source: true
+      - cargo-sources.json
+      - type: file
+        path: io.neovim.nvim.tool.ripgrep.metainfo.xml
+      - type: file
```
bbhtt, line 42:
```suggestion
      - type: patch
```

io.neovim.nvim.tool.ripgrep.yml
```diff
@@ -0,0 +1,43 @@
...
+build-options:
+  prefix: /app/tools/ripgrep
+  append-path: /usr/lib/sdk/rust-stable/bin
+  env:
+    - CARGO_HOME: /run/build/ripgrep/cargo
+modules:
+  - name: ripgrep
+    buildsystem: simple
+    build-commands:
+      - git apply config.toml.patch
```
bbhtt, line 20:
```suggestion
```

io.neovim.nvim.tool.ripgrep.yml
franute: nice! I didn't know I could just use a patch directly

bbhtt:
bot, build io.neovim.nvim.tool.ripgrep

barthalion:
Please discuss with freedesktop-sdk folks if they could provide a generic extension point for all apps. We shouldn't be packaging every single CLI tool for each IDE separately.

Lctrs:
Could be a dumb question, but why not building them as `org.freedesktop.Sdk.Extension.*` since all IDEs most likely use the SDK as runtime ?

franute:
> Could be a dumb question, but why not building them as `org.freedesktop.Sdk.Extension.*` since all IDEs most likely use the SDK as runtime ?

Sorry that I haven't had time to look into this. Just had a look at the [flatpak documentation on extensions](https://docs.flatpak.org/en/latest/extension.html) and I believe that's exactly what it suggests for the `texlive` example, so maybe it's worth a try. I'll see if I can ask if that's the right approach. Thanks for pointing that out @Lctrs !

franute:
There's an [open Issue](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1699) at `freedesktop-sdk`, but no activity so far

PR #4839, opened 2024-01-06
Add com.github.stsdc.monitor

stsdc:


com.github.stsdc.monitor.json
```diff
@@ -0,0 +1,347 @@
+{
+  "app-id": "com.github.stsdc.monitor",
```
hfiguiere, line 2: app-id must start with `io.github` as per the documentation.

com.github.stsdc.monitor.json
```diff
@@ -0,0 +1,347 @@
...
+  "runtime": "io.elementary.Platform",
```
hfiguiere, line 3: this runtime doesn't exist on flathub.

com.github.stsdc.monitor.json
```diff
@@ -0,0 +1,347 @@
...
+  "runtime-version": "7.3",
+  "sdk": "io.elementary.Sdk",
+  "command": "com.github.stsdc.monitor",
+  "finish-args": [
+    "--share=ipc",
+    "--socket=fallback-x11",
+    "--socket=wayland",
+    "--socket=session-bus",
+    "--socket=system-bus",
```
hfiguiere, line 12: `session-bus` and `system-bus` are not allowed. You need to list the talk name that are needed.

com.github.stsdc.monitor.json
```diff
@@ -0,0 +1,347 @@
...
+    "--device=shm",
```
hfiguiere, line 13: why do you need device=shm. *hint* you likely don't, it's deprecated.

com.github.stsdc.monitor.json
```diff
@@ -0,0 +1,347 @@
...
+    "--device=all",
+    "--share=network",
+    "--env=XDG_DATA_DIRS=/app/share:/usr/share:/usr/share/runtime/share:/run/host/user-share:/run/host/usr/share:/run/host/share:/var/lib/flatpak/exports/share:~/.local/share/flatpak/exports/share",
```
hfiguiere, line 16: you should never be setting this.

com.github.stsdc.monitor.json
```diff
@@ -0,0 +1,347 @@
...
+    "--filesystem=host:ro",
+    "--filesystem=/var/lib/flatpak/app:ro",
+    "--filesystem=/var/lib/flatpak/exports/share:ro",
+    "--filesystem=xdg-data/flatpak/app:ro",
+    "--filesystem=xdg-data/flatpak/exports/share:ro",
+    "--filesystem=/run/docker.sock",
+    "--filesystem=xdg-run/docker.sock",
+    "--filesystem=/run/podman/podman.sock",
+    "--filesystem=xdg-run/podman/podman.sock",
+    "--filesystem=/tmp"
```
hfiguiere, line 30: I have so many questions about these filesystem permissions. Notably with `/var/lib/flatpak`, why `xdg-data`, etc.

com.github.stsdc.monitor.json
```diff
@@ -0,0 +1,347 @@
...
+          "archive-type": "tar",
+          "url": "https://api.github.com/repos/NVIDIA/nvidia-settings/tarball/refs/tags/470.63.01",
+          "sha256": "0ede63515851d85ac0d1ed5f00e355f539968e6d1fd226120a27b2c66c3575de"
+        },
+        {
+          "type": "patch",
+          "path": "flatpak/patches/libxnvctrl_so_0.patch",
+          "x-checker-data": {
+            "type": "json",
+            "url": "https://api.github.com/repos/NVIDIA/nvidia-settings/tags",
```
hfiguiere, line 325: this is likely mis-placed.

com.github.stsdc.monitor.json
```diff
@@ -0,0 +1,347 @@
...
+    },
+    {
+      "name": "com.github.stsdc.monitor",
+      "buildsystem": "meson",
+      "config-opts": [
+        "-Dflatpak-workaround=enabled"
+      ],
+      "sources": [
+        {
+          "type": "script",
```
hfiguiere, line 340: this is likely incorrect.

you didn't even try to build this, did you?

com.github.stsdc.monitor.json
```diff
@@ -0,0 +1,347 @@
...
+      "build-options": {
+        "env": {
+          "LIB": "lib/x86_64-linux-gnu"
+        }
+      },
+      "no-autogen": true,
+      "no-make-install": true,
+      "build-commands": [
+        "mkdir -p ${FLATPAK_DEST}/lib",
+        "cp -a libXNVCtrl.so* ${FLATPAK_DEST}/lib/",
```
hfiguiere, line 309: so how does this work on aarch64?

hfiguiere, line 309:
```suggestion
        "install -D libXNVCtrl.so* -t ${FLATPAK_DEST}/lib/",
```

com.github.stsdc.monitor.json
stsdc: The same as here: https://github.com/flathub/net.nokyan.Resources/blob/master/net.nokyan.Resources.json#L16

com.github.stsdc.monitor.json
stsdc: Getting access to the podman and docker sockets. Precedent: https://github.com/flathub/com.github.sdv43.whaler/blob/master/com.github.sdv43.whaler.yml#L10-L13

com.github.stsdc.monitor.json
hfiguiere: two wrongs don't make a right.

com.github.stsdc.monitor.json
hfiguiere: two wrongs don't make a right.

com.github.stsdc.monitor.json
stsdc: If wrong is not right then why it is accepted?

com.github.stsdc.monitor.json
stsdc: If there is no GPU it will not show any stats for it. In overall it works, but some sensors may not be detected yet.

com.github.stsdc.monitor.json
hfiguiere: it is being QUESTION. You didn't explain why you need them. Just that you cut and pasted from somewhere.

com.github.stsdc.monitor.json
hfiguiere: you install an x86_64 binary on aarch64. That's the question.

com.github.stsdc.monitor.json
stsdc: It is needed to detect if a process is an app, to find its desktop file and icon.

com.github.stsdc.monitor.json
stsdc: The app lists running containers and their resources.

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+    "only-arches": ["x86_64"]
```
hfiguiere, line 2: why limiting like that when it's just one module?

com.github.stsdc.monitor.json
```diff
@@ -0,0 +1,344 @@
...
+    {
+      "name": "libNVCtrl",
+      "build-options": {
+        "env": {
+          "LIB": "lib/x86_64-linux-gnu"
+        }
+      },
+      "no-autogen": true,
+      "no-make-install": true,
+      "build-commands": [
```
hfiguiere, line 305: actually I misread, this should be 
```suggestion
      "post-install": [
```

and the module should be x86_64 only.

PR #816, opened 2019-01-20
Add com.teamspeak.TeamSpeak

Zlopez:

PR #4851, opened 2024-01-11
Add com.teamspeak.TeamSpeak3

Bixilon:

More like a followup of #816. The maintainer decided to ship the new major version of TeamSpeak (see 

There are quite some people, who want TeamSpeak 3 back for various reasons:
 - https://github.com/flathub/com.teamspeak.TeamSpeak/pull/53
 - https://github.com/flathub/com.teamspeak.TeamSpeak/issues/52
 - https://github.com/flathub/com.teamspeak.TeamSpeak/issues/47

TeamSpeak 3 is still supported by the official developers and TeamSpeak 5 is not even a stable release yet.

Everything in here drops just the last few commits of https://github.com/flathub/com.teamspeak.TeamSpeak and thus rollbacks to version 3.6.2


Bixilon:
bot, build com.teamspeak.TeamSpeak3

func0der:
You could use `TeamSpeak` as the display name, as it is the current stable and the other one is called `TeamSpeak5`

Bixilon:
@func0der I think we should leave it clearly that it starts version 3, as @Zlopez already said: TeamSpeak 5 seems to be the promoted version and maybe they decide to rename that version to just "TeamSpeak".

com.teamspeak.TeamSpeak3.yaml
```diff
@@ -0,0 +1,60 @@
+app-id: com.teamspeak.TeamSpeak3
+tags:
+  - proprietary
+runtime: org.freedesktop.Platform
+runtime-version: '23.08'
+sdk: org.freedesktop.Sdk
+command: teamspeak3
+finish-args:
+  - --socket=x11
+  - --socket=wayland
```
hfiguiere, line 10: it's either `x11` or `fallback-x11` + `wayland`. Also does it suddenly support Wayland? The other package doesn't have it.

Bixilon: Okay, let's drop this for now, will do that sometimes later.

But TS3 is a casual QT application and should work fine with that.

com.teamspeak.TeamSpeak3.yaml
```diff
@@ -0,0 +1,59 @@
...
+    buildsystem: cmake
+    sources:
+      - type: archive
+        url: https://github.com/libevent/libevent/releases/download/release-2.1.12-stable/libevent-2.1.12-stable.tar.gz
+        sha256: 92e6de1be9ec176428fd2367677e61ceffc2ee1cb119035037a27d346b0403bb
+  - name: teamspeak3
+    buildsystem: simple
+    build-commands:
+      - install -D apply_extra /app/bin/apply_extra
+      - install -D teamspeak3 /app/bin/teamspeak3
```
bbhtt, line 28:
```suggestion
      - install -Dm755 apply_extra /app/bin/apply_extra
      - install -Dm755 teamspeak3 /app/bin/teamspeak3
```

Bixilon: Resolved!

Zlopez:
Thanks @Bixilon for doing this.

func0der:
@Bixilon I see a lot of commits going into this. Shouldn't it be enough for an initial release to go back to this commit: https://github.com/flathub/com.teamspeak.TeamSpeak/commit/17526fac88a80f5cbe47ec0486d3c930f27f7164

and be done with version x? Save the the clean up for later basically?

That one worked back then and we would have something back out there for the folks that want it.

I am new to Flathub, so I am not sure if you could do stuff like the permissions on `apply_extra` etc. after the fact or if that is tied to a release on Flathub. (I am not even sure the permission changes are need, because the above commit worked fine :D )

com.teamspeak.TeamSpeak3.yaml
```diff
@@ -0,0 +1,59 @@
...
+        only-arches:
+          - x86_64
+        filename: TeamSpeak3-Client.run
+        url: https://files.teamspeak-services.com/releases/client/3.6.2/TeamSpeak3-Client-linux_amd64-3.6.2.run
+        sha256: 59f110438971a23f904a700e7dd0a811cf99d4e6b975ba3aa45962d43b006422
+        size: 110899830
+        x-checker-data:
+          type: html
+          url: https://files.teamspeak-services.com/releases/client/
+          version-pattern: (\d+\.\d+\.\d+)
```
hfiguiere, line 49: are you sure this won't grab 4.x version?

Bixilon: There is no 4.x version, they skipped to 5.0. I am not sure, but I think the whole check is obsolete (the url is just returning 404)

Bixilon: I implemented a new update checker, but it depends on https://github.com/flathub-infra/flatpak-external-data-checker/pull/410

hfiguiere: are you sure this won't grab 5.x version?


since you want to be pedantic, I rephrased the question.

Bixilon: As I said:
> I am not sure, but I think the whole check is obsolete (the url is just returning 404)

Its broken, I have a local check committed but it depends on the said pull request (its binary data) and does not work with the current yaml format.

hfiguiere: if the check doesn't work why is it here?

Bixilon: Because it was and I did not bother removing it.

PR #4878, opened 2024-01-18
Add org.dune3d.dune3d

carrotIndustries:

### DIFF
--- a/org.dune3d.dune3d.json
+++ b/org.dune3d.dune3d.json
@@ -1,0 +1,131 @@
+{
+    "app-id": "org.dune3d.dune3d",
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "45",
+    "sdk": "org.gnome.Sdk",
+    "command": "dune3d",
+    "finish-args": [
+        "--share=network",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--filesystem=home",
+        "--device=dri",
+        "--filesystem=/run/spnav.sock:ro",
+        "--env=SPNAV_SOCKET=/run/spnav.sock"
+    ],
+    "cleanup": [
+        "*.a",
+        "*.la",
+        "/include",
+        "/lib/*/*.h",
+        "/share/oce-0.18",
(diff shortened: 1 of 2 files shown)


bbhtt:
bot, build org.dune3d.dune3d

org.dune3d.dune3d.json
```diff
@@ -0,0 +1,131 @@
...
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "45",
+    "sdk": "org.gnome.Sdk",
+    "command": "dune3d",
+    "finish-args": [
+        "--share=network",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--filesystem=home",
```
bbhtt, line 12: Why do you need this? Seems to support filechooser portals.

carrotIndustries: The app has a builtin recent chooser that doesn't go through portals:

image

carrotIndustries: The app also needs to load files referenced in the document. While selecting them is done through a portal, loading them when re-opening the document doesn't go through the portal.

hfiguiere: if you use the path returned by the portal then it should just work.

bbhtt: The save dialogue doesn't work with the portal, it opens the correct filechooser but erros when trying to save. Launch it with `--nofilesystem=home`

carrotIndustries: This is probably due to the use of `g_file_set_contents` which creates a differently-named temporary file first:
`/home/lukas/scratch/rect.d3ddoc.UW6JH2`

carrotIndustries: > if you use the path returned by the portal then it should just work.

When the STEP model is in the same directory as the document, the app stores a relative path that's converted to an absolute path with the dirname of the document when opening the document. Since the portal only gives us access to the document and to to any other files in the directory, I think my app unfortunately needs home access.

barthalion:
bot, build org.dune3d.dune3d

PR #4903, opened 2024-01-29
Add com.nitrokey.nitrokey-app2

daringer:


hfiguiere:
what's the difference with this https://github.com/flathub/com.nitrokey.nitrokey-app ?

daringer:
The old nitrokey-app can be used for all the Nitrokey legacy devices, which are not actively developed but in maintenance mode. The new nitrokey-app2 is for all the Nitrokey 3-based devices - thus it makes sense to keep the old one.

pypi-dependencies.json
```diff
@@ -0,0 +1,2321 @@
...
+                }
+            ]
+        },
+        {
+            "name": "python3-cryptography",
+            "buildsystem": "simple",
+            "build-commands": [
+                "pip3 install --no-deps -I --verbose --exists-action=i --no-index --find-links=\"file://${PWD}\" --prefix=${FLATPAK_DEST} \"cryptography==41.0.7\" --no-build-isolation"
+            ],
+            "sources": [
```
bbhtt, line 324: This needs rust can a cargo manifest

pypi-dependencies.json
```diff
@@ -0,0 +1,2321 @@
...
+                {
+                    "type": "file",
+                    "url": "https://files.pythonhosted.org/packages/9b/89/a31c81e36bbb793581d8bba4406a8aac4ba84b2559301c44eef81f4cf5df/cffi-1.16.0-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl",
+                    "sha256": "7b78010e7b97fef4bee1e896df8a4bbb6712b7f05b7ef630f9d1da00f6444d2e"
+                },
+                {
+                    "type": "file",
+                    "url": "https://files.pythonhosted.org/packages/62/bd/69628ab50368b1beb900eb1de5c46f8137169b75b2458affe95f2f470501/cryptography-41.0.7-cp37-abi3-manylinux_2_28_x86_64.whl",
```
bbhtt, line 332: Why is this x86_64 only? Build happen on aarch64 too

pypi-dependencies.json
```diff
@@ -0,0 +1,2321 @@
...
+            ],
+            "sources": [
+                {
+                    "type": "file",
+                    "url": "https://files.pythonhosted.org/packages/07/bc/587a445451b253b285629263eb51c2d8e9bcea4fc97826266d186f96f558/pyserial-3.5-py2.py3-none-any.whl",
+                    "sha256": "c4451db6ba391ca6ca299fb3ec7bae67a5c55dde170964c7a14ceefec02f2cf0"
+                }
+            ]
+        },
+        {
```
bbhtt, line 1417: x86_64 only

pypi-dependencies.json
daringer: yes, `aarch64` builds shouldn't be done - added `flathub.json` with this restriction

pypi-dependencies.json
daringer: yes, `aarch64` builds shouldn't be done - added `flathub.json` with this restriction

pypi-dependencies.json
daringer: This is not entirely clear for me, I checked some other projects and it looks like rust is only needed if one builds `cryptography` from source, but here explicitly the binary `whl` is downloaded for x86_64 - which does not need a compilation so far I understand and observe.

projects compiling cryptography, thus need rust/cargo: [(1)](https://github.com/flathub/com.doycho.euterpe.gtk/blob/51664c5b7210a992faffab71e3f74a7362c07288/cryptography.json#L2) [(2)](https://github.com/flathub/org.qutebrowser.qutebrowser.Userscripts/blob/89dd583a323131974df27007a40769219b501586/python-stem/python-cryptography/python-cryptography.json)
projects pulling in the binary w/o compiling: [(1)](https://github.com/flathub/io.github.dgsasha.Remembrance/blob/master/python3-modules.json) [(2)](https://github.com/flathub/com.icanblink.blink/blob/master/com.icanblink.blink.json) [(3)](https://github.com/flathub/org.gajim.Gajim/blob/8d55c6e716bde390f0f9608c19eb9f43279aa5e9/org.gajim.Gajim.yaml#L101)

is this correct, or am I just misunderstanding something ?

pypi-dependencies.json
bbhtt: Why shouldn't there be aarch64 builds?

daringer: because the underlying `pynitrokey`, which serves as the central library for communication for the nitrokey 3s is also not officially supported on `aarch64` currently

pypi-dependencies.json
bbhtt: It shouldn't, it specifically prefers source packages when available https://github.com/flatpak/flatpak-builder-tools/blob/5880ad0ee58c7e8db73dae4e34fe93e458a7f205/pip/flatpak-pip-generator#L82

bbhtt: You just need to change it to a source package, download it, go to the rust folder and use the cargo-generator from that repo on it to get the cargo manifest and add that here.

You linked to a bunch of manifests compiling it from source, use those as examples.

daringer: ah, ok - will try this, thank you

daringer:
okay, that was "easy" xD ... but hopefully now as required.

* tried to automate as much as possible and included the toolchain in this repo, as I saw this in other repositories
* 3 different packages are built using cargo
* all other packages are built from source, `pyside6` and friends is excluded from this
* documented how an update should look like
* included the generators within `tools` not as submodule, as `flatpak-pip-generator` needs a minimal patch to work with the packages
* currently needs an ugly `sed` to set the correct version (forgot to to update the version in the origin repo)
* patch added to disable updating one Nitrokey 3 Variant due to the missing flatpak capability to dynamically attach serial devices

poetry.lock
bbhtt, line 1: Does the lock file need to be here?

python3-poetry.json
```diff
@@ -0,0 +1,259 @@
+{
+  "name": "poetry",
```
bbhtt, line 2: poetry is needed at build time only? If so you can clean it up using `"cleanup": [ "*" ]`

com.nitrokey.nitrokey-app2.yml
```diff
@@ -0,0 +1,43 @@
...
+  - shared-modules/libusb/libusb.json
+  - pre-pypi-dependencies.json
+  - rust-pypi-dependencies.json
+  - python3-poetry.json
+  - python3-hidapi.json
+  - app-pypi-dependencies.json
+  - name: nkapp
+    buildsystem: simple
+    build-commands:
+      - patch -p1 < disable_nrf_update.patch
```
bbhtt, line 30:
```suggestion
```

com.nitrokey.nitrokey-app2.yml
```diff
@@ -0,0 +1,43 @@
...
+      - sed -i -e 's|2.1.5|2.2.1|g' meta/${FLATPAK_ID}.metainfo.xml
+      - pip install --verbose --exists-action=i --no-index --find-links=\"file://${PWD}\" --prefix=${FLATPAK_DEST} . --no-build-isolation
+      - install -Dm644 meta/${FLATPAK_ID}.metainfo.xml ${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.metainfo.xml
+      - install -Dm644 meta/${FLATPAK_ID}.desktop ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop
+      - install -Dm644 meta/nk-app2.svg ${FLATPAK_DEST}/share/icons/hicolor/scaleable/apps/${FLATPAK_ID}.svg
+      - install -Dm644 meta/nk-app2.png ${FLATPAK_DEST}/share/icons/hicolor/128x128/apps/${FLATPAK_ID}.png
+    sources:
+      - type: file
```
bbhtt, line 39:
```suggestion
      - type: patch
```

com.nitrokey.nitrokey-app2.yml
```diff
@@ -0,0 +1,43 @@
...
+  - pre-pypi-dependencies.json
+  - rust-pypi-dependencies.json
+  - python3-poetry.json
+  - python3-hidapi.json
+  - app-pypi-dependencies.json
+  - name: nkapp
+    buildsystem: simple
+    build-commands:
+      - patch -p1 < disable_nrf_update.patch
+      - sed -i -e 's|2.1.5|2.2.1|g' meta/${FLATPAK_ID}.metainfo.xml
```
bbhtt, line 31: Can you make this a patch too or fix this upstream? sed is likely to replace unintended things.

app-requirements.txt
```diff
@@ -0,0 +1,78 @@
+appdirs==1.4.4 ; python_version >= "3.9" and python_version < "3.13"
```
bbhtt, line 1: Just a note, the pip generator tool can't yet parse python_version syntax. But you'll know the target python version beforehand from the SDK.

poetry.lock
daringer: that's the base for the automation for `app-pypi-dependencies.json` - so not necessarily

python3-poetry.json
daringer: done

com.nitrokey.nitrokey-app2.yml
daringer: will make a patch, this will be fixed with the next release - then the patch can be removed

app-requirements.txt
daringer: k thx, it's just there as this is the poetry export output and I kept it for the other files

com.nitrokey.nitrokey-app2.yml
daringer: done

com.nitrokey.nitrokey-app2.yml
daringer: done

com.nitrokey.nitrokey-app2.yml
daringer: done

bbhtt:
bot, build com.nitrokey.nitrokey-app2

daringer:
great, thanks! can report successful tests on some systems:
* fedora / gnome 
* kubuntu / kde
* arch / no-de compiz

PR #4904, opened 2024-01-29
Add dev.skynomads.Seabird

jgillich:

   * --filesystem=home is required to read certs when a user loads a kubeconfig
    * Yes, I own the domain
    * No additions

This is using official release binaries. I initially tried to build from source, but there's currently no working solution for Go dependencies in builder-tools. Apparently this isn't a concern anyway but I thought I'd mention it. We also have no arm64 binaries yet, they'll be added once some build troubles are ironed out.

No linter errors aside from `appstream-screenshots-not-mirrored-in-ostree`.

Upstream link: https://github.com/getseabird/seabird

### DIFF
--- a/dev.skynomads.Seabird.appdata.xml
+++ b/dev.skynomads.Seabird.appdata.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>dev.skynomads.Seabird</id>
+  <name>Seabird</name>
+  <summary>Kubernetes desktop client</summary>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>MPL-2.0</project_license>
+  <description>
+    <p>Seabird is the native desktop app that simplifies working with Kubernetes.</p>
+  </description>
+  <launchable type="desktop-id">dev.skynomads.Seabird.desktop</launchable>
+  <screenshots>
+    <screenshot type="default">
+      <caption>Seabird main application window</caption>
+      <image>https://getseabird.github.io/images/screenshot.png</image>
+    </screenshot>
+  </screenshots>
+  <url type="homepage">https://getseabird.github.io/</url>
+  <developer_name>Seabird Project</developer_name>
+  <content_rating type="oars-1.1" />
+  <releases>
+    <release version="v0.0.19" date="2024-01-29"/>
+  </releases>
+</component> 
--- a/dev.skynomads.Seabird.desktop
+++ b/dev.skynomads.Seabird.desktop
@@ -1,0 +1,11 @@
+[Desktop Entry]
+Type=Application
+
+Name=Seabird
+Comment=Kubernetes desktop client
+Categories=Development;Network;
+Keywords=kubernetes;
+
+Icon=dev.skynomads.Seabird
+Exec=seabird
+Terminal=false
--- a/dev.skynomads.Seabird.yaml
+++ b/dev.skynomads.Seabird.yaml
@@ -1,0 +1,38 @@
+app-id: dev.skynomads.Seabird
+runtime: org.gnome.Platform
+runtime-version: "45"
+sdk: org.gnome.Sdk
+command: seabird
+
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --device=dri
+  - --talk-name=org.freedesktop.Notifications
+  - --filesystem=home
+
+modules:
+  - name: seabird
+    buildsystem: simple
+    build-commands:
+      - |
+        install -Dm00755 -t /app/bin linux_amd64/seabird
+        install -Dm00644 seabird.svg /app/share/icons/hicolor/scalable/apps/$FLATPAK_ID.svg
+        install -Dm00644 -t /app/share/applications $FLATPAK_ID.desktop
+        install -Dm00644 -t /app/share/metainfo $FLATPAK_ID.appdata.xml
+    sources:
+      - type: file
+        path: dev.skynomads.Seabird.desktop
+      - type: file
+        path: dev.skynomads.Seabird.appdata.xml
+      - type: file
+        filename: seabird.svg
+        url: https://raw.githubusercontent.com/getseabird/seabird/d1c2366c5e6d5f12f070e446ceb99511f56ab7e8/icon/seabird.svg
+        sha256: 6593267d7c60a8a84c350250854fc3f2929c17af612a3b4e44d958b5301cd4fe
+      - type: archive
+        strip-components: 0
+        dest: linux_amd64
+        url: https://github.com/getseabird/seabird/releases/download/v0.0.19/seabird_linux_amd64.tar.gz
+        sha256: 79fdbc6a992e9a3cacb8526b073f03d8fe084c38b141a9000eb15d27f986401f
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.flatpak-builder


jgillich:
bot, build dev.skynomads.Seabird

dev.skynomads.Seabird.yaml
```diff
@@ -0,0 +1,38 @@
...
+        path: dev.skynomads.Seabird.appdata.xml
+      - type: file
+        filename: seabird.svg
+        url: https://raw.githubusercontent.com/getseabird/seabird/d1c2366c5e6d5f12f070e446ceb99511f56ab7e8/icon/seabird.svg
+        sha256: 6593267d7c60a8a84c350250854fc3f2929c17af612a3b4e44d958b5301cd4fe
+      - type: archive
+        strip-components: 0
+        dest: linux_amd64
+        url: https://github.com/getseabird/seabird/releases/download/v0.0.19/seabird_linux_amd64.tar.gz
+        sha256: 79fdbc6a992e9a3cacb8526b073f03d8fe084c38b141a9000eb15d27f986401f
```
bbhtt, line 38: Why not build from source?

dev.skynomads.Seabird.appdata.xml
bbhtt, line 1: Appdata should be in upstream repo

dev.skynomads.Seabird.desktop
bbhtt, line 1: Desktop file should be in upstream repo

dev.skynomads.Seabird.yaml
```diff
@@ -0,0 +1,38 @@
+app-id: dev.skynomads.Seabird
```
bbhtt, line 1: skynomads.dev doesn't seem to resolve. If you own it can you put something up?

dev.skynomads.Seabird.yaml
jgillich: As written above:

> there's currently no working solution for Go dependencies in builder-tools.

Network access during builds isn't permitted, so this is the only way right now

dev.skynomads.Seabird.yaml
jgillich: I put a redirect on https://seabird.skynomads.dev/, good enough?

dev.skynomads.Seabird.appdata.xml
jgillich: Done

dev.skynomads.Seabird.desktop
jgillich: And done

dev.skynomads.Seabird.yaml
hfiguiere: Youn mean tools like this https://github.com/flatpak/flatpak-builder-tools/tree/master/go-get


There is a certain number of go package built on Flathub.

dev.skynomads.Seabird.yaml
```diff
@@ -0,0 +1,34 @@
...
+sdk: org.gnome.Sdk
+command: seabird
+
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --device=dri
+  - --talk-name=org.freedesktop.Notifications
```
hfiguiere, line 13: why do you need this? Gtk4 does not need it.

dev.skynomads.Seabird.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - --filesystem=home
+
+modules:
+  - name: seabird
+    buildsystem: simple
+    build-commands:
+      - |
+        ls -la
```
hfiguiere, line 21:
```suggestion
```

jgillich: oops

dev.skynomads.Seabird.yaml
jgillich: Not sure, saw it somewhere and I assumed it was required for notifications. Not that there are notifications anyway. I removed it

dev.skynomads.Seabird.yaml
jgillich: They don't work with go modules I don't think. This has been [reported](https://github.com/flatpak/flatpak-builder-tools/issues/285) is a [PR](https://github.com/flatpak/flatpak-builder-tools/pull/369) for this but it's still WIP. Or do you have any examples of go packages using modules? Perhaps I did something wrong

hfiguiere: so you ask me to know about go. 

I don't.

I have no idea how a go module is different than the other stuff.

Maybe you can help finish the tools.

hfiguiere:
For the record, the x86_64 only thing is disappointing.

dev.skynomads.Seabird.yaml
jgillich: No worries, go redesigned how dependencies are installed a few years ago and `go get` is the legacy tool that's incompatible with the new module system. So, unfortunately, I don't think it's possible to build from source right now. I might be able to help finish the new tool in the future, but there's also an app I'm working on at the moment. :wink:

jgillich:
> For the record, the x86_64 only thing is disappointing.

I've tried to cross-compile to arm already, but apt on Debian unstable apt just freezes when trying to install gtk deps...well, I'll keep trying, it'll happen eventually

dev.skynomads.Seabird.yaml
hfiguiere: and this new system doesn't support vendoring packages?

hfiguiere:
> > For the record, the x86_64 only thing is disappointing.
> 
> I've tried to cross-compile to arm already, but apt on Debian unstable apt just freezes when trying to install gtk deps...well, I'll keep trying, it'll happen eventually

Good thing we don't cross compile.

dev.skynomads.Seabird.yaml
jgillich: Oh it does support vendoring, but I don't want to put all dependencies in git. I guess I could generate a source + deps tar.gz on release and use that for the flatpak build. But repackaging the binary should be fine for now, no? It can always be improved later

jgillich:
No offense but how is this helpful? I know repackaging binaries isn't ideal, which is why I mentioned it in the PR. I'd also like to provide arm64 binaries. We're in agreement on this. But neither of these are actually mandated by Flathub, correct? The bickering is honestly a little frustrating. This is the initial version, I plan to fix these later. But my time is limited and I don't see this as something that needs resolving immediately.

Again, no offense intended, I appreciate the work you guys do. But if you always strive for perfection, you'll never get anything done. :wink:

dev.skynomads.Seabird.yaml
bbhtt: Yes, thanks

flathub.json
```diff
@@ -0,0 +1,6 @@
+{
+  "only-arches": [
+      "x86_64"
+  ],
+  "automerge-flathubbot-prs": true
```
bbhtt, line 5: Please don't automerge on applications, this means builds will get pushed to users without any testing

bbhtt:
Seems fine otherwise.

jgillich:
bot, build dev.skynomads.Seabird

jgillich:
Dropped the automerge and made the switch to build from a vendored source archive

jgillich:
Thanks guys! Does I need to do anything to publish it? It shows a 404 on Flathub.org still. Docs say it should be auto-published after 3 hours, but the build completed over 4 hours ago: https://buildbot.flathub.org/#/apps/dev.skynomads.Seabird

bbhtt:
The publish job just started https://buildbot.flathub.org/#/builders/2/builds/47562 it'll take some time to appear on the website

jgillich:
I see, thanks!

PR #4913, opened 2024-02-02
Add org.asamk.SignalCli

AsamK:

  This is a cli application so it doesn't provide a .desktop file.
  I've added `"--own-name=org.asamk.Signal"` and `"--talk-name=org.asamk.Signal"`, as that is the dbus name that signal-cli currently uses.
  `"--filesystem=home"` is there, so that users can send attachments from files in their home directory.
  I've added the appstream and icon file here as well, as I only need them for flathub.


flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "skip-arches": ["aarch64"]
```

org.asamk.SignalCli.metainfo.xml
hfiguiere, line 1: this should be managed upstream.

hfiguiere (changes requested):
missing desktop file.

flathub.json
hfiguiere: even more since it looks like this is written in Java.

AsamK: It's because a dependency (libsignal-client) uses a native library (written in rust), that's unfortunately only included for x86_64.

org.asamk.SignalCli.json
```diff
@@ -0,0 +1,52 @@
...
+      "buildsystem": "simple",
+      "build-commands": [
+        "cp -r signal-cli/* /app/",
+        "install -Dm644 org.asamk.SignalCli.metainfo.xml /app/share/metainfo/org.asamk.SignalCli.metainfo.xml",
+        "install -Dm644 org.asamk.SignalCli.svg /app/share/icons/hicolor/scalable/apps/org.asamk.SignalCli.svg"
+      ],
+      "sources": [
+        {
+          "type": "archive",
+          "url": "https://github.com/AsamK/signal-cli/releases/download/v0.12.7/signal-cli-0.12.7.tar.gz",
```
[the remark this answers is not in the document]
AsamK: Because it's not so straight forward with gradle, that expects to be able to download dependencies during the build.

org.asamk.SignalCli.metainfo.xml
AsamK: Is it ok to leave it here for now?

AsamK:
> missing desktop file.

This is a commandline application (not TUI) so a desktop file is not useful for users.
It also shouldn't be required according to the [App Requirements](https://docs.flathub.org/docs/for-app-authors/requirements/#desktop-integration)

flathub.json
hfiguiere: why not building this library from source?

org.asamk.SignalCli.json
```diff
@@ -0,0 +1,52 @@
+{
+  "id": "org.asamk.SignalCli",
+  "runtime": "org.freedesktop.Platform",
+  "runtime-version": "23.08",
+  "sdk": "org.freedesktop.Sdk",
+  "sdk-extensions" : [ "org.freedesktop.Sdk.Extension.openjdk21" ],
+  "command": "signal-cli",
+  "finish-args": [
+    "--own-name=org.asamk.Signal",
+    "--talk-name=org.asamk.Signal",
```
bbhtt, line 10: This should be fixed upstream... The id and bus names you use internally should match with the Flatpak id too.

org.asamk.SignalCli.svg
bbhtt, line 1: Icon should be upstream

org.asamk.SignalCli.json
AsamK: I've removed them.

org.asamk.SignalCli.svg
AsamK: done

bbhtt:
bot, build org.asamk.SignalCli

bbhtt:
bot, build org.asamk.SignalCli

PR #4935, opened 2024-02-09
Add org.freedesktop.Sdk.Extension.swift5

david-swift:

- [ ] I am an upstream contributor to the project. If not, I contacted upstream developers about submitting their software to Flathub.
- [ ] Any additional patches or files have been submitted to the upstream projects concerned.

Do the last three todos also apply to SDK extensions?

Thanks for reviewing already in advance!

### DIFF
--- a/org.freedesktop.Sdk.Extension.swift.json
+++ b/org.freedesktop.Sdk.Extension.swift.json
@@ -22,47 +22,25 @@
     ],
     "modules": [
         {
-            "name": "libedit",
-            "buildsystem": "autotools",
-            "sources": [
-                {
-                    "type": "archive",
-                    "url": "https://www.thrysoee.dk/editline/libedit-20221030-3.1.tar.gz",
-                    "sha256": "f0925a5adf4b1bf116ee19766b7daa766917aec198747943b1c4edf67a4be2bb"
-                }
-            ],
-            "cleanup": [
-                "/include",
-                "/lib/pkgconfig",
-                "/share/man"
-            ]
-        },
-        {
             "name": "swift",
             "buildsystem": "simple",
             "sources": [
                 {
                     "type": "archive",
                     "only-arches": ["x86_64"],
-                    "url": "https://download.swift.org/swift-5.9.1-release/ubuntu2204/swift-5.9.1-RELEASE/swift-5.9.1-RELEASE-ubuntu22.04.tar.gz",
-                    "sha256": "ea2fe1190a9cb8abed8c5e7b94223a06a23c7dc8cd498850a1c79b8a87e7c251"
+                    "url": "https://download.swift.org/swift-5.9.2-release/amazonlinux2/swift-5.9.2-RELEASE/swift-5.9.2-RELEASE-amazonlinux2.tar.gz",
+                    "sha256": "bdfa4db698458963c1dfddff1523a69476626e25910177cc91add143e9130cde"
                 },
                 {
                     "type": "archive",
                     "only-arches": ["aarch64"],
-                    "url": "https://download.swift.org/swift-5.9.1-release/ubuntu2204-aarch64/swift-5.9.1-RELEASE/swift-5.9.1-RELEASE-ubuntu22.04-aarch64.tar.gz",
-                    "sha256": "8331bc3f50dbc484d1e4c939aad8399b567fced63d59167c7c2169ee5ff7168f"
+                    "url": "https://download.swift.org/swift-5.9.2-release/amazonlinux2-aarch64/swift-5.9.2-RELEASE/swift-5.9.2-RELEASE-amazonlinux2-aarch64.tar.gz",
+                    "sha256": "c7393dd0ff3bee8e8ea0ea35f8e0d8da6e2e31efd1cc1fd996403aa8976cfb64"
                 }
             ],
             "build-commands": [
                 "mkdir -p /usr/lib/sdk/swift",
-                "cp -R usr/* /usr/lib/sdk/swift",
-                "sed -i s/libedit.so.2/libedit.so.0/g /usr/lib/sdk/swift/bin/lldb-server",
-                "sed -i s/libedit.so.2/libedit.so.0/g /usr/lib/sdk/swift/lib/liblldb.so.13.0.0git",
-                "sed -i s/libedit.so.2/libedit.so.0/g /usr/lib/sdk/swift/lib/liblldb.so",
-                "sed -i s/libedit.so.2/libedit.so.0/g /usr/lib/sdk/swift/lib/liblldb.so.13git",
-                "ln -s /usr/lib/x86_64-linux-gnu/libtinfo.so.6 /usr/lib/sdk/swift/lib/libtinfo.so.5",
-                "ln -s /usr/lib/x86_64-linux-gnu/libncurses.so.6 /usr/lib/sdk/swift/lib/libncurses.so.5"
+                "cp -R usr/* /usr/lib/sdk/swift"
             ]
         },
         {
--- a/org.freedesktop.Sdk.Extension.swift.json
+++ b/org.freedesktop.Sdk.Extension.swift.json
@@ -43,8 +43,15 @@
             "sources": [
                 {
                     "type": "archive",
-                    "url": "https://download.swift.org/swift-5.9.1-release/centos7/swift-5.9.1-RELEASE/swift-5.9.1-RELEASE-centos7.tar.gz",
-                    "sha256": "85f76d092e50794cafc45229d6534c02abb25f4f62f8d923c83b2e7c08a6eada"
+                    "only-arches": ["x86_64"],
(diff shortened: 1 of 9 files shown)


org.freedesktop.Sdk.Extension.swift.json
```diff
@@ -0,0 +1,206 @@
...
+                "/share/man"
+            ]
+        },
+        {
+            "name": "swift",
+            "buildsystem": "simple",
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://download.swift.org/swift-5.9.1-release/centos7/swift-5.9.1-RELEASE/swift-5.9.1-RELEASE-centos7.tar.gz",
```
hfiguiere, line 46: how will this work on aarch64?

hfiguiere: also you can take something more recent that centos7. The Freedesktop SDK is a pretty recent stack

org.freedesktop.Sdk.Extension.swift.json
```diff
@@ -0,0 +1,206 @@
...
+                    "sha256": "85f76d092e50794cafc45229d6534c02abb25f4f62f8d923c83b2e7c08a6eada"
+                }
+            ],
+            "build-commands": [
+                "mkdir -p /usr/lib/sdk/swift",
+                "cp -R usr/* /usr/lib/sdk/swift",
+                "sed -i s/libedit.so.2/libedit.so.0/g /usr/lib/sdk/swift/bin/lldb-server",
```
hfiguiere, line 53: this is really really dirty to use sed to change the soversion because they don't match.

org.freedesktop.Sdk.Extension.swift.json
```diff
@@ -0,0 +1,213 @@
...
+                "sed -i s/libedit.so.2/libedit.so.0/g /usr/lib/sdk/swift/lib/liblldb.so.13.0.0git",
+                "sed -i s/libedit.so.2/libedit.so.0/g /usr/lib/sdk/swift/lib/liblldb.so",
+                "sed -i s/libedit.so.2/libedit.so.0/g /usr/lib/sdk/swift/lib/liblldb.so.13git",
+                "ln -s /usr/lib/x86_64-linux-gnu/libtinfo.so.6 /usr/lib/sdk/swift/lib/libtinfo.so.5",
+                "ln -s /usr/lib/x86_64-linux-gnu/libncurses.so.6 /usr/lib/sdk/swift/lib/libncurses.so.5"
```
hfiguiere, line 65: this will fail on aarch64

bbhtt:
What's the usecase for this on Flathub?

org.freedesktop.Sdk.Extension.swift.json
david-swift: I switched to Ubuntu 22.04, which has an x86_64 and an aarch64 download.

david-swift:
> What's the usecase for this on Flathub?

I'm working on [adwaita-swift](https://github.com/AparokshaUI/adwaita-swift) for creating GNOME apps using Swift in a readable, declarative way. I am e.g. working on [Flashcards](https://github.com/david-swift/Flashcards). The API is very similar to SwiftUI used for iOS/macOS development, so hopefully, it might attract some developers currently focused on Apple platforms into the Linux world. Swift could also be an option for other projects, as it offers good C and C++ interoperability.

org.freedesktop.Sdk.Extension.swift.json
david-swift: I switched once again, this time to the Amazon Linux 2 downloads, because I could remove the "really really dirty" part and the lines that fail on aarch64 that way.

hfiguiere:
any tooling to generate manifest for swift packages like we have of npn, cargo, pip?

hfiguiere:
> What's the usecase for this on Flathub?

There is at least one package on flathub that is built with Swift.

david-swift:
> any tooling to generate manifest for swift packages like we have of npn, cargo, pip?

There's the [Swift Package Manager](https://www.swift.org/documentation/package-manager/) which comes with the Swift download itself, so it is already included in this extension.

david-swift:
`swift package init` generates a starting point for a package including the manifest.

hfiguiere:
> There's the [Swift Package Manager](https://www.swift.org/documentation/package-manager/) which comes with the Swift download itself, so it is already included in this extension.

According to the doc:

> The role of the package manager is to reduce coordination costs by automating the process of downloading and building all of the dependencies for a project.

Flathub builds are done *offline* so none of this can happen during the build. 

"starting point" isn't really what I have in mind. Generating a manifest ready to use is what the tooling does (more or less with bugs and other).

hfiguiere:
bot, build org.freedesktop.Sdk.Extension.swift

org.freedesktop.Sdk.Extension.swift.json
```diff
@@ -0,0 +1,191 @@
...
+    "sdk-extensions": [
+        "org.freedesktop.Sdk.Extension.llvm16"
+    ],
+    "runtime-version": "23.08",
+    "separate-locales": false,
+    "appstream-compose": false,
+    "build-options": {
+        "prefix": "/usr/lib/sdk/swift",
+        "prepend-pkg-config-path": "/usr/lib/sdk/swift/lib/pkgconfig/",
+        "prepend-path": "/usr/lib/sdk/swift/bin/:/usr/lib/sdk/llvm14/bin",
```
hfiguiere, line 16: extension is llvm16

hfiguiere:
what's the plan for future version of the toolchain? It has been my understanding that Swift version brought need to update code. llvm use major versions as part of the package name. So does JDK. Or dotNET.

hfiguiere:
Also it would be nice it have an appstream file so the version can be displayed where appropriate.

david-swift:
> > There's the [Swift Package Manager](https://www.swift.org/documentation/package-manager/) which comes with the Swift download itself, so it is already included in this extension.
> 
> According to the doc:
> 
> > The role of the package manager is to reduce coordination costs by automating the process of downloading and building all of the dependencies for a project.
> 
> Flathub builds are done _offline_ so none of this can happen during the build.
> 
> "starting point" isn't really what I have in mind. Generating a manifest ready to use is what the tooling does (more or less with bugs and other).

Ah, you're talking about a Flatpak manifest file? In this extension, there are also two Swift packages built, and its dependencies were added manually to the manifest file. I think it makes sense to automate this, so I'll look into implementing this.

hfiguiere:
Not only for this extension. But for packagers using it. Usually these "always online" package managers are the biggest barrier. While our tooling isn't perfect https://github.com/flatpak/flatpak-builder-tools , it sure beat no support.

org.freedesktop.Sdk.Extension.swift.json
```diff
@@ -0,0 +1,191 @@
...
+    "sdk-extensions": [
+        "org.freedesktop.Sdk.Extension.llvm16"
+    ],
+    "runtime-version": "23.08",
+    "separate-locales": false,
+    "appstream-compose": false,
+    "build-options": {
+        "prefix": "/usr/lib/sdk/swift",
+        "prepend-pkg-config-path": "/usr/lib/sdk/swift/lib/pkgconfig/",
+        "prepend-path": "/usr/lib/sdk/swift/bin/:/usr/lib/sdk/llvm16/bin",
```
hfiguiere, line 16: funny thing the build seems to have passed even with the wrong path.

david-swift:
Just to be sure the message gets to you, @hfiguiere (you were _incredibly_ fast with the other reviews, so I'm not quite sure whether you saw the review request I opened yesterday - don't feel stressed though 😀): I renamed the extension to `org.swift.freedesktop.Sdk.Extension.swift5` and removed the useless build options so I think the PR would be ready for another review/test build.

david-swift:
bot, build org.freedesktop.Sdk.Extension.swift5

david-swift:
@hfiguiere Just realized I can request a build as well. I tested this build, it is working as expected.

org.freedesktop.Sdk.Extension.swift5.metainfo.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2024 david-swift -->
+<component type="addon">
+  <id>org.freedesktop.Sdk.Extension.swift5</id>
+  <extends>org.freedesktop.Sdk</extends>
+  <metadata_license>MIT</metadata_license>
+  <name>Swift 5</name>
+  <summary>Swift 5 compiler and tools for the Flatpak Freedesktop SDK</summary>
+  <description>This extension adds Swift support to Flatpak.</description>
```
bbhtt, line 9:
```suggestion
  <description>This extension adds Swift support to Flatpak</description>
```

david-swift:
Thank you!

PR #4948, opened 2024-02-15
Add com.visualstudio.code.tool.docker-compose

xduugu:


bbhtt:
CLI tools packaged as IDE extensions belongs in a generic extension point. You can ask in https://gitlab.com/freedesktop-sdk/freedesktop-sdk/ to provide one.

Otherwise the same extension needs to be packaged multiple times for multiple different IDEs.

xduugu:
@bbhtt, I've opened an [issue at freedesktop-sdk](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1699), but it doesn't look like there is much interest in this topic. Actually, I am not interested in it as well, because I think trying to package every development tool for flatpak is not the right direction.

That's why I use [devcontainers](https://code.visualstudio.com/docs/devcontainers/create-dev-container) with vscode, which only requires [podman-remote](https://github.com/flathub/com.visualstudio.code.tool.podman), which is already packaged for vscode, and possibly `docker-compose`. Since `docker-compose` is not called directly, but only by `podman-remote compose`, I see it as an extension for `podman-remote`. However, it does not fit into the existing `podman-remote` repository, that's why I opened this PR.

bbhtt:
We can't be hosting, multiple versions of the same extension for each family of IDE. That duplicates work and maintenance across the board.

Since most of the IDEs have flatpak-spawn access, you can easily launch host tools without needing to package them with flatpak.

PR #4968, opened 2024-02-19
Add net.codelogistics.webapps

eyekay-gh:

### DIFF
--- a/net.codelogistics.webapps.json
+++ b/net.codelogistics.webapps.json
@@ -32,7 +32,7 @@
                 {
                     "type" : "git",
                     "url" : "https://codeberg.org/eyekay/webapps.git",
-                    "tag" : "0.2.0"
+                    "tag" : "0.2.2"
                 }
             ]
         }
--- a/net.codelogistics.webapps.json
+++ b/net.codelogistics.webapps.json
@@ -46,7 +46,7 @@
                 {
                     "type" : "git",
                     "url" : "https://codeberg.org/eyekay/webapps.git",
-                    "tag" : "0.3.0"
+                    "tag" : "0.3.1"
                 }
             ]
         }
--- a/net.codelogistics.webapps.json
+++ b/net.codelogistics.webapps.json
@@ -46,7 +46,7 @@
                 {
                     "type" : "git",
                     "url" : "https://codeberg.org/eyekay/webapps.git",
-                    "tag" : "0.3.1"
+                    "tag" : "0.3.2"
                 }
             ]
         }
--- a/net.codelogistics.webapps.json
+++ b/net.codelogistics.webapps.json
@@ -46,7 +46,7 @@
                 {
                     "type" : "git",
                     "url" : "https://codeberg.org/eyekay/webapps.git",
-                    "tag" : "0.3.2"
+                    "tag" : "0.3.3"
                 }
             ]
         }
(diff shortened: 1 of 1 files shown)


net.codelogistics.webapps.json
```diff
@@ -0,0 +1,40 @@
+{
+    "app-id" : "net.codelogistics.webapps",
+    "runtime" : "org.gnome.Platform",
+    "runtime-version" : "44",
```
hfiguiere, line 4: '45' is the current.

net.codelogistics.webapps.json
```diff
@@ -0,0 +1,40 @@
...
+    "sdk" : "org.gnome.Sdk",
+    "command" : "webapps",
+    "finish-args" : [
+        "--share=network",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--device=dri",
+        "--socket=wayland",
+        "--filesystem=xdg-data"
```
hfiguiere, line 13: why do you need this?

hfiguiere:
bot, build net.codelogistics.webapps

net.codelogistics.webapps.json
hfiguiere: it's not allowed by default.

hfiguiere: You should use this instead https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.DynamicLauncher.html

eyekay-gh: Thank you so much, the program now uses portals, so I have removed that line from the manifest.

bbhtt:
bot, build net.codelogistics.webapps

net.codelogistics.webapps.json
```diff
@@ -0,0 +1,54 @@
+{
+    "app-id" : "net.codelogistics.webapps",
+    "runtime" : "org.gnome.Platform",
+    "runtime-version" : "45",
```
bbhtt, line 4: I'd suggest updating to 46.

bbhtt:
This seems to redirect to a regular Firefox instance when creating the webapp. I'm not sure if that is intended?

How is that a webapp?

eyekay-gh:
> This seems to redirect to a regular Firefox instance when creating the webapp. I'm not sure if that is intended?
> 
> How is that a webapp?

That should not be possible, the app does not call the system web browser or ask to open a URL at any point.

I just installed the app as a Flatpak on a fresh install, and when I create a new webapp, the Flatpak/ freedesktop portal to create a new desktop launcher appears (as expected) and the desktop file available through the menu calls `flatpak run --command=webapps 'net.codelogistics.webapps'`.

net.codelogistics.webapps.json
eyekay-gh: Done, thank you

bbhtt:
bot, build net.codelogistics.webapps

bbhtt:
Seems like it works but if I do https://mail.google.com it opens a youtube page in the default browser along with the webapp. I assume that's not coming from the app?


https://github.com/flathub/flathub/assets/62639087/6af0757a-ab5b-4a92-8c5d-935dd3ec6b90

eyekay-gh:
> Seems like it works but if I do https://mail.google.com it opens a youtube page in the default browser along with the webapp. I assume that's not coming from the app?
> Screencast.mp4

The issue here is that my app only allows webpages from the original domain name associated with the webapp to be opened inside it. Google apparently loads the youtube.com domain at that point of the login process, which gets opened in the default browser as it does not match the original domain (google.com). 

I tried this as well, you can safely close that tab. It does not affect the sign-in process.

PR #4980, opened 2024-02-21
Improve the PR template

bbhtt:
- Add info on how to mark items as done
- Add author/developer too to the "upstream contributor" contact line


hfiguiere:
I'd take the opportunity to add a section at the top:

"Please describe the application"

Often we have to guess or dig what it is doing.

hfiguiere (approved):

PR #5021, opened 2024-03-05
Add com.nickgirga.webready

nickgirga:

WebReady is a tool to make WEBP animations from video files in a matter of seconds. You can simply right-click a video, open it with WebReady, and press ENTER. By default, it creates WEBP animated images that are configured perfectly for use as live wallpapers in Gaming Mode on the Steam Deck. This is its primary purpose.

`--filesystem=home` is used to easily access chosen video files and the path to the output file. This also simplifies the "Open With" option for automatically choosing the file that the user opened WebReady with.

### DIFF
--- a/com.nickgirga.webready.yml
+++ b/com.nickgirga.webready.yml
@@ -10,24 +10,10 @@ finish-args:
   - --socket=fallback-x11
   # Wayland access
   - --socket=wayland
-  # access to home files; for locating WADs and custom installed source ports
+  # access to home files; for easy access to videos and "Open With"
   - --filesystem=home
 
 modules:
-  - name: webready
-    buildsystem: simple
-    build-commands:
-      - install -D webready /app/bin/webready
-      - install -D webready.glade /app/share/webready/webready.glade
-      - install -D res/icons/scalable.svg /app/share/icons/hicolor/scalable/apps/com.nickgirga.webready.svg
-      - install -D res/icons/128x.png /app/share/icons/hicolor/128x128/apps/com.nickgirga.webready.png
-      - install -D com.nickgirga.webready.desktop /app/share/applications/com.nickgirga.webready.desktop
-      - install -Dm644 com.nickgirga.webready.metainfo.xml -t /app/share/metainfo
-    sources:
-      - type: git
-        url: https://gitlab.com/nickgirga/webready.git
-        tag: 1.0.1
-
   - name: ffmpeg_progress_yield
     buildsystem: simple
     build-commands:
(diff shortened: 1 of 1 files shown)


com.nickgirga.webready.yml
```diff
@@ -0,0 +1,49 @@
...
+  - --filesystem=home
+
+modules:
+  - name: webready
+    buildsystem: simple
+    build-commands:
+      - install -D webready /app/bin/webready
+      - install -D webready.glade /app/share/webready/webready.glade
+      - install -D res/icons/scalable.svg /app/share/icons/hicolor/scalable/apps/com.nickgirga.webready.svg
+      - install -D res/icons/128x.png /app/share/icons/hicolor/128x128/apps/com.nickgirga.webready.png
```
bbhtt, line 23:
```suggestion
```

svg is enough

com.nickgirga.webready.yml
```diff
@@ -0,0 +1,49 @@
...
+  - name: libhandy
+    buildsystem: meson
+    config-opts:
+      - -Dglade_catalog=disabled
+      - -Dtests=false
+      - -Dexamples=false
+    sources:
+      - type: git
+        url: https://gitlab.gnome.org/GNOME/libhandy
+        tag: 1.0.3
```
bbhtt, line 49: Do you need 1.0.3? Runtime has 1.8.3

hfiguiere:
This is missing:

> Please describe your application briefly below.

com.nickgirga.webready.yml
```diff
@@ -0,0 +1,49 @@
...
+sdk: org.gnome.Sdk
+command: webready
+
+finish-args:
+  # X11 + XShm access
+  - --share=ipc
+  - --socket=fallback-x11
+  # Wayland access
+  - --socket=wayland
+  # access to home files; for locating WADs and custom installed source ports
```
hfiguiere, line 13: What's the relation with this comment and what the app seems to do?

com.nickgirga.webready.yml
nickgirga: I was using the 128x PNG as a resource for the about page. Just updated it to create a Pixbuf from the SVG. Will create a new release tag and update the flatpak manifest and stop including the PNG.

com.nickgirga.webready.yml
nickgirga: Oh, great. I didn't know that. I will remove libhandy as a module.

com.nickgirga.webready.yml
nickgirga: Sorry. I used an old manifest of mine for another app as a template. Forgot to update the comment. The permission is needed for this app. Will update it.

nickgirga:
> This is missing:
> 
> > Please describe your application briefly below.

Sorry. I thought filling out the checkboxes WAS describing the app. I'll update the MR/PR with an accurate text description of what my app does.

nickgirga:
The suggested changes have been made and I verified that my app still builds and runs correctly. I also verified that the manifest and the repo do not throw any warnings or errors with the linter.

com.nickgirga.webready.yml
```diff
@@ -0,0 +1,37 @@
...
+command: webready
+
+finish-args:
+  # X11 + XShm access
+  - --share=ipc
+  - --socket=fallback-x11
+  # Wayland access
+  - --socket=wayland
+  # access to home files; for easy access to videos and "Open With"
+  - --filesystem=home
```
bbhtt, line 14: Seems to use the portal file. Doesn't seem to be required.

nickgirga: The file chooser works okay without that permission, but the output file (which should end up right next to the source video file) will end up in the "/run/user/..." directories instead, which is not desirable behavior.

bbhtt: Ah right forgot about that

PR #5023, opened 2024-03-06
Add one.jwr.interstellar

jwr1:

Interstellar is a Flutter app that allows you to sign in and access your kbin, mbin, and lemmy fediverse accounts. It basically lets you browse, vote, and create posts and comments that are grouped into specific communities.

- [x] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about submitting their software to Flathub.


one.jwr.interstellar.desktop
bbhtt, line 1: Desktop file should be upstream

one.jwr.interstellar.metainfo.xml
bbhtt, line 1: Metainfo file should be upstream

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "skip-arches": ["aarch64"]
```
bbhtt, line 2: Can you provide aarch64 packages?

one.jwr.interstellar.metainfo.xml
```diff
@@ -0,0 +1,60 @@
...
+    <content_attribute id="social-audio">intense</content_attribute>
+  </content_rating>
+
+  <url type="bugtracker">https://github.com/jwr1/interstellar/issues</url>
+  <url type="contact">https://matrix.to/#/#kbin-interstellar:matrix.org</url>
+  <url type="vcs-browser">https://github.com/jwr1/interstellar</url>
+
+  <screenshots>
+    <screenshot type="default">
+      <image>https://raw.githubusercontent.com/jwr1/interstellar/main/assets/screenshots/desktop-1.png</image>
```
bbhtt, line 42: Screenshots should be from a tag or a commit

one.jwr.interstellar.yaml
```diff
@@ -0,0 +1,132 @@
...
+          - --disable-programs
+          - --disable-encoders
+          - --disable-muxers
+          - --enable-encoder=png
+          - --enable-libv4l2
+          - --enable-libdav1d
+        sources:
+          - type: git
+            url: https://git.ffmpeg.org/ffmpeg.git
+            tag: n6.0.1
```
bbhtt, line 102: ffmpge-full extension has 6.0.1

flathub.json
jwr1: The package is built using GitHub actions which only lets you use x86_64 architecture. I would like to support arm, but I'm just not sure how I would do that until github adds arm support to action runners.

hfiguiere:
```suggestion
  "only-arches": ["x86_64"]
```

run.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+LD_LIBRARY_PATH=/app/lib /app/interstellar/interstellar "$@"
```
hfiguiere, line 3:
```suggestion
exec /app/interstellar/interstellar "$@"
```

one.jwr.interstellar.yaml
```diff
@@ -0,0 +1,132 @@
...
+  - name: interstellar
+    buildsystem: simple
+    only-arches:
+      - x86_64
+    build-commands:
+      - 'cp -r interstellar/ /app/interstellar/'
+      - 'install -Dm644 interstellar/data/flutter_assets/assets/icons/logo.png /app/share/icons/hicolor/512x512/apps/one.jwr.interstellar.png'
+      - 'install run.sh /app/bin/interstellar'
+      - 'cp /app/lib/libmpv.so /app/lib/libmpv.so.1'
```
hfiguiere, line 111: why do you need to do that?

jwr1: Sorry, do what?

`cp -r interstellar/ /app/interstellar/`: this copies over the app binaries, libraries, and other assets.
`install -Dm644 interstellar/data/flutter_assets/assets/icons/logo.png /app/share/icons/hicolor/512x512/apps/one.jwr.interstellar.png`: this copies the app icon to the correct location.
`install run.sh /app/bin/interstellar` this copies over the script that runs the app correctly.
`cp /app/lib/libmpv.so /app/lib/libmpv.so.1` libmpv.so is a link to libmpv.so.2 so this makes it so libmpv.so.1 points to the libmpv library, as the the app expects the library to be at libmpv.so.1 instead of libmpv.so.2.

one.jwr.interstellar.metainfo.xml
jwr1: The app is cross-platform, and managing the metainfo.xml and desktop files from this end would be easier, especially considering they are flatpak-specific. I saw the [Requirements page](https://docs.flathub.org/docs/for-app-authors/requirements/#acceptable-but-should-be-submitted-upstream) in the flathub docs recommends having these files upsteam, but it does say it's acceptable to have it here as well, especially since it's cross-platform.

one.jwr.interstellar.yaml
hfiguiere: > cp /app/lib/libmpv.so /app/lib/libmpv.so.1 libmpv.so is a link to libmpv.so.2 so this makes it so libmpv.so.1 points to the libmpv library, as the the app expects the library to be at libmpv.so.1 instead of libmpv.so.2.

That's the only line I cared.

So basically you are just violating the ABI guarantees instead of building the right version. This is a bad idea. If it works then maybe you are in luck.

one.jwr.interstellar.metainfo.xml
hfiguiere: > The app is cross-platform,

Apparently not since it doesn't run on aarch64.

hfiguiere: this is NOT flatpak specific either.

jwr1: The linux version is not multi-architectured, but yes the app itself is cross-platform. It currently has builds for linux, android, and windows.

one.jwr.interstellar.yaml
jwr1: I certainly hadn't meant to break or violate anything. I'll try to update the .yaml file to build on mpv version 1.

jwr1: The correct libmpv version is now used. Thanks for the help, I wouldn't have seen that otherwise.

run.sh
jwr1: This is fixed now, once the "ugly hack" was removed.

one.jwr.interstellar.yaml
jwr1: I had to downgrade ffmpeg to 5.1.4. Does the ffmpeg extension have that version, if so how should I set up the extension?

bbhtt: No it doesn't

bbhtt:
bot, build one.jwr.interstellar

jwr1:
One question before this gets merged. I've been testing the flatpak version of the app, and it appears the gtk theme from the operating system isn't being passed in, is there an easy way to fix this? I'm assuming there's something I need to add to the `finish-args`.

bbhtt:
Themes should work automatically. If you are on wayland, the theme from host is communicated to flatpaks via portal and flatpak will automatically install the theme. If you are on X11 same thing happens xsettings daemon communicates it to flatpak.

This should work on both KDE/GNOME, other DEs/WMs might be lacking the proper support you have to check.

The theme needs to be packaged, so custom modifications/niche themes won't work.

jwr1:
OK, thanks. I'm assuming it's a bug with my desktop environment then, because I'm using KDE Wayland with breeze, but I did just upgrade to plasma 6 so maybe it's a bug they just didn't fix yet.

PR #5038, opened 2024-03-11
Add io.github.cudatext.CudaText-Qt

JordanL2:
This is a rename of io.github.cudatext.CudaText-Qt5, as this version has been built with Qt6.

The app ID io.github.cudatext.CudaText-Qt5 was originally chosen as the app can be built with different toolkits, and it was the version built with Qt5. However, I didn't consider what we'd do when it came time to upgrade the app to Qt6 or later. Keeping the suffix "-Qt5" obviously doesn't make sense, and having a different version for each Qt version wouldn't be great either (especially when I'm only planning to maintain the latest one). While we could remove the suffix altogether, I wouldn't like to potentially block (or make more difficult) a GTK version of the app being put on Flathub. So It seems the best option is renaming it to "io.github.cudatext.CudaText-Qt", which will be the version using whatever the latest version of Qt is.

The original io.github.cudatext.CudaText-Qt5 app will be deprecated and users will be migrated to this app.

-------

### DIFF
--- a/io.github.cudatext.CudaText-Qt.appdata.xml
+++ b/io.github.cudatext.CudaText-Qt.appdata.xml
@@ -1,0 +1,108 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>io.github.cudatext.CudaText-Qt</id>
+  <project_license>MPL-2.0</project_license>
+  <metadata_license>FSFAP</metadata_license>
+  <name>CudaText</name>
+  <summary>Open source and cross-platform text editor</summary>
+
+  <description>
+    <p>CudaText is a cross-platform text editor, written in Object Pascal. It is open source project and can be used free of charge, even for business. It starts quite fast: ~0.3 sec with ~30 plugins, on Linux on CPU Intel Core i3 3GHz. It is extensible by Python add-ons: plugins, linters, code tree parsers, external tools. Syntax parser is feature-rich, based on EControl engine (though not as fast as in some competitors).</p>
+
+    <p>Features</p>
+    <ul>
+      <li>Syntax highlight for lot of languages.</li>
+      <li>Code tree: structure of functions/classes/etc, if lexer allows it.</li>
+      <li>Code folding.</li>
+      <li>Multi-carets and multi-selections.</li>
+      <li>Find/Replace with regular expressions.</li>
+      <li>Configs in JSON format. Including lexer-specific configs.</li>
+      <li>Tabbed UI.</li>
+      <li>Split view to primary/secondary. Split window to 2/3/4/6 groups of tabs.</li>
+      <li>Command palette, with fuzzy matching.</li>
+      <li>Minimap. Micromap.</li>
+      <li>Show unprinted whitespace.</li>
+      <li>Support for many encodings.</li>
+      <li>Customizable hotkeys.</li>
+      <li>Binary/Hex viewer for files of unlimited size (can show 10 Gb logs).</li>
+      <li>Correctly saves binary files.</li>
+    </ul>
+
+    <p>Features for HTML/CSS coding</p>
+    <ul>
+      <li>Smart auto-completion for HTML, CSS.</li>
+      <li>HTML tags completion with Tab-key (Snippets plugin).</li>
+      <li>HTML color codes (#rgb, #rrggbb) underline.</li>
+      <li>Show pictures inside editor area (jpeg/png/gif/bmp/ico).</li>
+      <li>Show tooltip when mouse moves over picture tag, entity, color value.</li>
+    </ul>
+
+    <p>Features implemented as plugins</p>
+    <ul>
+      <li>Addons manager</li>
+      <li>Find in files</li>
+      <li>Snippets</li>
+      <li>External tools</li>
+      <li>Project manager</li>
+      <li>Session manager</li>
+      <li>Macro manager</li>
+      <li>Spell checker</li>
+      <li>FTP panel</li>
+      <li>Linters support (CudaLint)</li>
+      <li>Brackets highlight, go to pair bracket</li>
+      <li>Highlight all occurences</li>
+      <li>Color picker</li>
+      <li>Insert date/time</li>
+      <li>Formatters for HTML/ CSS/ JS/ XML/ SQL</li>
+      <li>Tabs list in sidebar</li>
+      <li>Create backup files</li>
+      <li>Menu configurator</li>
+      <li>and more...</li>
+    </ul>
+
+  </description>
+
+  <url type="homepage">https://cudatext.github.io/</url>
+  <developer_name>Alexey Torgashin</developer_name>
+
+  <content_rating type="oars-1.1" />
+
+  <launchable type="desktop-id">io.github.cudatext.CudaText-Qt.desktop</launchable>
+
+  <screenshots>
+    <screenshot type="default">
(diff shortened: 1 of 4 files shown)


bbhtt:
It says to mark them with `X`

io.github.cudatext.CudaText-Qt.appdata.xml
bbhtt, line 1: Metainfo should be upstream

io.github.cudatext.CudaText-Qt.yml
```diff
@@ -0,0 +1,188 @@
...
+    FPC_DIR: /app/lib/sdk/freepascal/fpcsrc
+    LAZARUSDIR: /app/lib/sdk/freepascal/share/lazarus
+cleanup-commands:
+  - rm -rf /app/lib/sdk/freepascal
+default-branch: stable
+command: cudatext
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --filesystem=home
```
bbhtt, line 19: why do you need home?

JordanL2:
> It says to mark them with `X`

Fixed.

io.github.cudatext.CudaText-Qt.yml
```diff
@@ -0,0 +1,188 @@
...
+      - lazbuild --ws=qt6 -q --pcp=. --lazarusdir="$LAZARUSDIR" "./src/CudaText/app/cudatext.lpi"
+      - install -Dm755 src/CudaText/app/cudatext -t $FLATPAK_DEST/bin
+      # Install all data and plugins
+      - mkdir -p $FLATPAK_DEST/share/cudatext
+      - cp -r src/CudaText/app/data $FLATPAK_DEST/share/cudatext/
+      - cp -r src/CudaText/app/py $FLATPAK_DEST/share/cudatext/
+      - cp -r src/CudaText/app/settings_default $FLATPAK_DEST/share/cudatext/
+      # Desktop file and icon - desktop file needs to have the icon name corrected first (and carriage-returns removed)
+      - sed -i -e"s/\r//g" src/CudaText/setup/debfiles/cudatext.desktop
+      - sed -i -e"s/Icon=cudatext-512/Icon=io.github.cudatext.CudaText\-Qt/" src/CudaText/setup/debfiles/cudatext.desktop
```
bbhtt, line 185: Use desktop-file-edit here

io.github.cudatext.CudaText-Qt.yml
```diff
@@ -0,0 +1,188 @@
...
+      - sed -i -e"s/\/usr\/share\/cudatext/\\$FLATPAK_DEST\/share\/cudatext/g" src/CudaText/app/proc_globdata.pas
+      - lazbuild --ws=qt6 -q --pcp=. --lazarusdir="$LAZARUSDIR" "./src/CudaText/app/cudatext.lpi"
+      - install -Dm755 src/CudaText/app/cudatext -t $FLATPAK_DEST/bin
+      # Install all data and plugins
+      - mkdir -p $FLATPAK_DEST/share/cudatext
+      - cp -r src/CudaText/app/data $FLATPAK_DEST/share/cudatext/
+      - cp -r src/CudaText/app/py $FLATPAK_DEST/share/cudatext/
+      - cp -r src/CudaText/app/settings_default $FLATPAK_DEST/share/cudatext/
+      # Desktop file and icon - desktop file needs to have the icon name corrected first (and carriage-returns removed)
+      - sed -i -e"s/\r//g" src/CudaText/setup/debfiles/cudatext.desktop
```
bbhtt, line 184: This should be fixed upstream

io.github.cudatext.CudaText-Qt.yml
```diff
@@ -0,0 +1,188 @@
...
+build-options:
+  append-path: /app/lib/sdk/freepascal/bin
+  env:
+    PPC_CONFIG_PATH: /app/lib/sdk/freepascal/etc
+    FPCDIR: /app/lib/sdk/freepascal/fpcsrc
+    FPC_DIR: /app/lib/sdk/freepascal/fpcsrc
+    LAZARUSDIR: /app/lib/sdk/freepascal/share/lazarus
+cleanup-commands:
+  - rm -rf /app/lib/sdk/freepascal
+default-branch: stable
```
bbhtt, line 14:
```suggestion
```

io.github.cudatext.CudaText-Qt.yml
JordanL2: Being a general text editor, the user might use it to open any text (or html, or code source etc) files that belong to them, so it doesn't make much sense to restrict it further than this IMO, although I suppose `xdg-documents` might be enough for most users.

io.github.cudatext.CudaText-Qt.appdata.xml
JordanL2: I'm maintaining this flatpak on behalf of (but with the permission of) the original app developer, @Alexey-T. Since this metadata is only used for this flatpak, it makes more sense for it to be here where I can maintain it (ie change the version number for each release) rather than imposing the burden on Alexey.

io.github.cudatext.CudaText-Qt.yml
JordanL2: Fixed.

io.github.cudatext.CudaText-Qt.appdata.xml
hfiguiere, line 1: this still hasn't been submitted upstream?

It's not flatpak specific

io.github.cudatext.CudaText-Qt.yml
```diff
@@ -0,0 +1,186 @@
...
+      - find . -type f -iname makefile.fpc -exec fpcmake -Tall '{}' ';'
+      - make lazbuild OPT='-gl'
+      - touch lazarus startlazarus
+      - make install
+      - sed -i -e "s#__LAZARUSDIR__#$LAZARUSDIR/#" -e "s#__FPCSRCDIR__#$FPCDIR/#" -e "s#/app/bin/fpc#$INSTALL_PREFIX/bin/fpc#"
+        tools/install/linux/environmentoptions.xml
+      - install -Dm644 -t $LAZARUSDIR tools/install/linux/environmentoptions.xml
+
+  - name: qt6pas
+    buildsystem: simple
```
hfiguiere, line 97: should this be `buildsystem: qmake`?

JordanL2: I think it needs to be `simple` as we need the sed command to fix the target.path parameter.

io.github.cudatext.CudaText-Qt.appdata.xml
JordanL2: @Alexey-T would you be ok with adding this file upstream in CudaText?

https://github.com/flathub/flathub/blob/97e9e6c4d725e4c459abfaf0a8df6ce4b82b565d/io.github.cudatext.CudaText-Qt.appdata.xml

Apparently it would be useful for distribution packagers as well as flatpak. It just needs to be updated with the latest version number and date on each release.

bbhtt:
bot, build io.github.cudatext.CudaText-Qt

io.github.cudatext.CudaText-Qt.appdata.xml
Alexey-T: but it is redundant burden for me, and I must have knowledge what all lines mean. so no, sorry.

JordanL2:
I've installed and tested the test build. All looks good to me.

PR #5058, opened 2024-03-16
Add org.freedesktop.Sdk.Extension.openjfx

trixon:
This is an extension that provides jmods for javafx aimed to be used when creating custom runtime images with jlink.
Like in this one: https://github.com/trixon/flatpak-java-cri/blob/main/se.trixon.HelloFx.yml

- [ ] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:**

### DIFF
--- a/org.freedesktop.Sdk.Extension.openjfx.appdata.xml
+++ b/org.freedesktop.Sdk.Extension.openjfx.appdata.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2024 Patrik Karlström <patrik@trixon.se> -->
+<component type="runtime">
+    <id>org.freedesktop.Sdk.Extension.openjfx</id>
+    <metadata_license>CC0-1.0</metadata_license>
+    <project_license>GPL-2.0 AND GPL-2.0-with-classpath-exception</project_license>
+    <name>OpenJFX SDK Extension</name>
+    <summary>The latest version of the OpenJFX jmods</summary>
+    <description>
+        <p>
+            This SDK extension allows the building custom runtime images with jmods for JavaFX.
+        </p>
+    </description>
+    <url type="homepage">https://wiki.openjdk.org/display/OpenJFX/Main/</url>
+    <url type="bugtracker">https://bugs.openjdk.org/</url>
+    <developer_name>Patrik Karlström</developer_name>
+    <update_contact>patrik@trixon.se</update_contact>
+    <releases>
+        <release version="openjfx-21.0.2+1" date="2024-01-16"/>
+    </releases>
+</component>
--- a/org.freedesktop.Sdk.Extension.openjfx.yaml
+++ b/org.freedesktop.Sdk.Extension.openjfx.yaml
@@ -1,0 +1,43 @@
+id: org.freedesktop.Sdk.Extension.openjfx
+branch: '23.08'
+runtime: org.freedesktop.Sdk
+runtime-version: '23.08'
+build-extension: true
+sdk: org.freedesktop.Sdk
+separate-locales: false
+appstream-compose: false
+build-options:
+  no-debuginfo: true
+  prefix: /usr/lib/sdk/openjfx
+
+modules:
+  - name: javafx_jmods
+    buildsystem: simple
+    sources:
+      - type: file
+        only-arches:
+          - x86_64
+        url: https://download.bell-sw.com/java/21.0.2+14/bellsoft-jdk21.0.2+14-linux-amd64-full.tar.gz
+        dest-filename: java-openjfx.tar.gz
+        sha512: 1b349703a585198af2ae56e2ec4ba1922c794be444bdf8abc6252a64a8cfe5bf26714971668ef8481f8223a3ff2a4bb14a1ba0cfe965eea1c32791e8cdb205f1
+      - type: file
+        only-arches:
+          - aarch64
+        url: https://download.bell-sw.com/java/21.0.2+14/bellsoft-jdk21.0.2+14-linux-aarch64-full.tar.gz
+        dest-filename: java-openjfx.tar.gz
+        sha512: 043e798e02433c03779a95c9583881908a5a5e12c9c2200fd18da96a24e48103d503795b4a6c191b5cf74f4fcda2e6b04c25336799c15d59f61e6832b73bf400
+    build-commands:
+#      - tar xfv java-openjfx.tar.gz --wildcards "jdk-21.0.2-full/jmods/javafx.*.jmod" -C $FLATPAK_DEST --strip-components 1
+      - tar xfv java-openjfx.tar.gz --wildcards "jdk-21.0.2-full/jmods/javafx.*.jmod" --strip-components 1
+      - mv jmods $FLATPAK_DEST
+
+  - name: appdata
+    buildsystem: simple
+    sources:
+      - type: file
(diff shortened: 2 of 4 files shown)


hfiguiere:
I don't see a justification for this

trixon:
In my view this would change how one as a developer/packager adds java applications using the javafx framework to flathub to the better. Just add the extension and the module(s) you need.

I think it would be wrong to compare javafx with any other random java dependency since there are two major graphical toolkits for java, swing and javafx. Swing is, for now, included in the jdk, where as javafx is not.

Having this extension would make it _easier_ to create smaller and faster flatpaks.

But then again, I don't know how inclusion would affect flathub as a whole, or if there's some special feelings in general involved when it comes to java.

org.freedesktop.Sdk.Extension.openjfx.appdata.xml
```diff
@@ -0,0 +1,21 @@
...
+    <name>OpenJFX SDK Extension</name>
+    <summary>The latest version of the OpenJFX jmods</summary>
+    <description>
+        <p>
+            This SDK extension allows the building custom runtime images with jmods for JavaFX.
+        </p>
+    </description>
+    <url type="homepage">https://wiki.openjdk.org/display/OpenJFX/Main/</url>
+    <url type="bugtracker">https://bugs.openjdk.org/</url>
+    <developer_name>Patrik Karlström</developer_name>
```
bbhtt, line 16: Unless you authored the software, this should be whoever made it.

org.freedesktop.Sdk.Extension.openjfx.yaml
```diff
@@ -0,0 +1,43 @@
...
+appstream-compose: false
+build-options:
+  no-debuginfo: true
+  prefix: /usr/lib/sdk/openjfx
+
+modules:
+  - name: javafx_jmods
+    buildsystem: simple
+    sources:
+      - type: file
```
bbhtt, line 17:
```suggestion
      - type: archive
```

org.freedesktop.Sdk.Extension.openjfx.yaml
```diff
@@ -0,0 +1,43 @@
...
+        only-arches:
+          - x86_64
+        url: https://download.bell-sw.com/java/21.0.2+14/bellsoft-jdk21.0.2+14-linux-amd64-full.tar.gz
+        dest-filename: java-openjfx.tar.gz
+        sha512: 1b349703a585198af2ae56e2ec4ba1922c794be444bdf8abc6252a64a8cfe5bf26714971668ef8481f8223a3ff2a4bb14a1ba0cfe965eea1c32791e8cdb205f1
+      - type: file
```
bbhtt, line 23:
```suggestion
      - type: archive
```

org.freedesktop.Sdk.Extension.openjfx.yaml
```diff
@@ -0,0 +1,43 @@
...
+        only-arches:
+          - aarch64
+        url: https://download.bell-sw.com/java/21.0.2+14/bellsoft-jdk21.0.2+14-linux-aarch64-full.tar.gz
+        dest-filename: java-openjfx.tar.gz
+        sha512: 043e798e02433c03779a95c9583881908a5a5e12c9c2200fd18da96a24e48103d503795b4a6c191b5cf74f4fcda2e6b04c25336799c15d59f61e6832b73bf400
+    build-commands:
+#      - tar xfv java-openjfx.tar.gz --wildcards "jdk-21.0.2-full/jmods/javafx.*.jmod" -C $FLATPAK_DEST --strip-components 1
+      - tar xfv java-openjfx.tar.gz --wildcards "jdk-21.0.2-full/jmods/javafx.*.jmod" --strip-components 1
```
bbhtt, line 31: `type: archive` will auto extract it, you should remove this and adjust the commands

org.freedesktop.Sdk.Extension.openjfx.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - mv jmods $FLATPAK_DEST
+
```
bbhtt, line 33: I think you need some scripts like this to let apps using the extension, easily copy everything to app and set, $PATH etc. https://github.com/flathub/org.freedesktop.Sdk.Extension.openjdk21/blob/30eb0497b1f21d250b7ef63f7b829838baba14e3/org.freedesktop.Sdk.Extension.openjdk21.yaml#L172-L195

org.freedesktop.Sdk.Extension.openjfx.yaml
trixon: This is by design since I'm only interested in a few javafx.jmods but of course I can let archive extract everything and move the wanted files.

org.freedesktop.Sdk.Extension.openjfx.yaml
trixon: jmods are not meant to be used at run time, in fact they can not. So there is nothing to copy.

They are used at compile and link time. Like in this sample here:
https://github.com/trixon/flatpak-java-cri/blob/1bcfb63706da7e1ff8bc5abc814085765454e8ce/se.trixon.HelloFx.yml#L19C1-L29C26

The runtime for that demo is 60MB, compare that to the copying of jre+javafx that lands on about 280MB.

The recommended way to distribute java applications today is to provide a custom runtime image, not the full blown "jre".

org.freedesktop.Sdk.Extension.openjfx.yaml
bbhtt: Ok then it is fine

org.freedesktop.Sdk.Extension.openjfx.yaml
hfiguiere: notice how this is `Sdk` and not runtime? It's for **building**.

Also the prefered way on flathub is to ship the JDK that come with the corresponding SDK Extension. That allow supporting more than just x86_64.

hfiguiere: And also allow easier security updates.

bbhtt: > notice how this is Sdk and not runtime? It's for building.

SDK extensions are not only for building though, all the regular JDK extensions use it and still give the option of copying the contents because they are required at runtime too.

Node extensions also use SDK but can be both used at build and runtime (for the `node` executable)

Anyway if this doesn't require it, it's fine.

org.freedesktop.Sdk.Extension.openjfx.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - name: appdata
+    buildsystem: simple
+    sources:
+      - type: file
+        path: org.freedesktop.Sdk.Extension.openjfx.appdata.xml
+    build-commands:
+      - mkdir -p ${FLATPAK_DEST}/share/metainfo
+      - cp org.freedesktop.Sdk.Extension.openjfx.appdata.xml ${FLATPAK_DEST}/share/metainfo
+      - appstream-compose --basename=org.freedesktop.Sdk.Extension.openjfx --prefix=${FLATPAK_DEST}
```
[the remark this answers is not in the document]
bbhtt: You can remove appstream-compose: false from the top and let flatpak builder take over. Nowadays it should work ootb with extensions

trixon:
Thank you for the comments, I will address them over the weekend.

PR #5078, opened 2024-03-22
Add io.github.eteran.edb-debugger

bztd:

A debugger
- [x] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about submitting their software to Flathub. https://github.com/eteran/edb-debugger/issues/808


bztd:
bot, build io.github.eteran.edb-debugger

bztd:
bot, build io.github.eteran.edb-debugger

io.github.eteran.edb-debugger.yml
```diff
@@ -0,0 +1,78 @@
...
+  - --socket=x11
+  - --filesystem=xdg-documents
+  - --device=dri
+
+modules:
+  - name: capstone
+    buildsystem: simple
+    build-commands:
+      - make V=1 PREFIX=/app/
+      - make DESTDIR=/app install
```
hfiguiere, line 21: `DESTDIR` doesn't do what you think it does. This should be `PREFIX` and then you can likely remove the following lines, including the `post-install`

io.github.eteran.edb-debugger.yml
```diff
@@ -0,0 +1,78 @@
...
+    sources:
+      - type: archive
+        url: https://gitlab.com/api/v4/projects/4207231/packages/generic/graphviz-releases/10.0.1/graphviz-10.0.1.tar.gz
+        sha256: 72e1a575e9c382c4fd2267837962bc4582b3d9ad746e338cd8024bdeca124f32
+
+  - name: edb
+    buildsystem: cmake-ninja
+    post-install:
+      #xterm path
+      - sed -i "s/\x2f\x75\x73\x72\x2f\x62\x69\x6e\x2f\x78\x74\x65\x72\x6d/\x78\x74\x65\x72\x6d\x00\x00\x00\x00\x00\x00\x00\x00\x00/g" /app/bin/edb
```
hfiguiere, line 68: DO NOT BINARY patch like that. Patch the source code.

bztd: I would prefer to leave the line because it is easier to maintain. If you have to patch, I think look, they wanted two files. It is quite unlikely that the line will end in the future causing a failure because it is very specific. (this is a translation)

io.github.eteran.edb-debugger.yml
bztd: the usr directory becomes empty and causes an error in my local tests.

io.github.eteran.edb-debugger.yml
hfiguiere: no it is not easier to maintain. It is completely obscure and it will break randomly as sed failure mode is inexistant.

io.github.eteran.edb-debugger.yml
hfiguiere: because there is no `/usr` directory as the prefix is `/app`.

edb.appdata.xml
bbhtt, line 1: can you submit this upstream?

bztd: Yes, at the end I make the request, in case there are changes.

io.github.eteran.edb-debugger.yml
bztd: without DESTDIR the installation is attempted in /usr which causes an error.

hfiguiere: I will repeat one more time: it's because you need to set prefix instead.

bztd:
bot, build io.github.eteran.edb-debugger

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+    "only-arches": ["x86_64"]
```
bbhtt, line 2: Why did it become x86_64 only?

bztd: no support for arm64

edb.appdata.xml
hfiguiere, line 1: this file is upstream. Why adding it here?

https://github.com/eteran/edb-debugger/blob/master/edb.appdata.xml

flathub.json
hfiguiere: thats not what I'm seeing:

image

Please remove the restriction.

edb.appdata.xml
bztd: If I'm not wrong, it was missing fields and is outdated.

flathub.json
bztd: I don't think it makes sense to do that.

According to the author's words:
"Really x86/x86-64 on Linux is really the only WELL supported platform."
https://github.com/eteran/edb-debugger/issues/808

Also, if the program compiles I cannot test the result.

hfiguiere: but if you don't build it no one can test it either. THere is explicit support for aarch64.

bztd: In fact the author says no. and it also says it in the code, it is possible that the software does not work well. I asked the developer, I'm waiting for the answer.

edb.appdata.xml
bztd: I'll do it then.

flathub.json
hfiguiere: link?

bztd: https://github.com/eteran/edb-debugger/issues/808

bztd:
bot, build io.github.eteran.edb-debugger

flathub.json
eteran: Author here 🙂

The only platform that edb can currently be expected to work in any meaningful way is x86-64 on Linux.

Due to the nature of the project, it has a lot of highly OS and architecture specific code and arm64 has never been in a state that can build, let alone work (so far!). We hope to improve this at some point, but it's challenging given that doing so involves buying new hardware, learning new arch specific details, and a lot of testing.

bbhtt:
would be nice if you can find a way to submit these patches upstream, xterm shouldn't be hardcoded to `/usr/bin` it should just follow `$PATH`.

bbhtt:
> @barthalion @hfiguiere Could you give permissions to @eteran?

Added

PR #5084, opened 2024-03-25
Add com.cerebralnomad.recipescribe

cerebralnomad:

- [ X] Please describe your application briefly.
Recipe Scribe is  a utility to record recipes, and store them as txt files with automatic formatting
Includes search and editing ability limited to only files inside your designated recipe folder and subfolders.

- [ X] I have read the [App Requirements][reqs] and [App Maintenance][maint] pages.
- [ X] My pull request follows the instructions at [App Submission][submission].
- [ X] I have [built][build] and tested the submission locally.
- [ X] I am using only the minimal set of permissions. *(If not, please explain each non-standard permission.)*
The only additional permission is to r/w in the home directory so the user can choose where
the files are saved.

- [ X] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:**
- [ X] The domain used for the application ID is controlled by the application developers either directly or through the code hosting (e.g. GitHub, GitLab, SourceForge, etc.). The [application id guidelines][app-id] are followed.
- [ X] Any additional patches or files have been submitted to the upstream projects concerned. *(If not, explain why.)*

### DIFF
--- a/com.cerebralnomad.recipescribe.yml
+++ b/com.cerebralnomad.recipescribe.yml
@@ -3,6 +3,13 @@ runtime: org.freedesktop.Platform
 runtime-version: '23.08'
 sdk: org.freedesktop.Sdk
 command: recipe_scribe-2.0.1
+
+finish-args:
+  - --filesystem=home:rw
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --share=ipc 
+
 modules:
   - name: recipescribe
     buildsystem: simple
@@ -17,8 +24,3 @@ modules:
         url: https://github.com/cerebralnomad/Recipe-Scribe/releases/download/v2.0.1-stable/recipe_scribe_2.0.1_flatpak_build_files.tar.gz
         sha256: 2b5ad38414a74f204f89b3006e83b4a7139f68cd3351a3851e1ccfc8b08683ed
 
-finish-args:
-  - --filesystem=home:rw
-  - --socket=fallback-x11
-  - --socket=wayland
-  - --share=ipc 
--- a/com.cerebralnomad.recipescribe.yml
+++ b/com.cerebralnomad.recipescribe.yml
@@ -6,8 +6,7 @@ command: recipe_scribe-2.0.1.py
 
 finish-args:
   - --filesystem=home:rw
-  - --socket=fallback-x11
-  - --socket=wayland
+  - --socket=x11
   - --share=ipc 
 
 modules:
--- a/com.cerebralnomad.recipescribe.yml
+++ b/com.cerebralnomad.recipescribe.yml
@@ -73,5 +73,5 @@ modules:
     sources:
       - type: archive
         url: https://github.com/cerebralnomad/Recipe-Scribe/releases/download/v2.0.1-stable/recipe_scribe_2.0.1_flatpak_source_files.tar.gz
-        sha256: 33d98dbf6b0b22cffda773d0dc01d6fa7f96e79109c1a7677be707bbd6edc724 
+        sha256: 5c7cfd04ae31432802aececd8569462a8d911cc9d6b63905ce1c5609591587b6
 
--- a/com.cerebralnomad.recipescribe.yml
+++ b/com.cerebralnomad.recipescribe.yml
@@ -7,7 +7,6 @@ modules:
   - name: recipescribe
     buildsystem: simple
     build-commands:
-      - rm README.txt
       - install -D recipe_scribe-2.0.1 /app/bin/recipe_scribe-2.0.1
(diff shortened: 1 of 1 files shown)


com.cerebralnomad.recipescribe.yml
```diff
@@ -0,0 +1,25 @@
+id: com.cerebralnomad.recipescribe
+runtime: org.freedesktop.Platform
+runtime-version: '23.08'
+sdk: org.freedesktop.Sdk
+command: recipe_scribe-2.0.1
+modules:
+  - name: recipescribe
+    buildsystem: simple
+    build-commands:
+      - rm README.txt
```
hfiguiere, line 10:
```suggestion
```

com.cerebralnomad.recipescribe.yml
```diff
@@ -0,0 +1,25 @@
...
+      - install -D recipe_scribe-2.0.1 /app/bin/recipe_scribe-2.0.1
```
hfiguiere, line 11: You install a binary, so how does this work on both x86_64 and aarch64?

com.cerebralnomad.recipescribe.yml
```diff
@@ -0,0 +1,25 @@
...
+      - install -D com.cerebralnomad.recipescribe.metainfo.xml /app/share/metainfo/com.cerebralnomad.recipescribe.metainfo.xml
+      - install -D com.cerebralnomad.recipescribe.desktop /app/share/applications/com.cerebralnomad.recipescribe.desktop
+      - install -D rc_128.png /app/share/icons/hicolor/128x128/apps/com.cerebralnomad.recipescribe.png
+      - install -D rc_256.png /app/share/icons/hicolor/256x256/apps/com.cerebralnomad.recipescribe.png
+    sources:
+      - type: archive
+        url: https://github.com/cerebralnomad/Recipe-Scribe/releases/download/v2.0.1-stable/recipe_scribe_2.0.1_flatpak_build_files.tar.gz
+        sha256: 2b5ad38414a74f204f89b3006e83b4a7139f68cd3351a3851e1ccfc8b08683ed
+
+finish-args:
```
hfiguiere, line 21: `finish-args` should be above `modules`.

com.cerebralnomad.recipescribe.yml
cerebralnomad: I didn't know if that was needed or not. The tarball is in my releases and i only included the readme in case it was downloaded by mistake instead o fthe proper tarball of the standalone program.
I'll remove that line

com.cerebralnomad.recipescribe.yml
cerebralnomad: Would I need something like

build-args:
   -  --arch=x86_64

bbhtt: https://docs.flathub.org/docs/for-app-authors/maintenance#only-build-on-x86_64

hfiguiere: how about building from source instead?

cerebralnomad: better question, if I started over and built from the source files instead of the binary, would the resulting flatpak run on either architecture?

cerebralnomad: I reworked it all to install from the source files instead. Had to pull in Tkinter as well. I used the binary initially to avoid that, I hope I did it correctly. I borrowed some code from another flatpak yaml for that part.
Thanks for your patience, this is my first try and I'm learning as I go.

bbhtt:
bot, build com.cerebralnomad.recipescribe

cerebralnomad:
bot, build com.cerebralnomad.recipescribe

com.cerebralnomad.recipescribe.yml
```diff
@@ -0,0 +1,77 @@
+id: com.cerebralnomad.recipescribe
+runtime: org.freedesktop.Platform
+runtime-version: '23.08'
+sdk: org.freedesktop.Sdk
+command: recipe_scribe-2.0.1.py
+
+finish-args:
+  - --filesystem=home:rw
+  - --socket=fallback-x11
+  - --socket=wayland
```
hfiguiere, line 10: are you sure this work with wayland?

cerebralnomad: The binary runs fine on wayland, but I just tested it and the flatpak does not. It didn't occur to me that it might be a problem. 
I'll have to dig into that and see what the issue is.
In the meantime I guess I'll have to use socket=x11 instead

bbhtt:
bot, build com.cerebralnomad.recipescribe

cerebralnomad:
bot, build com.cerebralnomad.recipescribe

cerebralnomad:
I realized that after changing to build from source the .desktop file was still calling the binary. 
I changed it to call the proper source file so the launcher in the menu works properly now.

PR #5138, opened 2024-04-06
add io.github.nyre221.kiview

Nyre221:
Kiview allows to quickly preview different file types in Dolphin without having to open the default application.

>  I am using only the minimal set of permissions. *(If not, please explain each non-standard permission.)*

Since my application needs to integrate with dolphin and have the ability to execute commands on the host system for `unrar` and `libreofffice` (optional) I had to add these two permissions:

```
      "--socket=session-bus",
      "--talk-name=org.freedesktop.Flatpak",
```
I tried not to use `--socket=session-bus` but unfortunately I didn't find a way to get the list of all available dolphin DBus interfaces without using it. My application uses that list to control which dolphin window is the active one.

`org.freedesktop.DBus /org/freedesktop/DBus org.freedesktop.DBus.ListNames
`

> --talk-name=org.freedesktop.Flatpak

Used to run `unrar` and `libreoffice` (optional)

If you want, since the application requires configuring a shortcut, I could remove the permissions from the .json and write them in the configuration guide.
For users launching the application without a shortcut, a message will appear with a link to the GitHub configuration page. This ensures the application remains functional even without the shortcut.


Nyre221:
bot, build io.github.nyre221.kiview

io.github.nyre221.kiview.json
```diff
@@ -0,0 +1,45 @@
...
+      "--socket=session-bus",
+      "--filesystem=host:ro",
+      "--filesystem=/tmp",
+      "--talk-name=org.freedesktop.Flatpak",
+      "--talk-name=org.kde.klipper",
+      "--talk-name=org.freedesktop.DBus",
+      "--talk-name=org.kde.dolphin"
+   ],
+   "cleanup-commands":[ "/app/cleanup-BaseApp.sh"],
+   "id": "io.github.nyre221.kiview",
```
hfiguiere, line 27: id should be the first thing.

io.github.nyre221.kiview.json
```diff
@@ -0,0 +1,45 @@
...
+               "url": "https://github.com/Nyre221/Kiview.git",
+               "commit":"fdc79c3a64f15005d6dca8bf9406c917a5cfa504"
+            }
+         ]
+      }
+
+   ],
+   "runtime": "org.kde.Platform",
+    "runtime-version": "6.6",
+   "sdk": "org.kde.Sdk"
```
hfiguiere, line 44: these three lines should also be toward the top. and please fix the indentation.

io.github.nyre221.kiview.json
```diff
@@ -0,0 +1,45 @@
...
+  },
+
+   "finish-args": [
+      "--share=ipc",
+      "--device=dri",
+      "--socket=fallback-x11",
+      "--socket=wayland",
+      "--env=QML_IMPORT_PATH=/app/qml" ,
+      "--socket=pulseaudio",
+      "--socket=session-bus",
```
hfiguiere, line 18: `session-bus` is not allowed nor did you justify.

io.github.nyre221.kiview.json
```diff
@@ -0,0 +1,45 @@
...
+      "--filesystem=host:ro",
+      "--filesystem=/tmp",
+      "--talk-name=org.freedesktop.Flatpak",
```
hfiguiere, line 21: this needs an exception.

io.github.nyre221.kiview.json
```diff
@@ -0,0 +1,45 @@
...
+      "--talk-name=org.kde.klipper",
+      "--talk-name=org.freedesktop.DBus",
```
hfiguiere, line 23: why do you need `org.freedesktop.DBus`.

hfiguiere:
> Used to run unrar and libreoffice (optional)

So what if the user on has libreoffice as a flatpak?

io.github.nyre221.kiview.json
Nyre221: It is used to obtain the list of all Dolphin windows and the name of their interface.
Example in bash:
`$dbus-send --session --print-reply --type=method_call --dest=org.freedesktop.DBus /org/freedesktop/DBus org.freedesktop.DBus.ListNames | grep -i dolphin `
`$  string "org.kde.dolphin-13165"`

I couldn't find another way to access `ListNames`

My application needs it to interact with dolphin and send signals to the top window.

As I said before, I can remove it from the .json file and write it in the guide to set up the shortcut.

io.github.nyre221.kiview.json
Nyre221: what do you mean?

io.github.nyre221.kiview.json
Nyre221: It's there from a previous attempt to not use `--socket=session-bus`
I can remove it

Nyre221:
> So what if the user only has libreoffice as a flatpak?

A message will appear with a link to the additional configuration needed for libreoffice (access to /tmp).

io.github.nyre221.kiview.json
Nyre221: I'll fix it soon

io.github.nyre221.kiview.json
Nyre221: I'll fix it soon

io.github.nyre221.kiview.json
Nyre221: removed

io.github.nyre221.kiview.json
Nyre221: fixed

io.github.nyre221.kiview.json
Nyre221: fixed

io.github.nyre221.kiview.json
bbhtt: https://docs.flathub.org/docs/for-app-authors/linter#exceptions

io.github.nyre221.kiview.json
bbhtt: Yes listnames is filtered inside the sandbox so that apps cannot know about applications running outside https://github.com/flatpak/xdg-dbus-proxy/blob/7466c8137fc06f863fde8486521984e43a26cd10/flatpak-proxy.c#L2032

bbhtt: If this stays it needs an exception too https://docs.flathub.org/docs/for-app-authors/linter#finish-args-arbitrary-dbus-access

io.github.nyre221.kiview.json
```diff
@@ -0,0 +1,44 @@
...
+   "sdk": "org.kde.Sdk",
+   "base": "io.qt.qtwebengine.BaseApp",
+   "base-version": "6.6",
+   "command": "kiview",
+   "build-options": {
+   "env": [
+      "QTWEBENGINE_DISABLE_SANDBOX=1"
+    ]
+  },
+
```
bbhtt, line 14: Why is sandbox disabled? It works fine for every app on Flathub.

bbhtt, line 14: Missing cleanup https://github.com/flathub/io.qt.qtwebengine.BaseApp?tab=readme-ov-file#cleanup

io.github.nyre221.kiview.json
bbhtt: What commands from libreoffice do you need? unrar can be easily included in the flatpak here.

io.github.nyre221.kiview.json
```diff
@@ -0,0 +1,44 @@
...
+   "finish-args": [
+      "--share=ipc",
+      "--device=dri",
+      "--socket=fallback-x11",
+      "--socket=wayland",
+      "--env=QML_IMPORT_PATH=/app/qml" ,
+      "--socket=pulseaudio",
+      "--socket=session-bus",
+      "--filesystem=host:ro",
+      "--filesystem=/tmp",
```
bbhtt, line 24: Why is this necessary?

io.github.nyre221.kiview.json
bbhtt: > It is used to obtain the list of all Dolphin windows and the name of their interface.

Also why exactly do you want this information? What's the use-case here?

io.github.nyre221.kiview.json
Nyre221: libreoffice --headless --nolockcheck --norestore --convert-to pdf file.[doc,ppt,etc,etc] --outdir /tmp

And I need to kill the process with `kill` if the user switches to another file.

How do I include unrar?

io.github.nyre221.kiview.json
Nyre221: https://github.com/flathub/io.qt.qtwebengine.BaseApp

>
> During packaging of a Flatpak application there is no session bus socket in the sandbox, so if the build process needs running QtWebEngine, then it will fail.
> The workaround is to disable sandboxing like this:
>
> build-options:
>   env:
>     QTWEBENGINE_DISABLE_SANDBOX: '1'

I can remove it if it's not needed.

bbhtt: No that's some old thing, it should work. I'll remove it.

io.github.nyre221.kiview.json
Nyre221: --socket=pulseaudio: to play audio/video files
--socket=session-bus: 

> It is used to obtain the list of all Dolphin windows and the name of their interface.
> Example in bash:
> $dbus-send --session --print-reply --type=method_call --dest=org.freedesktop.DBus /org/freedesktop/DBus org.freedesktop.DBus.ListNames | grep -i dolphin 
> $  string "org.kde.dolphin-13165"

Kiview needs to know which window is active and the name of the interface to send commands to.

--filesystem=host:ro: Must be able to access files and folders to preview them.
--filesystem=/tmp: /tmp is the folder where unrar/unzip and libreoffice put the unzipped/converted file. 
The reason why I put that permission is because from inside flatpak I don't have access to libreoffice and unrar.

example with the python version of Kiview: https://www.youtube.com/watch?v=mCfPZ5dgIZs

io.github.nyre221.kiview.json
Nyre221: you mean the open source version of unrar, right? I remember trying to use that version and it fails to open some .rar archives.

bbhtt: > How do I include unrar?

https://github.com/search?q=org%3Aflathub%20unrar&type=code

> kill if the user switches to another file.

`kill` is available inside the sandbox.

> libreoffice --headless --nolockcheck --norestore --convert-to pdf file.[doc,ppt,etc,etc] --outdir /tmp

Libreoffice provides some prebuild binary packages, can you try to see if including them in the flatpak works? https://www.libreoffice.org/download/appimage/ or the deb/rpm https://www.libreoffice.org/download/download-libreoffice/

This is not very ideal but still better than having a gaping hole.

io.github.nyre221.kiview.json
Nyre221: Kiview needs to send commands to the active dolphin window to know which file the user wants to view (if there is a file selected), or to deduce which folder the shortcut was invoked in (so it can display a preview of the folder).

Example with the python version of kiview (QuickView): https://www.youtube.com/watch?v=mCfPZ5dgIZs

io.github.nyre221.kiview.json
bbhtt: And if it is included in Flatpak, it works ootb for people.

You can try setting `TMPDIR="$XDG_RUNTIME_DIR/app/$FLATPAK_ID"` or to `/var/tmp` to drop the `filesystem=/tmp` access.

io.github.nyre221.kiview.json
bbhtt: Also I don't see the app needing a running Qtwebengine process when building, you can drop it.

Nyre221: removed

io.github.nyre221.kiview.json
bbhtt: Let's talk about `/tmp` in https://github.com/flathub/flathub/pull/5138#discussion_r1554638580 thread, it'd be only necessary if you are using flatpak-spawn to launch libreoffice from host.

Flatpak sandbox has its own private tmp.

io.github.nyre221.kiview.json
Nyre221: I'll test it tomorrow and let you know.

I'll probably remove support for  .rar files if the version to include is the open source one that doesn't always work.

Thanks for your help.

Nyre221: I managed to use /var/tmp:ro and removed /tmp, unfortunately I still haven't managed to "completely" start libreoffice in the sandbox (Appimage). Everything works except libreoffice-impress which gives me the error:

Schermata_20240407_221344

I can try with the .deb but I don't think it will change anything.

In any case I no longer need access to /tmp

I was thinking, is there any way to extend another application similarly to what I did with WebEngine?
I tried copying most of the libreoffice .json and couldn't get it to install.

I also included unrar.

Nyre221: @bbhtt 
Even installing openjdk didn't fix the error. 
I don't know what else to try.

bbhtt: I'll check later.

Nyre221: @bbhtt 
I think I found the problem:
https://bugs.documentfoundation.org/show_bug.cgi?id=119408
I'll see what I can do tomorrow.

Nyre221: I think I fixed the error.
If everything now works as I think, the only problematic permission is session-bus.

bbhtt: We can make an exception for session bus since there is no other way, please make a PR for the session-bus exception. https://docs.flathub.org/docs/for-app-authors/linter/#finish-args-arbitrary-dbus-access https://docs.flathub.org/docs/for-app-authors/linter#exceptions

Nyre221:
bot, build io.github.nyre221.kiview

io.github.nyre221.kiview.json
bbhtt: It's because the name ends with `-{id}`, the `*` wildcard only works when ending with `.*`, it's not a generic wildcard that expands everywhere.

The best you can do is `org.kde.*` probably, that's somewhat better than having a session-bus permission

You'll still need to make an exception request as that is blocked too
https://docs.flathub.org/docs/for-app-authors/linter#finish-args-wildcard-kde-talk-name

Nyre221:
bot, build io.github.nyre221.kiview

bbhtt:
You need an exception for this, see https://github.com/flathub/flathub/pull/5138#discussion_r1556173586

Nyre221:
@bbhtt 
> You need an exception for this

I was checking if I had configured the flathub.json file correctly. I don't understand why it keeps trying to build for aarch64 if it's disabled.

https://github.com/Nyre221/Kiview/blob/master/flathub.json

io.github.nyre221.kiview.json
```diff
@@ -0,0 +1,108 @@
...
+            "chmod +x ./LibreOffice*AppImage",
+            "./LibreOffice*AppImage --appimage-extract",
+            "mv ./squashfs-root /app/bin/libreoffice-dir",
+            "ln -s /app/bin/libreoffice-dir/AppRun /app/bin/libreoffice"
+         ],
+         "sources": [
+            {
+               "type": "file",
+               "url": "https://appimages.libreitalia.org/LibreOffice-24.2.0.3.basic-x86_64.AppImage",
+               "sha256": "c8ca641329bcfff31dc16f0dc945728afeb6feb0425bb4096bcac2759413afc4"
```
bbhtt, line 103: Add a `"dest-filename": "LibreOffice.AppImage"` here, then drop the `*` from the build commands above.

bbhtt:
> I was checking if I had configured the flathub.json file correctly. I don't understand why it keeps trying to build for aarch64 if it's disabled.

You haven't pushed flathub.json here

io.github.nyre221.kiview.json
bbhtt: Also a `"only-arches": ["x86_64"]` below

Nyre221:
> You haven't pushed flathub.json here

ah, that makes sense.

I added `dest-filename` and removed `*`, do you want me to update the commint to the io.github.nyre221.kiview.json file?

bbhtt:
Yes and push the flathub.json too

io.github.nyre221.kiview.json
```diff
@@ -0,0 +1,109 @@
...
+            "./LibreOffice.AppImage --appimage-extract",
+            "mv ./squashfs-root /app/bin/libreoffice-dir",
+            "ln -s /app/bin/libreoffice-dir/AppRun /app/bin/libreoffice"
+         ],
+         "sources": [
+            {
+               "type": "file",
+               "url": "https://appimages.libreitalia.org/LibreOffice-24.2.0.3.basic-x86_64.AppImage",
+               "sha256": "c8ca641329bcfff31dc16f0dc945728afeb6feb0425bb4096bcac2759413afc4",
+               "dest-filename": "LibreOffice.AppImage"
```
bbhtt, line 104:
```suggestion
               "dest-filename": "LibreOffice.AppImage",
               "only-arches": ["x86_64"]
```

Here too

io.github.nyre221.kiview.json
```diff
@@ -0,0 +1,109 @@
...
+               "url": "https://github.com/avahi/avahi/releases/download/v0.8/avahi-0.8.tar.gz",
+               "sha256": "060309d7a333d38d951bc27598c677af1796934dbd98e1024e7ad8de798fedda"
+            }
+         ]
+      },
+      {
+         "name": "libreoffice",
+         "buildsystem": "simple",
+         "build-commands": [
+            "chmod +x ./LibreOffice*AppImage",
```
bbhtt, line 94:
```suggestion
            "chmod +x ./LibreOffice.AppImage",
```

bbhtt:
bot, build io.github.nyre221.kiview

Nyre221:
Failed:
> {
>     "errors": [
>         "appstream-failed-validation"
>     ],
>     "warnings": [
>         "appstream-summary-too-long",
>         "appstream-screenshot-missing-caption"
>     ],
>     "appstream": [
>         "E: io.github.nyre221.kiview:43: tag-duplicated provides"
>     ],
>     "message": "Please consult the documentation at https://docs.flathub.org/docs/for-app-authors/linter"
> }

I'm fixing them.

Nyre221:
bot, build io.github.nyre221.kiview

Nyre221:
If there are other errors I will fix them tomorrow.

Nyre221:
> To test this build, install it from the testing repository:

I tested it and it works as it should.

Nyre221:
bot, build io.github.nyre221.kiview

PR #5161, opened 2024-04-14
Add net.cachapa.tudo

cachapa:

Description: tudo is an experiment on simple, private, synchronized to-do lists.

### DIFF
--- a/net.cachapa.tudo.yaml
+++ b/net.cachapa.tudo.yaml
@@ -1,0 +1,24 @@
+id: net.cachapa.tudo
+runtime: org.freedesktop.Platform
+runtime-version: '23.08'
+sdk: org.freedesktop.Sdk
+command: tudo
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+#  - --socket=pulseaudio
+  - --share=network
+modules:
+  - name: tudo
+    buildsystem: simple
+    only-arches:
+      - x86_64
+    build-commands:
+      - ./build-flatpak.sh
+    sources:
+      - type: dir
+        path: ../../build/linux/x64/release/bundle/
+      - type: dir
+        path: assets


net.cachapa.tudo.yaml
```diff
@@ -0,0 +1,24 @@
...
+    buildsystem: simple
+    only-arches:
+      - x86_64
+    build-commands:
+      - ./build-flatpak.sh
+    sources:
+      - type: dir
+        path: ../../build/linux/x64/release/bundle/
+      - type: dir
+        path: assets
```
bbhtt, line 24: SO how does this work? You should not use dir sources, it should be a direct link to the source code/tarball/binary etc.

net.cachapa.tudo.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+#  - --socket=pulseaudio
+  - --share=network
+modules:
+  - name: tudo
+    buildsystem: simple
+    only-arches:
+      - x86_64
```
hfiguiere, line 17: also what happens with the aarch64 flatpak?

net.cachapa.tudo.yaml
cachapa: To be honest this is my first time generating a Flatpak so I'm not sure if my approach is correct.

In the first case, the directory contains the executable + precompiled libs required for the application to run.

In the second the assets directory contains metadata for the Flatpak: build script, icon, .desktop, etc.

net.cachapa.tudo.yaml
cachapa: The application only supports x86_64 at the moment.

My goal is to add aarch64 later once the x86_64 is published and running.

net.cachapa.tudo.yaml
bbhtt: > the directory contains the executable

And how is said directory found? We cannot go and grab it from your PC nor we can ship your PC.

It has to be a link.

PR #5174, opened 2024-04-20
Add com.opera.opera.codecs

Gaeldrin:
An extension to Opera flatpak with proprietary codecs.

As for the extra-data part, I couldn't find any way to unzip archive downloaded via extra-data. I also have no idea if this is correct way/place to submit an extension so I ask for advice if this is not the way. O:)

### DIFF
--- a/com.opera.Opera.Codecs.yaml
+++ b/com.opera.Opera.Codecs.yaml
@@ -1,0 +1,30 @@
+id: com.opera.Opera.Codecs
+  #branch: 'nwjs-ffmpeg-0.86.0'
+runtime: com.opera.Opera
+sdk: org.freedesktop.Sdk//23.08
+build-extension: true
+separate-locales: false
+
+modules:
+  - name: codecs
+    buildsystem: simple
+    build-commands:
+      - unzip ffmpeg.zip
+      - install -Dm 644 libffmpeg.so -t ${FLATPAK_DEST}
+    sources:
+      - type: file
+        url: https://github.com/Ld-Hagen/nwjs-ffmpeg-prebuilt/releases/download/nwjs-ffmpeg-0.86.0/0.86.0-linux-x64.zip
+        sha256: 5a1cdf7d45f580f6a30b677e8a402453cba4856bcadd5f946550857d1c04497c
+        dest-filename: ffmpeg.zip
+        x-checker-data:
+          type: rotating-url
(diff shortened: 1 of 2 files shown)


com.opera.Opera.Codecs.yaml
```diff
@@ -0,0 +1,30 @@
+id: com.opera.Opera.Codecs
+  #branch: 'nwjs-ffmpeg-0.86.0'
```
hfiguiere, line 2:
```suggestion
```

com.opera.Opera.Codecs.yaml
```diff
@@ -0,0 +1,22 @@
+id: com.opera.Opera.Codecs
```
hfiguiere, line 1: this extension point doesn't exist in the opera flatpak

com.opera.Opera.Codecs.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+modules:
+  - name: codecs
+    buildsystem: simple
+    build-commands:
+      - unzip ffmpeg.zip
+      - install -Dm 644 libffmpeg.so -t ${FLATPAK_DEST}
+    sources:
+      - type: file
+        url: https://github.com/Ld-Hagen/nwjs-ffmpeg-prebuilt/releases/download/nwjs-ffmpeg-0.86.0/0.86.0-linux-x64.zip
```
hfiguiere, line 15: what happen when the build bot tries to build the aarch64 package?

com.opera.Opera.Codecs.yaml
```diff
@@ -0,0 +1,22 @@
...
+runtime: com.opera.Opera
+sdk: org.freedesktop.Sdk//23.08
+build-extension: true
+separate-locales: false
+
+modules:
+  - name: codecs
+    buildsystem: simple
+    build-commands:
+      - unzip ffmpeg.zip
```
hfiguiere, line 11: `type: archive` should handle that.

com.opera.Opera.Codecs.yaml
hfiguiere: also what is this random ffmpeg binary? Why not build from source instead?

hfiguiere:
IMHO this should be built part of the opera flaptak instead.

hfiguiere:
> I don't see the reason to burn computation time and energy on repeating the same process that somebody already did and keeps doing constantly for a few years.

This is irrelevant. There are plenty of other reasons to build from source.

Gaeldrin:
Well, there is not much to work with. Can you share some of the other reasons?

PR #5204, opened 2024-04-30
Add io.github.pieterdd.RcloneShuttle

pieterdd:

- [x] Please describe your application briefly. - **It's a GTK4 GUI frontend for Rclone.**
- [x] I am using only the minimal set of permissions. *(If not, please explain each non-standard permission.)* - **The application accesses the default Rclone config on the host and the locally installed `rclone` binary. I looked into including rclone from it's release builds but wasn't sure how to do it properly in an architecturally neutral way.**
- [x] All assets referenced in the manifest are redistributable by any party.  If not, the unredistributable parts are using an extra-data source type. - **I checked every Rust crate in Cargo.toml; they're all MIT-licensed.**


io.github.pieterdd.RcloneShuttle.yaml
```diff
@@ -0,0 +1,39 @@
+id: io.github.pieterdd.RcloneShuttle
+runtime: org.gnome.Platform
+runtime-version: '45'
```
hfiguiere, line 3: 46 is available.

io.github.pieterdd.RcloneShuttle.yaml
```diff
@@ -0,0 +1,39 @@
...
+build-options:
+  append-path: "/usr/lib/sdk/rust-stable/bin"
+  env:
+    CARGO_HOME: "/run/build/rust-flatpak/cargo"
+modules:
+  - name: rust-flatpak
+    buildsystem: simple
+    build-commands:
+    - /usr/lib/sdk/rust-stable/bin/cargo --offline fetch --manifest-path Cargo.toml --verbose
+    - /usr/lib/sdk/rust-stable/bin/cargo --offline build --release --verbose
```
hfiguiere, line 26:
```suggestion
    - cargo --offline fetch --manifest-path Cargo.toml --verbose
    - cargo --offline build --release --verbose
```

io.github.pieterdd.RcloneShuttle.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - mkdir -p /app/bin
+    - install -Dm755 ./target/release/rclone-shuttle /app/bin/rclone-shuttle
+    - ln -s /run/host/usr/bin/rclone /app/bin/rclone
```
hfiguiere, line 29: no. you should build rclone here.

io.github.pieterdd.RcloneShuttle.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - install -Dm644 -t /app/share/icons/hicolor/512x512/apps/ meta/io.github.pieterdd.RcloneShuttle.png 
+    - install -Dm644 -t /app/share/applications/ meta/io.github.pieterdd.RcloneShuttle.desktop
+    - install -Dm644 -t /app/share/metainfo/ meta/io.github.pieterdd.RcloneShuttle.metainfo.xml
+    sources:
+    - type: dir
+      path: .
```
hfiguiere, line 35: and how to you think it find the source code here?

io.github.pieterdd.RcloneShuttle.yaml
hfiguiere:
```
$ which rclone
/usr/bin/which: no rclone in ...
```

so it won't work

pieterdd: Can I use the releases from https://github.com/rclone/rclone/releases? I've read a little bit about `extra-data` before but I'm not sure how I can automatically get the right ZIP for the user's architecture and install the binary.

If rclone is missing from the user's path, the app handles that:
image

io.github.pieterdd.RcloneShuttle.yaml
pieterdd: I'm new to this, I only know the manual said I should add my manifest here. Should I mirror all source code here? Because that might get tricky to keep in sync with the real repo.

io.github.pieterdd.RcloneShuttle.yaml
hfiguiere: extra-data isn't apropriate. Just build it from source.

hfiguiere: > If rclone is missing from the user's path, the app handles that:

Flatpak are supposed to be self contained.

io.github.pieterdd.RcloneShuttle.yaml
hfiguiere: > Should I mirror all source code here?

No. There are literally THOUSANDS of package to look at for inspiration.

hfiguiere: tl;dr you specify a source, either a tarball or a git repository.

the basic example in the doc show how to do that with a tarball

https://docs.flathub.org/docs/for-app-authors/requirements

io.github.pieterdd.RcloneShuttle.yaml
pieterdd: I took a stab at building it from source. I found https://github.com/flatpak/flatpak-builder-tools/tree/master/go-modules and am currently figuring out how to deal with build errors like these:

```
go: inconsistent vendoring in /run/build/rclone:
	bazil.org/fuse@v0.0.0-20230120002735-62a210ff1fd5: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/Azure/azure-sdk-for-go/sdk/azcore@v1.9.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/Azure/azure-sdk-for-go/sdk/azidentity@v1.4.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
```

Stack Overflow says to run `go mod vendor`, which of course fails because Flatpak builds should not access the internet. I'll see if I can find a Flathub app written in Go for reference. I have no experience with the Go ecosystem so bear with me as I grind out the learning curve.

io.github.pieterdd.RcloneShuttle.yaml
```diff
@@ -0,0 +1,49 @@
...
+command: rclone-shuttle
+finish-args:
+- "--share=ipc"
+- "--share=network"
+- "--socket=fallback-x11"
+- "--socket=wayland"
+- "--device=dri"
+- "--env=G_MESSAGES_DEBUG=none"
+- "--env=RUST_BACKTRACE=1"
+- "--filesystem=~/.config/rclone"
```
hfiguiere, line 17: this should be `persist`...

hfiguiere:
bot, build io.github.pieterdd.RcloneShuttle

pieterdd:
Alright, I:

- Bumped to platform v46
- Removed the redundant absolute `cargo` path
- Added the Go SDK to inline the rclone build (it's MIT-licensed, should be fine)
- Descoped filesystem permissions to ~/.config/rclone
- Removed the `dir` source for Rclone Shuttle in favor of a `git` source pointing to a specific version tag

I have two new lint issues I didn't had before: `appstream-external-screenshot-url` and `appstream-remote-icon-not-mirrored`. I noticed lines like these appearing in `gunzip -c /app/share/app-info/xmls/io.github.pieterdd.RcloneShuttle.xml.gz`:

```
<image type="source" width="922" height="722">io/github/pieterdd.RcloneShuttle/fb4447bb71481061576206877e4a357f/screenshots/image-1_orig.png</image>
```

as opposed to:

```
<image type="source" width="922" height="722">https://dl.flathub.org/media/io/github/pieterdd.RcloneShuttle/fb4447bb71481061576206877e4a357f/screenshots/image-1_orig.png</image>
```

What determines whether I end up with a flathub.org URL or a generic relative path?

pieterdd:
Having tested `https://dl.flathub.org/build-repo/100851/io.github.pieterdd.RcloneShuttle.flatpakref`, I realize that getting rid of `--filesystem=host` breaks the drag-and-drop file upload functionality as well as host-mounted SMB targets.

I'm afraid I'll have to change it back to `--filesystem=host`.

bbhtt:
There is the Filetransfer portal for drag and drop without filesystem access.

pieterdd:
Good to know. I'll check if I can find a clear implementation example that works with GTK4's Rust bindings.

hfiguiere:
Gtk4 drag and drop use this by default. there is nothing special

pieterdd:
Then I wonder why it's not working for me without `--filesystem=host`. I instrumented [the application](https://github.com/pieterdd/RcloneShuttle/blob/3f4a81491a6fe9ae13a5859546f0a7f683e7739c/src/main.rs#L840-L841) to see what happens when I drop a file:

```
Dropped local path: /home/pieter/Desktop/tinyfile
Exists: false
```

I receive a path to a file that isn't accessible.

PR #5231, opened 2024-05-10
Add org.kde.sweeper

Mailaender:

> Utility to clear caches
- [X] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about submitting their software to Flathub. **Link:** https://invent.kde.org/teams/flathub/issues/-/issues/14
- [X] Any additional patches or files have been submitted to the upstream projects concerned. *(If not, explain why.)* https://invent.kde.org/utilities/sweeper/-/merge_requests/18

### DIFF
--- a/appstream.patch
+++ b/appstream.patch
@@ -1,0 +1,79 @@
+Date: Fri, 10 May 2024 16:40:35 +0000
+Subject: [PATCH 1/3] Add developer name.
+
+---
+ org.kde.sweeper.appdata.xml | 1 +
+ 1 file changed, 1 insertion(+)
+
+diff --git a/org.kde.sweeper.appdata.xml b/org.kde.sweeper.appdata.xml
+index d2dcd0a..22f7499 100644
+--- a/org.kde.sweeper.appdata.xml
++++ b/org.kde.sweeper.appdata.xml
+@@ -3,6 +3,7 @@
+   <id>org.kde.sweeper.desktop</id>
+   <metadata_license>FSFAP</metadata_license>
+   <project_license>LGPL-2.1+</project_license>
++  <developer_name translate="no">KDE</developer_name>
+   <name>Sweeper</name>
+   <name xml:lang="ar">الكناس</name>
+   <name xml:lang="ast">Barrenderu</name>
+-- 
+GitLab
+
+
+Date: Fri, 10 May 2024 23:03:10 +0200
+Subject: [PATCH 2/3] Add launchable.
+
+---
+ org.kde.sweeper.appdata.xml | 1 +
+ 1 file changed, 1 insertion(+)
+
+diff --git a/org.kde.sweeper.appdata.xml b/org.kde.sweeper.appdata.xml
+index 22f7499..f75b3fc 100644
+--- a/org.kde.sweeper.appdata.xml
++++ b/org.kde.sweeper.appdata.xml
+@@ -178,4 +178,5 @@
+     <release version="24.02.0" date="2024-02-28"/>
+     <release version="23.08.5" date="2024-02-15"/>
+   </releases>
++  <launchable type="desktop-id">org.kde.sweeper.desktop</launchable>
+ </component>
+-- 
+GitLab
+
+
+Date: Fri, 10 May 2024 20:49:00 +0000
+Subject: [PATCH 3/3] Fix AppStream ID.
+
+---
+ org.kde.sweeper.appdata.xml | 6 +++++-
+ 1 file changed, 5 insertions(+), 1 deletion(-)
+
+diff --git a/org.kde.sweeper.appdata.xml b/org.kde.sweeper.appdata.xml
+index f75b3fc..f1567fb 100644
+--- a/org.kde.sweeper.appdata.xml
++++ b/org.kde.sweeper.appdata.xml
+@@ -1,6 +1,6 @@
+ <?xml version="1.0" encoding="utf-8"?>
+ <component type="desktop">
+-  <id>org.kde.sweeper.desktop</id>
++  <id>org.kde.sweeper</id>
+   <metadata_license>FSFAP</metadata_license>
+   <project_license>LGPL-2.1+</project_license>
+   <developer_name translate="no">KDE</developer_name>
+@@ -170,8 +170,12 @@
+   </screenshots>
+   <project_group>KDE</project_group>
+   <provides>
++    <id>org.kde.sweeper.desktop</id>
+     <binary>sweeper</binary>
+   </provides>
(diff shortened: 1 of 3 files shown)


Mailaender:
bot, build org.kde.sweeper

org.kde.sweeper.json
```diff
@@ -0,0 +1,42 @@
...
+    "sdk": "org.kde.Sdk",
+    "rename-icon": "sweeper",
+    "command": "sweeper",
+    "finish-args": [
+        "--device=dri",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--talk-name=org.kde.sweeper",
+        "--filesystem=xdg-cache"
```
hfiguiere, line 14: why would it have this permission? it's the build failure.

org.kde.sweeper.json
```diff
@@ -0,0 +1,42 @@
...
+    "runtime-version": "6.7",
+    "sdk": "org.kde.Sdk",
+    "rename-icon": "sweeper",
+    "command": "sweeper",
+    "finish-args": [
+        "--device=dri",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--talk-name=org.kde.sweeper",
```
hfiguiere, line 13:
```suggestion
```

this is granted by default.

appstream.patch
```diff
@@ -0,0 +1,28 @@
...
+---
+ org.kde.sweeper.appdata.xml | 2 ++
+ 1 file changed, 2 insertions(+)
+
+diff --git a/org.kde.sweeper.appdata.xml b/org.kde.sweeper.appdata.xml
+index d2dcd0a..f75b3fc 100644
+--- a/org.kde.sweeper.appdata.xml
++++ b/org.kde.sweeper.appdata.xml
+@@ -3,6 +3,7 @@
+   <id>org.kde.sweeper.desktop</id>
```
hfiguiere, line 13: @bbhtt does this `id` need to be changed?

org.kde.sweeper.json
Mailaender: It needs access to `~/.cache`.

Mailaender:
bot, build org.kde.sweeper

appstream.patch
Mailaender: Yes https://invent.kde.org/teams/flathub/issues/-/issues/28#error-appstream-id-mismatch-flatpak-id

Mailaender:
bot, build org.kde.sweeper

appstream.patch
```diff
@@ -0,0 +1,66 @@
...
+
+diff --git a/org.kde.sweeper.appdata.xml b/org.kde.sweeper.appdata.xml
+index f75b3fc..94fd0ca 100644
+--- a/org.kde.sweeper.appdata.xml
++++ b/org.kde.sweeper.appdata.xml
+@@ -1,6 +1,6 @@
+ <?xml version="1.0" encoding="utf-8"?>
+ <component type="desktop">
+-  <id>org.kde.sweeper.desktop</id>
++  <id>org.kde.sweeper</id>
```
bbhtt, line 60: You have to add a provides tag. https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/#renaming-id-tag

Mailaender: This is not stated in https://invent.kde.org/teams/flathub/issues/-/issues/28#error-appstream-id-mismatch-flatpak-id

Mailaender:
bot, build org.kde.sweeper

appstream.patch
bbhtt: Needs to be documented there too.

appstream.patch
```diff
@@ -0,0 +1,77 @@
...
+   <metadata_license>FSFAP</metadata_license>
+   <project_license>LGPL-2.1+</project_license>
+   <developer_name translate="no">KDE</developer_name>
+@@ -179,4 +179,10 @@
+     <release version="23.08.5" date="2024-02-15"/>
+   </releases>
+   <launchable type="desktop-id">org.kde.sweeper.desktop</launchable>
++  <provides>
++    <id>org.kde.sweeper.desktop</id>
++  </provides>
```
bbhtt, line 70: Looks like there is already a provides tag in metainfo file somewhere.

Mailaender: Sorry, I am not very good at this patch juggling.

Mailaender:
bot, build org.kde.sweeper

bbhtt:
Looks good from my side waiting for a comment from KDE team.

travier:
Hum, doesn't this need full RW access to the home directory and all dot folder to be useful?

Mailaender:
I forgot to add a replacement for `xdg-cache` which was removed during review.

Mailaender:
bot, build org.kde.sweeper

travier:
bot, build org.kde.sweeper

travier:
Manifest LGTM and the app runs.

However I've not tried it yet as I don't know exactly what it removes and it has full home access so I will have to make backups before just to be safe here or run that in a VM.

Mailaender:
I am also unsure if it works at all because of the sandbox restrictions. Sounds like → `beta` to me.

Justinzobel:
Ping.

travier:
@Justinzobel We need someone to do the testing for this one.

Justinzobel:
I've never used it so probably not a good test candidate. Was just trying to see if we can move it along.

RacerBG:
@travier The application itself is working however something doesn't seem right.

1. The "General" settings are cleaned as they should in my case.
2. The "Web Browsing" though is a different beast. I assume that the problem might be that I'm using Firefox as a flatpak application. I tried to clean only certain stuff because I don't want to nuke my browsing history and cookies in the process. :smile: 
2.1. Favorite Icons and Form Completion Entries are claimed to be cleaned by Sweeper however they are not as far as I can tell (this applies more for the latter because I can't prove that for the former).
2.2. The Web Cache cleaning is definitely failing both in Sweeper and as shown in the browser. I used to have around 1.2 GB cache and it's still the same.

OS: Debian 12 Bookworm
Desktop: KDE 5.27.2
Flatpak: 1.14.10
Firefox (as a flatpak package): 131.0

travier:
So it looks like this works at least a bit but might need some development work. Maybe a good fit for beta?

RacerBG:
> So it looks like this works at least a bit but might need some development work. Maybe a good fit for beta?

Beta sounds good to me. At least Sweeper will be "out there" and more people could test it and report any potential issues. But it's far from ready for anything else.

PR #5240, opened 2024-05-13
Add org.dbgate.DbGate

janproch:
DbGate is SQL+noSQL database client supporting MySQL, PostgreSQL, MS SQL, SQLite, MongoDB, Redis

An additional permission - access to home directory is used, because DbGate stores it's data (eg. connections, scripts) in home directory


org.dbgate.DbGate.yml
```diff
@@ -0,0 +1,71 @@
...
+        for lang in ${FLATPAK_DEST}/dbgate/locales/*.pak
+        do
+          locale="$(basename -s .pak $lang)"
+          install -Dm644 -t "${FLATPAK_DEST}/share/runtime/locale/${locale%%-*}/" "$lang"
+          ln -sf "${FLATPAK_DEST}/share/runtime/locale/${locale%%-*}/$(basename $lang)" "${FLATPAK_DEST}/dbgate/locales/$(basename $lang)"
+        done
+    sources:
+      - type: file
+        only-arches: [x86_64]
+        url: https://github.com/dbgate/dbgate/releases/download/v5.2.8-beta.22/dbgate-5.2.8-beta.22-linux_x86_64.AppImage
```
hfiguiere, line 55: why not building from source?

org.dbgate.DbGate.yml
```diff
@@ -0,0 +1,71 @@
...
+base-version: '23.08'
+command: dbgate
+rename-icon: dbgate
+finish-args:
+  - --device=dri
+  - --share=ipc
+  - --share=network
+  - --socket=x11
+  - --filesystem=home
+  - --talk-name=org.freedesktop.Notifications
```
hfiguiere, line 15: this shouldn't be needed.

org.dbgate.DbGate.metainfo.xml
```diff
@@ -0,0 +1,39 @@
...
+    <url type="vcs-browser">https://github.com/dbgate/dbgate</url>
+    <url type="contact">https://dbgate.org/about/</url>
+    <url type="donation">https://github.com/sponsors/dbgate</url>
+    <url type="bugtracker">https://github.com/dbgate/dbgate/issues</url>
+
+    <launchable type="desktop-id">org.dbgate.DbGate.desktop</launchable>
+    <screenshots>
+        <screenshot type="default">
+            <image>
+                https://raw.githubusercontent.com/dbgate/dbgate/master/img/screenshot1.png</image>
```
hfiguiere, line 26: this should NEVER reference a branch, because if the file is remove it will break. a specific commit should be referenced.

org.dbgate.DbGate.yml
```diff
@@ -0,0 +1,71 @@
...
+          install -Dm644 -t "${FLATPAK_DEST}/share/runtime/locale/${locale%%-*}/" "$lang"
+          ln -sf "${FLATPAK_DEST}/share/runtime/locale/${locale%%-*}/$(basename $lang)" "${FLATPAK_DEST}/dbgate/locales/$(basename $lang)"
+        done
+    sources:
+      - type: file
+        only-arches: [x86_64]
+        url: https://github.com/dbgate/dbgate/releases/download/v5.2.8-beta.22/dbgate-5.2.8-beta.22-linux_x86_64.AppImage
+        # url: file:///home/jena/jenasoft/dbgate/app/dist/dbgate-5.0.0-alpha.1-linux_x86_64.AppImage
+        sha256: 5d55a62ac2fbed363acf226733655e284b9b7e530653bdadbe67e95a08169155
+        dest_filename: dbgate.AppImage
```
bbhtt, line 57:
```suggestion
        dest-filename: dbgate.AppImage
```

org.dbgate.DbGate.yml
```diff
@@ -0,0 +1,71 @@
...
+        
+      - type: file
+        only-arches: [aarch64]
+
+        url: https://github.com/dbgate/dbgate/releases/download/v5.2.8-beta.22/dbgate-5.2.8-beta.22-linux_arm64.AppImage
+        sha256: 9ddb17850d9676e9f2ac3d5ea2defbd072ca8462856d863b3b28a13c250211eb
+        dest_filename: dbgate.AppImage
```
bbhtt, line 64:
```suggestion
        dest-filename: dbgate.AppImage
```

org.dbgate.DbGate.yml
janproch: OK, I tested it and it's not, removed

org.dbgate.DbGate.metainfo.xml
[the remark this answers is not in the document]
janproch: Could you please be more specific?

I plan to add updating of this file to github pipeline, when new version is released, this file will be updated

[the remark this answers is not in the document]
bbhtt: It should live here https://github.com/dbgate/dbgate and it should be included in any packaging format you produce.

This is not at all specific to Flathub.

org.dbgate.DbGate.yml
```diff
@@ -0,0 +1,71 @@
...
+base: org.electronjs.Electron2.BaseApp
+base-version: '23.08'
+command: dbgate
+rename-icon: dbgate
+finish-args:
+  - --device=dri
+  - --share=ipc
+  - --share=network
+  - --socket=x11
+  - --filesystem=home
```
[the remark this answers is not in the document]
janproch: OK, it seems that --filesystem=~/.dbgate is enough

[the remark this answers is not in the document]
hfiguiere: this should be `--persist` instead. Or better, it should use XDG directories to store its data.

org.dbgate.DbGate.yml
janproch: Because network requests are prohibited while flatpak build. Prebuilt node modules are downloaded during the build and precaching is very difficult

hfiguiere: there is tooling for that

janproch: I saw several electron apps on flathub and they use appimage too. Electron build process is quite complicated, especially when using natrive modules, so I prefer to have core build only on one place - that is github pipeline

hfiguiere: > I saw several electron apps on flathub and they use appimage too.

org.dbgate.DbGate.yml
```diff
@@ -0,0 +1,71 @@
...
+  - --talk-name=org.freedesktop.Notifications
+modules:
+  - name: dbgate
+    buildsystem: simple
+    build-commands:
+      - chmod +x dbgate.AppImage
+      - ./dbgate.AppImage --appimage-extract
+      - rm dbgate.AppImage
```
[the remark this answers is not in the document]
janproch: fixed

org.dbgate.DbGate.yml
janproch: Sorry, now is build from source for me too difficult. Also it could easily break, when two complicated and very different pipelines (github, flathub) must be maintained. It is possible to use appimage for now?

org.dbgate.DbGate.metainfo.xml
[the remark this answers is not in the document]
bbhtt: > Should I add it to sources in org.dbgate.DbGate.yml file?

Yes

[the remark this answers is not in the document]
hfiguiere: > I prefer the solution with metainfo in flathub repo, 

the appstream file, unlike the manifest, is NOT specific to flatpak.

bbhtt:
bot, build org.dbgate.DbGate

org.dbgate.DbGate.yml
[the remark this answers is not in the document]
bbhtt: > this should be `--persist` instead. Or better, it should use XDG directories to store its data.

@janproch This isn't solved. Please test if it works with `--persist=.dbgate`

bbhtt:
Looks ok, except the last comment https://github.com/flathub/flathub/pull/5240#discussion_r1598488487.

org.dbgate.DbGate.yml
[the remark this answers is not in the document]
janproch: OK, I changed it to --persist switch, it works, thanks

PR #5300, opened 2024-05-31
Add io.github.rdybka.vht

rdybka:
vahatraker is a minimal midi sequencer for pipewire/JACK. It is written with live
use in mind and strives for fast workflow and minimal distractions.


net.sourceforge.projects.vht.yml
```diff
@@ -0,0 +1,19 @@
+id: net.sourceforge.projects.vht
+runtime: org.gnome.Platform
+runtime-version: '46'
+sdk: org.gnome.Sdk
+command: vht
+finish-args:
+  - --filesystem=home
+  - --filesystem=xdg-run/pipewire-0
+  - --socket=wayland
+  - --socket=fallback-x11
```
hfiguiere, line 10: missing `--share=ipc` (need to x11 performance)
why does it need `home` access?

net.sourceforge.projects.vht.yml
```diff
@@ -0,0 +1,19 @@
+id: net.sourceforge.projects.vht
```
hfiguiere, line 1: I have pretty sure this is incorrect. The project isn't even hosted on SourceForge. You you did 

> The domain used for the application ID is controlled by the application developers either directly or through the code hosting (e.g. GitHub, GitLab, SourceForge, etc.). The [application id guidelines](https://docs.flathub.org/docs/for-app-authors/requirements#application-id) are followed.

net.sourceforge.projects.vht.yml
```diff
@@ -0,0 +1,19 @@
...
+finish-args:
+  - --filesystem=home
+  - --filesystem=xdg-run/pipewire-0
+  - --socket=wayland
+  - --socket=fallback-x11
+modules:
+  - name: vht
+    buildsystem: simple
+    build-commands:
+      - pip3 install . --no-build-isolation --prefix=${FLATPAK_DEST}
```
hfiguiere, line 15: build is offline so it won't find the dependencies.

net.sourceforge.projects.vht.yml
rdybka: for config and saving files

net.sourceforge.projects.vht.yml
rdybka: This is correct. I have other project there, like zaz

net.sourceforge.projects.vht.yml
hfiguiere: if it uses the standard XDG directory (and it should) you don't need any permission.

net.sourceforge.projects.vht.yml
hfiguiere: This is not. You certainly don't have `projects.sourceforge.net`

net.sourceforge.projects.vht.yml
rdybka: Sorry, but I'm not sure what that means (I'm new to this). Is this about --no-build-isolation?

hfiguiere: do you try to build it locally?

net.sourceforge.projects.vht.yml
```diff
@@ -8,12 +8,6 @@ finish-args:
...
 modules:
-  - name: vht
-    buildsystem: simple
-    build-commands:
-      - pip3 install . --no-build-isolation --prefix=${FLATPAK_DEST}
-    sources:
-      - type: git
-        url: https://github.com/rdybka/vht
-        tag: v0.3.97
+  - python3-vht.json
```
hfiguiere, line 13: no how it was was fine.

I asked about the dependencies because upstream mention thee dependency `pipewire-audio-client-libraries [jack_capture]` and I didn't see it installed.

rdybka: oh, these are just to build it on a generic linux box

rdybka: jack_capture can get invoked for rendering of projects. The option will not display in sandbox as I'm not sure how to implement it yet.

net.sourceforge.projects.vht.yml
rdybka: would github.com.rdybka.vht work?

rdybka: ok, changed it

net.sourceforge.projects.vht.yml
rdybka: it builds and runs fine locally

hfiguiere:
bot, build io.github.rdybka.vht

net.sourceforge.projects.vht.yml
```diff
@@ -0,0 +1,22 @@
+id: net.sourceforge.projects.vht
+runtime: org.gnome.Platform
+runtime-version: '46'
+sdk: org.gnome.Sdk
+command: vht
+finish-args:
+  - --filesystem=xdg-home
+  - --filesystem=xdg-config
+  - --filesystem=xdg-data
```
hfiguiere, line 9:
```suggestion
```

There is no legitimate use of these.

net.sourceforge.projects.vht.yml
hfiguiere, line 1: actually remove this file. it no longer needed / used.

hfiguiere:
```
error: /srv/buildbot/worker/build-x86_64-5/build/builddir/export/share/icons/hicolor/scalable/apps/io.github.rdybka.vht.svg is not a valid icon: Expected a square icon but got: 353x342
```

io.github.rdybka.vht.yml
```diff
@@ -0,0 +1,21 @@
+id: io.github.rdybka.vht
+runtime: org.gnome.Platform
+runtime-version: '46'
+sdk: org.gnome.Sdk
+command: vht
+finish-args:
+  - --filesystem=xdg-config
+  - --filesystem=xdg-data
```
hfiguiere, line 8: no it shouldn't need these two.

net.sourceforge.projects.vht.yml
bbhtt: > but it needs the user to be able to save and open files from a gtk dialog. 

What does it use for filechooser? If it supports the portal then you need nothing.

net.sourceforge.projects.vht.yml
bbhtt: > would github.com.rdybka.vht work?

No

The current io.github one is ok.

net.sourceforge.projects.vht.yml
rdybka: It uses filechooser for saving and opening of user projects files. It expects (for now at least) to save and load user settings in ~/.config/vht and if it can't it will fail at start. It doesn't really need access to home itself but it just looks better for the user integration-wise imo. It also gets default instrument/controller definition files from the data directory, which (more advanced) users can later edit to suit their needs. Not saying it has to stay that way, just how to make it run in current state.

io.github.rdybka.vht.yml
rdybka: "just change where u get ur config path from, dude" - would have worked first time :)
Like I said, I'm new to this and sometimes assume things have to be certain way when they don't.

net.sourceforge.projects.vht.yml
bbhtt: > It uses filechooser for saving and opening of user projects files. 

GtkFilechooserNative ? 

> ~/.config/vht

How is the directory calculated? Relative to `$HOME` or $XDG_CONFIG_HOME`?

rdybka: .config dir was wrong - fixed it - it now takes it from GLib.get_user_config_dir()

It uses Gtk.FileChooserDialog and seems to save and open files fine, but after
I close the program and reopen, the files are "gone". Maybe it's just me doing it wrong.

bbhtt: > Gtk.FileChooserDialog

If that is gtk3, it does not support portals if you switch to GtkFilechooserNative it will work without any `--filesystem=` access.

rdybka: but then I have the same issue using qpwgraph so maybe it is just me...

rdybka: Native dialog did the trick. Thank you!

rdybka: Thank you for getting me to remove those unneeded perms.
I got nothing to add. Do I click something now?

bbhtt:
bot, build io.github.rdybka.vht

bbhtt:
https://github.com/rdybka/vht/blob/a6b77a7971d6e95cb04347fbf877e6e0f7f4b8ee/data/io.github.rdybka.vht.metainfo.xml#L29 does not match new desktop file, the app id changed.

bbhtt:
bot, build io.github.rdybka.vht

PR #5310, opened 2024-06-02
Add com.simutrans.Simutrans

Roboron3042:
Some notes:

1. In my local builds, flatpak always fails to download the miniupnpc source (wrong sha256, not sure what is the cause). This didn't happen when I wrote the manifest initially (two years ago), but it is happening now. Fortunately this is only an optional dependency, so I have commented/excluded it for now.
2. At some point in the cmake install step, there's a script that will try to download translation texts. It will not be able, of course, but it does not make the install step fail and can be completely ignored since translation texts are updated before every release.
3. If possible, add Simutrans lead developer @prissi to the final repository. 


hfiguiere:
> In my local builds, flatpak always fails to download the miniupnpc source (wrong sha256, not sure what is the cause). 

Tried twice to download it and got twice the same checksum. Maybe the failure was transient.
If it's a required dependency you need a solution otherwise there is no point in reviewing this.

hfiguiere:
> At some point in the cmake install step, there's a script that will try to download translation texts.

Upstream should fix this. The right way is to package the translation. So updating them regularly for release purposes.

com.simutrans.Simutrans.metadata.xml
hfiguiere, line 1: this should be submitted upstream. It's not flatpak specific

com.simutrans.Simutrans.yaml
```diff
@@ -0,0 +1,67 @@
...
+  - name: simutrans
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DSIMUTRANS_USE_REVISION=11266
+    builddir: true
+    post-install:
+      - mv $FLATPAK_DEST/share/icons/hicolor/scalable/apps/simutrans.svg $FLATPAK_DEST/share/icons/hicolor/scalable/apps/com.simutrans.Simutrans.svg
+      - mv $FLATPAK_DEST/share/applications/simutrans.desktop $FLATPAK_DEST/share/applications/com.simutrans.Simutrans.desktop
+      - sed -i "s/Icon=simutrans/Icon=com.simutrans.Simutrans/g" $FLATPAK_DEST/share/applications/com.simutrans.Simutrans.desktop
```
hfiguiere, line 53: don't use sed for that. Use the `desktop-file-edit` command.

com.simutrans.Simutrans.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+  - name: simutrans
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DSIMUTRANS_USE_REVISION=11266
+    builddir: true
+    post-install:
+      - mv $FLATPAK_DEST/share/icons/hicolor/scalable/apps/simutrans.svg $FLATPAK_DEST/share/icons/hicolor/scalable/apps/com.simutrans.Simutrans.svg
+      - mv $FLATPAK_DEST/share/applications/simutrans.desktop $FLATPAK_DEST/share/applications/com.simutrans.Simutrans.desktop
```
hfiguiere, line 52: you should use `rename-desktop-file` in the manifest instead as this would also fix the currently incorrect launchable in the appstream file

com.simutrans.Simutrans.yaml
```diff
@@ -0,0 +1,67 @@
...
+#        sha256: 741175f3a6edb68aa2a5a62a576ffc5596f28d775e6be0a447a41d5881f145ee
+
+  - name: simutrans
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DSIMUTRANS_USE_REVISION=11266
+    builddir: true
+    post-install:
+      - mv $FLATPAK_DEST/share/icons/hicolor/scalable/apps/simutrans.svg $FLATPAK_DEST/share/icons/hicolor/scalable/apps/com.simutrans.Simutrans.svg
```
hfiguiere, line 51: you should use `rename-icon` instead. This would also make the desktop-file modification uncessary (Line 53)

com.simutrans.Simutrans.yaml
```diff
@@ -0,0 +1,67 @@
...
+finish-args:
+  - --device=dri
+  - --share=ipc
+  - --share=network
+  - --socket=pulseaudio
+  - --socket=fallback-x11
+  - --socket=wayland
+modules:
+  - shared-modules/linux-audio/fluidsynth2.json
+  - name: bzip2
```
hfiguiere, line 15: there is `libbz2.so.1.0.8` in the runtime.

com.simutrans.Simutrans.metadata.xml
hfiguiere: and it currently is never installed.

com.simutrans.Simutrans.yaml
```diff
@@ -0,0 +1,67 @@
...
+#        - -DUPNPC_BUILD_STATIC=FALSE
+#        - -DUPNPC_BUILD_SHARED=TRUE
+#        - -DUPNPC_BUILD_TESTS=FALSE
+#        - -DUPNPC_BUILD_SAMPLE=FALSE
+#    cleanup:
+#        - /include
+#        - /share/doc
+#    sources:
+#      - type: archive
+#        url: http://miniupnp.free.fr/files/miniupnpc-2.2.7.tar.gz
```
bbhtt, line 41: Try https://miniupnp.tuxfamily.org/files/miniupnpc-2.2.7.tar.gz

com.simutrans.Simutrans.metadata.xml
```diff
@@ -0,0 +1,41 @@
...
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>Artistic-1.0</project_license>
+  <name>Simutrans</name>
+  <summary>Open source transportation simulation game</summary>
+  <description>
+    <p>
+		Simutrans is a freeware and open-source transportation simulator. Your goal is to establish a successful transport company. Transport passengers, mail and goods by rail, road, ship, and even air. Interconnect districts, cities, public buildings, industries and tourist attractions.
+    </p>
+  </description>
+  <launchable type="desktop-id">simutrans.desktop</launchable>
```
bbhtt, line 14: This should be com.simutrans.Simutrans.desktop if you use rename-desktop-file it will get updated automatically.

Roboron3042:
Thank you for your comments. I have successfully deal with everything you have pointed out:

- The metadata file is now upstream, and it is installed in the CMake install step.
- rename-desktop-file  and rename-icon are used instead of the previous post-install steps.
- The install step will no longer try to download translations.
- And compilation with miniupnpc works again! Thanks @bbhtt for the alternate source link.

bbhtt:
bot, build com.simutrans.Simutrans

Roboron3042:
Everything seems to work right. I can download paksets, play and join stable multiplayer servers with the test build 👍
imagen

PR #5328, opened 2024-06-11
Add im.nheko.Nheko

deepbluev7:
Because our old app id didn't match our github organization and also didn't match our domain, we weren't able to verify it. Additionally it didn't match our dbus interface and other ids we were using. As such switching to a new app id seemed like the easiest way forward. Additionally the Qt6 port seemed like a good time to do so.

Apart from that the packaging has barely any changes compared to io.github.NhekoReborn.Nheko. I plan to do an eol-rebase with that package soon.

# Description

Nheko is a Matrix client built with Qt, that aims to feel like a mainstream chat app.

# Permissions

The permissions are not as minimal as we would like, however they match what the old package used. Tray icon support sadly still requires claiming almost all of the kde namespace. Similarly we still default to x11 because of the screenshare support. We plan to default to wayland soon, after we gave the new pipewire screenshare support a bit more testing. Webcam support also still seems to need device permissions.

### DIFF
--- a/im.nheko.Nheko.yaml
+++ b/im.nheko.Nheko.yaml
@@ -4,19 +4,22 @@ runtime: org.kde.Platform
 runtime-version: '6.6'
 sdk: org.kde.Sdk
 finish-args:
-  - --device=dri
-  # needed for webcams, see #517
+  # needed for webcams, see https://github.com/Nheko-Reborn/nheko/issues/517
   - --device=all
   - --share=ipc
   - --share=network
   - --socket=pulseaudio
-  - --socket=wayland
+  # Since we need access to the X11 socket for sharing some X11 apps and the pipewire sharing is still experimental, default to X11.
+  # Since we default to X11, we are not supposed to add the wayland socket.
   - --socket=x11
+  #- --socket=wayland
   - --talk-name=org.freedesktop.Notifications
   - --talk-name=org.freedesktop.secrets
+  - --talk-name=org.kde.kwalletd5
+  - --talk-name=org.kde.kwalletd6
   - --talk-name=org.freedesktop.StatusNotifierItem
   # needed for tray icon
-  - --talk-name=org.kde.*
+  - --talk-name=org.kde.StatusNotifierWatcher
 cleanup:
   - /include
   - /lib/pkgconfig
@@ -183,6 +186,7 @@ modules:
         tag: 1.22.12
         type: git
         url: https://gitlab.freedesktop.org/gstreamer/gstreamer.git
+        disable-submodules: true
     config-opts:
       - --auto-features=disabled
       -  -Dgood=enabled
--- a/im.nheko.Nheko.yaml
+++ b/im.nheko.Nheko.yaml
@@ -1,0 +1,247 @@
+id: im.nheko.Nheko
+command: im.nheko.Nheko
+runtime: org.kde.Platform
+runtime-version: '6.6'
+sdk: org.kde.Sdk
+finish-args:
+  - --device=dri
+  # needed for webcams, see #517
+  - --device=all
+  - --share=ipc
+  - --share=network
+  - --socket=pulseaudio
+  - --socket=wayland
+  - --socket=x11
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.freedesktop.secrets
+  - --talk-name=org.freedesktop.StatusNotifierItem
+  # needed for tray icon
+  - --talk-name=org.kde.*
+cleanup:
+  - /include
+  - /lib/pkgconfig
+  - /bin/mdb*
+  - '*.a'
+  - /libexec
+  - /lib/cmake
+  - /bin/cmark
+  - /bin/event_rpcgen.py
+  - /bin/playout
+  - /bin/secret-tool
+  - /bin/gst-*
+  - /share/gdb
+  - /share/gst*
+  - /lib/girepository-1.0/
+  - /lib/gst-validate-launcher/
+  - /lib/gstreamer-1.0/include
+  - /lib/gstreamer-1.0/include/
+  - /lib/gstreamer-1.0/libgstcoreelements.so
+  - /lib/gstreamer-1.0/libgstopengl*
+  - /lib/gstreamer-1.0/libgstximagesink.so
+  - /lib/gstreamer-1.0/validate/
+  - /lib/libgst*
+modules:
+  - name: lmdb
+    sources:
+      - sha256: c0937223bba3c37c896809883a3c9b43049354578b239d6ed2476236a87f40c9
+        type: archive
+        url: https://git.openldap.org/openldap/openldap/-/archive/LMDB_0.9.32/openldap-LMDB_0.9.32.tar.gz
+    make-install-args:
+      - prefix=/app
+    no-autogen: true
+    subdir: libraries/liblmdb
+  - name: libevent
+    buildsystem: autotools
+    config-opts:
+      - --disable-shared
+      - --prefix=/app
+      - --disable-openssl
+    sources:
+      - sha256: 92e6de1be9ec176428fd2367677e61ceffc2ee1cb119035037a27d346b0403bb
+        type: archive
+        url: https://github.com/libevent/libevent/releases/download/release-2.1.12-stable/libevent-2.1.12-stable.tar.gz
+    make-install-args:
+      - prefix=/app
+    no-autogen: true
+  - name: cmark
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DCMARK_TESTS=OFF
+    sources:
+      - sha256: 6c7d2bcaea1433d977d8fed0b55b71c9d045a7cdf616e3cd2dce9007da753db3
+        type: archive
+        url: https://github.com/commonmark/cmark/archive/0.30.2.tar.gz
+  - name: spdlog
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DSPDLOG_BUILD_EXAMPLES=0
+      - -DSPDLOG_BUILD_BENCH=0
+      - -DSPDLOG_BUILD_TESTING=0
+    sources:
+      - sha256: 5197b3147cfcfaa67dd564db7b878e4a4b3d9f3443801722b3915cdeced656cb
+        type: archive
+        url: https://github.com/gabime/spdlog/archive/v1.8.1.tar.gz
+  - name: olm
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+    buildsystem: cmake-ninja
+    sources:
+      - commit: 92769cec711c604a1f682b95d6944578d2a1bb3d
+        disable-shallow-clone: true
+        tag: 3.2.12
+        type: git
+        url: https://gitlab.matrix.org/matrix-org/olm.git
+  - name: libsecret
+    buildsystem: meson
+    config-opts:
+      - -Dmanpage=false
+      - -Dvapi=false
+      - -Dgtk_doc=false
+      - -Dintrospection=false
+      # https://gitlab.gnome.org/GNOME/libsecret/-/issues/49
+      - -Dgcrypt=false
+    sources:
+      - commit: 3fe635e64efd4b8dbc9ec3548b0bc8034c7665c4
+        tag: 0.20.4
+        type: git
+        url: https://gitlab.gnome.org/GNOME/libsecret.git
+  #- config-opts:
+  #    - -DCMAKE_BUILD_TYPE=Release
+  #    - -DAVIF_CODEC_AOM=ON
+  #    #- -DBUILD_SHARED_LIBS=OFF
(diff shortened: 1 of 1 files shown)


deepbluev7:
Previous package for comparison: https://github.com/flathub/io.github.NhekoReborn.Nheko

im.nheko.Nheko.yaml
```diff
@@ -0,0 +1,247 @@
+id: im.nheko.Nheko
+command: im.nheko.Nheko
+runtime: org.kde.Platform
+runtime-version: '6.6'
```
hfiguiere, line 4: 6.7 is the current verison

im.nheko.Nheko.yaml
```diff
@@ -0,0 +1,247 @@
...
+sdk: org.kde.Sdk
+finish-args:
+  - --device=dri
+  # needed for webcams, see #517
+  - --device=all
+  - --share=ipc
+  - --share=network
+  - --socket=pulseaudio
+  - --socket=wayland
+  - --socket=x11
```
hfiguiere, line 14: it's either wayland with fallback or x11 alone.

im.nheko.Nheko.yaml
```diff
@@ -0,0 +1,247 @@
...
+  #      tag: v5.93.0
+  #      type: git
+  #      url: https://invent.kde.org/frameworks/kimageformats.git
+  - config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DBUILD_TEST_APPLICATION=OFF
+      - -DQTKEYCHAIN_STATIC=ON
+      - -DBUILD_WITH_QT6=ON
+    buildsystem: cmake-ninja
+    name: QtKeychain
```
hfiguiere, line 147: please make `name` the first item here. otherwise it's unreadable.

im.nheko.Nheko.yaml
```diff
@@ -0,0 +1,247 @@
...
+    sources:
+      - commit: 69f993c47efed7e557d79a30a367014d9a27d809
+        tag: 0.14.1
+        type: git
+        url: https://github.com/frankosterfeld/qtkeychain.git
+  - config-opts:
+      - -DJSON_BuildTests=OFF
+    buildsystem: cmake
```
hfiguiere, line 155: cmake-ninja?

im.nheko.Nheko.yaml
```diff
@@ -0,0 +1,247 @@
...
+    name: nlohmann
+    sources:
+      - sha256: d69f9deb6a75e2580465c6c4c5111b89c4dc2fa94e3a85fcd2ffcd9a143d9273
+        type: archive
+        url: https://github.com/nlohmann/json/archive/v3.11.2.tar.gz
+  - config-opts:
+      - -DKDSingleApplication_EXAMPLES=OFF
+      - -DKDSingleApplication_QT6=ON
+    buildsystem: cmake
```
hfiguiere, line 164: cmake-ninja?

im.nheko.Nheko.yaml
hfiguiere: all over the manifest, not just here

im.nheko.Nheko.yaml
```diff
@@ -0,0 +1,247 @@
...
+    name: re2
+    sources:
+      - sha256: f89c61410a072e5cbcf8c27e3a778da7d6fd2f2b5b1445cd4f4508bee946ab0f
+        type: archive
+        url: https://github.com/google/re2/archive/refs/tags/2022-06-01.tar.gz
+  - buildsystem: meson
+    name: gstreamer
+    sources:
+      - commit: 4d13eddc8b6d3f42ba44682ba42048acf170547f
+        tag: 1.22.7
```
hfiguiere, line 183: 1.22.12 is in the runtime. Why do you need to build it?

im.nheko.Nheko.yaml
```diff
@@ -0,0 +1,247 @@
...
+      -  -Dgst-plugins-good:qt-egl=enabled
+      -  -Dgst-plugins-good:qt-wayland=enabled
+      -  -Dgst-plugins-good:qt-x11=enabled
+      -  -Dbase=enabled
+      -  -Dgst-plugins-base:gl=enabled
+      -  -Dgst-plugins-base:gl_platform=glx,egl
+      -  -Dgst-plugins-base:gl_winsys=x11,wayland
+      -  -Dgst-plugins-base:x11=enabled
+      -  -Dgst-plugins-base:xshm=enabled
+  - buildsystem: cmake
```
hfiguiere, line 199: cmake-ninja?

im.nheko.Nheko.yaml
deepbluev7: Because the runtime doesn't include the qml module. We are only building the qml module.

im.nheko.Nheko.yaml
deepbluev7: Would the package then still work on systems without xwayland?

im.nheko.Nheko.yaml
deepbluev7: We haven't finished testing on 6.7 yet, which is why I went with 6.6 for now, which is what we tested for the last few months. I plan to bump it to 6.7 in a few weeks after we had some time to test it. Would that be okay?

im.nheko.Nheko.yaml
hfiguiere: that's what the fallback is about.

hfiguiere: `wayland` is NOT for `Xwayland` either.

im.nheko.Nheko.yaml
hfiguiere: then at least build from the same version, 1.22.12.

im.nheko.Nheko.yaml
```diff
@@ -0,0 +1,247 @@
...
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DBUILD_TEST_APPLICATION=OFF
+      - -DQTKEYCHAIN_STATIC=ON
+      - -DBUILD_WITH_QT6=ON
+    buildsystem: cmake-ninja
+    sources:
+      - commit: 69f993c47efed7e557d79a30a367014d9a27d809
+        tag: 0.14.1
+        type: git
+        url: https://github.com/frankosterfeld/qtkeychain.git
```
hfiguiere, line 152: whatever formatter you use should be reconfigure to not order by alphabetical order.

deepbluev7: That's not a formatter, that's just what we historically used. This file used to be the one for io.github.NhekoReborn.Nheko and we just adapted it to the new version and packages without reordering all fields. I don't remember why that order was chosen years ago, but there seem to be no rules for how to order things for flathub packages?

im.nheko.Nheko.yaml
deepbluev7: Fixed.

im.nheko.Nheko.yaml
deepbluev7: We currently want to run on x11 by default, because we need the x11 socket to share x11 windows. We can't use fallback-x11 in that case. However, if a system is configured without Xwayland and only supporting wayland apps, specifying the wayland socket should still allow Nheko to launch, only dropping the support for sharing the non-existing x11 windows.

So as far as I am concerned, "x11, wayland" is the correct socket configuration, since both work, but we need the x11 socket for screensharing. Am I misunderstanding something?

im.nheko.Nheko.yaml
deepbluev7: Actually, the order is probably from when we migrated from json to yaml.

im.nheko.Nheko.yaml
hfiguiere: no. the correct setting is `x11` and no `wayland`. A system without Xwayland broken.

deepbluev7: Nheko has quite a few users without xwayland, I would prefer to not break it for no reason?

hfiguiere: then you should enable wayland for everyone. it either work or doesn't

im.nheko.Nheko.yaml
bbhtt: Yea

im.nheko.Nheko.yaml
```diff
@@ -0,0 +1,247 @@
+id: im.nheko.Nheko
+command: im.nheko.Nheko
+runtime: org.kde.Platform
+runtime-version: '6.6'
+sdk: org.kde.Sdk
+finish-args:
+  - --device=dri
+  # needed for webcams, see #517
+  - --device=all
```
bbhtt, line 9: dri is redundant with all

im.nheko.Nheko.yaml
```diff
@@ -0,0 +1,247 @@
...
+  - --share=ipc
+  - --share=network
+  - --socket=pulseaudio
+  - --socket=wayland
+  - --socket=x11
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.freedesktop.secrets
+  - --talk-name=org.freedesktop.StatusNotifierItem
+  # needed for tray icon
+  - --talk-name=org.kde.*
```
bbhtt, line 19:
```suggestion
  - --talk-name=org.kde.StatusNotifierWatcher
```

https://github.com/flathub/io.github.NhekoReborn.Nheko/blob/59d2532add48f8b85536253ea8c0c52e080e0c33/io.github.NhekoReborn.Nheko.yaml#L17C17-L17C46

This should not be needed with recent runtimes, you just need to make sure it is not using `StatusNotifierItem-$PID-$ID`

bbhtt: If `org.kde.*` is needed, this needs an exception https://docs.flathub.org/docs/for-app-authors/linter#finish-args-wildcard-kde-own-name

But QT should not use it anymore. This is supposed to be fixed with Qt 6

im.nheko.Nheko.yaml
```diff
@@ -0,0 +1,247 @@
...
+      - sha256: f89c61410a072e5cbcf8c27e3a778da7d6fd2f2b5b1445cd4f4508bee946ab0f
+        type: archive
+        url: https://github.com/google/re2/archive/refs/tags/2022-06-01.tar.gz
+  - name: gstreamer
+    buildsystem: meson
+    sources:
+      - commit: d2c02bb704b5804ca057fc7e6c7b16b4466fd7d5
+        tag: 1.22.12
+        type: git
+        url: https://gitlab.freedesktop.org/gstreamer/gstreamer.git
```
bbhtt, line 185: Add `disable-submodules: true` here

im.nheko.Nheko.yaml
bbhtt: If you remove this, you will likely need to add the kwalletd interfaces (kde6 and 5 individually)

im.nheko.Nheko.yaml
bbhtt: > since both work, 

Unless you set QT_QPA_PLATFORM=wayland, Qt with this configuration will use Xwayland by default in a wayland session.

But since you can't set that, using x11 only seems fine? Those who have no Xwayland, will need to run with `--env=QT_QPA_PLATFORM=wayland` (or as an override) anyways. They can add `--socket=wayland` there as well?

deepbluev7: On my systems Qt seems to fallback to wayland transparently, but I can remove it and see, if people start complaining.

im.nheko.Nheko.yaml
deepbluev7: Thanks. I was looking for that option, but couldn't find it when I added that module originally!

im.nheko.Nheko.yaml
deepbluev7: Well, let's see if these permissions work. I wasn't aware that Qt6 removed the need for this and couldn't find documentation for it anywhere either.

im.nheko.Nheko.yaml
hfiguiere: as I said, if it works on wayland the proper permissions are `wayland + fallback-x11`

deepbluev7: No, they are not, because that breaks screensharing X11 windows, since you need the x11 socket for that. They might work using pipewire, but we haven't tested that enough yet.

bbhtt: > On my systems Qt seems to fallback to wayland transparently,

Yea seems to be the case for me too. I guess they didn't update the wiki.

im.nheko.Nheko.yaml
bbhtt: Tray seems to be working for me.

im.nheko.Nheko.yaml
bbhtt: Changed it https://github.com/flathub/flathub/pull/5328/commits/ea17b558f75d69a852ed7afa5e364c0f80ab6ecb

bbhtt:
Looks good from my end

deepbluev7:
> Looks good from my end. @deepbluev7 is it ready from your end?

From our side everything is looking fine. We are currently also testing Qt6.7 in our nightlies, so we might update the build to 6.7 in a few weeks or so.

Issue #5228, opened 2024-05-10
[Tracking]: Apps with trademark violations


bbhtt:
- [ ] https://flathub.org/apps/io.github.mimbrero.WhatsAppDesktop
- [ ] https://flathub.org/apps/com.github.eneshecan.WhatsAppForLinux
- [ ] https://flathub.org/apps/com.ktechpit.whatsie
- [ ] https://flathub.org/apps/com.github.IsmaelMartinez.teams_for_linux
- [ ] https://flathub.org/apps/io.github.gamingdoom.Datcord
- [ ] https://flathub.org/apps/io.github.tdesktop_x64.TDesktop

bbhtt:
Issues:

```
https://github.com/flathub/io.github.mimbrero.WhatsAppDesktop/issues/6
```

lionirdeadman:
Maybe this would be worth adding to this list: https://flathub.org/apps/com.github.k4zmu2a.spacecadetpinball

bbhtt:
What's the issue with that one?

lionirdeadman:
It's a reverse-engineering Microsoft game using *Microsoft assets* https://en.wikipedia.org/wiki/Full_Tilt!_Pinball#3D_Pinball_for_Windows_%E2%80%93_Space_Cadet

bbhtt:
That's a different issue than what this is tracking. 

Considering the original MS game is abandonware, the assets are publicly available all over the internet and it's being packaged as extra-data I don't think it's an issue for us at the moment.

PR #5339, opened 2024-06-16
Add com.maxizamorano.photopea-no-ads

MaxiZamorano:
**Name:** Photopea no Ads

**Description:** In addition to removing the ads from Photopea, it also removes the box that the ads occupy so that the work area is not reduced.

### DIFF
--- a/com.maxizamorano.photopea-no-ads/com.maxizamorano.photopea-no-ads.yml
+++ b/com.maxizamorano.photopea-no-ads/com.maxizamorano.photopea-no-ads.yml
@@ -1,0 +1,41 @@
+app-id: com.maxizamorano.photopea-no-ads
+runtime: org.freedesktop.Platform
+runtime-version: '23.08'
+sdk: org.freedesktop.Sdk
+base: org.electronjs.Electron2.BaseApp
+base-version: "23.08"
+command: photopea_no_ads.sh
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node20
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --socket=wayland
+  - --share=network
+  - --filesystem=xdg-download
+  - --device=dri
+build-options:
+  append-path: /usr/lib/sdk/node20/bin
+  env:
+    NPM_CONFIG_LOGLEVEL: info
+modules:
+  - name: photopea_no_ads
+    buildsystem: simple
+    build-options:
+        env:
+          XDG_CACHE_HOME: /run/build/photopea/cache
+    build-commands:
+      - install -D -t /app/bin photopea_no_ads.sh
+      - mkdir /app/main
+      - tar -xzf photopea_no_ads_build.tar.gz -C /app/main
+      - install -Dm755 /app/main/resources/data/com.maxizamorano.photopea-no-ads.desktop -t /app/share/applications
+      - install -Dm644 /app/main/resources/data/com.maxizamorano.photopea-no-ads.metainfo.xml -t /app/share/metainfo
+      - install -Dm644 /app/main/resources/icons/photopea.png /app/share/icons/hicolor/512x512/apps/com.maxizamorano.photopea-no-ads.png
+    sources:
+      - type: file
+        url: https://github.com/MaxiZamorano/photopea-no-ads/releases/download/v1.0.0/photopea_no_ads_build.tar.gz
+        sha256: fb9c3bf9a551ae73f41ebdf5301c757d91571923826e54ee82b8ea1124284a1c
+      - type: script
+        dest-filename: photopea_no_ads.sh
+        commands:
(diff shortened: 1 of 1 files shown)


com.maxizamorano.photopea-no-ads/com.maxizamorano.photopea-no-ads.yml
```diff
@@ -0,0 +1,41 @@
...
+    build-commands:
+      - install -D -t /app/bin photopea_no_ads.sh
+      - mkdir /app/main
+      - tar -xzf photopea_no_ads_build.tar.gz -C /app/main
+      - install -Dm755 /app/main/resources/data/com.maxizamorano.photopea-no-ads.desktop -t /app/share/applications
+      - install -Dm644 /app/main/resources/data/com.maxizamorano.photopea-no-ads.metainfo.xml -t /app/share/metainfo
+      - install -Dm644 /app/main/resources/icons/photopea.png /app/share/icons/hicolor/512x512/apps/com.maxizamorano.photopea-no-ads.png
+    sources:
+      - type: file
+        url: https://github.com/MaxiZamorano/photopea-no-ads/releases/download/v1.0.0/photopea_no_ads_build.tar.gz
```
hfiguiere, line 36: how does that work on aarch64?

com.maxizamorano.photopea-no-ads/com.maxizamorano.photopea-no-ads.yml
```diff
@@ -0,0 +1,41 @@
...
+sdk: org.freedesktop.Sdk
+base: org.electronjs.Electron2.BaseApp
+base-version: "23.08"
+command: photopea_no_ads.sh
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node20
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --socket=wayland
```
hfiguiere, line 13: it's either wayland with fallback-x11 or x11 alone. not both.

bbhtt:
The icon being used here is an exact copy of the original icon used by [Photopea](https://www.photopea.com/) https://github.com/MaxiZamorano/photopea-no-ads/blob/main/icons/photopea.png

You cannot do that. You cannot also use someone else's brand name for your application, as such it cannot be named "Photopea no ads" https://github.com/MaxiZamorano/photopea-no-ads/blob/981e06566e6ba3f434655812b521230b1bc3874d/data/com.maxizamorano.photopea-no-ads.metainfo.xml#L5

See https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/quality-guidelines#no-trademark-violations

com.maxizamorano.photopea-no-ads/com.maxizamorano.photopea-no-ads.yml
bbhtt: This should be built from source.

MaxiZamorano: > ¿Cómo funciona eso en Aarch64?

I haven't tested it on aarch64, only x86_64.

MaxiZamorano:
> The icon being used here is an exact copy of the original icon used by [Photopea](https://www.photopea.com/) https://github.com/MaxiZamorano/photopea-no-ads/blob/main/icons/photopea.png
> 
> You cannot do that. You cannot also use someone else's brand name for your application, as such it cannot be named "Photopea no ads" https://github.com/MaxiZamorano/photopea-no-ads/blob/981e06566e6ba3f434655812b521230b1bc3874d/data/com.maxizamorano.photopea-no-ads.metainfo.xml#L5
> 
> See https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/quality-guidelines#no-trademark-violations
> 
> I'm also a bit concerned whether this is acceptable for Flathub or not (but we'll have a discussion about that later)

There is currently a Photopea app on Flathub but it doesn't block ads, it also uses a Photopea icon and the Photopea name, what's the difference? I also see other apps on Flathub with brand names and icons.

bbhtt:
Two wrongs don't make a right. We reserve the right to remove them from the store at any point.
The issue is being tracked https://github.com/flathub/flathub/issues/5228.

com.maxizamorano.photopea-no-ads/com.maxizamorano.photopea-no-ads.yml
hfiguiere: but an aarch64 package is still built with these x86_64 binaries. Why not building from source?

MaxiZamorano:
> Two wrongs don't make a right. We reserve the right to remove them from the store at any point. The issue is being tracked #5228.

I understand, so I should modify icon, name and description? ID too?

bbhtt:
Name and icon, yes. The description should describe your app, not Photopea.

PR #5347, opened 2024-06-19
Add io.qt.QtDesignStudio

marbub:
Qt Design Studio is a graphical editor for Qml UI files. Qml files are the 'new' way to create UI interfaces with Qt. It shares some code with Qt Creator but diverges in UI and scope.

It still uses Qt 6.5. There are known issues with later Qt versions. Some Qt modules are still missing in the KDE runtime. We can add them later.

  We add host permissions because it is common to have Qml modules installed outside the home.

### DIFF
--- a/0001-Add-Design-Studio-for-Linux.patch
+++ b/0001-Add-Design-Studio-for-Linux.patch
@@ -152,7 +152,7 @@ index 00000000000..52f4ea27431
 +<?xml version="1.0" encoding="UTF-8"?>
 +<!-- Copyright (C) 2016 The Qt Company Ltd. Contact: https://www.qt.io/licensing/ -->
 +<component type="desktop">
-+ <id>io.qt.qtdesignstudio.desktop</id>
++ <id>io.qt.qtdesignstudio</id>
 + <name>Qt Design Studio</name>
 + <developer id="Qt Design Studio team">
 +   <name>Qt Project</name>
--- a/0001-Add-Design-Studio-for-Linux.patch
+++ b/0001-Add-Design-Studio-for-Linux.patch
@@ -1,0 +1,240 @@
+From 20e2823e89d3998498a46a251debdcd4b3ed3d32 Mon Sep 17 00:00:00 2001
+From: Marco Bubke <marco.bubke@qt.io>
+Date: Mon, 17 Jun 2024 14:52:53 +0200
+Subject: [PATCH] Add Design Studio for Linux
+
+Add AppStream data and desktop file for Design Studio.
+
+Change-Id: Icc5cca03a3c0945a2f90b39fdd4c7dcecd0e8d6b
+---
+ cmake/QtCreatorIDEBranding.cmake              |   2 +-
+ ...eator.png => QtProject-qtdesignstudio.png} | Bin
+ ...eator.png => QtProject-qtdesignstudio.png} | Bin
+ ...eator.png => QtProject-qtdesignstudio.png} | Bin
+ ...eator.png => QtProject-qtdesignstudio.png} | Bin
+ ...eator.png => QtProject-qtdesignstudio.png} | Bin
+ ...eator.png => QtProject-qtdesignstudio.png} | Bin
+ ...eator.png => QtProject-qtdesignstudio.png} | Bin
+ ...eator.png => QtProject-qtdesignstudio.png} | Bin
+ share/CMakeLists.txt                          |  10 +--
+ .../org.qt-project.qtdesignstudio.desktop     |  11 ++++
+ ...project.qtdesignstudio.appdata.xml.cmakein |  58 ++++++++++++++++++
+ src/app/CMakeLists.txt                        |   6 +-
+ src/app/app_logo.qrc.cmakein                  |   4 +-
+ src/app/main.cpp                              |   2 +-
+ 15 files changed, 82 insertions(+), 11 deletions(-)
+ rename dist/branding/qtdesignstudio/images/logo/128/{QtProject-qtcreator.png => QtProject-qtdesignstudio.png} (100%)
+ rename dist/branding/qtdesignstudio/images/logo/16/{QtProject-qtcreator.png => QtProject-qtdesignstudio.png} (100%)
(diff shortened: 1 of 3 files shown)


io.qt.QtDesignStudio.yaml
```diff
@@ -0,0 +1,51 @@
+app-id: io.qt.QtDesignStudio
+runtime: org.kde.Sdk
+runtime-version: '6.5'
+base: io.qt.qtwebengine.BaseApp
+base-version: '6.5'
+sdk: org.kde.Sdk
+command: qtdesignstudio
+rename-desktop-file: org.qt-project.qtdesignstudio.desktop
```
hfiguiere, line 8: I don't see why you would not use this app-id.

io.qt.QtDesignStudio.yaml
```diff
@@ -0,0 +1,51 @@
...
+          type: anitya
+          project-id: 4136
+          tag-template: v$version
+          stable-only: true
+      - type: patch
+        path: 0001-Add-Design-Studio-for-Linux.patch
+  - name: qtquickdesigner-components
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_PREFIX_PATH=/app
```
hfiguiere, line 46:
```suggestion
```

io.qt.QtDesignStudio.yaml
```diff
@@ -0,0 +1,51 @@
...
+  - --filesystem=host
+  - --device=dri
+  - --allow=devel
+  - --env=QTWEBENGINEPROCESS_PATH=/app/bin/QtWebEngineProcess
+  - --env=QML_IMPORT_PATH=/app/qml
+modules:
+  - name: qt-creator
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_PREFIX_PATH=/app
```
hfiguiere, line 25:
```suggestion
```

io.qt.QtDesignStudio.yaml
```diff
@@ -0,0 +1,51 @@
+app-id: io.qt.QtDesignStudio
+runtime: org.kde.Sdk
+runtime-version: '6.5'
```
hfiguiere, line 3: this runtime version is EOL. So either this has to use a more recent runtime, or this submission has to wait to be fixed.

marbub: Hmm, 6.7 is broken, what about 6.6?

bbhtt: How is it broken?

io.qt.QtDesignStudio.yaml
bbhtt: Seems resolved without any answer. If the app internally uses that id, the flatpak should use the same.

marbub: Because flatpak bilder is complaining about the '-'.

bbhtt: You demangle it dashes should be converted to underscore.

io.qt.QtDesignStudio.yaml
marbub: Behavior changes. Qt is not behavior-compatible. They are small but lead to a subpar experience. Actually, very complex software will suffer from it, because it is really hard to change something without breaking something: https://www.hyrumslaw.com/

io.qt.QtDesignStudio.yaml
marbub: Yes, but flatpak builder was not working.

marbub: error: 'org.qt-project.qtdesignstudio' is not a valid application name: Only last name segment can contain -

bbhtt: That's what I meant it should've been: `org.qt_project.qtdesignstudio`

io.qt.QtDesignStudio.yaml
bbhtt: 6.6 is going to be EOL ed 2 months.

io.qt.QtDesignStudio.yaml
marbub: But that is not the domain name, so I chose the other domain name that is under our control.

bbhtt: Dashes in the id are to be converted to underscore that's the rule for application ids. When converting them back to domain name, implementations reverse it.

See the appstream specification for id tag.

marbub: Oh, I love it. ;-)

So I abort this request and start a new one?

marbub: I changed the app id upstream.

0001-Add-Design-Studio-for-Linux.patch
```diff
@@ -0,0 +1,262 @@
...
+diff --git a/share/metainfo/io.qt.qtdesignstudio.appdata.xml.cmakein b/share/metainfo/io.qt.qtdesignstudio.appdata.xml.cmakein
+new file mode 100644
+index 00000000000..52f4ea27431
+--- /dev/null
++++ b/share/metainfo/io.qt.qtdesignstudio.appdata.xml.cmakein
+@@ -0,0 +1,58 @@
++<?xml version="1.0" encoding="UTF-8"?>
++<!-- Copyright (C) 2016 The Qt Company Ltd. Contact: https://www.qt.io/licensing/ -->
++<component type="desktop">
++ <id>io.qt.qtdesignstudio.desktop</id>
```
hfiguiere, line 155: an id that ends in `.desktop` isn't valid.

marbub:
bot, build io.qt.qtdesignstudio

marbub:
bot, build io.qt.qtdesignstudio

marbub:
Hi, I want to verify the application. But it seems I am not the author. Could I get the author role?

bbhtt:
image


You already have write access which is what you need. Then login with GitHub, if the app is not appearing hit refresh a bunch of times and it should appear.

image

marbub:
I mean, I can only see Qt Creator on https://flathub.org/developer-portal. So, I cannot verify Qt Design Studio.

marbub:
Now I can see it. Thanks!

PR #5370, opened 2024-06-30
Add com.alpineclient.AlpineClient

xEricL:

Alpine Client is an all-in-one modpack for Minecraft that offers a multitude of enhancements and optimizations to improve your gameplay. It brings together popular mods, exclusive features, player cosmetics, and multi-version support to curate the ultimate player experience.

- [x] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about this submission. **Link:** https://github.com/alpine-client/pinnacle


com.alpineclient.AlpineClient.metainfo.xml
```diff
@@ -0,0 +1,82 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+    <id>com.alpineclient.AlpineClient</id>
+
+    <name>Alpine Client</name>
+    <summary>A lightweight Minecraft mod-pack</summary>
+    <icon type="remote">https://api.alpineclientprod.com/uploads/dark_diamond_logo_256_15af794bce.png</icon>
```
bbhtt, line 7:
```suggestion
```

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "skip-arches": ["aarch64"]
```
bbhtt, line 2: Why not build from source?

com.alpineclient.AlpineClient.desktop
bbhtt, line 1: This should be upstream

com.alpineclient.AlpineClient.png
bbhtt, line 1: This should be upstream

com.alpineclient.AlpineClient.metainfo.xml
bbhtt, line 1: This should be upstream

flathub.json
hfiguiere: otherwise this should be "only-arches" since you are not merely excluding aarch64 but actually acknowledging one arch exist.

com.alpineclient.AlpineClient.yml
```diff
@@ -0,0 +1,40 @@
...
+        sha256: "df15106b56cefc91a7daf61313e36c97ee149e17ef7199673dfbbeb76a8a45f6"
+        dest-filename: "alpine-client"
+      - type: file
+        path: "com.alpineclient.AlpineClient.desktop"
+      - type: file
+        path: "com.alpineclient.AlpineClient.png"
+      - type: file
+        path: "com.alpineclient.AlpineClient.metainfo.xml"
+
+  - name: xrandr
```
hfiguiere, line 36: this should be build before the main module since it's a dependency. that's the logicial order.

xEricL: I will fix this.

flathub.json
xEricL: Good idea. I will fix this.

xEricL: > Why not build from source?

It seemed easier to just download the binary from the GitHub repo instead of making the build script download GoLang and compile it from source. Perhaps building from source would be better. I will update the manifest now.

bbhtt: We'd prefer if you can build from source https://docs.flathub.org/docs/for-app-authors/requirements#best-practices

xEricL: done.

xEricL:
All comments have been addressed. Please let me know if there's anything else I need to do.

flathub.json
bbhtt, line 1: Is this still needed? You can support both arches now?

bbhtt:
bot, build com.alpineclient.AlpineClient

com.alpineclient.AlpineClient.yml
```diff
@@ -0,0 +1,33 @@
...
+    sources:
+      - type: archive
+        url: "https://xorg.freedesktop.org/archive/individual/app/xrandr-1.5.1.tar.xz"
+        sha256: "7bc76daf9d72f8aff885efad04ce06b90488a1a169d118dea8a2b661832e8762"
+
+  - name: alpineclient
+    buildsystem: simple
+    build-commands:
+      - "install -D -m 755 alpine-client /app/bin/alpine-client"
```
hfiguiere, line 26: this file isn't found. maybe you actually forgot to build it?

xEricL: I'm sorry, I somehow forgot to commit the rest of my changes. Should be fixed now.

com.alpineclient.AlpineClient.yml
```diff
@@ -0,0 +1,43 @@
+id: com.alpineclient.AlpineClient
+runtime: org.freedesktop.Platform
+runtime-version: "23.08"
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.golang
+command: alpine-client
+finish-args:
+  - "--persist=.alpineclient"
```
hfiguiere, line 9: this would be so much better if it used standard xdg directories. Sandboxing or not.

flathub.json
xEricL: True. Fixed.

com.alpineclient.AlpineClient.yml
xEricL: The program currently uses `$HOME/.alpineclient`. Would you prefer it be `$XDG_DATA_HOME/.alpineclient`?

hfiguiere: this is supposed to be the rules for a while now. (it's not flatpak specific, quite the opposite)

xEricL: That is fine. I wasn't disputing it; I just wanted confirmation. Should I switch it to `$XDG_DATA_HOME/.alpineclient`? If so, do I remove `--persist=.alpineclient`? or change it to  `--persist=$XDG_DATA_HOME/.alpineclient`?

xEricL: I would prefer not to use $XDG_DATA_HOME. I'm on Fedora and the environmental variable isn't set on my system. [Minecraft put's their data in `$HOME/.minecraft`](https://github.com/flathub/com.mojang.Minecraft/blob/5551f63926c899b3e6794adf0b12cd8167340926/com.mojang.Minecraft.json#L12), so we followed and put our data in `$HOME/.alpineclient`.

hfiguiere: What Minecraft is doing is irrelevant.

Toolkits have functions to return these directories (ie implement the spec) and that what you should use.

hfiguiere: Even go seems to have packages to do it

https://github.com/adrg/xdg

(just a random search, not a recommendation)

xEricL: > What Minecraft is doing is irrelevant.

While you may not care, what Minecraft does is not irrelevant to us or our users. We are a Minecraft modpack. Our users expect to find our data directory in the same place that they have found the Minecraft data directory for over 13 years.

We are not alone in this. All other major Minecraft modpacks do it, even in their Flathub distributions:
- [Lunar Client](https://github.com/flathub/com.lunarclient.LunarClient/blob/master/com.lunarclient.LunarClient.yml#L12) uses `$HOME/.lunarclient`
- [LabyMod](https://github.com/flathub/net.labymod.LabyMod/blob/master/net.labymod.LabyMod.yml#L16) uses `$HOME/.minecraft`

> this is supposed to be the rules for a while now. (it's not flatpak specific, quite the opposite)

We understand the XDG Base Directory specification. We also understand that [we are far from the only application to not follow it](https://wiki.archlinux.org/title/XDG_Base_Directory#Partial). 

What is "supposed" to be a rule is of no concern to us. Given the above reasoning, if it is not a mandatory requirement then we are absolutely opposed to changing it.

hfiguiere: two wrongs don't make a right. They are doing it wrong, no reason to do the same.

tomwmth: Hey, one of the core developers here.

I'm sure you're familiar with the Linus Torvalds' quote ["If it's a bug that people rely on, it's not a bug - it's a feature"](https://youtu.be/Pzl1B7nB9Kc?t=271). While this is not a bug per se, I think it is pretty self explanatory how it applies here.

Users rely on this data directory behaviour to interface with Minecraft. As you likely know, Minecraft is a game where the average user is expected to tinker with the data directory. Through an over decade old decision that we had nothing to do with, users expect our data directory to be in `$HOME`. This has nothing to do with what we want or trying to make a right. No specification is going to change the fact that when a user goes to look for our data directory they will look in `$HOME`.

You reviewed both the Lunar Client and LabyMod pull requests. You did not raise this as an issue for either of them. You should understand that what we have done makes sense for this specific type of application.

As previously stated: unless this is a defined and enforced rule, we have no desire to change it.

bbhtt:
bor, build com.alpineclient.AlpineClient

xEricL:
> bor, build com.alpineclient.AlpineClient

You made a typo

bbhtt:
oh huh didn't see that. I started the build manually anyway here https://buildbot.flathub.org/#/builders/6/builds/131724

com.alpineclient.AlpineClient.yml
bbhtt: > We also understand that [we are far from the only application to not follow it](https://wiki.archlinux.org/title/XDG_Base_Directory#Partial).

Everyone else doing it is either a relic from the pre XDG days or someone doing it wrong.

It's a bit sad that current applications don't want to set the trend here of following the standard after it has been so widely adopted.

> quote ["If it's a bug that people rely on, it's not a bug - it's a feature"](https://youtu.be/Pzl1B7nB9Kc?t=271).

If there was a proper UI/UX and instructions for opening the data directory maybe people won't rely on antiquated behaviour.

Anyway this doesn't outright block the submission.

tomwmth: I understand your point of view. I agree that following the specification is what applications should do. If I was writing an application from scratch, I would follow it. The issue here is that we are not the application, merely a layer on top of another. We are aiming to feel like a natural extension of Minecraft. Sometimes this entails replicating their bad decisions for the sake of familiarity.

In matters like this, our users expect us to work the same way as the underlying application. They expect that things will be in the same reliable location that they have had memorized for years. I know this is primarily a Linux discussion but Minecraft is solely responsible for a majority of Windows users who know about the AppData directory and how to navigate there. It is *very* normal for a Minecraft player to poke around in the data directory.

Once again, I agree with you in principle. If tomorrow Mojang decided to start following the XDG Base Directory specification, we would the first to follow suit. Unfortunately in this case we vehemently believe that is in the best interest of ourselves and our users to ignore the specification.

Thank you for your understanding.

PR #5376, opened 2024-07-01
Add io.ossia.score

jcelerier:

- [X] Please describe the application briefly:

ossia score is a sequencer for audio-visual artists, designed to enable the creation of interactive shows, museum installations, intermedia digital artworks, interactive music and more in an intuitive user interface.
 
- [X] I am an author/developer/upstream contributor of the project.  **Link: https://ossia.io / https://github.com/ossia/score **

### DIFF
--- a/io.ossia.score.yml
+++ b/io.ossia.score.yml
@@ -2,6 +2,11 @@ id: io.ossia.score
 runtime: org.kde.Platform
 runtime-version: '6.7'
 sdk: org.kde.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.llvm18
+build-options:
+  env:
+    CXX: /usr/lib/sdk/llvm18/bin/clang++
 command: ossia-score
 rename-desktop-file: ossia-score.desktop
 rename-icon: ossia-score
--- a/io.ossia.score.yml
+++ b/io.ossia.score.yml
@@ -244,6 +244,18 @@ modules:
         url: https://github.com/avahi/avahi/releases/download/v0.8/avahi-0.8.tar.gz
         sha256: 060309d7a333d38d951bc27598c677af1796934dbd98e1024e7ad8de798fedda
 
+  # suil, necessary for LV2 availability
+  - name: suil
+    builddir: true
+    buildsystem: meson
+    config-opts:
+      - --prefix=/app
+
+    sources:
+      - type: archive
+        url: http://download.drobilla.net/suil-0.10.20.tar.xz
+        sha256: 334a3ed3e73d5e17ff400b3db9801f63809155b0faa8b1b9046f9dd3ffef934e
+
   - name: ossia-score
     buildsystem: cmake-ninja
     builddir: true
--- a/io.ossia.score.yml
+++ b/io.ossia.score.yml
@@ -10,7 +10,7 @@ build-options:
 command: ossia-score
 rename-desktop-file: ossia-score.desktop
 rename-icon: ossia-score
-rename-appdata-file: ossia-score
+rename-appdata-file: ossia-score.appdata.xml
 finish-args:
   # BLE support
   - --allow=bluetooth
--- a/io.ossia.score.yml
+++ b/io.ossia.score.yml
@@ -5,6 +5,7 @@ sdk: org.kde.Sdk
 command: ossia-score
 rename-desktop-file: ossia-score.desktop
 rename-icon: ossia-score
+rename-appdata-file: ossia-score
 finish-args:
   # BLE support
   - --allow=bluetooth
--- a/io.ossia.score.yml
+++ b/io.ossia.score.yml
@@ -274,4 +274,5 @@ modules:
      - install -Dm755 "${FLATPAK_DEST}/bin/ossia-score" "${FLATPAK_DEST}/bin/ossia-score.bin"
      - install -Dm755 ../run "${FLATPAK_DEST}/bin/ossia-score"
      - install -d /app/extensions/Plugins
+     - install -d /app/lib/ffmpeg
 
--- a/io.ossia.score.yml
+++ b/io.ossia.score.yml
@@ -260,8 +260,8 @@ modules:
     builddir: true
     sources:
         - type: archive
-          url: https://github.com/ossia/score/releases/download/v3.2.3-3/ossia.score-3.2.3-3-src.tar.xz
-          sha256: c66924e6490aafa5239310600149222b6288a977269592ff96d9491588e0784c
+          url: https://github.com/ossia/score/releases/download/v3.2.4/ossia.score-3.2.4-src.tar.xz
+          sha256: 78978021827dc7fe6450ceb5799bc5df23a6a228057e67f1dc33d3a4bee95963
         - type: script
           dest-filename: run
           commands:
--- a/io.ossia.score.yml
+++ b/io.ossia.score.yml
@@ -246,10 +246,7 @@ modules:
 
   # suil, necessary for LV2 availability
   - name: suil
-    builddir: true
     buildsystem: meson
-    config-opts:
-      - --prefix=/app
 
     sources:
       - type: archive
--- a/io.ossia.score.yml
+++ b/io.ossia.score.yml
@@ -273,4 +273,5 @@ modules:
     post-install:      
      - install -Dm755 "${FLATPAK_DEST}/bin/ossia-score" "${FLATPAK_DEST}/bin/ossia-score.bin"
      - install -Dm755 ../run "${FLATPAK_DEST}/bin/ossia-score"
+     - install -d /app/extensions/Plugins
 
--- a/io.ossia.score.yml
+++ b/io.ossia.score.yml
@@ -1,0 +1,270 @@
+id: io.ossia.score
+runtime: org.kde.Platform
+runtime-version: '6.7'
+sdk: org.kde.Sdk
+command: ossia-score
+rename-desktop-file: ossia-score.desktop
+rename-icon: ossia-score
+finish-args:
+  # BLE support
+  - --allow=bluetooth
+
+  # Access user's data and media libraries, plug-ins etc
+  - --filesystem=host
+
+  # Required for USB, Webcam devices support as well as raw GPIO I/O & GPU rendering
+  - --device=all
+
+  # X11 is used by default
+  - --socket=x11
+
+  # Used by X11
+  - --share=ipc
+
+  # Network access
+  - --share=network
+
+  # PulseAudio is needed for ALSA device access
+  - --socket=pulseaudio
+
+  # Audio support through pipewire
+  - --filesystem=xdg-run/pipewire-0
+
+  # Plugins
+  - --env=VST_PATH=/app/extensions/Plugins/vst
+  - --env=VST3_PATH=/app/extensions/Plugins/vst3
+
+  # Required to allow screensaver/idle inhibition
+  - --talk-name=org.freedesktop.ScreenSaver
+
+add-extensions:
+  # Linux audio plugins base extension
+  org.freedesktop.LinuxAudio.Plugins:
+     directory: extensions/Plugins
+     version: '23.08'
+     add-ld-path: lib
+     merge-dirs: vst;vst3;lv2
+     subdirectories: true
+     no-autodownload: true
+
+  # Full codec support for FFMPEG
+  org.freedesktop.Platform.ffmpeg-full:
+    directory: lib/ffmpeg
+    add-ld-path: .
+    version: "23.08"
+
+cleanup:
+  - /bin/avahi*
+  - /include
+  - '*.a'
+  - '*.la'
+  - /lib/cmake
+  - /lib/girepository-1.0
+  - /lib/pkgconfig
+  - /lib/avahi
+  - /lib/python*
+  - /share/gir-1.0
+  - /share/man
+  - /share/pkgconfig
+  - /share/aclocal
+  - /share/gtk-doc
+  - /share/doc
+  - /share/info
+  - /share/pixmaps
+  - /share/async-mqtt5
+  - /man
+
+modules:
+  # LV2 audio plugin support
+  - shared-modules/linux-audio/lv2.json
+  - shared-modules/linux-audio/lilv.json
+
+  # Audio I/O
+  - name: portaudio
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+      - -DCMAKE_POSITION_INDEPENDENT_CODE=ON
+      - -DBUILD_SHARED_LIBS=ON
+      - -DPA_ALSA_DYNAMIC=ON
+      - -DPA_USE_JACK=OFF
+    sources:
+      - type: archive
+        url: https://github.com/PortAudio/portaudio/archive/refs/tags/v19.7.0.tar.gz
+        sha256: 5af29ba58bbdbb7bbcefaaecc77ec8fc413f0db6f4c4e286c40c3e1b83174fa0
+
+  # Pitch shifting and resampling
+  - name: libsamplerate
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+      - -DCMAKE_POSITION_INDEPENDENT_CODE=ON
+      - -DBUILD_SHARED_LIBS=ON
+      - -DBUILD_TESTING=0
+    sources:
+      - type: archive
+        url: https://github.com/libsndfile/libsamplerate/archive/4858fb016550d677de2356486bcceda5aed85a72.zip
+        sha256: 11af7a8243dfb81e211ed140139d3439b39ffc7527b92d5c5c4a5e7338f595ee
+
+  # Time-stretch support
+  - name: rubberband
+    buildsystem: meson
+    config-opts:
+      - -Ddefault_library=shared
+      - -Dbuildtype=release
+      - -Dtests=disabled
+      - -Dcmdline=disabled
+      - -Dladspa=disabled
+      - -Dlv2=disabled
+      - -Dfft=fftw
+      - -Dresampler=libsamplerate
+    sources:
+      - type: archive
+        url: https://github.com/breakfastquay/rubberband/archive/refs/tags/v3.3.0.tar.gz
+        sha256: 2bb837fe00932442ca90e185af8a468f7591df0c002b4a9e27a1bced1563ac84
+
+  # Used for fuzzy-matching of various objects in text fields, etc.
+  - name: rapidfuzz-cpp
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+      - -DCMAKE_POSITION_INDEPENDENT_CODE=ON
+      - -DBUILD_SHARED_LIBS=ON
+    sources:
+      - type: archive
+        url: https://github.com/rapidfuzz/rapidfuzz-cpp/archive/refs/tags/v3.0.4.tar.gz
+        sha256: 18d1c41575ceddd6308587da8befc98c85d3b5bc2179d418daffed6d46b8cb0a
+
+  # Dependency of re2
+  - name: abseil-cpp
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+      - -DCMAKE_POSITION_INDEPENDENT_CODE=ON
+      - -DBUILD_SHARED_LIBS=ON
+    sources:
+      - type: archive
+        url: https://github.com/abseil/abseil-cpp/releases/download/20240116.2/abseil-cpp-20240116.2.tar.gz
+        sha256: 733726b8c3a6d39a4120d7e45ea8b41a434cdacde401cba500f14236c49b39dc
+
+  # Regex engine used within ossia
+  - name: re2
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+      - -DCMAKE_POSITION_INDEPENDENT_CODE=ON
+      - -DBUILD_SHARED_LIBS=ON
+    sources:
+      - type: archive
+        url: https://github.com/google/re2/archive/refs/tags/2023-06-01.tar.gz
+        sha256: 8b4a8175da7205df2ad02e405a950a02eaa3e3e0840947cd598e92dca453199b
+
+  # Needed for CoAP protocol support
+  - name: libcoap
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+      - -DCMAKE_POSITION_INDEPENDENT_CODE=ON
+      - -DBUILD_SHARED_LIBS=ON
+      - -DENABLE_WS=1
+      - -DENABLE_DTLS=0
+      - -DENABLE_EXAMPLES=0
+      - -DENABLE_DOCS=0
+      - -DENABLE_OSCORE=0
+      - -DENABLE_SERVER_MODE=0
+      - -DMAX_LOGGING_LEVEL=0
+    sources:
+      - type: archive
+        url: https://github.com/obgm/libcoap/archive/refs/tags/v4.3.4a.tar.gz
+        sha256: 45f1aabbb5f710e841c91d65fc3f37c906d42e8fc44dd04979e767d3960a77cf
+
+  # ossia uses boost
+  - name: boost
+    buildsystem: simple
+    build-commands:
+      - mkdir -p "${FLATPAK_DEST}/include/"
+      - cp -rf boost "${FLATPAK_DEST}/include/"
+    sources:
+      - type: archive
+        url: https://archives.boost.io/release/1.85.0/source/boost_1_85_0.tar.bz2
+        sha256: 7009fe1faa1697476bdc7027703a2badb84e849b7b0baad5086b087b971f8617
+
+  # HAP video decoding
+  - name: snappy
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+      - -DCMAKE_POSITION_INDEPENDENT_CODE=ON
+      - -DBUILD_SHARED_LIBS=ON
+      - -DSNAPPY_BUILD_TESTS=0
+      - -DSNAPPY_BUILD_BENCHMARKS=0
+    sources:
+      - type: archive
+        url: https://github.com/google/snappy/archive/refs/tags/1.2.1.tar.gz
+        sha256: 736aeb64d86566d2236ddffa2865ee5d7a82d26c9016b36218fcc27ea4f09f86
+
+  # ossia uses mDNS for multiple protocol discovery mechanisms
+  - name: avahi
+    builddir: true
+    config-opts:
+      - --with-distro=none
+      - --disable-qt3
+      - --disable-qt4
+      - --disable-qt5
+      - --disable-gtk
+      - --disable-gtk3
+      - --disable-glib
+      - --disable-gobject
+      - --disable-python
+      - --disable-mono
+      - --disable-monodoc
+      - --disable-libevent
+      - --disable-libdaemon
+      - --disable-dbm
+      - --disable-gdbm
+      - --disable-manpage
+
+    post-install:
+      - make -C avahi-common install
+      - make -C avahi-client install
+
+    sources:
+      - type: archive
+        url: https://github.com/avahi/avahi/releases/download/v0.8/avahi-0.8.tar.gz
+        sha256: 060309d7a333d38d951bc27598c677af1796934dbd98e1024e7ad8de798fedda
+
+  - name: ossia-score
+    buildsystem: cmake-ninja
+    builddir: true
+    sources:
+        - type: archive
+          url: https://github.com/ossia/score/releases/download/v3.2.3-3/ossia.score-3.2.3-3-src.tar.xz
+          sha256: c66924e6490aafa5239310600149222b6288a977269592ff96d9491588e0784c
+        - type: script
+          dest-filename: run
+          commands:
+            - export LV2_PATH=$HOME/.lv2:/app/extensions/Plugins/lv2:/app/lib/lv2
+            - exec ossia-score.bin "$@"
+
+    config-opts:
+      # Note: don't use RelWithDebInfo as ossia release builds explicitly expect Release build type.
+      - -DCMAKE_BUILD_TYPE=Release
+      - -Wno-dev
+      - -DSCORE_DEPLOYMENT_BUILD=1
+      - -DSCORE_STATIC_PLUGINS=1
+      - -DSCORE_FHS_BUILD=1
+      - -DCMAKE_UNITY_BUILD=1
+      - -DCMAKE_SKIP_RPATH=ON
+      - -DOSSIA_USE_SYSTEM_LIBRARIES=1
+      - -DSCORE_USE_SYSTEM_LIBRARIES=1
+      - -DSCORE_NO_QTCONF=1
+
+    post-install:      
+     - install -Dm755 "${FLATPAK_DEST}/bin/ossia-score" "${FLATPAK_DEST}/bin/ossia-score.bin"
+     - install -Dm755 ../run "${FLATPAK_DEST}/bin/ossia-score"
+
--- a/io.ossia.score.yml
+++ b/io.ossia.score.yml
@@ -252,6 +252,8 @@ modules:
       - type: archive
         url: http://download.drobilla.net/suil-0.10.20.tar.xz
         sha256: 334a3ed3e73d5e17ff400b3db9801f63809155b0faa8b1b9046f9dd3ffef934e
+      - type: patch
+        path: 0001-Add-an-x11_in_qt6-wrapper.patch
 
   - name: ossia-score
     buildsystem: cmake-ninja
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "shared-modules"]
+	path = shared-modules
+	url = https://github.com/flathub/shared-modules
--- a/0001-Add-an-x11_in_qt6-wrapper.patch
+++ b/0001-Add-an-x11_in_qt6-wrapper.patch
@@ -1,0 +1,340 @@
+From 2a0e0b7b3c23022c8302e9cfc8cd5ab04a296702 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Jean-Micha=C3=ABl=20Celerier?=
+ <jeanmichael.celerier@gmail.com>
+Date: Mon, 8 Jul 2024 05:17:23 -0400
+Subject: [PATCH] Add an x11_in_qt6 wrapper
+
+---
+ meson.build        |  21 +++++
+ meson_options.txt  |   3 +
+ src/instance.c     |  12 ++-
+ src/x11_in_qt6.cpp | 227 +++++++++++++++++++++++++++++++++++++++++++++
+ 4 files changed, 262 insertions(+), 1 deletion(-)
+ create mode 100644 src/x11_in_qt6.cpp
+
+diff --git a/meson.build b/meson.build
+index 3089ce8..684701a 100644
+--- a/meson.build
++++ b/meson.build
+@@ -124,6 +124,14 @@ qt5_x11_dep = dependency(
+   required: get_option('qt5').enabled() and get_option('x11').enabled(),
+ )
+ 
++qt6_dep = dependency(
++  'Qt6Widgets',
++  include_type: 'system',
++  version: '>=6.2.0',
++  required: get_option('qt6'),
++)
++
++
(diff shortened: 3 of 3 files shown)


jcelerier:
- Link to CI-built flatpak: https://github.com/ossia/score/actions/runs/9748167859

io.ossia.score.yml
```diff
@@ -0,0 +1,333 @@
...
+  - --allow=bluetooth
+
+  # Access user's data and media libraries, plug-ins etc
+  - --filesystem=host
+
+  # GPU rendering support
+  - --device=dri
+
+  # Raw input support
+  - --device=input
```
hfiguiere, line 17: this is NOT supported. It will fail on virtually any distro as there is no stable version of flatpak that has it. Also it's redundant with ALL.

io.ossia.score.yml
```diff
@@ -0,0 +1,333 @@
...
+
+  # Required for USB & MIDI devices support as well as JACK (/dev/shm) and raw GPIO I/O
+  - --device=all
```
hfiguiere, line 20: MIDI devices are exposed through the pulseaudio socket permission. The comment is incorrect.

io.ossia.score.yml
```diff
@@ -0,0 +1,333 @@
...
+
+  # Removable drives support
+  # https://github.com/flatpak/flatpak/issues/2713#issuecomment-1958510347
+  - --filesystem=/media
+  - --filesystem=/run/media
```
hfiguiere, line 25: given that you added `host`, these permissions are redundant.

io.ossia.score.yml
```diff
@@ -0,0 +1,333 @@
...
+
+  # X11 is used by default
+  - --socket=x11
+
+  # Wayland still has rendering regressions compared to X11 as of Qt 6.7
+  - --socket=wayland
```
hfiguiere, line 31: it is either or. Not both. If wayland is broken, don't add it.

io.ossia.score.yml
```diff
@@ -0,0 +1,333 @@
...
+  # Permit PipeWire and user music directory access
+  - --filesystem=xdg-run/pipewire-0
+  - --filesystem=xdg-music
+
+  # PulseAudio is needed for ALSA device access
+  - --socket=pulseaudio
+
+  # Talk to UPower and RTKit services on the session bus
+  - --system-talk-name=org.freedesktop.UPower
+  - --system-talk-name=org.freedesktop.RealtimeKit1
```
hfiguiere, line 48: the RealtimeKit will NOT work from inside the sandbox.

io.ossia.score.yml
```diff
@@ -0,0 +1,333 @@
...
+  - --env=VST3_PATH=/app/extensions/Plugins/vst3
+
+  # Required to allow screensaver/idle inhibition
+  - --talk-name=org.freedesktop.ScreenSaver
+
+  # For correct cursor scaling under Wayland
+  - --env=XCURSOR_PATH=/run/host/user-share/icons:/run/host/share/icons
+
+  # Plugins
+  - --env=LV2_PATH=$HOME/.lv2:/app/extensions/Plugins/lv2:/app/lib/lv2
```
hfiguiere, line 62: this doesn't work. `$HOME` is NOT substituted. You have to do that in a wrapper script.

io.ossia.score.yml
```diff
@@ -0,0 +1,333 @@
...
+
+  # Config persistence
+  - --persist=ossia
```
hfiguiere, line 65: it writes `~/ossia` ? This should be fixed upstream. This is ugly.

io.ossia.score.yml
```diff
@@ -0,0 +1,333 @@
...
+      - -DCMAKE_POSITION_INDEPENDENT_CODE=ON
+      - -DBUILD_SHARED_LIBS=OFF
+      - -DBUILD_TESTING=0
+    sources:
+      - type: archive
+        url: https://github.com/libsndfile/libsamplerate/archive/4858fb016550d677de2356486bcceda5aed85a72.zip
+        sha256: 11af7a8243dfb81e211ed140139d3439b39ffc7527b92d5c5c4a5e7338f595ee
+
+  # Used for some audio file formats not supported by FFMPEG, reading ACID tags, etc.
+  - name: libsndfile
```
hfiguiere, line 139: libsndfile is in the runtime.

```
$ pkg-config --modversion sndfile
1.2.0
```

io.ossia.score.yml
```diff
@@ -0,0 +1,333 @@
...
+      - -DENABLE_MPEG=0
+      - -DBUILD_REGTEST=0
+      - -DBUILD_TESTING=0
+    sources:
+      - type: archive
+        url: https://github.com/libsndfile/libsndfile/archive/58c05b87162264200b1aa7790be260fd74c9deee.zip
+        sha256: 3337d28c24f2aaa74b619c9c335db79fd4891d9dda5fa29e70b1c771c46ee342
+
+  # More efficient timestretch with rubberband
+  - name: fftw
```
hfiguiere, line 163: the `double` variant is in the runtieme. The `float` version please use the shared-modules.

io.ossia.score.yml
```diff
@@ -0,0 +1,333 @@
...
+      - -DSCORE_STATIC_PLUGINS=1
+      - -DSCORE_FHS_BUILD=1
+      - -DCMAKE_UNITY_BUILD=1
+      - -DCMAKE_SKIP_RPATH=ON
+      - -DOSSIA_USE_SYSTEM_LIBRARIES=1
+      - -DSCORE_USE_SYSTEM_LIBRARIES=1
+      - -DSCORE_NO_QTCONF=1
+
+    post-install:
+     - cmake -DCMAKE_INSTALL_DO_STRIP=1 -DCOMPONENT=OssiaScore -P cmake_install.cmake
```
hfiguiere, line 328: the install target isn't run by default?

io.ossia.score.yml
```diff
@@ -0,0 +1,333 @@
...
+     - install -Dm644 /run/build/ossia-score/cmake/Deployment/Linux/Flatpak/io.ossia.score.metainfo.xml /app/share/metainfo/io.ossia.score.metainfo.xml
```
hfiguiere, line 329:
```suggestion
     - install -Dm644 ../cmake/Deployment/Linux/Flatpak/io.ossia.score.metainfo.xml /app/share/metainfo/io.ossia.score.metainfo.xml
```

io.ossia.score.yml
```diff
@@ -0,0 +1,333 @@
...
+     - mv /app/share/applications/ossia-score.desktop /app/share/applications/io.ossia.score.desktop
```
hfiguiere, line 330: this should be `rename-desktop-file` in the manifest top-level.

io.ossia.score.yml
```diff
@@ -0,0 +1,333 @@
...
+     - sed -i 's/Icon=ossia-score/Icon=io.ossia.score/' /app/share/applications/io.ossia.score.desktop
+     - mkdir -p /app/share/icons/hicolor/512x512/apps
+     - mv /app/share/pixmaps/ossia-score.png /app/share/icons/hicolor/512x512/apps/io.ossia.score.png
```
hfiguiere, line 333:
```suggestion
     - install -Dm644  /app/share/pixmaps/ossia-score.png -t /app/share/icons/hicolor/512x512/apps
```

And then add `rename-icon` in the top-level of the manifest. Note that you could submit a patch to upstream to install the icon in the right location. (i.e. NOT pixmaps)

io.ossia.score.yml
```diff
@@ -0,0 +1,333 @@
...
+runtime: org.kde.Platform
+runtime-version: '6.7'
+sdk: org.kde.Sdk
+command: ossia-score
+finish-args:
+  # BLE support
+  - --allow=bluetooth
+
+  # Access user's data and media libraries, plug-ins etc
+  - --filesystem=host
```
hfiguiere, line 11: this is a very broad filesystem permission.

io.ossia.score.yml
```diff
@@ -0,0 +1,333 @@
...
+
+  # Used by X11
+  - --share=ipc
+
+  # Network access
+  - --share=network
+
+  # Permit PipeWire and user music directory access
+  - --filesystem=xdg-run/pipewire-0
+  - --filesystem=xdg-music
```
hfiguiere, line 41: xdg-music is also redundant with the over broad `host` filesystem permission.

io.ossia.score.yml
hfiguiere: btw the comment is incorrect, this is not the session bus.

io.ossia.score.yml
```diff
@@ -0,0 +1,333 @@
...
+
+  # Audio I/O
+  - name: portaudio
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+      - -DCMAKE_POSITION_INDEPENDENT_CODE=ON
+      - -DBUILD_SHARED_LIBS=OFF
+      - -DPA_ALSA_DYNAMIC=ON
+      - -DPA_USE_JACK=OFF
```
hfiguiere, line 119: why disabling jack when you enable the pipewire socket?

io.ossia.score.yml
```diff
@@ -0,0 +1,333 @@
...
+  - shared-modules/linux-audio/lv2.json
+  - shared-modules/linux-audio/lilv.json
+
+  # Audio I/O
+  - name: portaudio
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+      - -DCMAKE_POSITION_INDEPENDENT_CODE=ON
+      - -DBUILD_SHARED_LIBS=OFF
```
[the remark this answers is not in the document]
hfiguiere: in general all should be built shared.

io.ossia.score.yml
```diff
@@ -0,0 +1,333 @@
...
+  - name: ossia-score
+    buildsystem: cmake-ninja
+    builddir: true
+    no-make-install: true
+    sources:
+      - type: archive
+        url: https://github.com/ossia/score/releases/download/v3.2.3/ossia.score-3.2.3-src.tar.xz
+        sha256: 0
+    config-opts:
+      # Note: don't use RelWithDebInfo as ossia release builds explicitly expect Release build type.
```
hfiguiere, line 315: so no debug symbols?

io.ossia.score.yml
hfiguiere: it's also redundant with the broad filesystem permissions.

io.ossia.score.yml
hfiguiere: also this doesn't enable `/dev/shm` and JACK will never work that way. The platform provides `pw-jack` instead.

io.ossia.score.yml
hfiguiere: also VST and VST3 UI support will never work with Wayland so it's moot.

io.ossia.score.yml
jcelerier: oh, I read multiple comments that --device=all would enable /dev/shm. well :)

for MIDI : are also Alsa RawMidi devices available that way (basically raw access to a /dev, needed for some features not supported by the ALSA Sequencer API and necessary for hardware that supports more than classic MIDI 1 speed and large sysex: https://github.com/thestk/rtmidi/issues/214), as well as pipewire MIDI devices (needed for MIDI BLE support) ?

jcelerier: and - is raw JACK still possible ? even with pipewire 1.2.0 I still encounter cases where it's worth running JACK1 in terms of raw performance - or even raw ALSA PCM support (this is really necessary for instance when the software runs on e.g. Raspberry Pi Zero where just the pw daemon running and doing nothing is already using too much resources).

io.ossia.score.yml
jcelerier: okay, I could see many manifests which had both. How would an application handle running on either through QT_QPA_PLATFORM ? (+ also running directly on eglfs without X11 or Wayland involved - the app supports that and it's an important feature when running on embedded boards for art installations, etc.)

io.ossia.score.yml
jcelerier: hmm it writes .config/ossia and .cache/ossia, should I put that ?

jcelerier: as well as $XDG_DOCUMENTS_DIR/ossia

io.ossia.score.yml
jcelerier: okay, I need to see why the build system didn't pick it up then

io.ossia.score.yml
jcelerier: hmm same, it wasn't picked up by the build system, investigating..

io.ossia.score.yml
jcelerier: I disabled it explicitly - the default install target installs a lot of unnecessary things for this kind of linux packaging (but necessary in other cases). installing the "OssiaScore" component installs just the necessary things for the package.

io.ossia.score.yml
jcelerier: thanks, I had trouble understanding what was the pwd at this point

io.ossia.score.yml
bbhtt: > hmm it writes .config/ossia and .cache/ossia, should I put that ?

It should use XDG_CACHE_HOME and XDG_CONFIG_HOME instead of hardcoding `~/.config` then you don't need to do anything.

io.ossia.score.yml
jcelerier: hmm, users of the software generally have various data hard drives and folders mounted with audio, video content, etc. what would be the most frictionless way to handle this compared to the AppImage ? (every permission dialog generally ends up loosing less technical users from the UX testing I could run over the years)

io.ossia.score.yml
jcelerier: ah, this is where appimage expects it and right now I plan to support both so..

io.ossia.score.yml
jcelerier: ossia already provides direct implementation of JACK through the JACK API (as well as direct implementation of PipeWire audio and midi through the pipewire API) and this causes issues (crashes & deadlocks) when portaudio does its own jack init / teardown in Pa_Initialize / Pa_Terminate.

io.ossia.score.yml
jcelerier: as of now no though I can try to fix that upstream for a further version

io.ossia.score.yml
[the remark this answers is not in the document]
jcelerier: what's the value of this when no other process will be able to load these .so and thus won't benefit from deduplication ? This just adds indirection and increases binary size as the unused code cannot even be removed by the linker.

io.ossia.score.yml
jcelerier: okay, this is done through Qt's QStandardPaths so I assume the correct folders are already chosen automatically

io.ossia.score.yml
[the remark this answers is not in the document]
jcelerier: also technically I can build most of the individual classes in the software as their own shared libs - would that bring any value ?

io.ossia.score.yml
hfiguiere: this will break some plugins. It's subtle but necessary,

io.ossia.score.yml
jcelerier: basically all the header files for the development SDK of the app - I don't think it's suitable for the application flatpak to provide that right ?

io.ossia.score.yml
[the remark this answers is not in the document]
hfiguiere: deduplication is done at file level. That mean that if you update the main binary but not the dependencies then only the main binary is downloaded as part of the update. If you link everything static then you have humongus binaries.

The value is that we build libraries shared. There are a few excpetions, this is isn't one.

io.ossia.score.yml
hfiguiere: ok

io.ossia.score.metainfo.xml
```diff
@@ -0,0 +1,51 @@
...
+    <screenshot type="default">
+      <image>https://github.com/ossia/score/raw/master/docs/score.png?raw=true</image>
+      <caption>Sequence all kinds of media in ossia score</caption>
+    </screenshot>
+  </screenshots>
+  <releases>
+    <release version="3.2.2" date="2024-07-01">
+    <url type="details">https://github.com/ossia/score/releases/tag/v3.2.2</url>
+      <description>
+        ossia score 3.2.2 is a bugfix release with mostly non-Linux-related improvements.
```
hfiguiere, line 42: this is invalid markup. Text must be at least in `<p/>` tag.

io.ossia.score.yml
jcelerier: wouldn't it be possible to just bind-mount host libjack if it exists? (ossia loads jack through dlopen to enable support for systems that do not provide JACK at all so it does not care about a specific libjack version)

io.ossia.score.yml
```diff
@@ -0,0 +1,275 @@
...
+      - -DSCORE_FHS_BUILD=1
+      - -DCMAKE_UNITY_BUILD=1
+      - -DCMAKE_SKIP_RPATH=ON
+      - -DOSSIA_USE_SYSTEM_LIBRARIES=1
+      - -DSCORE_USE_SYSTEM_LIBRARIES=1
+      - -DSCORE_NO_QTCONF=1
+
+    post-install:
+     - cmake -DCMAKE_INSTALL_DO_STRIP=1 -DCOMPONENT=OssiaScore -P cmake_install.cmake
+     - install -Dm644 ../cmake/Deployment/Linux/Flatpak/io.ossia.score.metainfo.xml /app/share/metainfo/io.ossia.score.metainfo.xml
```
hfiguiere, line 274: since the appstream file is taken from the source tree why is it found in this repository?

io.ossia.score.metainfo.xml
hfiguiere, line 1: remove. see https://github.com/flathub/flathub/pull/5376/files#r1662418969

io.ossia.score.yml
hfiguiere: that's why `cleanup` exist.

io.ossia.score.yml
hfiguiere: No. It is not.

io.ossia.score.yml
jcelerier: I'd rather remove it from the upstream repository and have it there as this means that if I do e.g. a typo in the release notes I have to make a whole new release to fix it due to the "only released version" flatpak policy

io.ossia.score.metainfo.xml
jcelerier: hmm, when I did this appstream-validate gave me an issue : 
```
$ appstreamcli validate --explain io.ossia.score.metainfo.xml
I: io.ossia.score:11: description-first-word-not-capitalized
   The description line does not start with a capitalized word, project name or number.
```

can I safely ignore it ? (the software name is explicitly lower-case and should never be spelled Ossia Score or Ossia score for instance)

io.ossia.score.yml
jcelerier: I'll change it but I really think that flatpak should provide a way to specifiy which component to install instead of forcing to install a few thousand files just to remove them afterwards - this is just wasting CI time for no good reason

hfiguiere: You mean like `make-install-args` ?

(works with Ninja too)

hfiguiere: right now the regular install target is run anyway so I don't see your point.

io.ossia.score.yml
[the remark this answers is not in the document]
jcelerier: > If you link everything static then you have humongus binaries.

for reference the appimage of ossia, which links against Qt, ffmpeg, libllvm, libclang, and all the libraries mentioned here is 101 megabytes and not particularly optimized - in contrast, with flatpak's shared libs approach there's 1.5G to install to get to not even the same feature level, that's a 15x difference in size.. so that's going to be a tough pill for my users which I'll have a hard time arguing for :) but fair, I changed the build to use everything shared.

io.ossia.score.yml
jcelerier: okay, I had never come across those in 20-ish years of linux :) adding this to the main build for the next release

io.ossia.score.yml
hfiguiere: > otoh if `install-rule` calls `cmake --install` instead of the actual build system then it would work

no it calls `ninja` or `make` depending on the case.

io.ossia.score.yml
jcelerier: hmm, the flatpak documentation mentioned it being available since 1.15.6 (I have 1.15.8 on my system) - is it not correct ? https://docs.flatpak.org/en/latest/sandbox-permissions.html

io.ossia.score.yml
hfiguiere: Ardour, MuseScore, to list a few.
Most GNOME and KDE apps too.

jcelerier: changed in upstream to install it unconditionally

io.ossia.score.yml
```diff
@@ -0,0 +1,281 @@
...
+  # Audio support through pipewire
+  - --filesystem=xdg-run/pipewire-0
+
+  # Ignore ALSA configuration
+  - --env=ALSA_CONFIG_PATH=
+
+  # Plugins
+  - --env=VST_PATH=/app/extensions/Plugins/vst
+  - --env=VST3_PATH=/app/extensions/Plugins/vst3
+  - --env=LV2_PATH=/app/extensions/Plugins/lv2
```
hfiguiere, line 42: I'll reiterate what I said previously. This path for LV2 is incorrect because it doesn't include `~/.lv2` nor `/app/lib/lv2`. Unfortunately LV2 doesn't expand `~` and `$HOME` isn't expanded here.

io.ossia.score.yml
```diff
@@ -0,0 +1,281 @@
...
+  - /lib/debug
+  - /lib/python*
+  - /share/pixmaps
+  - /share/pkgconfig
+  - /share/aclocal
+  - /share/gtk-doc
+  - /share/doc
+  - /share/info
+  - /share/man
+  - /share/pixmaps
```
hfiguiere, line 90:
```suggestion
```

dupe

io.ossia.score.yml
hfiguiere: nevermind. just put it all in the wrapper script and remove it here.

ossia-score
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+export LV2_PATH="$HOME/.lv2:$LV2_PATH"
```
hfiguiere, line 3:
```suggestion
export LV2_PATH="$HOME/.lv2:/app/extensions/Plugins/lv2:/app/lib/lv2"
```

io.ossia.score.yml
```diff
@@ -0,0 +1,281 @@
...
+     version: '23.08'
+     add-ld-path: lib
+     merge-dirs: vst;vst3;lv2
+     subdirectories: true
+     no-autodownload: true
+
+  # Full codec support for FFMPEG
+  org.freedesktop.Platform.openh264:
+    directory: lib/openh264,
+    add-ld-path: .,
```
hfiguiere, line 60: I'm pretty sure this is incorrect the `,`.

And below.

hfiguiere: also I don't think ffmpeg uses openh264.

io.ossia.score.yml
```diff
@@ -0,0 +1,281 @@
...
+  - --share=network
+
+  # PulseAudio is needed for ALSA device access
+  - --socket=pulseaudio
+
+  # Audio support through pipewire
+  - --filesystem=xdg-run/pipewire-0
+
+  # Ignore ALSA configuration
+  - --env=ALSA_CONFIG_PATH=
```
bbhtt, line 37: Why do you need to ignore config from runtime?

io.ossia.score.yml
```diff
@@ -0,0 +1,281 @@
...
+     no-autodownload: true
+
+  # Full codec support for FFMPEG
+  org.freedesktop.Platform.openh264:
+    directory: lib/openh264,
+    add-ld-path: .,
+    version: "2.4.1"
+
+  org.freedesktop.Platform.ffmpeg-full:
+    directory: lib/ffmpeg,
```
bbhtt, line 64:
```suggestion
    directory: lib/ffmpeg
```

io.ossia.score.yml
```diff
@@ -0,0 +1,281 @@
...
+  - /include
+  - /share/gir-1.0
+  - /share/man
+  - '*.a'
+  - '*.la'
+  - /lib/cmake
+  - /lib/girepository-1.0
+  - /lib/pkgconfig
+  - /lib/*.la
+  - /lib/*.a
```
bbhtt, line 79:
```suggestion
```

The prevous "*.a, *.la` will clean this up too.

io.ossia.score.yml
```diff
@@ -0,0 +1,281 @@
...
+  - /lib/avahi
+  - /lib/debug
+  - /lib/python*
```
bbhtt, line 82: Is cleaning up python modules correct?

io.ossia.score.yml
```diff
@@ -0,0 +1,281 @@
...
+  - /share/man
+  - '*.a'
+  - '*.la'
+  - /lib/cmake
+  - /lib/girepository-1.0
+  - /lib/pkgconfig
+  - /lib/*.la
+  - /lib/*.a
+  - /lib/avahi
+  - /lib/debug
```
bbhtt, line 81:
```suggestion
```

`/lib/debug` should not be cleaned up.

io.ossia.score.yml
```diff
@@ -0,0 +1,281 @@
...
+      - -DSCORE_STATIC_PLUGINS=1
+      - -DSCORE_FHS_BUILD=1
+      - -DCMAKE_UNITY_BUILD=1
+      - -DCMAKE_SKIP_RPATH=ON
+      - -DOSSIA_USE_SYSTEM_LIBRARIES=1
+      - -DSCORE_USE_SYSTEM_LIBRARIES=1
+      - -DSCORE_NO_QTCONF=1
+
+    post-install:
+     - install -Dm755 /app/bin/ossia-score /app/bin/ossia-score.bin
```
bbhtt, line 279: why are you adding `.bin`? The command in the toplevel is `ossia-score`

bbhtt: Also your ossia-score script doesn't seem to be added as a source anywhere.

It needs to be added with `file: ossia-score`

io.ossia.score.yml
hfiguiere: this is the linter failure.

io.ossia.score.yml
```diff
@@ -0,0 +1,281 @@
...
+finish-args:
+  # BLE support
+  - --allow=bluetooth
+
+  # Access user's data and media libraries, plug-ins etc
+  - --filesystem=host
+
+  # GPU rendering support
+  - --device=dri
+
```
bbhtt, line 17:
```suggestion
```

`all` includes dri

io.ossia.score.yml
hfiguiere: 1.15.x is NOT a stable release. It is not available on most distribution.
It is an incompatible change. 

See https://github.com/flatpak/flatpak/issues/5681 if you don't believe me.

io.ossia.score.yml
hfiguiere: LV2 require a wrapper script.

ossia-score
hfiguiere, line 1: this file isn't in the source list.

io.ossia.score.yml
jcelerier: I saw this in all the pro-audio flatpak I saw - Ardour, Mixxx, Bitwig. Ardour has a comment that it is necessary for sound to work on SuSE : https://github.com/flathub/org.ardour.Ardour/blob/master/org.ardour.Ardour.json#L19C23-L21C31

io.ossia.score.yml
jcelerier: I saw that those were added there by LV2 but I'm pretty sure the app does not use python in any shape or form

io.ossia.score.yml
hfiguiere: It was done a bit blindly. I don't have SUSE to reproduce. And then it get copied.

io.ossia.score.yml
jcelerier: hmm if it's ffmpeg with all codecs supported by upstream it definitely does: https://ffmpeg.org/ffmpeg-codecs.html#libopenh264

this comes from this: https://bugzilla.mozilla.org/show_bug.cgi?id=1755361#c11

but is not strictly necessary either, I'll just remove it

hfiguiere: just making sure what the expectations are.

jcelerier:
made a new release upstream with the necessary changes, this will require sha256 adjustment once the source tarball has been generated

io.ossia.score.yml
jcelerier: rationale seems to be given here: https://github.com/flatpak/flatpak/issues/1875#issuecomment-520401097

jcelerier: removed anyways, hopefully it won't break for anyone !

jcelerier:
Hm, I noticed that LV2 support does not get built, cmake does not manage to find `lilv/liv.h`

When I try to build with

    $ flatpak-builder /home/jcelerier/builds/build-flatpak io.ossia.score.yml --force-clean --jobs=8 --ccache   --install --install-deps-from=flathub  --user --build-shell=ossia-score

and run 

    $ find / -name lilv.h
    /app/lib/debug/source/lilv/include/lilv/lilv.h

shouldn't it also be in /app/include/lilv-0/lilv/lilv.h from the shared lilv module ? (`shared-modules/linux-audio/lilv.json`)

hfiguiere:
it should locate it with pkg-config.

hfiguiere:
bot, build io.ossia.score

hfiguiere:
```
Failed to download sources: module ossia-score: Wrong sha256 checksum for ossia.score-3.2.3-1-src.tar.xz, expected "45fb630c8cdb48a444fcd0607126db6a8233cf97be0375b7fc0479b2c64d0e68", was "0b7495b3938e378c919c298c1cce6c522169fba0a708a58363be1ac30845a455"
```

jcelerier:
hadn't updated the checksum, done

hfiguiere:
bot, build io.ossia.score

io.ossia.score.yml
```diff
@@ -0,0 +1,270 @@
...
+      - type: archive
+        url: https://github.com/avahi/avahi/releases/download/v0.8/avahi-0.8.tar.gz
+        sha256: 060309d7a333d38d951bc27598c677af1796934dbd98e1024e7ad8de798fedda
+
+  - name: ossia-score
+    buildsystem: cmake-ninja
+    builddir: true
+    sources:
+        - type: archive
+          url: https://github.com/ossia/score/releases/download/v3.2.3-3/ossia.score-3.2.3-3-src.tar.xz
```
hfiguiere, line 246: this file is 404. You said "updating the sha256" but then changed the URL too.

jcelerier: sorry, the github release upload CI step had a spurious failure (https://dev.azure.com/ossia/libossia/_build/results?buildId=3905&view=logs&j=d3280fad-0e55-5adf-7e56-53b7a94fcec5&t=0093797f-f7df-5d77-022b-dd9b5deb130d&s=ea07b9a6-4434-52b2-ca0a-c05362050f83). It's good now !

bbhtt:
bot, build io.ossia.score

jcelerier:
Oops, missing rename metainfo step

hfiguiere:
probably missing `rename-appdata-file` in the manifest

jcelerier:
Done

hfiguiere:
also aarch6 build failed:

```
  116 | #error "ARM builds require Clang compiler. Disable checking with CMT_FORCE_NON_CLANG"
```

jcelerier:
Yes, I use a DSP library (KFR) which only supports clang under aarch64. Is it possible to make the flatpak build with clang in this case ?

jcelerier:
pushed with clang++ from the clang sdk extension being used, let's see. Can I trigger bot builds myself to not have to hassle you until it works ?

bbhtt:
bot, build io.ossia.score

jcelerier:
@bbhtt this doesn't look like it queued the build ? https://buildbot.flathub.org/#/pendingbuildrequests

jcelerier:
I see in the build log 

    Not exporting share/metainfo/ossia-score.appdata.xml, non-allowed export filename

but I added

    rename-appdata-file: ossia-score 

?

hfiguiere:
it's the filename `ossia-score.appdata.xml`

jcelerier:
but, if I read the docs correctly, shouldn't rename-appdata-file turn this into io.ossia.score.metainfo.xml ?

hfiguiere:
you pass it the **Filename** of the appstream file you want to rename.

hfiguiere:
the doc:

> Any appdata (metainfo) file with this name will be renamed to a name based on id during the cleanup phase.

jcelerier:
okay, done. any way I can run the validation checks locally to avoid this kind of things at least?

bbhtt:
It's in the 4th checlist item that you ticked

bbhtt:
bot, build io.ossia.score

bbhtt:
Looks good, let me know if once you test the build.

jcelerier:
Builds, runs and does what's expected of it :)

hfiguiere:
bot, build io.ossia.score

io.ossia.score.yml
```diff
@@ -0,0 +1,276 @@
...
+      - -DSCORE_FHS_BUILD=1
+      - -DCMAKE_UNITY_BUILD=1
+      - -DCMAKE_SKIP_RPATH=ON
+      - -DOSSIA_USE_SYSTEM_LIBRARIES=1
+      - -DSCORE_USE_SYSTEM_LIBRARIES=1
+      - -DSCORE_NO_QTCONF=1
+
+    post-install:      
+     - install -Dm755 "${FLATPAK_DEST}/bin/ossia-score" "${FLATPAK_DEST}/bin/ossia-score.bin"
+     - install -Dm755 ../run "${FLATPAK_DEST}/bin/ossia-score"
```
hfiguiere, line 275:
```suggestion
     - install -Dm755 ../run "${FLATPAK_DEST}/bin/ossia-score"
     - install -d /app/extensions/Plugins
```

this is required for the plugins to work. As I get 

```
bwrap: Can't mkdir parents for /app/extensions/Plugins: Read-only file system
```

hfiguiere:
bot, build io.ossia.score

io.ossia.score.yml
```diff
@@ -0,0 +1,277 @@
...
+     - install -d /app/extensions/Plugins
```
hfiguiere, line 276:
```suggestion
     - install -d /app/extensions/Plugins
     - install -d /app/lib/ffmpeg
```

that one too. I guess of have none of the extension kind installed when you tried.

jcelerier: what is this folder for ? I already see a fair list of A/V codecs available when running without this change

hfiguiere: it's for this

https://github.com/flathub/flathub/pull/5376/files#diff-a0e9197e52d315e8d7077671bea3cbb05bea4a79a29962f62cc6f013c0debaedR57

The mount points for extensions need to exist. I guess if you run the package without error it's because the extensions isn't installed.

hfiguiere:
bot, build io.ossia.score

hfiguiere:
Apparently `VST_PATH` is ignored:

image

jcelerier:
> and there is no other options but VST.

yes, the other supported plug-in apis haven't needed a config panel so far - the plug-ins should show up in the "processes" panel : 
here

hfiguiere:
so it doesn't find the VST3 either. (`VST3_PATH` which unlike `VST_PATH`, isn't standard is ignore too)

There is an open issue for the VST3 SDK to have a standard way to set a path, but it's still open.

Either this needs to be patched (like I do with JUCE), or `VST3_PATH` needs to be supported.

jcelerier:
does LV2 have an equivalent env. var? it seems to find *some* of the LV2s I have but not all

hfiguiere:
> does LV2 have an equivalent env. var? it seems to find _some_ of the LV2s I have but not all

that's what LV2_PATH is, and AFAIK since you use the standard lv2 support libraries it works out of the box. I see my LV2 plugins here.

jcelerier:
okay. note that for the LV2 plug-in UIs this requires a patch to upstream Suil as it didn't support Qt 6 so far (https://github.com/lv2/suil/pull/21)

hfiguiere:
you could add the patch to the suil build.

jcelerier:
added, but note that the software itself doesn't support that yet in any released version

hfiguiere:
bot, build io.ossia.score

bbhtt:
Is this ready or are there more changed to be made?

hfiguiere:
bot, build io.ossia.score

hfiguiere:
> Is this ready or are there more changed to be made?

plugins are found this time. So on my side it is :+1:

jcelerier:
Good for me to :)

io.ossia.score.yml
pedropaulosuzuki: Now that we have the 1.16 series, this is on stable. Still, many distros are on 1.14 or even 1.12, so yeah :/

We need a fallback system A.S.A.P

PR #5385, opened 2024-07-05
Add org.kde.CrowTranslate

Shatur:

A simple and lightweight translator that allows you to translate and speak text.

The application was previously submitted under [io.crow_translate.CrowTranslate](https://github.com/flathub/io.crow_translate.CrowTranslate), but I need to change the appid due to the transition under the KDE umbrella. After creating a new repository, I will send a PR with a eol rebase manifest to the previous repository. So users will be notified about the appid change.


io.crow_translate.CrowTranslate.yaml
```diff
@@ -0,0 +1,91 @@
...
+modules:
+  - name: crow-translate
+    buildsystem: cmake-ninja
+    sources:
+      - type: git
+        url: https://github.com/crow-translate/crow-translate.git
+        tag: 2.11.1
+        commit: 7113ff01a994584ed0a3ec557fc3a694fcb06b63
+        x-checker-data:
+          type: git
```
bbhtt, line 36: This repo is archived, I'm guessing this should be using KDE provided tarballs.

bbhtt:
cc @flathub/kde

io.crow_translate.CrowTranslate.yaml
bbhtt: Also is this the source now https://invent.kde.org/office/crow-translate ? (That'd justify the org.kde. ID)

Shatur: > This repo is archived, I'm guessing this should be using KDE provided tarballs.

It's because we don't have a new release yet. But I want to create a flathub entry first, otherwise [the app page](https://apps.kde.org/crowtranslate/) will point to an old release.

> 

> Also is this the source now https://invent.kde.org/office/crow-translate ? (That'd justify the org.kde. ID)

Yes!

hfiguiere:
cc @flathub/kde

io.crow_translate.CrowTranslate.yaml
travier: Then please use invent as the source, with a fixed commit, not GitHub

io.crow_translate.CrowTranslate.yaml
```diff
@@ -0,0 +1,91 @@
...
+        builddir: true
+        config-opts:
+          - -DBUILD_SHARED_LIBS:BOOL=ON
+        cleanup:
+          - /bin
+        sources:
+          - type: git
+            url: https://github.com/tesseract-ocr/tesseract.git
+            tag: 5.3.4
+            commit: 8ee020e14cf5be4e3f0e9beb09b6b050a1871854
```
travier, line 49: Please use release tarballs as much as possible, not direct Git repos.

io.crow_translate.CrowTranslate.yaml
```diff
@@ -0,0 +1,91 @@
...
+copy-icon: true
+command: crow
+cleanup:
+  - '*.a'
+  - '*.la'
+  - /include
+  - /lib/cmake
+  - /lib/pkgconfig
+modules:
+  - name: crow-translate
```
travier, line 28: This should come last

io.crow_translate.CrowTranslate.yaml
```diff
@@ -0,0 +1,91 @@
...
+          - cp -av * ${FLATPAK_DEST}/share/tessdata/
+        sources:
+          - type: git
+            url: https://github.com/tesseract-ocr/tessdata_fast.git
+            tag: 4.1.0
+            commit: 65727574dfcd264acbb0c3e07860e4e9e9b22185
+            x-checker-data:
+              type: git
+
+      - name: extra-cmake-modules
```
travier, line 84: This is already in the SDK

io.crow_translate.CrowTranslate.yaml
Shatur: Could you elaborate?

io.crow_translate.CrowTranslate.yaml
Shatur: Switched to using invent. I realized that we imported previous releases, so I can just change the URL and it will work.

io.crow_translate.CrowTranslate.yaml
Shatur: Removed!

io.crow_translate.CrowTranslate.yaml
Shatur: Switched to tarballs. I just used previously submitted yaml.

io.crow_translate.CrowTranslate.yaml
bbhtt: > This should come last

This is the only toplevel module. It's using nested modules.

bbhtt:
bot, build org.kde.CrowTranslate

bbhtt:
bot, build org.kde.CrowTranslate

io.crow_translate.CrowTranslate.yaml
bbhtt, line 1: Manifest filename is wrong `org.kde.CrowTranslate.yaml`

bbhtt:
Something is wrong with build requests, I'll start it manually once the filename is fixed.

bbhtt:
Ah Github is having an outage https://www.githubstatus.com/

io.crow_translate.CrowTranslate.yaml
Shatur: Updated, thanks!

Shatur:
Sorry, I accidentally calculated SHA1 instead. Fixed.

bbhtt:
bot, build org.kde.CrowTranslate

org.kde.CrowTranslate.yaml
```diff
@@ -0,0 +1,73 @@
...
+  - '*.la'
+  - /include
+  - /lib/cmake
+  - /lib/pkgconfig
+modules:
+  - name: crow-translate
+    buildsystem: cmake-ninja
+    sources:
+      - type: git
+        url: https://invent.kde.org/office/crow-translate/.git
```
bbhtt, line 32:
```suggestion
        url: https://invent.kde.org/office/crow-translate.git
```

org.kde.CrowTranslate.yaml
```diff
@@ -0,0 +1,73 @@
...
+            builddir: true
+            config-opts:
+              - -DBUILD_SHARED_LIBS:BOOL=ON
+            sources:
+              - type: archive
+                url: https://github.com/DanBloomberg/leptonica/archive/refs/tags/1.84.1.tar.gz
+                sha256: ecd7a868403b3963c4e33623595d77f2c87667e2cfdd9b370f87729192061bef
+
+      - name: tessdata_fast
+        disabled: true
```
bbhtt, line 64: If you are skipping this why is this added here?

Shatur: I just used what was added here: https://github.com/flathub/io.crow_translate.CrowTranslate/blob/ce7138a8686f7858ad2f99ce03ee11f5b8413f88/io.crow_translate.CrowTranslate.yaml#L70
Didn't even notice that it's disabled. Should I just remove it?

bbhtt: I have no idea if you need it or not. Why was it added to the manifest and why was it disabled?

bbhtt:
bot, build org.kde.CrowTranslate

Shatur:
Hm... Looks like it's unable to find the .desktop file.
Why it could be?
I generate it [here](https://invent.kde.org/office/crow-translate/-/blob/master/CMakeLists.txt?ref_type=heads#L253).
And install it [here](https://invent.kde.org/office/crow-translate/-/blob/master/CMakeLists.txt?ref_type=heads#L256).

org.kde.CrowTranslate.yaml
```diff
@@ -0,0 +1,61 @@
...
+  - --device=dri
+  - --socket=pulseaudio
+  - --share=network
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.kde.StatusNotifierWatcher
+  # The app supports capturing screen via KWin and Mutter DBus APIs, but
+  # access to it seems overly broad in spite of the app also supporting XDP API
+  #- --talk-name=org.kde.KWin
+  #- --talk-name=org.gnome.Shell
+rename-icon: crow-translate
```
bbhtt, line 18:
```suggestion
  #- --talk-name=org.gnome.Shell
rename-desktop-file: io.crow_translate.CrowTranslate.desktop
rename-icon: crow-translate
```

Shatur: But why? Isn't appid is `org.kde.CrowTranslate`?

bbhtt: the desktop file you are installing still has the old name...

Did you build this locally or just checked the 4th checkbox?

Shatur: > the desktop file you are installing still has the old name...

Ah, it's because I using the old commit that corresponds to the version with the old appid. What will be the best way to proceed? Add `rename-desktop-file` for now and remove after the release? Or should I pin to the latest commit?

> Did you build this locally or just checked the 4th checkbox?

I built it on KDE CI, but didn't check after making all these changes, sorry.

bbhtt: > Add rename-desktop-file for now and remove after the release? 

Yea that's fine. We can proceed with the build and test. You can remove it later.

bbhtt:
bot, build org.kde.CrowTranslate

bbhtt:
I'm guessing the metainfo is still using the old name too, use `rename-appdata-file` similarly in that case.

Shatur:
Makes sense, on this commit everything corresponds to the old ID.

bbhtt:
bot, build org.kde.CrowTranslate

Shatur:
Looks like something changed since the last submission and I need to add more fields. Let me try to run the linter locally.

Shatur:
@bbhtt I can update my appdata, but it's located in the repository. Should I try to use the latest commit instead?
Or I can temporarily add it here.

bbhtt:
You can add it as a patch https://docs.flatpak.org/en/latest/module-sources.html#patch-sources

bbhtt:
bot, build org.kde.CrowTranslate

Shatur:
> Tray is missing and icon and tray menu is missing icon on GNOME 46/Fedora 40 VM. I guess this would work in a KDE system.

Yes, I just tested the package on KDE and it works correctly. But it's bad that it doesn't work on GNOME properly. I remember [this](https://github.com/crow-translate/crow-translate/issues/665) issue and installing from PPA helped. Do I need to adjust something in the flatpak configuration to make it work with GNOME?

> Also why is it opening a tray by default? Why not show the window first by default?

It's a good question! I personally keep it enabled because I run the application automatically when the system starts. But the autostart itself is disabled by default. So your suggestion makes total sense. Just disabled "Start minimized" by default.

io.crow_translate.CrowTranslate.yaml
travier: Ah, good point, I hadn't realized that. Good then

travier:
I don't think it's worth spending time patching the old version. Let's get all the fixes upstream and use the latest commit here even if it's not released yet and we'll move to a release tarball once it's out.

org.kde.CrowTranslate.yaml
```diff
@@ -0,0 +1,64 @@
...
+runtime-version: 5.15-23.08
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --socket=pulseaudio
+  - --share=network
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.kde.StatusNotifierWatcher
```
travier, line 13: Let's sort this list

org.kde.CrowTranslate.yaml
```diff
@@ -0,0 +1,64 @@
...
+  # The app supports capturing screen via KWin and Mutter DBus APIs, but
+  # access to it seems overly broad in spite of the app also supporting XDP API
+  #- --talk-name=org.kde.KWin
+  #- --talk-name=org.gnome.Shell
```
travier, line 17: This should go into a README for the repo to make it easier to find.

Maybe also in a message in the application when running as Flatpak.

org.kde.CrowTranslate.yaml
```diff
@@ -0,0 +1,64 @@
...
+          type: git
+    modules:
+      - name: tesseract
+        buildsystem: cmake-ninja
+        builddir: true
+        config-opts:
+          - -DBUILD_SHARED_LIBS:BOOL=ON
+        cleanup:
+          - /bin
+        sources:
```
travier, line 48: You should add `x-checker-data` config for all sources

bbhtt:
> Do I need to adjust something in the flatpak configuration to make it work with GNOME?

Not it should work with the current configuration? What icon name are using for the tray? The icon needs to be exported by Flatpak on host so that it is available in XDG_DATA_DIRS. The tray extension cannot access an icon that Flatpak hasn't exported.

Flatpak will export icons from `$prefix/share/icons/hicolor/...` with the following patterns `$FLATPAK_ID, $FLATPAK_ID.foo, $FLATPAK_ID-foo` .png/.svg

Shatur:
Ah, I use `crow-translate` as the icon. That's the issue.
Will change it switch to the latest master.

org.kde.CrowTranslate.yaml
Shatur: Removed completely. `org.kde.KWin` and `org.gnome.Shell` aren't necessary.

org.kde.CrowTranslate.yaml
Shatur: I checked other projects and looks like they don't use `x-checker-data` for archives from GitHub.
For example, [org.flathub.flatpak-external-data-checker](https://github.com/flathub/org.flathub.flatpak-external-data-checker/blob/502b06d4d4e67372ee1c699be27d920995c70e90/org.flathub.flatpak-external-data-checker.yml#L159C27-L159C34) does not do this.

Shatur:
Done! Thank you a lot for the feedback. I'm new to Flatpak and looks like the app wasn't properly packaged before.
But it should be good now. I switched to the master and fixed the mentioned issues.

org.kde.CrowTranslate.yaml
```diff
@@ -0,0 +1,87 @@
...
+  - --share=ipc
+  - --share=network
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --socket=wayland
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.kde.StatusNotifierWatcher
+copy-icon: true
+command: crow
+separate-locales: false
```
bbhtt, line 24: Why is this disabled?

org.kde.CrowTranslate.yaml
```diff
@@ -0,0 +1,87 @@
+id: org.kde.CrowTranslate
+sdk: org.kde.Sdk
+runtime: org.kde.Platform
+runtime-version: 5.15-23.08
+add-extensions:
+  org.kde.CrowTranslate.tessdata:
```
bbhtt, line 6: What is this for?

org.kde.CrowTranslate.yaml
Shatur: The is unable to load localization without this line.

org.kde.CrowTranslate.yaml
Shatur: It's needed for tesseract models for screen reading. I used [this](https://github.com/flathub/io.github.manisandro.gImageReader/blob/master/io.github.manisandro.gImageReader.yml) app, which also uses Tesseract, as a reference.

org.kde.CrowTranslate.yaml
bbhtt: Are you sure? You need to install the `.Locale` extension for that locale. By default Flatpak will only install the locale extension for your host locale.

This disabled creating locale extension and bundles all locales into the app.

Shatur: Ah, I didn't know about it! In my application I have locale selection and it could be confusing when user picks a localization and it doesn't work.
Should I keep `separate-locales: false` or remove it?

bbhtt: Then keep it, usually apps follow the system locale.

Shatur:
Anything else needed from me?
Should we wait for the release drop before creating a repository?

org.kde.CrowTranslate.yaml
```diff
@@ -0,0 +1,87 @@
...
+            sources:
+              - type: archive
+                url: https://github.com/DanBloomberg/leptonica/releases/download/1.84.1/leptonica-1.84.1.tar.gz
+                sha256: 2b3e1254b1cca381e77c819b59ca99774ff43530209b9aeb511e1d46588a64f6
+                x-checker-data:
+                  type: anitya
+                  project-id: 1549
+                  url-template: https://github.com/DanBloomberg/leptonica/releases/download/$version/leptonica-$version.tar.gz
+
+      # The `tessdata_fast` variant is recommended for end users: https://tesseract-ocr.github.io/tessdoc/Data-Files
```
travier, line 72: Can you move that to a README file?

org.kde.CrowTranslate.yaml
```diff
@@ -0,0 +1,87 @@
...
+      - name: tessdata_fast
+        buildsystem: simple
+        build-commands:
+          - rm -rf ${FLATPAK_DEST}/share/tessdata
+          - mkdir -p ${FLATPAK_DEST}/share/tessdata
+          - cp -av * ${FLATPAK_DEST}/share/tessdata/
+        sources:
+          - type: git
+            url: https://github.com/tesseract-ocr/tessdata_fast
```
travier, line 81: Can this be fetched from a tarball instead of Git?

org.kde.CrowTranslate.yaml
```diff
@@ -0,0 +1,87 @@
+id: org.kde.CrowTranslate
+sdk: org.kde.Sdk
+runtime: org.kde.Platform
+runtime-version: 5.15-23.08
+add-extensions:
+  org.kde.CrowTranslate.tessdata:
+    directory: share/tessdata
+    version: stable
+    versions: stable;beta;test
```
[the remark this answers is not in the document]
hfiguiere: but then there is no such package. not even in the submission queue.

travier:
> Anything else needed from me? Should we wait for the release drop before creating a repository?

We can "merge" this once this is ready, which is once the @flathub/kde team is giving it 👍🏻 and then we'll ask Flathub Admins to trigger the repo creation.

org.kde.CrowTranslate.yaml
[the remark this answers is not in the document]
Shatur: Yes, I copied it from [this](https://github.com/flathub/io.github.manisandro.gImageReader/blob/master/io.github.manisandro.gImageReader.yml) app.

> but then there is no such package. not even in the submission queue.

Ah, I thought that it creates a extension for the app. Do I need to package tessdata separately? I don't see ` io.github.manisandro.gImageReader.tessdata` anywhere as well...

I'm new to flatpak.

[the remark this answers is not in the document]
hfiguiere: two wrongs don't make a right.

org.kde.CrowTranslate.yaml
travier: Ah, this copied from https://github.com/flathub/io.github.manisandro.gImageReader/blob/master/io.github.manisandro.gImageReader.yml?rgh-link-date=2024-07-09T12%3A52%3A57Z but I don't know how this works.

org.kde.CrowTranslate.yaml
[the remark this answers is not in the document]
travier: Let's try to figure out how to do properly. I'm not familiar with this so we'll have to search in the docs.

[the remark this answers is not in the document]
hfiguiere: `man flatpak-metadata`

[...]

```
       version (string)
           The branch to use when looking for the extension. If this is not specified, it defaults to the branch of the application
           or runtime that the extension point is for. Available since 0.4.1.

       versions (string)
           The branches to use when looking for the extension. If this is not specified, it defaults to the branch of the
           application or runtime that the extension point is for. Available since 0.9.1, and backported to the 0.8.x branch in
           0.8.4.

```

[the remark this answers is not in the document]
hfiguiere: What it doesn't say is that they are equivalent one being the plural form.

That doesn't make an extension being built. That just declare the extension point.

`man flatpak-manifest`

```
       bundle (boolean)
           If this is true, then the data created in the extension directory is omitted from the result, and instead packaged in a
           separate extension.

```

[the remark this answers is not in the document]
Shatur: > Let's try to figure out how to do properly

Reading it right now. Looks like I need to create a separate manifest for it and add a special entry to my app metadata.
We probably need to notify the maintainer of `io.github.manisandro.gImageReader` as well about it.

> man flatpak-metadata

We aren't talking about the version string, but about the whole `add-extensions` entry.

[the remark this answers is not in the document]
hfiguiere: > Reading it right now. Looks like I need to create a separate manifest for it and add a special entry to my app metadata.

no.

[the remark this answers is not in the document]
Shatur: > That doesn't make an extension being built. That just declare the extension point.

It's because `bundle` is false by default?
So instead of setting a version and creating a separate manifest, I can replace `version` with `bundle: true`?

org.kde.CrowTranslate.yaml
Shatur: Fixed, thanks!

bbhtt:
bot, build org.kde.CrowTranslate

org.kde.CrowTranslate.yaml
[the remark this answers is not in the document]
bbhtt: > Yes, I copied it from [this](https://github.com/flathub/io.github.manisandro.gImageReader/blob/master/io.github.manisandro.gImageReader.yml) app.

See the tip here https://docs.flatpak.org/en/latest/extension.html#extension-point Since you don't have a beta application you don't need this. `test` is not needed, it's a temporary repo.

> It's because bundle is false by default?

Yes

https://docs.flatpak.org/en/latest/extension.html#bundled-extensions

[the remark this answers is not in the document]
Shatur: I think I figured it out. I enabled bundling and now the extension creates properly. I also set version that matches the tessdata version.

Opened an issue no notify the maintainer of io.github.manisandro.gImageReader about it: https://github.com/flathub/io.github.manisandro.gImageReader/issues/70

@hfiguiere Thanks for pointing it out, but if your goal is to help the community, try to give helpful advice to newcomers instead of just saying "it's wrong".

org.kde.CrowTranslate.yaml
Shatur: Done!

org.kde.CrowTranslate.yaml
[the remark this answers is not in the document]
Shatur: That's what I thought, thanks!
It looks like, in this case, the tessada was installed as part of the main manifest, but the extension relied on a separate, non-existent manifest. The author has just applied the same fix: https://github.com/flathub/io.github.manisandro.gImageReader/pull/71

bbhtt:
The tray menu items are still missing icons

image

Shatur:
Thanks for testing!

I wondering why it happens... I use `window`, `configure` and `application-exit` icons for them respectively. It works in KDE. Are they missing in Advaita?

bbhtt:
The first two doesn't exist.

bbhtt:
You can probably use preferences-other for the configure, it seems to exist in Breeze and Adwaita legacy. The window one doesn't have an option that exists in both other than window-new.

Shatur:
Thanks! Used `preferences-other` for `configure` and `window-maximize` for `window`.

Shatur:
Switched to just using the application tray icon instead of `window-maximize`, I think it fits better.

bbhtt:
bot, build org.kde.CrowTranslate

travier:
bot, build org.kde.CrowTranslate

travier:
Works well for me. LGTM

Shatur:
> After creating a new repository, I will send a PR with a eol rebase manifest to the previous repository. So users will be notified about the appid change.

Should I proceed with this?

bbhtt:
Yea

PR #5390, opened 2024-07-06
Add com.one_ware.OneWare

HendrikMennen:

- [X] I am an author/developer/upstream contributor of the project.

This project is already released as a snap. I am the author. It is an IDE for electronic development and needs access to USB.  It is a free to use software but there might be some premium extensions in the future. It is my first time building a flatpak, I carefully followed the guides. The only issues I had while linting were about the mirrored screenshots/icons despite using the --mirror arguments, I guess this will be fixed with release?

The website is https://one-ware.com/studio


com.one_ware.OneWare.desktop
hfiguiere, line 1: this should be upstream

com.one_ware.OneWare.metainfo.xml
hfiguiere, line 1: this should be upstream

com.one_ware.OneWare.metainfo.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.one_ware.OneWare</id>
+  <name>OneWare Studio</name>
+  <summary>IDE for Electronics Development</summary>
+  <developer id="one-ware.com">
+    <name translatable="no">Hendrik Mennen</name>
+  </developer>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>LicenseRef-proprietary</project_license>
```
hfiguiere, line 10: you said " It is a free software " but then this. Which is it?

com.one_ware.OneWare.png
hfiguiere, line 1: this should be upstream

flathub.json
```diff
@@ -0,0 +1,5 @@
+{
+    "only-arches": [
+        "x86_64"
```
hfiguiere, line 3: this is disappointing.

com.one_ware.OneWare.yml
```diff
@@ -0,0 +1,49 @@
...
+  - name: oneware
+    buildsystem: simple
+    build-commands:
+      - install -D apply_extra ${FLATPAK_DEST}/bin/apply_extra
+      - install -D -m755 oneware ${FLATPAK_DEST}/bin/oneware
+      - install -D -m644 ${FLATPAK_ID}.desktop ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop
+      - install -D -m644 ${FLATPAK_ID}.png ${FLATPAK_DEST}/share/icons/hicolor/512x512/apps/${FLATPAK_ID}.png
+      - install -D -m644 ${FLATPAK_ID}.metainfo.xml ${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.metainfo.xml
+    sources:
+      - type: extra-data
```
hfiguiere, line 27: why do you use extra-data? You said you were the upstream developer. So that imply you allow redistribution (non withstanding the " It is a free software "), so extra-data should not be used here.

com.one_ware.OneWare.yml
```diff
@@ -0,0 +1,49 @@
...
+tags:
+  - proprietary
+finish-args:
+  - --persist=.local/share/OneWareStudio
+  - --device=all
+  - --share=ipc
+  - --share=network
+  - --socket=pulseaudio
+  - --socket=x11
+  - --filesystem=host
```
hfiguiere, line 16: this is a very broad file system permission?

com.one_ware.OneWare.metainfo.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.one_ware.OneWare</id>
+  <name>OneWare Studio</name>
+  <summary>IDE for Electronics Development</summary>
+  <developer id="one-ware.com">
```
bbhtt, line 6:
```suggestion
  <developer id="com.one_ware">
```

com.one_ware.OneWare.yml
```diff
@@ -0,0 +1,49 @@
+app-id: com.one_ware.OneWare
+runtime: org.freedesktop.Platform
+runtime-version: '23.08'
+sdk: org.freedesktop.Sdk
+command: oneware
+separate-locales: false
+tags:
+  - proprietary
+finish-args:
+  - --persist=.local/share/OneWareStudio
```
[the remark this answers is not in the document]
bbhtt: persist does not work here since you used host Did you see the line here? https://docs.flatpak.org/en/latest/sandbox-permissions.html#filesystem-access

> The --persist=DIR option can be used to map directories from the user’s home directory into the sandbox filesystem. This only works if the application has no home or a broader permission like host that includes home.

com.one_ware.OneWare.yml
```diff
@@ -0,0 +1,49 @@
...
+  - --device=all
```
bbhtt, line 11: why do you need `all` here?

com.one_ware.OneWare.yml
HendrikMennen: I was looking at other examples for IDEs and in my opinion it does make sense to have broad file system permissions for a code editor, similar to Rider or VSCode?

com.one_ware.OneWare.yml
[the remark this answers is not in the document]
HendrikMennen: > persist does not work here since you used host Did you see the line here? https://docs.flatpak.org/en/latest/sandbox-permissions.html#filesystem-access
> 
> > The --persist=DIR option can be used to map directories from the user’s home directory into the sandbox filesystem. This only works if the application has no home or a broader permission like host that includes home.

Understood, So I have two options: Remove the persist and keep it on host (like other IDEs) or try to make it work with XDG directories. Since this is a code editor that can be used for every kind of code, I prefer the first option.

com.one_ware.OneWare.metainfo.xml
HendrikMennen: I think I said this wrong, I appologize. By free I wanted to say free to use, unlike many Jetbrains IDEs for example where you need to login first. The source is not entirely free (yet), which we might do at some point. A big part of the IDE system is already open source, but not everything.

com.one_ware.OneWare.yml
```diff
@@ -0,0 +1,49 @@
...
+  - --share=ipc
+  - --share=network
+  - --socket=pulseaudio
```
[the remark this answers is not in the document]
HendrikMennen: While this IDE (without extensions) doesn't use sound devices out of the box, this functionality will be probably added later with extensions to add some features regarding audio hardware development.

com.one_ware.OneWare.yml
HendrikMennen: I need access to USB devices for programming the hardware and using components like Serial Monitor.
In the [docs](https://docs.flatpak.org/en/latest/sandbox-permissions.html#device-access) it was not clear to me if dri and input would be enough

flathub.json
HendrikMennen: While compiling avalonia for ARM is possible, some binaries that are needed to make this IDE work fully are not available (yet). Also I would prefer to test this before allowing arm builds

com.one_ware.OneWare.yml
HendrikMennen: While the IDE System I made is open source, the Studio itself (including FPGA hardware features) is not.
Thats why the sources are downloaded from our cdn.

HendrikMennen:
Thanks for the quick review. I think I need to clarify something:

https://github.com/one-ware/OneWare is only the IDE system used by this project, it does not contain any hardware development features. Maybe I should change the name to make this more obvious.

The software available on our website or on Snap (https://snapcraft.io/oneware) uses this system as a submodule and adds Hardware Development functionality which is currently not open source.

com.one_ware.OneWare.yml
hfiguiere: note that I used the word "redistribution" here. A Free Software license would imply it, but it's not required. Since you are the upstream developer you should know whether it can be reditributed.

The use of `extra-data` is only for situation where things can be redistributed.

com.one_ware.OneWare.yml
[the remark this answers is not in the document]
hfiguiere: so it doesn't.
```suggestion
```

com.one_ware.OneWare.yml
HendrikMennen: I see, guess I missed that in the documentation. I was following an example from another flatpak that is also using Avalonia. So I guess I should move away from extra-data and instead download the sources in the build process. I will try this again later

com.one_ware.OneWare.yml
bbhtt: all is ok, input is not widely available.

com.one_ware.OneWare.yml
HendrikMennen: I changed it tom home already. This will force users to install their Quartus (or other FPGA toolchains) in home aswell. At least for quartus thats the default install location anyways so its not so bad.
image

HendrikMennen:
I changed the png to a svg according to the guide.
I put the resources (manifest, svg, .desktop) in a different repo to get shared with snap: 
https://github.com/one-ware/OneWareStudio.Resources

hfiguiere:
they should be with the upstreamm source. None of these are specific to flatpak or snap. It's part of the Xdg spec.

HendrikMennen:
> they should be with the upstreamm source. None of these are specific to flatpak or snap. It's part of the Xdg spec.

The upstream source is not public. And I guess its not making a lot of sense to put them into the tar.gz archives. Or am I missing something?

hfiguiere:
it actually does. because that's part of the distribution.

HendrikMennen:
it is now contained inside the tar.gz as requested

HendrikMennen:
Decided to open source the entire software. So I can build it inside the flatpak. Consider this on hold until I'm done with this

HendrikMennen:
Building the app inside the flatpak is not easily possible since there is no network available to restore nuget packages and seems to be broken the intended way.

Here is my try doing that:
https://github.com/one-ware/com.one_ware.OneWare/tree/dotnet-plugin

```
/run/build/oneware/source/studio/OneWare.Studio.Desktop/OneWare.Studio.Desktop.csproj : error NU1101: Unable to find package Microsoft.NETCore.App.Crossgen2.linux-x64. No packages exist with this id in source(s): /run/build/oneware/nuget-sources
```


Even if the generator would work, it would still be very annoying to always build these sources files.
Thats why I think its easier and more reliable to just use the prebuilt binaries.

I am happy with this now, except for the filesystem=home, since it messes up the built-in terminal aswell:
image

The .desktop / metainfo can be found here now:
https://github.com/one-ware/OneWare/tree/main/studio/OneWare.Studio.Desktop

Any further suggestions?

bbhtt:
bot, build om.one_ware.OneWare

bbhtt:
bot, build com.one_ware.OneWare.yml

bbhtt:
bot, build com.one_ware.OneWare

HendrikMennen:
bot, build com.one_ware.OneWare

HendrikMennen:
bot, build com.one_ware.OneWare

bbhtt:
Unless you are fixing issues, you don't have to update versions in this PR. You can do that once the repo is created.

HendrikMennen:
I have some memoryaccess violations with my integrated terminal that only happen inside the flatpak container. I thought I fixed them, which is why I updated here. Sadly the issue persists for now, will convert to draft until I find a solution

HendrikMennen:
I found a solution, which consists of not compiling with dotnet "ReadyToRun" flags. This also makes it possible to use the nuget generator and allows me to build it from source.

Is it okay to keep the nuget sources inside this repo?
I would set up a pipeline that updates them using this script https://github.com/flatpak/flatpak-builder-tools/tree/master/dotnet automatically when I release an update.

bbhtt:
bot, build com.one_ware.OneWare

com.one_ware.OneWare.yml
```diff
@@ -0,0 +1,53 @@
...
+runtime-version: '23.08'
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.dotnet8
+build-options:
+  prepend-path: "/usr/lib/sdk/dotnet8/bin"
+  append-ld-library-path: "/usr/lib/sdk/dotnet8/lib"
+  env:
+    PKG_CONFIG_PATH: "/app/lib/pkgconfig:/app/share/pkgconfig:/usr/lib/pkgconfig:/usr/share/pkgconfig:/usr/lib/sdk/dotnet8/lib/pkgconfig"
+  arch:
```
bbhtt, line 12:
```suggestion
  prepend-path: "/usr/lib/sdk/dotnet8/bin"
  append-ld-library-path: "/usr/lib/sdk/dotnet8/lib"
  prepend-pkg-config-path: "/usr/lib/sdk/dotnet8/lib/pkgconfig"
  arch:
```

com.one_ware.OneWare.yml
```diff
@@ -0,0 +1,53 @@
...
+    arm:
+      env:
+        RUNTIME: "linux-arm"
```
bbhtt, line 15:
```suggestion
```

we build for aarch64

com.one_ware.OneWare.yml
```diff
@@ -0,0 +1,53 @@
...
+      - install -D -m755 oneware ${FLATPAK_DEST}/bin/oneware
+      - install -D -m644 studio/${FLATPAK_ID}.desktop ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop
+      - install -D -m644 studio/Assets/${FLATPAK_ID}.svg ${FLATPAK_DEST}/share/icons/hicolor/scalable/apps/${FLATPAK_ID}.svg
+      - install -D -m644 studio/${FLATPAK_ID}.metainfo.xml ${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.metainfo.xml
+      - cp -r studio ${FLATPAK_DEST}/bin/studio
+    sources:
+      - type: git
+        url: https://github.com/one-ware/OneWare.git
+        disable-shallow-clone: true
+        branch: main
```
bbhtt, line 47: This needs a tag or at least a commit https://docs.flatpak.org/en/latest/module-sources.html#git-sources

HendrikMennen:
bot, build com.one_ware.OneWare

bbhtt:
Is there any part still proprietary? The license and tag in metainfo says that, I thought you said something about open sourcing it.

HendrikMennen:
changed it to GPL

HendrikMennen:
bot, build com.one_ware.OneWare

com.one_ware.OneWare.yml
```diff
@@ -0,0 +1,49 @@
...
+      env:
+        RUNTIME: "linux-arm64"
+    x86_64:
+      env:
+        RUNTIME: "linux-x64"
+
+command: oneware
+separate-locales: false
+tags:
+  - proprietary
```
bbhtt, line 22:
```suggestion
```

HendrikMennen: ah I missed this one, I changed it in the metainfo

bbhtt:
bot, build com.one_ware.OneWare

PR #5392, opened 2024-07-07
Add org.texmacs.TeXmacs

bztd:

- [x] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about this submission. **Link:** https://mail.gnu.org/archive/html/texmacs-dev/2024-07/msg00000.html
a text editor that uses latex in the background and with which you can interact directly with multiple CAS and other types of software.

### DIFF
--- a/org.texmacs.TeXmacs.appdata.xml
+++ b/org.texmacs.TeXmacs.appdata.xml
@@ -9,58 +9,52 @@
     <developer id="org.texmacs">
    	 <name>TeXmacs Team</name>
     </developer>
+
     <description>
 	<p>
-  	    
-	</p>
-	<p>
-	    
-	</p>
-   	<p>
-	    
+  	    TeXmacs is a free scientific editing platform designed
+            to create beautiful technical documents using LaTeX in 
+            the background. It provides a unified and user friendly 
+            framework for editing structured documents with different 
+            types of content: text, mathematics, graphics, interactive 
+            content, slides, etc.
 	</p>
     </description>
 
     <screenshots>
         <screenshot type="default">
-          <image>https://bztd.github.io/LyX/</image>
-          <caption>The startup screen of TeXmacs</caption>
+          <image>https://www.texmacs.org/tmweb/images/collage-full.png</image>
+          <caption>Welcome to TeXmacs!</caption>
         </screenshot>
 	    
         <screenshot>
-          <image>https://bztd.github.io/</image>
-        </screenshot>
-	    
-        <screenshot>
-          <image>https://bztd.github.io/</image>
+          <image>https://www.texmacs.org/tmweb/images/texmacs-micro-typography.png</image>
         </screenshot>
 
-	<screenshot>
-          <image>https://bztd.github.io/LyX/</image>
-          <caption></caption>
+        <screenshot>
+          <image>https://www.texmacs.org/tmweb/images/texmacs-latex-import.png</image>
+          <caption>Importing from LaTeX</caption>
         </screenshot>
+	    
     </screenshots>
 
     <categories>
         <category>Science</category>
         <category>Education</category>
     </categories>
+
     <url type="homepage">https://www.texmacs.org</url>
     <url type="help">http://forum.texmacs.cn/</url>
-    <kudos>
-        <kudo>HiDpiIcon</kudo>
-        <kudo>HighContrast</kudo>
-        <kudo>UserDocs</kudo>
-    </kudos>
+    
     <launchable type="desktop-id">texmacs.desktop</launchable>
     <provides>
         <binary>texmacs</binary>
     </provides>
-    <translation type="gettext">LyX</translation>
+    <translation type="gettext">TeXmacs</translation>
     <content_rating type="oars-1.1">
         <content_attribute id="social-info">mild</content_attribute>
     </content_rating>
     <releases>
-        <release version="2.4.1" date="2024-06-02"></release>
+        <release version="2.1.4" date="2024-06-02"></release>
     </releases>
 </component>
--- a/org.texmacs.TeXmacs.appdata.xml
+++ b/org.texmacs.TeXmacs.appdata.xml
@@ -1,0 +1,66 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2012-2024 texmacs.org -->
+<component type="desktop">
+    <id>org.texmacs.TeXmacs</id>
+    <metadata_license>CC0-1.0</metadata_license>
+    <project_license>GPL-2.0-or-later</project_license>
+    <name>TeXmacs</name>
+    <summary>Scientific Document Editor</summary>
+    <developer id="org.texmacs">
+   	 <name>TeXmacs Team</name>
+    </developer>
+    <description>
+	<p>
+  	    
+	</p>
+	<p>
+	    
+	</p>
+   	<p>
+	    
+	</p>
+    </description>
+
+    <screenshots>
+        <screenshot type="default">
+          <image>https://bztd.github.io/LyX/</image>
+          <caption>The startup screen of TeXmacs</caption>
+        </screenshot>
+	    
+        <screenshot>
+          <image>https://bztd.github.io/</image>
+        </screenshot>
+	    
+        <screenshot>
+          <image>https://bztd.github.io/</image>
+        </screenshot>
+
+	<screenshot>
+          <image>https://bztd.github.io/LyX/</image>
+          <caption></caption>
+        </screenshot>
+    </screenshots>
+
+    <categories>
+        <category>Science</category>
+        <category>Education</category>
+    </categories>
+    <url type="homepage">https://www.texmacs.org</url>
+    <url type="help">http://forum.texmacs.cn/</url>
+    <kudos>
+        <kudo>HiDpiIcon</kudo>
+        <kudo>HighContrast</kudo>
+        <kudo>UserDocs</kudo>
+    </kudos>
+    <launchable type="desktop-id">texmacs.desktop</launchable>
+    <provides>
+        <binary>texmacs</binary>
+    </provides>
+    <translation type="gettext">LyX</translation>
+    <content_rating type="oars-1.1">
+        <content_attribute id="social-info">mild</content_attribute>
+    </content_rating>
+    <releases>
+        <release version="2.4.1" date="2024-06-02"></release>
+    </releases>
+</component>
--- a/org.texmacs.TeXmacs.appdata.xml
+++ b/org.texmacs.TeXmacs.appdata.xml
@@ -19,6 +19,10 @@
             types of content: text, mathematics, graphics, interactive 
             content, slides, etc.
 	</p>
+        <p>
+            This package contains:
+                TeXLive, Gnuplot, SymPy, Graphviz.
+        </p>
     </description>
 
     <screenshots>
--- a/org.texmacs.TeXmacs.yml
+++ b/org.texmacs.TeXmacs.yml
@@ -2,7 +2,7 @@ app-id: org.texmacs.TeXmacs
 runtime: org.kde.Platform
 runtime-version: "5.15-23.08"
 sdk: org.kde.Sdk
-command: texmacs
+command: startTeXmacs.sh
 rename-desktop-file: texmacs.desktop
 rename-icon: texmacs
 finish-args:
@@ -11,10 +11,6 @@ finish-args:
   - --share=ipc
   - --filesystem=xdg-documents
   - --persist=.TeXmacs
-  - --env=PATH=/app/texlive/bin/:/app/texlive/bin/x86_64-linux:/app/texlive/bin/aarch64-linux:/app/bin:/usr/bin
-  - --env=ASPELL_CONF=dict-dir /app/share/dicts
-  - --env=LD_LIBRARY_PATH=/app/texlive/lib:/app/texlive/lib/perl5/5.38.2/aarch64-linux/CORE:/app/texlive/lib/perl5/5.38.2/x86_64-linux/CORE
-  - --env=PERL5LIB=/app/texlive/lib/perl5/5.38.2:/app/texlive/lib/perl5/site_perl/5.38.2
 
 add-extensions:
   org.freedesktop.Sdk.Extension.texlive:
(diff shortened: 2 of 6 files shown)


bztd:
bot, build org.texmacs.TeXmacs

bztd:
bot, build org.texmacs.TeXmacs

bztd:
bot, build org.texmacs.TeXmacs

org.texmacs.TeXmacs.yml
```diff
@@ -0,0 +1,116 @@
...
+  - --env=LD_LIBRARY_PATH=/app/texlive/lib:/app/texlive/lib/perl5/5.38.2/aarch64-linux/CORE:/app/texlive/lib/perl5/5.38.2/x86_64-linux/CORE
+  - --env=PERL5LIB=/app/texlive/lib/perl5/5.38.2:/app/texlive/lib/perl5/site_perl/5.38.2
+
+add-extensions:
+  org.freedesktop.Sdk.Extension.texlive:
+    version: "23.08"
+    directory: texlive
+    autodelete: false
+  org.lyx.LyX.Plugin:
+    versions: stable;beta;test
```
hfiguiere, line 25:
```suggestion
    versions: stable
```

This should be only ONE. Or there is actually no point having versions.

org.texmacs.TeXmacs.yml
```diff
@@ -0,0 +1,116 @@
...
+    config-opts:
+      - --disable-static 
+      - --disable-error-on-warning
+    sources:
+      - type: archive
+        url: http://ftp.gnu.org/gnu/guile/guile-1.8.8.tar.gz
+        sha256: c3471fed2e72e5b04ad133bbaaf16369e8360283679bcf19800bc1b381024050
+      - type: patch
+        path: guile-1.8.8-configure.patch
+  - name: freetype
```
hfiguiere, line 66: there is freetype2 2.16 in the runtime. You should need to build it.

org.texmacs.TeXmacs.yml
```diff
@@ -0,0 +1,116 @@
...
+      - make datadir=/app/share/dicts dictdir=/app/share/dicts install
+    sources:
+      - type: archive
+        url: https://ftp.gnu.org/gnu/aspell/dict/en/aspell6-en-2020.12.07-0.tar.bz2
+        sha256: 4c8f734a28a088b88bb6481fcf972d0b2c3dc8da944f7673283ce487eac49fb3
+
+  - name: xdg # temporary
+    buildsystem: simple
+    build-commands:
+      - ar x xdg-utils_1.1.3-4.1_all.deb
```
hfiguiere, line 97: no. you should buid this from source. not unpacking a .deb.

org.texmacs.TeXmacs.yml
```diff
@@ -0,0 +1,116 @@
...
+      - tar -xJf data.tar.xz
+      - cp -ru ./usr/bin/* /app/bin/
+    sources:
+      - type: file
+        url: http://ftp.de.debian.org/debian/pool/main/x/xdg-utils/xdg-utils_1.1.3-4.1_all.deb
+        sha256: 0e31caa8c34643f7eedb4d373ee61943c09061275b5fa727524fc568d0a9e332
+  - name: texmacs
+    buildsystem: autotools
+    post-install:
+      - rm /app/bin/xdg-*
```
hfiguiere, line 107: then why do you build xdg?

org.texmacs.TeXmacs.yml
```diff
@@ -0,0 +1,116 @@
...
+    buildsystem: autotools
+    sources:
+      - type: archive
+        url: https://download.savannah.gnu.org/releases/freetype/freetype-2.13.2.tar.xz
+        sha256: 12991c4e55c506dd7f9b765933e62fd2be2e06d421505d7950a132e4f1bb484d
+
+  - name: dictEs-aspell
+    buildsystem: simple
+    build-commands:
+      - ./configure
```
hfiguiere, line 76: should this be `buildsystem: autotools` ?

org.texmacs.TeXmacs.yml
```diff
@@ -0,0 +1,116 @@
...
+      - make datadir=/app/share/dicts dictdir=/app/share/dicts
+      - make datadir=/app/share/dicts dictdir=/app/share/dicts install
+    sources:
+      - type: archive
+        url: https://ftp.gnu.org/gnu/aspell/dict/es/aspell6-es-1.11-2.tar.bz2
+        sha256: ad367fa1e7069c72eb7ae37e4d39c30a44d32a6aa73cedccbd0d06a69018afcc
+  - name: dictEn-aspell
+    buildsystem: simple
+    build-commands:
+      - ./configure
```
hfiguiere, line 86: should this be `buildsystem: autotools` ?

org.texmacs.TeXmacs.yml
bztd: They are temporary files, used in the installation of texmacs.

org.texmacs.TeXmacs.yml
bztd: Since they are scripts and temporary, I don't think there is a need to compile.

org.texmacs.TeXmacs.yml
hfiguiere: then a `cleanup` with `*` in the corresponding module.

org.texmacs.TeXmacs.yml
hfiguiere: and? Use the tarball. Not a deb.

bztd:
bot, build org.texmacs.TeXmacs

org.texmacs.TeXmacs.yml
bztd: the --prefix gives an error.

org.texmacs.TeXmacs.yml
hfiguiere: apparently not. Despite being GNU.

bztd:
bot, build org.texmacs.TeXmacs

bztd:
bot, build org.texmacs.TeXmacs

bztd:
bot, build org.texmacs.TeXmacs

org.texmacs.TeXmacs.yml
```diff
@@ -0,0 +1,118 @@
...
+  org.lyx.LyX.Plugin:
+    versions: stable
+    directory: plugins
+    subdirectories: true
+    no-autodownload: true
+    autodelete: true
+    add-ld-path: lib
+
+cleanup:
+  - /bin/xdg-*
```
hfiguiere, line 33: this should be `*` BUT as a local rule to `xdg-utils`

bztd:
bot, build org.texmacs.TeXmacs

org.texmacs.TeXmacs.yml
```diff
@@ -0,0 +1,116 @@
...
+command: texmacs
+rename-desktop-file: texmacs.desktop
+rename-icon: texmacs
+finish-args:
+  - --socket=x11
+  - --device=dri
+  - --share=ipc
+  - --filesystem=xdg-documents
+  - --persist=.TeXmacs
+  - --env=PATH=/app/texlive/bin/:/app/texlive/bin/x86_64-linux:/app/texlive/bin/aarch64-linux:/app/bin:/usr/bin
```
bbhtt, line 14: You should set this through a wrapper.

org.texmacs.TeXmacs.yml
```diff
@@ -0,0 +1,116 @@
...
+  - --env=ASPELL_CONF=dict-dir /app/share/dicts
+  - --env=LD_LIBRARY_PATH=/app/texlive/lib:/app/texlive/lib/perl5/5.38.2/aarch64-linux/CORE:/app/texlive/lib/perl5/5.38.2/x86_64-linux/CORE
+  - --env=PERL5LIB=/app/texlive/lib/perl5/5.38.2:/app/texlive/lib/perl5/site_perl/5.38.2
```
bbhtt, line 17: These two too, the perl version in the SDK changing will break this.

org.texmacs.TeXmacs.yml
```diff
@@ -0,0 +1,116 @@
...
+rename-desktop-file: texmacs.desktop
+rename-icon: texmacs
+finish-args:
+  - --socket=x11
+  - --device=dri
+  - --share=ipc
+  - --filesystem=xdg-documents
+  - --persist=.TeXmacs
+  - --env=PATH=/app/texlive/bin/:/app/texlive/bin/x86_64-linux:/app/texlive/bin/aarch64-linux:/app/bin:/usr/bin
+  - --env=ASPELL_CONF=dict-dir /app/share/dicts
```
bbhtt, line 15: This is not a valid env var

org.texmacs.TeXmacs.yml
```diff
@@ -0,0 +1,116 @@
...
+  - --env=LD_LIBRARY_PATH=/app/texlive/lib:/app/texlive/lib/perl5/5.38.2/aarch64-linux/CORE:/app/texlive/lib/perl5/5.38.2/x86_64-linux/CORE
+  - --env=PERL5LIB=/app/texlive/lib/perl5/5.38.2:/app/texlive/lib/perl5/site_perl/5.38.2
+
+add-extensions:
+  org.freedesktop.Sdk.Extension.texlive:
+    version: "23.08"
+    directory: texlive
+    autodelete: false
+  org.lyx.LyX.Plugin:
```
bbhtt, line 24: Is this correct? Why is it using extensions of another app?

org.texmacs.TeXmacs.yml
```diff
@@ -0,0 +1,116 @@
...
+    build-commands:
+      - install -d /app/texlive
+      - install -d /app/plugins
+
+  - name: dictEs-aspell
+    buildsystem: simple
+    build-commands:
+      - ./configure
+      - make dictdir=/app/share/dicts datadir=/app/share/datadicts
+      - make dictdir=/app/share/dicts datadir=/app/share/datadicts install
```
bbhtt, line 47: Use autotools and `make-args` `make-install-args`.

org.texmacs.TeXmacs.yml
```diff
@@ -0,0 +1,116 @@
...
+        sha256: c3471fed2e72e5b04ad133bbaaf16369e8360283679bcf19800bc1b381024050
+      - type: patch
+        path: guile-1.8.8-configure.patch
+
+  - name: xdg-utils # temporary #fake
+    buildsystem: simple
+    build-commands:
+      - touch /app/bin/xdg-icon-resource /app/bin/xdg-mime /app/bin/xdg-desktop-menu 
+      - touch /app/bin/xdg-desktop-icon /app/bin/xdg-icon-resource
+      - chmod +x /app/bin/xdg-*
```
bbhtt, line 102: What is this for?

org.texmacs.TeXmacs.yml
```diff
@@ -0,0 +1,116 @@
...
+    sources:
+      - type: archive
+        url: https://ftp.gnu.org/gnu/aspell/dict/es/aspell6-es-1.11-2.tar.bz2
+        sha256: ad367fa1e7069c72eb7ae37e4d39c30a44d32a6aa73cedccbd0d06a69018afcc
+  - name: dictEn-aspell
+    buildsystem: simple
+    build-commands:
+      - ./configure
+      - make dictdir=/app/share/dicts datadir=/app/share/datadicts
+      - make dictdir=/app/share/dicts datadir=/app/share/datadicts install
```
bbhtt, line 57: Use autotools and make-args make-install-args.

org.texmacs.TeXmacs.appdata.xml
bbhtt, line 1: Has this been submitted upstream?

org.texmacs.TeXmacs.yml
bztd: Yes, I thought that, but it works. The runtime has support for Huspell and has Aspell installed but does not have dictionaries. If not this way I wouldn't know how.

org.texmacs.TeXmacs.yml
bztd: In this case "--prefix=.." is injected, which causes the configuration script to fail.

org.texmacs.TeXmacs.yml
bztd: Some installation scripts use xdg-utils, but in this case they have no effect, the installation is aborted if they are not found.

org.texmacs.TeXmacs.appdata.xml
bztd: I already make the request

org.texmacs.TeXmacs.yml
bbhtt: Add a shell wrapper

```
export PATH=/app/texlive/bin:/app/texlive/bin/$(gcc --print-multiarch):${PATH}

exec command_here "$@"
```

instead of setting it statically here

Same with perl. use `perl -e'print substr($^V, 1)'` to get the version.

org.texmacs.TeXmacs.yml
bbhtt: You should put it under quotes then this weird.

org.texmacs.TeXmacs.yml
bbhtt: Weird.

org.texmacs.TeXmacs.yml
bztd: I understand now. I thought doing this kind of thing was frowned upon.

bbhtt: The other solution would be to symlink everything from `/app/texlive/bin/x86_64-linux` to `/app/bin`.

hfiguiere: > The other solution would be to symlink everything from `/app/texlive/bin/x86_64-linux` to `/app/bin`.

this is not how it's used elsewhere - due to how the SDK is built. The wrapper script is a good way.

bztd:
bot, build org.texmacs.TeXmacs

bztd:
bot, build org.texmacs.TeXmacs

bbhtt:
Looks good.

bztd:
Thanks for the contributions, I will also make modifications in LyX.

PR #5399, opened 2024-07-09
Add org.gnome.Showtime

kra-mo:
A PyGObject GStreamer video player app.

### DIFF
--- a/org.gnome.Showtime.json
+++ b/org.gnome.Showtime.json
@@ -1,0 +1,90 @@
+{
+  "id": "org.gnome.Showtime",
+  "runtime": "org.gnome.Platform",
+  "runtime-version": "46",
+  "sdk": "org.gnome.Sdk",
+  "sdk-extensions": [
+    "org.freedesktop.Sdk.Extension.rust-stable",
+    "org.freedesktop.Sdk.Extension.llvm16"
+  ],
+  "build-options": {
+    "append-path": "/usr/lib/sdk/rust-stable/bin"
+  },
+  "command": "showtime",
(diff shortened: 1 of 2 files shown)


org.gnome.Showtime.json
```diff
@@ -0,0 +1,91 @@
+{
+  "id": "org.gnome.Showtime",
```
hfiguiere, line 2: is this an official GNOME app?

org.gnome.Showtime.json
```diff
@@ -0,0 +1,91 @@
...
+  },
+  "command": "showtime",
+  "finish-args": [
+    "--share=ipc",
+    "--socket=fallback-x11",
+    "--device=dri",
+    "--socket=wayland",
+    "--socket=pulseaudio",
+    "--filesystem=xdg-pictures/Screenshots:create",
+    "--own-name=org.mpris.MediaPlayer2.org.gnome.Showtime"
```
hfiguiere, line 21: this name is actually granted by default. (based on the  app-id)

org.gnome.Showtime.json
kra-mo: Yes, it is in [Incubator](https://gitlab.gnome.org/GNOME/Incubator).

org.gnome.Showtime.json
hfiguiere: I don't see what the doc properly update. A lot of app don't use the proper naming which is why they need to own the name.

bbhtt:
bot, build org.gnome.Showtime

PR #5400, opened 2024-07-09
Add org.kuchelmeister.ToolboxTuner

13hannes11:
### Description
A graphical application intended for use on Fedora Silverblue to manage containertoolbx.org containers. To do this the application interfaces with podman, toolbox and gnome-terminal.


13hannes11:
This application has been submitted once before during earlier stages of development but I renamed it slightly: https://github.com/flathub/flathub/pull/3397

org.kuchelmeister.ToolboxTuner.json
```diff
@@ -0,0 +1,61 @@
...
+        "org.freedesktop.Sdk.Extension.rust-stable",
+        "org.freedesktop.Sdk.Extension.llvm15"
+    ],
+    "command" : "toolbox-tuner",
+    "finish-args" : [
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--device=dri",
+        "--socket=wayland",
+        "--talk-name=org.freedesktop.Flatpak"
```
hfiguiere, line 16: why is this necessary.

org.kuchelmeister.ToolboxTuner.json
```diff
@@ -0,0 +1,61 @@
...
+    ],
+    "build-options" : {
+        "append-path" : "/usr/lib/sdk/rust-stable/bin",
+        "build-args" : []
```
hfiguiere, line 20:
```suggestion
        "append-path" : "/usr/lib/sdk/rust-stable/bin"
```

org.kuchelmeister.ToolboxTuner.json
```diff
@@ -0,0 +1,61 @@
...
+                    "CARGO_HOME": "/run/build/toolbox-tuner/cargo"
+                }
+            },
+            "config-opts": [
+                "-Dprofile=flathub"
+            ],
+            "sources" : [
+                {
+                    "type" : "archive",
+                    "url" : "https://github.com/13hannes11/toolbox-tuner/archive/refs/tags/dev.tar.gz",
```
hfiguiere, line 49: "dev" as a tag. I guess there is no release yet?

org.kuchelmeister.ToolboxTuner.json
```diff
@@ -0,0 +1,61 @@
+{
+    "id" : "org.kuchelmeister.ToolboxTuner.Devel",
```
hfiguiere, line 2:
```suggestion
    "id" : "org.kuchelmeister.ToolboxTuner.Devel",
```

this doesn't match. `Devel` isn't a proper app name.

hfiguiere:
This is missing

> Please describe the application briefly.

org.kuchelmeister.ToolboxTuner.json
13hannes11: The application needs to talk to certain host applications via the commandline. These are _toolbox_ and _podman_.  Also it is used to open the terminal. The commands used for achieving this are usually something like: 
* `flatpak-spawn --host gnome-terminal --version`
* `flatpak-spawn --host podman stop <hash>`
* `flatpak-spawn --host toolbox list`

I'd appreciate suggestions on how to avoid using  "--talk-name=org.freedesktop.Flatpak"

org.kuchelmeister.ToolboxTuner.json
13hannes11: This was an accidental copy paste leftover from the development config. I fixed it in 77bcfab7c6b

org.kuchelmeister.ToolboxTuner.json
13hannes11: I was planning to have the release and review in parallel. in principle the software is ready for release but it might be that this review surfaces some problems that need fixing. Thus, I was planning to not yet have a full release build.

hfiguiere:
> A graphical application intended for use on Fedora Silverblue

What happen on other distros?

hfiguiere:
bot, build org.kuchelmeister.ToolboxTuner

13hannes11:
I realised there was an issue with the buildprofile fixed by [b389184](https://github.com/flathub/flathub/pull/5400/commits/b389184342a2c814d5e15457e09c17f67dfd915a)

org.kuchelmeister.ToolboxTuner.json
13hannes11: The version v0.1.0 is now published

13hannes11:
> > A graphical application intended for use on Fedora Silverblue
> 
> What happen on other distros?

The application checks for the presence of toolbox and gnome-terminal. If these applicatiions are not present a window this window is displayed:
image

In the future the applicatioin should support a wider range of terminals.

org.kuchelmeister.ToolboxTuner.json
```diff
@@ -0,0 +1,59 @@
+{
+    "id" : "org.kuchelmeister.ToolboxTuner",
+    "runtime" : "org.gnome.Platform",
+    "runtime-version" : "46",
+    "sdk" : "org.gnome.Sdk",
+    "sdk-extensions" : [
+        "org.freedesktop.Sdk.Extension.rust-stable",
+        "org.freedesktop.Sdk.Extension.llvm15"
```
Erick555, line 8: are you sure you need such old version of llvm? latest one is llvm18

13hannes11: Good catch! Bumped the version to llvm18 and ran a testbuild locally.

org.kuchelmeister.ToolboxTuner.json
bbhtt: You need to ask an exception to this https://docs.flathub.org/docs/for-app-authors/linter#exceptions

The error code is https://docs.flathub.org/docs/for-app-authors/linter#finish-args-flatpak-spawn-access

13hannes11:
A bit strange that flatpak builder locally did not complain about the checksums but when I download just now from github the checksums are different than the ones in the manifest.

I updated the checksums to match the release output.

13hannes11:
bot, build org.kuchelmeister.ToolboxTuner

13hannes11:
bot, build org.kuchelmeister.ToolboxTuner

hfiguiere:
> A bit strange that flatpak builder locally did not complain about the checksums but when I download just now from github the checksums are different than the ones in the manifest.
> 
> I updated the checksums to match the release output.

This is what happen when you retag or serve file that change because they are not version.  Retagging is bad.

flatpak-builder doesn't redownload what it already has. Since you didn't change the filename, it didn't try to redownload it.

13hannes11:
> > A bit strange that flatpak builder locally did not complain about the checksums but when I download just now from github the checksums are different than the ones in the manifest.
> > I updated the checksums to match the release output.
> 
> This is what happen when you retag or serve file that change because they are not version. Retagging is bad.
> 
> flatpak-builder doesn't redownload what it already has. Since you didn't change the filename, it didn't try to redownload it.

I was not aware of this thanks for the explanation.

13hannes11:
Sorry for the double build I somehow thought the previous one had failed 😅

bbhtt:
Can you please add support for a few different terminals if that's possible? I think GNOME console is shipped by default nowadays instead of GNOME terminal. Some on the KDE side would be nice to.

Also the dialogue box doesn't detect the presence of toolbox and the terminal separately. If one is installed but not the other one - the dialogue says both are missing. Probably a good idea to just say what is missing.

PR #5408, opened 2024-07-11
add io.github.getnf.embellish

ronniedroid:

  A graphical interface built with GTK4 and libadwaita that helps the user install nerd fonts and manage them on their system. Written in go.


io.github.getnf.embellish.yml
```diff
@@ -0,0 +1,37 @@
...
+
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.golang
+
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc
+  - --share=network
+  - --filesystem=xdg-data/embellish:create
```
bbhtt, line 15:
```suggestion
```

This won't be needed. Flatpak sets it to `~/.var/app/$FLATPAK_ID/data`

io.github.getnf.embellish.yml
```diff
@@ -0,0 +1,37 @@
...
+  - --filesystem=xdg-data/fonts:create
```
bbhtt, line 16:
```suggestion
```

Fonts from here are automounted by Flatpak, not needed.

bbhtt, line 16: Why do you need this? Fonts from here are automounted by Flatpak.

io.github.getnf.embellish.yml
bbhtt: This should be XDG base directory compliant https://github.com/getnf/embellish/blob/bbec1d3f4796c162b9901577a7a45351971a4d63/internal/types/types.go#L114

You are supposed to access this path relative to XDG_DATA_HOME not HOME.

ronniedroid: Embellish also has a CLI/TUI binary available on github, and I am thinking it would be nice if both the GUI and TUI/CLI builds have access to the same db.

If you still don't think it is necessary, I will change it.

io.github.getnf.embellish.yml
ronniedroid: For the same reason of the GUI being compatable with the TUI/CLI build. 

Also, will fonts automounted by flatpak be acceessible by non-flatpak apps?

io.github.getnf.embellish.yml
bbhtt: Package the CLI in the Flatpak too? You can run it with `flatpak run --command=cli-command <appid>`

io.github.getnf.embellish.yml
bbhtt: > Also, will fonts automounted by flatpak be acceessible by non-flatpak apps?

Not sure what you mean, the sandbox is only for this app, the font files are already on host, flatpak will bind mount them inside the sandbox.

As the fonts are already on host they are accessible by other host apps.

ronniedroid: Oh i misunderstood you.

I need access to the fonta directory because my tool is used to install and uninstall fonta from the system.

My tool won't be able to function without this access.

io.github.getnf.embellish.yml
ronniedroid: But then each tool has its own data home and won't be able too see the same db file.

bbhtt: I mean in this flatpak. Then they will share the same `XDG_DATA_HOME` which is `~/.var/app/$FLATPAK_ID/data`

io.github.getnf.embellish.yml
bbhtt: Does `--filesystem=~/.local/share/fonts` work or do you need it to be xdg-data/fonts?

ronniedroid: `--filesystem=~/.local/share/fonts` does work

bbhtt: So can you change this one and drop the other one?

bbhtt:
bot, build io.github.getnf.embellish

bbhtt:
Appstream needs some fixing, if you follow the links there's more info in the docs site.

ronniedroid:
All appstream issues were fixed, sorry, I had used the mateinfo generator so I thought I was good to go.

bbhtt:
bot, build io.github.getnf.embellish

bbhtt:
Where are the fonts being installed? I installed the first one and I don't see anything in `~/.local/share/fonts`

bbhtt:
Ok seems like it doesn't work if the directory is already not present on the system.

io.github.getnf.embellish.yml
```diff
@@ -0,0 +1,38 @@
...
+
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.golang
+
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc
+  - --share=network
+  - --filesystem=~/.local/share/fonts
```
bbhtt, line 15:
```suggestion
  - --filesystem=~/.local/share/fonts:create
```

bbhtt:
bot, build io.github.getnf.embellish

PR #5412, opened 2024-07-13
Add net.sourceforge.symphonic.jsymphonic

georgewoodall82:

  A file manager for old sony network walkman devices

- [ ] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about this submission. **Link:** unable to contact upstream as the program is quite old, would i need to make a fork?


flathub.json
bbhtt, line 1: Why is this x86_64 only?

net.sourceforge.symphonic.jsymphonic.desktop
bbhtt, line 1: This should be in the upstream repo

net.sourceforge.symphonic.jsymphonic.metainfo.xml
bbhtt, line 1: This should be in the upstream repo

net.sourceforge.symphonic.jsymphonic.png
bbhtt, line 1: This should be in the upstream repo

net.sourceforge.symphonic.jsymphonic.yml
```diff
@@ -0,0 +1,37 @@
+id: net.sourceforge.symphonic.jsymphonic
+runtime: org.freedesktop.Platform
+runtime-version: '23.08'
+finish-args:
+  - --socket=x11
+  - --share=ipc
+  - --filesystem=host
```
bbhtt, line 7: Why is host needed?

net.sourceforge.symphonic.jsymphonic.yml
```diff
@@ -0,0 +1,37 @@
...
+      - install -D ${FLATPAK_ID}.png /app/share/icons/hicolor/512x512/apps/${FLATPAK_ID}.png
+      - install -D ${FLATPAK_ID}.desktop /app/share/applications/${FLATPAK_ID}.desktop
+      - install -Dm755 jsymphonic.sh /app/bin/jsymphonic
+      - install -D *.jar /app/jsymphonic/jsymphonic.jar
+    sources:
+      - type: script
+        dest-filename: jsymphonic.sh
+        commands:
+          - mkdir -p /var/config/jsymphonic
+          - cd /var/config/jsymphonic/
```
bbhtt, line 27: Why is this required? Why can't the app initiate its own config directory?

bbhtt:
The upstream status says abandonded https://sourceforge.net/projects/symphonic/ ?

flathub.json
georgewoodall82: i just removed the json file, it should build for arm64 too now

net.sourceforge.symphonic.jsymphonic.desktop
georgewoodall82: I have no access to the upstream repo as it is abandoned

net.sourceforge.symphonic.jsymphonic.yml
georgewoodall82: it needs to access usb for the sony network walkman, and the home folder for music files to import and export. it doesn't support portals.

net.sourceforge.symphonic.jsymphonic.yml
georgewoodall82: it creates the config in the same directory that it is run, so the config will be different for every place you run it by default. this makes it the same regardless of where it is ran.

georgewoodall82:
> The upstream status says abandonded https://sourceforge.net/projects/symphonic/ ?

Yeah it is, do i need to specify that in the metainfo.xml?

net.sourceforge.symphonic.jsymphonic.yml
hfiguiere: `host` doesn't give access to USB.

Ad for music. `xdg-music`

net.sourceforge.symphonic.jsymphonic.yml
hfiguiere: this is not persistent in any way.

net.sourceforge.symphonic.jsymphonic.yml
```diff
@@ -0,0 +1,37 @@
...
+          - /app/jre/bin/java -jar /app/jsymphonic/jsymphonic.jar
+      - type: archive
+        url: https://downloads.sourceforge.net/project/symphonic/jsymphonic/JSymphonic version 0.3.0 Ode To Freedom/JSymphonic_v0.3.0_Ode_To_Freedom.zip
```
hfiguiere, line 30: what is this and why is it needed?

net.sourceforge.symphonic.jsymphonic.yml
georgewoodall82: Usbs are usually mounted at /run/media or /media which host gives access to

net.sourceforge.symphonic.jsymphonic.yml
georgewoodall82: Would it be better to leave it as the default behaviour? (Creating the config in the current directory)

net.sourceforge.symphonic.jsymphonic.yml
hfiguiere: That's not USB. That's removable drives.

net.sourceforge.symphonic.jsymphonic.yml
hfiguiere: no. Fixing the code is the best.

bbhtt:
> Yeah it is, do i need to specify that in the metainfo.xml?

No that's fine. Are you willing to maintain this/patch things, if necessary to keep the Flatpak updated and working? Every year or so you have to bump the runtime to a major version as old ones are EOL-ed.

Upstream being gone means you likely need to patch it in times like that.

georgewoodall82:
just made a fork that may fix these issues, ill make a new pull request

PR #5421, opened 2024-07-16
Migrate FontManager with upstream App-Id change

Eonfge:
This is an existing app, that I've been maintaining for some time. This is just an administrative change based on this issue: https://github.com/flathub/org.gnome.FontManager/issues/40

- [X] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about this submission. [**Link**](https://github.com/FontManager/font-manager/issues/319)

### DIFF
--- a/com.github.FontManager.FontManager.yaml
+++ b/com.github.FontManager.FontManager.yaml
@@ -1,0 +1,91 @@
+app-id: com.github.FontManager.FontManager
+runtime: org.gnome.Platform
+runtime-version: '46'
+sdk: org.gnome.Sdk
+
+command: font-manager
+
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc
+  - --share=network
+
+  # XDG Data/Fonts location
+  - --filesystem=xdg-data/fonts
+  - --persist=xdg-data/fonts
+
+  # XDG complient locations for font config
+  - --filesystem=xdg-config/fontconfig
+  - --persist=xdg-config/fontconfig
+
+cleanup:
+  - /app/include
+  - /app/lib/*.a
+  - /app/lib/*.la
+  - /app/lib/pkgconfig
+
+modules:
+
+  - shared-modules/intltool/intltool-0.51.json
+  - shared-modules/libsoup/libsoup-2.4.json
+
+  # Borrowed from Poedit
+  # https://github.com/flathub/net.poedit.Poedit
+  - name: unifdef
+    no-autogen: true
+    make-install-args:
+      - prefix=${FLATPAK_DEST}
+    sources:
+      - type: archive
+        url: https://dotat.at/prog/unifdef/unifdef-2.12.tar.xz
+        sha256: 43ce0f02ecdcdc723b2475575563ddb192e988c886d368260bc0a63aee3ac400
+    cleanup:
+      - '*'
+
+  - name: backtrace
+    buildsystem: autotools
+    sources:
+      - type: archive
+        url: https://github.com/ianlancetaylor/libbacktrace/archive/7ead8c1ea2f4aeafe9c5b9ef8a9461a9ba781aa8.zip
+        sha256: 3b07ffe9a6302ccfdf5771ab8631814303f5bc15d19a2f5a1907c64a391ec29f
+
+  - name: webkit2gtk-4.0
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DPORT=GTK
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DENABLE_DOCUMENTATION=OFF
+      - -DENABLE_BUBBLEWRAP_SANDBOX=OFF
+      - -DENABLE_WEBDRIVER=OFF
+      - -DENABLE_GAMEPAD=OFF
+    sources:
+      - type: archive
+        url: https://webkitgtk.org/releases/webkitgtk-2.44.2.tar.xz
+        sha256: 523f42c8ff24832add17631f6eaafe8f9303afe316ef1a7e1844b952a7f7521b
+        x-checker-data:
+          type: html
+          url: https://webkitgtk.org/releases/
+          version-pattern: LATEST-STABLE-(\d[\.\d]+\d)
+          url-template: https://webkitgtk.org/releases/webkitgtk-$version.tar.xz
+
+  - name: fontmanager
+    buildsystem: meson
+    config-opts:
+      - -Dreproducible=true
+      - -Dviewer=false
+      - -Dbuildtype=release
+    sources:
+      - type: git
+        url: https://github.com/FontManager/font-manager.git
+        tag: 0.9.0
+        commit: 091d2896d623c3e5aac7613a95da694a0b36699a
+        x-checker-data:
+          type: git
+          tag-pattern: ^([\d.]+)$
+    post-install:
+      # Changes to minetype support are currently on the roadmap
+      - rm /app/share/applications/mimeinfo.cache
+      - sed -i '6 i <developer_name>Font Manager project</developer_name>' /app/share/metainfo/com.github.FontManager.FontManager.appdata.xml
+      - desktop-file-edit /app/share/applications/com.github.FontManager.FontManager.desktop --remove-key=MimeType
+
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+build-dir/
+.flatpak-builder/
+testing-repo/
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "shared-modules"]
+	path = shared-modules
+	url = git@github.com:flathub/shared-modules.git
--- a/run.sh
+++ b/run.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+clear
+flatpak-builder --repo=testing-repo --force-clean build-dir com.github.FontManager.FontManager.yaml
+flatpak --user remote-add --if-not-exists --no-gpg-verify fm-testing-repo testing-repo
+flatpak --user install fm-testing-repo com.github.FontManager.FontManager -y
+flatpak --user install fm-testing-repo com.github.FontManager.FontManager.Debug -y
+flatpak update -y


com.github.FontManager.FontManager.yaml
```diff
@@ -0,0 +1,91 @@
+app-id: com.github.FontManager.FontManager
```
bbhtt, line 1: This should be `io.github.` see the appid rules in checklist. The last part should be `font-manager` to match the repo name.

com.github.FontManager.FontManager.yaml
```diff
@@ -0,0 +1,91 @@
...
+command: font-manager
+
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc
+  - --share=network
+
+  # XDG Data/Fonts location
+  - --filesystem=xdg-data/fonts
```
bbhtt, line 15: Needs exception

Eonfge:
For reference, the application is moving from `org.gnome.FontManager` to `com.github.FontManager.Fontmanager`. This is discussed in this issue upstream(https://github.com/FontManager/font-manager/issues/319) and GNOME is really looking forward to the App-Id change. This is one of those applications which is old enough to pre-date some of the current rules we have related to application Ids.

com.github.FontManager.FontManager.yaml
```diff
@@ -0,0 +1,91 @@
...
+  - --persist=xdg-data/fonts
+
+  # XDG complient locations for font config
+  - --filesystem=xdg-config/fontconfig
```
bbhtt, line 19: Needs exception

com.github.FontManager.FontManager.yaml
```diff
@@ -0,0 +1,91 @@
...
+  - --persist=xdg-config/fontconfig
+
+cleanup:
+  - /app/include
+  - /app/lib/*.a
+  - /app/lib/*.la
```
bbhtt, line 25:
```suggestion
  - "*.a"
  - "*.la"
```

com.github.FontManager.FontManager.yaml
```diff
@@ -0,0 +1,91 @@
...
+  # XDG Data/Fonts location
+  - --filesystem=xdg-data/fonts
+  - --persist=xdg-data/fonts
+
+  # XDG complient locations for font config
+  - --filesystem=xdg-config/fontconfig
+  - --persist=xdg-config/fontconfig
+
+cleanup:
+  - /app/include
```
bbhtt, line 23:
```suggestion
  - /include
```

com.github.FontManager.FontManager.yaml
```diff
@@ -0,0 +1,91 @@
...
+  - /app/lib/*.a
+  - /app/lib/*.la
+  - /app/lib/pkgconfig
```
bbhtt, line 26:
```suggestion
  - /lib/pkgconfig
```

com.github.FontManager.FontManager.yaml
```diff
@@ -0,0 +1,91 @@
...
+        url: https://github.com/FontManager/font-manager.git
+        tag: 0.9.0
+        commit: 091d2896d623c3e5aac7613a95da694a0b36699a
+        x-checker-data:
+          type: git
+          tag-pattern: ^([\d.]+)$
+    post-install:
+      # Changes to minetype support are currently on the roadmap
+      - rm /app/share/applications/mimeinfo.cache
+      - sed -i '6 i <developer_name>Font Manager project</developer_name>' /app/share/metainfo/com.github.FontManager.FontManager.appdata.xml
```
bbhtt, line 89: Use a patch for this, sed will fail and break the metainfo file.

com.github.FontManager.FontManager.yaml
bbhtt: I don't know how upstream chose the `com.github` id... they don't obviously own `github.com`

Eonfge: Discussed upsteam. They want to carry this app Id and they maintain the metadata themselves:
https://github.com/FontManager/font-manager/issues/286

Eonfge: What you want me to do? `org.gnome` also isn't ideal...

bbhtt: Unless the id is fixed this cannot be accepted. New apps cannot have a com.github id. I suggest you contact upstream

bbhtt: Or otherwise patch it to `io.github.fontManager.font-manager` if upstream is unwilling to fix it.

com.github.FontManager.FontManager.yaml
Eonfge: Do I have to submit the exception, or is that something out of my direct control?

bbhtt: You have to submit it https://docs.flathub.org/docs/for-app-authors/linter#exceptions but the id thing should be figured out first. 

You can run the linter on the manifest to get the error code. See #3 https://docs.flathub.org/docs/for-app-authors/submission#before-submission

com.github.FontManager.FontManager.yaml
Eonfge: Distributions like Debian Sid and Arch allow `com.github.FontManager.FontManager` so I doubt that Upsteam will care that much.

Perfect is the enemy of good. I'm afraid that it will take an unreasonable amount of time for me to patch the App-id. In which case, the package will stick with `org.gnome.FontManager`. 

Would this be worth an exception?

com.github.FontManager.FontManager.yaml
Eonfge: No longer applicable, Upsteam has accepted my change and I forgot about it with the 0.9.0 release

com.github.FontManager.FontManager.yaml
bbhtt: No this breaks future verification and the reverse DNS check, with the current ID we calculate the domain to be fontmanager.github.com which does not exist.

bbhtt: @JerryCasiano can you change the ID upstream to `io.github`. An ID should not be chosen on how it looks, it carries meaning that you control the domain. Otherwise anyone could claim a reversed domain as their ID.

This standard has been present for the past 20+ years in Linux.

See https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names

> For instance, if the owner of example.com is developing a D-Bus API for a music player, they might define interfaces called com.example.MusicPlayer1, com.example.MusicPlayer1.Track and com.example.MusicPlayer1.Seekable. 

https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic

> Ownership of {vendor}.{tld} in the domain name system guarantees uniqueness of IDs. 

You were linked to both Flatpak and Flathub documentation for this in https://github.com/FontManager/font-manager/issues/286

JerryCasiano: > You were linked to both Flatpak and Flathub documentation for this in FontManager/font-manager#286 (comment)

Interesting tone, almost feel like a child being scolded.

That said I'm still happy to review a pull request from you.

hfiguiere: Just to clarify, there won't be an exception for the app-id.

bbhtt: > Interesting tone, almost feel like a child being scolded.

No one is scolding you. But this is the second time the app has the wrong ID when you were linked to the docs that clearly says not to use `com.github`. This is just frustrating for everyone.

JerryCasiano: > No one is scolding you. But this is the second time the app has the wrong ID when you were linked to the docs that clearly says not to use com.github. This is just frustrating for everyone.

Second time?

I'm not sure what you're on about. 

I've changed the ID exactly once and it was changed to an ID which is clearly in line with one suggested as acceptable in the very thread you referenced, apparently you missed that part or simply chose to ignore it.


In any case, this is obviously not headed in a productive direction so if someone decides to make the necessary changes, great, I'll be happy to merge before the next point release otherwise oh well.

Unsubscribing from this noise.

bbhtt: That suggestion was wrong the docs clearly said that. And the first id was `org.gnome` per https://github.com/FontManager/font-manager/issues/319 which was also a domain you didn't own.

Anyway not my problem, but this won't get into Flathub unless the ID is fixed either upstream or through a Flathub specific patch.

Eonfge: @JerryCasiano 

> Interesting tone, almost feel like a child being scolded.

I'm sorry that you're being dragged into this. I maintain applications with creative App-Ids like `com.github.Flacon` and `io.github.ezQuake` and `org.gnome.gitlab.Cowsay`, without previous problems. Sadly, at some point in the past 4 years the policy and atmosphere has changed drastically.

I think we've done enough for now, and I'll gladly re-evaluate this issue once you've signed off on a merge request with the appropriate changes.

hfiguiere: > com.github.Flacon and io.github.ezQuake

I can tell you that I am 100% sure these two app-id wouldn't be accepted. They can't even be verified.

Eonfge: > > com.github.Flacon and io.github.ezQuake
> 
> I can tell you that I am 100% sure these two app-id wouldn't be accepted. They can't even be verified.

We're trying to make the world (or at least, Flathub) a better place one small step at the time. Back when I submitted those applications, the rules were less strict. Importantly though, we all collaborated under the assumption that functional applications we're better then no applications. 

My point is not that those applications would now be allowed, but that there was a time when Flathub was more pragmatic. Perfect is the enemy of good.

bbhtt: Those were wrong then, it's still wrong now. The only difference is that it wasn't enforced.

bbhtt: Using a `or.gnome` ID made the GNOME foundation contact the project asking to change, using `com.github` makes it vulnerable to Github/MS down the line.

PR #5426, opened 2024-07-17
Add org.kde.okular.kirigami

jbruechert:
This is based directly on the Okular manifest, and the changes are split into separate commits to make review easy.

  Mobile interface for the Okular PDF viewer

- [x] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about this submission. **Link:** https://matrix.to/#/!iotruiWARuYNvXiTDB:kde.org/$XRybn93-Pn2iBITRe964G2O-FlZk0kSPD4_dLYW1zcc?via=matrix.org&via=kde.org&via=im.kde.org

### DIFF
--- a/0001-mobile-Clip-DocumentView.patch
+++ b/0001-mobile-Clip-DocumentView.patch
@@ -1,0 +1,25 @@
+From 30456eddb351e6b9977b73144329df881143ecfb Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Jonah=20Br=C3=BCchert?= <jbb@kaidan.im>
+Date: Wed, 3 Jul 2024 16:55:07 +0200
+Subject: [PATCH 1/2] mobile: Clip DocumentView
+
+---
+ mobile/components/DocumentView.qml | 2 ++
+ 1 file changed, 2 insertions(+)
+
+diff --git a/mobile/components/DocumentView.qml b/mobile/components/DocumentView.qml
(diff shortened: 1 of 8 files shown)


hfiguiere:
cc @flathub/kde 

Is this a valid app-id?

travier:
This is a valid KDE App. Is the question that we should use another id?

hfiguiere:
the appstream documentation doesn't take flathub into account.

hfiguiere:
No. This is the question I asked in https://github.com/flathub/flathub/pull/5426#issuecomment-2234432799

bbhtt:
> This is a valid KDE App. Is the question that we should use another id?

What ID does it internally use? If it uses a D-Bus name or an internal ID, it should match with that.

bbhtt:
So I guess the ID is fine then.

> Its appstream metadata uses org.kde.okular.kirigami.desktop.

Since provides and replaces works fine, it'd be better if you can add the tags upstream and rename it to `org.kde.okular.kirigami`. https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/#provides

bbhtt:
Seems fine, waiting for the review from flathub/kde.

tsdgeos:
Did you copy/adapt the manifest from the "regular" okular one in flathub?

tsdgeos:
Looks reasonable to me

travier:
Tested, LGTM. Thanks!

PR #5473, opened 2024-08-03
Add rs.ruffle.Ruffle

kjarosh:

    Ruffle is a Flash Player emulator that allows users to play Flash games and movies (SWF files). https://ruffle.rs/

    I've got an error `appstream-external-screenshot-url`, but I assume it's due to building locally.

    I'm a maintainer of Ruffle: https://github.com/orgs/ruffle-rs/people

This PR supersedes https://github.com/flathub/flathub/pull/4179 and is based on @Doomsdayrs's wonderful work.


flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64"]
```
hfiguiere, line 2: so no on aarch64?

rs.ruffle.Ruffle.yaml
```diff
@@ -0,0 +1,45 @@
...
+        dest-filename: rs.ruffle.Ruffle.desktop
+      - type: file
+        url: https://raw.githubusercontent.com/ruffle-rs/ruffle/efe754cdcc57f0e1f2820e66983f6a5a91e5d2c8/desktop/assets/icon.svg
+        sha256: 9ad3452f957260c4b3e23cc9e3b1a6da05cfa36cde2a584031c4a1f3aba15635
+        dest-filename: rs.ruffle.Ruffle.svg
+      # TODO Remove this patch when stable releases are present
+      - type: patch
+        path: patches/release.patch
+      - type: archive
+        url: https://github.com/ruffle-rs/ruffle/releases/download/nightly-2024-08-03/ruffle-nightly-2024_08_03-linux-x86_64.tar.gz
```
hfiguiere, line 39: why no building from source instead?

flathub.json
kjarosh: Currently we do not build nor test on aarch64, however there are some plans for the future to support it.

You can see all the artifacts we build here: https://github.com/ruffle-rs/ruffle/releases

rs.ruffle.Ruffle.yaml
kjarosh: That's something I'm trying to wrap my head around. I wasn't sure how to do it and just left downloading binaries as in the original PR.

I saw https://github.com/flatpak/flatpak-builder-tools/tree/master/cargo, but it seems that it requires manually generating the JSON each release in flathub's repo or publishing it upstream, right?

kjarosh: Okay, I think I got it! Ruffle is being built from sources. Let me know if everything's alright now

flathub.json
bbhtt: If it is built from source, you can remove this file?

rs.ruffle.Ruffle.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - --share=network
+
+modules:
+  - name: ruffle
+    buildsystem: simple
+    build-options:
+      env:
+        CARGO_HOME: /run/build/ruffle/cargo
+    build-commands:
+      - java --version
```
bbhtt, line 26:
```suggestion
```

rs.ruffle.Ruffle.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - cargo --offline fetch --manifest-path Cargo.toml --verbose
+      - cargo --offline build --release --verbose
+      - install -D -T ./desktop/assets/icon.svg /app/share/icons/hicolor/scalable/apps/rs.ruffle.Ruffle.svg
+      - install -D -T ./desktop/packages/linux/rs.ruffle.Ruffle.desktop /app/share/applications/rs.ruffle.Ruffle.desktop
+      - install -D -T ./desktop/packages/linux/rs.ruffle.Ruffle.metainfo.xml /app/share/metainfo/rs.ruffle.Ruffle.metainfo.xml
+      - install -D -T ./target/release/ruffle_desktop /app/bin/ruffle
+    sources:
+      - type: git
+        url: https://github.com/ruffle-rs/ruffle.git
+        tag: nightly-2024-08-04
```
bbhtt, line 36: Is there a stable tag for this? Or do ruffle only publishes nightly tags? 

https://docs.flathub.org/docs/for-app-authors/requirements#stable-releases

flathub.json
kjarosh: Sure, removed! Although I haven't tested it yet to see if it builds and works properly on aarch64

rs.ruffle.Ruffle.yaml
kjarosh: Oh, whoops!

rs.ruffle.Ruffle.yaml
kjarosh: We are in the process of releasing Ruffle 1.0.0, currently we have only nightly releases, but we do not plan to update flatpak daily.

When 1.0.0 gets released (and it should be soon), I will switch this to the stable version tag and remove `patches/release.patch`, as we will probably be updating our metainfo with releases at that point.

Is it a blocker or is it acceptable to merge and switch to stable later?

bbhtt:
bot, build rs.ruffle.Ruffle

rs.ruffle.Ruffle.yaml
bbhtt: I think as long as this doesn't get updated daily, it is probably fine. You should not enable external-data-checker (x-checker) too as the PRs will pile up due to high release frequency. It runs hourly, daily.

You can update it weekly or semi-weekly etc.

kjarosh: Alright! I'll set up the checker when we have stable releases (we will probably be releasing stable ~monthly).

kjarosh:
Tested the build with:
* https://www.transformice.com/TransformiceChargeur.swf
* http://deadmaze.com/alpha/chargeur-deadmeat.swf
* a SWF from https://www.newgrounds.com/portal/view/178172

I noticed that CPU was used for rendering graphics so I added `--device=dri` to fix that.

kjarosh:
bot, build rs.ruffle.Ruffle

kjarosh:
The newest build seems to work fine! Is there anything else I should do before merge?

hadess:
Maintainer of the [Flash Player package](https://flathub.org/apps/com.adobe.Flash-Player-Projector) here.

I tried to play Samorost 2:
- opening the file chooser seems to hang the application, which means that I get pops up about the application hanging while trying to choose a file
- loading neighbouring files doesn't work, and doesn't throw any errors in the UI, just on the command-line. So while waiting for https://github.com/flatpak/xdg-desktop-portal/issues/463 to be fixed, might be best to require more filesystem permissions, or have a dialogue about the error which the user can act upon:
```
2024-08-04T15:54:38.380091Z ERROR ruffle_core::loader: Error during movie loading of "file:///run/user/1000/doc/de0436cc/01intro/intro.swf": FetchError("No such file or directory (os error 2)")
```

hadess:
You might also want to add a target to regenerate the cargo file (and document how it was generated at the same time).

kjarosh:
> opening the file chooser seems to hang the application, which means that I get pops up about the application hanging while trying to choose a file

That's an upstream bug:
* https://github.com/ruffle-rs/ruffle/issues/11336

kjarosh:
> You might also want to add a target to regenerate the cargo file (and document how it was generated at the same time).

Added a target, to regenerate cargo sources: `make regenerate-sources RUFFLE_COMMIT=<commit>`.

kjarosh:
> loading neighbouring files doesn't work, and doesn't throw any errors in the UI, just on the command-line

That's a very good remark. We generally do not show errors to the user (only panics) and every game that does not work will require collecting logs to see what's wrong, even outside of flatpak.

We would have to have full filesystem access, without that multi-file movies will not work locally (they will work if loaded from a URL or bundled into a single file though).

I asked others from Ruffle about that, would be also great to know Flathub's preference — having full disk access vs missing a functionality?

hadess:
I went with full host access, but read-only, for the Flash Player, and it seems to be acceptable for now:
https://github.com/flathub/com.adobe.Flash-Player-Projector/blob/master/com.adobe.Flash-Player-Projector.json#L15

kjarosh:
Is there a specific reason why it's `host` and not `home`? I think usually users have their SWFs in their home directory

Doomsdayrs:
> Is there a specific reason why it's `host` and not `home`? I think usually users have their SWFs in their home directory

Home makes more sense for Ruffle. Given that the Adobe flash is legacy software from an era of no sandboxing.

If a user has swfs on some separate partition for whatever reason, they also know how to handle file permissions.

kjarosh:
Unfortunately just adding `--filesystem=home:ro` does not work, as file chooser still translates paths (e.g. to `/run/user/1000/doc/1c898eec/file.swf`) and other files are looked up with paths relative to the original. I can play those movies when I know their absolute paths, but I cannot use the file chooser to select them.

Doomsdayrs:
> Unfortunately just adding `--filesystem=home:ro` does not work, as file chooser still translates paths (e.g. to `/run/user/1000/doc/1c898eec/file.swf`) and other files are looked up with paths relative to the original. I can play those movies when I know their absolute paths, but I cannot use the file chooser to select them.

I don't recall this being how it happens, will test on my system soon.

kjarosh:
I tested it on two different systems and it works the same way; giving filesystem permissions does not help.

I think that problem will have to be addressed upstream along with other issues related to prompts, but that requires major changes.

We discussed this on the Ruffle Discord server and came into the conclusion that it's not a blocker: the vast majority of games people will use are single-file SWFs (or URLs), and everything works properly for them.

kjarosh:
bot, build rs.ruffle.Ruffle

Doomsdayrs:
Just to nitpick, is it just me or is the UI elements slightly blurry? (is this found upstream as well?)

bbhtt:
Is there any games to test this "loading neighbouring files does not work" issue?

hadess:
> Are there any games to test this "loading neighbouring files does not work" issue?

I use Amarost 2 from the first Humble Bundle, but I don't know if the version that's for sale still uses Flash.

kjarosh:
> Just to nitpick, is it just me or is the UI elements slightly blurry? (is this found upstream as well?)

Wdym? It looks fine for me, even on hidpi screens.

> Are there any games to test this "loading neighbouring files does not work" issue?

You can use this: [juicy-breakout.zip](https://github.com/user-attachments/files/16508071/juicy-breakout.zip)

I'll be working on adding an option to "open a directory" upstream, which will give Ruffle access to the whole directory with all assets. However, as I said before, we do not see this issue as a blocker for now.

bbhtt:
I think you should add `home:ro` or `xdg-download:ro` regardless, because typing in the full path seems to work. Without it opening these games don't work.

bbhtt:
`:ro` makes the path from portal show up, which seems to be a separate issue, only read-write ie. `home` makes the "real" path work.

bbhtt:
This issue sounds related https://github.com/flatpak/flatpak/issues/5413

kjarosh:
Okay, I think I figured this out. I was testing it with `--filesystem=home:ro` in finish-args, but when the file picker popped up I had "Open files read-only" unchecked. That meant the portal had to provide the file in `/run/flatpak/doc` in order for it to be writable.

When I either use `--filesystem=home` or check "Open files read-only" it works.

The problem is that the checkbox is disabled by default and it's unintuitive to check it in order for multi-file movies to work (I fell for it, so the average user will for sure).

I'll add `--filesystem=home` in that case, because I don't see any easy way to pick files with read-only pre-selected for us. That should fix the problem of multi-file movies for now.

kjarosh:
bot, build rs.ruffle.Ruffle

kjarosh:
Seems to be working fine now

kjarosh:
Please also add @sombraguerrero as an additional maintainer after merge if possible, thanks!

And thank you for your reviews and time!

flathub.json
lvyitian: I tried building it on Aarch64 Linux and it succeeded

PR #5474, opened 2024-08-03
Add io.github.bkbilly.lnxlink

bkbilly:

It collects data from the system and sends them to an MQTT broker and has support for Home Assistant that it automatically creates sensors.

- [X] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about this submission. **Link:** https://github.com/bkbilly/lnxlink


flatpak-modules-opencv.yaml
hfiguiere, line 1: this is not autogenerated it should be inlined in the main manifest instead.

flatpak-modules-system.json
hfiguiere, line 1: this is not auto generated it should be in the main manifest.

io.github.bkbilly.lnxlink.metainfo.xml
hfiguiere, line 1: this file should be managed / submitted upstream.

io.github.bkbilly.lnxlink.metainfo.xml
```diff
@@ -0,0 +1,50 @@
...
+    <release version="2024.8.0" date="2024-08-05">
+      <url type="details">https://github.com/bkbilly/lnxlink/releases</url>
+      <description>
+        <p>First Flatpak Release</p>
+          <ul>
+            <li>Easy installation</li>
+          </ul>
+      </description>
+    </release>
+  </releases>
```
hfiguiere, line 48: missing screenshots.

io.github.bkbilly.lnxlink.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - flatpak-lnxlink.yaml
+  - flatpak-modules-system.json
+  - name: metadata
+    buildsystem: simple
+    build-commands:
+      - install -Dm644 -t ${FLATPAK_DEST}/share/metainfo ${FLATPAK_ID}.metainfo.xml
+    sources:
+      - type: file
+        path: io.github.bkbilly.lnxlink.metainfo.xml
+finish-args:
```
hfiguiere, line 20: finish args should be above `module`

io.github.bkbilly.lnxlink.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - --share=network
+  # Access to audio
+  - --socket=pulseaudio
+  # Access to Bluetooth
+  - --allow=bluetooth
+  # Needs to save files locally
+  - --filesystem=xdg-documents
+  # Access to filesystem
+  - --filesystem=/tmp/
+  - --filesystem=/var/run/reboot-required
```
hfiguiere, line 45: why do you need this?

io.github.bkbilly.lnxlink.yaml
```diff
@@ -0,0 +1,50 @@
...
+  # host-os
+  - --filesystem=host-os
```
hfiguiere, line 47: why do you need this?

lnxlink.png
hfiguiere, line 1: screenshot should be upstream or on the app website if there is one.

io.github.bkbilly.lnxlink.yaml
```diff
@@ -0,0 +1,50 @@
...
+  # Access to camera, gamepad
+  - --device=all
+  # - --device=input
```
hfiguiere, line 50:
```suggestion
```

io.github.bkbilly.lnxlink.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - type: file
+        path: io.github.bkbilly.lnxlink.metainfo.xml
+finish-args:
+  # Wayland access
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  # DBus
+  # - --socket=system-bus
+  # - --socket=session-bus
```
hfiguiere, line 27:
```suggestion
```

hfiguiere (changes requested):
I don't see an icon nor a desktop file installed.

io.github.bkbilly.lnxlink.yaml
bkbilly: The /tmp path is required for the notification image: https://github.com/bkbilly/lnxlink/blob/master/lnxlink/modules/notify.py

The reboot-required is for checking if the system needs to be rebooted: https://github.com/bkbilly/lnxlink/blob/master/lnxlink/modules/required_restart.py

io.github.bkbilly.lnxlink.metainfo.xml
bkbilly: Could you explain what you mean with submitting it upstream?

lnxlink.png
bkbilly: I've added a screenshot

io.github.bkbilly.lnxlink.yaml
bkbilly: It seems that it works without it, so I removed it.

io.github.bkbilly.lnxlink.yaml
bkbilly: Both of them are needed for my application to work properly.

io.github.bkbilly.lnxlink.yaml
bkbilly: Fixed it.

io.github.bkbilly.lnxlink.metainfo.xml
bkbilly: I've added a screenshot.

flatpak-modules-system.json
bkbilly: Removed the files and added to the main yaml file.

flatpak-modules-opencv.yaml
bkbilly: Removed the files and added to the main yaml file.

bkbilly:
This is not a desktop application, it runs from the terminal, so I don't think that a desktop file is required.

hfiguiere:
> This is not a desktop application, it runs from the terminal, so I don't think that a desktop file is required.

but then you open sockets for wayland and x11 fallback.

hfiguiere:
also, just setting the expectation. cli aren't a good use case for flatpak.

bkbilly:
The sockets are required for using the xdotool to move the mouse and for others that I am using.
I find the flatpak very useful for my application because it will make it much easier for users to install it without needing to manually install python and all it's dependencies.

flatpak-lnxlink.yaml
```diff
@@ -0,0 +1,55 @@
...
+    sha256: 82fee1fc78add43492d3a1898bfa6d8a904cc97d8427f683ed8e798d07761aa0
+  - type: file
+    url: https://files.pythonhosted.org/packages/35/cb/6d564f8a3f25d9516298dce151670d01e43a4b3b769c1c15f40453179cd5/inotify-0.2.10.tar.gz
+    sha256: 974a623a338482b62e16d4eb705fb863ed33ec178680fc3e96ccdf0df6c02a07
+  - type: file
+    url: https://files.pythonhosted.org/packages/11/f1/0f1436d1bb7fae54f557a9965ab2b61163d57853c6f4e2a97236973acd28/jc-1.25.3-py3-none-any.whl
+    sha256: ea17a8578497f2da92f73924d9d403f4563ba59422fbceff7bb4a16cdf84a54f
+  - type: git
+    url: https://git@github.com/bkbilly/lnxlink.git
+    commit: 8dbb784
```
bbhtt, line 31: If this is the app it should be in the main manifest

io.github.bkbilly.lnxlink.metainfo.xml
bbhtt: https://github.com/bkbilly/lnxlink

io.github.bkbilly.lnxlink.metainfo.xml
```diff
@@ -0,0 +1,56 @@
...
+      <li>No sudo required: No need to be root user to install and use, unless used on server setup.</li>
+      <li>Easily expanded: Any new module is automatically imported and custom modules can be added.</li>
+    </ul>
+  </description>
+
+  <screenshots>
+    <screenshot type="default">
+      <image>https://private-user-images.githubusercontent.com/518494/351081444-1b7f3fc2-4387-4cd1-8fcf-25d77137c3fe.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjI3MTgxMjksIm5iZiI6MTcyMjcxNzgyOSwicGF0aCI6Ii81MTg0OTQvMzUxMDgxNDQ0LTFiN2YzZmMyLTQzODctNGNkMS04ZmNmLTI1ZDc3MTM3YzNmZS5wbmc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwODAzJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDgwM1QyMDQzNDlaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1hNzhhN2MyNThlODNlYzRhY2QyNDRiNDdjNDIwZTRiODRmNmIzZjlhZTJjZjMyZGM4NzUzN2M3YmU1M2VkYWU5JlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.nZWVOTZOZJPJZSwXIBkd43iT8wwFZqqnidlk0eW9Iuk</image>
```
bbhtt, line 30: This link seems dead?

io.github.bkbilly.lnxlink.yaml
bbhtt: Don't hardcode `tmp` use the tempfile library https://docs.python.org/3/library/tempfile.html

Then set TMPDIR in the finish-args.

io.github.bkbilly.lnxlink.yaml
```diff
@@ -0,0 +1,165 @@
...
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  # For DBus
+  - --system-talk-name=org.freedesktop.NetworkManager
+  - --system-talk-name=org.freedesktop.login1
+  - --system-talk-name=org.freedesktop.UPower
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.gnome.Mutter.IdleMonitor
+  - --talk-name=org.mpris.MediaPlayer2.*
```
bbhtt, line 17: Why is this needed? Is this a MPRIS client?

io.github.bkbilly.lnxlink.yaml
```diff
@@ -0,0 +1,165 @@
...
+command: lnxlink
+finish-args:
+  # Wayland access
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  # For DBus
+  - --system-talk-name=org.freedesktop.NetworkManager
+  - --system-talk-name=org.freedesktop.login1
+  - --system-talk-name=org.freedesktop.UPower
```
bbhtt, line 14: Why are these needed?

lnxlink.png
bbhtt, line 1: Screenshot should be upstream

io.github.bkbilly.lnxlink.yaml
bbhtt: Don't hardcode `tmp` use the tempfile library https://docs.python.org/3/library/tempfile.html

Then set TMPDIR in the finish-args.

io.github.bkbilly.lnxlink.yaml
bkbilly: This is used to get information from the current playing media players: https://github.com/bkbilly/lnxlink/blob/master/lnxlink/modules/media.py

io.github.bkbilly.lnxlink.yaml
bkbilly: The Network Manager is to get information about the WiFi: https://github.com/bkbilly/lnxlink/blob/master/lnxlink/modules/wifi.py
The login1 is used to shutdown/restart the computer: https://github.com/bkbilly/lnxlink/blob/master/lnxlink/modules/shutdown.py
The UPower is used for getting the battery information of connected devices: https://github.com/bkbilly/lnxlink/blob/master/lnxlink/modules/battery.py

lnxlink.png
bkbilly: How should I do this? I don't understand what you mean with upstream...

io.github.bkbilly.lnxlink.metainfo.xml
bkbilly: Fixed it.

flatpak-lnxlink.yaml
bkbilly: Added it to the main configuration

lnxlink.png
bkbilly: I've uploaded an image preview and an icon. Could you check if I did this correctly?

io.github.bkbilly.lnxlink.metainfo.xml
bkbilly: I've uploaded the icon

bkbilly: I think I've already uploaded it. It's the last line that looks like this:
```yaml
  - name: metadata
    buildsystem: simple
    build-commands:
      - install -Dm644 io.github.bkbilly.lnxlink.metainfo.xml -t ${FLATPAK_DEST}/share/metainfo
      - install -Dm644 io.github.bkbilly.lnxlink.logo.png -t $FLATPAK_DEST/share/icons/hicolor/scalable/apps
    sources:
      - type: file
        path: io.github.bkbilly.lnxlink.metainfo.xml
      - type: file
        path: io.github.bkbilly.lnxlink.logo.png
```

io.github.bkbilly.lnxlink.metainfo.xml
```diff
@@ -0,0 +1,56 @@
...
+      <li>Sensors: Automatically discover sensors that monitor and control the system.</li>
+      <li>Home Assistant: Uses MQTT Autodiscovery to create entities and shows if update is required.</li>
+      <li>No sudo required: No need to be root user to install and use, unless used on server setup.</li>
+      <li>Easily expanded: Any new module is automatically imported and custom modules can be added.</li>
+    </ul>
+  </description>
+
+  <screenshots>
+    <screenshot type="default">
+      <image>https://raw.githubusercontent.com/bkbilly/lnxlink/master/images/preview.png</image>
```
bbhtt, line 30: screenshot should be from a tag or a commit not `master`

io.github.bkbilly.lnxlink.metainfo.xml
bbhtt: upstream means your github repository, the icon and screenshot should be there too.

io.github.bkbilly.lnxlink.yaml
bbhtt: don't hardcode /tmp like that. Use the tempfile library, then set TMPDIR in the manifest.

lnxlink.png
bbhtt: Upstream means your GitHub repository not here

io.github.bkbilly.lnxlink.metainfo.xml
bkbilly: I've added the latest tag.

lnxlink.png
bkbilly: I removed the screenshot from here.

io.github.bkbilly.lnxlink.metainfo.xml
bkbilly: I've uploaded the screenshot and logo to my repository: https://github.com/bkbilly/lnxlink/tree/master/images

io.github.bkbilly.lnxlink.metainfo.xml
```diff
@@ -0,0 +1,56 @@
...
+      <li>Home Assistant: Uses MQTT Autodiscovery to create entities and shows if update is required.</li>
+      <li>No sudo required: No need to be root user to install and use, unless used on server setup.</li>
+      <li>Easily expanded: Any new module is automatically imported and custom modules can be added.</li>
+    </ul>
+  </description>
+
+  <screenshots>
+    <screenshot type="default">
+      <image>https://raw.githubusercontent.com/bkbilly/lnxlink/d6e42e9/lnxlink/logo.png</image>
+      <caption>A caption</caption>
```
hfiguiere, line 31: this is not a screenshot. This is just the product icon.

bkbilly: Oh, I got it confused. I fixed it.

io.github.bkbilly.lnxlink.yaml
bkbilly: How should I set the TMPDIR?

bbhtt: `--env=TMPDIR=value` if you want persistent use `/var/tmp` for non-persistent `${XDG_RUNTIME_DIR}/app/${FLATPAK_ID}`

bkbilly: I still can't understand how to use it.
Under the finish-args I've removed the `filesystem=/tmp/` and added the TMPDIR environment but it doesn't have access to the `/tmp` filesystem:
```yaml
- --env=TMPDIR=/tmp/
```

I've also tried adding the variable you suggested, but this time it doesn't even build:
```yaml
- --filesystem=${XDG_RUNTIME_DIR}/app/${FLATPAK_ID}
```

I am using the tempfile library which defaults to the `/tmp/` folder, not the `/var/tmp/`.

bbhtt: You don't need `/tmp`, the directory used in TMPDIR will be your new temp folder.

https://docs.python.org/3/library/tempfile.html

> The default directory is chosen from a platform-dependent list, but the user of the application can control the directory location by setting the TMPDIR, TEMP or TMP environment variables.****

bkbilly: My application creates temporary images that they need to be deleted by the OS when it restarts, but as long as the OS is running they need to exist because they are used by the system notification. 
The user has the option to open it at any time, as long as the notification is active via action buttons. This is why I set the tempfile with these options:
```python
tempfile.NamedTemporaryFile(prefix="lnxlink_icon_", delete=False, delete_on_close=False)
```

It seems to me that setting the TMPDIR won't work for the requirements of my application.

bbhtt: Then what's the issue with setting `--env=TMPDIR=${XDG_RUNTIME_DIR}/app/${FLATPAK_ID}` ?

bbhtt: Also why does notification images need to be temporary files?

bbhtt: Yea, right missed that.

bkbilly: The notification needs temporary files for the images and sounds. These are downloaded to a temporary file and they are called by the notification system or can be shown as full screen when the user clicks on a provided button. 
The user can click on that button at any point that the OS is turned on, so they need to exist and be deleted once it restarts because otherwise it will have a lot of unwanted files.
This image could be used for any application that the user wants by creating automations on Home Assistant. For example it could be used to be sent as an email attachment or anything else.

bbhtt: I'm guessing notifications here mean desktop notifications? So these icons need to be accessible outside the sandbox for the notification daemon on host to show the icon?

bbhtt:
Can you move metainfo and icon upstream?

io.github.bkbilly.lnxlink.yaml
bkbilly: Yeah, that's how the icons are used.

bkbilly:
I've added the metainfo to upstream, the icon already exists.

io.github.bkbilly.lnxlink.yaml
bkbilly: I tried that, but unfortunately it doesn't work because when the application starts, the directory `/tmp/lnxlink` doesn't exist so I create it.
```python
if not os.path.exists("/tmp/lnxlink"):
    os.makedirs("/tmp/lnxlink")
```
Once it is created with flatpak, the system can't see the newly created directory.
In order to work the way you suggest, the directory `/tmp/lnxlink` has to exist before running the application.

bbhtt: The `:create` prefix exists for that. `--filesystem=/tmp/lnxlink:create` will create the directory if it doesn't exist first.

bkbilly: This works perfectly! Thanks for your suggestion!

bbhtt:
bot, build io.github.bkbilly.lnxlink

io.github.bkbilly.lnxlink.metainfo.xml
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="console-application">
+  <content_rating type="oars-1.0" />
```
bbhtt, line 3:
```suggestion
  <content_rating type="oars-1.1" />
```

io.github.bkbilly.lnxlink.yaml
```diff
@@ -0,0 +1,221 @@
...
+    - cp -r ./build/*  /app/
+    sources:
+    - type: archive
+      url: https://github.com/jordansissel/xdotool/releases/download/v3.20211022.1/xdotool-3.20211022.1.tar.gz
+      sha256: 96f0facfde6d78eacad35b91b0f46fecd0b35e474c03e00e30da3fdd345f9ada
+  - name: metadata
+    buildsystem: simple
+    build-commands:
+      - install -Dm644 io.github.bkbilly.lnxlink.metainfo.xml -t ${FLATPAK_DEST}/share/metainfo
+      - install -Dm644 io.github.bkbilly.lnxlink.png -t $FLATPAK_DEST/share/icons/hicolor/scalable/apps
```
bbhtt, line 216:
```suggestion
      - install -Dm644 io.github.bkbilly.lnxlink.png -t $FLATPAK_DEST/share/icons/hicolor/512x512/apps
```

Why is png going to scalable folder?

io.github.bkbilly.lnxlink.metainfo.xml
```diff
@@ -0,0 +1,56 @@
...
+
+  <provides>
+    <binary>lnxlink</binary>
+  </provides>
+
+  <categories>
+    <category>System</category>
+    <category>Monitor</category>
+  </categories>
+
```
bbhtt, line 43:
```suggestion
  </categories>
  <icon type="stock">io.github.bkbilly.lnxlink</icon>

```

io.github.bkbilly.lnxlink.metainfo.xml
bbhtt: > I've uploaded the screenshot and logo to my repository: https://github.com/bkbilly/lnxlink/tree/master/images

Remove them from here after changes and use the ones from your repo.

Change it to:

```
type: file
url: https://raw.githubusercontent.com/bkbilly/lnxlink/d6e42e9b007798d18f22b5f0b6f1b613fd7373f4/images/logo.png
sha256: ...
```

bbhtt:
Can you show a screen record of how this works? I tried to launch it but it is asking for some configs

bkbilly:
Here is a preview of running the application for the 1st time:
https://github.com/user-attachments/assets/43f7e835-7a46-4358-9053-7dc732c31cd6

bbhtt:
bot, build io.github.bkbilly.lnxlink

PR #5497, opened 2024-08-11
Add net.launchpad.qpdfview

hg:

qpdfview is an excellent tabbed document reader that supports PDF, DjVu, and PostScript formats and is the closest analog to SumatraPDF that I could find. I've used it for years.

---

- [x] ~~I am an author/developer/upstream contributor of the project.~~ If not, I contacted upstream developers about this submission.

I wrote the manifest a little over a year ago, and [asked the upstream][q] to consider submitting it themselves, but have received no feedback.

[q]: https://answers.launchpad.net/qpdfview/+question/707222

---

## filesystem access

Because of `filesystem=host:ro`, users will not be able to save copies of documents from within the application, and this functionality is disabled at build time. Considering how exploitable PDF readers are, and how niche this feature is, I think we should rely on sandboxing Flatpak provides us to protect the majority of users from e.g. cryptominers.

Another option is to *not* disable the menu entries, and somehow let users who need this feature know that they should override the default permission.

What do you think?

## metainfo

I understand that metainfo changes should be upstreamed, but this project runs at slow pace, with releases being cut every 2-4 years. If we decide to block on this, the submission will be delayed by at least another year.

I will try to upstream these changes if/when this is merged, and (if successful) will remove the patch on next release.

Original screenshots have very low resolution, so I made my own.


bbhtt:
Pretty sure `net.launchpad` ID means it would be unverifiable. Does launchpad allow project's to have static sites on a subdomain? What does the app internally use as ID?

metainfo.patch
```diff
@@ -0,0 +1,50 @@
+diff --git a/miscellaneous/qpdfview.appdata.xml b/miscellaneous/qpdfview.appdata.xml
+index bdcc460..6ddce9a 100644
+--- a/miscellaneous/qpdfview.appdata.xml
++++ b/miscellaneous/qpdfview.appdata.xml
+@@ -32,24 +32,39 @@
+     <launchable type="desktop-id">qpdfview.desktop</launchable>
+     <screenshots>
+       <screenshot type="default">
+-        <image>https://screenshots.debian.net/shrine/screenshot/14328/simage/small-587677349ebef2f9ab1b56d7cbc7888e.png</image>
++        <image>https://i.imgur.com/VzApQWY.png</image>
```
bbhtt, line 10: I'd just commit them to the PR here under `screenshots/`.

net.launchpad.qpdfview.yml
```diff
@@ -0,0 +1,156 @@
...
+rename-appdata-file: qpdfview.appdata.xml
+
+finish-args:
+  - --filesystem=host:ro # open file dialog; drag & drop
+  - --device=dri # hardware acceleration
+  - --share=ipc # x11
+  - --socket=wayland # gui
+  - --socket=fallback-x11 # gui
+  - --socket=cups # printing
+  - --own-name=local.qpdfview # unique instance
```
bbhtt, line 19: Does the app really use `.local` ?

net.launchpad.qpdfview.yml
```diff
@@ -0,0 +1,156 @@
...
+
+cleanup:
+  - /include
+  - /lib/*.a
+  - /lib/*.la
```
bbhtt, line 24:
```suggestion
  - "*.a"
  - "*.la"
```

net.launchpad.qpdfview.yml
```diff
@@ -0,0 +1,156 @@
...
+          project-id: 1724
+          stable-only: true
+          url-template: http://libspectre.freedesktop.org/releases/libspectre-$version.tar.gz
+
+  - name: qpdfview
+    buildsystem: simple
+    build-commands:
+      # Saving copies will not be available because of `filesystem=host:ro`
+      - sed -i 's|return m_document->canSave();|return false;|' sources/documentview.cpp
+      - sed -i 's|/usr|/app|' qpdfview.pri
```
bbhtt, line 136: Can you use a patch for this?

bbhtt:
bot, build net.launchpad.qpdfview

net.launchpad.qpdfview.yml
```diff
@@ -0,0 +1,156 @@
+app-id: net.launchpad.qpdfview
+default-branch: stable
```
bbhtt, line 2:
```suggestion
```

net.launchpad.qpdfview.yml
hfiguiere: this is a bug if that's the case.

net.launchpad.qpdfview.yml
```diff
@@ -0,0 +1,156 @@
...
+          url-template: http://libspectre.freedesktop.org/releases/libspectre-$version.tar.gz
+
+  - name: qpdfview
+    buildsystem: simple
+    build-commands:
+      # Saving copies will not be available because of `filesystem=host:ro`
+      - sed -i 's|return m_document->canSave();|return false;|' sources/documentview.cpp
+      - sed -i 's|/usr|/app|' qpdfview.pri
+      - lrelease qpdfview.pro
+      - qmake qpdfview.pro
```
hfiguiere, line 138: should this be `buildsystem: qmake` ?

metainfo.patch
bbhtt: Use these links for now https://raw.githubusercontent.com/flathub/flathub/8f5611acd0a2a6cc4a58c130ae8d952f5af9b226/screenshots/1-main.png

You can update it once the submission is accepted into a new repository.

net.launchpad.qpdfview.yml
bbhtt: This should be fixed upstream. You should probably open an issue

bbhtt:
bot, build net.launchpad.qpdfview

bbhtt:
I'm a bit concerned about the ID, as the developers of this software don't own, `launchpad.net`, which is under the control of Canonical. If upstream is active, the issue about correct dbus names should be raised.

The ID can follow the proper dbus name if that is fixed.

PR #5501, opened 2024-08-13
Add io.github.limo_app.limo

limo-app:

- [ X] Please describe the application briefly.
Limo is a mod manager, a type of application used to manage collections of files ("mods") meant to overwrite files
used by other applications (often games).

Note on permissions:
talk-name=org.freedesktop.Flatpak is required for the following reasons:
 - allow users to add launch commands for custom modding tools which are to be run on the host. 
 - allow users to run a modded application through Limo
 - the functionality includes archive extraction. Some RAR5 archives can only reliably be extracted using non open source
  tools like unrar. For those cases a user can optionally set an unrar executable to be used.

A pull request for the above exception has been created

Note on the linter:
According to the the [documentation](https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/#description),
ul and p tags are allowed in the description. Since the linter does not allow them, i don't know which standard to follow.

- [X] I am an author of the project. [Link](https://github.com/limo-app/limo)


io.github.limo_app.limo.json
```diff
@@ -0,0 +1,351 @@
...
+          "tag": "v1.3.1"
+        }
+      ],
+      "cleanup":
+      [
+        "/lib/libz.so*"
+      ]
+    },
+    {
+      "name": "curl",
```
hfiguiere, line 134: libcurl is in the runtime

io.github.limo_app.limo.json
```diff
@@ -0,0 +1,351 @@
...
+      [
+        {
+          "type": "git",
+          "url": "https://github.com/zeux/pugixml.git",
+          "tag": "v1.14"
+        }
+      ]
+    },
+    {
+      "name": "zlib",
```
hfiguiere, line 112: zlib is in the runtime

io.github.limo_app.limo.json
```diff
@@ -0,0 +1,351 @@
...
+      [
+        {
+          "type": "archive",
+          "url": "https://github.com/boostorg/boost/releases/download/boost-1.85.0/boost-1.85.0-cmake.tar.xz",
+          "sha256": "0a9cc56ceae46986f5f4d43fe0311d90cf6d2fa9028258a95cab49ffdacf92ad"
+        }
+      ]
+    },
+    {
+      "name": "icu",
```
hfiguiere, line 207: icu is in the runtime.

io.github.limo_app.limo.json
```diff
@@ -0,0 +1,351 @@
...
+      "buildsystem": "simple",
+      "build-commands":
+      [
+        "cd source && ./configure --prefix /app/local --enable-static=yes --enable-shared=no",
```
hfiguiere, line 211: this should be `buildsystem: autotools` anyway. You are probably looking at `subdir: source`. But see above: it's in the runtime.

io.github.limo_app.limo.json
```diff
@@ -0,0 +1,351 @@
...
+    },
+    {
+      "name": "libloot",
+      "buildsystem": "cmake-ninja",
+      "config-opts":
+      [
+        "-DCMAKE_BUILD_TYPE=Release",
+        "-DLIBLOOT_BUILD_TESTS=OFF",
+        "-DLIBLOOT_INSTALL_DOCS=OFF",
+        "-DESPLUGIN_URL=/run/build/libloot/6.1.0.tar.gz",
```
hfiguiere, line 295: use `$FLATPAK_BUILDER_BUILDDIR` here and below... instead for `/run/build/...`

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64"]
```
hfiguiere, line 2: why so?

limo-app: I do not own an ARM device for testing. Other than that, i see no reason to restrict building to x86

limo-app:
I've included and tested the requested changes.
Question on architecture support: Should i remove the x86 restriction even though i currently cannot test on other architectures?

hfiguiere:
if it's not build how do you think it will get tested?

limo-app:
I was assuming that is to be done by building locally before enabling it on Flathub.
Will remove the restriction now

io.github.limo_app.limo.json
```diff
@@ -0,0 +1,293 @@
...
+    "--filesystem=/media",
+    "--filesystem=/run/media",
+    "--talk-name=org.freedesktop.Flatpak",
+    "--system-talk-name=org.freedesktop.UDisks2"
+  ],
+  "cleanup":
+  [
+    "/lib/debug/bin",
+    "/lib/debug/lib",
+    "/lib/debug/source",
```
bbhtt, line 40:
```suggestion
```

io.github.limo_app.limo.json
```diff
@@ -0,0 +1,293 @@
...
+    "/bin/bsd*",
+    "/bin/cbindgen",
+    "/bin/curl*",
+    "/bin/mk-ca-bundle.pl",
+    "/share/doc",
+    "/share/man",
+    "/share/pkgconfig",
+    "/include",
+    "/lib/*.a",
```
bbhtt, line 49:
```suggestion
    "*.a",
```

io.github.limo_app.limo.json
```diff
@@ -0,0 +1,293 @@
...
+    "--system-talk-name=org.freedesktop.UDisks2"
+  ],
+  "cleanup":
+  [
+    "/lib/debug/bin",
+    "/lib/debug/lib",
+    "/lib/debug/source",
+    "/bin/bsd*",
+    "/bin/cbindgen",
+    "/bin/curl*",
```
bbhtt, line 43:
```suggestion
```

This won't be necessary anymore ?

bbhtt:
> Note on the linter:
According to the the [documentation](https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/#description),

They do work, your issue is using a `p` tag inside another. Also can you properly indent these?

https://github.com/limo-app/limo/blob/fe8c20dea1a69687e795a9805614990c9bef7522/flatpak/io.github.limo_app.limo.metainfo.xml#L22C6-L22C23

```xml
  <description>
    <p>
      Limo is a general-purpose mod manager for games and other applications.
      Built-in support for the Load Order Optimization Tool (LOOT) allows modding various Bethesda titles such as Skyrim or Fallout 4.
    </p>
    <p>Features:</p>
      <ul>
        <li>Multiple target directories per application</li>
        <li>Automatic adaptation of mod file names to prevent issues with case mismatches</li>
        <li>Auto-Tagging system for filtering</li>
        <li>Sort load order according to conflicts</li>
        <li>Import installed games from Steam</li>
        <li>Simple backup system</li>
        <li>LOOT integration</li>
        <li>NexusMods API support</li>
    </ul>
    <p>For a guide on how to set up Limo, refer to the GitHub wiki page.</p>
  </description>
```

io.github.limo_app.limo.json
```diff
@@ -0,0 +1,293 @@
...
+    "--share=ipc",
+    "--device=dri",
+    "--share=network",
+    "--socket=wayland",
+    "--socket=fallback-x11",
+    "--filesystem=home",
+    "--filesystem=/mnt",
+    "--filesystem=/media",
+    "--filesystem=/run/media",
+    "--talk-name=org.freedesktop.Flatpak",
```
bbhtt, line 33: > allow users to add launch commands for custom modding tools which are to be run on the host.

What kind of commands are these usually? Do you have examples?

`unrar` can be included in the Flatpak package

io.github.limo_app.limo.json
```diff
@@ -0,0 +1,293 @@
...
+      [
+        {
+          "type": "archive",
+          "url": "https://github.com/boostorg/boost/releases/download/boost-1.85.0/boost-1.85.0-cmake.tar.xz",
+          "sha256": "0a9cc56ceae46986f5f4d43fe0311d90cf6d2fa9028258a95cab49ffdacf92ad"
+        }
+      ]
+    },
+    {
+      "name": "cbindgen",
```
bbhtt, line 167: Is this a build dependency only? You are cleaning up `cbindgen` which is the only thing this builds.

If so, then add `cleanup: ["*"]` here

limo-app: Yes, cbindgen is only used to build libloot

io.github.limo_app.limo.json
limo-app: These are usually commands that run some external tool meant to edit an applications files in a way that cannot be done by simply replacing files. [Here](https://steamcommunity.com/sharedfiles/filedetails/?id=2442018077) is an example guide where one such tool is added to a similar application.
Launching these external tools through this application is technically not required for its function but does provide a convenient way for users to collect them all in one place.

limo-app:
I have replaced calls to the unrar binary with a shared lib of unrar, so 1 of the 3 reasons for needing flatpak-spawn --host is longer vailid

io.github.limo_app.limo.json
```diff
@@ -0,0 +1,345 @@
...
+        "dependencies/loot-condition-interpreter-sources.json"
+      ]
+    },
+    {
+      "name": "unrar",
+      "buildsystem": "simple",
+      "build-commands":
+      [
+        "make lib",
+        "make install-lib DESTDIR=/app",
```
bbhtt, line 310: This should be buildsystem: make, you can use `make-install-args` and `make-args` to pass arguments.

The two commands below should be in `post-install`.

limo-app: Sorry for all the extra work i'm causing but i can't seem to get make as a buildsystem to work:
Error: module unrar: Invalid buildsystem: "make"
I've moved the 2 commands to post-install.

bbhtt: Ah yea my bad. I was thinking about buildstream...

This should work:

```json
        {
            "name": "unrar",
            "no-autogen": true,
            "no-make-install": true,
            "make-args": [
                "lib"
            ],
            "post-install": [
                "install -D dll.hpp /app/include/unrar/dll.hpp",
                "install -D libunrar.a /app/lib/libunrar.a"
            ],
            "sources": [
                {
                    "type": "git",
                    "url": "https://github.com/aawc/unrar.git",
                    "tag": "v7.0.9"
                }
            ]
        }
```

limo-app: That works, thanks

bbhtt:
bot, build io.github.limo_app.limo

limo-app:
It looks like the build fails on ARM because a dependency of libloot is missing a dependency. As i have no experience with that architecture and no device to test on, i would appreciate any advice you can give me.

bbhtt:
Don't think it's an issue with dependencies, seems like the aarch64 target isn't defined:

```
error[E0463]: can't find crate for `core`
  |
  = note: the `x86_64-unknown-linux-gnu` target may not be installed
  = help: consider downloading the target with `rustup target add x86_64-unknown-linux-gnu`
```

it should not be `x86_64-unknown-linux-gnu` in aarch64 ci.

bbhtt:
I guess ask upstream https://github.com/Ortham/esplugin/issues

hfiguiere:
or more they get out of their way to undefine aarch64. By default, it just works™

limo-app:
The CMAKE script of libloot calls `cargo build` with `--target x86_64-unknown-linux-gnu` when building esplugin for Linux, which is most likely the issue here. Are Flathub builds for aarch64 native or cross complied? It seems like if they are native i just have to get rid of the `--target` option to fix my issue.

bbhtt:
Yes it is native.

limo-app:
I've added a patch for libloots CMAKE script that removes the `--target` option from cargo. This builds successfully on x86 and it should, hopefully, work on native ARM now.

bbhtt:
Feel free to start the build.

bot, build io.github.limo_app.limo

PR #5509, opened 2024-08-16
Add org.overte.Overte

JulianGro:

Overte is a social-VR platform. This Flatpak specifically includes the client called "interface".

- [x] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about this submission. **Link:** https://github.com/orgs/overte-org/people


org.overte.Overte.yml
```diff
@@ -0,0 +1,88 @@
...
+      - type: git
+        url: https://github.com/refi64/unappimage
+        commit: d7f86f2a0d7ec3a69211125207d5f127386b849a
+
+  - name: Overte
+    buildsystem: simple
+    cleanup:
+      - squashfs-root
+    sources:
+      - type: extra-data
```
hfiguiere, line 43: why is this extra-data?

org.overte.Overte.yml
```diff
@@ -0,0 +1,88 @@
...
+        x-checker-data:
+          type: html
+          url: https://overte.org/downloads.html
+          version-pattern: "The current release of Overte is version ([\\d\\.-]+)"
+          url-pattern: "(https://public.overte.org/build/overte/release/([\\d\\.-]+)/Overte-([\\d\\.-]+)-x86_64.AppImage)"
+
+      - type: extra-data
+        only-arches:
+          - aarch64
+        url: http://188.245.105.160/Overte-colorspace-x86_64.AppImage
```
hfiguiere, line 60: this says `x86_64`. also a host without fqdn and no https.... that's doesn't inspire confidence.

org.overte.Overte.yml
```diff
@@ -0,0 +1,88 @@
+app-id: org.overte.Overte
+runtime: org.kde.Platform
+runtime-version: '5.15-23.08'
+sdk: org.kde.Sdk
+# QtWebEngine is not part of the KDE runtime,
+# so we use this package, which is also maintained by KDE developers.
```
hfiguiere, line 6:
```suggestion
```

org.overte.Overte.yml
JulianGro: This is an AppImage being extracted. Is there a preferable method?
Maybe the extra-data is what calls the required apply_extra.sh script?

org.overte.Overte.yml
```diff
@@ -0,0 +1,88 @@
...
+      - make -C squashfs-tools install INSTALL_DIR=/app/bin
+    sources:
+      - type: git
+        url: https://github.com/refi64/unappimage
+        commit: d7f86f2a0d7ec3a69211125207d5f127386b849a
+
+  - name: Overte
+    buildsystem: simple
+    cleanup:
+      - squashfs-root
```
hfiguiere, line 41:
```suggestion
```

this does nothing.

org.overte.Overte.yml
JulianGro: There is an x86_64 version which is an official release. There is also an aarch64 version which I build the day before yesterday. It works but is intended to be a placeholder until the next release, which should be next month. I can upload this aarch64 release to public.overte.org in a couple of days, I just don't have access to it right now.

JulianGro: Oh. And the filename is wrong. Good point. That is hardcoded into the AppImage builder :sweat_smile:

org.overte.Overte.yml
hfiguiere: > This is an AppImage being extracted. 

this bear no relation to extra-data

org.overte.Overte.yml
JulianGro: Are comments like this problematic? They are intended for my colleagues. In this case, it is not obvious that the KDE Qt runtime doesn't ship QtWebEngine since that is part of Qt.

org.overte.Overte.yml
hfiguiere: why not just building from source?

org.overte.Overte.yml
hfiguiere: Comment bring no value.

org.overte.Overte.yml
```diff
@@ -0,0 +1,88 @@
...
+# QtWebEngine is not part of the KDE runtime,
+# so we use this package, which is also maintained by KDE developers.
+base: io.qt.qtwebengine.BaseApp
+base-version: '5.15-23.08'
+command: Overte
+separate-locales: false
+rename-icon: Overte
+finish-args:
+  - --share=network
+  - --share=ipc # For Discord integration
```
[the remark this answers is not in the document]
JulianGro: You mean this is required by X11 as well?
We have Discord integration, which also uses IPC.

org.overte.Overte.yml
JulianGro: Does this not delete the `squashfs-root` directory created by extracting the AppImage?

org.overte.Overte.yml
JulianGro: > why not just building from source?

Flathub doesn't allow an internet connection while building from source, which is required in this case.

JulianGro: I just checked and it indeed is the reason for the script being called, which is needed to extract the AppImage.

> When installing the Flatpak package containing an extra data source, Flatpak calls the apply_extra script( …)

https://docs.flatpak.org/en/latest/module-sources.html

Is there any better way to run such operations while building the Flatpak?

hfiguiere: or use a standard archive format for the build artifacts. nonethess making your build process being able to work offline would improve on everything.

`extra-data` still isn't suited. This is reserved for license issue (ie when stuff isn't redistributable).

hfiguiere: > I just checked and it indeed is the reason for the script being called, which is needed to extract the AppImage.

Maybe I was unclear. The use of extra-data is irrelevant to extracting appimage, which is probably one of the worst way to provide the binaries here....

You should NOT use extra-data unless for licensing reasons.

bbhtt: > I just checked and it indeed is the reason for the script being called, which is needed to extract the AppImage.

You missed the first line of extra-data which clearly states it is for avoiding license and redistribution clauses https://docs.flatpak.org/en/latest/module-sources.html#extra-data

> Extra data is often used for proprietary applications to avoid redistribution by downloading the binary package provided by the original distributors during installation by the user.

JulianGro: Switched away from extra-data.

org.overte.Overte.yml
JulianGro: The file was renamed and uploaded to public.overte.org

icons/org.overte.Overte
hfiguiere, line 1: There is no icon in the appimage?

org.overte.Overte.desktop
hfiguiere, line 1: there is no desktop file in the appimage?

org.overte.Overte.yml
hfiguiere: still applies.

org.overte.Overte.yml
```diff
@@ -0,0 +1,74 @@
...
+          - exec /app/bin/overte/interface/interface $@
+
+      - type: file
+        path: org.overte.Overte.appdata.xml
+
+      - type: file
+        path: org.overte.Overte.desktop
+
+      - type: file
+        path: icons/org.overte.Overte.png
```
hfiguiere, line 63: this file doesn't exist.

org.overte.Overte.yml
```diff
@@ -0,0 +1,74 @@
...
+        only-arches:
+          - aarch64
+        url: https://public.overte.org/experiment/overte/2024.07.1/Overte-2024.07.1-modified-colorspace-aarch64.AppImage
+        sha256: 391b45f36ac8d4441ca0b532abc48d5c1ce5c9ff067c6d1901338ba099640912
+        dest-filename: Overte.AppImage
+
+      - type: script
+        dest-filename: Overte
+        commands:
+          - exec /app/bin/overte/interface/interface $@
```
hfiguiere, line 54:
```suggestion
          - exec /app/bin/overte/interface/interface "$@"
```

org.overte.Overte.appdata.xml
```diff
@@ -0,0 +1,119 @@
...
+    <!-- 10 mbit/s downstream and 2 mbit/s upstream -->
+    <internet bandwidth_mbitps="10">always</internet>
+  </recommends>
+
+  <provides>
+    <binary>Overte</binary>
+  </provides>
+
+<!--
+  <releases>
```
hfiguiere, line 110: as previously said, `releases` is mandatory.

icons/org.overte.Overte
hfiguiere: also this is likely imporperly named as even github can't figure out it's an image.

JulianGro: There is. I just had trouble getting through validation without the one in here. That was before I switched away from data-extra though, so maybe I could get rid of this one now.

org.overte.Overte.desktop
JulianGro: There is; It has some syntax errors in it though. I would like to keep the desktop file in here, until it is fixed upstream.

hfiguiere: the is desktop-file-edit to fix desktop files or a patch.

icons/org.overte.Overte
hfiguiere: that's because extra-data.

org.overte.Overte.yml
JulianGro: I don't see how an extra helpful comment would be problematic.

icons/org.overte.Overte
JulianGro: Switched to icon provided by AppImage.

org.overte.Overte.desktop
hfiguiere: this is still relevant.

JulianGro: Switched to desktop file provided in the AppImage.

hfiguiere, line 1: you can remove this file.

org.overte.Overte.yml
```diff
@@ -0,0 +1,70 @@
...
+        commands:
+          - exec /app/bin/overte/interface/interface "$@"
+
+      - type: file
+        path: org.overte.Overte.appdata.xml
+
+    build-commands:
+      - chmod +x Overte.AppImage
+      - ./Overte.AppImage --appimage-extract
+      - mv squashfs-root/* ${FLATPAK_DEST}/bin/
```
hfiguiere, line 62: I am pretty sure this should be 
```suggestion
      - mv squashfs-root/* ${FLATPAK_DEST}/
```

JulianGro: It's deliberate. The AppImage isn't exactly very clean on the inside, as it is ment to be self-contained. Even putting it outside of bin/, it would not integrate with the Flatpaks directory structure, so it doesn't really matter where it is put.

hfiguiere:
bot, build org.overte.Overte

JulianGro:
@bbhtt is this still awaiting changes?

org.overte.Overte.appdata.xml
bbhtt, line 1: This file should be upstream, preferably inside the appimage.

org.overte.Overte.yml
```diff
@@ -0,0 +1,70 @@
+app-id: org.overte.Overte
+runtime: org.kde.Platform
+runtime-version: '5.15-23.08'
+sdk: org.kde.Sdk
+# QtWebEngine is not part of the KDE runtime,
+# so we use this package, which is also maintained by KDE developers.
+base: io.qt.qtwebengine.BaseApp
```
bbhtt, line 7: You forgot to add the baseapp cleanup.

bbhtt: https://github.com/flathub/io.qt.qtwebengine.BaseApp?tab=readme-ov-file#cleanup

bbhtt:
> @bbhtt is this still awaiting changes?

Yes, the process is explained in the submission docs.

org.overte.Overte.yml
JulianGro: Thanks, I missed that.

org.overte.Overte.appdata.xml
JulianGro: Is this a blocker? Because it would need to wait for the next release at least.

bbhtt: If you moved it to a git repo or somewhere you can add it as an URL to the manifest

JulianGro: Moved to git URL.

org.overte.Overte.yml
```diff
@@ -0,0 +1,75 @@
...
+      # Remove Qt shipped with AppImage
+      - rm -rf ${FLATPAK_DEST}/bin/usr/lib64/libQt*
+      - install Overte "${FLATPAK_DEST}/bin/"
+      - install -Dm644 ${FLATPAK_DEST}/bin/org.overte.interface.desktop "${FLATPAK_DEST}/share/applications/org.overte.Overte.desktop"
+      - desktop-file-edit --set-key="Exec" --set-value="Overte" --set-icon=org.overte.Overte --set-key="Categories" --set-value="Game;Network;Qt;" "${FLATPAK_DEST}/share/applications/org.overte.Overte.desktop"
+      - install -Dm644 org.overte.Overte.appdata.xml "${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.appdata.xml"
+      - install -Dm644 ${FLATPAK_DEST}/bin/Overte.png ${FLATPAK_DEST}/share/icons/hicolor/256x256/apps/org.overte.Overte.png
+      - rm ${FLATPAK_DEST}/bin/Overte.png
+
+cleanup-commands:
```
hfiguiere, line 74: so should also be more toward the top (above `modules`)

JulianGro:
bot, build

JulianGro:
I addressed the outstanding review comments and updated the Flatpak manifest. GitHub doesn't give me the option to reopen the PR. (I can of course create a new PR if wanted.)

JulianGro:
@bbhtt sorry for the ping. Could you reopen the PR, or should I open a new one?

JulianGro:
bot, build

JulianGro:
bot, build

org.overte.Overte.yml
```diff
@@ -0,0 +1,71 @@
...
+  - /app/cleanup-BaseApp.sh
+modules:
+  - name: Overte
+    buildsystem: simple
+    sources:
+      - type: file
+        only-arches:
+          - x86_64
+        url: https://public.overte.org/build/overte/release/2025.12.1/Overte-2025.12.1-x86_64.AppImage
```
bbhtt, line 31: it's open source so this needs be built from source.

JulianGro: I was hoping you might turn a blind eye, because that rule was changed just two weeks before I tried to reopen the PR, and the original PR is much older of course. https://github.com/flathub-infra/documentation/commit/f1cba5b0dce099de4e222a3aac7d572009df6200

On another note: What would constitute build from source? We might be able to feasibly build from source in an environment without a network connection nowadays, but it would be using a preassembled cache of sources.
From a technical standpoint it would look like this:
1. We build an Overte Flatpak on our own CI pipeline using a lockfile, in order to fill the Conan cache with everything we need.
2. We delete the resulting binaries from our Conan cache, leaving only the source files and recipes. (I am not sure if Conan allows deleting binaries without their caches.)
3. We export the Conan cache, put it on a web-server, and use it as a source in the Flathub Flatpak manifest.
4. I set up all this in the Flathub Flatpak manifest (setup Conan, its cache, generate all the dependencies with Conan, etc.)
5. Then theoretically, we should be able to just continue our build as normal.

Of course, I would like to avoid this convoluted process, since it doesn't provide any benefit, as far as I am aware. But if I have to do it, I want to at least be sure that it is indeed what needs to be done, rather than going through the trouble and having it rejected anyway.

hfiguiere: > I was hoping you might turn a blind eye, because that rule was changed just two weeks before I tried to reopen the PR,

Not really. Building from source have been here for a long time. And the commit you linked definitely doesn't make it new.

PR #5512, opened 2024-08-17
Add dev.toastbits.spmp

toasterofbread:

A YouTube Music client with a focus on customisation of colours and song metadata built using Compose Multiplatform.

I hope including the package files as a submodule (as I've done here) is OK. I couldn't find any information about this, but `shared-modules` seems to be intended as a submodule so I figured this would be fine as well.


hfiguiere (changes requested):
the manifest is missing.

hfiguiere:
> I hope including the package files as a submodule (as I've done here) is OK.

it's not. The manifest MUST be in this repository.

hfiguiere:
The .desktop and appstreamm file should be part of upstrea repository and installed by the build system.

toasterofbread:
> the manifest is missing.

What do you mean? `dev.toastbits.spmp.yml` is in the root directory.

> it's not. The manifest MUST be in this repository.

Is there any documentation on this? It builds perfectly fine for me using the symlinks.

hfiguiere:
This is how it looks

image

> Is there any documentation on this?

https://docs.flathub.org/docs/for-app-authors/requirements#manifest It says 

> The [application manifest](https://docs.flatpak.org/en/latest/manifests.html) must be at the top level and named after the [application ID](https://docs.flathub.org/docs/for-app-authors/requirements#application-id) with the extension .json, .yml or .yaml depending on whether it is JSON or YAML.

It's not there. I guess maybe we should add more "don't" for stuff that seems to be obvious.

toasterofbread:
That's just how Github represents [symlinks](https://en.wikipedia.org/wiki/Symbolic_link). If you clone the repository and checkout the branch, the files will be there.

toasterofbread:
image

hfiguiere:
OK, so I will be clearer.

**The manifest shall NOT be a symlink to anywhere else**

toasterofbread:
Okay, I can change it then :+1:.

So this is just an undocumented rule? It seems like a fairly standard thing to do to avoid code duplication (I'll have to maintain the manifest in two separate repositories otherwise).

hfiguiere:
`shared-modules` is something else. 

It not in the doc because we don't always think about the unreasonable things that people woul do, to add them to the documentation. Including those that make things harder for the **volunteer** reviewers to review.

hfiguiere:
Also this prevent the (volunteer) admins from fixing things. Like if you go AWOL and there is no one else that have access to said repository.

toasterofbread:
> It not in the doc because we don't always think about the unreasonable things that people woul do, to add them to the documentation. Including those that make things harder for the volunteer reviewers to review.

As I explained, I don't think it's all that unreasonable? Fair enough that it makes it harder to review though, would be nice if Github displayed symlink contents.

> Also this prevent the (volunteer) admins from fixing things. Like if you go AWOL and there is no one else that have access to said repository.

Okay, that makes a lot of sense.

dev.toastbits.spmp.desktop
hfiguiere, line 1: this belongs upstream

dev.toastbits.spmp.metainfo.xml
hfiguiere, line 1: this belong upsream

dev.toastbits.spmp.metainfo.xml
```diff
@@ -0,0 +1,97 @@
...
+                <p>Added desktop version, landscape UI, Chinese language, and other changes.</p>
+            </description>
+        </release>
+    </releases>
+
+    <provides>
+        <binary>spmp</binary>
+    </provides>
+
+    <developer id="tld.vendor">
```
hfiguiere, line 89: this developer id is wrong.

flathub.json
```diff
@@ -0,0 +1,4 @@
+{
+    "only-arches": ["x86_64"]
```
hfiguiere, line 2: why limiting architectures?

spmp.sh
```diff
@@ -0,0 +1 @@
+LD_LIBRARY_PATH="/app/lib" /app/dist/bin/spmp --flatpak "$@"
```
hfiguiere, line 1: you don't need to set library path.

flathub.json
toasterofbread: Because ARM isn't supported.

dev.toastbits.spmp.yml
```diff
@@ -0,0 +1,123 @@
...
+  - --talk-name=org.gnome.SettingsDaemon.MediaKeys
+  - --talk-name=org.kde.StatusNotifierWatcher
+
+modules:
+  - name: openjdk
+    buildsystem: simple
+    build-commands:
+      - /usr/lib/sdk/openjdk/install.sh
+
+  - name: spmp
```
hfiguiere, line 42: since this is the main app, it should be last.

dev.toastbits.spmp.yml
```diff
@@ -0,0 +1,123 @@
...
+      - mv ./* /app/dist
+    only-arches: [x86_64]
+    sources:
+      - type: file
+        path: spmp.sh
+      - type: archive
+        url: https://github.com/toasterofbread/spmp/releases/download/v0.4.0/spmp-v0.4.0-linux-x86_64.tar.gz
+        sha256: 1489923471cf10260e731a6bbadda09f1d3ae10b6d657d68100153e2cee25057
+
+  - name: other-files
```
hfiguiere, line 56: this should be part of the main module (the app)

dependencies/libayatana-appindicator3-0.1.yml
```diff
@@ -0,0 +1,68 @@
...
+      - -DENABLE_IDO=NO
+    cleanup:
+      - /lib/debug
+    sources:
+      - type: git
+        url: https://github.com/AyatanaIndicators/libayatana-indicator.git
+        tag: 0.9.4
+        commit: 611bb384b73fa6311777ba4c41381a06f5b99dad
+
+  - name: intltool
```
hfiguiere, line 40: you should use the intltool from the `shared-modules`

spmp.sh
toasterofbread: It doesn't work without that (tested it again now just to be sure), so I'd say I do need to set it.

dependencies/libayatana-appindicator3-0.1.yml
```diff
@@ -0,0 +1,68 @@
...
+        url: https://github.com/AyatanaIndicators/ayatana-ido.git
+        tag: 0.10.3
+        commit: 9906cc1d22f17c3629d32521473b172ae594ddcf
+  - name: libayatana-indicator
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DENABLE_LOADER=NO
+      - -DENABLE_IDO=NO
+    cleanup:
+      - /lib/debug
```
hfiguiere, line 33:
```suggestion
```

dependencies/libayatana-appindicator3-0.1.yml
```diff
@@ -0,0 +1,68 @@
...
+    tag: 0.5.93
+    commit: 238c8b02718fa5b4af95ede72beeed762094f4cc
+modules:
+  - name: ayatana-ido
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DENABLE_LOADER=NO
+      - -DENABLE_IDO=NO
+    cleanup:
+      - /lib/debug
```
hfiguiere, line 21:
```suggestion
```

dependencies/libayatana-appindicator3-0.1.yml
```diff
@@ -0,0 +1,68 @@
+name: libayatana-appindicator3-0.1
+buildsystem: cmake-ninja
+config-opts:
+  - -DENABLE_BINDINGS_MONO=NO
+  - -DENABLE_BINDINGS_VALA=NO
+  - -DENABLE_GTKDOC=NO
+cleanup:
+  - /lib/debug
```
hfiguiere, line 8:
```suggestion
```

dependencies/libayatana-appindicator3-0.1.yml
```diff
@@ -0,0 +1,68 @@
...
+      - --disable-dumper
+      - --disable-static
+      - --enable-tests
+      - --disable-gtk-doc
+      - --enable-introspection=no
+      - --disable-vala
+    cleanup:
+      - /share/gtk-doc
+      - /share/libdbusmenu
+      - /lib/debug
```
hfiguiere, line 64:
```suggestion
```

flathub.json
hfiguiere: but it's written in Java, isn't it supposed to run everywhere?

toasterofbread:
For the files you mentioned should be moved upstream, there's no need for either file in the base repository and I've seen several projects include those files directly in the Flathub repo. Is this a strict requirement?

flathub.json
toasterofbread: It uses several native libraries. I haven't gone through the process of testing the use of these on ARM.

toasterofbread: I'm also not sure if Compose Multiplatform (the UI framework) even supports ARM on Linux.

hfiguiere:
> there's no need for either file in the upstream repository 

A desktop application that doesn't ship a .desktop, it is a bug. As for the appstream file it is NOT flatpak specific either and is used also by other package formats.

> and I've seen several projects include those files directly in the Flathub repo. 

hfiguiere:
bot, build dev.toastbits.spmp

dev.toastbits.spmp.desktop
bbhtt, line 1: This belongs upstream.

dev.toastbits.spmp.metainfo.xml
```diff
@@ -0,0 +1,97 @@
...
+            <caption>Home page</caption>
+        </screenshot>
+        <screenshot>
+            <image>https://raw.githubusercontent.com/toasterofbread/spmp/35d683bcbcef06b6622961c98464de610b5e5925/metadata/en-US/images/phoneScreenshots/landscape_0.png</image>
+            <caption>Player page</caption>
+        </screenshot>
+    </screenshots>
+
+    <icon type="stock">dev.toastbits.spmp</icon>
+
```
bbhtt, line 40:
```suggestion
```

dev.toastbits.spmp.metainfo.xml
```diff
@@ -0,0 +1,97 @@
...
+    <metadata_license>CC0-1.0</metadata_license>
+    <project_license>GPL-3.0-only</project_license>
+    <content_rating type="oars-1.0" />
```
bbhtt, line 43:
```suggestion
    <content_rating type="oars-1.1" />
```

dev.toastbits.spmp.metainfo.xml
```diff
@@ -0,0 +1,97 @@
...
+            <description>
+                <p>Added desktop version, landscape UI, Chinese language, and other changes.</p>
+            </description>
+        </release>
+    </releases>
+
+    <provides>
+        <binary>spmp</binary>
+    </provides>
+
```
bbhtt, line 88:
```suggestion
```

spmp.sh
bbhtt: > It doesn't work without that (tested it again now just to be sure), so I'd say I do need to set it.

What's the issue? Flatpak handles this by default.

dev.toastbits.spmp.yml
```diff
@@ -0,0 +1,118 @@
...
+
+  # Notifications
+  - --talk-name=org.freedesktop.Notifications
+
+  # Audio playback
+  - --filesystem=xdg-run/pipewire-0:ro
+  - --socket=pulseaudio
+
+  # App indicator
+  - --talk-name=org.gnome.SettingsDaemon.MediaKeys
```
bbhtt, line 33:
```suggestion
```

This interface was removed a long time ago, in GNOME 42

.gitmodules
hfiguiere, line 1: this should be removed too, there is no submodule. (it should have been)

toasterofbread: This has nothing to do with the submodule, it's for temporary build files.

spmp.sh
toasterofbread: LD_LIBRARY_PATH is used by Java to set `java.library.path`, which is used to find and load libraries dynamically at runtime. Whatever Flatpak does doesn't cover this case I guess,

.gitmodules
hfiguiere: Not sure what kind of fiction you are writing. Just remove it.

toasterofbread: Sorry, I just realised this was for `.gitmodules`. Github UI made me think it was `.gitignore` for some reason. No need to be so rude about it though...

dev.toastbits.spmp.yml
```diff
@@ -0,0 +1,124 @@
...
+
+  - name: spmp
+    buildsystem: simple
+    build-commands:
+      - install -D icon.png /app/share/icons/hicolor/512x512/apps/dev.toastbits.spmp.png
+
+      - install -D dev.toastbits.spmp.metainfo.xml /app/share/metainfo/dev.toastbits.spmp.metainfo.xml
+
+      - sed -i '0,/^Exec=/s/^Exec=.*/Exec=spmp/' dev.toastbits.spmp.desktop
+      - sed -i '0,/^Icon=/s/^Icon=.*/Icon=dev.toastbits.spmp/' dev.toastbits.spmp.desktop
```
bbhtt, line 102: Use `desktop-file-edit` here

dev.toastbits.spmp.metainfo.xml
bbhtt: this hasn't been moved

toasterofbread: I [asked](https://github.com/flathub/flathub/pull/5512#issuecomment-2294948128) if this was a strict requirement and I didn't get a yes. The file isn't needed upstream so keeping it in the one place it's needed would be preferable to me.

bbhtt: The answer to that was 

>  As for the appstream file it is NOT flatpak specific either and is used also by other package formats.

I don't see any reason to not move it upstream.

toasterofbread: It's not Flatpak-specific, but it's specific to this repository. Why keep a file in an unrelated repository (upstream), instead of the single place it's actually used? I'd need to commit to upstream and to here if I wanted to make a single metadata change, for example.

bbhtt: > It's not Flatpak-specific, but it's specific to this repository. 

What content makes it specific to this repository but not Flatpak specific?

toasterofbread: It's not the content, it's the format. I don't have a need for a `metainfo.xml` for anything other than publishing to Flathub, so nowhere other than this repository is likely to ever need that file.

hfiguiere: metainfo (appstream) files are used widely on Linux distributions. While some DE make it standard to support them (see GNOME software and KDE Discover), they are not required. Flathub however require one.

There are many applications that shipped one before even Flathub existed. Debian and Fedora install them.

bbhtt: "don't need" is because no distribution enforces it as a requirement, but you should include it in every other packaging format so that when people install it from those they see the same metadata.

Having it in multiple places will cause divergence.

You already said "It's not Flatpak-specific" so there's no reason to not include it in other packaging formats

toasterofbread: What other package formats/distributions support metainfo? This is my first time encountering the format so I'm not very familiar with it. If other packaging formats I'm using (or will use) support it then moving the file upstream would make sense.

bbhtt: everything else, check any app from your distro. You install it to the same location as here when packaging for distro: $PREFIX/share/metainfo.

bbhtt: https://archlinux.org/packages/extra-testing/x86_64/firefox/ has a metainfo file.

toasterofbread: Oh, it's a Freedesktop thing (wish I'd realised that sooner). I probably won't be packaging for anything that uses it, but I guess it makes sense to have it upstream if it's a widely adopted standard (again, sorry for the ignorance). Will get it moved soon.

bbhtt:
bot, build dev.toastbits.spmp

PR #5517, opened 2024-08-19
Add org.syntalos.syntalos

ximion:
Syntalos is a software for timestamp-synchronized parallel data acquisition from diverse data sources, such as cameras, microendoscopes, Intan electrophysiology amplifiers or Firmata-based serial interfaces. The software also allows closed-loop interventions via its built-in Python scripting support. It is specifically designed for use in (neuro)scientific in vivo behavior experiments. 

This software already exists on Flathub as https://github.com/flathub/io.github.bothlab.syntalos
It has recently been rehomed, which allowed us to simplify some procedures (and shorten file paths ^^).

The project will also release version 2.0 this week, which contains breaking changes. So right now is a good time to rename the project's ID to reflect the new location as well, therefore I am following the procedure outlined at https://docs.flathub.org/docs/for-app-authors/maintenance#renaming-the-flatpak-id


io.github.syntalos.yaml
```diff
@@ -0,0 +1,244 @@
+id: io.github.syntalos
```
bbhtt, line 1: Code hosting IDs needs 4 components at least https://docs.flathub.org/docs/for-app-authors/requirements#control-over-domain-or-repository

`io.github.syntalos.syntalos` would be fine to match the repo name, we automaitcally check through the linter if the repo exists by reversing the rDNS

io.github.syntalos.yaml
```diff
@@ -0,0 +1,244 @@
...
+rename-desktop-file: syntalos.desktop
+
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=x11
+  - --device=all
+  - --filesystem=host
+  - --socket=system-bus
+  - --socket=session-bus
```
bbhtt, line 20: These two needs exceptions https://docs.flathub.org/docs/for-app-authors/linter#finish-args-arbitrary-dbus-access

https://docs.flathub.org/docs/for-app-authors/linter#exceptions

bbhtt: why are they needed?

io.github.syntalos.yaml
```diff
@@ -0,0 +1,244 @@
...
+command: syntalos
+rename-icon: syntalos
+rename-desktop-file: syntalos.desktop
+
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=x11
+  - --device=all
+  - --filesystem=host
```
bbhtt, line 18: why is host needed?

io.github.syntalos.yaml
```diff
@@ -0,0 +1,244 @@
...
+
+add-extensions:
+  org.freedesktop.Platform.GL:
+    directory: lib/GL
+    version: '23.08'
+    subdirectories: true
+    add-ld-path: lib
+    merge-dirs: vulkan/icd.d;glvnd/egl_vendor.d
+    download-if: active-gl-driver
+    enable-if: active-gl-driver
```
bbhtt, line 38:
```suggestion
```

This is installed automatically by the runtime

io.github.syntalos.yaml
```diff
@@ -0,0 +1,244 @@
...
+
+modules:
+- name: essential-dirs
+  buildsystem: simple
+  build-commands:
+    - mkdir -p /app/lib/GL
+
```
bbhtt, line 45:
```suggestion

```

io.github.syntalos.yaml
```diff
@@ -0,0 +1,244 @@
...
+    - type: archive
+      url: https://github.com/KhronosGroup/OpenCL-Headers/archive/v2023.12.14/OpenCL-Headers-v2023.12.14.tar.gz
+      sha256: 407d5e109a70ec1b6cd3380ce357c21e3d3651a91caae6d0d8e1719c69a1791d
+
+- name: meson
+  buildsystem: simple
+  build-commands:
+    - python3 setup.py build
+    - python3 setup.py install --prefix=/app --root=/
+  sources:
```
bbhtt, line 58: Add 

```yaml
cleanup:
- "*"
```

io.github.syntalos.yaml
```diff
@@ -0,0 +1,244 @@
...
+    - type: archive
+      url: https://github.com/Cyan4973/xxHash/archive/refs/tags/v0.8.2.tar.gz
+      sha256: baee0c6afd4f03165de7a4e67988d16f0f2b257b51d0e3cb91909302a26a79c4
+
+- name: toml++
+  buildsystem: meson
+  builddir: true
+  config-opts:
+    - --libdir=lib
```
bbhtt, line 81: is this correct? flatpak-builder sets `/app/lib` as default for meson. So this could be not needed.

io.github.syntalos.yaml
```diff
@@ -0,0 +1,244 @@
...
+- name: pybind11
+  buildsystem: simple
+  build-commands:
+    - python3 setup.py build
+    - cmake -DPYBIND11_INSTALL=ON -DPYBIND11_TEST=OFF -DCMAKE_INSTALL_INCLUDEDIR:PATH=${FLATPAK_DEST}/include
+      -DCMAKE_INSTALL_LIBDIR:PATH=${FLATPAK_DEST}/lib -DCMAKE_INSTALL_DATAROOTDIR:PATH=${FLATPAK_DEST}/share
+      .
+    - python3 setup.py install --prefix=${FLATPAK_DEST}
+    - cmake --build .
+    - cmake --install .
```
bbhtt, line 188: Using ninja should be faster here.

bbhtt: Also why use cmake? pybind can be installed with pip directly

io.github.syntalos.yaml
```diff
@@ -0,0 +1,244 @@
...
+  sources:
+    - type: archive
+      url: https://github.com/pybind/pybind11/archive/refs/tags/v2.10.4.tar.gz
+      sha256: 832e2f309c57da9c1e6d4542dedd34b24e4192ecb4d62f6f4866a737454c9970
+
+- name: python3-numpy
+  buildsystem: simple
+  build-commands:
+    - python3 setup.py build -j6
```
bbhtt, line 197:
```suggestion
    - python3 setup.py build -j$FLATPAK_BUILDER_N_JOBS
```

modules/tiscamera.yaml
```diff
@@ -0,0 +1,41 @@
...
+- name: libzip
+  buildsystem: cmake-ninja
+  config-opts:
+    - -DCMAKE_INSTALL_LIBDIR=lib
+  cleanup:
+    - "/include"
+    - "/bin"
+    - "/share"
+    - "/lib/pkgconfig"
+    - "/lib/*.la"
```
bbhtt, line 13:
```suggestion
    - "*.la"
```

io.github.syntalos.yaml
ximion: This rule feels a bit arbitrary - we would really like to keep the much shorter name and not have to rename the app again, as due to how Flatpak works the ID shows up in many file paths that users will have to access.

There are also many examples on Flathub that do not follow this rule, e.g. `io.github.Hexchat`, `io.github.ImEditor`, `io.github.OpenToonz` and many more...

bbhtt: Those were submitted before the rule came, the issues with 3 components is explained in https://docs.flatpak.org/en/latest/conventions.html#application-ids

io.github.syntalos.yaml
ximion: People use this software in environments where they will want to store data in locations like `/mnt`, `/srv` or even custom other paths depending on what policy the research organization has for live data storage.
We have no idea where data will be saved in advance, and given that the software is already very weakly sandboxed, just allowing most locations will result in the fewest complaints from users when they use the Flatpak bundle over a native package  :-)

io.github.syntalos.yaml
bbhtt: This convention existed for a long time, we started enforcing it since 2022

io.github.syntalos.yaml
ximion: There were some issues with that in the past, but this config section has existed for a long while... So, let's drop it and see what happens!
Thanks for the suggestion!

io.github.syntalos.yaml
bbhtt: > These requirements aren't new though

You need the exception PR still I guess just rename the existing one https://github.com/flathub-infra/flatpak-builder-lint/blob/f52259b6f3c20fb1c40889d068633a94b893137b/flatpak_builder_lint/staticfiles/exceptions.json#L968-L969

But the ID needs to be fixed first, exceptions are tied to the ID.

io.github.syntalos.yaml
ximion: It is, actually - there is a module in Syntalos that compiles & runs C++ code and uses Meson.

But we will need to test that further in the Flatpak bundle and see how we want to deal with this module in Flatpak at all, so I'm fine with the cleanup rule until we have figured the C++ compiling out (this will be tricky and I made it a problem of future-me).

io.github.syntalos.yaml
ximion: I faintly remember this being required, because Meson otherwise did weird things... But I could remove it and see what happens, this is also fairly old.

io.github.syntalos.yaml
ximion: Okay, this is a problem actually (as silly as that sounds...) - I will have to have some conversations here to see what we can do about it.

modules/tiscamera.yaml
```diff
@@ -0,0 +1,41 @@
+name: tiscamera
+modules:
+
+- name: libzip
+  buildsystem: cmake-ninja
+  config-opts:
+    - -DCMAKE_INSTALL_LIBDIR=lib
```
bbhtt, line 7:
```suggestion
```

This is set by default as well for a long time.

modules/tiscamera.yaml
```diff
@@ -0,0 +1,41 @@
...
+  sources:
+    - type: archive
+      url: https://libzip.org/download/libzip-1.8.0.tar.xz
+      sha256: f0763bda24ba947e80430be787c4b068d8b6aa6027a26a19923f0acfa3dac97e
+
+- name: tiscamera
+  buildsystem: cmake-ninja
+  builddir: true
+  config-opts:
+    - -DCMAKE_INSTALL_LIBDIR=lib
```
bbhtt, line 23:
```suggestion
```

modules/tiscamera.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - -DTCAM_INSTALL_SYSTEMD=/app/lib/systemd/system/
+    - -DTCAM_BUILD_WITH_GUI=OFF
+    - -DTCAM_BUILD_DOCUMENTATION=OFF
+    - -DTCAM_DOWNLOAD_MESON=OFF
+    - -DTCAM_INTERNAL_ARAVIS=OFF
+    - -Daravis_INCLUDE_DIR=/app/include/aravis-0.8
+    - -DGOBJECT_INTROSPECTION_1.0_GIRDIR=/app/share/gir-1.0/
+    - -DGOBJECT_INTROSPECTION_1.0_TYPELIBDIR=/app/lib/girepository-1.0
+    - -DGSTREAMER_1.0_PLUGINSDIR=/app/lib/gstreamer-1.0
+    - -DGSTREAMER_1.0_INCLUDEDIR=/include/gstreamer-1.0
```
bbhtt, line 35: Is this correct? shouldn't it be `/app/include/gstreamer-1.0` ?

io.github.syntalos.yaml
ximion: (With "here" I mean at my institution, not "in this issue" ^^ - either we move to a new domain or have to change the ID again, I guess)

bbhtt: Preferably `io.github.syntalos.syntalos`, then linter can calculate the repo to be github.com/syntalos/syntalos and check it exists.

If you differ on the last component, then it'd need an exception.

modules/tiscamera.yaml
ximion: I guess it should be... It's just weird that this has never failed. This section has not been touched for a long time.

hfiguiere: mayeb it is not needed and can be dropped?

io.github.syntalos.yaml
ximion: Curious question: What if a `vcs-browser` URL is also present in the metadata? https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-url

Will you ignore the ID then, instead of distilling meaning other than "project owns web location" from it? What happens if those two values differ and point to different VCS browser locations?

io.github.syntalos.yaml
bbhtt: You can but I can't merge it until the ID is settled.

io.github.syntalos.yaml
bbhtt: Appstream tags aren't read for this. We verify the ID and domain/webpage manually and through linter.

Appstream tags can be changed without going through us, the ID cannot.

io.github.syntalos.yaml
ximion: Yeah, that's to be expected. Give me a day to talk to some people (I really want to avoid having to change that ID more than once).

hfiguiere: get them to buy a domain and use that....

io.github.syntalos.yaml
```diff
@@ -0,0 +1,228 @@
...
+  - --share=ipc
+  - --socket=x11
+  - --device=all
+  - --filesystem=host
+  - --socket=system-bus
+  - --socket=session-bus
+  - --allow=devel
+  - --allow=bluetooth
+  - --env=PATH=/app/bin:/usr/bin:/usr/sbin
+  - --env=GST_PLUGIN_SYSTEM_PATH=/app/lib/gstreamer-1.0:/usr/lib/extensions/gstreamer-1.0:/usr/lib/x86_64-linux-gnu/gstreamer-1.0
```
bbhtt, line 24: The last path hardcodes the arch triplet in it and the second one should not be needed. No extensions are mounted there.

bbhtt:
Doing a build to check for any build issues.

bot, build io.github.syntalos.yaml

bbhtt:
bot, build io.github.syntalos

bbhtt:
Ah yea forgot about the exceptions, it won't work until the exception on the new ID is merged.

io.github.syntalos.yaml
bbhtt: You can set this via a wrapper for the main executable since that'd deal with runtime changes automatically.


finish-args can't do variable substitutions.

io.github.syntalos.yaml
ximion: > get them to buy a domain and use that....

Well, that was exactly my plan, and it succeeded :-)
(it has been in the making before, and I just had to accelerate things a bit)

io.github.syntalos.yaml
bbhtt: > I hate wrappers... But maybe we can avoid one still, I have to test the result.

The other option is to just put the aarch64 libdir in there too since one will not have any effect on the other.

ximion:
@bbhtt I've created https://github.com/flathub-infra/flatpak-builder-lint/pull/469 to update the ID, and am currently building this locally again :-)

The ID is renamed now.

bbhtt:
waiting for the exception to get deployed.

ximion:
So, funny thing... The application does not run because it relies on `clone3` in a critical path and relies on the features it offers...
https://github.com/flatpak/flatpak/blob/main/common/flatpak-run.c#L1906

This is extremely crappy. I'll have a look if there is any reliable way to change the code so it can obtain its pidfds when spawning processes reliably.

I doubt there is any way to whitelist `clone3` in Flatpak for this app, and I also never thought that this would be an issue (it didn't feel particularly security-sensitive).
If you have ideas, please let me know! I'll have a look if any substitution is possible.

ximion:
That was a lot easier than thought, but RtKit communication fails. I guess that is because RtKit takes a PID as parameter for the DBus calls, and within the sandbox, the PIDs are different from the ones outside.

I can maybe write something to prefer the portal if possible... Anyway, since most things work, this shouldn't block the merging of this change and I will work on a solution in the coming days (if I can't get one done today).

ximion:
> I will work on a solution in the coming days (if I can't get one done today)

Well, this also was easier than assumed. The current 2.0 build runs as well as the 1.x Flatpak builds :-)

bbhtt:
bot, build org.syntalos.syntalos

ximion:
> Build [141276 failed](https://buildbot.flathub.org/#/builders/6/builds/141276)

Weird linking error in OpenCV on arm64 - I assume one of the convolution functions in the DNN module hasn't been properly compiled on that architecture. Disabling DNN on arm64 is fine for now, so if the change works as intended, this should build on arm64 now (that is, if my assumption is correct that cmake takes the last passed parameter in the list, overriding any parameter passed prior).

bbhtt:
Feel free to start the build on your own

bot, build org.syntalos.syntalos

ximion:
> Feel free to start the build on your own

Ah, I didn't know I had permission! :-) I hope this is an easy fix, arm64 build are very annoying to test locally.

bbhtt:
The docs link in help menu is 404.

I assume you tested this already.

ximion:
> The docs link in help menu is 404.

Yes, it's already fixed :-) Otherwise, things seem to work fine (but before 2.0 is properly released, we will do a bit more testing on the Flatpak bundle, of course).

Thank you for looking into it as well! :-)

ximion:
By the way, is it okay if I wait a few days with deprecating the old ID, following the procedure outlined in the docs? That would be very convenient for us, allowing us to test the 2.0 Flatpak build a bit better and then make the switch. The planned release date for 2.0 is this Saturday (24.08.), so it wouldn't be a long delay.

bbhtt:
Yea that's fine.

ximion:
bot, build org.syntalos.syntalos

ximion:
This new build works fine as well, all it changed was to fix some links anyway :-)

PR #5537, opened 2024-08-27
Add org.damask_multiphysics.DAMASK_mesh

MarDiehl:

    DAMASK is a unified multi-physics crystal plasticity simulation package. The solution of continuum mechanical boundary value problems requires a constitutive response that connects deformation and stress at each material point. This problem is solved in DAMASK on the basis of crystal plasticity using a variety of constitutive models and homogenization approaches. However, treating mechanics in isolation is no longer sufficient to study emergent advanced high-strength materials. In these materials, deformation happens interrelated with displacive phase transformation, significant heating, and potential damage evolution. Therefore, DAMASK is capable of handling multi-physics problems. Following a modular approach, additional field equations are solved in a fully coupled way using a staggered approach.

    The mesh solver (DAMASK_mesh) operates on unstructered meshes.

- [x] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about this submission. https://damask-multiphysics.org/development


bbhtt:
Is this a different solver that damask_grid submission? Any reason you can't build both in the same package?

MarDiehl:
> Is this a different solver that damask_grid submission? Any reason you can't build both in the same package?

DAMASK_grid and DAMASK_mesh are two different executables which are build separately but share some backend code (the "material point model"). When distributing via other channels (https://damask-multiphysics.org/installation/package_manager) we use the same approach and additionally provide a ``damask'' metapackage which also includes Python tools for pre- and post-processing. But since Python libraries are not within the scope of Flathub (at lease I could not find things like numpy etc.) I did not consider that here.

bbhtt:
You can ship python stuff within the flatpak, anyone can use run `flatpak run --command=python org.damask_multiphysics.DAMASK_mesh` -c "import damask; foo bar" but using extra python modules outside of the runtime/sdk involves setting some env vars manually.

> we use the same approach

So what's the actual reason here? Does it get too big etc. ? I don't see why you can't build both executables in the same one.

MarDiehl:
> You can ship python stuff within the flatpak, anyone can use run `flatpak run --command=python org.damask_multiphysics.DAMASK_mesh -c "import damask; foo bar"` but using extra python modules outside of the runtime/sdk involves setting some env vars manually.

The workflow consists of pre-processing (using the Python library interactively or via a script), running the simulation (using DAMASK_mesh) and post-processing (again with the Python library). For serious use, the simulation is often conducted on a more powerful computer. The Python library needs not compiled code but depends on other Python packages (NumPy, SciPy, h5py, and a few others) and can be installed easily via PIP. The bottleneck for starters using their own computer and not a workstation is always the compiled code, for which Flathub is handy.

> 
> > we use the same approach
> 
> So what's the actual reason here? Does it get too big etc. ? I don't see why you can't build both executables in the same one.

No, both packages are indeed tiny. But we want to communicate to the users that both solvers are independent and also allow to install the solvers individually.

How would it actually work for Flatpak to call different executables?

bbhtt:
You have to pass `--command=`, you can make two wrapper scripts in `~/.local/bin` like so:

```
#!/bin/sh

exec flatpak run --command=damask_mesh org.damask_multiphysics.DAMASK_mesh ${@}
```

Or you can make a wrapper in the Flatpak that asks if the user wants to start the mesh or grid binary.

bbhtt:
> require something like

Yea correct. The default without a `--command` flag would be whatever you set in the `command` key in manifest

PR #5541, opened 2024-08-27
Add to.holepunch.Keet

butera-simone:

Chat directly with your friends. Send messages, files, high quality videos and photos with no limitations. Private and Encrypted. Always Peer-to-Peer. Never on a server.

- [X] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about this submission. **Link:** https://holepunch.to/


to.holepunch.Keet.yml
```diff
@@ -0,0 +1,52 @@
+id: to.holepunch.Keet
+runtime: org.gnome.Platform
+runtime-version: '46'
+sdk: org.gnome.Sdk
+command: /app/bin/usr/bin/keet
```
hfiguiere, line 5: comman should be in `/app/bin` and here it should be a simple command name.

to.holepunch.Keet.yml
```diff
@@ -0,0 +1,52 @@
...
+base: org.electronjs.Electron2.BaseApp
+base-version: '23.08'
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --socket=x11
```
hfiguiere, line 12: it's either wayland + fallback-x11 or x11. not both.

to.holepunch.Keet.yml
```diff
@@ -0,0 +1,52 @@
...
+  - --socket=pulseaudio
+  - --device=all
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --env=XCURSOR_PATH=/run/host/user-share/icons:/run/host/share/icons
+  - --talk-name=org.freedesktop.ScreenSaver
+  - --filesystem=host
```
hfiguiere, line 18: why does a chat program need host access?

to.holepunch.Keet.yml
```diff
@@ -0,0 +1,52 @@
...
+modules:
+  - name: assets
```
hfiguiere, line 20: these 3 modules should only be one.

to.holepunch.Keet.yml
```diff
@@ -0,0 +1,52 @@
...
+    buildsystem: simple
+    build-commands:
+      - mkdir -p /app/share/metainfo/
+      - mkdir -p /app/share/icons/hicolor/512x512/apps/
+      - mkdir -p /app/share/applications/
+      - cp -r assets/. /app/bin/ 
+      - cp to.holepunch.Keet.metainfo.xml /app/share/metainfo/
+      - cp to.holepunch.Keet.png /app/share/icons/hicolor/512x512/apps/
+      - cp to.holepunch.Keet.desktop /app/share/applications/
```
hfiguiere, line 29:
```suggestion
      - cp -r assets/. /app/bin/ 
      - install -Dm644 to.holepunch.Keet.metainfo.xml -t /app/share/metainfo/
      - install -Dm644 to.holepunch.Keet.png -t /app/share/icons/hicolor/512x512/apps/
      - install -Dm644 to.holepunch.Keet.desktop -t /app/share/applications/
```

to.holepunch.Keet.yml
```diff
@@ -0,0 +1,52 @@
...
+    sources:
+      - type: dir
```
hfiguiere, line 31: don't use type `dir`.

to.holepunch.Keet.yml
```diff
@@ -0,0 +1,52 @@
...
+        path: .
+  - name: keet-x64
+    only-arches:
+      - x86_64
+    buildsystem: simple
+    build-commands:
+      - cp keet /app/bin/usr/bin/
+    sources:
+      - type: archive
+        url: https://keet.io/downloads/latest/keet-x64-binary.tar.gz
```
hfiguiere, line 41: this is an unversionned URL. Which mean the build WILL break. It should be a specific version.

to.holepunch.Keet.yml
```diff
@@ -0,0 +1,52 @@
...
+        sha512: 9ad6b93d874dc6a421f0e0e619d1e520238c65da980a14f0304604eb4e5c9cf821e86439e51c45e3c0e066d045bc1e703b074616302f2897e940ddf0fbb6c305
+  - name: keet-arm64
+    only-arches:
+      - aarch64
+    buildsystem: simple
+    build-commands:
+      - cp keet /app/bin/usr/bin/
+    sources:
+      - type: archive
+        url: https://keet.io/downloads/latest/keet-arm64-binary.tar.gz
```
hfiguiere, line 51: same

to.holepunch.Keet.yml
```diff
@@ -0,0 +1,52 @@
+id: to.holepunch.Keet
+runtime: org.gnome.Platform
```
hfiguiere, line 2: why do you need the gnome platform here since it's Electron based.

to.holepunch.Keet.desktop
hfiguiere, line 1: the file should be upstream.

assets/icon.png
hfiguiere, line 1: what is this used for ? It's not the icon

assets/splash.png
hfiguiere, line 1: what is this used for?

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64"]
```
hfiguiere, line 2: but yet you try to unpack an aarch64 binary as well. Just remove this.

to.holepunch.Keet.metainfo.xml
```diff
@@ -0,0 +1,55 @@
...
+  </developer>
+  <branding>
+    <color type="primary" scheme_preference="light">#FFFFFF</color>
+    <color type="primary" scheme_preference="dark">#111921</color>
+  </branding>
+  <launchable type="desktop-id">to.holepunch.Keet.desktop</launchable>
+  <releases>
+      <release version="1.0.0" date="2024-07-17"/>
+  </releases>
+  <content_rating type="oars-1.0">
```
hfiguiere, line 25:
```suggestion
  <content_rating type="oars-1.1">
```

to.holepunch.Keet.png
hfiguiere, line 1: icon should be upstream

to.holepunch.Keet.yml
butera-simone: The keet binary requires a certain directory structure to work correctly (the one from /assets). The path to the assets is hardcoded [in the C code](https://github.com/holepunchto/libpear/blob/c9931ebf26aa48a171276edfab983aa4d745542b/src/pear.c#L105). I guess we could change the name of the directories to avoid the redundant "bin" in the flatpak? How would you prefer it to be called?

to.holepunch.Keet.yml
butera-simone: (I'm not sure this is the right permission, but) the app can send/receive files. Also it needs to install the Pear platform in `.config`

to.holepunch.Keet.yml
butera-simone: the installer for the Pear platform uses native GUI for its splash screen

assets/splash.png
butera-simone: it's for the Pear platform installer splash screen

to.holepunch.Keet.yml
hfiguiere: define "native GUI"

flathub.json
butera-simone: we had a problem with flatpak on ARM so I added this temporarily to avoid users downloading it on that architecture

assets/splash.png
hfiguiere: then why isn't this part of the distribution tarball?

flathub.json
hfiguiere: then don't put it in the manifest, it's misleading. and disappoiting.

screenshots/1.RandomChat.png
hfiguiere, line 1: screenshots should be upstrean.

to.holepunch.Keet.yml
hfiguiere: this is unreliable in general. `/usr` is a prefix and finding it anywhere down is not standard. Here the prefix is `/app`.

Maybe you should be from source if that coded like that. at least the flatpak content is immutable.

to.holepunch.Keet.yml
hfiguiere: it maybe not. it should install stuff in `XDG_CONFIG_HOME`. Also what is that "pear platform"/"pear installer" ? I thought this was chat program

to.holepunch.Keet.yml
hfiguiere: `type: file`. If you feel there are too many then see my comment about stuff that should be upstream: all but the manifest and eventual `flathub.json`.

to.holepunch.Keet.yml
butera-simone: sorry I don't understand what you wrote, could you please rephrase that?

to.holepunch.Keet.yml
butera-simone: it's a messaging app like Discord, but p2p, and it uses our [Pear platform ](https://docs.pears.com/)

to.holepunch.Keet.yml
butera-simone: it calls into GTK

hfiguiere: so basically you don't need. gtk3 is part of the freedesktop runtime.

to.holepunch.Keet.yml
bbhtt: > assets is hardcoded [in the C code](https://github.com/holepunchto/libpear/blob/c9931ebf26aa48a171276edfab983aa4d745542b/src/pear.c#L105).

This should be fixed

butera-simone: I'm not sure I understand, do you want the assets used by the app upstream to be stored in flatpak-specific paths?

bbhtt: $prefix is not Flatpak specific.

butera-simone:
@hfiguiere could you review it again? I think it should fit now

to.holepunch.Keet.yml
```diff
@@ -0,0 +1,28 @@
...
+  - --device=all
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --env=XCURSOR_PATH=/run/host/user-share/icons:/run/host/share/icons
+  - --talk-name=org.freedesktop.ScreenSaver
+  - --filesystem=host
+modules:
+  - name: assets
+    buildsystem: simple
+    build-commands:
+      - cp -r assets/. /app/bin/
```
bbhtt, line 21:
```
file assets/keet 
assets/keet: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=027fd1480e7f57dabda0d34b95e1852e549a7642, for GNU/Linux 3.2.0, not stripped
```

this won't work on aarch64, also why not build from source?

to.holepunch.Keet.yml
bbhtt: > (I'm not sure this is the right permission, but) the app can send/receive files.

why does it not use portals?

to.holepunch.Keet.yml
```diff
@@ -0,0 +1,28 @@
+id: to.holepunch.Keet
+runtime: org.gnome.Platform
+runtime-version: '46'
```
bbhtt, line 3: 47 is latest

to.holepunch.Keet.yml
bbhtt: this is still a path, it just needs to be `command: foo`.

to.holepunch.Keet.yml
```diff
@@ -0,0 +1,28 @@
...
+  - name: assets
+    buildsystem: simple
+    build-commands:
+      - cp -r assets/. /app/bin/ 
+      - install -Dm644 to.holepunch.Keet.metainfo.xml -t /app/share/metainfo/
+      - install -Dm644 to.holepunch.Keet.png -t /app/share/icons/hicolor/512x512/apps/
+      - install -Dm644 to.holepunch.Keet.desktop -t /app/share/applications/
+    sources:
+      - type: archive
+        url: https://keet.io/downloads/flatpak/keet-3.18.1.tar
```
bbhtt, line 27: why is the id to.holepunch while this goes to https://keet.io/?

to.holepunch.Keet.yml
hfiguiere: > it's a messaging app like Discord, but p2p, and it uses our [Pear platform ](https://docs.pears.com/)

why doesn't ship it instead?

If it still fail to use XDG dir, then it should persist `.config` instead instead of a wide open filesystem access.

to.holepunch.Keet.yml
hfiguiere: you moved from two binaries / two arches but one disabled, to one x86_64 binaries for both architecture.

butera-simone: what would be your preferred way to specify that this is only for x86 ?

to.holepunch.Keet.yml
butera-simone: what should the id be ?

to.holepunch.Keet.yml
butera-simone: re: building from source, my understanding was that it was not a hard requirement for inclusion on the store?

butera-simone: some of the dependencies are not open source

bbhtt: > what would be your preferred way to specify that this is only for x86 ?

it's in the doc https://docs.flathub.org/docs/for-app-authors/requirements#flathubjson

to.holepunch.Keet.yml
bbhtt: `io.keet.keet`? is that a problem?

to.holepunch.Keet.yml
butera-simone: that was in the submission but I was [told to remove it](https://github.com/flathub/flathub/pull/5541#discussion_r1733432063). I'll include it again

to.holepunch.Keet.yml
butera-simone: I don't think so

to.holepunch.Keet.yml
bbhtt: well you removed the aarch64 source as well

to.holepunch.Keet.yml
butera-simone: what is the problem with filesystem access? Could you please be more specific? Like, "giving full access is against our rules because of X, you have to use Y instead, which will limit the app to Z"

to.holepunch.Keet.yml
hfiguiere: no you were asked to remove the `only-arches`. Really.

to.holepunch.Keet.yml
bbhtt: > what is the problem with filesystem access?

filesystem access bypasses the sandbox, making it limited avoiding widespread access should be avoided.

butera-simone: Ok so, we need: 
- a persistent `.config` that we can write/read/execute for the platform. It can be in the container but it needs to be updatable for when we release new versions of the platform
- access to the wider filesystem to read/write, for the file sharing functionality of the chat app

What would be a permission setup that would allow us to do that while being accepted on the store?

bbhtt: > a persistent pear folder in .config that we can write/read/execute 

Does it use $XDG_CONFIG_HOME or $HOME/.config?

> access to the wider filesystem to read/write, for the file sharing functionality of the chat app (the app can send/receive files like most chat apps)

Does it support filechooser portals?

butera-simone: The platform is at ` $HOME/.config`

File access is done via https://docs.libuv.org/en/v1.x/misc.html#c.uv_os_homedir

to.holepunch.Keet.yml
```diff
@@ -0,0 +1,28 @@
+id: to.holepunch.Keet
+runtime: org.gnome.Platform
+runtime-version: '46'
+sdk: org.gnome.Sdk
+command: /app/bin/keet
```
hfiguiere, line 5: this shall not be a path. I think this has already been said. and ignored.

to.holepunch.Keet.yml
bbhtt: > The platform is at  $HOME/.config

Can you make it follow $XDG_CONFIG_HOME? Then nothing is needed.

PR #5543, opened 2024-08-27
Add io.github.breadmakesyoufull.bpm2fps

BreadMakesYouFull:

  Match song beats to their closest frame for animation.

### DIFF
--- a/io.github.breadmakesyoufull.bpm2fps.yaml
+++ b/io.github.breadmakesyoufull.bpm2fps.yaml
@@ -12,6 +12,8 @@ finish-args:
   - --socket=fallback-x11
   - --socket=wayland
   - --device=dri
+  - --filesystem=xdg-documents
+  - --filesystem=xdg-desktop
 modules:
   - name: bpm2fps
     buildsystem: simple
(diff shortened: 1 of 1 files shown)


bbhtt:
bot, build io.github.breadmakesyoufull.bpm2fps

bbhtt:
The file doesn't get saved unless the extension is added, and the only option in the filechooser seems to be json.

[gedG.webm](https://github.com/user-attachments/assets/8bf582d5-b077-4645-b1f0-f9f6359c01ae)

BreadMakesYouFull:
Thanks for taking a look so swiftly @bbhtt and the thorough testing. Apologies for missing this, still learning about flatpak permissions portals etc, and should've known opening the dialog isn't enough to check the export is working!

I can see one way forwards would be to expose xdg-documents or home:
``flatpak run --filesystem=xdg-documents io.github.breadmakesyoufull.bpm2fps``

But there is not a way to tell the QML dialog that defers to the system dialog that it should only show a specific folder:
https://doc.qt.io/qt-6/qml-qtquick-dialogs-filedialog.html 

I'll have a think about how to fix, this but it seems like the approaches available are:
* expose xdg-documents / home, warn user with popup if they accept on non accessible folder after the fact.
* expose xdg-documents, remove file save dialog, pop-up "file saved to: <xdg-documents>/<unique-name>.<format>

BreadMakesYouFull:
Opted to expose xdg-documents + desktop and log more explicit message in terminal if file save fails

bbhtt:
bot, build io.github.breadmakesyoufull.bpm2fps

PR #5558, opened 2024-09-05
Add org.sil.scripture-app-builder

chrisvire:

Create customized mobile apps with synchronized Scripture text and audio


chrisvire:
We have 4 related but separate "App Builders" that we are switching from Debian packages for Ubuntu to Flatpak.

Scripture App Builder (this one) 
- home: https://software.sil.org/scriptureappbuilder/
- repo: https://github.com/sillsdev/flathub/tree/org.sil.scripture-app-builder
Dictionary App Builder
- home: https://software.sil.org/dictionaryappbuilder/
- repo: https://github.com/sillsdev/flathub/tree/org.sil.dictionary-app-builder
Reading App Builder
- home: https://software.sil.org/readingappbuilder/
- repo: https://github.com/sillsdev/flathub/tree/org.sil.reading-app-builder
Keyboard App Builder
- home: https://software.sil.org/keyboardappbuilder/
- repo: https://github.com/sillsdev/flathub/tree/org.sil.keyboard-app-builder

Should we submit them all at once or work through any issues with this one first and then submit the other 3 later (and make the appropriate fixes to them if anything is found in this one)?

Thanks!
Chris

bbhtt:
How large are these and are they related to one another ? If size is not an issue having them in one Flatpak is better similar to how Libreoffice works.

org.sil.scripture-app-builder.desktop
```diff
@@ -0,0 +1,11 @@
+[Desktop Entry]
+Version=1.0
+Name=Scripture App Builder
+Comment=Build customized Scripture apps for Android smartphones and tablets
+Exec=/app/bin/sab-run.sh %u
```
bbhtt, line 5:
```suggestion
Exec=sab-run.sh %u
```

org.sil.scripture-app-builder.yml
```diff
@@ -0,0 +1,114 @@
...
+  # Or the printer socket
+  # - --socket=cups
+  # Network access
+  - --share=network
+  # XShm access
+  - --share=ipc
+  # All external devices (catch all for devices we interact with and graphical acceleration)
+  - --device=all
+  # Local file browsing access
+  - --filesystem=host
```
bbhtt, line 29: why is this needed?

org.sil.scripture-app-builder.yml
```diff
@@ -0,0 +1,114 @@
...
+  # Temporary file access
+  - --filesystem=/tmp
+  # Filesystem mounts
+  - --filesystem=/mnt
```
bbhtt, line 33:
```suggestion
```

host includes this

org.sil.scripture-app-builder.yml
```diff
@@ -0,0 +1,114 @@
...
+  # Network access
+  - --share=network
+  # XShm access
+  - --share=ipc
+  # All external devices (catch all for devices we interact with and graphical acceleration)
+  - --device=all
+  # Local file browsing access
+  - --filesystem=host
+  # Temporary file access
+  - --filesystem=/tmp
```
bbhtt, line 31: why is /tmp needed?

org.sil.scripture-app-builder.yml
```diff
@@ -0,0 +1,114 @@
...
+  # Filesystem mounts
+  - --filesystem=/mnt
+  # USB Drives
+  - --system-talk-name=org.freedesktop.Udisks2
+  # File saving access
+  - --system-talk-name=org.freedesktop.FileManager1
```
bbhtt, line 37: this does not run on system bus.
```suggestion
  - --talk-name=org.freedesktop.FileManager1
```

org.sil.scripture-app-builder.yml
```diff
@@ -0,0 +1,114 @@
...
+
+# Build args give resource access during the build
+build-options:
+  build-args:
+    - --share=network
+
```
bbhtt, line 43: Network access during build is not allowed on Flathub

org.sil.scripture-app-builder.yml
```diff
@@ -0,0 +1,114 @@
...
+- sab-node20.yml
+
+- openjdk17.yml
+
+- name: scripture-app-builder
+  buildsystem: simple
+  build-commands:
+    - pwd
+    # Source tarball from a docker image and extract to flatpak app
+    - cd /tmp
```
bbhtt, line 67:
```suggestion
```

python3-aeneas.yml
```diff
@@ -0,0 +1,99 @@
...
+  # install extra build modules. The latest version of numpy is 2.0.0 which
+  # will require 'meson-python' to build successfully.
+  # A setup error in version 1.17.5 of numpy suggested the "setup.py clean"
+  # function was deprecated for builds. This is why I was testing new versions.
+  # FYI: The last version of numpy released as a .zip instead of a .tar.gz
+  # was 1.22.4: https://pypi.org/project/numpy/1.22.4/#files
+name: python3-aeneas
+buildsystem: simple
+build-options:
+    append-ld-library-path: /app/lib
```
bbhtt, line 14:
```suggestion
```

/app/lib is already in ld path

python3-aeneas.yml
```diff
@@ -0,0 +1,99 @@
...
+build-commands:
+    - echo PWD="${PWD}"  and  FLATPAK_DEST="${FLATPAK_DEST}"
```
bbhtt, line 16:
```suggestion
```

python3-aeneas.yml
```diff
@@ -0,0 +1,99 @@
...
+    - python3 -m pip install --upgrade pip
```
bbhtt, line 17:
```suggestion
```

python3-aeneas.yml
```diff
@@ -0,0 +1,99 @@
...
+    # Ensuring python3 points to the right place if needed
+    #- rm -f /app/bin/python
+    - python3 -m pip list
```
bbhtt, line 20:
```suggestion
```

python3-aeneas.yml
```diff
@@ -0,0 +1,99 @@
...
+    - type: file
+      url: https://files.pythonhosted.org/packages/e5/92/88d421001bb257588df4864ceca24d570e2e822db4f40f48737a78b648b2/aeneas-1.7.3.0.tar.gz
+      sha256: b492d8ba36f12cce69c29c35e7075857a9d036c4743cc8f4689922be980826da
+    
+    # Latest: 2.5
+    - type: file
+      url: https://files.pythonhosted.org/packages/4c/f3/038b302fdfbe3be7da016777069f26ceefe11a681055ea1f7817546508e3/soupsieve-2.5-py3-none-any.whl
+      sha256: eaa337ff55a1579b6549dc679565eac1e3d000563bcb1c8ab0d0fefbc0c2cdc7
+    
+    #OLD
```
bbhtt, line 34: Can you remove the commented out sources?

python3-aeneas.yml
```diff
@@ -0,0 +1,99 @@
...
+    #   url: https://files.pythonhosted.org/packages/9c/d8/909c4089dbe4ade9f9705f143c9f13f065049a9d5e7d34c828aefdd0a97c/beautifulsoup4-4.11.1-py3-none-any.whl
+    #   sha256: 58d5c3d29f5a36ffeb94f02f0d786cd53014cf9b3b3951d42e0080d8a9498d30
+modules:
+    - name: espeak
+      buildsystem: simple
+      build-commands:
+          - |
+              set -xueo pipefail
+              cd src
+              make install
```
bbhtt, line 91: Use `no-autogen: true`, `subdir: src`, drop build-commands and buildsystem: simple

python3-requests.yaml
```diff
@@ -0,0 +1,23 @@
+# Generated with flatpak-pip-generator --runtime=org.freedesktop.Sdk//23.08 requests --yaml
+name: python3-requests
+buildsystem: simple
+build-options:
+    append-ld-library-path: /app/lib
```
bbhtt, line 5:
```suggestion
```

sab-node20.yml
bbhtt, line 1: why not just use the node 20 extension?

org.sil.scripture-app-builder.yml
```diff
@@ -0,0 +1,114 @@
...
+    - python3 /run/build/scripture-app-builder/docker_pull.py ghcr.io/sillsdev/scripture-app-builder:12.1
+    - python3 /run/build/scripture-app-builder/extract_layer.py /app/output
```
bbhtt, line 69: why pull from a docker image? Where is the source of the application ?

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+    "only-arches": ["x86_64"]
+}
```
bbhtt, line 3: If the end result is x86_64 only why are there aarch64 sources around ? Like OpenJFX

docker_pull.py
bbhtt, line 1: You can use it directly from the source instead of putting it here.

```yaml
- type: file
  url: https://raw.githubusercontent.com/NotGlop/docker-drag/2d9880e2e1756d3bc8f265e8a30a662387ab67ad/docker_pull.py
  sha256: 
```

bbhtt:
Desktop file and metainfo file belongs upstream.

org.sil.scripture-app-builder.yml
bbhtt: Fwiw this will not work on Flathub, there is no network access during build.

org.sil.scripture-app-builder.yml
```diff
@@ -0,0 +1,114 @@
...
+    - rm sillsdev_scripture-app-builder.tar
+    - cd /run/build/scripture-app-builder
```
hfiguiere, line 71: this should already be the current directory. If you get out of part of these steps then it's too complicated.

org.sil.scripture-app-builder.yml
```diff
@@ -0,0 +1,114 @@
...
+    - cp -RPn /app/output/fonts /app/output/images /app/output/html /app/output/info /app/output/lib /app/output/docs /app/output/tools /app/output/pwa /app/scripture-app-builder/
+    # JavaFX SDK
+    - unzip -q -j -d sdk sdk.zip '*lib/*'
+      #- mkdir /app/lib
+    - mv sdk /app/lib/sdk
+    # Icons, desktop entry, metainfo and runner script
+    - mkdir -p /app/share/icons/hicolor/128x128/apps
+    - mkdir -p /app/share/icons/hicolor/scalable/apps
+    - mv /app/output/app-icons/scripture-app-builder.png /app/share/icons/hicolor/128x128/apps/org.sil.scripture-app-builder.png
+    - mv /app/output/app-icons/scripture-app-builder.svg /app/share/icons/hicolor/scalable/apps/org.sil.scripture-app-builder.svg
```
hfiguiere, line 85: use `install`. It creates the target directories at the same tieme.

org.sil.scripture-app-builder.yml
```diff
@@ -0,0 +1,114 @@
...
+    - mkdir -p  /app/share/applications
+    - desktop-file-install --dir /app/share/applications org.sil.scripture-app-builder.desktop
+    - install -Dm644 org.sil.scripture-app-builder.metainfo.xml -t /app/share/metainfo/
+    - install -Dm755 sab-run.sh /app/bin/sab-run.sh
+    - rm -rf /app/output/*
```
hfiguiere, line 90: maybe output shouldn't be in `/app`

python3-aeneas.yml
hfiguiere, line 1: this should be in the main manifest since it's not autogenerated.

sab-run.sh
```diff
@@ -0,0 +1,16 @@
...
+# echo $( pwd )
+# echo $( /app/jre/bin/java -version )
+# echo $( /app/jdk17.0.11/bin/javac -version )
+# echo $( python3 --version )
+# echo $( which python3 )
+# echo $( python3 -m aeneas.diagnostics )
+# echo " "
+# echo "$@"
+# echo " "
+exec /app/jre/bin/java --module-path "/app/lib/sdk" --add-modules javafx.web,javafx.fxml,javafx.swing,javafx.media --add-opens=javafx.fxml/javafx.fxml=ALL-UNNAMED -jar /app/scripture-app-builder/bin/scripture-app-builder.jar $@
```
hfiguiere, line 16:
```suggestion
exec /app/jre/bin/java --module-path "/app/lib/sdk" --add-modules javafx.web,javafx.fxml,javafx.swing,javafx.media --add-opens=javafx.fxml/javafx.fxml=ALL-UNNAMED -jar /app/scripture-app-builder/bin/scripture-app-builder.jar "$@"
```

xterm.yml
hfiguiere, line 1: this should be in the main manifest since it's not autogenerated.

hfiguiere: but then I question building `xterm`

xterm.yml
```diff
@@ -0,0 +1,40 @@
...
+          commit: cc378e4f0cc2ab5b66d378050ba4956612a01197
+  - name: libXaw
+    sources: &libXaw-sources
+      - type: archive
+        url: https://www.x.org/archive/individual/lib/libXaw-1.0.16.tar.xz
+        sha256: 731d572b54c708f81e197a6afa8016918e2e06dfd3025e066ca642a5b8c39c8f
+        # OLD
+        # url: https://www.x.org/archive/individual/lib/libXaw-1.0.15.tar.xz
+        # sha256: ab35f70fde9fb02cc71b342f654846a74328b74cb3a0703c02d20eddb212754a
+  - name: libXaw-32bit
```
hfiguiere, line 34: what is this for?

chrisvire:
> How large are these and are they related to one another ? If size is not an issue having them in one Flatpak is better similar to how Libreoffice works.

We proposed that to the lead developer.  For "political" reasons, they should be separate.

python3-aeneas.yml
chrisvire: The definition is "shared" with [org.sil.Bloom](https://github.com/flathub/org.sil.Bloom/) and easier to maintain.

org.sil.scripture-app-builder.yml
chrisvire: The application builds mobile apps based on project files that are in the file system. It also needs to browse to add audio files and text files to the project from different locations in the file system.

After a build completes, it uses adb to install the app to an Android device so it needs to connect to the device through USB.

org.sil.scripture-app-builder.yml
chrisvire: The build process writes files to /tmp.

org.sil.scripture-app-builder.yml
chrisvire: We download a docker image of the pre-built binaries from a private repo. Is this flag not necessary to do that?

org.sil.scripture-app-builder.yml
bbhtt: These permissions have no effect on the build process.

org.sil.scripture-app-builder.yml
chrisvire: What is different?

org.sil.scripture-app-builder.yml
bbhtt: This is clearly mentioned in the requirements page linked in the PR checklist.

You will have to find an alternate building mechanism

python3-aeneas.yml
chrisvire: What is different? Replace all 4 lines with just the 1 line?

This file comes from a app already accepted into flathub:

https://github.com/flathub/org.sil.Bloom/blob/master/python3-aeneas.yml

python3-aeneas.yml
chrisvire: What is different?

python3-aeneas.yml
bbhtt: $PWD is a shell built in and $FLATPAK_DEST is already set. The source you are pointing at is doing it wrong.

echo -ing this is not a necessary part of the build process

python3-aeneas.yml
bbhtt: not a part of any build process.

org.sil.scripture-app-builder.yml
bbhtt: If you want to use prebuilt binaries, you can host them somewhere and directly use those in the source. Not the docker image.

org.sil.scripture-app-builder.yml
chrisvire: Due to relationships with Bible societies, the source code is in a private repo.  We do a Build in GitHub Actions which creates a tarball of the artifacts and imports that into a docker image. Since we have a nonprofit GitHub account, there is no cost for storing this in GHCR.  It is an economical way to store the results of a build to share and is versioned with tags.

[Saving Images and Containers as Tar Files for Sharing](https://dockerlabs.collabnix.com/beginners/saving-images-as-tar/)

Here is the step we use in our GitHub Action:
```
      - name: "Push AppBuilders Tarball to GHCR"
        run: |
          docker import release/output/app-builders.tgz app-builders:latest
          docker tag app-builders:latest ghcr.io/$OWNER/app-builders:latest
          docker tag app-builders:latest ghcr.io/$OWNER/app-builders:${{ steps.version.outputs.SAB }}
          docker push ghcr.io/$OWNER/app-builders:latest
          docker push ghcr.io/$OWNER/app-builders:${{ steps.version.outputs.SAB }}
```

I don't quite understand `there is no network access during build`. How would installing python modules or OpenJFX into the Flatpak image work without network access?

flathub.json
chrisvire: Actually, I just noticed that the OpenJFX with aarch64 is faulty and should be removed.  We are using Java 17 which is required due to some built libraries being shared between the desktop and android compiled app and Java 17 is the correct version to use with the Android Plugin for Gradle.  There are no builds of OpenJFX for Java 17 and aarch64.

docker_pull.py
chrisvire: We were worried about it going away and breaking our build.  It is pretty essential for the build to work.

chrisvire: Looking at the license, it is GPL3, so we should pull it from their repo to use it since our license is MIT?

xterm.yml
chrisvire: Yeah, I don't like it. One of the reasons for creating a Flatpak is so that we can work on multiple distributions/desktop flavors.  One of the output types from our platform is a PWA.  To test it, the user need to run `vite preview` which launches hosts a web server. We want that in a terminal window so that it is obvious that it is a separate running process, they can interact with the vite process, and close it when they are done.

When we were creating a Debian package for Ubuntu, we would just launch the gnome terminal.  That doesn't seem appropriate if we want to run on a wide variety of dsektops like Gnome, KDE, XFCE, etc. 

Do you have a better suggestion?

chrisvire: > this should be in the main manifest since it's not autogenerated.

It seemed better to modularize it.

org.sil.scripture-app-builder.yml
chrisvire: So we can't specify a docker image to download and then extract the files?  It needs to be a zip file hosted somewhere with a sha256?

hfiguiere: docker images are not an archive format.

xterm.yml
hfiguiere: no. it's not modularize. its just complexified. It's hard to maintain.

org.sil.scripture-app-builder.yml
chrisvire: We will be moving the prebuilt binaries to an S3 Bucket and access through url and sha256

org.sil.scripture-app-builder.yml
chrisvire: Will change to tar ball stored in an S3 bucket.

sab-node20.yml
chrisvire: The node 20 extension was building node from source with yarn and taking an extra 20 minutes for a build.  This extract a pre-built node 20.

org.sil.scripture-app-builder.yml
chrisvire: When I switch from docker file to tarball, this should go away.

xterm.yml
chrisvire: I will see if this can be removed.

sab-run.sh
chrisvire: This should not be quoted since it could be a series of command-line arguments.

hfiguiere: no. 

```
@      Expands to the positional parameters, starting from one. [...] When the expansion occurs within double quotes, each parameter expands to a separate word.  That
is, "$@" is equivalent to "$1" "$2" ...
```

org.sil.scripture-app-builder.yml
chrisvire: The application builds android apps from projects located on the user's computer.  Once the android app is built, it connects via usb to the phone to install and run the app.

org.sil.scripture-app-builder.yml
chrisvire: Sorry, when I said the build process, I should have said the mobile app build process which is done during the running of the desktop app.

sab-run.sh
chrisvire: Thanks for the clarification.  I will try that out.

org.sil.scripture-app-builder.desktop
chrisvire: Moved to source repo and fixed.

org.sil.scripture-app-builder.yml
chrisvire: Removed

org.sil.scripture-app-builder.yml
chrisvire: Changed

org.sil.scripture-app-builder.yml
chrisvire: Removed build-args for `--share=network` and using zip file from S3 bucket.

org.sil.scripture-app-builder.yml
chrisvire: Removed

python3-aeneas.yml
chrisvire: Removed

python3-aeneas.yml
chrisvire: Removed

python3-aeneas.yml
chrisvire: Removed

python3-aeneas.yml
chrisvire: Removed

python3-aeneas.yml
chrisvire: Removed

python3-aeneas.yml
chrisvire: Changed to your recommendation

python3-requests.yaml
chrisvire: Removed

org.sil.scripture-app-builder.yml
chrisvire: Removed and using zip from S3 bucket.

flathub.json
chrisvire: Removed incorrect aarch64 reference to just documenation.

docker_pull.py
chrisvire: Removed.

org.sil.scripture-app-builder.yml
chrisvire: Removed

org.sil.scripture-app-builder.yml
chrisvire: Changed to using `install`.

org.sil.scripture-app-builder.yml
chrisvire: Stopped using /app/output

sab-run.sh
chrisvire: Changed.  This worked.  Thanks!

xterm.yml
chrisvire: Removed libXaw-32bit.

chrisvire:
@bbhtt @hfiguiere I believe I have made all the requested changes and hope I have addressed all of your concerns.

sab-node20.yml
bbhtt: > The node 20 extension was building node from source with yarn and taking an extra 20 minutes for a build. This extract a pre-built node 20.

Do you need node available at runtime for this?

If you use the node20 extension there is no need to rebuild node itself. You can copy the contents of that extension into `/app`

It provides two scripts to copy the contents `install.sh, install-sdk.sh`

org.sil.scripture-app-builder.yml
bbhtt: > Sorry, when I said the build process, I should have said the mobile app build process which is done during the running of the desktop app.

Does it need to be `/tmp` from host? You can set the TMPDIR variable to XDG_RUNTIME_DIR or `/var/tmp`. One is tmpfs, the second one is persistent.

org.sil.scripture-app-builder.yml
```diff
@@ -0,0 +1,86 @@
...
+      - x86_64
+      dest-filename: sdk.zip
+      url: https://download2.gluonhq.com/openjfx/17.0.11/openjfx-17.0.11_linux-x64_bin-sdk.zip
+      sha256: f46a1fbea32b83cca6715d74cba7d9c24ce320f1da2daf8ff852f133e9f15674
+    # App Builder prebuilt binaries
+    - type: file
+      only-arches:
+      - x86_64
+      url: https://sil-app-builders-linux-artifacts.s3.amazonaws.com/artifacts/59/scripture-app-builder.tgz
+      sha256: 1689bf31291802614401042f5c263c1c30d10a546982570ada7ac2d6924e77ff
```
bbhtt, line 86: This is unversioned ? It will break whenever the source is rotated because sha256 will not match anymore.

python3-aeneas.yml
```diff
@@ -0,0 +1,49 @@
...
+      url: https://files.pythonhosted.org/packages/b1/fe/e8c672695b37eecc5cbf43e1d0638d88d66ba3a44c4d321c796f4e59167f/beautifulsoup4-4.12.3-py3-none-any.whl
+      sha256: b80878c9f40111313e55da8ba20bdba06d8fa3969fc68304167741bbf9e082ed
+modules:
+    - name: espeak
+      # buildsystem: simple
+      # build-commands:
+      #     - |
+      #         set -xueo pipefail
+      #         cd src
+      #         make install
```
bbhtt, line 39:
```suggestion
```

org.sil.scripture-app-builder.yml
chrisvire: It is versioned by the build number (59).  Every time we do a release, I will update the url and the sha256 code.

python3-aeneas.yml
chrisvire: Sorry!  I commented it out since I wasn't sure it would work. and forgot to delete it.

chrisvire:
@bbhtt I think it is ready this time.

chrisvire:
@bbhtt @hfiguiere I did another pass through the files and cleaned up some comments.  Thanks for reviewing!

sab-node20.yml
chrisvire: We need the full sdk at runtime since we do a build of a Progressive Web App (PWA) from the user's project data + a PWA template (https://github.com/sillsdev/appbuilder-pwa).

Our application is expecting it to be installed in a specific place (and be a specific version) since we can't just use whatever node is on the path (we have had problems in the past).

If we used the extension, I would need to make code changes to the source of the app to look for it in the install location.

Is it ok to do this?

chrisvire:
@bbhtt @hfiguiere Please let me know what is left to be addressed.  Thanks.

sab-node20.yml
chrisvire: I have switched to using the node20 extension.  I assume this is OK:

`ln -s /app/node /app/scripture-app-builder/tools/node`

chrisvire:
@bbhtt @hfiguiere I have removed our download of node20 and using the node20 extension.

What else do I need to do?

Thanks!

org.sil.scripture-app-builder.yml
chrisvire: I did some testing.  Access to /tmp was being used for builds.  I changed the sab-run.sh to have `-Djava.io.tmpdir=/var/tmp` so that the default tmp dir is set to /var/tmp instead of /tmp.

python3-aeneas.yml
chrisvire: I have combined python3-aeneas.xml into the main manifest.

xterm.yml
chrisvire: I have combined xterm.xml and openjdk.xml into the main manifest.

python3-aeneas.yml
chrisvire: The whole file has been removed and the contents moved to the main manifest.  It does not contain the above commented out code.

chrisvire:
@bbhtt @hfiguiere, I have:
* combined all non-autogen manifest entries into the same file
* removed access to /tmp
* changed to use the node20 extension

What else would you like changed?

org.sil.scripture-app-builder.yml
```diff
@@ -0,0 +1,160 @@
...
+      # jdk
+      - "/usr/lib/sdk/openjdk17/installjdk.sh"
+      # exports
+      - "/usr/lib/sdk/openjdk17/enable.sh"
+      # Rename
+      - mkdir /app/jdk17.0.11
+      - mv /app/jdk/* /app/jdk17.0.11/
+      - rm -r /app/jdk
+    finish-args:
+      - "--env=PATH=/app/jre/bin:/usr/bin:$PATH"
```
bbhtt, line 115:
```suggestion
```

This does not work here. Also variable substitution does not work in `finis-args`

You need to pass the entire thing: `--env=PATH=/app/jre/bin:/app/bin:/usr/bin`

sab-node20.yml
bbhtt: > I have switched to using the node20 extension. I assume this is OK:

Yes if it work.

chrisvire: It works for me.  Thanks

org.sil.scripture-app-builder.yml
chrisvire: I have cleaned up openjdk installation.

chrisvire:
@bbhtt @hfiguiere So that you know, the original work on creating the flatpak packaging of our software was an intern project.  The result worked and I didn't completely understand all the parts of flatpak packaging.  So I submitted to get feedback.

Your feedback was invaluable!  Over the past couple of days, I feel I have been able to clean up and reorganize it into a more easily manageable script.

The latest changes:
* Reorganize the modules into layers from least likely to change to most likely to change. It seems like flatpak is similar to docker and this would be better.
* Simplify the openjdk module.
* Break out the openjfx module and use paths that are more relevant (/app/lib/openjfx instead of /app/lib/sdk)
* Simplify the extraction of prebuilt binaries into the proper place from the beginning and install desktop entry and metainfo from that directory.
* Clean up files from extraction that are not needed in that directory.

Please let me know of any additional suggestions that you have.

bbhtt:
bot, build org.sil.scripture-app-builder

bbhtt:
Please test if it's working.

chrisvire:
@bbhtt The test build passes.  Thanks.

bbhtt:
Did you install it and run ?

chrisvire:
I installed and ran it and did the following tests:
- launched app from search (uses desktop settings)
- build Android app (uses JDK and top)
- build and run PWA app ( uses node20 and xterm)
- fine tune timings (uses resources)
- synchronize text/audio (uses python and Aeneas libraries)
- used book viewer (uses openjfx)

chrisvire:
@bbhtt: I reviewed the changes with our intern and we found some more simplifications:
1. the python3-requests.yaml file was only needed for the python scripts that extracted the docker file so those are not needed anymore.
2. the run shell script is only one line long so I saw (when looking that the node20 manifest) that it could be included in the manifest.

Do we need to do another test build?

bbhtt:
You can start them if you want

bbhtt:
bot, build org.sil.scripture-app-builder

chrisvire:
I tested the build and it was successful.  Thanks!

chrisvire:
@bbhtt Is there anything else to resolve? I am curious about the timetable for the next step.  Thanks.

bbhtt:
No from you, the process is explained in https://docs.flathub.org/docs/for-app-authors/submission#faq

PR #5583, opened 2024-09-12
Add com.tibia.Tibia

ranisalt:

Tibia is a classic MMORPG launched in 1997 by CipSoft. Set in a fantasy world, players create characters, explore vast landscapes, complete quests, battle monsters, and interact with other players. Tibia is known for its old-school graphics, open-world environment, and high level of difficulty, offering players the freedom to explore and engage in both PvE and PvP activities. It's one of the longest-running online games with a loyal community.

The contact forms for the developer do not have public registry. I tried to contact them via their website contact forms, to which I got no answer. I have also emailed them a detailed information about Flatpak, Flathub and how it would be beneficial for distribution of their software, and got a request to submit the same information via other forms to which I didn't get responses:

<details>
Dear Customer Support Team (and hopefully technical development team too!)

Firstly, I want to express my gratitude for providing a Linux client for Tibia, despite it not being officially supported. Many game companies tend to overlook the Linux platform, and your acknowledgment of Linux users is commendable.

I avoided posting on the proposal forums because they mainly focus on in-game content proposals, and discussions related to Linux support often lead to flaming and unproductive debates among users. To ensure a more direct and focused presentation, I chose to address the Flatpak distribution proposal privately to Cipsoft.

I would like to propose allowing the Tibia community to distribute the game via the Flathub repository for Flatpak (https://flathub.org/), a standardized software distribution system for Linux. This initiative offers numerous benefits for both Cipsoft and the players, without requiring a significant investment of resources from the company.

**Benefits for Cipsoft:**

1. Wider Linux Reach: By embracing Flatpak, Tibia will become accessible to a broader audience of Linux users, potentially expanding the game's player base - while the Linux audience is nowadays small, it keeps growing consistently after the launch of the Steam Deck.

2. Community Engagement: Supporting Linux through an official Flatpak version demonstrates Cipsoft's commitment to the Linux community, fostering goodwill and stronger player engagement. Tibia has a longstanding good relationship with players that develop tooling, such as the in-game analyzers and the TibiaMaps fansite.

3. Minimal Effort Required: The creation and maintenance of the Flatpak package would primarily be handled by the community, requiring a minimum (if not zero) amount of work from Cipsoft. A supported Flatpak version provides Cipsoft with better control over the distribution process, ensuring a consistent and optimized gaming experience on Linux.

4. Simplified Testing: With Flatpak, Cipsoft would only need to test Tibia on Linux against a single runtime. This reduces the complexity of testing across various Linux distributions and their package versions, and we know the ecosystem is very fragmented.

5. Security and Stability: Flatpak's sandboxing feature enhances security, reducing the risk of potential security breaches, and contributes to a more stable gaming environment.

**Benefits for Players:**

1. Simplified Installation: Flatpak eliminates dependency concerns, allowing players to install Tibia seamlessly on any Linux distribution with a single click.

2. Cross-Distribution Compatibility: Players can enjoy Tibia on their preferred Linux distribution without worrying about compatibility issues. There is no incompatibility or inconsistency between dependencies.

3. Enhanced Security: Flatpak's sandboxing ensures a secure gaming experience, safeguarding the player's system from potential threats.

4. Consistent Performance: With an encapsulated runtime, Tibia will run consistently on different Linux systems, minimizing performance discrepancies.

In conclusion, embracing Flatpak for distributing Tibia builds acknowledges the Linux community's presence, and demonstrates Cipsoft's support for Linux users. The majority of the work involved in creating and maintaining the Flatpak version will be undertaken by the community, requiring minimal effort from Cipsoft. We believe this collaborative effort will have significant benefits for both Cipsoft and the Tibia community.
</details>

(there's more to the email below)

They do have a bad URL naming where all release have exactly the same filename and there is no mention on the page to what version is currently available. It is possible to find the current launcher version on the following URL: https://static.tibia.com/launcher/launcher-linux-current/package.json.version but this does not change the URL. I have requested that they altered it, with implementation details, in the same email:

<details>
For the proposed Flatpak distribution of Tibia, in its simplest form, it would be necessary to implement a rotating URL system. This entails using a fixed URL that dynamically redirects to different locations, each hosting the respective Tibia version with the version number included in the filename. Such a rotating URL mechanism ensures that users always access the latest version of Tibia without the need to update the Flatpak package continuously.

More specifically, the current download URL

```
https://static.tibia.com/download/tibia.x64.tar.gz
```

Should be a redirect to the latest version, i.e.

```
https://static.tibia.com/download/tibia-13.20.13623.x64.tar.gz
```

I suppose you would still want the file to be downloaded and saved as tibia.x64.tar.gz, and to achieve that, you can send the following header alongside the versioned URL of the file:

```
Content-Disposition: attachment; filename="tibia.x64.tar.gz"
```

If you'd like to know more about and experiment with it, I have a working Flatpak manifest ready that I can share. However, without a rotating URL system, it becomes complex to keep it regularly updated, therefore it is not suitable for submission to Flathub. I invite you to discuss this further and collaborate to enhance the Linux gaming experience for Tibia players. Let's explore this opportunity together. Feel free to reach out for further details and cooperation.
</details>

Because of the fixed URL with changing content, the manifest will fail to build every time there is an update to it, as the checksum will change. flatpak-external-data-checker doesn't help since it expects the URL to change when there's a new version. It would be possible to have a script to work around it, though.

I believe it is very unlikely that the developer will have interest in maintaining it themselves, as they consider the Linux client as "unsupported" (even though it has been working well and they do reply to support requests).

I want to push this to Flathub as the client currently looks awful, with the wrong font and doesn't follow system style, in Arch and possibly other distros, but wrapped in Flatpak it looks almost perfect (there's a longstanding known bug with font alignment):

<details>

**Flatpak**

image

**Native

image
</details>


com.tibia.Tibia.metainfo.xml
```diff
@@ -0,0 +1,44 @@
...
+  <id>com.tibia.Tibia</id>
+
+  <name>Tibia</name>
+  <summary>Become a part of Tibia, the MMORPG classic that has been expanded for more than 25 years</summary>
+
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>LicenseRef-proprietary</project_license>
+
+  <recommends>
+    <display_length compare="ge">1020</display_length>
```
ranisalt, line 12: The game window cannot be resized below 1020x650. Is this the correct way of using this property?

bbhtt:
>  flatpak-external-data-checker doesn't help since it expects the URL to change when there's a new version. It would be possible to have a script to work around it, though.

There is a "rotating-url-checker" that expects a link to the latest version and changes only shasum accordingly. Does that work?

com.tibia.Tibia.metainfo.xml
bbhtt: Yea seems fine.

com.tibia.Tibia.metainfo.xml
```diff
@@ -0,0 +1,44 @@
...
+  <url type="faq">https://www.tibia.com/support/</url>
+  <url type="help">https://www.tibia.com/gameguides/</url>
+
+  <developer id="com.cipsoft">
+    <name>Cipsoft GmbH</name>
+    <url type="homepage">https://www.cipsoft.com/</url>
+  </developer>
+
+  <releases>
+    <release version="5cb4ccd3a17b99a571201e92a8f925f5f19e6c63" date="2024-09-12"/>
```
bbhtt, line 42: How is the version a string ? Or is this a commit ?

com.tibia.Tibia.yml
```diff
@@ -0,0 +1,63 @@
...
+runtime: org.freedesktop.Platform
+runtime-version: '24.08'
+sdk: org.freedesktop.Sdk
+command: tibia-wrapper
+tags:
+  - proprietary
+
+finish-args:
+  - --device=dri # OpenGL rendering
+  - --filesystem=xdg-data/tibia:create
```
bbhtt, line 12: why is xdg-data needed ?

com.tibia.Tibia.yml
```diff
@@ -0,0 +1,63 @@
...
+    build-commands:
+      - install -Dm644 -t ${FLATPAK_DEST}/share/applications com.tibia.Tibia.desktop
+      - install -Dm644 -t ${FLATPAK_DEST}/share/metainfo ${FLATPAK_ID}.metainfo.xml
+
+    sources:
+      - type: file
+        path: com.tibia.Tibia.metainfo.xml
+
+      - type: file
+        path: com.tibia.Tibia.desktop
```
bbhtt, line 63: Merge this with the main module.

modules/icoutils.json
bbhtt, line 1: Add this to the main manifest

com.tibia.Tibia.metainfo.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.tibia.Tibia</id>
+
+  <name>Tibia</name>
+  <summary>Become a part of Tibia, the MMORPG classic that has been expanded for more than 25 years</summary>
+
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>LicenseRef-proprietary</project_license>
```
bbhtt, line 9: Does this allow redistribution ? If not you have to use extra-data https://docs.flatpak.org/en/latest/module-sources.html#extra-data

com.tibia.Tibia.metainfo.xml
```diff
@@ -0,0 +1,44 @@
...
+    </p>
+    <p>
+      For more than 25 years now, players have been visiting the medieval world of Tibia. At present, more than 500,000 players from all over the world inhabit the Tibian continent enjoying the numerous game features.
+    </p>
+    <p>
+      Acting as knights, paladins, sorcerers or druids, players are faced with the challenge of developing the skills of their selected characters, exploring a large variety of areas and dangerous dungeons and interacting with other players on a social and diplomatic level. Besides the sophisticated chat tools it is especially the unique freedom players enjoy in Tibia that create an enormously immersive gaming experience.
+    </p>
+    <p>
+      NOTE: This wrapper is not verified by, affiliated with, or supported by Cipsoft GmbH.
+    </p>
```
bbhtt, line 27: Put it as the first para.

com.tibia.Tibia.desktop
bbhtt, line 1: There is no desktop file inside the tarball ?

com.tibia.Tibia.metainfo.xml
ranisalt: I don't know how their versioning works, it is not sequential and is available in the link I added in the description. Probably an internal repository commit or checksum.

They update the launcher every once in a while but there are no public version numbers. They don't separate launcher and client update news.

This is the _client_ version (which is downloaded by the launcher, this package: https://static.tibia.com/launcher/tibiaclient-linux-current/package.json.version)

Sorry if this is confusing, I don't get it either 😆

com.tibia.Tibia.yml
ranisalt: The game stores configuration in this folder. Should I comment/document it?

com.tibia.Tibia.metainfo.xml
ranisalt: Unclear, it's not mentioned anywhere on the website that redistribution is not allowed, and many websites redistributed files over the years, especially before the launcher.

If it is valuable, I can try to contact the company again about redistribution policy.

com.tibia.Tibia.metainfo.xml
ranisalt: OK, but the generator is using it in a different order.

com.tibia.Tibia.desktop
ranisalt: No. They instruct the user to run a script present in the tarball instead: https://www.tibia.com/support/?subtopic=gethelp&entryid=121

ranisalt:
> There is a "rotating-url-checker" that expects a link to the latest version and changes only shasum accordingly. Does that work?

It does not work because the rotating URL checker expects the URL to... rotate 😆 at least change with updates. The URL for this game will always be the same.

com.tibia.Tibia.yml
bbhtt: XDG_DATA_HOME inside Flatpak is `~/.var/app/$FLATPAK_ID/data`. Why does it need host's xdg-data which is usually `~/.local/share` ?

com.tibia.Tibia.yml
ranisalt: All of this metadata module, or just the desktop file?

com.tibia.Tibia.yml
ranisalt: Indeed it respects `XDG_DATA_HOME`, will remove it

com.tibia.Tibia.yml
bbhtt: this whole module, metainfo + desktop file

com.tibia.Tibia.metainfo.xml
bbhtt: If they don't explicitly say it's allowed, it means it's not and you have to use extra-data.

If you get permission from them then it's fine. You can contact them.

ranisalt: I will try to get their permission then! Will return in a few days since they don't reply on weekends :wink:

guihkx: If you're not able to get permission for distribution from them, I'm pretty sure it'll be impossible to use the official tarball as a source, because its hash will always be changing.

However, given that the game seems to have a built-in auto-update feature, you could probably archive the link of the tarball to [web.archive.org](https://web.archive.org) and use that as a source instead...

guihkx:
bot, build com.tibia.Tibia

com.tibia.Tibia.metainfo.xml
bbhtt: That won't do... it's still the official tarball but with a different non official link.

guihkx: I understand, but why does Flathub take issue with that? For all intents and purposes, it's the official program, just from a different (and trustworthy) mirror...

bbhtt: Because you will sill be distributing their binaries/sources now from a different link. changing the hosting place does not mean their terms and conditions over it disappear.

guihkx: I think I wasn't clear in my previous post. I only suggested archiving the tarball if @ranisalt didn't get permission to redistribute from them, i.e. his only option is to use `extra-data`.

If he gets permission to distribute, this checksum issue won't be a major problem unless Flathub requires reproducible builds.

bbhtt: If they don't have permission to distribute the original link can be used with extra-data. Why does it need archiving ?

We will allow automerge in such cases so that the Flatpak remains installable.

bbhtt: Ok I missed the reply about x-checker below. I see what you meant now.

guihkx: >Why does it need archiving ?

Because unless Flathub is able to keep the tarball checksum in-sync with upstream (down to the second), some unlucky Flathub users will inevitably have issue while installing the game simply because the hash of the tarball doesn't match the one expected in `extra-data`...

bbhtt:
> It does not work because the rotating URL checker expects the URL to... rotate 😆 at least change with updates. The URL for this game will always be the same.

No it doesn't, It is supposed to work with just:

```
        x-checker-data:
          type: rotating-url
          url: https://static.tibia.com/download/tibia.x64.tar.gz
```

You can see it in action here https://github.com/flathub/com.hamrick.VueScan/blob/master/com.hamrick.VueScan.yaml#L68-L70

But the issue is `static.tibia.com` is blocking connections from x-checker probably due to cloudflare. It works on a browser so my guess is that it does some user agent filtering.

ranisalt:
I got a statement from the company behind the game after asking for permission to redistribute:


> Hello,
> 
> Thank you for reaching out.
> 
> Please understand that redistributing the Tibia launcher tarball or its files in any form, including packaging it for a Linux distribution, is not permitted. CipSoft does not allow redistribution of its software. This policy helps us maintain control over the distribution and integrity of our products. I hope you understand.
> 
> Kind regards,
> Manina
> Tibia Customer Support

So I changed it to use extra-data instead.

> But the issue is `static.tibia.com` is blocking connections from x-checker with 403 Forbidden probably due to cloudflare or some firewall. It works on a browser so my guess is that it does some user agent filtering.

It is blocking requests that do not accept compressed responses. It used to work but was disabled in flathub-infra/flatpak-external-data-checker#230

I commented on the original issue back in May flathub-infra/flatpak-external-data-checker#159 but I'm waiting for a response, I'd fix it myself but I lack the knowledge

com.tibia.Tibia.metainfo.xml
ranisalt: Considering f-e-d-c is currently unable to handle the requirements to fetch the file (see below), I'm inclined to use an archive of the file. That would avoid the checksum sync issue, and it will auto-update anyway, one does not need the most up to date launcher at all times.

Is there any facility in place to trigger Internet Archive to redownload a file every once in a while?

bbhtt:
It seems they are against any kind of packaging and the way their binaries are hosted seems very anti packager. I don't like the idea of mirroring their binaries to some other site, it'd still be in their right to take down that mirror and make the Flatpak defunct.

guihkx:
>It seems they are against any kind of packaging and the way their binaries are hosted seems very anti packager

I agree with that. In the past, they did distribute versioned packages of the game (there was no launcher, too). It's sad they decided to change that.

>it'd still be in their right to take down that mirror

True, but...

>make the Flatpak defunct.

I fail to see how that'd be feasible unless they threatened Flathub legally, or something like that.

And I highly doubt they'd do that. Even if they did, I suspect such action wouldn't even have legal merits, since we wouldn't be distributing their binaries...

Ultimately, this is not my decision, of course. Just sharing my 2¢.

bbhtt:
Extra data means the user downloads it on their machine, if the archive.org link is gone through a DMCA takedown the Flatpak is immediately not installable.

guihkx:
Yes, but if that happens (which, again, i highly doubt it), we'd just use another mirror. Ad infinitum.

bbhtt:
And also if they are truly against distributing or packaging it in any form on Flathub, they can still ask us to take down.

Ultimately the distribution story here is flimsy. It's not very useful to have an app published then it becomes defunct after a year and we have to EOL it.

bbhtt:
> FWIW I fixed this in flatpak-builder a while ago https://github.com/flatpak/flatpak-builder/pull/556

As a side note that fix caused a lot of problems on Flathub. Your assumption that webserver will simply ignore empty accept-encoding matches the spec but not the real world. They simply error out and cause the download to fail.

A new `disable-http-decompression` option had to be introduced to mitigate the fallout from that, which is now widely used.

guihkx:
@ranisalt Maybe you could send them another email laying out some of the benefits of having their game on Flathub, and luckily they could reconsider it...

Off the top of my head:

- Potential increase of their player base due to Flathub's popularity on Linux
- Decrease of distro-specific issues (e.g. the font issue you mentioned)
- Complete control of the release process (if they want to)

com.tibia.Tibia.metainfo.xml
guihkx: >one does not need the most up to date launcher at all times.

Right. Though I think you'd need a wrapper launcher script that would copy all launcher files from `/app/bin` to a place where they can be overwritten (`$XDG_DATA_HOME` is a common choice).

Similarly to [what JDownloader does](https://github.com/flathub/org.jdownloader.JDownloader/blob/master/jd-wrapper.sh).

>Is there any facility in place to trigger Internet Archive to redownload a file every once in a while?

I'm not aware of such tool. If I were to maintain this package, I'd use [nvchecker](https://github.com/lilydjwg/nvchecker) to periodically monitor `https://static.tibia.com/launcher/launcher-linux-current/package.json` for changes on the `version` key. Once that key changed, I'd just manually re-archive their `.tar.gz` on [web.archive.org](https://web.archive.org/).

bbhtt:
You could ask them to provide versioned download links that aren't removed. Not sure if they would, but that'd be the most reliable option.

ranisalt:
I asked the company to support Flathub or make changes to the URL, and this was the last reply:

> Hello,
> 
> thank you for your reply, and for the proposal. As already pointed out, we currently have no plans at present to extend availability of any of our clients onto other platforms, but who knows ... thinks may change with time. At any rate, we will keep your feedback in mind.
> 
> In the meantime, why not post your thoughts on our [Proposals board](https://www.tibia.com/forum/?action=board&boardid=10) as well? You can discuss your ideas with other players there, and it would also be a good way to see if the idea resonates with the community.

It seems to me that they would not seek after a Flathub package or any other repackaging (it has been [available on Arch's AUR](https://aur.archlinux.org/packages/tibia) for 9 years) but would not be interested in doing it themselves unless it has enough traction.

I'm confident the community would adopt it given the multiple known issues that occur when running the game natively (lack of fonts, rendering offset issues, sluggish mouse movement, etc) that don't happen in Flatpak.

Remaining issues are:
- f-e-d-c doesn't support their download URL due to lack of compression support
- when there is an update, it fails to load because it expects to modify something in the folder where the executable lives

Update issue could be solved by copying to a mutable folder before launching. It doesn't happen very often, average once a month https://release-monitoring.org/project/371491/ (note this month they updated Qt5 to Qt6 and there are more bugs than usual)

bbhtt:
Are they ok with mirroring their tarballs over to archive.org or any other place ?

guihkx:
If you haven't done the above yet, you should probably explain to them why that would be needed in the first place, and make it very clear that no file is going to be modified.

ranisalt:
I have asked them about sourcing from internet archives, and at the same time created a proposal on their official forums to gather community support: https://www.tibia.com/forum/?action=thread&postid=39483041

@guihkx I have the impression you are also a Tibia player, and I'd enjoy your feedback and support on the official forums :wink:

guihkx:
I used to play Tibia a lot in my younger days, but only on Open Tibia servers...

I actually made an account on official servers last year just to see how much the game has changed, but I tried to log in now and apparently CipSoft deleted it. :S

ranisalt:
Well...

> I am afraid we oppose the distribution of our launcher on any other platform save our own.  I regret that we cannot be more forthcoming - we know you mean no harm - but I'm afraid we are really quite conservative that way. I hope you understand.

So, since we need to download the launcher from tibia.com, I think the extra-data approach + copying to xdg-run so that the launcher can self-update without the Flatpak itself updating (for the time window after a new launcher is available but the Flatpak hasn't been updated) is the way to go.

guihkx:
That's really unfortunate. 

I might be speaking out of my ass here, but I don't think Flathub admins/collaborators will agree to publish the game after that statement, even if we use `extra-data` (especially since that won't be 100% reliable, unless we archive the launcher on [web.archive.org](https://web.archive.org/), which is an idea that's been shot down already).

If the above is true, I think the last option would be hosting your own Flatpak repository here on GitHub, something like [Sober](https://github.com/vinegarhq/sober) (a popular Roblox app) [does](https://github.com/vinegarhq/sober/tree/gh-pages).

PR #5597, opened 2024-09-15
Add org.trelby.Trelby

priframr:

Trelby is a libre wxpython application for creation of screenplays. It can create, edit, and export screenplays in proper formatting for creation of television or film scripts.

- [X] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about this submission. [**Link:**](https://github.com/trelby/trelby/issues/26) (I am making the PR first before asking. I will also ask if someone from there can be the official maintainer instead of me)


com.trelby.Trelby.desktop
```diff
@@ -0,0 +1,6 @@
+[Desktop Entry]
+Type=Application
+Name=trelby
+Exec=/app/bin/trelby
```
hfiguiere, line 4:
```suggestion
Exec=trelby
```

com.trelby.Trelby.desktop
hfiguiere, line 1: this file should be (submitted) upstream

com.trelby.Trelby.metainfo.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id>com.trelby.Trelby</id>
```
hfiguiere, line 3: this doesn't match the homepage that is in `.org`

priframr:
One issue I found with local build is that it cannot find default PDF reader which is /usr/bin/xpdf. How should I resolve that?

com.trelby.Trelby.png
hfiguiere, line 1: icon should also be upstream.

hfiguiere: like from here  https://github.com/trelby/trelby/tree/master/resources

com.trelby.Trelby.yaml
```diff
@@ -0,0 +1,33 @@
+app-id: com.trelby.Trelby
+runtime: org.freedesktop.Platform
+runtime-version: '24.08'
+sdk: org.freedesktop.Sdk
+command: trelby
+finish-args:
+  - --device=all
```
hfiguiere, line 7: why do you need ALL devices?

priframr: I'll remove that, sorry, I wasn't really sure how liberal I should be with permissions.

com.trelby.Trelby.desktop
hfiguiere: it already exist and should be used

https://github.com/trelby/trelby/blob/master/trelby.desktop

hfiguiere:
`xpdf` is kinda deprecated as a PDF reader. It should just `xdg-open` the PDF and this will open is the user chosen PDF reader.

priframr:
> `xpdf` is kinda deprecated as a PDF reader. It should just `xdg-open` the PDF and this will open is the user chosen PDF reader.

This was not my choice, it was the default set. I could just sed it out.

hfiguiere:
1. use patches no sed
2. submit a change upstream that allow to configure it. Using `xpdf` is not a good default.

priframr:
> 1. use patches no sed
> 
>     2. submit a change upstream that allow to configure it. Using `xpdf` is not a good default.

Took me considerable amount of trial and error, but I managed to get xdg-open up and running and it works. So far so good.

priframr:
bot, build

priframr:
bot, build com.trelby.Trelby

Fixed-For-Flatpak.patch
```diff
@@ -0,0 +1,32 @@
+diff --git a/src/misc.py b/src/misc.py
+index 937d078..9d185ca 100644
+--- a/src/misc.py
++++ b/src/misc.py
+@@ -53,7 +53,7 @@ def init(doWX = True):
+             progPath = str(
+                 os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+ 
+-            confPath = str(os.environ["HOME"]) + "/.trelby"
++            confPath = str(os.environ["HOME"]) + "/.var/app/com.trelby.Trelby/config/trelby"
```
hfiguiere, line 10: no. this should be `XDG_CONFIG_HOME`. Not this hardcoded path.

Fixed-For-Flatpak.patch
```diff
@@ -0,0 +1,32 @@
...
+     elif misc.isUnix:
+         progs = [
+-            ("/usr/local/Adobe/Acrobat7.0/bin/acroread", "-tempFile"),
+-            ("acroread", "-tempFile"),
+-            ("xpdf", ""),
+-            ("evince", ""),
+-            ("gpdf", ""),
+-            ("kpdf", ""),
+-            ("okular", ""),
++            ("/app/bin/xdg-open", ""),
```
hfiguiere, line 29:
```suggestion
+            ("xdg-open", ""),
```

com.trelby.Trelby.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - --socket=cups
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --talk-name=org.freedesktop.FileManager1
+  - --filesystem=home
+  - --filesystem=/tmp
+
+modules:
+  - name: flatpak-xdg-utils
```
hfiguiere, line 16: why do you need this?

limburgher:
Hi! I'm one of the upstream Trelby developers, and should be the primary maintainer of this flatpak.

com.trelby.Trelby.yaml
priframr: Because I couldn't at the time figure out how to get xdg-open to work, I think I'll try again and remove that. Sorry.

Fixed-For-Flatpak.patch
hfiguiere: there is `xdg-open` in the runtime. Where it is expected.

com.trelby.Trelby.yaml
hfiguiere: it is there in the runtime. `xdg-open` just work

priframr:
Do you still want me to change com.trelby.Trelby to org.trelby.Trelby?

hfiguiere:
> Do you still want me to change com.trelby.Trelby to org.trelby.Trelby?

What don't you understand with the question you acknowledged:

> The domain used for the application ID is [controlled by the application developers](https://docs.flathub.org/docs/for-app-authors/requirements/#control-over-domain-or-repository) and the [application id guidelines](https://docs.flathub.org/docs/for-app-authors/requirements#application-id) are followed.

priframr:
@hfiguiere 
Okay, what do you think so far? Better?

com.trelby.Trelby.desktop
hfiguiere: this was not addressed. There is still a desktop file here.

com.trelby.Trelby.png
hfiguiere: same thing here. this wasn't addressed.

hfiguiere:
bot, build org.trelby.Trelby

priframr:
So I think we're good, this can be merged? And @limburgher can be given maintainers permissions?

priframr:
@hfiguiere @bbhtt
Whichever one of you can approve the merge request, I think it's good to go unless you have any last minute requests.

hfiguiere:
maybe, before, you answer all the questions.

org.trelby.Trelby.yaml
```diff
@@ -0,0 +1,40 @@
...
+sdk: org.freedesktop.Sdk
+command: trelby
+finish-args:
+  - --socket=cups
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --talk-name=org.freedesktop.FileManager1
+  - --filesystem=home
+  - --filesystem=/tmp
```
bbhtt, line 13: Why is `/tmp` needed?

pip-requirements.json
```diff
@@ -0,0 +1,93 @@
...
+                },
+                {
+                    "type": "file",
+                    "url": "https://files.pythonhosted.org/packages/08/69/1fadb986011de190d0a7c7df404e471b5ccd0d3d3df4b9bf095c29c22d21/reportlab-4.2.2-py3-none-any.whl",
+                    "sha256": "927616931637e2f13e2ee3b3b6316d7a07803170e258621cff7d138bde17fbb5"
+                }
+            ]
+        },
+        {
+            "name": "python3-pytest",
```
bbhtt, line 64: Is pytest a runtime dependency here ?

org.trelby.Trelby.yaml
priframr: Because when you click file > print via PDF, it stores a temporary copy under /tmp, opens your PDF reader for viewing and then you have the choice to do what you wish with that open file.

com.trelby.Trelby.desktop
priframr: Completed

com.trelby.Trelby.png
priframr: Finished

Fixed-Desktop-File.patch
hfiguiere, line 1: use `desktop-file-edit` in post-install.

priframr: Completed

org.trelby.Trelby.yaml
bbhtt: is that all handled in this app or you mean a different app for viewing ?

priframr: > is that all handled in this app or you mean a different app for viewing ?

If my understanding of the source code is correct, it doesn't print to the printer like GTK provides. It instead saves the PDF as a new file, puts it in /tmp, then then requires a PDF viewer to open said file. That is why there is a setting for selecting your PDF reader. It requires another app, so I used xdg-open to let the user select a PDF reader. Once it's open, you can then choose what you wish to do such as skim through it or use the built in printing options in the PDF viewer to do the actual printing be it to a file outside of tmp so it is saved permanent or straight to the printer itself as the user gets to decide.

bbhtt: /tmp is a bit weird location to save. Why not save in `~/Documents` ? Why does it need to be temporary if the user needs to use different apps to view it

bbhtt: Yes something other than `/tmp` would be better as this permission leaks whatever is in host's `/tmp`. You can use `~/.var/cache/$FLATPAK_ID/tmp` by setting `--env=TMPDIR=/var/tmp`, the only issue is that, it's not tmpfs like host's `/tmp`

priframr: > the only issue is that, it's not tmpfs like host's `/tmp`
What kind of consequences or side effects would happen as a result?

bbhtt: tmpfs means things will be cleaned up on restart. Nowadays most distros mount `/tmp` as tmpfs. Whereas that location is persistent storage. There's no cleanup on restart.

priframr: > tmpfs means things will be cleaned up on restart. Nowadays most distros mount `/tmp` as tmpfs. Whereas that location is persistent storage. There's no cleanup on restart.

Going through the python source code, it uses the built in python3 temporary file creation functions, so it was always intended to be a temporary file with print menu option. I changed the tmp to use from /tmp (host tmp) to /var/tmp and the python program explicitly deletes it upon closing trelby. Any other changes I should make?

bbhtt:
bot, build org.trelby.Trelby

priframr:
However, if I don't set filesystem for /var/tmp, it doesn't popup the program selection dialogue.

bbhtt:
That sounds like something else. If you add `--filesystem=/var/tmp` it will mount host's /var/tmp inside sandbox's /var/tmp. Without it, it's an alias to `~/.var/app/$FLATPAK_ID/tmp`

having `--filesystem=/var/tmp` is equivalent to having `--filesystem=/tmp` because in both cases you are trying to access host's paths.

bbhtt:
No that has the same effect as setting it via TMPDIR. Can you try setting it to `$XDG_CACHE_HOME/tmp` via a wrapper ?

priframr:
> No that has the same effect as setting it via TMPDIR. Can you try setting it to `$XDG_CACHE_HOME/tmp` via a wrapper ?

Got it running by setting my wrapper to set TMPDIR to the location you asked for.

priframr:
@bbhtt 
Should we run tests again after a few minor updates since the first test run? Or are we good here?

bbhtt:
Doesn't seem to run here? Fedora 40 VM wayland session.

```
/app/opt/trelby/src/screenplay.py:415: SyntaxWarning: invalid escape sequence '\-'
  pattern = re.compile("#([a-zA-Z0-9\-]+) (.*)")
/app/opt/trelby/src/titles.py:143: SyntaxWarning: invalid escape sequence '\q'
  tmp += " \qc"
/app/opt/trelby/src/titles.py:145: SyntaxWarning: invalid escape sequence '\q'
  tmp += " \qr"
Unable to access the X Display, is $DISPLAY set properly?
```

priframr:
> Doesn't seem to run here? Fedora 40 VM wayland session.
> 
> ```
> /app/opt/trelby/src/screenplay.py:415: SyntaxWarning: invalid escape sequence '\-'
>   pattern = re.compile("#([a-zA-Z0-9\-]+) (.*)")
> /app/opt/trelby/src/titles.py:143: SyntaxWarning: invalid escape sequence '\q'
>   tmp += " \qc"
> /app/opt/trelby/src/titles.py:145: SyntaxWarning: invalid escape sequence '\q'
>   tmp += " \qr"
> Unable to access the X Display, is $DISPLAY set properly?
> ```

The first few warnings are normal and to be expected. It works under X11, even in debian 12 VM. I don't know if we or upstream has to set wayland support. Do you have by any chance xwayland perhaps?

priframr:
bot, build org.trelby.Trelby

org.trelby.Trelby.yaml
```diff
@@ -0,0 +1,35 @@
+app-id: org.trelby.Trelby
+runtime: org.freedesktop.Platform
+runtime-version: '24.08'
+sdk: org.freedesktop.Sdk
+command: trelby
+finish-args:
+  - --socket=cups
+  - --share=ipc
+  - --socket=fallback-x11 #Known bug in wxWidgets, it is hardcoded to use X11 no matter what, for now it will only support X or XWayland. https://github.com/wxWidgets/Phoenix/issues/2557
```
hfiguiere, line 9: this should be `x11`

priframr: Sorry about that, I assumed it didn't matter because it worked with either setting be it x11 or x11-fallback

priframr:
bot, build org.trelby.Trelby

priframr:
@bbhtt 
Give it a shot now in wayland and tell me how it works for you.

PR #5613, opened 2024-09-20
Add org.freecad.FreeCAD

adrianinsaval:

This is intended as an update and id change for [org.freecadweb.FreeCAD](https://github.com/flathub/org.freecadweb.FreeCAD), during 1.0 development cycle freecad has obtained the freecad.org domain and switched to it as primary domain https://github.com/FreeCAD/FreeCAD/pull/10125
FreeCAD 1.0 has not been released yet but there is a release candidate so this is should go only to the beta channel for now, my idea is to go through the approval process before release so that at release day I can just publish it directly, is that ok?

I assume the linter will report error `finish-args-flatpak-spawn-access`, this exception has already been [granted for org.freecadweb.FreeCAD](https://github.com/flathub-infra/flatpak-builder-lint/blob/c2462da11e247d7ef6d6cd2916d410a65a3d3a9e/flatpak_builder_lint/staticfiles/exceptions.json#L1351-L1353) so it just has to be updated to allow org.freecad.FreeCAD too, should I file a PR for that?

- [x] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about this submission. **Link:** https://github.com/orgs/FreeCAD/people
and of course you can see this PR is coming from a repo in the FreeCAD org and I'm already a maintainer of the existing org.freecadweb.FreeCAD flatpak


hfiguiere:
> should I file a PR for that?

yes

flathub.json
hfiguiere, line 1: you don't need this if this builds on both arch.

org.freecad.FreeCAD.yaml
```diff
@@ -0,0 +1,801 @@
...
+default-branch: beta
+runtime: org.kde.Platform
+runtime-version: '6.7'
+sdk: org.kde.Sdk
+command: FreeCAD
+desktop-file-name-prefix: "(beta) "
+finish-args:
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
```
hfiguiere, line 11: does this work better than with Qt 5.15? I remember with wayland it was very buggy.

org.freecad.FreeCAD.yaml
```diff
@@ -0,0 +1,801 @@
...
+  - --device=dri
+  - --share=network
+  - --filesystem=host # needed to make file saving work
+  - --filesystem=xdg-run/gvfs # make GnomeVFS accessible
+  - --filesystem=/run/spnav.sock:ro # allow access to spacenavd
+  - --talk-name=org.freedesktop.Flatpak # to launch openscad
```
hfiguiere, line 17: Note to reviewers:

I am the one who implemented this in the original package.

to import openscad file it runs openscad. The wrapper script in there will launch the flatpak of openscad. It has been around for a while.

org.freecad.FreeCAD.yaml
adrianinsaval: better than qt5 but still has some problems, we need people using wayland and reporting the problems though

org.freecad.FreeCAD.yaml
```diff
@@ -0,0 +1,801 @@
...
+  - --env=TMPDIR=/var/tmp
```
adrianinsaval, line 18: most things seem to work but this is giving me trouble, many unit tests fail when this is set but if I manually set it to /home/adrian/.cache/tmp it all works

org.freecad.FreeCAD.yaml
```diff
@@ -0,0 +1,801 @@
...
+            revcount=38642
+            sed -i 's|${PACKAGE_WCREF}|'"$revcount"' (Git)|' src/Build/Version.h.cmake
+            sed -i 's|@PACKAGE_VERSION@|@PACKAGE_VERSION@.'"$revcount"'|' src/XDGData/org.freecad.FreeCAD.metainfo.xml.in
+            sed -i 's|${PACKAGE_WCURL}|https://github.com/FreeCAD/FreeCAD.git|' src/Build/Version.h.cmake
+          - | # fix paths for flatpak environment
+            sed -i 's|/usr|/app|g' cMake/FindOCC.cmake src/Gui/GraphvizView.cpp
+            sitepackagesDir=$(python3 -c 'import sysconfig; print(sysconfig.get_paths()["platlib"])' | sed 's|/usr|/app|')
+            sed -i 's|${python_libs}|'"$sitepackagesDir"'|' src/Ext/freecad/CMakeLists.txt
+          # We need to change the doc icon name so it can be exported too
+          - sed -i "s|application-x-extension-fcstd|${FLATPAK_ID}.application-x-extension-fcstd|" src/XDGData/org.freecad.FreeCAD.xml src/Gui/CMakeLists.txt
```
hfiguiere, line 801: no. please use `rename-mime-icons`.

flathub.json
adrianinsaval: oh I forgot about this, do you have equipment to test aarch64? I ran out of disk space after ifcopenshell compilation so I didn't actually finish that. I'll use an external hardrive and give it another go tomorrow.

org.freecad.FreeCAD.yaml
adrianinsaval: docs said it was preferable to rename the source file but I don't mind, does `rename-mime-icons` go at the top level or in the freecad step? If I use `rename-mime-icons` do I still have to patch the xml file or that takes care of that too?

hfiguiere: like `rename-desktop-file` and `rename-icon`

it will rename the icons and fix the mime file.

org.freecad.FreeCAD.yaml
adrianinsaval: run unit tests with
```
org.freecad.FreeCAD -c --run-test 0
```

hfiguiere:
bot, build org.freecad.FreeCAD

flathub.json
hfiguiere: there will be a build that you can try.

My RPi is off right now. And building was complicated and long.

hfiguiere:
this will fail as it need the exception in.

hfiguiere:
bot, build org.freecad.FreeCAD

hfiguiere:
bot, build org.freecad.FreeCAD

org.freecad.FreeCAD.yaml
adrianinsaval: solved, it was a faulty openscad test that messed up many subsequent tests because it left cwd pointing to an nonexistent directory

adrianinsaval:
some issues on aarch64 but I'm running asahi on an m1, hard to know if the app is to blame or the os

hfiguiere:
> some issues on aarch64 but I'm running asahi on an m1, hard to know if the app is to blame or the os

Like? 

When I tried the aarch64 build on a RaspberryPi, it used to crash on startup, until I fixed the runtime to build "desktop GL".

hfiguiere:
And just to clarify unless some other aarch64, Raspberry Pi also has an actual GL driver. I know Asahi it too. (don't have the hardware)

adrianinsaval:
> When I tried the aarch64 build on a RaspberryPi, it used to crash on startup, until I fixed the runtime to build "desktop GL".

no crash on startup but it did crash after opening and closing some documents, don't know what I exactly I did to make it crash. Supposedly the qt6 runtime has both desktop opengl and opengl es support at the same time

bbhtt:
why does the desktop file say beta ?

org.freecad.FreeCAD.yaml
```diff
@@ -0,0 +1,802 @@
...
+  - '*.a'
+  - '*.la'
+build-options:
+  append-path: /usr/lib/sdk/llvm15/bin
+  append-ld-library-path: /usr/lib/sdk/llvm15/lib
+modules:
+
+  - shared-modules/glu/glu-9.json
+
+  - name: pyside
```
bbhtt, line 40: we have a pyside baseapp now. can that be used ?

org.freecad.FreeCAD.yaml
```diff
@@ -0,0 +1,802 @@
...
+  - name: Cython
+    buildsystem: simple
+    build-commands:
+      - python3 setup.py install --prefix=/app --root=/
+    sources:
+      - type: archive
+        url: https://files.pythonhosted.org/packages/2a/97/8cc3fe7c6de4796921236a64d00ca8a95565772e57f0d3caae68d880b592/Cython-0.29.37.tar.gz
+        sha256: f813d4a6dd94adee5d4ff266191d1d95bf6d4164a4facc535422c021b2504cfb
+
+  - name: numpy
```
bbhtt, line 254: why are these being manually generated ? calling setup.py is deprecated. why not use pip-generator ?

org.freecad.FreeCAD.yaml
```diff
@@ -0,0 +1,802 @@
...
+    buildsystem: simple
+    build-commands:
+      - python3 setup.py build -j 1 install --prefix=/app
```
bbhtt, line 257: why `-j 1` ?

org.freecad.FreeCAD.yaml
adrianinsaval: do they disable py_limited_api in that build?

org.freecad.FreeCAD.yaml
adrianinsaval: @hfiguiere might now, this is coming from org.freecadweb.FreeCAD unchanged, I can try without it

hfiguiere:
> why does the desktop file say beta ?

It's to be merged in beta until final release
It's what FreeCAD has been doing, as well as some other apps (like GIMP) to distinguish beta from stable.

org.freecad.FreeCAD.yaml
adrianinsaval: ignornace I guess, is there any docs about the proper way of doing this? and again this is coming straight from the current flatpak. https://github.com/flathub/org.freecadweb.FreeCAD/blame/105074876d9b26ab0469d6b68589faeeff2fd59a/org.freecadweb.FreeCAD.yaml#L235C11-L246

org.freecad.FreeCAD.yaml
hfiguiere: I believe in might have failed with too many CPU. Unsure exactly if this is fixed or not.

org.freecad.FreeCAD.yaml
bbhtt: Don't think so, the default is to build without limited API.

org.freecad.FreeCAD.yaml
hfiguiere: that's what the original package does. It probably predate all that tooling, the original maintainer went MIA a while back so I just maintained it.

bbhtt: https://github.com/flatpak/flatpak-builder-tools/tree/master/pip#flatpak-pip-generator point it to your requirements.txt or individual modules.

The ones below are generated with pip-generator.

hfiguiere: I use the tool for recent dependencies, but the whole thing is fragile. I'd leave it like that until necessary.

Numpy is a pain to build.

adrianinsaval:
bot, build org.freecad.FreeCAD

hfiguiere:
We could also publish it with the old name on the flathub-beta until this get sorted out. I tried earlier but PySide for 5.15 failed to build.... which is kind of puzzling as it didn't change. Didn't try with 6.7

adrianinsaval:
is this expected to take long?

hfiguiere:
the build takes a few hours, yes.

adrianinsaval:
hours aren't a problem. I don't think we're going to make an official stable release until next week

adrianinsaval:
but if you think it's more appropiate to publish on the old name for now I can make a PR there too

hfiguiere:
oh you meant merge this? On my side, it the build is ok for you it's ok for me so we should be able to get it quick.

adrianinsaval:
I'm not having trouble with the latest aarch64 build except that openscad is very old there (2015) so the pipe transfer mechanism doesn't work. Any particular reason why openscad is so out of date on aarch64?

hfiguiere:
I'm pretty sure this is a very old build until it got disabled because it didn't build (instead of fixing).
https://github.com/flathub/org.openscad.OpenSCAD/commit/ea2e3427cb6a6edae777858434b520b533a3ba98 is the commit that disabled it.

hfiguiere:
It's a question for them.

adrianinsaval:
from a quick test it might be the same qt5 opengl es issue, in any case that's not a blocker for us.

bbhtt:
One of the dependencies became unreachable here since the last test build, https://buildbot.flathub.org/#/builders/6/builds/149175

You have to fix that and trigger a rebuild to get this published.

PR #5614, opened 2024-09-21
Add com.fyralabs.Exponential

lleyton:

A repack of the [Linear](https://linear.app/) desktop app for Linux.

I have contacted the developers through their customer Slack channel. 
https://files.catbox.moe/xonqhz.png


com.fyralabs.Exponential.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
+Type=Application
+Name=Exponential
+Comment=A purpose-built tool for planning and building products
+Icon=com.fyralabs.Exponential
+Exec=/app/bin/com.fyralabs.Exponential %U
```
hfiguiere, line 6: this is never a path.

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "skip-arches": ["aarch64"]
```
hfiguiere, line 2:
```suggestion
  "only-arches": ["x86_64"]
```

com.fyralabs.Exponential.yaml
```diff
@@ -0,0 +1,56 @@
...
+      - install -Dm644 ${FLATPAK_ID}.metainfo.xml /app/share/appdata/${FLATPAK_ID}.metainfo.xml
+      - install -Dm644 ${FLATPAK_ID}.desktop /app/share/applications/${FLATPAK_ID}.desktop
+      - mkdir /app/exponential
+      - unzip electron.zip -d /app/exponential
+      - rm -rf /app/exponential/resources/*
+      - unzip -j -d "/app/exponential/resources" linear.zip 'Linear.app/Contents/Resources/*'
+      - patch-desktop-filename ${FLATPAK_DEST}/exponential/resources/app.asar
+    sources:
+      - type: file
+        url: https://download.todesktop.com/200315glz2793v6/Linear%201.27.0%20-%20Build%202408152uraifa26-x64-mac.zip
```
hfiguiere, line 39: why does the build say "mac"? This is not macOS.

com.fyralabs.Exponential.yaml
```diff
@@ -0,0 +1,56 @@
...
+base-version: "23.08"
+
+command: com.fyralabs.Exponential
+separate-locales: false
+
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  # - --socket=wayland
+  # - --socket=fallback-x11
```
hfiguiere, line 17:
```suggestion
```

hfiguiere:
> A repack of the [Linear](https://linear.app/) desktop app for Linux.

And what is it?

com.fyralabs.Exponential.metainfo.xml
```diff
@@ -0,0 +1,63 @@
...
+<component type="desktop-application">
+  <id>com.fyralabs.Exponential</id>
+  <name>Exponential</name>
+  <developer id="com.fyralabs">
+    <name>Fyra Labs</name>
+  </developer>
+  <summary>A purpose-built tool for planning and building products</summary>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>LicenseRef-proprietary</project_license>
+  <url type="homepage">https://linear.app</url>
```
hfiguiere, line 11: why doesn't this correspond to the app-id?

com.fyralabs.Exponential.metainfo.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.fyralabs.Exponential</id>
+  <name>Exponential</name>
+  <developer id="com.fyralabs">
+    <name>Fyra Labs</name>
+  </developer>
+  <summary>A purpose-built tool for planning and building products</summary>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>LicenseRef-proprietary</project_license>
```
hfiguiere, line 10: since this is proprietary and there is no further information we assume it's not redistributable so 'extra-data' must be used.

com.fyralabs.Exponential.yaml
lleyton: We use the .asar file (which is universal) from the macOS build.

com.fyralabs.Exponential.metainfo.xml
lleyton: The Linear team is comfortable with us redistributing, as long as the app metadata doesn't use the Linear branding. I can use extra-data though.

com.fyralabs.Exponential.metainfo.xml
lleyton: Good point, might be best to just omit it. The desktop app wraps the Linear app, which is why I put it here.

com.fyralabs.Exponential.desktop
lleyton: Shouldn't this install to that path:
https://github.com/flathub/flathub/pull/5614/files#diff-72eb0b8bf04f6d2d4464f1f57a3d341ce93fae36ae7408a2fcaa84b2ccf7355dR26

bbhtt: Exec does not need a path, `/app/bin` is already in $PATH inside the sandbox, it is automatically located.

com.fyralabs.Exponential.yaml
bbhtt: what is the relation of todesktop.com with fyralabs.com ?

com.fyralabs.Exponential.metainfo.xml
TingPing: Unless they have a license written down nobody can verify that.

com.fyralabs.Exponential.yaml
lleyton: the situation is a bit confusing, but I can try to explain:
fyralabs.com is a domain I control, which I'm using for the repack, upstream wanted us to use distinct branding from "Linear", which is why the app is called "Exponential", rather than "Linear"
linear.app is the product that is being repacked
todesktop.com is a service that Linear uses to actually build the desktop application (and bundle), and acts as a CDN for the build artifacts

com.fyralabs.Exponential.metainfo.xml
lleyton: understandable, let me update it

com.fyralabs.Exponential.yaml
bbhtt: So linear.app is the upstream of the main app here ? So you are submitting it under com.fryalabs because they asked you to ?

> upstream wanted us to use distinct branding from "Linear", which is why the app is called "Exponential", rather than "Linear"

Is this simply a rebrand or a fork ?

Can we get some kind of confirmation on that ? If not this should be called `app.linear.foo.bar`

bbhtt: Apps must use the original branding, upstream's app ID. If they want it "unofficial" there are other ways to do that like having a note on the description and the unverified badge.

If you submit it under your domain, it allows someone else to verify someone else's apps

lleyton: > you are submitting it under com.fryalabs because they asked you to

Correct, they seemed to not want any of the app metadata (besides a mention in the description) tied to them.

> Is this simply a rebrand or a fork ?

rebrand

> Apps must use the original branding, upstream's app ID.

I'll ask again, I originally suggested the idea of just having a note in the description, but they didn't think that was sufficient.

hfiguiere: This make the license dubious at best. This is not open source and you want us to believe it's okay based on a screenshot of a random discord conversation.

bbhtt: Marking it as blocked because:

a) we don't know if branding modifications are allowed as it's under a proprietary license

b) the ID does not follow the requirements: https://docs.flathub.org/docs/for-app-authors/requirements#control-over-domain-or-repository

> It must be a domain directly related to the original project and the author/developer/project of the application must have control over the domain portion of the ID

Even if the modification was allowed, the ID must be `app.linear`, you can make the user visible parts like metainfo, desktop file, icon use the new branding.

PR #5629, opened 2024-09-24
Add org.sil.reading-app-builder

chrisvire:

Build apps with synced text and audio


flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+    "only-arches": ["x86_64"]
```
hfiguiere, line 2: disappointing

org.sil.reading-app-builder.yml
```diff
@@ -0,0 +1,158 @@
...
+  # Audio access
+  - --socket=pulseaudio
+  # Network access
+  - --share=network
+  # XShm access
+  - --share=ipc
+  # All external devices (catch all for devices we interact with and graphical acceleration)
+  - --device=all
+  # Local file browsing access
+  - --filesystem=host
```
hfiguiere, line 23: why host?

org.sil.reading-app-builder.yml
```diff
@@ -0,0 +1,158 @@
...
+  # USB Drives
+  - --system-talk-name=org.freedesktop.Udisks2
+  # File saving access
+  - --talk-name=org.freedesktop.FileManager1
+
+# App's run command
+command: rab-run.sh
+
+modules:
+  - name: xterm
```
hfiguiere, line 33: why  do you need an xtern?

org.sil.reading-app-builder.yml
```diff
@@ -0,0 +1,158 @@
...
+    sources:
+      - type: git
+        url: https://github.com/ThomasDickey/xterm-snapshots.git
+        tag: xterm-394
+        commit: 47a95ffdc1bef07ec62836ed887536b5d92166af
+        x-checker-data:
```
hfiguiere, line 39: why do you need this updated on snapshots?

org.sil.reading-app-builder.yml
```diff
@@ -0,0 +1,158 @@
...
+finish-args:
+  # Window x11 socket access
+  - --socket=x11
+  # Audio access
+  - --socket=pulseaudio
+  # Network access
+  - --share=network
+  # XShm access
+  - --share=ipc
+  # All external devices (catch all for devices we interact with and graphical acceleration)
```
hfiguiere, line 20: which devices?

flathub.json
chrisvire: We are dependent on JDK 11 (we can't do a newer version yet) and OpenJFX which only has x86_64 builds for JDK 11.

org.sil.reading-app-builder.yml
chrisvire: This desktop app builds mobile apps based on projects that are stored in the user's file system (which reference additional files stored on the file system). We also use Python 3 which is installed on the platform

flathub.json
hfiguiere: there is an aarchh64 build of jdk 11

org.sil.reading-app-builder.yml
chrisvire: We will be adding a way to build a Progressive Web App (PWA) output from the project data along with an Android app. That is not ready yet. I had removed the Node usage (https://github.com/flathub/flathub/pull/5629/commits/aaf816249051ec8056bf1c9aff6ddc1ed97a52d6) and I could also remove xterm for now. 

The purpose of using xterm is to allow `npm run preview` to run in a separate visible process so that the user can interact with it (get the local ip address / port to access the PWA in preview mode) and close the development web server running.

Should I remove it for now and then add it back in later when the feature is ready?

org.sil.reading-app-builder.yml
chrisvire: When we build an Android app, the user can select to download and run on their connected Android device.

flathub.json
chrisvire: We also have native code embedded in our Java code which doesn't have support for aarch64.  I have tried, but I haven't been able to get the native code included to build for aarch64. I am developing on a MacStudio with Apple Silicon M1 Max. I would love to support aarch64.

org.sil.reading-app-builder.yml
hfiguiere: > We also use Python 3 which is installed on the platform

From inside the sandbox, don't do that. If it works, it's coincidental.

org.sil.reading-app-builder.yml
hfiguiere: if you don't need it it has no reason to be here.

org.sil.reading-app-builder.yml
hfiguiere: that doesn't explain which device.

flathub.json
chrisvire: The above should be JDK 17 and OpenJFX (not JDK 11). My mistake.

https://gluonhq.com/products/javafx/
When selecting JavaFX 17.0.12 [LTS], Linux, and aarch64, the only download link is for Javadoc.

And we still have our own native library code that doesn't work on aarch64. I will schedule time in October to attempt porting that native code to aarch64 again.

org.sil.reading-app-builder.yml
chrisvire: xterm has been removed.

org.sil.reading-app-builder.yml
chrisvire: xterm has been removed

org.sil.reading-app-builder.yml
```diff
@@ -0,0 +1,158 @@
...
+        sources:
+          - type: git
+            url: https://gitlab.freedesktop.org/xorg/lib/libxmu.git
+            commit: cc378e4f0cc2ab5b66d378050ba4956612a01197
+      - name: libXaw
+        sources:
+          - type: archive
+            url: https://www.x.org/archive/individual/lib/libXaw-1.0.16.tar.xz
+            sha256: 731d572b54c708f81e197a6afa8016918e2e06dfd3025e066ca642a5b8c39c8f
+      - name: ncurses
```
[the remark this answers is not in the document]
chrisvire: xterm and dependencies have been removed.

chrisvire:
I have installed the test version and it passes my tests.

PR #5647, opened 2024-09-28
Add com.cinecred.cinecred

LoadingByte:

  Cinecred generates fully customizable film credits from spreadsheets: https://cinecred.com

- [x] I am the author of the project. **Link**: https://github.com/LoadingByte/cinecred


com.cinecred.cinecred.yml
```diff
@@ -0,0 +1,41 @@
...
+      - /share/doc
+      - /share/man
+    sources:
+      - type: archive
+        url: https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs10040/ghostscript-10.04.0.tar.gz
+        sha256: c764dfbb7b13fc71a7a05c634e014f9bb1fb83b899fe39efc0b6c3522a9998b1
+  - name: cinecred
+    buildsystem: simple
+    build-commands:
+      - sed -i "s/Icon=cinecred/Icon=$FLATPAK_ID/" cinecred.desktop
```
hfiguiere, line 28: use `desktop-file-edit` not `sed`.

com.cinecred.cinecred.yml
```diff
@@ -0,0 +1,41 @@
...
+      - sed -i "s/>cinecred/>$FLATPAK_ID/" cinecred.metainfo.xml
```
hfiguiere, line 29: if you install it with its original name and put `rename-appdata-file` in the top-level of the manifest, you don't have to do that obscure sed command. Also install the desktop file with its origianal name and use `rename-desktop-file` and it does the right thing too.

com.cinecred.cinecred.yml
```diff
@@ -0,0 +1,41 @@
...
+      - mkdir -p /app/share/{applications/,metainfo/} /app/share/icons/hicolor/{scalable/apps/,256x256/apps/}
+      - mv cinecred.desktop "/app/share/applications/$FLATPAK_ID.desktop"
+      - mv cinecred.metainfo.xml "/app/share/metainfo/$FLATPAK_ID.metainfo.xml"
+      - mv cinecred.svg "/app/share/icons/hicolor/scalable/apps/$FLATPAK_ID.svg"
+      - mv cinecred.png "/app/share/icons/hicolor/256x256/apps/$FLATPAK_ID.png"
```
hfiguiere, line 34:
```suggestion
      - install -Dm644 cinecred.desktop "/app/share/applications/$FLATPAK_ID.desktop"
      - install -Dm644 cinecred.metainfo.xml "/app/share/metainfo/$FLATPAK_ID.metainfo.xml"
      - install -Dm644 cinecred.svg "/app/share/icons/hicolor/scalable/apps/$FLATPAK_ID.svg"
      - install -Dm644 cinecred.png "/app/share/icons/hicolor/256x256/apps/$FLATPAK_ID.png"
```

nonwithstanding installing them with the original names as mentionned above.

com.cinecred.cinecred.yml
hfiguiere: also installing with the original name and using `rename-icon` will also fix that automatically.

com.cinecred.cinecred.yml
```diff
@@ -0,0 +1,41 @@
...
+      - mv bin/* /app/bin/
+      - mv lib/* /app/lib/
+    sources:
+      - type: archive
+        url: https://cinecred.com/dl/1.7.0/cinecred-1.7.0-linux-x86_64.tar.gz
```
hfiguiere, line 39: is redistribution allowed?

com.cinecred.cinecred.yml
```diff
@@ -0,0 +1,41 @@
...
+runtime: org.freedesktop.Platform
+runtime-version: "24.08"
+sdk: org.freedesktop.Sdk
+command: cinecred
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=x11
+  - --device=dri
+  - --filesystem=host
```
hfiguiere, line 11: this is a very broad filesystem permission.

com.cinecred.cinecred.yml
LoadingByte: I'm the original author and also host the website and downloads, so yes.

hfiguiere: no aarch64 build? why not building from source?

com.cinecred.cinecred.yml
LoadingByte: The full file system access is required because users can drag-and-drop folders into the welcome window to open them, and portals don't support that:

Drag-and-drop demo

com.cinecred.cinecred.yml
LoadingByte: Cinecred is a JVM application with various native components, which have been pre-built and checked into the source tree: https://github.com/LoadingByte/cinecred/tree/master/src/main/natives/linux-x86_64

On Linux, these natives have only been built for x86_64, so compiling the JVM app from source wouldn't suffice to add aarch64 support. At some point in the future, I'll probably check in aarch64 natives, and at that point the Flatpak will also be able to support it.

com.cinecred.cinecred.yml
hfiguiere: the filetransfer portal work with folders.

com.cinecred.cinecred.yml
LoadingByte: I now use the `rename-*` properties and the `install` command; this should also resolve the two previous comments. One `sed` is however still needed to delete the `<icon>` tag in the AppStream XML file, because that tag not automatically renamed by `flatpak-builder`, but instead later amended by the "Composing metadata..." step.

com.cinecred.cinecred.yml
hfiguiere: you mean it can't be rebuilt from source? So much for GPL-3.0.

com.cinecred.cinecred.yml
hfiguiere: no, `sed` is not necessary. and in any case patches should be used as `sed` failure mode is...  a failure.

Ideally you'll just use the naming across the board, it's not flatpak specific.

com.cinecred.cinecred.yml
LoadingByte: They can be rebuilt, and that's documented, but it would make the Flatpak build a lot more complicated: https://github.com/LoadingByte/cinecred/blob/master/MAINTENANCE.md#compiling-native-libraries

com.cinecred.cinecred.yml
LoadingByte: When I build the Flatpak locally without the remaining `sed` command, I get the following error:

```
Renaming cinecred.metainfo.xml to share/metainfo/com.cinecred.cinecred.metainfo.xml
Renaming cinecred.desktop to com.cinecred.cinecred.desktop
Renaming icon hicolor/256x256/apps/cinecred.png to hicolor/256x256/apps/com.cinecred.cinecred.png
Renaming icon hicolor/scalable/apps/cinecred.svg to hicolor/scalable/apps/com.cinecred.cinecred.svg
Rewriting contents of com.cinecred.cinecred.desktop
Running appstreamcli compose
Only accepting components: com.cinecred.cinecred, com.cinecred.cinecred.desktop
Processing directory: .../flathub/.flatpak-builder/rofiles/rofiles-MgpDk2/files
Composing metadata...
Run failed, some data was ignored.
Errors were raised during this compose run:
general
  E: filters-but-no-output

com.cinecred.cinecred
  E: icon-not-found
Refer to the generated issue report data for details on the individual problems.
Error: ERROR: appstreamcli compose failed: Der Kindprozess wurde mit Status 1 beendet
```

The cause is that the upstream `cinecred.metainfo.xml` contains the tag `<icon type="stock">cinecred</icon>`, and `flatpak-builder` doesn't rename it, so deleting it via `sed` is the simplest solution. If this `sed` call failed, the whole Flatpak build would fail, as is evident from the error log above. So in this case, the failure condition of `sed` is to make the whole build fail, which I think is right.

com.cinecred.cinecred.yml
LoadingByte: Oh, I didn't know about that yet. Thanks a lot for letting me know! I'll try adding support for that portal in the next release.

Still, there are places in the UI where no portal solution would work. For example, in the delivery window, the user can select the output file either with a portal file chooser, or by pasting a file path into a text field. In addition, when the user switches the render format, the file extension is automatically updated. In the latter two cases, I think it's impossible to obtain writing permission for the file via portals.

Delivery window

com.cinecred.cinecred.yml
```diff
@@ -0,0 +1,42 @@
...
+      - /share/doc
+      - /share/man
+    sources:
+      - type: archive
+        url: https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs10040/ghostscript-10.04.0.tar.gz
+        sha256: c764dfbb7b13fc71a7a05c634e014f9bb1fb83b899fe39efc0b6c3522a9998b1
+  - name: cinecred
+    buildsystem: simple
+    build-commands:
+      - sed -i "/<icon/d" cinecred.metainfo.xml
```
bbhtt, line 31: use a patch for this.

LoadingByte: I updated the submission to use a patch.

metainfo.patch
```diff
@@ -0,0 +1,7 @@
+diff --git i/cinecred.metainfo.xml w/cinecred.metainfo.xml
+index dbf835f..c75dce5 100644
+--- cinecred.metainfo.xml
++++ cinecred.metainfo.xml
+@@ -11 +11 @@
+-  <icon type="stock">cinecred</icon>
++  <icon type="stock">com.cinecred.cinecred</icon>
```
bbhtt, line 7: this is a desktop-application right ? Then you don't need this tag. Remove it altogether.

bbhtt:
bot, build com.cinecred.cinecred

metainfo.patch
LoadingByte: Done. Thanks for the hint!

Just a quick question: it's only for Flatpak where the icon tag isn't needed since Flatpak adds it by itself during the build process, right? Because the AppStream docs state:

> The `<icon/>` tag describes the component icon. It is mostly used for GUI applications (component-type `desktop-application`).

So to me it looks correct to keep the icon tag in the general upstream metainfo, and remove it only for the Flatpak build (i.e., what we're doing right now :) ).

bbhtt: it's talking about catalogue data not metainfo. It's needed for console apps when they have no desktop file.

For desktop-applications: appstream gets the desktop file using `launchable` -> desktop file points to icon -> icon is found if installed.

LoadingByte: Ah, I see. Then I'll remove the `<icon>` tag upstream in the next release, which will allow me to also get rid of this patch. Thanks!

LoadingByte:
I just tested the server build myself and everything seems to work, including the bundled Ghostscript.

com.cinecred.cinecred.yml
```diff
@@ -0,0 +1,43 @@
+id: com.cinecred.cinecred
+runtime: org.freedesktop.Platform
+runtime-version: "24.08"
+sdk: org.freedesktop.Sdk
+command: cinecred
+finish-args:
+  - --share=ipc
+  - --share=network
```
bbhtt, line 8: why do you need network access ?

LoadingByte: There's optional integration with Google Sheets: https://cinecred.com/guide/project-folder/#creating-a-project

LoadingByte:
Thank you for the quick review!

One final question from my side: Cinecred supports optional playout to [DeckLink](https://www.blackmagicdesign.com/products/decklink) cards, which requires `libDeckLinkAPI.so`. That library is only distributed by Blackmagic and isn't redistributable. To install the library on the host, users need to set up the official [Desktop Video](https://www.blackmagicdesign.com/support/family/capture-and-playback) software. However, since Flatpaks can't use any libraries from the host, Cinecred can't find the library. And due to the licensing, it's impossible to distribute the library with the Flatpak.

It seems that other projects are facing the same issue (e.g., obsproject/obs-studio#6004). Do you know whether there are any plans to add (very rarely granted) exceptions to the isolation principle for cases like this?

bbhtt:
Make sure to change the license tag in appstream and check for compatibility if you are going to add it.

PR #5648, opened 2024-09-28
Add com.jwestall.Forecast

jwestall:
### Please confirm your submission meets all the criteria

- [x] Please describe the application briefly.
  A simple weather application written in rust and libcosmic.

- [X] The domain used for the application ID is [controlled by the application developers][appid-domain] and the [application id guidelines][appid] are followed.
- [X] I have read the and followed all the [Submission and licence requirements][reqs].
- [X] I have [built][build] and tested the submission locally.
- [X] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about this submission.

[appid-domain]: https://docs.flathub.org/docs/for-app-authors/requirements/#control-over-domain-or-repository
[appid]: https://docs.flathub.org/docs/for-app-authors/requirements#application-id
[reqs]: https://docs.flathub.org/docs/for-app-authors/requirements
[build]: https://docs.flathub.org/docs/for-app-authors/submission/#before-submission

PR #5649, opened 2024-09-29
Add dev.edfloreshz.CosmicTweaks

edfloreshz:

 A tweaking tool for the COSMIC desktop. 

### DIFF
--- a/dev.edfloreshz.CosmicTweaks.json
+++ b/dev.edfloreshz.CosmicTweaks.json
@@ -34,7 +34,7 @@
         {
           "type": "git",
           "url": "https://github.com/cosmic-utils/tweaks.git",
-          "commit": "4c73f09e16f34d014e0379b55c21225694f178e3"
+          "commit": "9010f5963f313c17fbbb33bc516f64f13b71cd56"
         },
         "cargo-sources.json"
       ]
--- a/dev.edfloreshz.CosmicTweaks.json
+++ b/dev.edfloreshz.CosmicTweaks.json
@@ -1,0 +1,43 @@
+{
+  "id": "dev.edfloreshz.CosmicTweaks",
+  "runtime": "org.freedesktop.Platform",
+  "runtime-version": "24.08",
+  "sdk": "org.freedesktop.Sdk",
+  "sdk-extensions": ["org.freedesktop.Sdk.Extension.rust-stable"],
+  "command": "cosmic-ext-tweaks",
+  "finish-args": [
+    "--share=ipc",
+    "--socket=fallback-x11",
+    "--socket=wayland",
+    "--device=dri",
+    "--talk-name=com.system76.CosmicSettingsDaemon",
+    "--filesystem=xdg-config/cosmic:rw"
+  ],
+  "build-options": {
+    "append-path": "/usr/lib/sdk/rust-stable/bin",
+    "env": {
+      "CARGO_HOME": "/run/build/tweaks/cargo"
+    }
+  },
+  "modules": [
+    {
+      "name": "tweaks",
+      "buildsystem": "simple",
+      "build-commands": [
+        "cargo --offline build --release --verbose",
+        "install -Dm0755 ./target/release/cosmic-ext-tweaks -t /app/bin/",
+        "install -Dm644 ./res/app_icon.svg /app/share/icons/hicolor/scalable/apps/dev.edfloreshz.CosmicTweaks.svg",
+        "install -Dm644 ./res/metainfo.xml /app/share/metainfo/dev.edfloreshz.CosmicTweaks.metainfo.xml",
+        "install -Dm644 ./res/desktop_entry.desktop /app/share/applications/dev.edfloreshz.CosmicTweaks.desktop"
+      ],
+      "sources": [
+        {
+          "type": "git",
+          "url": "https://github.com/cosmic-utils/tweaks.git",
+          "commit": "4c73f09e16f34d014e0379b55c21225694f178e3"
+        },
+        "cargo-sources.json"
+      ]
+    }
+  ]
+}
--- a/dev.edfloreshz.CosmicTweaks.json
+++ b/dev.edfloreshz.CosmicTweaks.json
@@ -34,7 +34,7 @@
         {
           "type": "git",
           "url": "https://github.com/cosmic-utils/tweaks.git",
-          "commit": "7924c71399a1376b7a88325801edf12446a06527"
+          "commit": "37e3e3b9ca229e83a2cc699e2c1271e529cb0d1a"
         },
         "cargo-sources.json"
       ]
--- a/dev.edfloreshz.CosmicTweaks.json
+++ b/dev.edfloreshz.CosmicTweaks.json
@@ -34,7 +34,7 @@
         {
           "type": "git",
           "url": "https://github.com/cosmic-utils/tweaks.git",
-          "commit": "9010f5963f313c17fbbb33bc516f64f13b71cd56"
+          "commit": "d3eb1e4eb2747386730b5aadf02b29ab8b4e1478"
         },
         "cargo-sources.json"
       ]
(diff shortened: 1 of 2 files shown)


bbhtt:
You can start the build when the exception is up

bbhtt:
bot, build dev.edfloreshz.CosmicTweaks

dev.edfloreshz.CosmicTweaks.json
```diff
@@ -0,0 +1,43 @@
...
+  "sdk": "org.freedesktop.Sdk",
+  "sdk-extensions": ["org.freedesktop.Sdk.Extension.rust-stable"],
+  "command": "cosmic-ext-tweaks",
+  "finish-args": [
+    "--share=ipc",
+    "--socket=fallback-x11",
+    "--socket=wayland",
+    "--device=dri",
+    "--talk-name=com.system76.CosmicSettingsDaemon",
+    "--filesystem=xdg-config/cosmic:rw"
```
bbhtt, line 14: You made the exception for ro but this is rw Which one is needed ?

edfloreshz: Oops, sorry I missed that, rw is the one needed.

edfloreshz: Here's the [PR](https://github.com/flathub-infra/flatpak-builder-lint/pull/487)

bbhtt:
bot, build dev.edfloreshz.CosmicTweaks

bbhtt:
Remove cosmic from categories, it was added last month to xdg-specs and no release has been made since then, so it is failing validation.

Also according to @jackpot51, they are going to use a different tag for marking apps suited for COSMIC see the discussion https://github.com/mariinkys/starrydex/pull/3#issuecomment-2380880405

bbhtt:
You are also missing releases https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/#release

bbhtt:
I think for OnlyShowIn to work you need a new desktop-file-utils release as well.

bbhtt:
I will update desktop-file-utils to the latest master commit. So you can just fix the metainfo file - remove cosmic from category and add releases.

edfloreshz:
Thanks! Latest commit should fix the issues, let me know if there's anything else missing.

bbhtt:
bot, build dev.edfloreshz.CosmicTweaks

bbhtt:
```
install: cannot stat './res/app_icon.svg': No such file or directory

```

Looks like the icon is no more.

edfloreshz:
Sorry about that, should be fixed now.

bbhtt:
bot, build dev.edfloreshz.CosmicTweaks

jackpot51:
@edfloreshz COSMIC category needs to be removed.

edfloreshz:
Should be fixed now

ahoneybun:
bot, build dev.edfloreshz.CosmicTweaks

bbhtt:
Exit buttons seem to be invisible and only a shadow shows up on hover. Is that intended ?


https://github.com/user-attachments/assets/388ba0ce-0c7e-48fb-9584-aab560961f91

bbhtt:
I also see a 

```
error sending request for url (https://cosmic-themes.org/api/themes/?order=name&limit=15&offset=15)
```

in terminal

wiiznokes:
> @edfloreshz COSMIC category needs to be removed.

Did you mean in category or onlyShowIn ? Because onlyShowIn is somehow useful

wiiznokes:
Or maybe we can use X-COSMIC instead. I haven't tested it tho

bbhtt:
OnlyShowIn=COSMIC will work now

I bumped desktop-file-utils to the latest commit in infra.

wiiznokes:
> Exit buttons seem to be invisible and only a shadow shows up on hover. Is that intended ?
> 
> 
> https://github.com/user-attachments/assets/388ba0ce-0c7e-48fb-9584-aab560961f91
> 
> 
> 
> 

https://github.com/pop-os/libcosmic/issues/550

bbhtt:
A baseapp with just the icon theme right now is not worth it. If it grows to multiple shared dependencies that all libcosmic apps need then it's fine.

If bundling the icon theme in the app, solves it then it's fine to do that for now.

wiiznokes:
Technically, with the onlyShowIn, I think it will works (at least it showed the icon when I was building it manually)

bbhtt:
I guess that's true for this one, but other apps will need to do it that aren't specific to only cosmic DE

wiiznokes:
Could it be an extension of `org.freedesktop.Sdk` ?

bbhtt:
There is an extension point for that `org.freedesktop.Platform.Icontheme`, but it doesn't autodownload. What you want is the icon theme gets automatically downloaded for everyone but that's not possible.

If you want you can submit a `org.freedesktop.Platform.Icontheme.COSMIC` but people have to manually install it.

edfloreshz:
Should I bring `OnlyShowIn=COSMIC` back? It was failing the last time it compiled.

bbhtt:
Yea should work now.

wiiznokes:
> There is an extension point for that `org.freedesktop.Platform.Icontheme`, but it doesn't autodownload. What you want is the icon theme gets automatically downloaded for everyone but that's not possible.
> 
> If you want you can submit a `org.freedesktop.Platform.Icontheme.COSMIC` but people have to manually install it.

So there is nothing we can specify in the manifest that will include the icon theme ?

bbhtt:
Bundle it ?

wiiznokes:
Yeah but i don't think the theme is cheap in size

edfloreshz:
I'm noticing some issues with the app, themes are not being applied, I'll try to figure out what's happening before we merge this.

bbhtt:
> Yeah but i don't think the theme is cheap in size

I think having a fallback icon here would solve this, unless you want to preserve same look and feel everywhere. 

If you are on COSMIC then you already have the theme installed with your DE and Flatpak will just use those from the host.

For people not on COSMIC either it uses a bundled version or a fallback icon theme.

If you want COSMIC icon theme for people not on COSMIC as well, then right now you have to:

1. define a new extension point in the GNOME runtime and Freedesktop Runtime `com.s76.COSMIC.IconTheme.FOobar` one that has `autodownload: true`. 
2. Then submit the theme as
`com.s76.COSMIC.IconTheme.FOobar.CosmicIconTheme`.

3. Then app developers can request it with:

```yaml
add-extensions:
 com.s76.COSMIC.IconTheme.FOobar.CosmicIconTheme
```

and it will get autoinstalled whenever people install it.

A baseapp will not help here as stuff from the baseapp is copied to the target app as is. So there won't be in savings in size.

bbhtt:
> I'm noticing some issues with the app, themes are not being applied, I'll try to figure out what's happening before we merge this.

Ok let men know when it's ready again.

wiiznokes:
Ok, seems like a nice solution for the future

wiiznokes:
> I'm noticing some issues with the app, themes are not being applied, I'll try to figure out what's happening before we merge this.

Good luck with that, i already tried to debugging this without success

bbhtt:
As a note, if you are adding new extension points to the fdsdk runtime, they can only be part of the next major release.

ie. If you add it now, it will be available for 25.08 + (October 2025).

These are part of the runtime ABI/API and fdsdk maintains both forward and backwards ABI.

edfloreshz:
> error sending request for url (https://cosmic-themes.org/api/themes/?order=name&limit=15&offset=15)

This is due to it not having network access, I know `--share=network` is not allowed in the manifest, how should I handle this? @bbhtt

bbhtt:
It's only not allowed as a `build-arg` ie. you can't request network at build time. At runtime in `finish-args` it's completely fine if you need it.

edfloreshz:
Okay, thanks, that should do it, we're ready for building and if everything goes well, publishing.

bbhtt:
bot, build dev.edfloreshz.CosmicTweaks

edfloreshz:
Working as expected, okay to merge

bbhtt:
> Yeah but i don't think the theme is cheap in size

https://github.com/flathub/flathub/pull/5648 bundled the theme and it's 9.5 MB in total. If the entirety of the icon theme is only 4-5 MB it's better to just bundle it than going through the overhead of create an extension point, then submit the theme as extension.

Issue #5654, opened 2024-10-01
Add COSMIC runtime


jackpot51:
According to https://docs.flathub.org/docs/for-app-authors/runtimes/#submission-process, "Runtime developers may submit a ticket to the Flathub issue tracker requesting that their runtime be added to Flathub."

I am working on a runtime for COSMIC applications to simplify some common items, namely installing `just` and the required icon themes. The source code can be found here: https://github.com/pop-os/cosmic-flatpak-runtime/. The guidelines around runtimes are sparse, and I need help to understand them. For example, there is a listed requirement "CI/CD workflow: Flathub uploads must be automated as a part of the regular build/update process for the runtime, so that security updates are not delayed". I would of course want to do this, but I do not have the ability to upload to flathub at this time so I am unable to test such CI.

This would simplify #5649 and #5648, I'm requesting review by @bbhtt who has reviewed these COSMIC apps.

bbhtt:
Having another runtime is probably not worth it due to the maintenance overhead of maintaining multiple branches, guaranteeing security updates/rebuilds and ABI/API compatibility and also to prevent fragmentation for end users.

Unless you have several 100s of modules like KDE has for Qt + KDE components.

Looking at the modules, `just` seems to be a build only dependency and isn't needed by end users (runtimes are installed by end users). The rest are icon and theme which only account for a few MBs.

What you most likely want is a baseapp similar to the Elementary baseapp we have?

bbhtt:
> A baseapp could be good, is there an example of an app derived from the elementary baseapp?

Elementary moved things to their own store a while ago, so we don't get regular submissions for it. But there is a bunch of older ones using it. https://github.com/search?q=org%3Aflathub+%22io.elementary.BaseApp%22&type=code&p=1

jackpot51:
Thanks, I will look into this.

jackpot51:
Would a base app mean that the icons would be included in every app deriving from it, or would they just be downloaded once?

jackpot51:
If I want to make a baseapp, what do I do next?

jackpot51:
So, if two apps use the same baseapp with the icons, will they be stored on disk just once?

jackpot51:
This sounds ideal, then, and simpler than a runtime by far.

PR #5655, opened 2024-10-01
Add com.system76.Cosmic.BaseApp

jackpot51:

A COSMIC BaseApp that Includes the just command runner, pop-icon-theme, and cosmic-icons which most COSMIC applications need.

- [x] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about this submission.

### DIFF
--- a/com.system76.CosmicBaseApp.json
+++ b/com.system76.CosmicBaseApp.json
@@ -1,0 +1,63 @@
+{
+    "id": "com.system76.CosmicBaseApp",
+    "branch": "master",
+    "runtime": "org.freedesktop.Platform",
+    "sdk": "org.freedesktop.Sdk",
+    "runtime-version": "24.08",
+    "modules": [
+        {
+            "name": "just",
+            "buildsystem": "simple",
+            "build-commands": [
+                "install -Dm0755 just /app/bin/just"
+            ],
+            "sources": [
+                {
+                    "type": "archive",
+                    "strip-components": 0,
+                    "dest-filename": "just.tar.gz",
+                    "url": "https://github.com/casey/just/releases/download/1.35.0/just-1.35.0-x86_64-unknown-linux-musl.tar.gz",
+                    "sha256": "c4172306e241bd175c07316156a610593fa2b687ac49418520a375605d7cead4",
+                    "only_arches": [
+                        "x86_64"
+                    ]
+                },
+                {
+                    "type": "archive",
+                    "strip-components": 0,
+                    "dest-filename": "just.tar.gz",
+                    "url": "https://github.com/casey/just/releases/download/1.35.0/just-1.35.0-aarch64-unknown-linux-musl.tar.gz",
+                    "sha256": "a94c75426239fce8c1e7fdfccd20d60d987924f91ac45c843a14cd72d4d71e73",
+                    "only_arches": [
+                        "aarch64"
+                    ]
+                }
+            ]
+        },
+        {
+            "name": "pop-icon-theme",
+            "buildsystem": "meson",
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://github.com/pop-os/icon-theme/archive/refs/tags/v3.5.1.tar.gz",
+                    "sha256": "92e73f252e5e9c8a1124623a7d421410ef43c798a4a992955e6eb4f666f5c8a0"
+                }
+            ]
+        },
+        {
+            "name": "cosmic-icons",
+            "buildsystem": "simple",
+            "sources": [
+                {
+                    "type": "archive",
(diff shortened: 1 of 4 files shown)


jackpot51:
This replaces the COSMIC runtime proposed in #5654, and will simplify COSMIC application submissions #5648 and #5649. Requesting review by @bbhtt, who recommended this approach.

com.system76.CosmicBaseApp.json
```diff
@@ -0,0 +1,63 @@
+{
+    "id": "com.system76.CosmicBaseApp",
+    "branch": "1.0.0-alpha.2",
```
bbhtt, line 3: Does these components have an API/ABI involved ? Or apps are supposed to always use the latest version of theme from here ?

If so I'd drop the `branch`.

jackpot51:
Can I use finish_args here to allow COSMIC applications by default to access the settings daemon dbus service and the cosmic config folder?

com.system76.CosmicBaseApp.json
jackpot51: There are no ABI/API versioned components here as of now.

jackpot51: I have removed `branch`

wash2:
`separate-locales: false` is included in the elementary base app. Should we disable that as well?

bbhtt:
> `separate-locales: false` is included in the elementary base app. Should we disable that as well?

Currently no locale is produced here, but sure it's fine to add it.

bbhtt:
bot, build com.system76.CosmicBaseApp

hfiguiere:
there should be a script to cleanup stuff like `just` that is not needed at runtime.

also what toolkit does COSMIC uses?

jackpot51:
COSMIC uses libcosmic, which is statically compiled. Just is left in as it is the default build system for COSMIC applications.

jackpot51:
I think I can add the just cleanup because it will be available for the build of an app based on this, right?

hfiguiere:
> I think I can add the just cleanup 

if you do that, there is virtually no point in having that package.

bbhtt:
> there should be a script to cleanup stuff like just that is not needed at runtime.

Yea I'd had a script like this https://github.com/flathub/com.riverbankcomputing.PyQt.BaseApp/blob/d86580d472384d877a366d5aba3b61e79bcce1f3/com.riverbankcomputing.PyQt.BaseApp.yaml#L20-L28

Apps just add it in `cleanup-commands` https://github.com/flathub/com.riverbankcomputing.PyQt.BaseApp?tab=readme-ov-file#how-to-use like this:

```
cleanup-commands:
  - /app/cleanup-BaseApp.sh
```

> I think I can add the just cleanup because it will be available for the build of an app based on this, right?

No, cleanups run at the end of the build, so if you add it here the baseapp won't have `just`

jackpot51:
I have added a cleanup script

bbhtt:
bot, build com.system76.CosmicBaseApp

jackpot51:
The errors are  "finish-args-not-defined" and "toplevel-no-command". I don't believe either finish-args or a command should be defined, they are not for io.elementary.BaseApp.

jackpot51:
Should I submit a PR for exceptions?

jackpot51:
It looks like there is a problem with the ID being used. It needs to end with ".BaseApp". https://github.com/flathub-infra/flatpak-builder-lint/blob/6ca3f8131fed842123f82e1bc4b8f03bbef9069d/flatpak_builder_lint/checks/appid.py#L29

Is it okay if I change the ID to be `com.system76.Cosmic.BaseApp`?

jackpot51:
I have changed the ID and verified locally that it passes the linter.

jackpot51:
I have added flathub.json in order to skip appstream checks: https://github.com/flathub/io.elementary.BaseApp/blob/branch/juno-23.08/flathub.json

bbhtt:
Ah yea you need to end the ID with `.BaseApp` that's how the linter skips checks on it that don't matter. I forgot about that.

The `flathub.json` is unused, those parameters were dropped.

bot, build com.system76.Cosmic.BaseApp

jackpot51:
I have deleted flathub.json

jackpot51:
Yes, it is

jackpot51:
@barthalion if it helps, I can add a `branch`. It was recommended to drop it as we don't have any ABI components, in https://github.com/flathub/flathub/pull/5655#discussion_r1783226509

jackpot51:
@bbhtt or @barthalion, what else do I need to do for this?

bbhtt:
PR is ok, I think barthalion is just busy, usually barthalion handles the merges.

PR #5651, opened 2024-09-30
Add io.github.buonhobo.KatharaGUI

BuonHobo:
### Please confirm your submission meets all the criteria

- [X] Please describe the application briefly.
  
KatharaGUI is a simple GUI for the Kathara network emulation tool. 
It allows you to start and stop labs and connect to the containers in a convenient way. 
KatharaGUI requires Docker to be installed and correctly configured on your system. 
This app is written in Python and uses GTK 4, libadwaita, and VTE.

- [X] The domain used for the application ID is [controlled by the application developers][appid-domain] and the [application id guidelines][appid] are followed.
- [X] I have read the and followed all the [Submission and licence requirements][reqs].
- [X] I have [built][build] and tested the submission locally.
- [X] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about this submission. **Link:**

[appid-domain]: https://docs.flathub.org/docs/for-app-authors/requirements/#control-over-domain-or-repository
[appid]: https://docs.flathub.org/docs/for-app-authors/requirements#application-id
[reqs]: https://docs.flathub.org/docs/for-app-authors/requirements
[build]: https://docs.flathub.org/docs/for-app-authors/submission/#before-submission

PR #5659, opened 2024-10-02
Add io.github.buonhobo.Katharsis

BuonHobo:
This is a remake of this PR, https://github.com/flathub/flathub/pull/5651.
I had to make a new one because I renamed the application to Katharsis.

Please refer to what's written over there


bbhtt:
There wasn't any need to open a new PR, you could've just renamed the files.


bot, build io.github.buonhobo.Katharsis

bbhtt:
Doesn't seem to be running this time:

```
  File "/app/lib/python3.12/site-packages/urllib3/connectionpool.py", line 789, in urlopen
    response = self._make_request(
               ^^^^^^^^^^^^^^^^^^^
  File "/app/lib/python3.12/site-packages/urllib3/connectionpool.py", line 495, in _make_request
    conn.request(
  File "/app/lib/python3.12/site-packages/urllib3/connection.py", line 441, in request
    self.endheaders()
  File "/usr/lib/python3.12/http/client.py", line 1331, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.12/http/client.py", line 1091, in _send_output
    self.send(msg)
  File "/usr/lib/python3.12/http/client.py", line 1035, in send
    self.connect()
  File "/app/lib/python3.12/site-packages/docker/transport/unixconn.py", line 26, in connect
    sock.connect(self.unix_socket)
FileNotFoundError: [Errno 2] No such file or directory

During handling of the above exception, another exception occurred:
```

BuonHobo:
Weird, it works on my end, Katharsis requires docker to be configured on the host and requires you to be in the docker group. Did you do that?

Can you show me if on your host system there is a docker socket in /run/docker.sock? It's supposed to be there AFAIK

bbhtt:
there is yes.

image

bbhtt:
Previous PR was working, so something changed in between

BuonHobo:
Does the previous PR still work for you? Does it work if you try to reload?
Does the upstream build https://github.com/BuonHobo/Katharsis/actions/runs/11141908734 work for you?

I've tried on a second computer and it still works for me. This is really weird and i don't really know how to debug this since I can't replicate and the issue seems to be in the underlying kathara library (it handles docker connection by itself, I didn't touch that part).

One thing I changed is that now Katharsis does an automatic refresh on startup, but it didn't do it in the previous PR.
So maybe yesterday it allowed you to open it but it would still have broken when trying to reload?

Very weird, I assume you are in the docker group so it's probably not a configuration issue on your part. It's probably an issue with the underlying library. Can you give me more information about your system? Even though it should't really matter...

bbhtt:
With this PR I don't get any application window, https://github.com/flathub/flathub/pull/5651#issuecomment-2385262462 <- installing this one at least the app window shows up.

BuonHobo:
The window shows up, but if it can't connect to docker then there is no functionality at all so there's not much difference.
It looks like for some reason it cannot find the docker socket even though it's there and it is visible by root and the docker group.

Can you show the output of `id`?

BuonHobo:
I'll update the app in the coming days so that it shows an alert when docker is not working

bbhtt:
```
uid=1000(bbhtt) gid=1000(bbhtt) groups=1000(bbhtt),10(wheel),135(mock),972(docker) context=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
```

System is Fedora 40, GNOME wayland

BuonHobo:
I'm using fedora 41 on gnome Wayland, so there's probably no difference with regard to docker...

This is extremely weird, I tested it on 3 computers and it's always worked. Would it be possible to get a third person to try this?

bbhtt:
Ok got it to work. Looks like I had `export DOCKER_HOST=unix:///run/user/$UID/podman/podman.sock` in zshrc to use it via podman which seems to cause the issue.

BuonHobo:
Glad we could figure it out! Thank you for all the time you spent on this!

BuonHobo:
I agree that it's inconvenient to miss support for rootless docker, however this is a limitation with the underlying Kathara library. What I developed is just a GUI and unfortunately I have no control over the development of the Kathara framework.

I tried running Kathara with docker rootless and it throws errors when trying to install the Kathara network plugin. I'd be glad to add support for that feature once it's supported by Kathara.

As I said, I should really put an alert on startup, for now I'll avoid doing the automatic reload on launch, so that at least the GUI will show up but it won't detect running containers...

BuonHobo:
I pushed the change. Now the GUI will always startup and display a helpful text about configuring it properly.

BuonHobo:
bot, build io.github.buonhobo.Katharsis

BuonHobo:
Just pushed the final update before dedicating myself to work and study again.
What are the next steps in order to push this to flathub? 
Students in my university could benefit from this tool since Kathara is what we use for networking exams.

bot, build io.github.buonhobo.Katharsis

BuonHobo:
Ok nevermind THIS is the last one lol

bot, build io.github.buonhobo.Katharsis

BuonHobo:
Sorry, there was a bug.

bot, build io.github.buonhobo.Katharsis

BuonHobo:
I've greatly improved the UX by allowing users to reattach terminals and focus them when clicking on the entry. Without having to look for the correct window

bot, build io.github.buonhobo.Katharsis

Issue #5621, opened 2024-09-23
Verifying net.sf.fuse_emulator


bertogg:
Hi,

I'm the maintainer of `net.sf.fuse_emulator` (a ZX Spectrum emulator) and also a contributor to the upstream project.

I would like to have it verified in Flathub but I'm not sure if it's going to be possible with the current app ID.

The upstream home page is https://fuse-emulator.sourceforge.net/ , and when I chose the ID for the Flatpak app I used `net.fs.fuse_emulator` (probably inspired by `net.sf.VICE` or other apps with similar names).

In hindsight I realize that this is a problem because the verification process requests that I create a file at https://sf.net/.well-known/org.flathub.VerifiedApps.txt which I obviously cannot do.

My understanding is that I should rename the app in order to be able to have it verified.

Is this correct?

What would be the appropriate name scheme in this case?

1. `net.sourceforge.fuse-emulator.Fuse` (with hyphen) ?
2. `net.sourceforge.fuse_emulator.Fuse` (with underscore) ?
3. Something else?

Is there are any other pitfalls that I should be aware of?

Thanks!

PR #5672, opened 2024-10-05
Add net.sourceforge.fuse_emulator.Fuse

bertogg:

  Fuse is a Sinclair ZX Spectrum emulator. This a renamed version of the Fuse app already in Flathub, see details below.

### Notes about this submission

This app replaces https://github.com/flathub/net.sf.fuse_emulator

The reason why I'm renaming it is because I cannot verify the old one since Flathub asks me to create a file at https://sf.net/.well-known/org.flathub.VerifiedApps.txt (which I don't control). When this app was originally uploaded the verification mechanism didn't exist yet.

As explained in https://github.com/flathub/flathub/issues/5621, with this new name I should be able to verify the app with a file at https://fuse-emulator.sourceforge.net/.well-known/org.flathub.VerifiedApps.txt (which I do control).

The changes compared to the latest published version of net.sf.fuse_emulator are the following:

- App renamed from `net.sf.fuse_emulator` to `net.sourceforge.fuse_emulator.Fuse`.
- `<id>` tag renamed from `net.sf.fuse_emulator.desktop` to `net.sourceforge.fuse_emulator.Fuse`
- Added `<provides>` and `<replaces>` tags with the old id.
- `net.sf.fuse_emulator.desktop` renamed to `net.sourceforge.fuse_emulator.Fuse.desktop`
- `net.sf.fuse_emulator.sdl.desktop` renamed to `net.sourceforge.fuse_emulator.Fuse.SDL.desktop`
- Freedesktop runtime version updated from `23.08` to `24.08`.

A couple of additional notes:

- I made all the changes listed above on top of the existing https://github.com/flathub/net.sf.fuse_emulator repository. This means that this pull request contains the whole history. If this is a problem I can change it, but if not I would rather keep the complete history of the app (before and after the rename) in the same branch. 
- The screenshots are still URLs to the old flathub repository (at `net.sf.fuse_emulator`) because `flatpak-builder-lint` complains if I point to a repository that doesn't exist yet. I can update the links once the new repository is live, although if it's acceptable to ignore those lint warnings I can modify this pull request to use the new links already.


bertogg:
The only changes to this new app compared to the version already published are in commit https://github.com/flathub/flathub/pull/5672/commits/ce08769726a6da3fe02e437f4f0906b574728c73

net.sourceforge.fuse_emulator.Fuse.yaml
```diff
@@ -0,0 +1,123 @@
...
+        sha256: cdc60df19ab08bfe55344395739bb08f50fc15c92da3962fac334d3bff116965
+      - type: patch
+        path: audiofile-build.patch
+    cleanup:
+      - /bin
+  # Library to handle Spectrum snapshots, tapes and other files
+  - name: libspectrum
+    sources:
+      - type: archive
+        url: https://netcologne.dl.sourceforge.net/project/fuse-emulator/libspectrum/1.5.0/libspectrum-1.5.0.tar.gz
```
hfiguiere, line 60: please use generic downloads link for sourceforge.

```suggestion
        url: https://download.sourceforge.net/project/fuse-emulator/libspectrum/1.5.0/libspectrum-1.5.0.tar.gz
```

net.sourceforge.fuse_emulator.Fuse.yaml
```diff
@@ -0,0 +1,123 @@
...
+        sha256: a353cb46e9b1a281061d816353ea010d0a6fe78e6a17aa0b7b74271ca5e4acfc
+  # Misc utilities (not needed if you simply want to run Fuse)
+  - name: fuse-utils
+    sources:
+      - type: archive
+        url: https://netcologne.dl.sourceforge.net/project/fuse-emulator/fuse-utils/1.4.3/fuse-utils-1.4.3.tar.gz
```
hfiguiere, line 66:
```suggestion
        url: https://downloads.sourceforge.net/project/fuse-emulator/fuse-utils/1.4.3/fuse-utils-1.4.3.tar.gz
```

net.sourceforge.fuse_emulator.Fuse.yaml
```diff
@@ -0,0 +1,123 @@
...
+  # The emulator itself
+  - name: fuse
+    config-opts:
+      - --with-x
+      - --with-audio-driver=sdl
+      - --enable-desktop-integration
+    rm-configure: true
+    sources:
+      - type: archive
+        url: https://netcologne.dl.sourceforge.net/project/fuse-emulator/fuse/1.6.0/fuse-1.6.0.tar.gz
```
hfiguiere, line 77:
```suggestion
        url: https://downloads.sourceforge.net/project/fuse-emulator/fuse/1.6.0/fuse-1.6.0.tar.gz
```

net.sourceforge.fuse_emulator.Fuse.yaml
```diff
@@ -0,0 +1,123 @@
...
+      - --with-audio-driver=sdl
+      - --disable-desktop-integration
+    make-args:
+      - EXEEXT=-sdl
+    make-install-args:
+      - EXEEXT=-sdl
+    rm-configure: true
+    sources:
+      - type: archive
+        url: https://netcologne.dl.sourceforge.net/project/fuse-emulator/fuse/1.6.0/fuse-1.6.0.tar.gz
```
hfiguiere, line 105:
```suggestion
        url: https://downloads.sourceforge.net/project/fuse-emulator/fuse/1.6.0/fuse-1.6.0.tar.gz
```

net.sourceforge.fuse_emulator.Fuse.yaml
```diff
@@ -0,0 +1,123 @@
...
+        paths:
+          - fuse-config-dir.patch
+          - set-generic-icon.patch
+      - type: script
+        dest-filename: autogen.sh
+        commands:
+          - autoreconf -f -i -I m4
+      - type: shell
+        commands:
+          - cp /app/share/applications/fuse.desktop /app/share/applications/net.sourceforge.fuse_emulator.Fuse.SDL.desktop
```
hfiguiere, line 117: you should use `rename-desktop-file` in the top level of the manifest instead.

net.sourceforge.fuse_emulator.Fuse.appdata.xml
```diff
@@ -0,0 +1,136 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2018 The Fuse development team -->
+<component type="desktop">
+  <id>net.sourceforge.fuse_emulator.Fuse</id>
+  <provides>
+    <id>net.sf.fuse_emulator.desktop</id>
+  </provides>
+  <replaces>
+    <id>net.sf.fuse_emulator.desktop</id>
+  </replaces>
```
hfiguiere, line 10: it's either `provides` or `replaces`. In that case `replaces` is most appropriate.

net.sourceforge.fuse_emulator.Fuse.appdata.xml
```diff
@@ -0,0 +1,136 @@
...
+      Fuse can load most formats of snapshots, tapes and floppy disks
+      and it can emulate many peripherals such as joysticks, printers,
+      storage devices and network interfaces, among others.
+    </p>
+  </description>
+  <screenshots>
+    <screenshot width="1048" height="590" type="default">
+      <caption>Fuse running the 128 BASIC editor</caption>
+      <image>
+        https://raw.githubusercontent.com/flathub/net.sf.fuse_emulator/master/screenshots/fuse1.png
```
hfiguiere, line 79: this should be pinned on a specific revision. Sames fro the other screenshots. And remove the dimensions. Otherwise it's prone to breakages.

net.sourceforge.fuse_emulator.Fuse.appdata.xml
```diff
@@ -0,0 +1,136 @@
...
+        https://raw.githubusercontent.com/flathub/net.sf.fuse_emulator/master/screenshots/fuse4.png
+      </image>
+    </screenshot>
+  </screenshots>
+  <developer id="net.sourceforge.fuse-emulator"><name>The Fuse development team</name></developer>
+  <update_contact>fuse-emulator-devel_AT_lists.sourceforge.net</update_contact>
+  <url type="homepage">https://fuse-emulator.sourceforge.net/</url>
+  <url type="bugtracker">https://sourceforge.net/p/fuse-emulator/bugs/</url>
+  <content_rating type="oars-1.1">
+    <content_attribute id="violence-cartoon">none</content_attribute>
```
hfiguiere, line 106: all the `none` values can be removed.

net.sourceforge.fuse_emulator.Fuse.yaml
```diff
@@ -0,0 +1,123 @@
+app-id: net.sourceforge.fuse_emulator.Fuse
+branch: stable
```
hfiguiere, line 2:
```suggestion
```

don't use flathub branches

net.sourceforge.fuse_emulator.Fuse.yaml
hfiguiere: oh nevermind there are two desktop files.

net.sourceforge.fuse_emulator.Fuse.yaml
```diff
@@ -0,0 +1,123 @@
...
+        dest-filename: autogen.sh
+        commands:
+          - autoreconf -f -i -I m4
+      - type: shell
+        commands:
+          - cp /app/share/applications/fuse.desktop /app/share/applications/net.sourceforge.fuse_emulator.Fuse.SDL.desktop
+          - >-
+             desktop-file-edit --set-key=Name --set-value='SDL Fuse'
+             --set-key=Exec --set-value='fuse-sdl %f'
+             --set-key=Icon --set-value='net.sourceforge.fuse_emulator.Fuse'
```
hfiguiere, line 121:
```suggestion
             --set-icon=$FLATPAK_ID
```

set-generic-icon.patch
hfiguiere, line 1: I don't understand the purpose of this patch. 

Also to have the icons available they must be exported. There is a `rename-mime-icons` that take a list of icon names that will deal with that. Also the mime file shall be exported with the right name, and there is `rename-mime-file` for that. (works like `rename-desktop-file`)

net.sourceforge.fuse_emulator.Fuse.yaml
```diff
@@ -0,0 +1,123 @@
...
+          - set-generic-icon.patch
+      - type: file
+        path: net.sourceforge.fuse_emulator.Fuse.appdata.xml
+      - type: script
+        dest-filename: autogen.sh
+        commands:
+          - autoreconf -f -i -I m4
+      - type: shell
+        commands:
+          - install -m644 -D net.sourceforge.fuse_emulator.Fuse.appdata.xml /app/share/appdata/net.sourceforge.fuse_emulator.Fuse.appdata.xml
```
hfiguiere, line 91: this setion should be a `post-install`

net.sourceforge.fuse_emulator.Fuse.yaml
```diff
@@ -0,0 +1,123 @@
...
+  # Sound
+  - --socket=pulseaudio
+  - --device=all
+  # Network access (for Spectranet)
+  - --share=network
+  # Home directory (for loading and saving Spectrum software)
+  - --filesystem=home
+build-options:
+  cflags: -O2 -g
+  cxxflags: -O2 -g
```
hfiguiere, line 23:
```suggestion
```

there are unecesary and conflict with the defaults.

net.sourceforge.fuse_emulator.Fuse.yaml
bertogg: There are two `.desktop` files because I build the app twice with two different UIs. The first (main) desktop file is renamed with `rename-desktop-file`, but I don't know how to rename the second one without doing it by hand.

set-generic-icon.patch
bertogg: Hmmm, this is not for the host file chooser (which would require exporting the mime file and the icons, as you say).

This is so the GTK file chooser from the application (which runs inside the sandbox and has access to the host filesystem) shows the right icons.

The former is probably good to have, but I think the mime file has way too many icons and there's an open upstream bug about reducing the list a bit to keep only the relevant ones. As soon as I'm confident that we have a better mime file I'll be more confident to export it.

The reason why I'm changing `icon` to `generic-icon` is because otherwise the icons don't appear at all. From the original commit description:

```
The Freedesktop theme spec says that icons are first searched in
the current theme, then in its parents and finally in the default
"hicolor" theme.

"hicolor" is where Fuse installs its icons, however if generic-icon is
not specified in fuse.xml then it defaults to application-x-generic.

This means that if the current theme provides an icon named
application-x-generic it will be used instead of the ZX Spectrum one.
```

bertogg: More about `icon` and `generic-icon`: https://gitlab.gnome.org/GNOME/nautilus/-/issues/2190#note_1410485

net.sourceforge.fuse_emulator.Fuse.appdata.xml
```diff
@@ -0,0 +1,105 @@
...
+      </image>
+    </screenshot>
+  </screenshots>
+  <developer id="net.sourceforge.fuse-emulator"><name>The Fuse development team</name></developer>
+  <update_contact>fuse-emulator-devel_AT_lists.sourceforge.net</update_contact>
+  <url type="homepage">https://fuse-emulator.sourceforge.net/</url>
+  <url type="bugtracker">https://sourceforge.net/p/fuse-emulator/bugs/</url>
+  <content_rating type="oars-1.1"/>
+  <launchable type="desktop-id">net.sourceforge.fuse_emulator.Fuse.desktop</launchable>
+  <launchable type="desktop-id">net.sourceforge.fuse_emulator.Fuse.SDL.desktop</launchable>
```
bbhtt, line 104:
```suggestion
```

Add only the main desktop file matching the ID.

net.sourceforge.fuse_emulator.Fuse.appdata.xml
```diff
@@ -0,0 +1,105 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2018 The Fuse development team -->
+<component type="desktop">
+  <id>net.sourceforge.fuse_emulator.Fuse</id>
+  <replaces>
+    <id>net.sf.fuse_emulator.desktop</id>
+  </replaces>
```
bbhtt, line 7: can you add the same thing with the provides tag? replaces is newer and not everyone supports it.

net.sourceforge.fuse_emulator.Fuse.appdata.xml
bertogg: Sure, I was told [some months ago](https://github.com/flathub/net.sf.fuse_emulator/pull/9#issuecomment-2002265075) to add two `<launchable>` tags if I had two `.desktop` files, but I see [in the spec](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-launchable) that I should try to avoid that.

I'll keep only the main one.

net.sourceforge.fuse_emulator.Fuse.appdata.xml
bertogg: Ok, @hfiguiere suggested to use `<replaces>`, but I can use `<provides>` instead.

Shall I switch to `<replaces>` in one or two years, or what do you recommend?

bbhtt: Use both, not every software store supports replaces.

bbhtt: It's in the docs https://docs.flathub.org/docs/for-app-authors/maintenance#renaming-the-flatpak-id

bbhtt:
bot, build net.sourceforge.fuse_emulator.Fuse

bbhtt:
bot, build net.sourceforge.fuse_emulator.Fuse

bbhtt:
You should include the appdata file in the tarball, given it's a cross desktop standard and the screenshots should be in the project website: fuse-emulator.sourceforge.net

bertogg:
I just tested the build, it looks good.

About moving the screenshots to the upstream website and the appdata file to the tarball, that is planned.

Thanks!

PR #5680, opened 2024-10-07
Add io.github.loissascha.mangoverlay

loissascha:

  MangOverlay is an open source GUI app for managing your MangoHud configuration. It aims to provide as much flexibility with it's settings as possible.


loissascha:
lint gives me the error "content-rating-missing", but I've updated my metainfo file to include it and it still shows it.

io.github.loissascha.mangoverlay.yml
```diff
@@ -0,0 +1,24 @@
...
+runtime-version: '47'
+sdk: org.gnome.Sdk
+command: mangoverlay
+finish-args:
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --share=ipc
+  - --filesystem=host:ro
+  - --filesystem=home
+  - --talk-name=org.freedesktop.Flatpak
```
bbhtt, line 12: why is this needed ?

io.github.loissascha.mangoverlay.yml
```diff
@@ -0,0 +1,24 @@
...
+modules:
+  - name: MangOverlay
+    sources:
+      - type: archive
+        url: https://github.com/loissascha/MangOverlay/releases/download/v0.2.9/mangoverlay-v0.2.9-linux-amd64.zip
```
bbhtt, line 17: this does not work on aarch64. Did you see https://docs.flathub.org/docs/for-app-authors/requirements#flathubjson ?

bbhtt: this seems go, why not build from source ? There are lots of go apps on Flathub.

bbhtt:
> lint gives me the error "content-rating-missing", but I've updated my metainfo file to include it and it still shows it.

why does the archive include files that are not in the git repo ?

loissascha:
> > lint gives me the error "content-rating-missing", but I've updated my metainfo file to include it and it still shows it.
> 
> why does the zip archive include files that are not in the git repo ?

the metainfo file is in the git repo inside the "build" folder. I included it in the zip because I wasn't sure how to get just that file from the git repo

loissascha:
@bbhtt I've now updated my yml file to get the metainfo, .desktop file and icon from the git repo and only the binary from the archive. I was not able to successfully setup a build from source, I think I'd have to ship all the go dependencies for wails and for my project which I don't think would be right to do.

bbhtt:
why not?

loissascha:
On every update I'd have to go through all the dependencies from both my backend and also my frontned (npm dependencies) and verify if any of them have changed/updated and then download all of them and bundle them as zip (?) and ship them? So I'd have to ship like 3 separate things. Wails.io dependencies, MangOverlay backend dependencies, MangOverlay frontend dependencies. It feels like **a lot** additional work where mistakes could easily happen compared to just shipping the binary which is available anyways.

loissascha:
at least that's what I think would be required as it doesn't have network access in the build system right? Please correct me if I'm wrong, this is the first time I'm trying to ship anything to flathub.

bbhtt:
There are tools to generate a manifest by pointing it to yarn.lock etc. You don't need to manually do it, same for go.

bbhtt:
Rest of the comments are unanswered still.

loissascha:
> Rest of the comments are unanswered still.

What do you mean by that? I have answered every single comment you left like 5 days ago already, have I not? It says I started a Review and all my comments have a label "Pending". Are they not visible to you? 


> There are tools to generate a manifest by pointing it to yarn.lock etc. You don't need to manually do it, same for go.

I wasn't able to find any information/guide on that. Could you point me towards something? That's why I initially asked in one of my comments if the way I'm trying to do the manual build is even correct, because I didn't find any information, tried and at some point gave up after I couldn't figure out how to get my dependencies.

bbhtt:
No you haven't

bbhtt:
> I wasn't able to find any information/guide on that. Could you point me towards something? That's why I initially asked in one of my comments if the way I'm trying to do the manual build is even correct, because I didn't find any information, tried and at some point gave up after I couldn't figure out how to get my dependencies.

io.github.loissascha.mangoverlay.yml
bbhtt: you have to allow network locally, using `build-args` generate the manifest, remove build-args and add the manifest.

PR #5687, opened 2024-10-08
Add org.kde.klevernotes

LouisSchul:

KleverNotes is a note-taking and management application using markdown.


org.kde.klevernotes.json
```diff
@@ -0,0 +1,61 @@
+{
+    "id": "org.kde.klevernotes",
+    "runtime": "org.kde.Platform",
+    "runtime-version": "6.6",
```
bbhtt, line 4: 6.7 is the latest, any reason you aren't using it ?

org.kde.klevernotes.json
```diff
@@ -0,0 +1,61 @@
...
+    "sdk": "org.kde.Sdk",
+    "base": "io.qt.qtwebengine.BaseApp",
```
bbhtt, line 6: you have to cleanup the baseapp, see https://github.com/flathub/io.qt.qtwebengine.BaseApp?tab=readme-ov-file#cleanup

org.kde.klevernotes.json
```diff
@@ -0,0 +1,61 @@
...
+    "base-version": "6.6",
+    "command": "klevernotes",
+    "finish-args": [
+        "--share=ipc",
+        "--socket=x11",
+        "--socket=wayland",
+        "--device=dri",
+        "--filesystem=home",
```
bbhtt, line 14: why is home needed ?

org.kde.klevernotes.json
```diff
@@ -0,0 +1,61 @@
...
+        "--env=QML_IMPORT_PATH=/app/qml",
+        "--env=QTWEBENGINE_DICTIONARIES_PATH=/app/qtwebengine_dictionaries",
+        "--talk-name=org.freedesktop.Flatpak"
```
bbhtt, line 17: this needs an exception why is this needed ?

org.kde.klevernotes.json
LouisSchul: You are right, I did not noticed it, I will bump the version.

org.kde.klevernotes.json
LouisSchul: Will add this.

org.kde.klevernotes.json
LouisSchul: I can't switch/create klevernotes storage if I remove this. 
Do you suggest using a different filesystem than `home` ?

org.kde.klevernotes.json
LouisSchul: It seems like I can remove `"--env=QTWEBENGINE_DICTIONARIES_PATH=/app/qtwebengine_dictionaries"` without any issue.

- `"--env=QML_IMPORT_PATH=/app/qml"` is for the WebEngineView, it won't load without it.
- `"--talk-name=org.freedesktop.Flatpak"` is used to be able to run external program, for example, syntax highlighters.

bbhtt: I'm talking about `org.freedesktop.Flatpak`. what syntax highlighters are these and can these be added to the Flatpak package ?

org.kde.klevernotes.json
bbhtt: what is "klevernotes storage" and where is it created ?

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+    "only-arches": ["x86_64"]
```
hfiguiere, line 2: why?

hfiguiere:
cc @flathub/kde

org.kde.klevernotes.json
LouisSchul: It's just a folder. The app need to be able to create and read from a folders/files. The user usually choose the location, that's why I choose `home` snce there's no predefined location.

org.kde.klevernotes.json
LouisSchul: I'm not sure it could be added as a Flatpak.

It basically rely on external syntax highlighting to do the work, such as [chroma](https://github.com/alecthomas/chroma) or [pygments](https://pygments.org/).

flathub.json
hfiguiere: it sure will never be tested. All `org.kde` is built also on aarch64.

LouisSchul: Alright, let's remove this.

org.kde.klevernotes.json
LouisSchul: Those are optionnal, it is up to the user to decide if they want to enable it or not. If they want it, they just install the required package and the option will appear, if not, the app is still completely usable.

bbhtt: if it's optional why have this permission by default for everyone ? Like you said those wanting it can add it themselves.

LouisSchul: I'm not very familiar with the way flatpak works. 
If this can be added afterward then I'm okay to remove it from this build, no problem.

I think we've covered all the issues with this initial commit, I'll update the PR based on your feedback, thank you.

hfiguiere: > Those are optionnal, it is up to the user to decide if they want to enable it or not. If they want it, they just install the required package and the option will appear, if not, the app is still completely usable.

Enabled vs installed.

My take is that they should be added to the flatpak so that the user can enable. Otherwise there is no point. It's not user friendly.

LouisSchul: I currently allow `chroma`, `pygmentize` and `ksynthaxhighlighter` for the external highlighters to choose from.
I also allow the use of `plantuml` to generate UML graph.

All of those are optionnals and disabled by default. To use them you install the required package and enable the options, that's it.

Having them installed by default would mean that I would need to ship 4 external program on top of KleverNotes inside this flatpak. 
What if I want to support more things in this way in the future ? Even more programs into the flatpak ?

I don't like this idea.

bbhtt: Then don't have this permission by default, users can add it with `flatpak override --talk-name=org.freedesktop.Flatpak` or through Flatseal or I think KDE settings allow overriding permissions as well.

org.kde.klevernotes.json
```diff
@@ -0,0 +1,60 @@
...
+    "runtime": "org.kde.Platform",
+    "runtime-version": "6.7",
+    "sdk": "org.kde.Sdk",
+    "base": "io.qt.qtwebengine.BaseApp",
+    "base-version": "6.6",
+    "command": "klevernotes",
+    "finish-args": [
+        "--share=ipc",
+        "--socket=x11",
+        "--socket=wayland",
```
bbhtt, line 12: it's either x11 only or wayland+fallback-x11 not both.

bbhtt:
bot, build org.kde.klevernotes

org.kde.klevernotes.json
```diff
@@ -0,0 +1,60 @@
+{
+    "id": "org.kde.klevernotes",
+    "runtime": "org.kde.Platform",
+    "runtime-version": "6.7",
+    "sdk": "org.kde.Sdk",
+    "base": "io.qt.qtwebengine.BaseApp",
+    "base-version": "6.6",
```
travier, line 7: Shouldn't this be 6.7?

org.kde.klevernotes.json
travier: The location should be chosen via the file portal where the user would select the directory and then the app would get access to it. I don't think it needs full home access.

org.kde.klevernotes.json
```diff
@@ -0,0 +1,60 @@
...
+    "command": "klevernotes",
+    "finish-args": [
+        "--share=ipc",
+        "--socket=wayland",
+        "--socket=fallback-x11",
+        "--device=dri",
+        "--filesystem=home",
```
travier, line 14: Let's sort this list.

travier (changes requested):

org.kde.klevernotes.json
bbhtt: indeed

bbhtt, line 7:
```suggestion
    "base-version": "6.7",
```

org.kde.klevernotes.json
```diff
@@ -0,0 +1,60 @@
...
+    "runtime": "org.kde.Platform",
+    "runtime-version": "6.7",
+    "sdk": "org.kde.Sdk",
+    "base": "io.qt.qtwebengine.BaseApp",
+    "base-version": "6.7",
+    "command": "klevernotes",
+    "finish-args": [
+        "--device=dri",
+        "--env=QML_IMPORT_PATH=/app/qml",
+        "--filesystem=home",
```
travier, line 12: Should be removed unless it clearly does not work without (and a bug should be filed then).

LouisSchul: You're right, thing seems to work fine without `"--filesystem=home"`. 
Only the displayed path to the storage seems to be affected but that's no big deal.

travier:
bot, build org.kde.klevernotes

travier:
@LouisSchul Note that you have to trigger a build manually here with a comment with `bot, build org.kde.klevernotes` every time you push.

bbhtt:
waiting for someone from flathub/kde to test it, otherwise seems fine.

org.kde.klevernotes.json
```diff
@@ -0,0 +1,59 @@
...
+    "base-version": "6.7",
+    "command": "klevernotes",
+    "finish-args": [
+        "--device=dri",
+        "--env=QML_IMPORT_PATH=/app/qml",
+        "--share=ipc",
+        "--socket=wayland",
+        "--socket=fallback-x11"
+    ],
+    "cleanup-commands": "/app/cleanup-BaseApp.sh",
```
travier, line 16: We should also clean `include`

LouisSchul:
bot, build org.kde.klevernotes

travier:
So, we've just merged this app, and right away we get errors on the first fedc PR: https://github.com/flathub/org.kde.klevernotes/pull/1

Is there something missing in the validation process here?

hfiguiere:
because f-e-d-c is failing here (or is it anitya?):

image

Note the versions of the main module, that is the one that provide the appstream file fail validation.

travier:
Arg, I missed that. Makes sense. Sorry for the noise.

PR #5709, opened 2024-10-13
Add me.amankhanna.opendeck

ninjadev64:

OpenDeck is a desktop application for using stream controller devices like the Elgato Stream Deck.


flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+	"only-arches": ["x86_64"]
```
hfiguiere, line 2: that's disappointing given that "cross-platform" is in it's description.

me.amankhanna.opendeck.yaml
```diff
@@ -0,0 +1,35 @@
+id: me.amankhanna.opendeck
+
+runtime: org.gnome.Platform
+runtime-version: "46"
```
hfiguiere, line 4: 47 is the current version

me.amankhanna.opendeck.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - --share=network
+  - --talk-name=org.freedesktop.Flatpak
+
+modules:
+  - shared-modules/libayatana-appindicator/libayatana-appindicator-gtk3.json
+  - name: binary
+    buildsystem: simple
+    sources:
+      - type: file
+        url: https://github.com/ninjadev64/OpenDeck/releases/download/v2.1.2/opendeck_2.1.2_amd64.deb
```
hfiguiere, line 22: why not building from source?

me.amankhanna.opendeck.desktop
hfiguiere, line 1: this should be upstream.

me.amankhanna.opendeck.metainfo.xml
hfiguiere, line 1: this file should be upstream.

me.amankhanna.opendeck.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --device=all
+  - --share=ipc
+  - --share=network
+  - --talk-name=org.freedesktop.Flatpak
+
+modules:
+  - shared-modules/libayatana-appindicator/libayatana-appindicator-gtk3.json
+  - name: binary
```
hfiguiere, line 18: please use an actually relevant name. Like "opendeck".

me.amankhanna.opendeck.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - type: file
+        path: me.amankhanna.opendeck.desktop
+      - type: file
+        path: me.amankhanna.opendeck.metainfo.xml
+    build-commands:
+      - ar -x *.deb
+      - tar -xf data.tar.gz
+      - install -Dm755 usr/bin/opendeck /app/bin/opendeck
+      - install -Dm644 usr/share/icons/hicolor/512x512/apps/opendeck.png /app/share/icons/hicolor/512x512/apps/me.amankhanna.opendeck.png
+      - install -Dm644 me.amankhanna.opendeck.desktop /app/share/applications/me.amankhanna.opendeck.desktop
```
hfiguiere, line 34: so the .deb comes without a .desktop? It's a bug if it does.

flathub.json
ninjadev64: Cross-platform as in it supports Windows, macOS, and GNU+Linux. I don't think Tauri has ARM support.

me.amankhanna.opendeck.yaml
ninjadev64: Started this a few days ago when it wasn't, apologies.

me.amankhanna.opendeck.yaml
ninjadev64: https://github.com/tauri-apps/tauri/issues/11210#issuecomment-2408877526

me.amankhanna.opendeck.desktop
ninjadev64: There is an upstream .desktop file. However, the icon it specifies is "opendeck.png" and not "me.amankhanna.opendeck.png". If you want, I can use the upstream .desktop file but this will require me to include the icon as "opendeck.png".

me.amankhanna.opendeck.metainfo.xml
ninjadev64: Is that so? I have looked at the Flathub repos of other Tauri applications and they all seem to include the metainfo in the Flathub repo, not upstream. Especially considering it is not useful upstream as it is Flatpak-specific.

me.amankhanna.opendeck.yaml
ninjadev64: Will do. The Tauri docs use binary and I copy-pasted.

me.amankhanna.opendeck.yaml
ninjadev64: See my other comment about the .desktop file.

flathub.json
hfiguiere: it does support. build from source.

me.amankhanna.opendeck.yaml
hfiguiere: a few days ago it was already the current version. Has been for a few weeks now (September)

me.amankhanna.opendeck.yaml
hfiguiere: this is not a justification.

me.amankhanna.opendeck.desktop
hfiguiere: there are plenty of way to fix this icon. It's all documented. You should NOT have the .desktop file in this repository.

me.amankhanna.opendeck.metainfo.xml
hfiguiere: yes it is even written in the documentation. Two wrongs don't make a right.

me.amankhanna.opendeck.yaml
hfiguiere: the tauri documentation is wrong.

me.amankhanna.opendeck.yaml
hfiguiere: see my other comment too.

hfiguiere (changes requested):
Some changes still need to be made.

me.amankhanna.opendeck.yaml
```diff
@@ -0,0 +1,35 @@
...
+sdk: org.gnome.Sdk
+
+command: opendeck
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --device=all
+  - --share=ipc
+  - --share=network
+  - --talk-name=org.freedesktop.Flatpak
```
hfiguiere, line 14: this talk name is NOT allowed by default. Why do you need this?

me.amankhanna.opendeck.yaml
ninjadev64: Maybe I just took "not found" to mean not existing, not not installed. It's fixed now anyway

me.amankhanna.opendeck.desktop
ninjadev64: I'm asking if it is OK for the icon to not be named with the full ID, because that will allow be to use the icon from the .deb.

me.amankhanna.opendeck.yaml
ninjadev64: OpenDeck supports running optionally plugins compiled for Windows using Wine, and running Node.js plugins using Node. This talk name is used to run the system install of Wine and Node so these optional features don’t have to be bundled with OpenDeck and triple its size as well as cause redundancy if these tools are already installed on the host system.

me.amankhanna.opendeck.yaml
ninjadev64: I see. There seems to be no dependency manifest generator for Deno, so how generate a dependency manifest for Deno?

me.amankhanna.opendeck.metainfo.xml
ninjadev64: I am sorry if this seems rude, but I have just gone to the Flathub organisation main page on GitHub and clicked on 10 random applications that I recognise, 9/10 of them had the metainfo.xml file included in their Flathub org repo and not upstream.

See a few examples.
- https://github.com/flathub/com.vscodium.codium/blob/master/com.vscodium.codium.metainfo.xml
- https://github.com/flathub/com.modrinth.ModrinthApp/blob/master/com.modrinth.ModrinthApp.metainfo.xml
- https://github.com/flathub/com.rustdesk.RustDesk/blob/master/com.rustdesk.RustDesk.metainfo.xml

me.amankhanna.opendeck.yaml
ninjadev64: Yes OK, I had fixed this already.

me.amankhanna.opendeck.metainfo.xml
hfiguiere: Four wrongs don't make a right. You picked the worst examples possible.

ninjadev64: Alright, fine. You say "it is even written in the documentation". Well, maybe you should read the documentation.

https://docs.flathub.org/docs/for-app-authors/submission/#submission-pr: "Add the required files using git, commit them and push."

"Required files" links to https://docs.flathub.org/docs/for-app-authors/requirements#appstream, where the metainfo file is specified.

Therefore, your documentation not lists this file as required in the Flathub repo but also explicitly states that it should be added using git to the new_pr branch for the PR submission.

Also, I have found many more than 3 examples, I just linked a few. It makes perfect sense for this file to be here as it is Flatpak-specific and has no use upstream.

flathub.json
ninjadev64: Please see [my other comment](https://github.com/flathub/flathub/pull/5709#discussion_r1798579209) asking how to build from source when using Deno which there is no dependency manifest generator for.

ninjadev64:
Apologies if I have come across as being rude, I did not intend to do so.

me.amankhanna.opendeck.yaml
bbhtt: wine is available as a baseapp, that's a large sandbox hole to try using host packages, for some optional features.

me.amankhanna.opendeck.metainfo.xml
```diff
@@ -0,0 +1,55 @@
...
+  <launchable type="desktop-id">me.amankhanna.opendeck.desktop</launchable>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://github.com/ninjadev64/OpenDeck/raw/main/.github/readme/mainmenu.png</image>
+    </screenshot>
+    <screenshot>
+      <image>https://github.com/ninjadev64/OpenDeck/raw/main/.github/readme/multiaction.png</image>
+    </screenshot>
+    <screenshot>
+      <image>https://github.com/ninjadev64/OpenDeck/raw/main/.github/readme/plugins.png</image>
```
bbhtt, line 39: these screenshots should be from a commit, not `main`

me.amankhanna.opendeck.yaml
```diff
@@ -0,0 +1,35 @@
...
+    buildsystem: simple
+    sources:
+      - type: file
+        url: https://github.com/ninjadev64/OpenDeck/releases/download/v2.1.2/opendeck_2.1.2_amd64.deb
+        sha256: 567ed33d1eb4220a698751676eded4e9ec34e7ccb7ed30165e0e6fc5047c1442
+        only-arches: [ x86_64 ]
+      - type: file
+        path: me.amankhanna.opendeck.desktop
+      - type: file
+        path: me.amankhanna.opendeck.metainfo.xml
```
bbhtt, line 28: add these to the deb.

me.amankhanna.opendeck.metainfo.xml
bbhtt: It doesn't say to include them in the PR and you missed the heading of the entire metadata section:

```
These metadata files should directly come from upstream whenever possible as these are widely adopted Freedesktop standards that can be used by other distributions as well.
```

bbhtt: Turned it to a note in case it was not visible enough.

bbhtt: > "Required files" links to https://docs.flathub.org/docs/for-app-authors/requirements#appstream, where the metainfo file is specified.

And I can't find where "Required files" links to appstream, where do you see this? Those two are under different headers.

me.amankhanna.opendeck.yaml
ninjadev64: Done

me.amankhanna.opendeck.yaml
ninjadev64: I have done some superficial research around baseapps and have generally found that it is a sort of template for a Flatpak application, for example an Electron baseapp that provides the libraries and so forth that every Electron app would use. I don't think that a Wine baseapp (if I could even find it) would be the right fit for this, as OpenDeck itself does not run in Wine but instead various plugins.

Various plugins also run their own commands on the host system to interface with user-installed applications and the user also has the capability to specify their own commands to run - it would be very irritating for the user to only be able to run, say, xdg-open because no other commands are available from within the sandbox. I see that this permission was approved for the StreamController application by Core447 for the same reasons, so could the same please be done for OpenDeck?

me.amankhanna.opendeck.metainfo.xml
ninjadev64: I have pulled in the metainfo file from where I have now included it upstream.

me.amankhanna.opendeck.desktop
ninjadev64: I am now using the opendeck.desktop and opendeck.png files from the upstream .deb.

me.amankhanna.opendeck.yaml
ninjadev64: I see that the absence of a pnpm dependency manifest generator has been enough reason to allow other applications to build from the .deb (see GitButler for an example). Could the same please be done in this case as there is no Deno dependency manifest generator?

me.amankhanna.opendeck.yaml
ninjadev64: About 50% of them, as OpenDeck is designed to run plugins made for the official Windows/macOS-only official Stream Deck software, and so any plugins written in a compiled language have been compiled only for Windows and macOS. The other 50% are those written with web technologies or those published by myself or other plugin developers who want to support OpenDeck/GNU+Linux better.

ninjadev64:
Just pushed, by the way, because I made some mistakes, so check the most recent commit in case you're looking over it right now.

me.amankhanna.opendeck.metainfo.xml
ninjadev64: Done

me.amankhanna.opendeck.yaml
ninjadev64: As for the plugins that run their own commands, my own Starter Pack plugin has a "run command" action used by at least 50 people that I know of for sure that allows the user to specify a command to run on button press and release. For example, interfacing with pipewire to mute and unmute audio devices.

ninjadev64: https://github.com/ninjadev64/opendeck-starterpack/blob/main/src/run_command.rs

It is yet to be updated to check for the `container` env var to check if it is running in Flatpak and prepend `flatpak-spawn --host` if it is, but you get the idea.

ninjadev64:
I've just found this while looking at other submissions. I'll update to use these, they look perfect for what I need!

rename-desktop-file (string)

    Any desktop file with this name will be renamed to a name based on id during the cleanup phase. The appdata file launchable will be updated if needed.
rename-appdata-file (string)

    Any appdata (metainfo) file with this name will be renamed to a name based on id during the cleanup phase. The id in the file will be updated as needed and the subsequent provides. (since 1.4.1)
rename-mime-file (string)

    Any mimetypes file with this name will be renamed to a name based on id during the cleanup phase. (since 1.4.0)
rename-icon (string)

    Any icon with this name will be renamed to a name based on id during the cleanup phase. Note that this is the icon name, not the full filename, so it should not include a filename extension. 
rename-mime-icons (array of string)

    Any mime icons with any of these names will be renamed to a name prefixed with id during the cleanup phase. Note that these are the icon names, not the full filenames, so they should not include a filename extension. (since 1.4.0)

ninjadev64:
@hfiguiere Why didn't you tell me about those options when I asked about the icon?

me.amankhanna.opendeck.yaml
ninjadev64: For example this Toggle Action with Run Command that one user has:
> `amixer set Capture nocap` and `amixer set Capture cap` for unmute/mute, lol, and it just alternates what command it runs

bbhtt: can you open an exception pr for this? https://docs.flathub.org/docs/for-app-authors/linter#exceptions


error code is https://docs.flathub.org/docs/for-app-authors/linter#finish-args-flatpak-spawn-access

bbhtt:
bot, build me.amankhanna.opendeck

ninjadev64:
The build is functional

ninjadev64:
What else do I need to do for this to be merged?

bbhtt:
have some patience.

bbhtt:
I don't have the device, can you show a screenshot or a screerecord of the UI and plugin once connected?

ninjadev64:
> have some patience. 

I was just asking, since the changes-requested label is still present, if there was anything that I had to do that was missing.

ninjadev64:
Uploading Screencast From 2024-10-16 07-31-10.mp4…

ninjadev64:
https://github.com/user-attachments/assets/e2290db5-3898-4cf4-ab46-5111ee676a70

ninjadev64:
Thank you so much for your help @bbhtt!

ninjadev64:
So, the URL in the new repository's description (https://flathub.org/apps/details/me.amankhanna.opendeck) is a 404? I created an account on flathub.org and a few minutes later me.amankhanna.opendeck appeared in my Developer Portal, but when I actually click it, it takes me to the same 404 page. I've also just verified my domain, if that's something that needs to be done for the app to appear.

bbhtt:
https://docs.flathub.org/docs/for-app-authors/submission#how-long-does-it-take-to-get-the-build-published-after-merge

ninjadev64:
> https://docs.flathub.org/docs/for-app-authors/submission#how-long-does-it-take-to-get-the-build-published-after-merge

Thanks!

PR #5747, opened 2024-10-23
Add org.dash.dash-core

PastaPastaPasta:

    Fully validating Dash peer-to-peer network node, wallet and GUI
    
    See https://github.com/dashpay/dash

    Implementation for flatpack very similar to the https://github.com/flathub/org.bitcoincore.bitcoin-qt repo


org.dash.dash-core.json
```diff
@@ -0,0 +1,87 @@
...
+    "sdk": "org.freedesktop.Sdk",
+    "runtime-version": "24.08",
+    "command": "dash-entrypoint.sh",
+    "finish-args": [
+        "--socket=x11",
+        "--socket=pulseaudio",
+        "--device=dri",
+        "--share=ipc",
+        "--share=network",
+        "--filesystem=~/.dashcore:create",
```
bbhtt, line 13: use `persist=.dashcore`

org.dash.dash-core.json
```diff
@@ -0,0 +1,87 @@
...
+            "post-install": [
+                "/app/bin/internal/test_dash"
+            ],
+            "sources": [
+                {
+                    "type": "archive",
+                    "only-arches": [
+                        "x86_64"
+                    ],
+                    "url": "https://github.com/dashpay/dash/releases/download/v21.1.1/dashcore-21.1.1-x86_64-linux-gnu.tar.gz",
```
bbhtt, line 37: can this be built from source?

org.dash.dash-core.json
```diff
@@ -0,0 +1,87 @@
...
+                }
+            ]
+        },
+        {
+            "name": "misc",
+            "buildsystem": "simple",
+            "build-commands": [
+                "desktop-file-edit --set-key=\"Icon\" --set-value=${FLATPAK_ID} dash-qt.desktop",
+                "install -Dp -m 644 dash-qt.desktop /app/share/applications/${FLATPAK_ID}.desktop",
+                "install -Dp -m 644 dash.png /app/share/icons/hicolor/scalable/apps/org.dash.dash-core.png",
```
bbhtt, line 56: scalable is not PNG

org.dash.dash-core.metainfo.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>org.dash.dash-core</id>
+  <name>Dash Core</name>
+  <project_license>MIT</project_license>
+  <metadata_license>CC0-1.0</metadata_license>
+  <summary>
+    Fully validating Dash peer-to-peer network node, wallet and GUI
+  </summary>
```
bbhtt, line 9:
```suggestion
  <summary>Fully validating Dash peer-to-peer network node, wallet and GUI</summary>
```

org.dash.dash-core.metainfo.xml
```diff
@@ -0,0 +1,40 @@
...
+  <url type="homepage">
+    https://dash.org/
+  </url>
```
bbhtt, line 12:
```suggestion
  <url type="homepage">https://dash.org/</url>
```

bbhtt: Did you autoformat this or something? This must not have linebreaks like that unless it's inside description.

org.dash.dash-core.metainfo.xml
```diff
@@ -0,0 +1,40 @@
...
+    <screenshot type="default">
+      <image type="source">
+        https://docs.dash.org/en/stable/_images/overview.png
+      </image>
+    </screenshot>
+  </screenshots>
+  <releases>
+    <release date="2024-10-22" version="21.1.1"/>
+  </releases>
+  <content_rating type="oars-1.1"/>
```
bbhtt, line 36: is this rating correct? There should be something in the OARS generator related to money.

org.dash.dash-core.metainfo.xml
bbhtt, line 1: This should be submitted upstream.

org.dash.dash-core.json
PastaPastaPasta: It **can**, however, we use guix for deterministic builds for production see: https://github.com/dashpay/dash/blob/develop/contrib/guix/README.md and https://github.com/dashpay/guix.sigs. Multiple developers all build the exact same source and depends and produce the exact same bit for bit binaries. I think it is more valuable to use the deterministic binary. 

I guess it may be possible for you to use guix in your building, but guix takes a LONG time, and also, I've found it to be flaky in CI like environments

org.dash.dash-core.metainfo.xml
PastaPastaPasta: Just went through the generator and appears correct. Only stuff financial is as shown below

image

org.dash.dash-core.metainfo.xml
PastaPastaPasta: Just pulled it from https://github.com/flathub/org.bitcoincore.bitcoin-qt/blob/master/org.bitcoincore.bitcoin-qt.metainfo.xml

org.dash.dash-core.metainfo.xml
PastaPastaPasta: Can you clarify what you mean?

org.dash.dash-core.metainfo.xml
hfiguiere: Gambling should be "intense".

org.dash.dash-core.metainfo.xml
bbhtt: Open a PR adding this to https://github.com/dashpay/dash or an issue there

hfiguiere: This should be hosted and managed where the app is developped. It is not flatpak specific. The `install` rule of the build system should also install it in the right place. For the record, distro packages will use it too.

org.dash.dash-core.metainfo.xml
bbhtt: Yea fix the linebreaks first. That one is also wrong.

hfiguiere: two wrongs don't make a right.

org.dash.dash-core.metainfo.xml
PastaPastaPasta: Thanks for your response, can you provide any justification or documentation to support this? The application does not include any gambling functionality, and org.bitcoincore.bitcoin-qt is also marked as oars-1.1, the functionality in the two applications is very similar.

bbhtt: Going by the screenshot, does the send/receive allow to send or receive cryptocurrencies? Then it should have `<content_attribute id="money-purchasing">intense</content_attribute>`

org.dash.dash-core.metainfo.xml
PastaPastaPasta: Yes, I will have some pushes coming, just answering how they got there

org.dash.dash-core.metainfo.xml
PastaPastaPasta: Is there any documentation to support this? based on my view looking at:
- monero: https://github.com/flathub/org.getmonero.Monero/blob/master/extra/org.getmonero.Monero.metainfo.xml#L6C1-L7C1
- bitcoin: https://github.com/flathub/org.bitcoincore.bitcoin-qt/blob/master/org.bitcoincore.bitcoin-qt.metainfo.xml#L53
- exodus: https://github.com/flathub/io.exodus.Exodus/blob/master/io.exodus.Exodus.appdata.xml#L154

All three support sending / receiving cryptocurrency, and all three share the same OARS as this application currently. I cannot find any examples that show that this should qualify under "money-purchasing" but I also can't seem to find a definition anywhere besides in the generator itself? I'm not so opposed to changing it, but I'd prefer to see a definition for what exactly "money-purchasing" means

bbhtt: I'd say all 3 should have it but they were submitted a long time ago when I was not here. Money purchasing is more like dealing with any transactions that can involve actual money or things that can be converted to money. 

Unfortunately we don't control the OARS standard and it's quite an old specification which isn't actively developed anymore. So the docs are missing and the categories do not reflect the real world usage today.

org.dash.dash-core.json
PastaPastaPasta: I'm a bit concerned that with this users will have issues finding their wallets or other information

Assuming a user was previously using Dash Core (so had a ~/.dashcore), and install the flatpak. Flatpak will copy ~/.dashcore into ~/.var/app/org.dash.dash-core/data it appears?

But if a user wants to go find their wallets, they will probably look in ~/.dashcore and then get confused when it is not there, (especially since the directory there still exists). This leads me to simply wanting to keep it as `"--filesystem=~/.dashcore:create",`

Please let me know if I'm missing something?

org.dash.dash-core.metainfo.xml
PastaPastaPasta: done

org.dash.dash-core.metainfo.xml
PastaPastaPasta: https://github.com/flathub/flathub/pull/5747/commits/2dd1609f5206d38173bb47b652129cd76b408324

org.dash.dash-core.json
PastaPastaPasta: https://github.com/flathub/flathub/pull/5747/commits/bdc3b760a5ef05028f1f75b5d59ea9bca3f98ebe

org.dash.dash-core.json
bbhtt: Why would you/the user want to inspect `~/.dashcore`?

bbhtt:
bot, build org.dash.dash-core

org.dash.dash-core.json
PastaPastaPasta: Wallets are stored in ~/.dashcore (by default). Maybe a user has it setup to auto backup their wallets from ~/.dashcore. If they upgrade to the flatpak, I wouldn't want that to silently break (ie no new backups would be made because flatpak will be writing new wallets into ~/.app/org.dash.dash-core)

bbhtt: If that's an important thing, then use it. Otherwise persist is generally better for sandboxing.

org.dash.dash-core.json
```diff
@@ -0,0 +1,88 @@
...
+            ],
+            "sources": [
+                {
+                    "type": "file",
+                    "url": "https://raw.githubusercontent.com/dashpay/dash/v21.1.1/contrib/debian/dash-qt.desktop",
+                    "sha256": "59b6125eb0732137026d33e1e61dc8b84e20f90baba176cc84899c41cf8201df"
+                },
+                {
+                    "type": "file",
+                    "url": "https://raw.githubusercontent.com/dashpay/dash/4d1b648bad417ef9b1c5a96edfa04b0e382f096f/contrib/flatpak/org.dash.dash-core.metainfo.xml",
```
bbhtt, line 69: it's content_rating

PastaPastaPasta:
I'm potentially still getting some linter issues such as 
"appstream-screenshots-files-not-found-in-ostree" and "appstream-screenshots-not-mirrored-in-ostree". Are these blockers?

bbhtt:
No those will work in CI it's in the submission docs. You can start the build .

PastaPastaPasta:
bot, build org.dash.dash-core

bbhtt:
I get an error on starting:


```
Error: Prune mode is incompatible with -txindex
```

bbhtt:
just flatpak run org.dash.dash-core

PastaPastaPasta:
I think I figured out why. On launch did you select "limit block chain storage to..."? It appears that there is a bug there, where it enables pruning in a way that doesn't disable txindex, so it refuses to continue. If you don't select that option all appears to work (don't worry it won't use a ton of storage initially, so you should be able to launch without it, and as long as you don't run it for too long I think a lack of storage may be fine)

bbhtt:
> I think I figured out why. On launch did you select "limit block chain storage to..."?

That's selected by default for me.

PastaPastaPasta:
> > I think I figured out why. On launch did you select "limit block chain storage to..."?
> 
> That's selected by default for me.

Weird. I think it may auto-detect based on storage available? It isn't auto selected on my systems with plenty of storage available.

PastaPastaPasta:
Anything blocking this? is that bug blocking this? it'll be a few weeks until it's fixed in the production version.

bbhtt:
I was waiting for a version with the fix so that I can test it.

PastaPastaPasta:
I've pushed a commit bumping the build artifacts to v22.0.0-rc.1

PR #5758, opened 2024-10-30
Add io.github.ladaapp.lada

ladaapp:

Its an app that let's you remove and recover pixelated content in adult videos (common in Japan)
https://github.com/ladaapp/lada

- [X] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about this submission. 

Thank you for reviewing!


share/applications/io.github.ladaapp.lada.desktop
hfiguiere, line 1: this is currently never referenced and should be upstream

share/icons/hicolor/128x128/apps/io.github.ladaapp.lada.png
hfiguiere, line 1: this is currently never referenced and should be upstream

share/metainfo/io.github.ladaapp.lada.metadata.xml
```diff
@@ -0,0 +1,38 @@
...
+        <screenshot type="default">
+            <image>https://github.com/ladaapp/lada/blob/main/assets/screenshot_gui_2.png</image>
+            <caption>Exporting restored video</caption>
+        </screenshot>
+        <screenshot type="default">
+            <image>https://github.com/ladaapp/lada/blob/main/assets/screenshot_cli_1.png</image>
+            <caption>Exporting restored video via CLI</caption>
+        </screenshot>
+    </screenshots>
+</component>
```
hfiguiere, line 38: missing mandatory `releases` tag

io.github.ladaapp.lada.yaml
```diff
@@ -0,0 +1,48 @@
...
+runtime: org.gnome.Platform
+runtime-version: '47'
+sdk: org.gnome.Sdk
+command: lada
+finish-args:
+  - "--socket=wayland"
+  - "--socket=fallback-x11"
+  - "--device=dri"
+  - "--share=ipc"
+  - "--socket=pulseaudio"
```
hfiguiere, line 11: why does this need sound access?

io.github.ladaapp.lada.yaml
```diff
@@ -0,0 +1,48 @@
...
+  - name: lada
+    buildsystem: simple
+    build-commands:
+      - |
+        mkdir -p /app/model_weights/3rd_party
+        mv lada_mosaic_detection_model.pt lada_mosaic_edge_detection_model.pth lada_mosaic_restoration_model_bj_pov.pth lada_mosaic_restoration_model_generic.pth /app/model_weights/
+        # mv clean_youknow_video.pth /app/model_weights/3rd_party/ # todo
+      - python3 -m pip install --prefix=/app --no-deps '.[basicvsrpp,gui]'
+      - |
+        [ -d /app/share ] || mkdir /app/share
```
hfiguiere, line 23:
```suggestion
        install -d /app/share
```

README.md
```diff
@@ -0,0 +1,14 @@
+Run in project root directory:
+
+```shell
+flatpak remote-add --if-not-exists --user flathub https://dl.flathub.org/repo/flathub.flatpakrepo
+flatpak install --user -y flathub org.flatpak.Builder
+python -m pip install '.[gui,basicvsrpp,packaging-flatpak]'
+pip-compile --extra gui,basicvsrpp -o flatpak/requirements.txt setup.py
+req2flatpak --requirements-file flatpak/requirements.txt --target-platforms 312-x86_64 > flatpak/lada-pip-dependencies.json.tmp
```
hfiguiere, line 8: so what happen on aarch64?

share/applications/io.github.ladaapp.lada.desktop
ladaapp: Oh right, thanks. changed it.

io.github.ladaapp.lada.yaml
ladaapp: The app includes a simple media player which plays the video including sound

share/metainfo/io.github.ladaapp.lada.metadata.xml
ladaapp: sure will add a release in upstream then

README.md
ladaapp: For now I don't target to support ARM as I cannot test it myself. I currently also assume the user to have a CUDA capable / Nvidia GPU.

hfiguiere: This is not what the build think. It will try to build an aarch64

Also assuming nvidia is very wrong. What happen if the user doesn't have?

ladaapp: At least one of my core dependencies [MMC](https://mmcv.readthedocs.io/en/latest/get_started/installation.html#install-with-pip) only mentions CUDA GPU support. Pytorch seems to try to support more AMD and Intel GPUs recently but the ecosystem as a whole still seems to assume CUDA only.

I could try to export the models to ONNX in a later version as it has more compatibility across systems.

hfiguiere: this doesn't tell what happens.  Does it crash? Does it just do nothing? Does it tell the user?

ladaapp: When no GPU could be detected the app still technically works and will use the CPU (tested on x86_64) but I've added a banner to indicate that it's probably too slow to be useful.

screenshot_no_gpu_banner

Regarding aarch64: I wasn't successful trying to install pip requirements for aarch64. I'm not sure how the app would behave if started on aarch64 right now. Guess I have to look into aarch64 on x86_64 emulation so I can test this...

io.github.ladaapp.lada.yaml
```diff
@@ -0,0 +1,48 @@
...
+        # mv clean_youknow_video.pth /app/model_weights/3rd_party/ # todo
+      - python3 -m pip install --prefix=/app --no-deps '.[basicvsrpp,gui]'
+      - |
+        install -d /app/share
+        mv flatpak/share/* /app/share/
+      - |
+        patch -i $FLATPAK_BUILDER_BUILDDIR/patches/adjust_mmengine_resume_dataloader.patch -p2 -d /app
+        patch -i $FLATPAK_BUILDER_BUILDDIR/patches/bump_mmagic_mmcv_dependency_bound.patch -p2 -d /app
+        patch -i $FLATPAK_BUILDER_BUILDDIR/patches/enable_loading_vgg19_from_local_file.patch -p2 -d /app
+        patch -i $FLATPAK_BUILDER_BUILDDIR/patches/fix_diffusers_import.patch -p2 -d /app
```
bbhtt, line 29: use type: patch instead of manually doing this https://docs.flatpak.org/en/latest/module-sources.html#patch-sources

Make sure to have the correct path in the patch files.

README.md
ladaapp: The following dependencies do not provide aarch64 wheels
```
nvidia-cublas-cu12
nvidia-cuda-cupti-cu12
nvidia-cuda-nvrtc-cu12
nvidia-cuda-runtime-cu12
nvidia-cudnn-cu12
nvidia-cufft-cu12
nvidia-curand-cu12
nvidia-cusolver-cu12
nvidia-cusparse-cu12
nvidia-nvtx-cu12
triton
mmcv
```
For nvidia there seems to be options directly from nvidia but not provided by pytorch as far as I can tell
Triton (dependency of pytorch) is not available yet on aarch64: [GH issue](https://github.com/pytorch/pytorch/issues/130558)

I tried to skip them as they are technically not necessary and torch should fall back to CPU inference on aarch64

The other dependency mmcv is necessary but they only provide x86 builds. aarch64 seems to be possible but would have to built it from source

[GH discussion](https://github.com/open-mmlab/mmcv/discussions/2334)
[source build docs](https://mmcv.readthedocs.io/en/latest/get_started/build.html)

Setting up cross compilation builds and integrating this into flatpak manifest source creation via req2flatpak is over my head.

**TLDR;** If aarch64 support is a pre-requisite for submission to flathub I'd need to withdraw my submission for this app

Please let me know, the I'll close the PR

io.github.ladaapp.lada.yaml
ladaapp: I've spent more than two hours trying to do this and couldn't make it run.
I cannot apply the patch directly on the build dir where the the files are downloaded into as these files are basically zip files.
Only after pip install the files are extracted to installed to the /app dir
I hoped to make it work by adding another module just to run these patches which runs after the above steps happened
```
  - name: patch-installed-pip-dependencies
    buildsystem: simple
    build-commands: []
    sources:
      - type: patch
        paths:
          - "../patches/adjust_mmengine_resume_dataloader.patch"
          - "../patches/bump_mmagic_mmcv_dependency_bound.patch"
          - "../patches/enable_loading_vgg19_from_local_file.patch"
          - "../patches/fix_diffusers_import.patch"
        strip-components: 2
        options: ["-d", "/app"]
```
Which I hoped to be the equivalent to my shell command version above
This didn't work though:
```
Applying patch adjust_mmengine_resume_dataloader.patch
can't find file to patch at input line 3
Perhaps you used the wrong -p or --strip option?
The text leading up to this was:
--------------------------
|--- old/.venv/lib/python3.12/site-packages/mmengine/runner/loops.py	2024-09-22 22:11:32.120340983 +0200
|+++ new/.venv/lib/python3.12/site-packages/mmengine/runner/loops.py	2024-09-26 21:51:19.303710074 +0200
```

bbhtt: why do you have patches in the git repository? 

It'd be easier to apply them in a different branches in your git repo and use the commit of that branch, in the `sources` here.

ladaapp: Its easier to have the patches in my upstream repo for folks that don't use flatpak.
But there doesn't seem to be an issue in finding the patches only when they're applied.

Is there an issue with running patch in the `build-commands` instead of the git source?

bbhtt:
bot, build io.github.ladaapp.lada

bbhtt:
bot, build io.github.ladaapp.lada

README.md
hfiguiere: The prerequisite it to not build broken packages. If you package doesn't work on aarch64, it current is still build on that platform.

ladaapp:
bot, build io.github.ladaapp.lada

README.md
ladaapp: Thanks for the hint. Just found out about `flathub.json` and that I can specify which archs to build for

ladaapp:
As far as I can tell I've fixed all linting and build errors.
The last pipeline run still failed though on `Committing builds`
with
```
| Error: HTTP status client error (401 Unauthorized) for url (https://hub.flathub.org/api/v1/build/141836/extended)
```
I assume this is an infra issue.

bbhtt:
bot, build io.github.ladaapp.lada

bbhtt:
what makes this app so big? almost 3.5 GB download

ladaapp:
> what makes this app so big? almost 3.5 GB download

Runtime dependencies: Pytorch comes in at around ~1.6G and its's transitive Nvidia/Cuda dependencies of add another ~2.6G.
Couldn't find an [available runtime](https://docs.flatpak.org/en/latest/available-runtimes.html) for torch /w cuda.

ladaapp:
@bbhtt  I'll try to switch to onnxruntime-gpu. If I understand it correctly it also provides CUDA runtime but way smaller at <500M.
Should this be done in scope of this PR or a follow-up? (cannot tell if its a quick change, if there are incompatibility issues maybe needs more time/work)

ladaapp:
Okay not an easy thing. Ended up in dependency hell as the deployment/exporting tool (mmdeploy) which could convert it to onnx is not well maintained. I could build the tool only against some ancient onnxruntime-gpu version for CPU but couldn't make it work to compile against current CUDA version.
I don't think I can make this work. As this dependency and others from the same dev I use do not seem to be well maintained anymore I will have to look for alternatives. But this will be a bigger change as its one of my core dependencies unfortunately.

Is it fine to merge it as is (accepting the huge size)? Otherwise will have to withdraw the submission and maybe come back later once I've migrated to something else :man_shrugging:

bbhtt:
> @bbhtt I'll try to switch to onnxruntime-gpu. If I understand it correctly it also provides CUDA runtime but way smaller at <500M.

No it is fine, I was just curious. I'll test this tomorrow as it will take a while to download on the VM.

ladaapp:
Got it.
Just remember that you practically will need to connect a CUDA capable GPU to your VM if you want to use the app.
Technically you can run it on CPU but even the CUDA version barely runs in real-time (30-90 fps, depending on video source (resolution, number of visible mosaic regions) and GPU type).
Just tried it in an Ubuntu 22.04 VM with 4 vCores and took maybe 10 mins to show first few frames.
Just so you know what to expect, Good luck!

bbhtt:
Yea I can't get it to work in the VM, can you show a screenrecord of how the app works? 10-20s would be fine.

ladaapp:
@bbhtt , here you go:

https://github.com/user-attachments/assets/241e57ee-43b0-4414-96ab-c78df88e1016

Notice how the pixelized region is somewhat restored. The app is intended for nsfw content restoration, not cat videos so it doesn't work well but it should show you how it would perform...

ladaapp:
Is there anything still missing from my side you're waiting for or do I just need to wait a bit longer?

bbhtt:
I am a bit concerned about the size - 3.5 G seems too big for users to download but I don't have a suggestion as those are python modules and downloading/integrating them at runtime can be difficult.

bbhtt:
bot, build io.github.ladaapp.lada

PR #5833, opened 2024-11-10
Add net.zdechov.app.C_evo

Mailaender:
### Please confirm your submission meets all the criteria

- [X] Please describe the application briefly.
> A clone of Civilization II.
- [X] The domain used for the application ID is [controlled by the application developer(s)][appid-domain] and the [application id guidelines][appid] are followed.
- [X] I have read and followed all the [Submission requirements][reqs] and the [Submission guide][reqs2].
- [X] I have [built][build] and tested the submission locally.
- [X] I am an author/developer/upstream contributor to the project. If not, I contacted upstream developers about this submission. **Link:** http://c-evo.org/bb/viewtopic.php?f=4&t=229&p=1869#p1869

[appid-domain]: https://docs.flathub.org/docs/for-app-authors/requirements/#control-over-domain-or-repository
[appid]: https://docs.flathub.org/docs/for-app-authors/requirements#application-id
[reqs]: https://docs.flathub.org/docs/for-app-authors/requirements
[reqs2]: https://docs.flathub.org/docs/for-app-authors/submission
[build]: https://docs.flathub.org/docs/for-app-authors/submission/#before-submission

PR #5837, opened 2024-11-12
Add net.zdechov.app.C-evo

chronoscz:

C-evo is turn-based empire building strategy game inspired by Civilization II game.
* Is a fork and Lazarus/FPC port of the original C-evo game which can be found at [​www.c-evo.org](http://www.c-evo.org/).
* Is a free & open source game (FOSS). The source code is available with public domain license. You can freely download and modify/improve it.

### DIFF
--- a/net.zdechov.app.C-evo.yml
+++ b/net.zdechov.app.C-evo.yml
@@ -1,0 +1,50 @@
+app-id: net.zdechov.app.C-evo
+runtime: org.freedesktop.Platform
+runtime-version: '23.08'
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.freepascal
+command: c-evo
+finish-args:
+  - --socket=x11
+  - --socket=pulseaudio
+  - --share=ipc
+  - --device=dri
+modules:
+  - shared-modules/gtk2/gtk2.json
+  - name: c-evo
+    sources:
+      - type: svn
+        url: https://svn.zdechov.net/c-evo/tags/1.3.6
+        revision: r634
+    buildsystem: simple
+    build-commands:
+      - |
+        . /usr/lib/sdk/freepascal/enable.sh
+        lazbuild --build-mode=Release AI/StdAI/StdAI.lpi
+        mv AI/StdAI/libstdai.so AI/StdAI/libstdai-amd64.so
+        lazbuild --build-mode=Release Integrated.lpi
+      - install -Dm755 c-evo -t $FLATPAK_DEST/share/c-evo
+      - install -d $FLATPAK_DEST/bin
+      - ln -s $FLATPAK_DEST/share/c-evo/c-evo $FLATPAK_DEST/bin/c-evo
+      - install -Dm644 Language.txt -t $FLATPAK_DEST/share/c-evo
+      - install -Dm644 Language2.txt -t $FLATPAK_DEST/share/c-evo
+      - install -Dm644 Fonts.txt -t $FLATPAK_DEST/share/c-evo
+      - install -Dm644 Install/flatpak/${FLATPAK_ID}.desktop -t $FLATPAK_DEST/share/applications
+      - install -Dm644 Install/deb/c-evo.xml -t $FLATPAK_DEST/share/mime/packages
+      - install -Dm644 Graphics/c-evo_256x256.png $FLATPAK_DEST/share/icons/hicolor/256x256/apps/${FLATPAK_ID}.png
+      - install -Dm644 Graphics/c-evo_256x256.png -t $FLATPAK_DEST/share/pixmaps/application-cevo.png
+      - install -Dm644 Graphics/c-evo_256x256.png -t $FLATPAK_DEST/share/pixmaps/application-cevomap.png
+      - install -Dm644 AI/StdAI/libstdai-amd64.so -t $FLATPAK_DEST/share/c-evo/AI/StdAI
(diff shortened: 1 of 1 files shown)


chronoscz:
I am the author and the maintainer of this fork. Please close that another pull request done by another user. He notified me about that and I would like to use my own request and my configuration for this game. I already had unfinished flatpak files prepared for Qt5 variant which was not fully supported by the game. He helped to update flatpak build to with older GTK2. I will be updating this flatpak with future releases.

hfiguiere:
Please talk to the person that submitted https://github.com/flathub/flathub/pull/5833

We don't arbitrate that.

hfiguiere:
And please see the review comment there to match here if you really want the submission to be considered in lieu of the other one. Re-submitting create unecessary work for the reviewers.

net.zdechov.app.C-evo.yml
```diff
@@ -0,0 +1,50 @@
...
+  - --socket=pulseaudio
+  - --share=ipc
+  - --device=dri
+modules:
+  - shared-modules/gtk2/gtk2.json
+  - name: c-evo
+    sources:
+      - type: svn
+        url: https://svn.zdechov.net/c-evo/tags/1.3.6
+        revision: r634
```
hfiguiere, line 19: this is an invalid revision

hfiguiere: >           revision (string)
>               A specific revision number to use

This is not a number you put.

hfiguiere:
bot, build net.zdechov.app.C-evo

net.zdechov.app.C-evo.yml
```diff
@@ -0,0 +1,50 @@
...
+    sources:
+      - type: svn
+        url: https://svn.zdechov.net/c-evo/tags/1.3.6
+        revision: 634
+    buildsystem: simple
+    build-commands:
+      - |
+        . /usr/lib/sdk/freepascal/enable.sh
+        lazbuild --build-mode=Release AI/StdAI/StdAI.lpi
+        mv AI/StdAI/libstdai.so AI/StdAI/libstdai-amd64.so
```
hfiguiere, line 25: what happen on aarch64?

net.zdechov.app.C-evo.yml
chronoscz: Revision number in terms of Subversion is both r123 and 123. From their manual https://svnbook.red-bean.com/en/1.7/svn.tour.revs.specifiers.html "Referring to revision numbers with an “r” prefix (r314, for example) is an established practice in Subversion communities, and is both supported and encouraged by many Subversion-related tools. In most places where you would specify a bare revision number on the command line, you may also use the rNNN syntax."

Apparently flatpak builder accepts both r123 and 123. So I guess that it is just your review observation and it is not based on any real check. Also flatpak-builder-lint doesn't complain that it is not just a number.

Anyway I have updated that as both works ok.

net.zdechov.app.C-evo.yml
chronoscz: Good point. I have fixed the naming of StdAI library based on the architecture. Compilation itself should be ok.

hfiguiere: why does it need a different naming?

chronoscz: That is how it is currently done. Each AI has its own definition file where multiple architectures can be specified so the author of the particular AI can distribute his AI as closed source with multiple pre-compiled libraries for different architectures. StdAI as standard open source implementation has them pre-compiled in the repository so main project can be built and run without need to always recompile it for given architecture. If the app is packaged for particular architecture then it doesn't matter as much but still compiled library needs to have matching name so the game knows what to load.

#PATH StdAI.dll
#PATH_WIN32 StdAI-win32.dll
#PATH_WIN64 StdAI-win64.dll
#PATH_LINUX_I386 libstdai-i386.so
#PATH_LINUX_AMD64 libstdai-amd64.so
#PATH_LINUX_ARM32 libstdai-arm32.so
#PATH_LINUX_ARM64 libstdai-arm64.so

hfiguiere:
bot, build net.zdechov.app.C-evo

net.zdechov.app.C-evo.yml
```diff
@@ -0,0 +1,58 @@
+app-id: net.zdechov.app.C-evo
+runtime: org.freedesktop.Platform
+runtime-version: '23.08'
```
hfiguiere, line 3: the freedpascal SDK is not available for 24.08 so it would be better to use that.

```suggestion
runtime-version: '24.08'
```

chronoscz: That looks like contradiction. I habe also tested 24.08 and came to the same conclusion. Freepascal SDK is not yet available so newer runtime can't be used yet. Then there is no action here.

bbhtt: it's now available for 24.08, that's a typo

hfiguiere: yeah I'm the king of typos. When it's not the autocarrot.

chronoscz:
Interesting github feature to be able to directly commit suggested change. I see that you have created a new pascal sdk brach https://github.com/flathub/org.freedesktop.Sdk.Extension.freepascal/blob/branch/24.08/org.freedesktop.Sdk.Extension.freepascal.yml
Thanks. I will also later update our pascal wiki.

bbhtt:
bot, build net.zdechov.app.C-evo

PR #5911, opened 2024-12-06
Add com.super_productivity.SuperProductivity

aripollak:
Please add @johannesjo as a collaborator. Thanks for your review!

- [x] Please describe the application briefly: Advanced to-do list app with timeboxing, time tracking, and third-party integration
- [x] I am an author/developer/upstream contributor to the project. If not, I contacted upstream developers about this submission. **Link:** https://github.com/johannesjo/super-productivity/issues/25#issuecomment-2522749255

### DIFF
--- a/com.super_productivity.SuperProductivity.metainfo.xml
+++ b/com.super_productivity.SuperProductivity.metainfo.xml
@@ -8,7 +8,7 @@
   <url type="homepage">https://super-productivity.com/</url>
   <url type="vcs-browser">https://github.com/johannesjo/super-productivity/</url>
   <url type="donation">https://github.com/sponsors/johannesjo</url>
-  <launchable type="desktop-id">com.super_productivity.SuperProductivity</launchable>
+  <launchable type="desktop-id">com.super_productivity.SuperProductivity.desktop</launchable>
   <icon type="stock">com.super_productivity.SuperProductivity</icon>
   <categories>
     <category>Office</category>
@@ -36,18 +36,23 @@
   <screenshots>
     <screenshot type="default">
       <image>https://super-productivity.com/static/7d2933a034590b3f40301a65e1d52538/34ae2/standard.png</image>
+      <caption>An overview of today's tasks, including time estimates and progress tracking</caption>
     </screenshot>
     <screenshot>
       <image>https://super-productivity.com/static/80c1ca336f5d3ad06390fdfecc7c6ad7/34ae2/jira.png</image>
+      <caption>Task details imported from GitHub</caption>
     </screenshot>
     <screenshot>
       <image>https://super-productivity.com/static/f259c271688b81febdf8082b22645823/34ae2/break-reminder.png</image>
+      <caption>Break reminder</caption>
     </screenshot>
     <screenshot>
       <image>https://super-productivity.com/static/d56fca55ab08d2464fb7b0bbd15d43a0/34ae2/timeline-calendar.png</image>
+      <caption>A timeline view of tasks, including integration with calendar events</caption>
     </screenshot>
     <screenshot>
       <image>https://super-productivity.com/static/92c8d60be9f55a4bf6a8ebc97e3bb6a5/34ae2/worklog.png</image>
+      <caption>Worklog view, showing a summary of tasks broken down by year, month, week, and day</caption>
     </screenshot>
   </screenshots>
   <releases>
--- a/com.super_productivity.SuperProductivity.metainfo.xml
+++ b/com.super_productivity.SuperProductivity.metainfo.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0"?>
+<component type="desktop">
+  <id>com.super_productivity.SuperProductivity</id>
+  <name>Super Productivity</name>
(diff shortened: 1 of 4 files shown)


com.super_productivity.SuperProductivity.yml
```diff
@@ -0,0 +1,73 @@
...
+separate-locales: false
+rename-icon: superproductivity
+finish-args:
+  - --device=dri
+  - --share=ipc
+  - --socket=wayland
+  # Use socket=x11 instead of socket=fallback-x11 to make it easier for people to just
+  # unset ELECTRON_OZONE_PLATFORM_HINT if they have problems with Electron's
+  # experimental Wayland support
+  - --socket=x11
```
hfiguiere, line 17: it either `wayland` + `fallback-x11` or just `x11`. not both.

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "skip-arches": ["aarch64"]
```
hfiguiere, line 2: this is disappointing. why not building from sources?

com.super_productivity.SuperProductivity.yml
hfiguiere: then you disable wayland.

the linter will block this.

flathub.json
hfiguiere: There are a lot things f-e-d-c doesn't do well anyway.

As you just said this would make aarch64 packages and that's the idea.

com.super_productivity.SuperProductivity.yml
hfiguiere: > [Signal](https://github.com/flathub/org.signal.Signal) does something similar but does not set ELECTRON_OZONE_PLATFORM_HINT by default.

two wrongs do not make a right.

com.super_productivity.SuperProductivity.yml
```diff
@@ -0,0 +1,73 @@
...
+command: run.sh
+separate-locales: false
+rename-icon: superproductivity
+finish-args:
+  - --device=dri
+  - --share=ipc
+  - --socket=wayland
+  # Use socket=x11 instead of socket=fallback-x11 to make it easier for people to just
+  # unset ELECTRON_OZONE_PLATFORM_HINT if they have problems with Electron's
+  # experimental Wayland support
```
hfiguiere, line 16:
```suggestion
```

either is work or it doesn't. there is no middle ground where people have to tweak things.

com.super_productivity.SuperProductivity.yml
```diff
@@ -0,0 +1,73 @@
...
+  - --socket=x11
+  - --socket=pulseaudio
+  - --share=network
+  # Allow local file attachments
+  - --filesystem=home:ro
```
hfiguiere, line 21: why does this need full home access?

aripollak: The application has the ability to "attach" any local file to a task item. As far as I could tell, there was no way to allow this without an explicit filesystem mount, but maybe that could be fixed upstream with a different way of opening files.

flathub.json
hfiguiere:
```suggestion
  "only-arches": ["x86_64"]
```

This is a better way to write it. Still disappointed that you don't want to do it.

aripollak: I do want to do it, just not right at this moment since it would require more work and I'd like to get any flatpak at all released first. At least 88 people have voted for the [upstream issue](https://github.com/johannesjo/super-productivity/issues/25) requesting this package, but as far as I know there haven't been any requests for aarch64 native binary packages for Linux.

com.super_productivity.SuperProductivity.yml
aripollak: For reference, I believe the app uses [shell.openPath](https://www.electronjs.org/docs/latest/api/shell#shellopenpathpath), which seems to call out to [xdg-open](https://github.com/electron/electron/blob/578c1e5dae1cb341c97d329902abc089b4f7c612/shell/common/platform_util_linux.cc#L342). I don't know that there's currently a more correct way to do this. It seems like opening a file:// URI might do what the user expects.

com.super_productivity.SuperProductivity.metainfo.xml
bbhtt, line 1: was this submitted to https://github.com/johannesjo/super-productivity ?

bbhtt:
bot, build com.super_productivity.SuperProductivity

com.super_productivity.SuperProductivity.metainfo.xml
hfiguiere: for sure it is not found (not installed?)

aripollak: Yup. https://github.com/johannesjo/super-productivity/issues/25#issuecomment-2522749255

aripollak: Oh weird, I thought I ran lint repo successfully but I guess not. I'll install the metainfo.

aripollak:
bot, build com.super_productivity.SuperProductivity

aripollak:
bot, build com.super_productivity.SuperProductivity

PR #5920, opened 2024-12-10
Add studio.assetmanager.ams 

NathanaelA:

This application allows you to download and control your assets from local downloads, Epic's FAB, and Epic's Unreal Marketplace (UEM is depreciated, but some assets are still only on UEM behind the scenes and they haven't been migrated).  It also allows you to manage your projects, and Unreal Engines.


flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+    "only-arches": ["x86_64"]
```

studio.assetmanager.ams.yaml
```diff
@@ -0,0 +1,77 @@
...
+command: ams
+finish-args:
+  # X11 performance
+  - --share=ipc
+  # Access to network
+  - --share=network
+  # Access to X11
+  - --socket=x11
+  # Access to wayland
+  - --socket=wayland
```
hfiguiere, line 16: it's either, or, not both.

studio.assetmanager.ams.yaml
```diff
@@ -0,0 +1,77 @@
...
+  # Audio Access
+  - --socket=pulseaudio
+  # Access to all files
+  - --filesystem=host
```
hfiguiere, line 20: this is very broad

studio.assetmanager.ams.yaml
```diff
@@ -0,0 +1,77 @@
...
+        do
+          locale="$(basename -s .pak $lang)"
+          install -Dm644 -t "${FLATPAK_DEST}/share/runtime/locale/${locale%%-*}/" "$lang"
+          ln -sf "${FLATPAK_DEST}/share/runtime/locale/${locale%%-*}/$(basename $lang)" "${FLATPAK_DEST}/asset-manager-studio/locales/$(basename $lang)"
+        done
+    sources:
+      - type: file
+        only-arches:
+          - x86_64
+        url: https://assetmanager.studio/dl/0.0.27/asset-manager-studio_0.0.27_amd64.deb
```
hfiguiere, line 68: why not building from source?

studio.assetmanager.ams.metainfo.xml
```diff
@@ -0,0 +1,56 @@
...
+  <project_license>GPL-3.0</project_license>
+  <name>Asset Manager Studio</name>
+  <summary>The asset manager and launcher for the rest of us...</summary>
+  <description>
+    <p>
+      AMS is the easiest and fastest method to access all your FAB and Unreal Marketplace assets.
+    </p>
+    <p>
+      AMS supports:
+      <ul>
```
hfiguiere, line 15: `<ul>` inside `<p>` is invalid

studio.assetmanager.ams.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+rename-desktop-file: asset-manager-studio.desktop
+rename-icon: asset-manager-studio
+modules:
+  - name: ams-desktop
+    buildsystem: simple
+    build-commands:
+      # Extract .deb & cleanup
+      - ar -x asset-manager-studio*.deb
+      - rm -f asset-manager-studio*.deb
```
hfiguiere, line 35:
```suggestion
```

studio.assetmanager.ams.yaml
```diff
@@ -0,0 +1,77 @@
...
+      - tar -xf data.tar.xz
+      - rm -f control.tar.gz data.tar.xz debian-binary
```
hfiguiere, line 37:
```suggestion
```

studio.assetmanager.ams.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+      # Copy application to Flatpack destination, and then clean it up
+      - cp --preserve --recursive opt/* usr/* $FLATPAK_DEST
+      - rm -rf opt usr
```
hfiguiere, line 41:
```suggestion
```

studio.assetmanager.ams.yaml
```diff
@@ -0,0 +1,77 @@
...
+  - name: ams-desktop
+    buildsystem: simple
+    build-commands:
+      # Extract .deb & cleanup
+      - ar -x asset-manager-studio*.deb
+      - rm -f asset-manager-studio*.deb
+      - tar -xf data.tar.xz
+      - rm -f control.tar.gz data.tar.xz debian-binary
+
+      # Copy application to Flatpack destination, and then clean it up
```
hfiguiere, line 39:
```suggestion
      # Copy application to Flatpak destination, and then clean it up
```

studio.assetmanager.ams.metainfo.xml
hfiguiere, line 1: this file should be managed upstream.

studio.assetmanager.ams.yaml
```diff
@@ -0,0 +1,77 @@
...
+      - cp --preserve --recursive opt/* usr/* $FLATPAK_DEST
+      - rm -rf opt usr
+
+      # Install our two data files
+      - install ams.sh $FLATPAK_DEST/bin/ams
+      - install -Dm0644 "${FLATPAK_ID}.metainfo.xml" "${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.metainfo.xml"
+
+      # update Desktop file
+      - desktop-file-edit --set-key=Exec --set-value="ams %U" $FLATPAK_DEST/share/applications/asset-manager-studio.desktop
+      - desktop-file-edit --set-key=Icon --set-value="${FLATPAK_ID}" "${FLATPAK_DEST}/share/applications/asset-manager-studio.desktop"
```
hfiguiere, line 49:
```suggestion
      - desktop-file-edit --set-icon="${FLATPAK_ID}" "${FLATPAK_DEST}/share/applications/asset-manager-studio.desktop"
```

studio.assetmanager.ams.yaml
```diff
@@ -0,0 +1,77 @@
...
+      - desktop-file-edit --set-key=X-Flatpak-RenamedFrom --set-value='asset-manager-studio.desktop;' "${FLATPAK_DEST}/share/applications/asset-manager-studio.desktop"
```
hfiguiere, line 50:
```suggestion
```

you never need to do that. What bit of the documentation said so (so we fix it)

studio.assetmanager.ams.yaml
NathanaelA: I copied this from Signal's https://github.com/flathub/org.signal.Signal/blob/0e46425c4bafd2076351bd6d1244a11d38fc1e2d/org.signal.Signal.yaml#L12-L15

studio.assetmanager.ams.yaml
NathanaelA: Yeah, unfortunately, people can put their UE projects, asset vaults, & unreal engines ANYWHERE on the machine.   

On my machine I have two vaults in my main home/document folder (mine, & Epic default vault) and another one on a my unreal drive mount point.   My UE's are on a that same unreal mount point, and my projects are also on the same unreal mount point.   I don't see any easy way around this as there is no "set" locations for anything when it comes to Unreal.

studio.assetmanager.ams.yaml
hfiguiere: two wrongs don't make a right

flathub.json
hfiguiere: how about now?

studio.assetmanager.ams.yaml
NathanaelA: Well in this case the .so file that is built using an older version of the OS GLIBC so that it works on virtually all linux machines.   Not sure if this impacts Flatpak, but it does impact the AppImage/DEB versions.   ;-)   

Second It is far simpler to extract out everything from a  working .deb file like many of the other Electron apps do, since the Electron-builders flatpak support is extremely poor.

studio.assetmanager.ams.yaml
NathanaelA: Why wouldn't we want to delete this file, we don't want to leave it after it is d/l'd.   Or is this not a factor for some reason I don't understand?

hfiguiere: yep you don't understand that the build dir doesn't exist after the module is build.

studio.assetmanager.ams.yaml
NathanaelA: This is common in many of the other published flatpacs using the extraction of the deb file.   
See for example: 
https://github.com/flathub/com.bitwarden.desktop/blob/3603fbffe8c50fa4ea3e2c6e34d6f2545740507b/com.bitwarden.desktop.yaml#L31-L34

Is this no longer needed?   I looked at over a dozen electron from deb that are actually published to figure out what I needed to do.  ;-)

hfiguiere: because I never reviewed that package. it's not because they do it that it is right.

studio.assetmanager.ams.yaml
hfiguiere: so basically you just do like all the others. and don't want to provide an aarch64 which you'd get building from source, since this seems to be the only reason there is no aarch64 build here.

studio.assetmanager.ams.yaml
NathanaelA: Sorry, first time using Flatpak-builder,   It is quite the learning curve especially when the messages are non-explanatory on what is broken.    I realize the comment isn't terribly useful, but a year from now having comments that even state the obvious makes reading the yaml file easier to find whatever area I need to find.

studio.assetmanager.ams.yaml
NathanaelA: Copied it from Signal's repo; I'd rather have overkill and have everything work then wonder "Why" they did something and have weird issues.   So I copied anything that I wasn't sure why they were using.  :D

https://github.com/flathub/org.signal.Signal/blob/0e46425c4bafd2076351bd6d1244a11d38fc1e2d/org.signal.Signal.yaml#L82

studio.assetmanager.ams.metainfo.xml
hfiguiere: once again two wrong don't make a right.

A lot of the electron package are done by people unrelated to the upstream project where upstream doesn't care.

That doesn't change. The appstream file should be managed upstream. IT IS NOT FLATPAK SPECIFIC.

But you keep pushing back like you try to find excuses for virtually everything.

studio.assetmanager.ams.yaml
hfiguiere: once again two wrong don't make a right. Signal is not the documentation.

studio.assetmanager.ams.yaml
NathanaelA: 1. Electron-builder's support for Flatpak can only generate a standalone .flatpak not one that uses the proper image files.    At this point switching build systems would be very complex as I have multiple custom rules so that my CI/CD machines can build both Linux and Windows and sign the windows apps at the same time...

2. The .so file is **critically** important that it is linked (at least for the .deb/appimage) to the oldest GLIBC I can support, because Ubuntu 18, 20 (and many other distro's) use an older GLibC.   If I let the "default" build occur  I get an app that crashes on a decent chunk of linux machines because of GLibC issues.     I honestly don't know if this affects flatpak's...

studio.assetmanager.ams.yaml
NathanaelA: Biggest issue, is that when I see things like that it makes it appear as if that is the supported way for wayland support since it is on an actual published flatpak.     Electron can use x11 AND wayland, so my assumption is it would need access to both to choose which method to use.

studio.assetmanager.ams.metainfo.xml
NathanaelA: Ah, thank; I found this: https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html so now I'll try and figure out how to include it into the .deb file using Electron Builder.  :grin:

studio.assetmanager.ams.yaml
NathanaelA: Unfortunately the documentation has lots of issues (much outdated it seems) when dealing with Electron apps.  Going to actual published electron flatpak apps was the only way I made it work using Flatpak.

NathanaelA:
I believe I have resolved all your issues.   Thanks for the input and help understanding and simplifying some of the yaml!

studio.assetmanager.ams.yaml
```diff
@@ -0,0 +1,71 @@
...
+      - ar -x asset-manager-studio*.deb
+      - tar -xf data.tar.xz
+      - cp --preserve --recursive opt/* usr/* $FLATPAK_DEST
+
+      # Install our application script
+      - install ams.sh $FLATPAK_DEST/bin/ams
+
+      # Copy our metadata file from upstream deb
+      - mkdir -p "${FLATPAK_DEST}/share/metainfo/"
+      - cp --preserve -f "opt/asset-manager-studio/resources/${FLATPAK_ID}.metainfo.xml" "${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.metainfo.xml"
```
hfiguiere, line 41:
```suggestion
      -install -Dm644 "opt/asset-manager-studio/resources/${FLATPAK_ID}.metainfo.xml" "${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.metainfo.xml"
```

studio.assetmanager.ams.yaml
```diff
@@ -0,0 +1,71 @@
...
+  - name: ams-desktop
+    buildsystem: simple
+    build-commands:
+      # Extract .deb & and copy the data to where it belongs
+      - ar -x asset-manager-studio*.deb
+      - tar -xf data.tar.xz
+      - cp --preserve --recursive opt/* usr/* $FLATPAK_DEST
+
+      # Install our application script
+      - install ams.sh $FLATPAK_DEST/bin/ams
```
hfiguiere, line 37:
```suggestion
      - install -Dm755 ams.sh $FLATPAK_DEST/bin/ams
```

hfiguiere:
bot, build studio.assetmanager.ams

studio.assetmanager.ams.yaml
NathanaelA: Updated

studio.assetmanager.ams.yaml
NathanaelA: Updated

bbhtt:
bot, build studio.assetmanager.ams

PR #5921, opened 2024-12-10
Add com.ilnonop.sonnewebapp

IlNonoP:

- [x] Please describe the application briefly. This program allows you to turn websites into apps on your linux device, they will be added to the application menu quickly.


SonneWebApp.desktop
```diff
@@ -0,0 +1,8 @@
+[Desktop Entry]
+Name=Sonne Web App
+Comment=Program that transform a web site into an app
+Exec=/app/bin/run.sh
```
hfiguiere, line 4: exec is never a full path.

app/main.py
hfiguiere, line 1: the flathub repo is NOT meant to host the upstream code.

manifest.yml
hfiguiere, line 1: the manifest file name MUST match the app-id

manifest.yml
```diff
@@ -0,0 +1,74 @@
+app-id: com.ilnonop.sonnewebapp
+runtime: org.freedesktop.Platform
+runtime-version: '23.08'
```
hfiguiere, line 3: 24.08 is the current

manifest.yml
```diff
@@ -0,0 +1,74 @@
...
+sdk: org.freedesktop.Sdk
+command: /app/bin/run.sh
```
hfiguiere, line 5: this is NEVER a full path.

manifest.yml
```diff
@@ -0,0 +1,74 @@
...
+finish-args:
+  - --filesystem=home:rw
+  - --socket=x11
+  - --socket=wayland
```
hfiguiere, line 9: it's either or, not both (and wayland has the fallback in that case).

manifest.yml
```diff
@@ -0,0 +1,74 @@
+app-id: com.ilnonop.sonnewebapp
+runtime: org.freedesktop.Platform
+runtime-version: '23.08'
+sdk: org.freedesktop.Sdk
+command: /app/bin/run.sh
+finish-args:
+  - --filesystem=home:rw
```
hfiguiere, line 7: why does it need fill home access?

manifest.yml
```diff
@@ -0,0 +1,74 @@
...
+      - type: file
+        path: app/main.py
+      - type: file
+        path: app/icon.png
+      - type: file
+        path: app/run.sh
+      - type: file
+        path: SonneWebApp.desktop
+
+  - name: tkinter
```
hfiguiere, line 32: this is in the runtime (24.08),

hfiguiere:
missing appstream file.

hfiguiere:
> * The domain used for the application ID is [controlled by the application developer(s)](https://docs.flathub.org/docs/for-app-authors/requirements/#control-over-domain-or-repository) and the [application id guidelines](https://docs.flathub.org/docs/for-app-authors/requirements#application-id) are followed.

`ilnonop.com` doesn't exist.

manifest.yml
IlNonoP: Revolved

manifest.yml
IlNonoP: Because it has to access the folders ~/.config/SonneWebApp and the folder ~/.local/share/applications

manifest.yml
IlNonoP: corrected!

manifest.yml
```diff
@@ -0,0 +1,74 @@
...
+command: /app/bin/run.sh
+finish-args:
+  - --filesystem=home:rw
+  - --socket=x11
+  - --socket=wayland
+  - --share=network
+  - --talk-name=org.freedesktop.FileManager1
+
+modules:
+  - name: SonneWebApp
```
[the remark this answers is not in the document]
IlNonoP: Corrected!

IlNonoP (commented):
I tried to correct everything, but I have doubts

app/main.py
hfiguiere: This was unresolved. It's a blocker.

SonneWebApp.desktop
hfiguiere: it's not. It still a path, albeit it's even worse.

manifest.yml
hfiguiere: not it wasn't. See above.

manifest.yml
hfiguiere: not really. You removed sockets for the UI now.

manifest.yml
hfiguiere: it was indeed a request for you to use the latest runtime. And see below about how tkinter is in that version.

manifest.yml
hfiguiere: `~/.config/SonneWebApp`  is defintely wrong. If it uses the standard directory functions it won't be this in the sandbox, and you don't need any special permissions.

manifest.yml
```diff
@@ -0,0 +1,73 @@
...
+        build-commands:
+          - ./configure --prefix=${FLATPAK_DEST}
+          - make
+          - make install
+        post-install:
+          - chmod +w ${FLATPAK_DEST}/lib/libtk8.6.so
+
+           - name: SonneWebApp
+    buildsystem: simple
+    build-commands:
```
hfiguiere, line 57: if yaml syntax didn't have significant whitespace, this would be just cosmetic, but unfortunately whitespace define the semantic of YAML, therefor this is just wrong. Obviously you didn't try to build this (despite acknowledging doing so)

manifest.yml
bbhtt: Yes the basename of the filename should be the same as app ID.

PR #5930, opened 2024-12-13
Add org.kde.optiimage

tjkirch:

- [x] Please describe the application briefly: OptiImage is an image compressor for KDE that supports PNG, JPEG, WebP and SVG file types.  Homepage: https://apps.kde.org/optiimage/
- [x] I am an author/developer/upstream contributor to the project. If not, I contacted upstream developers about this submission. **Link:** https://discuss.kde.org/t/this-week-in-kde-apps-optiimage-first-release-itinerary-redesign-and-more-kde-blogs/26268/6

The message linked above is from @CarlSchwan, a KDE developer who's the author of OptiImage.  I also got some tips from them in PMs, and confirmed that they were OK with me updating dependencies, and that they'd like this to be marked official so they can update it in the future.  They apparently have access to any org.kde.* appids.

The lints gave warnings about remote icons and screenshots, which the guide says may be a false positive; I did run a build with `--mirror-screenshots-url=https://dl.flathub.org/media/` so hopefully that was enough.  It also warned about the screenshot missing a caption; I submitted [an MR](https://invent.kde.org/graphics/optiimage/-/merge_requests/12) for this upstream so hopefully it's not a hard blocker for this 1.0 release.

### DIFF
--- a/org.kde.optiimage.json
+++ b/org.kde.optiimage.json
@@ -1,7 +1,7 @@
 {
     "id": "org.kde.optiimage",
     "runtime": "org.kde.Platform",
-    "runtime-version": "6.6",
+    "runtime-version": "6.8",
     "sdk": "org.kde.Sdk",
     "sdk-extensions": [
         "org.freedesktop.Sdk.Extension.rust-stable"
--- a/org.kde.optiimage.json
+++ b/org.kde.optiimage.json
@@ -73,8 +73,9 @@
             "buildsystem": "cmake-ninja",
             "sources": [
                 {
-                    "type": "git",
-                    "url": "https://invent.kde.org/libraries/kirigami-addons.git"
+                    "type": "archive",
+                    "url": "https://download.kde.org/stable/kirigami-addons/kirigami-addons-1.6.0.tar.xz",
+                    "sha256": "376dae6fc5acac7d0905ce9fef3211be0705c6e2df52bb80dfde1eaa20fe1bfa"
                 }
             ]
         },
--- a/org.kde.optiimage.json
+++ b/org.kde.optiimage.json
@@ -12,8 +12,7 @@
         "--share=ipc",
         "--socket=fallback-x11",
         "--socket=wayland",
-        "--device=dri",
-        "--share=network"
+        "--device=dri"
     ],
     "modules": [
         {
--- a/org.kde.optiimage.json
+++ b/org.kde.optiimage.json
@@ -80,8 +80,13 @@
         {
             "name": "optiimage",
             "buildsystem": "cmake-ninja",
-            "builddir": true,
-            "sources": [ { "type": "dir", "path": ".", "skip": [".git"] } ]
+            "sources": [
+               {
+                    "type": "archive",
+                    "url": "https://download.kde.org/stable/optiimage/optiimage-1.0.0.tar.xz",
+                    "sha256": "05ef948c0fbf2b80dfa8f1f5a42457f8fd39e1b988b8af99679348e401c5f515"
+               }
+            ]
         }
     ]
 }
--- a/org.kde.optiimage.json
+++ b/org.kde.optiimage.json
@@ -8,6 +8,7 @@
     ],
     "command": "optiimage",
     "finish-args": [
+        "--filesystem=home",
         "--share=ipc",
         "--socket=fallback-x11",
         "--socket=wayland",
--- a/org.kde.optiimage.json
+++ b/org.kde.optiimage.json
@@ -43,8 +43,8 @@
             "sources": [
                 {
                     "type": "archive",
-                    "url": "https://github.com/shssoichiro/oxipng/archive/refs/tags/v9.0.0.tar.gz",
-                    "sha256": "534fa8f349f52b01c2ee4332cef25ce1311edca04209ac6d972e38b171550a1b"
+                    "url": "https://github.com/shssoichiro/oxipng/archive/refs/tags/v9.1.1.tar.gz",
+                    "sha256": "f50db90c7ef95c0964dc980f6596b821f362e15d6d4bab247f1eb4aab7554db8"
                 },
                 "oxipng-sources.json"
             ]
--- a/org.kde.optiimage.json
+++ b/org.kde.optiimage.json
@@ -7,8 +7,6 @@
         "org.freedesktop.Sdk.Extension.rust-stable"
     ],
     "command": "optiimage",
-    "tags": ["nightly"],
-    "desktop-file-name-suffix": " (Nightly)",
     "finish-args": [
         "--share=ipc",
         "--socket=fallback-x11",
--- a/org.kde.optiimage.json
+++ b/org.kde.optiimage.json
@@ -1,0 +1,89 @@
+{
+    "id": "org.kde.optiimage",
+    "runtime": "org.kde.Platform",
+    "runtime-version": "6.6",
+    "sdk": "org.kde.Sdk",
+    "sdk-extensions": [
+        "org.freedesktop.Sdk.Extension.rust-stable"
+    ],
+    "command": "optiimage",
+    "tags": ["nightly"],
+    "desktop-file-name-suffix": " (Nightly)",
+    "finish-args": [
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--device=dri",
+        "--share=network"
+    ],
+    "modules": [
+        {
+            "name": "jpegoptim",
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://github.com/tjko/jpegoptim/archive/refs/tags/v1.5.5.tar.gz",
+                    "sha256": "90a309d1c092de358bb411d702281ac3039b489d03adb0bc3c4ef04cf0067d38"
+                }
+            ]
+        },
+        {
+            "name": "oxipng",
+            "buildsystem": "simple",
+            "build-commands": [
+                "cargo --offline fetch --manifest-path Cargo.toml --verbose",
+                "cargo --offline build --release --verbose",
+                "install -Dm 755 target/release/oxipng -t /app/bin"
+            ],
+            "build-options": {
+                "append-path": "/usr/lib/sdk/rust-stable/bin",
+                "env": {
+                    "CARGO_HOME": "/run/build/oxipng/cargo"
+                }
+            },
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://github.com/shssoichiro/oxipng/archive/refs/tags/v9.0.0.tar.gz",
+                    "sha256": "534fa8f349f52b01c2ee4332cef25ce1311edca04209ac6d972e38b171550a1b"
+                },
+                "oxipng-sources.json"
+            ]
+        },
+        {
+            "name": "python3-scour",
+            "buildsystem": "simple",
+            "build-commands": [
+                "pip3 install --verbose --exists-action=i --no-index --find-links=\"file://${PWD}\" --prefix=${FLATPAK_DEST} \"scour\" --no-build-isolation"
+            ],
+            "sources": [
+                {
+                    "type": "file",
+                    "url": "https://files.pythonhosted.org/packages/75/19/f519ef8aa2f379935a44212c5744e2b3a46173bf04e0110fb7f4af4028c9/scour-0.38.2.tar.gz",
+                    "sha256": "6881ec26660c130c5ecd996ac6f6b03939dd574198f50773f2508b81a68e0daf"
+                }
+            ]
+        },
+        {
+            "name": "kirigami-addons",
+            "config-opts": [
+                "-DBUILD_TESTING=OFF",
+                "-DCMAKE_BUILD_TYPE=RelWithDebInfo",
+                "-DBUILD_WITH_QT6=ON"
+            ],
+            "buildsystem": "cmake-ninja",
+            "sources": [
+                {
+                    "type": "git",
+                    "url": "https://invent.kde.org/libraries/kirigami-addons.git"
+                }
+            ]
+        },
+        {
+            "name": "optiimage",
+            "buildsystem": "cmake-ninja",
+            "builddir": true,
+            "sources": [ { "type": "dir", "path": ".", "skip": [".git"] } ]
+        }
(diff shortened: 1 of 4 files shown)


hfiguiere:
cc @flathub/kde

hfiguiere:
bot, build org.kde.optiimage

tjkirch:
Looks like it doesn't build on aarch64.  I added a flathub.json to skip it.

bot, build org.kde.optiimage

hfiguiere:
> Looks like it doesn't build on aarch64. I added a flathub.json to skip it.

Why not fixing it instead?

tjkirch:
> Why not fixing it instead?

I'm not the developer, I'm just helping to get the application into Flathub.  If it doesn't work on ARM today, it's still better to have it available for x86 users, I think.  When ARM support is fixed/added upstream, we can add that to Flathub too :)

org.kde.optiimage.json
```diff
@@ -0,0 +1,93 @@
...
+    "id": "org.kde.optiimage",
+    "runtime": "org.kde.Platform",
+    "runtime-version": "6.8",
+    "sdk": "org.kde.Sdk",
+    "sdk-extensions": [
+        "org.freedesktop.Sdk.Extension.rust-stable"
+    ],
+    "command": "optiimage",
+    "finish-args": [
+        "--filesystem=host",
```
bbhtt, line 11: why is host needed? it does not use portals?

org.kde.optiimage.json
```diff
@@ -0,0 +1,93 @@
...
+            "build-options": {
+                "append-path": "/usr/lib/sdk/rust-stable/bin",
+                "env": {
+                    "CARGO_HOME": "/run/build/oxipng/cargo"
+                }
+            },
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://github.com/shssoichiro/oxipng/archive/refs/tags/v9.1.3.tar.gz",
```
bbhtt, line 45: according to https://github.com/shssoichiro/oxipng/issues/518 they support aarch64, so might want to report it there.

bbhtt: i see at least one other app building 9.1.1 on aarch64

org.kde.optiimage.json
CarlSchwan: home should be enough

Unfortunately, while we use the portal for selecting files to optimize, when the safe mode is enabled, we write to a new file with a suffix next to the original file. This can't be done with portals :(

org.kde.optiimage.json
CarlSchwan: yeah here is the update to 9.1.1 for the nighlty flatpak build https://invent.kde.org/graphics/optiimage/-/merge_requests/13

org.kde.optiimage.json
tjkirch: I initially had used `filesystem=home`, but OptiImage doesn't work for me like that.  The tools it calls (jpegoptim, oxipng, etc.) each give different errors, but they either fail to open existing files or create new files, even though everything I'm doing is in my home folder.  It only started working when I used `filesystem=host`.

org.kde.optiimage.json
tjkirch: I could update the PR to use oxipng 9.1.1 instead of 9.1.3, to match that upstream MR, but I'm not sure it would help.  The error was `error: linker 'aarch64-linux-gnu-gcc' not found`, which would be an issue with the build environment, not the oxipng source, right?

org.kde.optiimage.json
bbhtt: what errors are you seeing? can you post it?

tjkirch: This is what I get if I only have `filesystem=home` and not `filesystem=host`.  Different errors from each tool, but none of them work.

optiimage-errors-filesystem-home

org.kde.optiimage.json
bbhtt: no

org.kde.optiimage.json
hfiguiere: is your home `/home/tjk` or `/var/home/tjk`.

on a distro like SilverBlue it's `/var/home` with a symlink for `/home`, which make things unreachable with only home permissions.

bbhtt: that doesn't tell much, but if it is under `home` there is no reason `host` should be needed. `host` exposes home + `/run/media /mnt, /opt, /srv`, they don't make a difference here

tjkirch: > on a distro like SilverBlue it's `/var/home` with a symlink for `/home`, which make things unreachable with only home permissions.

It's `/var/home`.  So I should change the PR to `filesystem=home` and people with this type of system have to change permissions manually?

org.kde.optiimage.json
tjkirch: Ah, I see.  They did make a change in oxipng 9.1.3 that might be related: https://github.com/shssoichiro/oxipng/commit/e2d58c112ba8041f47dd94e5d130b142208562f8

I'll change this PR to use 9.1.1 like upstream OptiImage.

org.kde.optiimage.json
hfiguiere: Yes the permission should be `home`.

No they don't need to change the permission manually. The input path are incorrect. I only ran into the trouble because some path were stored and I downgraded from Workstation (where it is `/home`) to SilverBlue (where it is `/var/home`)

tjkirch: > No they don't need to change the permission manually. The input path are incorrect. I only ran into the trouble because some path were stored and I downgraded from Workstation (where it is `/home`) to SilverBlue (where it is `/var/home`)

What should the user do if they have the `/var/home` setup, and this Flatpak with `filesystem=host` gives them the errors I showed?  It seems like adding permissions is the only way to use the tool.  Unless Flatpak has a change coming to better handle that type of filesystem layout.

org.kde.optiimage.json
hfiguiere: > according to [shssoichiro/oxipng#518](https://github.com/shssoichiro/oxipng/issues/518) they support aarch64, so might want to report it there.

If you read the MR, it's just wrong. They don't support aarch64. They support cross-compiling for it.

https://github.com/shssoichiro/oxipng/pull/534/files#diff-9a4f3e4537ebb7474452d131b0d969d89a51286f4269aac5ef268e712be17268R1

```
[📦 org.freedesktop.Sdk ~]$ which aarch64-linux-gnu-gcc
/usr/bin/which: no aarch64-linux-gnu-gcc in (/app/bin:/usr/bin)
[📦 org.freedesktop.Sdk ~]$ arch 
aarch64
```

org.kde.optiimage.json
hfiguiere: > What should the user do if they have the /var/home 

Then file path in their home directoruy start with `/var/home`

`~` should expand correctly. So do `$HOME`. If not then their distro is misconfigured.

bbhtt: why does `host` make any difference? `host` doesn't change or expose anything in `/var`.

It specifically says `/var` is excluded.

```
            host
               The entire host file system, except for directories that
               are handled specially by Flatpak. In particular, this
               shares /home, /media, /opt, /run/media and /srv if they
               exist.

               /dev is not shared: use devices=all; instead.

               Parts of /sys are always shared. This option does not
               make additional files in /sys available.

               Additionally, this keyword provides all of the same
               directories in /run/host as the host-os and host-etc
               keywords. If this keyword is used in conjunction with one
               of the host- keywords, whichever access level is higher
               (more permissive) will be used for the directories in
               /run/host: for example, host:rw;host-os:ro; is equivalent
               to host:rw;.

               These other reserved directories are currently excluded:
               /app, /bin, /boot, /efi, /etc, /lib, /lib32, /lib64,
               /proc, /root, /run, /sbin, /tmp, /usr, /var.

               Available since 0.3.
```

tjkirch:
I pushed the changes (oxipng 9.1.1, and `filesystem=home`) and for some reason the PR was closed.  It won't let me reopen it.

org.kde.optiimage.json
bbhtt: then a good idea to ask them to support it natively. The original issue was talking about native aarch64

org.kde.optiimage.json
tjkirch: > why does host make any difference? host doesn't change or expose anything in /var.

I don't know.  I guess my distro is misconfigured, as hfiguiere said.  For now, I have to add host permissions to run this app.  Whether it's a distro problem or a Flatpak problem, I don't think it's relevant to this application being added, though.

tjkirch:
bot, build org.kde.optiimage

tjkirch:
Ah, now it's letting me reopen the PR, but it's requiring that I make a comment.  Sorry for the noise.

bbhtt:
It's not open, you changed how the branch was created. It's not branched from new-pr now.

tjkirch:
You're right, the "reopen and comment" button did not reopen the PR.  I've never had this issue with GitHub before.  It does appear that it's merging into new-pr on my screen, though.

Should I start a new PR?

org.kde.optiimage.json
hfiguiere: > why does `host` make any difference? `host` doesn't change or expose anything in `/var`.

Because. With `home` it export `/var/home`

```
% flatpak run --filesystem=home org.freedesktop.Sdk//24.08
[📦 org.freedesktop.Sdk ~]$ echo $HOME
/var/home/hub
[📦 org.freedesktop.Sdk ~]$ ls -ld /home
ls: cannot access '/home': No such file or directory
[📦 org.freedesktop.Sdk ~]$ ls -ld /var/home/hub/
drwx------. 1 hub hub 7908 Dec 13 21:28 /var/home/hub/
```

With host:

```
[📦 org.freedesktop.Sdk ~]$ ls -ld /home
lrwxrwxrwx. 1 hub hub 8 Dec 13 21:30 /home -> var/home
```

The rest is the same.
That's why it work with `host` but not `home`, because it can resolve the `/home` symlink.

bbhtt:
No, `git checkout org.kde.optiimage` -> `git rebase new-pr` -> `git push -f`, that will put it on top of new-pr again

Maybe it lets us open after that.

org.kde.optiimage.json
hfiguiere: it was copiloted.

(ie Microsoft wasting resource on stuff no one asked for, instead of fixing stuff people asked for like aarch64 runners)

org.kde.optiimage.json
hfiguiere: Just to clarify. The content of `$HOME` is what matters. The fact the `/home` works by symlink is just compatibility for broken software.

bbhtt: sounds like a Flatpak bug.

bbhtt: https://github.com/flatpak/flatpak/issues/3582

hfiguiere: no. it is not.

bbhtt: so is this accessing the path like `/home` instead of using `$HOME` or `~/`?

hfiguiere: that's what the screenshot shows, but I'm not sure how it was initiated.

bbhtt: It seems to be using `file://` URIs as Fedora provides the `home` symlinks the URI is `file:///home/foo/bar.png` which is wrong inside the Flatpak.

bbhtt: Looks like the difference between home and host, is that `host` shares `/home` (so the symlink gets created as well) while `home` shares `$HOME`.

hfiguiere: Worked fine here. Selected the file with the file dialog.

`--filesystem=home`.

image

Note that it used the Gtk file dialog, so I suspect it might be using the portal.

And I tested and it works without filesystem permissions.

hfiguiere: > And I tested and it works without filesystem permissions.

But because of how the portals work the output is stuck in the wrong place.

tjkirch:
I pushed the fixed commits, which you can see in [my branch](https://github.com/flathub/flathub/compare/new-pr...tjkirch:flathub:org.kde.optiimage), but this PR isn't giving me an option to change/fix the base branch, and it won't let me reopen it.

bbhtt:
branch is fine now, but github is being annoying

image

hfiguiere:
I :heart: github

image


#sarcasm

bbhtt:
Let's open a new PR, using a new branch from your end. this is annoying

bbhtt:
Looks like aarch64 of oxipng build works with the following hack:

```
ln -s /usr/bin/gcc /app/bin/aarch64-linux-gnu-gcc
```

tjkirch:
> Looks like aarch64 build of oxipng build works with the following hack:
> 
> ```
> ln -s /usr/bin/gcc /app/bin/aarch64-linux-gnu-gcc
> ```

Would you recommend adding that to the `build-commands` for oxipng, before cargo?

org.kde.optiimage.json
tjkirch: It looks like we're having the same error after downgrading to oxipng 9.1.1 (see https://buildbot.flathub.org/#/builders/31/builds/15654/steps/7/logs/stdio ) so if the workaround from https://github.com/flathub/flathub/pull/5930#issuecomment-2542723538 works, should I update back to the latest oxipng release, 9.1.3?

bbhtt:
Sure, and add a cleanup with `/bin/aarch64-linuc-gnu`

org.kde.optiimage.json
bbhtt: Yes if the workaround works it should be fine with 9.1.3 too.

tjkirch:
> Sure, and add a cleanup with `/bin/aarch64-linuc-gnu-gcc`

Do you mean a second symlink, so also running `ln -s /usr/bin/gcc /bin/aarch64-linux-gnu-gcc` ?

I can do that, re-update to oxipng 9.1.3, and create a new PR.

bbhtt:
No cleanup the symlink after the build is done that symlink is only needed for building

tjkirch:
New PR here: https://github.com/flathub/flathub/pull/5934

PR #5965, opened 2024-12-24
Add org.freedesktop.Sdk.Extension.symfony

francoism90:

This would SDK would be addition over https://github.com/flathub/org.freedesktop.Sdk.Extension.php84. The current php Sdk is to limited when working with popular frameworks like Symfony and Laravel.

Both of these popular frameworks need more extensions for IDE completion support, and it would be useful to enable connecting support to your database(s) or sockets, when using these kind of extensions on VSCode.

The current workaround is to either expose your IDE to a `toolbox` by breaking the sandbox, or run it inside a container using `podman remote`. However both are really hard to setup and maintain, and it requires an user to have knowledge of Toolbox or run Podman.

Both SDKs would be useful, this one would be compared to `ffmpeg-full`. Sometimes you need more codecs, in this you would need more extensions. It's still limited to the current Flatpak SDK, but it should offer a good base for any Symfony/Laravel developer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+.flatpak-builder
+.build
+repo
+_build
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
-# org.freedesktop.Sdk.Extension.php84-full
+# org.freedesktop.Sdk.Extension.symfony
 
-This extension adds PHP 8.4 support to Flatpak.
+This extension adds [Symfony](https://symfony.com/) support to Flatpak.
+
+Compared to [org.freedesktop.Sdk.Extension.php84](https://github.com/flathub/org.freedesktop.Sdk.Extension.php84), it offers more extensions but this package is large in size.
 
 PHP installs to `/usr/lib/sdk/php84` inside the sandbox.
 
@@ -52,24 +54,30 @@ To use the [Laravel extension](https://marketplace.visualstudio.com/items?itemNa
 bash-5.0$ php -m
 [PHP Modules]
 apcu
+bcmath
 bz2
 Core
 ctype
 curl
 date
 dom
+exif
 fileinfo
 filter
+gd
+gettext
 hash
 iconv
 intl
 json
 libxml
 mbstring
+mysqlnd
 openssl
 pcntl
 pcre
 PDO
+pdo_mysql
 pdo_sqlite
 Phar
 posix
@@ -78,6 +86,7 @@ redis
 Reflection
 session
 SimpleXML
+sockets
 SPL
 sqlite3
 standard
@@ -86,6 +95,7 @@ xdebug
 xml
 xmlreader
 xmlwriter
+xsl
 Zend OPcache
 zip
 zlib
@@ -98,5 +108,5 @@ Zend OPcache
 ## Build
 
 ```bash
-flatpak run org.flatpak.Builder --force-clean --sandbox --user --install --install-deps-from=flathub --ccache --repo=repo builddir org.freedesktop.Sdk.Extension.php84-full.json
+flatpak run org.flatpak.Builder --force-clean --sandbox --user --install --install-deps-from=flathub --ccache --repo=repo builddir org.freedesktop.Sdk.Extension.symfony.json
 ```
--- a/README.md
+++ b/README.md
@@ -98,5 +98,5 @@ Zend OPcache
 ## Build
 
 ```bash
-flatpak-builder --repo repo .build org.freedesktop.Sdk.Extension.php84.json --force-clean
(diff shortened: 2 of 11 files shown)


francoism90:
I've created a PR (https://github.com/flathub/org.freedesktop.Sdk.Extension.php84/pull/4#issuecomment-2560419002) on `org.freedesktop.Sdk.Extension.php84`, however I disagree with the `docker` and `podman` workarounds.

Both are very limited supported, and because it hooks in your current container, it may cause your container(s) to crash or you may end up with weird behavior (https://github.com/containers/podman/discussions/24898).

This may happen on containers running PHP in memory. Because the IDE (e.g. VSCode) may kill the PHP-process, the container will restart frequently or may even overlap. The only (good) workaround would be running an additional 'dev' container, but must users won't even know how to setup this.

I do understand the `podman remote` solution (didn't tested Docker), but stated before, this is very difficult to setup, and sometimes you don't even have a container instance running. Like I do developing on a Chromebook, which isn't even capable of running any containers. But it does push it upstream to my homelab.

Lctrs:
Still not convinced by the need of this extension...

That being said, I think the name is confusing as it can be used for developing with other frameworks as well. Why not following `ffmpeg` example and naming it `org.freedesktop.Sdk.Extension.php-full` ?

hfiguiere:
why not just adding it all to the PHP extensions?

bbhtt:
> Both SDKs would be useful, this one would be compared to ffmpeg-full.

ffmpeg-full exists because of legal reasons for shipping patented codecs and it is extending something shipped in the runtime by default.

>  Sometimes you need more codecs, in this you would need more extensions.

Not more, if this is installed, then php-84 extension doesn't seem to be of use, since you are shipping a competing copy of PHP too.


Ideally it'd be in the original extension.

francoism90:
Except from 4 extensions, these are all built in PHP extensions. They don't require any additional packages or libs.

@bbhtt These are the original provided from PHP itself. GD for example is, and same for pdo mysql. The only package I've added is Redis. The reason for this, is that it's a very popular cache backend and some frameworks expect it to exists as a class.

The author of the php84 sdk doesn't want to enable more inbuilt extensions. That's why I've created this SDK.

It would be great if Flatpak could somehow allow the customizable of SDKs. Like for node, sometimes you don't need npm, but need yarn or any other package manager.. which isn't something the current SDK allows.

The world doesn't consists of a minimal. People use more languages than C or Rust you know.

Lctrs:
It's not that I don't want to add more extensions, it's just that I'm very surprised that a development tool need to be able to run your project to work properly.

francoism90:
@Lctrs This is needed for IDE completion, it doesn't run the project itself.

But let's say you're on a clean PHP project (without any framework or tools), and you need to connect to a remote MySQL database. When not using flatpaks, you would install php-pdo as a package, which would offer you adapters to connect to most major databases. Since VSCode doesn't run in a sandbox in this example, it's able to use all the extensions you have setup in your main environment.

Now say you want to run PHP in VSCode and connect to a database, you can't. The php binary doesn't have the required extensions to talk to any database using the inbuilt terminal or extensions that allow you to run code (e.g. run as php).

I fully agree about not adding extensions just because it's possible. But like I stated before not everyone has container experienced, and sometimes you don't even want to use it. You just want to run the code directly in your IDE instead.

Again, not saying you shouldn't, but you'd be surprised how many do this, or just need a basic set (like me) that does understand how to complete inbuilt PHP methods and classes.

bbhtt:
> It would be great if Flatpak could somehow allow the customizable of SDKs. Like for node, sometimes you don't need npm, but need yarn or any other package manager.. which isn't something the current SDK allows.

Not possible, it's read-only once built and put to the repo.

francoism90:
@bbhtt I think that's a valid point. :)

One final question about `ffmpeg-ful`:

> ffmpeg-full exists because of legal reasons for shipping patented codecs and it is extending something shipped in the runtime by default.

Do you recommend against something like this (e.g. php-full)? If that's the case, the only thing I can think of are dev-containers (this requires podman and the podman-sdk).

But this still requires a lot of loops and things aren't that complete either https://github.com/flathub/com.visualstudio.code/issues/55#issuecomment-913864265

As you can see this is only for VSCode + Podman, for other apps I don't have any idea.

I'll close this PR for now, as I don't see this becoming a thing.

bbhtt:
I'm not convinced entirely with a second `-full` extension like this as I said. There is no precedent for this other than `ffmpeg-full` which was purely for legal reasons and maintained alongside the runtime sharing the sources and dependencies with the copy in the runtime.


Ideally it should be in the current PHP extensions, if there are concerns with maintenance bandwidth, maybe co-maintenance can be set up.

francoism90:
@bbhtt Just see the trouble of adding your own things: https://github.com/flathub/com.visualstudio.code/issues/55#issuecomment-2561905012

The problem isn't the number of missing extensions in this Sdk, it's flatpak not offering any real solution for this.

PR #5967, opened 2024-12-25
Add io.github.sxyazi.yazi

lihaohong6:

- [X] Please describe the application briefly. A terminal file manger.
- [X] I am an author/developer/upstream contributor to the project. If not, I contacted upstream developers about this submission. Submitted an issue on this: sxyazi/yazi#2080


update_sources.py
bbhtt, line 1: what is the source for this?

io.github.sxyazi.yazi.yml
```diff
@@ -0,0 +1,170 @@
...
+  # fuzzy finder
+  - modules/fzf/fzf.yml
+  # Recursively searching for files in the current directory using a regex
+  - modules/ripgrep/ripgrep.yml
+  # Alternative of find.
+  - modules/fd/fd.yml
+  # For quickly jumping to another directory
+  - modules/zoxide/zoxide.yml
+  # For spawning processes outside the flatpak sandbox
+  - modules/host-spawn/host-spawn.yml
```
bbhtt, line 97: these should be in the main manifest.

bbhtt:
I posted some comments but note that we may not accept this submission. I don't see this as a good fit, there is no sandboxing involved due to the permissions you have. This is equivalent to a host package.

bbhtt:
You can never browse files inside `/usr, /var` etc. from the Flatpak, since the sandbox overrides that directory. You can never perform any privileged operation (common for file managers) from inside the sandbox.

io.github.sxyazi.yazi.yml
```diff
@@ -0,0 +1,170 @@
...
+    add-ld-path: .
+    version: *runtime-version
+    no-autodownload: false
+    autodelete: true
+
+finish-args:
+  # For downloading plugins
+  - --share=network
+  # Use host-spawn to call utilities not packaged in the sandbox
+  - --talk-name=org.freedesktop.Flatpak
```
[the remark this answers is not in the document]
lihaohong6: The package manager (`ya`) needs networking access to download plugins and themes. This is why `git` got packaged as well. 

For `host-spawn`, the main use case is to launch editors on the host (e.g. neovim). There are some other situations such as plugins requiring additional commands to function. 

If any of them is too much permission to grant, I can disable it by default and add a note about it in the README in case someone needs to use `ya` or launch a program outside the sandbox.

io.github.sxyazi.yazi.yml
```diff
@@ -0,0 +1,170 @@
...
+  # File manager needs access to every file
+  - --filesystem=host
+  - --filesystem=home
+  - --filesystem=xdg-config/yazi:create
+  - --filesystem=xdg-data/yazi:create
+  - --filesystem=xdg-data/zoxide:create
```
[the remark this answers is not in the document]
lihaohong6: I'll remove the two entries for Yazi. The `zoxide` directory would be helpful for some users since they might be using zoxide in their shell alongside Yazi. It would be helpful for the zoxide database to be synced across Yazi and the shell.

update_sources.py
lihaohong6: I wrote it myself to avoid the work of copy-pasting the output of `flatpak-builder-tools`. There are some shell scripts in other flatpak repositories that do similar things, but I'm not very familiar with shell scripting, and writing a custom script turned out to be very little work.

io.github.sxyazi.yazi.appdata.xml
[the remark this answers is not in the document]
lihaohong6: I encountered some issues when trying to upstream the metadata file before and am wondering if you could help clarify. The `releases` section is supposed to contain the newest version of the program, which is displayed on flathub. However, if the maintainer forgets to modify it when releasing a new version. The flatpak build would produce outdated version numbers. What would be a good solution for that? I reverted to having the appdata file downstream, which is not ideal.

io.github.sxyazi.yazi.yml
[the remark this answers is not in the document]
bbhtt: > zoxide directory would be helpful for some users since they might be using zoxide in their shell alongside Yazi.

then they can store the zoxide config in `~/.var/app/$app-id/config`. You cannot trust configs from host, the version of zoxide on host may not match with the one shipped in the flatpak.

It's entirely possible a newer version supports a config option that makes older versions error with "unrecognized config" or similar.

io.github.sxyazi.yazi.appdata.xml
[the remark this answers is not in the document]
bbhtt: > The flatpak build would produce outdated version numbers. What would be a good solution for that? I reverted to having the appdata file downstream, which is not ideal.

Depends on the release workflow of the maintainer. Generally people just update it manually before doing the release but you can also set up automation to edit the tags and insert the latest release. But still it'd depend on how the release workflow is carried out.

io.github.sxyazi.yazi.yml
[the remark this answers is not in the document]
bbhtt: I think, not allowing to launch external editors is fine considering some other things don't work with the Flatpak either. People who want it can manually add it.

> There are some other situations such as plugins requiring additional commands to function.

What are these generally? Can these be included in the manifest?

update_sources.py
bbhtt: Probably best to assign a license and (C) header then.

io.github.sxyazi.yazi.yml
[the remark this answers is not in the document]
lihaohong6: I removed both permissions since they are needed only in specific situations. I expanded the README file with instructions on how to enable network and `host-spawn` access when needed.

There is a list of plugins [here](https://yazi-rs.github.io/docs/resources/). A lot of them expand the preview feature with a library, so it would be impractical to package every single one of them. These libraries are also subject to change as more plugins appear. I think packaging whatever the base program needs is sufficient. Plugin users can use workarounds.

io.github.sxyazi.yazi.yml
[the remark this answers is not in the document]
lihaohong6: Good point. I removed these permissions and added a note in the README.

io.github.sxyazi.yazi.appdata.xml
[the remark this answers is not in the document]
lihaohong6: Got it. Thanks! I have created a PR for the screenshot. Once it gets merged upstream, the metadata file will have a permanent upstream url for the screenshot and can be submitted upstream as well.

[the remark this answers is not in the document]
lihaohong6: I found an appdata patch with you involved on https://github.com/flathub/org.mypaint.MyPaint. I'll follow that example for future releases.

io.github.sxyazi.yazi.yml
```diff
@@ -0,0 +1,263 @@
...
+        url: https://github.com/sxyazi/yazi.git
+        tag: v0.4.2
+        commit: 5cfcab305ef0a02771b3bd651ed343e436fc7f7e
+        x-checker-data:
+          type: anitya
+          project-id: 370571
+          stable-only: true
+          tag-template: v$version
+      - type: dir
+        path: .
```
bbhtt, line 262: don't use dir sources, these cannot be cached.

bbhtt:
bot, build io.github.sxyazi.yazi

io.github.sxyazi.yazi.yml
lihaohong6: Done. Not sure why I had them. Probably a blanket copy-paste from another repo.

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "disable-external-data-checker": true
```
hfiguiere, line 2: why not removing the f-e-d-c stuff from the manifest instead?

io.github.sxyazi.yazi.appdata.xml
```diff
@@ -0,0 +1,53 @@
...
+
+  <provides>
+    <binary>yazi</binary>
+    <binary>ya</binary>
+  </provides>
+
+  <screenshots>
+    <screenshot type="default">
+      <!--Does not yet exist but will be created upon acceptance to flathub-->
+      <image>https://raw.githubusercontent.com/flathub/io.github.sxyazi.yazi/master/screenshot.png</image>
```
hfiguiere, line 35: this should NEVER reference the tip of a branch.

io.github.sxyazi.yazi.appdata.xml
```diff
@@ -0,0 +1,53 @@
...
+    </screenshot>
+  </screenshots>
+
+  <!--Just a file manager-->
+  <content_rating type="oars-1.1" />
+
+  <launchable type="desktop-id">io.github.sxyazi.yazi</launchable>
```
hfiguiere, line 42: this should match the .desktop file name

modules/zoxide/zoxide.yml
hfiguiere, line 1: remove this file it's never used.

io.github.sxyazi.yazi.appdata.xml
lihaohong6: It has been changed to a screenshot on my fork of upstream. I'll change it to the upstream version once a PR gets merged.

hfiguiere: also the repo doesn't exist so the appstream validation will fail.

flathub.json
lihaohong6: Could you elaborate on what you meant by f-e-d-c stuff? The reason to disable the check was that rust and go dependencies cannot be automatically upgraded since the versions of all the dependencies of dependencies need to be updated as well.

io.github.sxyazi.yazi.appdata.xml
lihaohong6: I used a wrong link for the image, but the current link (on commit [dc03863](https://github.com/flathub/flathub/pull/5967/commits/dc038639a166220b6d18992d3c3cffe80b8fe13e)) should work.

flathub.json
hfiguiere: why did you put the `x-checker-data`?

lihaohong6: I saw some other manifests that use it to update their dependencies, so I added them in this manifest. Although the auto-PR feature from flathub wouldn't be too useful, being able to check for updates for all dependencies by simply running `flatpak-external-data-checker` is still nice to have, in my opinion.

lihaohong6:
The screenshot is now upstream. It belongs to a dedicated [asset repository](https://github.com/yazi-rs/assets) instead of the main code repository. Hopefully that's fine.

As for the metadata file, would it be okay to have it also hosted in the asset repository and referenced with a commit hash? If the metadata is in the main repository, the maintainer needs to remember to update the `releases` tag before a new release, otherwise a downstream patch is required. Being in a separate repository allows the metadata to receive updates after a new version is released.

bbhtt:
> As for the metadata file, would it be okay to have it also hosted in the asset repository and referenced with a commit hash? 

Yes it is fine, as long as it gets updated with new releases. Otherwise you have to patch it on new release.

lihaohong6:
bot, build io.github.sxyazi.yazi

lihaohong6:
The metadata file has been unstreamed. I believe every issue is addressed?

bbhtt:
Does this require a nerd font to be installed? 

image

lihaohong6:
> Does this require a nerd font to be installed?

Yes. The missing icons show up as empty spots on my machine without nerd-fonts, so I didn't it expect it to be so bad in other configurations. I could either try package nerd-fonts along with the flatpak (haven't tried this yet) or [create a default configuration](https://yazi-rs.github.io/docs/faq/#dont-like-nerd-fonts) that disables nerd-fonts.

lihaohong6:
I did some testing on my end. The font installed at `/app/etc/fonts` is not loaded as a fallback, possibly because I launched `yazi` with Konsole, which only uses fonts available on the host. Looks like some font patching might be necessary, which would be too intrusive as it involves changing configurations on the host. 

Perhaps it'll be easier to disable nerd-fonts by default by creating a config file (if it's not already there)?

bbhtt:
> Perhaps it'll be easier to disable nerd-fonts by default by creating a config file (if it's not already there)?

Yea maybe that's better, better if the user can override this too.

lihaohong6:
> Yea maybe that's better, better if the user can override this too.

I modified the way `yazi` is launched: before starting `yazi`, a shell script will check if `theme.toml` is present in the config directory. If not, a `theme.toml` file that disables nerd-fonts will be created. 

If the users wants to enable nerd-fonts, they can empty the `theme.toml` file. 

This is not an elegant solution. I was looking for a way to create this `theme.toml` file when installing the flatpak instead of checking for it every time before launch, but I didn't manage to find any information. Making modifications to the host's `$HOME` during installation also sounds sketchy in terms of security.

io.github.sxyazi.yazi.yml
```diff
@@ -0,0 +1,275 @@
...
+      - install -Dm00755 target/release/yazi ${FLATPAK_DEST}/bin/yazi
+      # `ya` is a package manager for themes/plugins
+      - install -Dm00755 target/release/ya ${FLATPAK_DEST}/bin/ya
+      # Upstream icon is 1254x1260. Shrink to 512x512 to conform to icon standard. 
+      - ffmpeg -hide_banner -i assets/logo.png -vf scale=512:512 logo.png
+      - install -Dm00644 logo.png ${FLATPAK_DEST}/share/icons/hicolor/512x512/apps/${FLATPAK_ID}.png
+      # The icon specified in the upstream dot desktop file does not play nicely with Flatpak.
+      - desktop-file-edit assets/yazi.desktop --set-key=Icon --set-value="${FLATPAK_ID}"
+      # Run the script to disable nerd fonts before launching Yazi.
+      - desktop-file-edit assets/yazi.desktop --set-key=Exec --set-value="/app/bin/disable-nerd-fonts.sh"
```
bbhtt, line 253: call this yazi-wrapper.sh and it should be just the filename not the full path.

io.github.sxyazi.yazi.yml
```diff
@@ -0,0 +1,275 @@
+id: io.github.sxyazi.yazi
+runtime: org.freedesktop.Platform
+runtime-version: &runtime-version '24.08'
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+  - org.freedesktop.Sdk.Extension.golang
+command: yazi
```
bbhtt, line 8: change this too.

bbhtt:
> This is not an elegant solution. I was looking for a way to create this 

It needs to support a global config directory /usr or /etc where vendors can ship their own config, which would get overridden by `~/.config`

lihaohong6:
Thanks for the suggestion. I checked the docs. It is possible to specify an alternate config directory through an environment variable, but nothing mentions having separate configs of different priorities. The current solution seems to work well enough: I don't notice any performance problems with the extra checks.

bbhtt:
bot, build io.github.sxyazi.yazi

bbhtt:
Please make sure to enable 2fa in github account.

Issue #5388, opened 2024-07-06
Random failures while building the `i386` NVIDIA drivers


guihkx:
The latest `i386` build of `org.freedesktop.Platform.GL.nvidia`, has failed:

- https://buildbot.flathub.org/#/builders/6/builds/132946

These failures seem random and they've happened many times before. I'm not sure what causes them, but I have a suspicion:

It seems like the only buildbot worker able to build the i386 driver successfully is `cncf-x86-4`.

Examples:

- https://buildbot.flathub.org/#/builders/38/builds/20
- https://buildbot.flathub.org/#/builders/38/builds/23
- https://buildbot.flathub.org/#/builders/38/builds/25

Meanwhile, the `cncf-x86-3` worker has failed every time:

- https://buildbot.flathub.org/#/builders/38/builds/21
- https://buildbot.flathub.org/#/builders/38/builds/22
- https://buildbot.flathub.org/#/builders/38/builds/24

But then again, maybe those are just coincidences...

If I may suggest something, perhaps making the Freedesktop 1.6 i386 SDK publicly available on Flathub, would prevent this.

```
$ flatpak remote-info flathub org.freedesktop.Sdk/i386/1.6
error: Error searching remote flathub: Can't find ref org.freedesktop.Sdk/i386/1.6
```

guihkx:
Yes, but so has the 1.6 x86_64 SDK, which is still downloadable through Flathub regardless:

```
$ flatpak remote-info flathub org.freedesktop.Sdk/x86_64/1.6
         ID: org.freedesktop.Sdk
        Ref: runtime/org.freedesktop.Sdk/x86_64/1.6
       Arch: x86_64
     Branch: 1.6
 Collection: org.flathub.Stable
   Download: 717,1 MB
  Installed: 1,9 GB

     Commit: c02e7a722d26f43a5af534cb70b0d61f593c0c12f44d5aa3000db2227a5a90e0
     Parent: d1e0781a699ab6c89e3fad6ad1362f4903834c45df6fb900dbdf4981405ff18c
End-of-life: The Freedesktop 1.6 SDK is no longer supported and does not receive security updates. We strongly recommend migrating to a supported runtime.
    Subject: build of org.freedesktop.Sdk, Tue Nov  6 17:56:02 UTC 2018
       Date: 2018-11-07 00:08:07 +0000
```

guihkx:
[...]

PR #5971, opened 2024-12-28
Add org.freedesktop.Sdk.Extension.nvidia-base

guihkx:

- [x] Please describe the application briefly. *Base extension with pre-built static libraries to be used by the [NVIDIA driver extension for Flatpak](https://github.com/flathub/org.freedesktop.Platform.GL.nvidia) to speed up packaging time.*

---

This base extension contains pre-built static libraries for zlib, liblzma, libzstd and libarchive, meaning we don't have to download and compile these libraries for every version of the NVIDIA driver we package [here](https://github.com/flathub/org.freedesktop.Platform.GL.nvidia) (currently there are 373).

The speed gain is massive: Without this extension, packaging 5 different driver versions on my personal computer (quad-core CPU @ 3.6 GHz) took a whopping 6 minutes.

With the base extension, the same packaging process took only 23 seconds, almost 16 times faster!

Flathub currently takes [about 5 hours](https://buildbot.flathub.org/#/builders/6/builds/170856) to package all 373 NVIDIA drivers, so I expect that figure to also drop greatly if this gets merged.

Please also add @404NetworkError and @TingPing as collaborators.

### DIFF
--- a/org.freedesktop.Sdk.Extension.nvidia-base.metainfo.xml
+++ b/org.freedesktop.Sdk.Extension.nvidia-base.metainfo.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="addon">
+  <id>org.freedesktop.Sdk.Extension.nvidia-base</id>
+  <extends>org.freedesktop.Sdk</extends>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>MIT</project_license>
+  <name>nvidia-base</name>
+  <summary>Package NVIDIA drivers faster</summary>
+  <url type="homepage">https://github.com/flathub/org.freedesktop.Sdk.Extension.nvidia-base</url>
+  <description>
+    <p>
+      Base extension with pre-built static libraries to be used by the NVIDIA driver extension for Flatpak (https://github.com/flathub/org.freedesktop.Platform.GL.nvidia) to speed up packaging time.
+    </p>
+  </description>
+</component>
--- a/org.freedesktop.Sdk.Extension.nvidia-base.yaml
+++ b/org.freedesktop.Sdk.Extension.nvidia-base.yaml
@@ -1,0 +1,103 @@
+id: org.freedesktop.Sdk.Extension.nvidia-base
+runtime: org.freedesktop.Sdk
+runtime-version: '24.08'
+build-extension: true
+sdk: org.freedesktop.Sdk
+separate-locales: false
+
+build-options:
+  prefix: /usr/lib/sdk/nvidia-base
+
+cleanup:
+  - /bin
+  - /include
+  - /share
+
+modules:
+  - name: zlib
+    config-opts:
+      - --static
+    sources:
+      - type: archive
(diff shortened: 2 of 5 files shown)


guihkx:
bot, build org.freedesktop.Sdk.Extension.nvidia-base

guihkx:
The build actually passed (the failure was caused by `flat-manager-client` while committing the build).

In any case, I'll wait for reviews before triggering a new build.

bbhtt:
Fwiw, for transparency, I suggested doing this extension in https://github.com/flathub/flathub/issues/5388

org.freedesktop.Sdk.Extension.nvidia-base.yaml
```diff
@@ -0,0 +1,104 @@
+id: org.freedesktop.Sdk.Extension.nvidia-base
+branch: '24.08'
```
bbhtt, line 2: Does this need to be branched? You aren't going to maintain multiple branches for Nvidia, it will follow only one runtime version and so will this extension.

guihkx: Yeah, that makes sense. I had just copied this part from some other extension and didn't really think about it. I've removed it now.

org.freedesktop.Sdk.Extension.nvidia-base.metainfo.xml
```diff
@@ -0,0 +1,20 @@
...
+  <project_license>MIT</project_license>
+  <name>nvidia-base</name>
+  <summary>Package NVIDIA drivers faster</summary>
+  <url type="homepage">https://github.com/flathub/org.freedesktop.Sdk.Extension.nvidia-base</url>
+  <description>
+    <p>
+      Base extension with pre-built static libraries to be used by the NVIDIA driver extension for Flatpak (https://github.com/flathub/org.freedesktop.Platform.GL.nvidia) to speed up packaging time.
+    </p>
+  </description>
+  <releases>
```
bbhtt, line 15: Release doesn't make sense here, I'd drop this it's not mandatory for extensions.

guihkx: Great, I didn't know that. Removed it as well.

bbhtt:
bot, build org.freedesktop.Sdk.Extension.nvidia-base

bbhtt:
Is the final thing working now?

guihkx:
Yup. Once this gets merged and published I'll update org.freedesktop.Platform.GL.nvidia, but I've already packaged  some drivers locally and it works.

bbhtt:
Ok will merge next week, there are some infra things that need to be dealt with before merging new submissions.

bbhtt:
Please make sure to enable 2FA for your github account.

/merge @404NetworkError @TingPing

org.freedesktop.Sdk.Extension.nvidia-base.yaml
bbhtt: I forgot that this extension point needs to be branched anyway as that is how it is defined in the SDK. I merged it to `branch/24.08`. This was redundant in either case.

PR #5975, opened 2024-12-31
Add com.vargary.focal

vargab95:

- [x] Please describe the application briefly. Focal is a tree based note taking and task management app, currently available on android and planned to be released on all major platforms.


flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64"]
```

com.vargary.focal.png
hfiguiere, line 1: this should ship with the upstream distribution

com.vargary.focal.metainfo.xml
hfiguiere, line 1: this should ship with the upstream distribution

com.vargary.focal.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Type=Application
+
+Name=Focal
+Categories=Utility;
+
+Icon=com.vargary.focal
+Exec=/app/focal
```
hfiguiere, line 8: this should NOT be a full path. This should just be the name of the executable and the executable, of course, should be in the path, ie in `/app/bin`

com.vargary.focal.yaml
```diff
@@ -0,0 +1,35 @@
+app-id: com.vargary.focal
+runtime: org.freedesktop.Platform
+runtime-version: "23.08"
+sdk: org.freedesktop.Sdk
+command: /app/focal
```
hfiguiere, line 5: this should NOT be a full path.

com.vargary.focal.desktop
hfiguiere, line 1: this should ship with the upstream distribution

com.vargary.focal.metainfo.xml
```diff
@@ -0,0 +1,85 @@
...
+  <summary>Organize tasks and notes in a tree structure for focus and productivity.</summary>
+
+  <developer id="com.vargary">
+    <name>Balázs János Varga</name>
+  </developer>
+
+  <url type="homepage">https://focal.vargary.com/</url>
+
+  <metadata_license>CC-BY-4.0</metadata_license>
+  <project_license>LicenseRef-proprietary</project_license>
```
hfiguiere, line 15: what is the actual license? Is redistribution allowed?

com.vargary.focal.yaml
```diff
@@ -0,0 +1,35 @@
+app-id: com.vargary.focal
+runtime: org.freedesktop.Platform
+runtime-version: "23.08"
```
hfiguiere, line 3: current version is 24.08

com.vargary.focal.yaml
```diff
@@ -0,0 +1,35 @@
...
+sdk: org.freedesktop.Sdk
+command: /app/focal
+tags:
+  - proprietary
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --socket=pulseaudio
```
bbhtt, line 13: why is network and sound needed for note taking app?

flathub.json
vargab95: Why? It is written in flutter which supports linux-x64 and linux-arm64. However, x64 to arm cross compilation is currently not supported, so that's more or less the only arch what I can support.

com.vargary.focal.yaml
vargab95: Thanks for the finding, I'll update it in my next commit.

com.vargary.focal.yaml
vargab95: Network is added for crash reports and in the next version, I plan to deliver a synchronization feature which will be used to synchronize across different devices. However, the sound is just a copy-paste error, thanks for pointing it out. I'll remove it in the next commit.

com.vargary.focal.desktop
vargab95: Thanks for the finding, I'll fix it in my next commit.

com.vargary.focal.png
vargab95: Thanks for the finding, I'll fix it in my next commit.

com.vargary.focal.metainfo.xml
vargab95: Thanks for the finding, I'll fix it in my next commit.

com.vargary.focal.desktop
vargab95: Thanks for the finding, I'll fix it in my next commit.

com.vargary.focal.yaml
vargab95: Thanks for the finding, I'll fix it in my next commit.

flathub.json
vargab95: As I see, other flutter apps have the same config. Like https://github.com/flathub/de.wger.flutter/blob/master/flathub.json and https://github.com/flathub/com.maplerain.telsis_translator_flutter/blob/master/flathub.json.

com.vargary.focal.desktop
vargab95: Fixed in commit [bb74d01](https://github.com/flathub/flathub/pull/5975/commits/bb74d019e1001aac8ee6baa1856e6a8c0a821c1b).

com.vargary.focal.yaml
vargab95: Fixed in the new [commit](https://github.com/flathub/flathub/pull/5975/commits/bb74d019e1001aac8ee6baa1856e6a8c0a821c1b).

com.vargary.focal.yaml
vargab95: Fixed in the new [commit](https://github.com/flathub/flathub/pull/5975/commits/bb74d019e1001aac8ee6baa1856e6a8c0a821c1b).

com.vargary.focal.metainfo.xml
vargab95: There is no license yet. I've only distributed apps on google play so far and there wasn't necessary to provide a license. So when I wrote the metainfo.xml, then I just searched for another closed source app on flathub and copied the line from there. (https://github.com/flathub/com.hamrick.VueScan/blob/de29c398416a715f71f3549f71e470aacf086e8f/com.hamrick.VueScan.appdata.xml#L27)

hfiguiere: two wrong don't make a right. This is the package we are reviewing, not the other one.

flathub.json
hfiguiere: 1. I don't see what "cross compilation" has anything to do here.
2. So "because the other do it wrong" is your mantra? 

I thought that the use of Flutter was "because it's cross platform". Yet it can't even manage the same platform in a different architecture?

vargab95: > I thought that the use of Flutter was "because it's cross platform". Yet it can't even manage the same platform in a different architecture?

I can build it on linux-x64 for android and linux-x64. That's all. I have to spin up a windows virtual machine to have the windows build and a mac to have the mac + ios ones. I also don't like it, but that's the case.

> I don't see what "cross compilation" has anything to do here.

If target platform options are linux-x64 and linux-arm64 and you would only have an x64 machine, then how would you define the process of compiling a project to arm64 on your x64 machine? According to my prior knowledge and google search results, it's called cross compilation...

> So "because the other do it wrong" is your mantra?

It's hard to call something a mantra which was stated only once... Anyway, the explanation is that, I could not find any description in the flutter documentation about how to deploy to flathub, I also could not find a flutter specific guide in the flatpak documentation, so the next best option was to try to figure it out based on the documentation and other examples. Maybe you're some kind of dev god, but mere mortals try to learn from others while doing new stuff and there is no official documentation. Due to this and because I could not compile to any other arch on my linux machine, it seemed to be the correct way.

com.vargary.focal.metainfo.xml
vargab95: Yeah, but how can I solve it in any other way? The flathub docs have only a few lines about non-FOSS licenses. [This](https://docs.flathub.org/docs/for-app-authors/requirements#license) page states, that it must allow legal redistribution and must be correctly specified in the app's MetaInfo file. And [this](https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/#license) page states that "Proprietary applications/custom licenses can use LicenseRef-proprietary with a link to the license". It does not help a lot. I'm just a software dev and have no idea what such a file should exactly contain. In my open source projects it's quite straightforward, I can use MIT or GPL or so. However, in this case, I have no idea. I guess if I would just give a link to something where I write a sentence or two (which can be completely incorrect from a legal standpoint) would not help so much. That's why I was glad that I found another example in an already accepted PR...

bbhtt:
bot, build com.vargary.focal

com.vargary.focal.metainfo.xml
bbhtt: So the problem is you are asking "how to write a proprietary license"?

vargab95: How to write one / link one which is acceptable by flathub. As I see other projects are including this without a link, like Zoom and Steam. Some refers to their EULA, like Spotify and Synology. Is it better if I put the terms & conditions link there or to use it without a link or any third option?

bbhtt: There is no regulation for that. Since you are the default copyright holder for your work, any terms and conditions you state or impose is acceptable.

It all depends on how you want your work to be used / reused. We don't have a say over that. That's upto you.

vargab95: Thanks for the explanation. I've updated the terms & conditions to also fit for desktop applications, added the link to the metainfo.xml and built a new package. The latest commit contains the hash for the new package. With this, I think this finding is resolved.

bbhtt:
bot, build com.vargary.focal

bbhtt:
Please make sure to enable 2FA for your github account.

/merge

PR #5989, opened 2025-01-04
Add org.freedesktop.Sdk.Extension.rustup-stable-minimal

JanKoudijs:

**This is an (alternative) Rust SDK extension that provides a rustup installation for build scenarios that require this. See the README of this PR for more details.**
- [ ] The domain used for the application ID is [controlled by the application developer(s)][appid-domain] and the [application id guidelines][appid] are followed.
**Being a Freedesktop SDK extension the ID is derived from that.**
- [ ] I am an author/developer/upstream contributor to the project. If not, I contacted upstream developers about this submission. **Link:**
**rustup, cargo and rustc use the Apache-2.0/MIT license**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+.flatpak-builder/
+repo/
--- a/README.md
+++ b/README.md
@@ -1,0 +1,14 @@
+# org.freedesktop.Sdk.Extension.rustup-stable-minimal
+
+This freedesktop SDK extension contains a rustup installation from the `stable` channel using the `minimal` profile. Rustup with profile minimal includes cargo and the rustc compiler.
+
+This extension exists for those use cases were a rustup installation is needed at build time.
+Tools like [Cargokit](https://github.com/irondash/cargokit) and [flutter_rust_bridge](https://github.com/fzyzcjy/flutter_rust_bridge) use the Rust tools via rustup and fail to build without it.
+
+For use cases that do not depend on rustup there is also the [org.freedesktop.Sdk.Extension.rust-stable
+](https://github.com/flathub/org.freedesktop.Sdk.Extension.rust-stable) SDK extension. It has an [explicit statement](https://github.com/flathub/org.freedesktop.Sdk.Extension.rust-stable/issues/4) that rustup is out of its scope.
+
+The read-only nature of SDK extensions limits the rustup features that can be used at application build time. The stable toolchain for the architecture is included, but there is no possibility to perform a `rustup toolchain install`. Additionally, updates have to be done by updating the extension, a `rustup self update` will fail.
+
+## Automation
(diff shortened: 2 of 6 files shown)


JanKoudijs:
The merging of this PR will allow me to build my [Community Remote](https://flathub.org/apps/com.theappgineer.community_remote) app from source. This build depends on rustup and flutter. See this [draft PR](https://github.com/flathub/com.theappgineer.community_remote/pull/14) for the app build manifest.

The flutter SDK is currently an included module, but can potentially be converted into an SDK extension as well.

bbhtt:
This should be inside the rust extension, there is no need for a separate rustup extension.

JanKoudijs:
As stated in the README, the existing rust extension is build-up in a different way, doesn't include rustup, and states it has [no plans to do so](https://github.com/flathub/org.freedesktop.Sdk.Extension.rust-stable/issues/4). Technically the rustup approach can be extended to support all the tools included in the current one, and become the new rust extension. But as a start I included the minimal set to allow building rust while limiting maintenance.

Do you see possibilities in converting `org.freedesktop.Sdk.Extension.rust-stable` to the rustup approach I took?

bbhtt:
rust-std, cargo, rustc are already in the rust extension and these are quite big. I don't see any reason for a second extension.

why do you need rustup? The main purpose is to install a rust toolchain channel on the system which is already being provided by the extension.

JanKoudijs:
My app uses a code generator that relies on rustup, without it code generation fails with the message that rust is not found and needs to be installed using rustup:

> Alright. So what's the catch here?
Well, there is one. It's small, but it's there: The developer using your plugin (not the end user of course) needs to have Rust installed. Through [rustup](https://rustup.rs/).

Source: https://matejknopp.com/post/flutter_plugin_in_rust_with_no_prebuilt_binaries/ 

I can also solve this by having this as a module in my app manifest, but thought it would be a good idea to give this back to the Flathub community in the form of an extension.

bbhtt:
Any reason it specifically requires rustup? Does it not work with the other rust extension? What's the issue here?

hfiguiere:
> My app uses a code generator that relies on rustup, without it code generation fails with the message that rust is not found and needs to be installed using rustup:

Then provide it with your package.
SDK extension are not a repository of bad ideas.

JanKoudijs:
> Then provide it with your package.
SDK extension are not a repository of bad ideas.

Yes, this is a workaround, but one resulting from the enforced restriction that Flatpak builds have to be offline. In my opinion also [flatpak-cargo-generator](https://github.com/flatpak/flatpak-builder-tools/tree/master/cargo) is a workaround in this category.

It is really not helpful that when you apply an app that uses prebuilt binaries, first thing mentioned is "why is it not build from source?" and if you try to come up with solutions these are considered bad ideas.

JanKoudijs:
The tool requiring rustup is a Flutter add-on, and probably uses it to download the native toolchain for the host OS and architecture (Flutter being multiplatform). My proposed solution only bundles the toolchain for `x86_64-unknown-linux-gnu` or `aarch64-unknown-linux-gnu`, depending on the host.

But it is clear that there is no interest, I will not bother you with a Flutter extension for offline builds.

hfiguiere:
nothing stops you from shipping this in the package you want to make. Or fix things so that this is not required.

PR #5995, opened 2025-01-05
Add io.github.dzheremi2.Chronograph

Dzheremi2:

Syncing songs lyrics using timestamps and exporting it to LRCLib

### DIFF
--- a/io.github.dzheremi2.Chronograph.yaml
+++ b/io.github.dzheremi2.Chronograph.yaml
@@ -1,0 +1,105 @@
+id: io.github.dzheremi2.Chronograph
+runtime: org.gnome.Platform
+runtime-version: "47"
+sdk: org.gnome.Sdk
+command: chronograph
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=fallback-x11
+  - --device=dri
+  - --socket=wayland
+  - --socket=pulseaudio
+  - --filesystem=host
+  - --filesystem=xdg-run/gvfsd
+cleanup:
+  - /include
+  - /lib/pkgconfig
+  - /man
+  - /share/doc
+  - /share/gtk-doc
+  - /share/man
+  - /share/pkgconfig
+  - "*.la"
+  - "*.a"
+modules:
+  - name: python3-modules
+    buildsystem: simple
+    build-commands: []
+    modules:
+      - name: python3-mutagen
+        buildsystem: simple
+        build-commands:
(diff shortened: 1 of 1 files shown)


Dzheremi2:
bot, build io.github.dzheremi2.Chronograph

hfiguiere:
why does this need a rename?

Dzheremi2:
> why does this need a rename?

Because I'm going to apply for GNOME Circle participance and, in general, to better follow App Naming Guidelines

bbhtt:
ID is not the app name, don't see why naming guidelines would apply to the ID.

Dzheremi2:
> ID is not the app name, don't see why naming guidelines would apply to the ID.

It wouldn't. I just think that it will be weird, if the app name will differ from the ID. It's just less user friendly, especially when using a terminal.

Dzheremi2:
bot, build io.github.dzheremi2.Chronograph

bbhtt:
> > ID is not the app name, don't see why naming guidelines would apply to the ID.
> 
> It wouldn't. I just think that it will be weird, if the app name will differ from the ID. It's just less user friendly, especially when using a terminal.

It's not weird, a bunch of apps within gnome have the ID not matching the visible name, epiphany, nautilus, loupe etc. The entrypoint for the app is the desktop file, people launching through terminal should have the ability to figure it out.

Dzheremi2:
> > > ID is not the app name, don't see why naming guidelines would apply to the ID.
> > 
> > It wouldn't. I just think that it will be weird, if the app name will differ from the ID. It's just less user friendly, especially when using a terminal.
> 
> It's not weird, a bunch of apps within gnome have the ID not matching the visible name, epiphany, nautilus, loupe etc. The entrypoint for the app is the desktop file, people launching through terminal should have the ability to figure it out.

 Why don't you want to just merge? Maybe I don't understand something, but is changing app id that "destructive" so you don't want to merge it?

bbhtt:
we aren't obligated to merge every arbitrary rename you make.

Dzheremi2:
But this rename isn't arbitrary. I've already explained why I want to change ID. The app was fully rewritten, all from zero. So I want to make the new id — the new life for the app, so to speak.

bbhtt:
There is no reason to change the ID when you rewrite it or decide to submit to some curated collection like GNOME circle. The ID is supposed to be constant.

Specifically here you are only changing the last component, which does not matter.

Dzheremi2:
Okayy...

PR #6012, opened 2025-01-12
Add org.srb2.srb2-legacy

P-AS:

SRB2 Legacy is an updated fork of [Sonic Robo Blast 2](https://www.srb2.org/) version 2.1 that includes bugfixes and QOL improvements from future releases. This submission is an adaption of the manifest [from upstream SRB2](https://github.com/flathub/org.srb2.SRB2).
The source code is hosted at https://git.do.srb2.org/PAS/srb2-legacy. Other forks of SRB2 that are hosted there also use the `org.srb2` components for flathub.
- [X] I am an author/developer/upstream contributor to the project. If not, I contacted upstream developers about this submission. **Link:** https://git.do.srb2.org/PAS/srb2-legacy


bbhtt:
Why does this need a legacy version? Creating a submission for "legacy" every time they release a major update is not sustainable.

org.srb2.srb2-legacy.metainfo.xml
```diff
@@ -0,0 +1,54 @@
...
+    <p>
+      SRB2 Legacy is an updated fork of Sonic Robo Blast 2 2.1.25.
+      The goal of SRB2 Legacy is to include essential fixes and QOL
+      improvements seen in 2.2.
+    </p>
+  </description>
+  <launchable type="desktop-id">org.srb2.srb2-legacy.desktop</launchable>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://file.garden/ZuUiCPHqjV5ssvQM/metadata/tiltescreen.png</image>
```
bbhtt, line 19: If this is a temporary filehost, then don't use it.

I also can't reach https://file.garden

org.srb2.srb2-legacy.metainfo.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
bbhtt, line 1: This file should be upstream.

org.srb2.srb2-legacy.yaml
```diff
@@ -0,0 +1,97 @@
...
+
+  - name: libopenmpt
+    config-opts:
+      - --disable-examples
+      - --disable-openmpt123
+      - --disable-static
+      - --disable-tests
+      - --without-portaudio
+      - --without-portaudiocpp
+    cleanup:
```
bbhtt, line 38: All these cleanups should be in top level, no need to repeat it for every modile

org.srb2.srb2-legacy.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - name: srb2legacy
+    buildsystem: simple
+    build-options:
+      arch:
+        aarch64:
+          env:
+            ARCH_MAKE_ARGS: LINUX64=1
+        x86_64:
+          env:
+            ARCH_MAKE_ARGS: LINUX64=1
```
hfiguiere, line 63:
```suggestion
    build-options:
      env:
        ARCH_MAKE_ARGS: LINUX64=1
```

hfiguiere: or maybe just set it on line 65.

org.srb2.srb2-legacy.metainfo.xml
P-AS: It is a permanent filehost. The images are uploaded to my account. I'm not sure why you are unable to reach it. I could use github as an alternative, but then the links would look like this: https://github.com/user-attachments/assets/8eb891b7-47d6-4d02-8daa-1f639a5ef870. Would it be better to switch it to github images?

P-AS:
> Why does this need a legacy version? Creating a submission for "legacy" every time they release a major update is not sustainable.

The main reason for 2.1 to have one is that the content is almost completely different. All the music, levels, sprites were remade in 2.2. The content is distinct enough to warrant having a separate version. I don't know if that will still be true for future releases, though.

org.srb2.srb2-legacy.metainfo.xml
P-AS: I have decided to move them to https://git.do.srb2.org/PAS/srb2-legacy/-/wikis/Screenshots. That way you should be able to access them, and it's directly connected to the git.do.srb2.org repo.

P-AS:
I have:
- Moved the metainfo file upstream
- Moved the cleanup to the top level
- Moved `LINUX64=1` directly to the `make` command
- Moved the screenshots off filegarden to https://git.do.srb2.org/PAS/srb2-legacy/-/wikis/Screenshots
I hope this is valid now.

bbhtt:
> The main reason for 2.1 to have one is that the content is almost completely different. All the music, levels, sprites were remade in 2.2. The content is distinct enough to warrant having a separate version. I don't know if that will still be true for future releases, though.

This seems a bit unsustainable if they do another set of major changes and someone wants to make a another Flatpak.

It would be valid if they broke game data or config file compatibility again under very specific circumstances.

bbhtt:
bot, build org.srb2.srb2-legacy

PR #6036, opened 2025-01-17
Add io.github.moonlight_mod.moonlight-installer

slonkazoid:

A desktop application to install [moonlight](https://github.com/moonlight-mod/moonlight).
- [x] I am an author/developer/upstream contributor to the project. If not, I contacted upstream developers about this submission. **Link:** https://github.com/moonlight-mod/moonlight-installer


metainfo.xml
hfiguiere, line 1: this should be (submitted) upstream

hfiguiere:
Why an installer instead of the actual thing?

slonkazoid:
> Why an installer instead of the actual thing?

the 'installer' patches Discord with the moonlight mod

io.github.moonlight_mod.moonlight-installer.yml
```diff
@@ -0,0 +1,43 @@
...
+      - install -Dm644 installer/metainfo.xml /app/share/metainfo/${FLATPAK_ID}.metainfo.xml
+      - install -Dm644 installer/moonlight-installer.desktop /app/share/applications/${FLATPAK_ID}.desktop
+      - install -Dm644 icon.png /app/share/icons/hicolor/256x256/apps/${FLATPAK_ID}.png
+    sources:
+      - type: git
+        commit: 22d2bd29f984a0e6de15fe040ee91c73aaba88a7
+        url: https://github.com/moonlight-mod/moonlight-installer.git
+        dest: installer
+      - type: file
+        url: https://raw.githubusercontent.com/moonlight-mod/moonlight-mod.github.io/a66607f32b8130de63177d5dfa748e86f5166af6/public/favicon.png
```
hfiguiere, line 40: why isn't the icon with the source code?

hfiguiere:
> > Why an installer instead of the actual thing?
> 
> the 'installer' patches Discord with the moonlight mod, it is not a standalone app

Then why not shipping an patched version ?

slonkazoid:
> > > Why an installer instead of the actual thing?
> > 
> > 
> > the 'installer' patches Discord with the moonlight mod, it is not a standalone app
> 
> Then why not shipping an patched version ?

That would be a logistical nightmare, and also illegal I think, as Discord is a proprietary application

io.github.moonlight_mod.moonlight-installer.yml
slonkazoid: it's in the repo where public assets are stored. so, still part of the upstream, but i can move it into the source code repo if needed

io.github.moonlight_mod.moonlight-installer.yml
```diff
@@ -0,0 +1,43 @@
...
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --filesystem=xdg-data/flatpak:rw
+  - --filesystem=xdg-data/Discord:rw
+  - --filesystem=xdg-data/DiscordPTB:rw
+  - --filesystem=xdg-data/DiscordCanary:rw
+  - --filesystem=xdg-data/DiscordDevelopment:rw
+  - --filesystem=xdg-config/moonlight-mod:rw
```
bbhtt, line 19: why?

bbhtt:
> That would be a logistical nightmare, and also illegal I think, as Discord is a proprietary application.

how "legal" is the current thing, considering you are patching it anyways on end user machine?

io.github.moonlight_mod.moonlight-installer.yml
```diff
@@ -0,0 +1,43 @@
...
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+command: moonlight-installer
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --filesystem=xdg-data/flatpak:rw
```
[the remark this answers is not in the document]
slonkazoid: explained in https://github.com/flathub-infra/flatpak-builder-lint/pull/563

io.github.moonlight_mod.moonlight-installer.yml
slonkazoid: those are the discord installation paths that it can patch
as for the config, thats where moonlight is stored

io.github.moonlight_mod.moonlight-installer.yml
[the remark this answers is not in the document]
bbhtt: doesn't seem very safe if you can access other Flatpak's data.

slonkazoid:
> > That would be a logistical nightmare, and also illegal I think, as Discord is a proprietary application.
> 
> how "legal" is the current thing, considering you are patching it anyways on end user machine?

Surely more legal than [Vesktop](https://flathub.org/apps/dev.vencord.Vesktop). Not a lawyer but as far as I know, an end user modifying programs is always legal under EU jurisdiction.

> Also an installer does not qualify as an app submission.

It's an application that extends existing Discord installations, not an installer of another application really. It's more of a patcher but calling it that is kind of unfriendly.

If stuff like [Grapejuice](https://github.com/flathub/net.brinkervii.grapejuice), which is just a fancy Roblox installer, was allowed, this should be too.

io.github.moonlight_mod.moonlight-installer.yml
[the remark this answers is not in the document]
NotNite: It's probably possible to restrict this to flatpak/overrides, no?

hfiguiere:
> If stuff like [Grapejuice](https://github.com/flathub/net.brinkervii.grapejuice), which is just a fancy Roblox installer, was allowed, this should be too.

Sometime mistake are made as shall not be repeated. Or two wrong don't make a right. Flathub has gotten more strict over time so thing that used to be accepted may no longer be. "they already do it" is usually a negative argument.

io.github.moonlight_mod.moonlight-installer.yml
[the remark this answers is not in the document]
bbhtt: yes restrict it to `overrides`

io.github.moonlight_mod.moonlight-installer.yml
bbhtt: I'm asking about `xdg-config/moonlight-mod`.

io.github.moonlight_mod.moonlight-installer.yml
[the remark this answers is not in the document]
bbhtt: ok this was done already.

io.github.moonlight_mod.moonlight-installer.yml
slonkazoid: the files put in there must be accessible between

- all instances of moonlight-installer GUI, flatpak or not
- all instances of Discord, flatpak or not
- the moonlight-installer CLI (not included in this package)

as it includes configuration and the moonlight-mod injector

bbhtt: > all instances of moonlight-installer GUI, flatpak or not

Why? The Flatpak gets it's own xdg-config inside `~/.var/app/$app-id/config`. Why do you want to share config with an host instance of moonlight?

> the moonlight-installer CLI (not included in this package)

Why not include this?

slonkazoid: > Why? The Flatpak gets it's own xdg-config inside `~/.var/app/$app-id/config`. Why do you want to share config with an host instance of moonlight?

that's where the moonlight modification itself, along with the channel information for the moonlight updater is stored. if a patched Discord instance can not access it, it will not boot.  
it should ideally be xdg-data/moonlight-mod but that'd be a breaking change for existing installations and upstream has no mechanism for migration.

> Why not include this?

it's not distributed in any way unlike the GUI and intended to be built from source by whoever wants to use it. it also wouldn't be too useful in a flatpak as it doesn't even support patching flatpak installations of Discord anyways

bbhtt:
bot, build io.github.moonlight_mod.moonlight-installer

NotNite:
Build runs! :tada: There seems to be an issue with our code where it's not detecting Flatpak installs, give us a bit to figure out what's up with that...

NotNite:
Okay, we've figured it out - this seems to be some undefined behavior with Flatpak. To quote a community member (@maddymeows):

> when flatpak is mounting into
> the xdg-data-home
> it runs mkdir
> it doesn't create parents
> so
> mkdir -p ~/.var/app/io.github.moonlight_mod.moonlight-installer/data/flatpak/app

This is why us mounting the entire Flatpak directory worked, but mounting specifically Discord's data doesn't. Is there a way we can work around this?

bbhtt:
I'm not sure what you are trying to do but Flatpak does bind mounts.

NotNite:
The issue lies in the bind mounts - we mount `xdg-data/flatpak/app/com.discordapp.Discord`, but it doesn't make that mount recursively. If `xdg-data/flatpak` or `xdg-data/flatpak/app` doesn't exist, the mount fails. If you make the directory yourself (see the `mkdir` above), it mounts fine, because the parent directory now exists.

This is why our mount for `xdg-data/flatpak` worked - the parent directory already existed, so it mounted fine. This also explains why we had some trouble catching it, because it was not immediately apparent what was wrong unless you deleted all data for the Flatpak and retried it - the directory would exist when testing, which is why it worked on @slonkazoid's machine and not on my machine.

That leaves us the options of:

1. Grant `xdg-data/flatpak` - much more permissions than it needs! - to avoid this issue.
2. Ship this currently (non-Flatpak Discord patching works & Flatpak Discord patching works after making the directory) and document the issue somewhere.
3. Upstream a fix somewhere (Flatpak? bwrap?) to recursively create bind mounts. A community member is looking into this option at the moment.

bbhtt:
Did you try `:create`? It seems obvious to me that if the first one host doesn't exist it will fail to mount, because it is done at runtime.

slonkazoid:
> Did you try `:create`? It seems obvious to me that if the directories on host doesn't exist it will fail to mount, because it is done at runtime.

I have, and it doesn't work either. I'm assuming it's because the parent directory doesn't exist either (`$XDG_DATA_HOME/flatpak` in this case).

Granting `xdg-data/flatpak/overrides` is a sandbox escape anyways, so I do not see the harm in granting just `xdg-data/flatpak` in this case. We will make the permissions more strict once this issue is fixed.

bbhtt:
To be honest I think it should not have access to the overrides directory either. That alone is enough to mess with permissions of other Flatpaked apps.

bbhtt:
> I have, and it doesn't work either. I'm assuming it's because the parent directory doesn't exist either ($XDG_DATA_HOME/flatpak in this case).

It seems to work fine here, I renamed my flatpak folder and added a `--filesystem=xdg-data/flatpak/test/foo:create` to a random app and it creates a new `xdg-data/flatpak` folder which for me is on `~/.local/share/flatpak` on host.

NotNite:
I agree, but I don't think there's a better way to scope it unfortunately - we need to allow Discord to write config files for moonlight. If there's a way to scope it to specific files in that directory, we could maybe use that (but it might run into the same issue we're having with the directory not existing in the target sandbox).

bbhtt:
You could just ask the user to run a command and show a dialogue box instead. Considering they are using a patcher/patched client it shouldn't be a big ask.

maddymeows:
> It seems to work fine here, I renamed my flatpak folder and added a `--filesystem=xdg-data/flatpak/test/foo:create` to a random app and it creates a new `xdg-data/flatpak` folder which for me is on `~/.local/share/flatpak` on host.

The problem right now is that `xdg-data/flatpak/test/foo` wouldn't appear in the app's container, just on the host. We need it available in the app's container at its xdg-data, not just on the host, and it only bind mounts those into the container if the parents of those folders exist (e.g. `~/.var/app/.../data/flatpak/test` needs to exist before `~/.var/app/.../data/flatpak/test/foo` mounts correctly).

bbhtt:
Ok that seems like a bug if it appears on host but not in the sandbox.

NotNite:
> Ok that seems like a bug if it appears on host but not in the sandbox.

PR open: https://github.com/flatpak/flatpak/pull/6084

> You could just ask the user to run a command and show a dialogue box with copy/paste instead. Considering they are using a patcher/patched client it shouldn't be a big ask.

I agree that this wouldn't be that hard for users to do, but it's bad UX and requires us to add new logic for it. Having the entire overrides exposed is a bad and rare thing, but there's not much we can do to restrict it further (we can't mount/create individual files tmk).

I kindly request that we can keep this override - removing it would make it inconvenient for users, tedious for us to implement, and the only attack vector is [changing it in the source](https://github.com/moonlight-mod/moonlight-installer/blob/08e47434e8163c616f3fb266177a3f61ee2902e6/crates/libmoonlight/src/installer.rs#L290-L293) and getting a new build published (but I know Flathub maintainers probably aren't auditing every line of code, so maybe this point is somewhat invalid).

bbhtt:
Ok, so is this working minus the edge case of not exposing the directories in the sandbox?

NotNite:
Yes, and that can be fixed with a `mkdir` (we'll add documentation for it) or with that Flatpak PR linked above.

bbhtt:
Ok I patched the Discord installation from Flathub but it fails to launch now:

```
Disabling updates already done
A JavaScript error occurred in the main process
Uncaught Exception:
Error: Cannot find module '/home/test/.var/app/com.discordapp.Discord/config/moonlight-mod/dist/injector.js'
Require stack:
- /app/discord/resources/app/injector.js
- 
    at Module._resolveFilename (node:internal/modules/cjs/loader:1152:15)
    at s._resolveFilename (node:electron/js2c/browser_init:2:124611)
    at Module._load (node:internal/modules/cjs/loader:993:27)
    at c._load (node:electron/js2c/node_init:2:17025)
    at Module.require (node:internal/modules/cjs/loader:1240:19)
    at require (node:internal/modules/helpers:179:18)
    at Object.<anonymous> (/app/discord/resources/app/injector.js:30:1)
    at Module._compile (node:internal/modules/cjs/loader:1373:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1432:10)
    at Module.load (node:internal/modules/cjs/loader:1215:32)
Gtk-Message: 09:46:13.882: Failed to load module "canberra-gtk-module"
Gtk-Message: 09:46:13.882: Failed to load module "pk-gtk-module"
Gtk-Message: 09:46:13.883: Failed to load module "canberra-gtk-module"
Gtk-Message: 09:46:13.883: Failed to load module "pk-gtk-module"
```

bbhtt:
one injector.js is in `/home/test/.local/share/flatpak/app/com.discordapp.Discord/x86_64/stable/active/files/discord/resources/app`

slonkazoid:
i'm confused on how you got it to detect Discord as it should only be able to do that if:
- using flatpak with maddy's PR applied
- with `--filesystem=xdg-data/flatpak`
- `xdg-data/flatpak/app` already having been created inside the sandbox

what are the contents of .local/share/flatpak/overrides/com.discordapp.Discord`?

bbhtt:
It works if you create the directory on host before running the Flatpak. 

```
cat ~/.local/share/flatpak/overrides/com.discordapp.Discord
[Context]
filesystems=xdg-config/moonlight-mod:rw;
```

slonkazoid:
what's in xdg-config/moonlight-mod on the host?

slonkazoid:
Screenshot_20250119_140429

I have deleted all the data of Discord and moonlight installer and cannot replicate this

NotNite:
I can reproduce when ~/.config/moonlight-mod is missing - seems it's not being created. `flatpak run --filesystem=xdg-config/moonlight-mod:create io.github.moonlight_mod.moonlight-installer` works for me, so we probably need to update the mount - cc @slonkazoid.

slonkazoid:
updating lints

bbhtt:
bot, build io.github.moonlight_mod.moonlight-installer

NotNite:
Tested on a fresh user account after installing from Flatpak and Discord's tar.gz - works, but the overrides folder isn't being created. Another `:create`, I guess...

slonkazoid:
> Tested on a fresh user account after installing from Flatpak and Discord's tar.gz - works, but the overrides folder isn't being created. Another `:create`, I guess...

to clarify it is getting created, but only in the sandbox

bbhtt:
bot, build io.github.moonlight_mod.moonlight-installer

bbhtt:
hm new exception wasn't published yet you can restart the build in a few hours.

slonkazoid:
> hm new exception wasn't published yet, you can restart the build in a few hours.

seems like they are published now, do I just run the command again?

slonkazoid:
bot, build io.github.moonlight_mod.moonlight-installer

slonkazoid:
> Build [175483 successful](https://buildbot.flathub.org/#/builders/6/builds/175483) To test this build, install it from the testing repository:
> 
> ```
> flatpak install --user https://dl.flathub.org/build-repo/158627/io.github.moonlight_mod.moonlight-installer.flatpakref
> ```

it seems to work - it patched discord without errors, and moonlight-mod was functional in it

1. uninstalled existing moonlight-installer with flatpak remove --delete-data
2. installed this one
3. ran the mkdir

NotNite:
The easiest way is to `flatpak run com.discordapp.Discord` and look for a log from stdout along the lines of `[INFO] [core/extension/loader] Loaded <number> extensions`.

NotNite:
I've also tested it with my own account after signing in and it works okay - here's the settings menu that you can view after signing into Discord:
image

bbhtt:
I managed to get a test account through.

PR #6044, opened 2025-01-19
Add de.bwravencl.ControllerBuddy

bwRavencl:

ControllerBuddy is a highly advanced gamepad mapping software that supports the creation of input profiles for complex target applications such as flight simulators.

  In addition to simply mapping the buttons and axes of a physical game controller to keyboard and mouse input, ControllerBuddy also supports sending input commands to a virtual joystick.

  ControllerBuddy’s goal is to allow the user to control target applications using only a gamepad, without ever having to reach for a keyboard or mouse.

### DIFF
--- a/de.bwravencl.ControllerBuddy-update-profiles.desktop
+++ b/de.bwravencl.ControllerBuddy-update-profiles.desktop
@@ -1,0 +1,7 @@
+[Desktop Entry]
+Type=Application
+Name=Update ControllerBuddy-Profiles
+Icon=text-x-script
+Exec=/app/bin/update-profiles.sh
+Terminal=true
+Categories=Game
--- a/de.bwravencl.ControllerBuddy-update-profiles.desktop
+++ b/de.bwravencl.ControllerBuddy-update-profiles.desktop
@@ -2,6 +2,6 @@
 Type=Application
 Name=Update ControllerBuddy-Profiles
 Icon=text-x-script
-Exec=/app/bin/update-profiles.sh
+Exec=update-profiles.sh
 Terminal=true
 Categories=Game
--- a/de.bwravencl.ControllerBuddy.desktop
+++ b/de.bwravencl.ControllerBuddy.desktop
@@ -2,5 +2,5 @@
 Type=Application
 Name=ControllerBuddy
 Icon=de.bwravencl.ControllerBuddy
-Exec=/app/bin/launch.sh -autostart local -tray
+Exec=launch.sh -autostart local -tray
 Categories=Game
--- a/de.bwravencl.ControllerBuddy.desktop
+++ b/de.bwravencl.ControllerBuddy.desktop
@@ -1,0 +1,6 @@
+[Desktop Entry]
+Type=Application
+Name=ControllerBuddy
+Icon=de.bwravencl.ControllerBuddy
+Exec=/app/bin/launch.sh -autostart local -tray
+Categories=Game
--- a/de.bwravencl.ControllerBuddy.metainfo.xml
+++ b/de.bwravencl.ControllerBuddy.metainfo.xml
@@ -86,7 +86,7 @@
       <image type="source">https://raw.githubusercontent.com/bwRavencl/ControllerBuddy/master/screenshot_8.png</image>
     </screenshot>
   </screenshots>
-  <content_rating type="oars-1.1" />
+  <content_rating type="oars-1.1"/>
   <branding>
     <color type="primary" scheme_preference="light">#459dc6</color>
     <color type="primary" scheme_preference="dark">#2b2d31</color>
--- a/de.bwravencl.ControllerBuddy.metainfo.xml
+++ b/de.bwravencl.ControllerBuddy.metainfo.xml
@@ -1,0 +1,95 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>de.bwravencl.ControllerBuddy</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <name>ControllerBuddy</name>
+  <summary>Highly advanced game controller mapping</summary>
+  <icon type="remote">https://raw.githubusercontent.com/bwRavencl/ControllerBuddy/refs/heads/master/icon.svg</icon>
+  <description>
+    <p>
+      ControllerBuddy is an advanced gamepad mapping software that supports the creation of input profiles for complex target applications such as flight simulators.
+    </p>
+    <p>
+      In addition to the simple mapping of buttons and axes of a physical game controller to keyboard and mouse input, ControllerBuddy also supports feeding input commands to a virtual joystick device (vJoy / uinput).
+    </p>
+    <p>
+      The goal of ControllerBuddy is to allow the user to control target applications exclusively with a gamepad without ever having to reach for a keyboard or mouse.
+    </p>
+  </description>
+  <categories>
+      <category>Game</category>
+      <category>Utility</category>
+      <category>Java</category>
+  </categories>
+    <keywords>
+      <keyword>game</keyword>
+      <keyword>flight</keyword>
+      <keyword>simulation</keyword>
+      <keyword>controller</keyword>
+      <keyword>gamepad</keyword>
+      <keyword>mapping</keyword>
+      <keyword>xbox</keyword>
+      <keyword>dualshock</keyword>
+      <keyword>dualsense</keyword>
+      <keyword>steamdeck</keyword>
+    </keywords>
+  <url type="homepage">https://controllerbuddy.org</url>
+  <url type="bugtracker">https://github.com/bwRavencl/ControllerBuddy/issues</url>
+  <url type="faq">https://controllerbuddy.org/#faq</url>
+  <url type="vcs-browser">https://github.com/bwRavencl/ControllerBuddy</url>
+  <launchable type="desktop-id">de.bwravencl.ControllerBuddy.desktop</launchable>
+  <releases>
+    <release version="1.3.57-4a47fb6" date="2025-01-17">
+      <description/>
+    </release>
+  </releases>
+  <supports>
+    <control>pointing</control>
+    <control>keyboard</control>
+    <control>gamepad</control>
+  </supports>
+  <project_license>GPL-3.0</project_license>
+  <developer id="de.bwravencl">
+    <name>Matteo Hausner</name>
+    <url>https://bwravencl.de</url>
+  </developer>
+  <screenshots>
+    <screenshot type="default">
+      <caption>Visualization Tab</caption>
+      <image type="source">https://raw.githubusercontent.com/bwRavencl/ControllerBuddy/master/screenshot_6.png</image>
+    </screenshot>
+    <screenshot>
+      <caption>Overlay and On-Screen Keyboard</caption>
+      <image type="source">https://raw.githubusercontent.com/bwRavencl/ControllerBuddy/master/screenshot_7.png</image>
+    </screenshot>
+    <screenshot type="default">
+      <caption>Modes Tab</caption>
+      <image type="source">https://raw.githubusercontent.com/bwRavencl/ControllerBuddy/master/screenshot_1.png</image>
+    </screenshot>
+    <screenshot type="default">
+      <caption>Assignments Tab</caption>
+      <image type="source">https://raw.githubusercontent.com/bwRavencl/ControllerBuddy/master/screenshot_2.png</image>
+    </screenshot>
+    <screenshot>
+      <caption>Component Editor (Button)</caption>
+      <image type="source">https://raw.githubusercontent.com/bwRavencl/ControllerBuddy/master/screenshot_3.png</image>
+    </screenshot>
+    <screenshot>
+      <caption>Component Editor (Axis)</caption>
+      <image type="source">https://raw.githubusercontent.com/bwRavencl/ControllerBuddy/master/screenshot_4.png</image>
+    </screenshot>
+    <screenshot type="default">
+      <caption>Overlay Tab (Dark Theme)</caption>
+      <image type="source">https://raw.githubusercontent.com/bwRavencl/ControllerBuddy/master/screenshot_5.png</image>
+    </screenshot>
+    <screenshot>
+      <caption>Overlay and On-Screen Keyboard (VR)</caption>
+      <image type="source">https://raw.githubusercontent.com/bwRavencl/ControllerBuddy/master/screenshot_8.png</image>
+    </screenshot>
+  </screenshots>
+  <content_rating type="oars-1.0" />
+  <branding>
+    <color type="primary" scheme_preference="light">#459dc6</color>
+    <color type="primary" scheme_preference="dark">#2b2d31</color>
+  </branding>
+</component>
--- a/de.bwravencl.ControllerBuddy.metainfo.xml
+++ b/de.bwravencl.ControllerBuddy.metainfo.xml
@@ -87,7 +87,7 @@
       <image type="source">https://raw.githubusercontent.com/bwRavencl/ControllerBuddy/master/screenshot_8.png</image>
     </screenshot>
   </screenshots>
-  <content_rating type="oars-1.0" />
+  <content_rating type="oars-1.1" />
   <branding>
     <color type="primary" scheme_preference="light">#459dc6</color>
     <color type="primary" scheme_preference="dark">#2b2d31</color>
--- a/de.bwravencl.ControllerBuddy.metainfo.xml
+++ b/de.bwravencl.ControllerBuddy.metainfo.xml
@@ -4,7 +4,6 @@
   <metadata_license>CC0-1.0</metadata_license>
   <name>ControllerBuddy</name>
   <summary>Highly advanced game controller mapping</summary>
-  <icon type="remote">https://raw.githubusercontent.com/bwRavencl/ControllerBuddy/refs/heads/master/icon.svg</icon>
   <description>
     <p>
       ControllerBuddy is an advanced gamepad mapping software that supports the creation of input profiles for complex target applications such as flight simulators.
--- a/de.bwravencl.ControllerBuddy.yml
+++ b/de.bwravencl.ControllerBuddy.yml
@@ -8,7 +8,6 @@ finish-args:
   - --env=JAVA_TOOL_OPTIONS=-Dawt.useSystemAAFontSettings=gasp
   - --filesystem=xdg-documents
   - --persist=.java
-  - --persist=.local/share/applications/ControllerBuddy
   - --share=network
   - --socket=x11
   - --share=ipc
--- a/de.bwravencl.ControllerBuddy.yml
+++ b/de.bwravencl.ControllerBuddy.yml
@@ -3,6 +3,17 @@ runtime: org.freedesktop.Platform
 runtime-version: '24.08'
 sdk: org.freedesktop.Sdk
 command: launch.sh
+finish-args:
+  - --device=all
+  - --env=JAVA_TOOL_OPTIONS=-Djava.io.tmpdir=/var/tmp -Dawt.useSystemAAFontSettings=gasp
+  - --filesystem=xdg-documents
+  - --persist=.java
+  - --persist=.local/share/applications/ControllerBuddy
+  - --share=network
+  - --socket=x11
+  - --share=ipc
+  - --talk-name=org.freedesktop.ScreenSaver
+  - --talk-name=org.freedesktop.Flatpak
 modules:
   - name: ControllerBuddy
     buildsystem: simple
@@ -135,14 +146,3 @@ modules:
               type: anitya
               project-id: 1766
               url-template: https://xorg.freedesktop.org/releases/individual/lib/libXaw-$version.tar.gz
-finish-args:
-  - --device=all
-  - --env=JAVA_TOOL_OPTIONS=-Djava.io.tmpdir=/var/tmp -Dawt.useSystemAAFontSettings=gasp
-  - --filesystem=xdg-documents
-  - --persist=.java
-  - --persist=.local/share/applications/ControllerBuddy
-  - --share=network
-  - --socket=x11
-  - --share=ipc
-  - --talk-name=org.freedesktop.ScreenSaver
-  - --talk-name=org.freedesktop.Flatpak
--- a/de.bwravencl.ControllerBuddy.yml
+++ b/de.bwravencl.ControllerBuddy.yml
@@ -72,13 +72,13 @@ modules:
             fi
             if [ ! -d "$(xdg-user-dir DOCUMENTS)/ControllerBuddy-Profiles" ]
             then
-              xterm -fa Mono -e /app/bin/update-profiles.sh
+              /app/bin/update-profiles.sh
             fi
             /app/bin/ControllerBuddy "$@"
       - type: script
         dest-filename: update-profiles.sh
         commands:
-          - /app/bin/InstallControllerBuddy.sh profiles
+          - xterm -fa Mono -e /app/bin/InstallControllerBuddy.sh profiles
       - type: file
         path: de.bwravencl.ControllerBuddy.desktop
       - type: file
--- a/de.bwravencl.ControllerBuddy.yml
+++ b/de.bwravencl.ControllerBuddy.yml
@@ -137,14 +137,3 @@ modules:
         branch: 1.3
         commit: 0e2deb591d6695aba4d4ae51ddc6bc43572a8359
         dest: ControllerBuddy-Profiles
-  - name: zenity
-    buildsystem: meson
-    sources:
-      - type: archive
-        url: https://download.gnome.org/sources/zenity/3.44/zenity-3.44.5.tar.xz
-        sha256: 29ac447f4246b1da783a3c28ddc80b0ef7c80eccd5836ad9140abf89841d0eab
-        x-checker-data:
-          type: gnome
-          name: zenity
-          versions:
-            <: 3.90.0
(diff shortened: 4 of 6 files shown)


de.bwravencl.ControllerBuddy.metainfo.xml
```diff
@@ -0,0 +1,95 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>de.bwravencl.ControllerBuddy</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <name>ControllerBuddy</name>
+  <summary>Highly advanced game controller mapping</summary>
+  <icon type="remote">https://raw.githubusercontent.com/bwRavencl/ControllerBuddy/refs/heads/master/icon.svg</icon>
```
hfiguiere, line 7:
```suggestion
```

de.bwravencl.ControllerBuddy.metainfo.xml
```diff
@@ -0,0 +1,95 @@
...
+  </supports>
+  <project_license>GPL-3.0</project_license>
+  <developer id="de.bwravencl">
+    <name>Matteo Hausner</name>
+    <url>https://bwravencl.de</url>
+  </developer>
+  <screenshots>
+    <screenshot type="default">
+      <caption>Visualization Tab</caption>
+      <image type="source">https://raw.githubusercontent.com/bwRavencl/ControllerBuddy/master/screenshot_6.png</image>
```
hfiguiere, line 59: this should NEVER reference the tip of a branch.

(applies to the others)

de.bwravencl.ControllerBuddy.metainfo.xml
```diff
@@ -0,0 +1,95 @@
...
+    <screenshot type="default">
+      <caption>Overlay Tab (Dark Theme)</caption>
+      <image type="source">https://raw.githubusercontent.com/bwRavencl/ControllerBuddy/master/screenshot_5.png</image>
+    </screenshot>
+    <screenshot>
+      <caption>Overlay and On-Screen Keyboard (VR)</caption>
+      <image type="source">https://raw.githubusercontent.com/bwRavencl/ControllerBuddy/master/screenshot_8.png</image>
+    </screenshot>
+  </screenshots>
+  <content_rating type="oars-1.0" />
```
hfiguiere, line 90:
```suggestion
  <content_rating type="oars-1.1" />
```

de.bwravencl.ControllerBuddy.yml
```diff
@@ -0,0 +1,145 @@
...
+      - |
+        cp -R ControllerBuddy/* /app/ &&
+        install -Dm 755 launch.sh -t /app/bin &&
+        install -Dm 755 update-profiles.sh -t /app/bin &&
+        install -Dm 755 InstallControllerBuddy.sh /app/bin/InstallControllerBuddy.sh &&
+        install -Dm 644 icon.svg "/app/share/icons/hicolor/scalable/apps/$FLATPAK_ID.svg" &&
+        install -Dm 644 "$FLATPAK_ID.metainfo.xml" -t /app/share/metainfo &&
+        shortcuts_dir=/app/share/applications &&
+        mkdir -p "$shortcuts_dir" &&
+        echo -e "[Desktop Entry]\nType=Application\nName=ControllerBuddy\nIcon=de.bwravencl.ControllerBuddy\nExec=/app/bin/launch.sh -autostart local -tray\nCategories=Game" > "$shortcuts_dir/de.bwravencl.ControllerBuddy.desktop" &&
```
hfiguiere, line 19: you should NEVER make a file with echo like that. Also the desktop file should come from upstream

de.bwravencl.ControllerBuddy.yml
```diff
@@ -0,0 +1,145 @@
...
+        echo -e "[Desktop Entry]\nType=Application\nName=Update ControllerBuddy-Profiles\nIcon=text-x-script\nExec=/app/bin/update-profiles.sh\nTerminal=true\nCategories=Game" > "$shortcuts_dir/de.bwravencl.ControllerBuddy.Update.desktop"
+    sources:
+      - type: archive
+        only-arches:
+          - x86_64
+        url: https://github.com/bwRavencl/ControllerBuddy/releases/download/1.3.57-4a47fb6/ControllerBuddy-linux-x86-64-1.3.57-4a47fb6.tgz
```
hfiguiere, line 25: why not building from source?

de.bwravencl.ControllerBuddy.yml
```diff
@@ -0,0 +1,145 @@
...
+        buildsystem: autotools
+        sources:
+          - type: archive
+            url: https://xorg.freedesktop.org/releases/individual/lib/libXaw-1.0.16.tar.gz
+            sha256: 012f90adf8739f2f023d63a5fee1528949cf2aba92ef7ac1abcfc2ae9cf28798
+            x-checker-data:
+              type: anitya
+              project-id: 1766
+              url-template: https://xorg.freedesktop.org/releases/individual/lib/libXaw-$version.tar.gz
+finish-args:
```
hfiguiere, line 136: this should be above "modules"

de.bwravencl.ControllerBuddy.yml
```diff
@@ -0,0 +1,145 @@
...
+  - --device=all
+  - --env=JAVA_TOOL_OPTIONS=-Djava.io.tmpdir=/var/tmp -Dawt.useSystemAAFontSettings=gasp
+  - --filesystem=xdg-documents
+  - --persist=.java
+  - --persist=.local/share/applications/ControllerBuddy
+  - --share=network
+  - --socket=x11
+  - --share=ipc
+  - --talk-name=org.freedesktop.Flatpak
```
hfiguiere, line 145: why does it need this dbus access?

de.bwravencl.ControllerBuddy.yml
```diff
@@ -0,0 +1,145 @@
...
+          type: anitya
+          project-id: 376395
+          stable-only: false
+          url-template: https://github.com/bwRavencl/ControllerBuddy-Install-Script/releases/download/$version/InstallControllerBuddy.sh
+      - type: script
+        dest-filename: launch.sh
+        commands:
+          - |
+            script_file="$XDG_RUNTIME_DIR/app/$FLATPAK_ID/InstallControllerBuddy.sh"
+            install -Dm 755 /app/bin/InstallControllerBuddy.sh "$script_file"
```
hfiguiere, line 65: why does the start script install something?

de.bwravencl.ControllerBuddy.metainfo.xml
hfiguiere, line 1: this should be upstream.

de.bwravencl.ControllerBuddy.yml
bwRavencl: There are no desktop files.
They get created by the install script in the same way.

de.bwravencl.ControllerBuddy.yml
```diff
@@ -0,0 +1,145 @@
...
+    sources:
+      - type: archive
+        url: https://www.kernel.org/pub/software/scm/git/git-2.48.1.tar.xz
+        sha256: 1c5d545f5dc1eb51e95d2c50d98fdf88b1a36ba1fa30e9ae5d5385c6024f82ad
+        x-checker-data:
+          type: anitya
+          project-id: 5350
+          stable-only: true
+          url-template: https://www.kernel.org/pub/software/scm/git/git-$version.tar.xz
+  - name: xterm
```
hfiguiere, line 103: why is xterm needed?

de.bwravencl.ControllerBuddy.yml
hfiguiere: There should be .desktop files. And Flatpak is not to package insall scripts.

de.bwravencl.ControllerBuddy.yml
bwRavencl: Because it is a Java application and handling the build process without access to the internet is next to impossible.

Background:
During the build of ControllerBuddy the JVM that is usually installed on the system is only used to orchestrate the build process, the actual building of the application happens with a toolchain JVM that must be downloaded by gradle if it is not present locally. This ensures the correct JVM (version and distributor) is used.
Details on this subject can be found in the [gradle documentation on toolchains](https://docs.gradle.org/current/userguide/toolchains.html)

My first attempt using [Flatpak Gradle Generator from flatpak/flatpak-builder-tools](https://github.com/flatpak/flatpak-builder-tools/tree/master/gradle) failed miserably because the generator is nothing more than a log parser and includes no deeper logic to determine which URLs are actually relevant to the build process and which not.
E.g. it was unable to resolve the correct URLs for gradle plugins, for which the JAR and plugin name differs.

I made another attempt with [jwharm/flatpak-gradle-generator](https://github.com/jwharm/flatpak-gradle-generator), which works better, but it ultimately failed too, because it does not cover the whole toolchain part, I described above.

de.bwravencl.ControllerBuddy.yml
bwRavencl: ControllerBuddy talks to dbus to inhibit the screen-saver from becoming active or the screen going to energy save, while gaming with a controller, due to the lack of phyiscal keyboard / mouse inputs.

de.bwravencl.ControllerBuddy.yml
bwRavencl: The start script runs the `InstallControllerBuddy.sh` script two times:

1. Outside the sandbox to allow the user to automatically add the necessary udev rules for access to uinput and HID level access to Sony DualShock and DualSense controllers.
2. Once more, from inside the Sandbox to obtain the official ControllerBuddy-Profiles if they are not present yet.

de.bwravencl.ControllerBuddy.yml
bwRavencl: Because a terminal emulator is needed for the user to interact with the ControllerBuddy-Install-Script

de.bwravencl.ControllerBuddy.metainfo.xml
bwRavencl: Please elaborate.

de.bwravencl.ControllerBuddy.yml
hfiguiere: there are actual DBUS API for that. It feels done wrong.

de.bwravencl.ControllerBuddy.yml
bwRavencl: The application itself does not and will not come with .desktop files.
Would an acceptable alternative be to put them inside the Flathub Repo?

de.bwravencl.ControllerBuddy.yml
hfiguiere: why? This feel like poor UX.

de.bwravencl.ControllerBuddy.yml
bwRavencl: I'm pretty sure I am using those exact APIs:
https://github.com/bwRavencl/ControllerBuddy/blob/1.3.57-4a47fb6/src/main/java/de/bwravencl/controllerbuddy/runmode/dbus/ScreenSaverType.java#L21-L24

de.bwravencl.ControllerBuddy.yml
hfiguiere: > The application itself does not and will not come with .desktop files.

Why? It's a bug to not have one for a graphical application. .desktop are going on being a quarter century old.

de.bwravencl.ControllerBuddy.yml
hfiguiere: then why do you ask for this DBus name? Did you even test it?

de.bwravencl.ControllerBuddy.yml
bwRavencl: Because in ControllerBuddy the setup / system preparation process is not part of the application itself, but done via the ControllerBuddy-Install-Script, which is text-mode only.

de.bwravencl.ControllerBuddy.yml
bwRavencl: Because the application is multiplatform, primarily Windows, but also Linux and macOS.
And shortcut files get generated during the installation process by the installer.

de.bwravencl.ControllerBuddy.yml
bwRavencl: It seems I misunderstood.
Are you referring to the `--talk-name=org.freedesktop.Flatpak`?
It's there for the `InstallControllerBuddy.sh` script to leave the sandbox for adding udev rules.

hfiguiere: >  Are you referring to the --talk-name=org.freedesktop.Flatpak?

I don't see any other DBus permission...

bwRavencl: > ControllerBuddy talks to dbus to inhibit the screen-saver from becoming active or the screen going to energy save, while gaming with a controller, due to the lack of phyiscal keyboard / mouse inputs.

Okay, the `org.freedesktop.ScreenSaver` D-Bus API was actually missing and I added it now with commit e9eb0ea047213ae28cba6e522f2773f797382b09.

de.bwravencl.ControllerBuddy.yml
bwRavencl: Fixed with 374cb8ca08ca0862b4cc1ca98e2e3492f329e5ab

de.bwravencl.ControllerBuddy.yml
bwRavencl: With 947db6172f8ebd7aa5362f11d616d239f59e5c2a the .desktop files are now static files. As I explained, it makes no sense to include them in the upstream main app repo, that's why they have to reside here - like it seems to be done with many other Flatpaks.

de.bwravencl.ControllerBuddy.metainfo.xml
bbhtt: Remote icon not needed and not allowed.

de.bwravencl.ControllerBuddy-update-profiles.desktop
```diff
@@ -0,0 +1,7 @@
+[Desktop Entry]
+Type=Application
+Name=Update ControllerBuddy-Profiles
+Icon=text-x-script
+Exec=/app/bin/update-profiles.sh
```
bbhtt, line 5: exec should not be a path

de.bwravencl.ControllerBuddy.desktop
```diff
@@ -0,0 +1,6 @@
+[Desktop Entry]
+Type=Application
+Name=ControllerBuddy
+Icon=de.bwravencl.ControllerBuddy
+Exec=/app/bin/launch.sh -autostart local -tray
```
bbhtt, line 5: same here, exec should not be a path

de.bwravencl.ControllerBuddy.yml
```diff
@@ -0,0 +1,148 @@
...
+runtime: org.freedesktop.Platform
+runtime-version: '24.08'
+sdk: org.freedesktop.Sdk
+command: launch.sh
+finish-args:
+  - --device=all
+  - --env=JAVA_TOOL_OPTIONS=-Djava.io.tmpdir=/var/tmp -Dawt.useSystemAAFontSettings=gasp
+  - --filesystem=xdg-documents
+  - --persist=.java
+  - --persist=.local/share/applications/ControllerBuddy
```
bbhtt, line 11: why? why not use XDG compliant location?

de.bwravencl.ControllerBuddy.yml
bbhtt: >  it makes no sense to include them in the upstream main app repo

a desktop app with desktop file makes no sense either.

de.bwravencl.ControllerBuddy.yml
bbhtt: what udev rules are you installing and where do I see them? the script in the manifest contains none. and I'm not sure why installing udev rules would require calling xterm. 

just installing them silently and show a dialogue box? 

Is this only for installing udev rules or you need it for other things as well?

de.bwravencl.ControllerBuddy-update-profiles.desktop
bwRavencl: Fixed with f52e3d0fd06e13cde90e6e188c58d6e42311d24e

de.bwravencl.ControllerBuddy.desktop
bwRavencl: Fixed with f52e3d0fd06e13cde90e6e188c58d6e42311d24e

de.bwravencl.ControllerBuddy.yml
bwRavencl: Good catch - this was a left-over from an earlier iteration, I removed the override with 3e93f04113ed83b2bd07c7e3587a14aef48a9696

de.bwravencl.ControllerBuddy.yml
bwRavencl: > > it makes no sense to include them in the upstream main app repo
> 
> a desktop app without desktop file makes no sense either.

As I already explained, when installed with the installer - when not packaged as a Flatpak - shortcuts are generated dynamically by the installer, depending on the OS, on Windows .lnk files are created and on Linux .desktop.
As I understand, for the Flatpak static resources are required, that's why they have to be included here.
Generally, I don't want Flatpak specific stuff in the main repo, as it is only way to ship the app.

de.bwravencl.ControllerBuddy.yml
bwRavencl: The udev rules get added by the `InstallControllerBuddy.sh` script, effectively the following lines of the script are run outside the sandbox, when launched with the `prepare` cmd option:
https://github.com/bwRavencl/ControllerBuddy-Install-Script/blob/master/InstallControllerBuddy.sh#L645-L674

A terminal app is required, because the script is interactive and it will e.g. ask for the sudo password, because I can't assume there is a terminal app present on all systems, xterm gets shipped with the Flatpak.

bbhtt: desktop file is not "Flatpak stuff", the standard exists almost for 20 years well before flatpak

de.bwravencl.ControllerBuddy.yml
bbhtt: desktop file is not "Flatpak stuff", the standard exists almost for 20 years well before flatpak

> generated dynamically by the installer

why not do the same here?

de.bwravencl.ControllerBuddy.yml
bbhtt: I'm not sure what you are replying too but the comment applies to `persist`. it shows the line I commented to.

image

de.bwravencl.ControllerBuddy.yml
bbhtt: if you are executing on host anway better to use polkit to present a GUI?

de.bwravencl.ControllerBuddy.yml
```diff
@@ -0,0 +1,148 @@
...
+        x-checker-data:
+          type: anitya
+          project-id: 376395
+          stable-only: false
+          url-template: https://github.com/bwRavencl/ControllerBuddy-Install-Script/releases/download/$version/InstallControllerBuddy.sh
+      - type: script
+        dest-filename: launch.sh
+        commands:
+          - |
+            script_file="$XDG_RUNTIME_DIR/app/$FLATPAK_ID/InstallControllerBuddy.sh"
```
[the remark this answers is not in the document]
bbhtt: i see it also tries to  do a sort of self update, so the hash of the script here means nothing if you go to `/tmp` on host and update it there and execute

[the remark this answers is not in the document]
bbhtt: it should be a simpler script in the manifest with reduced scope, you only need to install the udev rules I assume?

de.bwravencl.ControllerBuddy.yml
bbhtt: this permission needs an exception https://docs.flathub.org/docs/for-app-authors/linter#exceptions https://docs.flathub.org/docs/for-app-authors/linter#finish-args-flatpak-spawn-access, but with the current script you are executing it's complicated. something would reduced scope will help as I said before.

de.bwravencl.ControllerBuddy.yml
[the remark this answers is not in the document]
bbhtt: i also saw you mentioned "profiles". why can't these profiles be included in the manifest?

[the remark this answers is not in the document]
bwRavencl: > it should be a simpler script in the manifest with reduced scope, you only need to install the udev rules I assume?
> i also saw you mentioned "profiles". why can't these profiles be included in the manifest?

Valid points, I took the time to rework my approach with a93726c16a1949a4a50e9dc829891bf2253b0718 and got entirely rid of the dependency on the original install script.  
Instead I implemented a much condensed launch script, that uses `pkexec` and `zenity` instead of `sudo` and thus no longer requires a terminal emulator to run.

The profiles are now also no longer getting downloaded via a script but shipped as part of the manifest as you suggested.

I'm looking forward to get your feedback on those changes!

de.bwravencl.ControllerBuddy.yml
bwRavencl: Also a left-over - removed with 93a7c5d987a46374e68b7fafe72dd712bb0c2790

de.bwravencl.ControllerBuddy.yml
bwRavencl: Fixed with a93726c16a1949a4a50e9dc829891bf2253b0718 which introduces a condensed Flatpak-specific launch-script based on `pkexec` and `zenity` and thus no longer requires text-mode.

de.bwravencl.ControllerBuddy.yml
bwRavencl: The dependency on `InstallControllerBuddy.sh` got entirely removed with a93726c16a1949a4a50e9dc829891bf2253b0718

de.bwravencl.ControllerBuddy.yml
bwRavencl: > if you are executing on host anway better to use polkit to present a GUI?

Implemented with a93726c16a1949a4a50e9dc829891bf2253b0718

de.bwravencl.ControllerBuddy.yml
```diff
@@ -0,0 +1,150 @@
...
+            fi
+
+            CONTROLLER_BUDDY_PROFILES_DIR=/app/share/ControllerBuddy-Profiles ControllerBuddy "$@"
+      - type: file
+        path: de.bwravencl.ControllerBuddy.desktop
+      - type: file
+        path: de.bwravencl.ControllerBuddy.metainfo.xml
+      - type: git
+        url: https://github.com/bwRavencl/ControllerBuddy-Profiles.git
+        branch: 1.3
```
hfiguiere, line 137: NEVER git branches.

bwRavencl: Well there are no tags or releases for the profiles repo.
Basically I want flathubbot to open a PR whenever there's a push to the branch (currently 1.3) that matches the version of the App.
What are the alternatives?

hfiguiere:
```suggestion
```

Apparently I wasn't clear.

de.bwravencl.ControllerBuddy.yml
```diff
@@ -0,0 +1,150 @@
...
+        commit: 0e2deb591d6695aba4d4ae51ddc6bc43572a8359
+        dest: ControllerBuddy-Profiles
+  - name: zenity
```
hfiguiere, line 140: zenity is in the runtime

bwRavencl: Module removed with 02d0d9fc4acd610ad8e6d8ae75e0b86c36196ed6

de.bwravencl.ControllerBuddy.yml
bwRavencl: > Apparently I wasn't clear.

You were very clear, but unfortunately you didn't answer the question I posed:
> Basically I want flathubbot to open a PR whenever there's a push to the branch (currently 1.3) that matches the version of the App.

When running `org.flathub.flatpak-external-data-checker --update` on the manifest, with the branch info removed, it says:
`INFO    src.checkers.gitchecker: Skipping source ControllerBuddy-Profiles.git, commit is specified, but neither tag nor branch`

Having to manually update the manifest every time there is a new commit in the repo is a no-go for me, this needs to be automated, i.e. I'd like to get an automatic pull request that only have to accept.

hfiguiere: nightly builds are NOT allowed.

de.bwravencl.ControllerBuddy.yml
```diff
@@ -0,0 +1,139 @@
...
+      - type: file
+        url: https://raw.githubusercontent.com/bwRavencl/ControllerBuddy/refs/tags/1.3.66-639d789/icon.svg
+        sha256: 22cb081b000a30f248e6c5f842a9519d933339d01449fe2889e3379cc5628523
+        x-checker-data:
+          type: anitya
+          project-id: 376393
+          stable-only: false
+          url-template: https://raw.githubusercontent.com/bwRavencl/ControllerBuddy/refs/tags/$version/icon.svg
+      - type: script
+        dest-filename: launch.sh
```
bbhtt, line 59: i think this will be easier to maintain as a separate file, long lines, yaml and no highlight makes it hard to read.

bwRavencl:
I'm closing this PR because the "nightly builds are NOT allowed" policy doesn't work at all with the way my development cycle looks like.
I already set up a dedicated Flatpak repo for ControllerBuddy instead of pursuing acceptance into Flathub.  
I'd like to thank @bbhtt for his suggestions and feedback, which inspired me to improve my approach on packaging.

PR #6087, opened 2025-01-27
Add de.desy.constellation

stephanlachnit:

- [x] Please describe the application briefly. Control and Data Acquisition System
- [x] I am an author/developer/upstream contributor to the project. If not, I contacted upstream developers about this submission. **Link:** https://gitlab.desy.de/constellation/constellation/-/merge_requests/554 (@simonspa is coauthor as well)

### DIFF
--- a/de.desy.constellation.metainfo.xml
+++ b/de.desy.constellation.metainfo.xml
@@ -1,0 +1,81 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<component type="desktop-application">
+  <id>de.desy.constellation</id>
+  <metadata_license>CC0-1.0</metadata_license>
+
+  <name>Constellation</name>
+  <developer id="de.desy">
+    <name>Deutsches Elektronen-Synchrotron DESY</name>
+  </developer>
+  <summary>Control and Data Acquisition System</summary>
+  <icon type="stock">de.desy.constellation</icon>
+  <project_license>EUPL-1.2</project_license>
+  <url type="homepage">https://constellation.pages.desy.de/</url>
+  <url type="vcs-browser">https://gitlab.desy.de/constellation/constellation</url>
+
+  <launchable type="desktop-id">de.desy.constellation.missioncontrol.desktop</launchable>
+  <launchable type="desktop-id">de.desy.constellation.observatory.desktop</launchable>
+
+  <description>
+    <p>
+      Constellation is an autonomous control and data acquisition system for small-scale experiments and experimental setup with volatile and dynamic constituents such as testbeam environments or laboratory test stands. Constellation aims to provide a flexible framework that requires minimal effort for the integration of new devices.
+    </p>
+    <p>
+      Devices participating in the DAQ system are called satellites. Satellite are autonomous and communicate over the network with each other and graphical user interfaces for controlling and monitoring. Constellation provides two user interfaces: MissionControl for controlling and Observatory for logging.
+    </p>
+    <p>
+      Detailed documentation on how to setup and run Constellation can be found on the homepage.
+    </p>
+  </description>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://constellation.pages.desy.de/_images/missioncontrol_run.png</image>
+      <caption>MissionControl main window</caption>
+    </screenshot>
+  </screenshots>
+  <categories>
+      <category>Science</category>
+      <category>Physics</category>
+  </categories>
+  <keywords>
+    <keyword translate="no">DAQ</keyword>
+    <keyword>testbeam</keyword>
+  </keywords>
+  <branding>
+    <color type="primary" scheme_preference="light">#6fbbd8</color>
+    <color type="primary" scheme_preference="dark">#d97e7c</color>
+  </branding>
+
+  <content_rating type="oars-1.1" />
+  <requires>
+    <display_length compare="ge">768</display_length>
+  </requires>
+  <recommends>
+    <control>keyboard</control>
+    <control>pointing</control>
+  </recommends>
+  <supports>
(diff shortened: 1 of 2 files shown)


de.desy.constellation.yml
```diff
@@ -0,0 +1,41 @@
...
+      - -Dlistener_observatory=true
+      - -Dsatellite_dev_null_receiver=true
+      - -Dsatellite_eudaq_native_writer=true
+      - -Dsatellite_random_transmitter=true
+      - -Dsatellite_sputnik=true
+    sources:
+      - type: archive
+        url: https://gitlab.desy.de/constellation/constellation/-/releases/v0.3/downloads/Constellation-v0.3.tar.xz
+        sha256: 8d30f5c7a1203ded7768a00e58281fffc907abfd1257a50de0b533f99a56ddad
+  - name: Constellation-metainfo
```
hfiguiere, line 35: no need for a separate module. this should be merged with the "main" one.

de.desy.constellation.yml
```diff
@@ -0,0 +1,41 @@
+id: de.desy.constellation
+runtime: org.kde.Platform
+runtime-version: "6.8"
+sdk: org.kde.Sdk
+command: Satellite
+finish-args:
+  - --share=network
+  - --filesystem=host
```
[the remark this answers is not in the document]
stephanlachnit: Yeah, this is a tricky one. Constellation is a DAQ system where independent components take over specialized tasks (called "satellites"). E.g. you have one satellite that takes data, and one that stores data. The data path where to store the data is configured in a config file, which is transmitted to the satellite that stores the data. This means that we cannot use portals for this. The alternative would be that users allow every single data path via Flatseal or similar, but this is a bit tedious for a DAQ system.

de.desy.constellation.yml
stephanlachnit: I.e. merged upstream? We can do that, but not retroactively for this release - can we fix this in the next release?

hfiguiere: yes having the appstream file upstream for a future release would be great but that's not what I mean here. There is no reason to have s separate module to install the appstream file.
(the module list start line 17)

stephanlachnit: Right, but how do I then install the metainfo file? Can I add build commands to meson?

de.desy.constellation.metainfo.xml
bbhtt, line 1: this should be submitted to https://gitlab.desy.de/constellation/constellation

de.desy.constellation.metainfo.xml
```diff
@@ -0,0 +1,70 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<component type="desktop-application">
+  <id>de.desy.constellation</id>
+  <metadata_license>CC0-1.0</metadata_license>
+
+  <name>Constellation</name>
+  <developer id="de.desy"><name>Deutsches Elektronen-Synchrotron DESY</name></developer>
```
bbhtt, line 7: indent this properly.

de.desy.constellation.yml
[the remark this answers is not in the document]
bbhtt: why not `home`?

de.desy.constellation.yml
bbhtt: and metainfo should be installed by meson, not manually

de.desy.constellation.yml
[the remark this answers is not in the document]
stephanlachnit: > why not `home`?

Changed to home. Often people store data in e.g. `/data`, but those aren't included in `host` anyway, so changed to home.

de.desy.constellation.metainfo.xml
stephanlachnit: Done: https://gitlab.desy.de/constellation/constellation/-/merge_requests/599

stephanlachnit:
I changed the metainfo to be less Flathub specific, will be merged upstream for the next release

bbhtt:
bot, build de.desy.constellation

bbhtt:
flatpak run does not work:

```
test@fedora:~$ flatpak run de.desy.constellation -g
|2025-01-28 21:03:39.906| CRITICAL [DEFAULT] Argument parsing failed: Too few arguments for '-g'.
|2025-01-28 21:03:39.906| CRITICAL [DEFAULT] Run "satellite --help" for help
```

why is this not pointing to whatever the desktop file is opening?

bbhtt:
Also can you post a screenrecord of how the app works?

stephanlachnit:
> flatpak run does not work:
> 
> ```
> test@fedora:~$ flatpak run de.desy.constellation -g
> |2025-01-28 21:03:39.906| CRITICAL [DEFAULT] Argument parsing failed: Too few arguments for '-g'.
> |2025-01-28 21:03:39.906| CRITICAL [DEFAULT] Run "satellite --help" for help
> ```
> 
> why is this not pointing to whatever the desktop file is opening?

This is because Constellation comes with multiple applications for different use cases (like the LibreOffice Flatpak).

There are GUI components (the MissionControl and Observatory desktop files), but the actual DAQ software does not run in a UI (which would be a really bad idea). When invoking `flatpak run`, I decided to use the command that starts DAQ satellites instead one of the GUIs.

Most of the time, the Satellites that run the actual DAQ will not run from flatpak, since they are quite device specific and thus live in obscure git repos and are only installed on the computer that is connected to the hardware.

The main point of the flatpak is to make it easy to install the GUI to operate and monitor these satellite over the network. However, I still think it is useful that one can test the flatpak on its own without additional software installed somewhere.

> Also can you post a screenrecord of how the app works?

https://streamable.com/heb1rx

PR #6119, opened 2025-02-03
Add dev.heppen.webapps

hepp3n:

This is re-up Quick Web Apps app with changed app_id. I made new domain and website and wanted to change app_id also to it. Link to app: https://flathub.org/apps/io.github.elevenhsoft.WebApps

I need to submit some exceptions otherwise linter shouting about it. Basically I need access `ro` for host-os so we can read what browsers user have installed. Also, we need to read `xdg-data/flatpak` for the same reason but for flatpak browsers and also we need to communicate with `org.freedesktop.Flatpak` because our web apps are `flatpak-spawn`ed with `--host` argument.

This app allows to create web app launchers by passing special arguments to each browser, so it will behave like a web app or PWA.

### DIFF
--- a/dev.heppen.webapps.json
+++ b/dev.heppen.webapps.json
@@ -16,9 +16,9 @@
     "--device=dri",
     "--share=network",
     "--filesystem=host-os:ro",
-    "--filesystem=xdg-data/flatpak/exports/share/applications:ro",
+    "--filesystem=xdg-data/applications",
+    "--filesystem=xdg-data/flatpak",
     "--talk-name=com.system76.CosmicSettingsDaemon",
-    "--talk-name=org.freedesktop.Flatpak"
   ],
   "build-options": {
     "append-path": "/usr/lib/sdk/rust-stable/bin"
@@ -41,7 +41,7 @@
         {
           "type": "git",
           "url": "https://github.com/cosmic-utils/web-apps",
-          "commit": "fa33340f166adc2040f9a4480512be063d38fe97",
+          "commit": "92d3713703ff80753f32a17b4e06d8fd6313a002",
           "builddir": true
         },
         "cargo-sources.json"
--- a/dev.heppen.webapps.json
+++ b/dev.heppen.webapps.json
@@ -16,7 +16,7 @@
     "--device=dri",
     "--share=network",
     "--filesystem=host-os:ro",
-    "--filesystem=xdg-data/flatpak:ro",
+    "--filesystem=xdg-data/flatpak/exports/share/applications:ro",
     "--talk-name=com.system76.CosmicSettingsDaemon",
     "--talk-name=org.freedesktop.Flatpak"
   ],
--- a/dev.heppen.webapps.json
+++ b/dev.heppen.webapps.json
@@ -16,9 +16,9 @@
     "--device=dri",
     "--share=network",
     "--filesystem=host-os:ro",
+    "--filesystem=xdg-data/flatpak:ro",
     "--filesystem=xdg-data/applications",
-    "--filesystem=xdg-data/flatpak",
-    "--talk-name=com.system76.CosmicSettingsDaemon",
+    "--talk-name=com.system76.CosmicSettingsDaemon"
   ],
   "build-options": {
     "append-path": "/usr/lib/sdk/rust-stable/bin"
@@ -41,7 +41,7 @@
         {
           "type": "git",
           "url": "https://github.com/cosmic-utils/web-apps",
-          "commit": "92d3713703ff80753f32a17b4e06d8fd6313a002",
+          "commit": "3577ce8a781dfa98985f325ccd7ddc00e256e89e",
           "builddir": true
         },
         "cargo-sources.json"
--- a/dev.heppen.webapps.json
+++ b/dev.heppen.webapps.json
@@ -1,0 +1,51 @@
+{
+  "app-id": "dev.heppen.webapps",
+  "runtime": "org.freedesktop.Platform",
+  "runtime-version": "24.08",
+  "sdk": "org.freedesktop.Sdk",
+  "sdk-extensions": [
+    "org.freedesktop.Sdk.Extension.rust-stable"
+  ],
+  "base": "com.system76.Cosmic.BaseApp",
+  "base-version": "stable",
+  "command": "quick-webapps",
+  "finish-args": [
+    "--share=ipc",
+    "--socket=fallback-x11",
+    "--socket=wayland",
+    "--device=dri",
+    "--share=network",
+    "--filesystem=host-os:ro",
+    "--filesystem=xdg-data/flatpak:ro",
+    "--talk-name=com.system76.CosmicSettingsDaemon",
+    "--talk-name=org.freedesktop.Flatpak"
+  ],
+  "build-options": {
+    "append-path": "/usr/lib/sdk/rust-stable/bin"
+  },
+  "modules": [
+    {
+      "name": "quick-webapps",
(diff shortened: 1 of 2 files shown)


hepp3n:
Should I submit exceptions right now?

hepp3n:
I just bought new domain for my personal website and want to migrate some my projects to it. Also I hope I will be able to make a some landing page for this app in this domain.

dev.heppen.webapps.json
```diff
@@ -0,0 +1,51 @@
...
+  "base-version": "stable",
+  "command": "quick-webapps",
+  "finish-args": [
+    "--share=ipc",
+    "--socket=fallback-x11",
+    "--socket=wayland",
+    "--device=dri",
+    "--share=network",
+    "--filesystem=host-os:ro",
+    "--filesystem=xdg-data/flatpak:ro",
```
bbhtt, line 19: > Basically I need access ro for host-os so we can read what browsers user have installed. Also, we need to read xdg-data/flatpak for the same reason but for flatpak browsers 

what does this read? Desktop files? Then limit it to the subdirectory.

dev.heppen.webapps.json
```diff
@@ -0,0 +1,51 @@
...
+    "--talk-name=com.system76.CosmicSettingsDaemon",
+    "--talk-name=org.freedesktop.Flatpak"
```
bbhtt, line 21: what do you mean by this?

> also we need to communicate with org.freedesktop.Flatpak because our web apps are flatpak-spawned with --host argument.

If it's using host browser you would start it like `firefox ...` and if it is usign flatpak, you would use `flatpak run org.mozilla.fireox ..`, where is flatpak-spawn coming from?

hepp3n: I'm using Dynamic Launcher Portal for creating and launching web apps from Quick Web Apps. When I'm doing `Exec=` with only binary name, it will make it to `flatpak run --command=firefox` which will fails later. My solution was to use `flatpak-spawn` so it works.

dev.heppen.webapps.json
hepp3n: Well, i updated permissions to his:

```
    "--filesystem=/usr/share/applications:ro",
    "--filesystem=/usr/local/share/applications:ro",
    "--filesystem=/var/lib/flatpak:ro",
    "--filesystem=xdg-data/applications:ro",
    "--filesystem=xdg-data/flatpak:ro",
    "--talk-name=com.system76.CosmicSettingsDaemon",
    "--talk-name=org.freedesktop.Flatpak"
```
but then i got warning, `/usr` is restricted so i can't read system browsers... can i leave it as it is? it works right now ^^

bbhtt: I commented on `xdg-data/flatpak`


>     "--filesystem=/usr/share/applications:ro",
    "--filesystem=/usr/local/share/applications:ro",

Won't work https://docs.flatpak.org/en/latest/sandbox-permissions.html#reserved-paths

>     "--filesystem=xdg-data/flatpak:ro",

this can be `xdg-data/flatpak/exports/share/applications:ro`, same for /var/lib right?

dev.heppen.webapps.json
bbhtt: that seems complicated, whenever it is host browser why not write the desktop file directly skipping the portal?

hepp3n: yes, to be honest it was recommended me in this topic: https://discourse.flathub.org/t/how-to-work-with-system-paths-from-flatpak/8796/5

i'm not really sure how to work in flatpak sandbox ^^ im just wrote this app and then thought it's a good solution to distribute this app to others, however it's kind of complicated right now :/

dev.heppen.webapps.json
hepp3n: ah yes, sorry then, will correct this :)

dev.heppen.webapps.json
hepp3n: the portal definitely helps with storing desktop file and icon for sure, otherwise i would handle this on my own

bbhtt: is there any case where flatpak-spawn permission can be avoided? This is worse that than the filesystem permissions. Or do you need this anyway even creating the desktop files manually?

hepp3n: > Or do you need this anyway even creating the desktop files manually?

In this case no. I would create desktop entry manually and if put in ~/.local/share/applications it will be visible from launchers and it can `Exec=firefox` without issues. The problem is with dynamic launcher portal, it creates entry (when created from flatpak sandbox) with `flatpak run` at start of `Exec=` and of course we don't have firefox in our sandbox so it can't launch this browser.

I need to definitely get back to previous apporach when creating file manually. Unfortunatelly I will need `rw` permissions for `xdg-data/applications`. Is it acceptable?

bbhtt: > permissions for xdg-data/applications. I

Yes that's a lot better that having flatpak-spawn access.

hepp3n: Yeah, then I will make a needed changes in the app and push new commit. Sorry, for that. As i said, I'm struggling with working with sandbox, it requires sometimes different approaches and I would like to make this app working with and without flatpak :D

hepp3n:
Ok, so I pushed 2 commits since last review. I tried to sort out those permissions and I think it looks good now.

dev.heppen.webapps.json
```diff
@@ -0,0 +1,51 @@
...
+  "command": "quick-webapps",
+  "finish-args": [
+    "--share=ipc",
+    "--socket=fallback-x11",
+    "--socket=wayland",
+    "--device=dri",
+    "--share=network",
+    "--filesystem=host-os:ro",
+    "--filesystem=xdg-data/applications",
+    "--filesystem=xdg-data/flatpak",
```
bbhtt, line 20:
```suggestion
    "--filesystem=xdg-data/flatpak",
```

this should be `xdg-data/flatpak/exports/share/applications:ro` shouldn't it? You don't install desktop files there, only read.

bbhtt:
> Ok, so I pushed 2 commits since last review. I tried to sort out those permissions and I think it looks good now.

You can open an exception PR, the last 3 permission needs exceptions https://docs.flathub.org/docs/for-app-authors/linter#exceptions

Running the manifest check will show the codes.

dev.heppen.webapps.json
bbhtt: Ah those are symlinks I forgot

bbhtt: Adding the target should fix it `xdg-data/flatpak/app:ro`, `dg-data/flatpak/exports/share/applications:ro`

hepp3n: Yeah i thought that's the reason and tried different ways to achieve this but with this permission it's just easier :P

bbhtt: you can keep `xdg-data/flatpak` but this should be :ro anyways

hepp3n: oh yeah, it works with `ro`, I don't know, I'm pretty sure I tested it before and it did not, haha. Ok, I will change this :D

hepp3n:
Ok, got it! :smile:

bbhtt:
you can start the build by commenting, in 30 minutes - 1h once the exceptions are up.

hepp3n:
Ok, let's see :D

bbhtt:
bot, build dev.heppen.webapps

bbhtt:
/merge

You can EOL rebase the old one once this is published.

PR #6139, opened 2025-02-08
Add io.github.syt0r.kanji-dojo

syt0r:

- [X] Please describe the application briefly. Japanese language learning application

There's only the following lint error, I'm not sure what to do with it but according to the docs seems like it can be ignored?
```
{
    "errors": [
        "appstream-screenshots-not-mirrored-in-ostree"
    ],
    "message": "Please consult the documentation at https://docs.flathub.org/docs/for-app-authors/linter"
}
```

### DIFF
--- a/flathub.json
+++ b/flathub.json
@@ -1,0 +1,3 @@
+{
+    "skip-arches": ["aarch64"]
+}
--- a/io.github.syt0r.kanji-dojo.yml
+++ b/io.github.syt0r.kanji-dojo.yml
@@ -1,0 +1,37 @@
+id: io.github.syt0r.kanji-dojo
+runtime: org.freedesktop.Platform
+runtime-version: '24.08'
+sdk: org.freedesktop.Sdk
+command: kanji-dojo.sh
+finish-args:
+  - --socket=x11
+  - --share=ipc
+  - --share=network
+  - --device=dri
+  - --socket=pulseaudio
+  - --filesystem=xdg-download
+  - --filesystem=~/.local/share/kanji-dojo/
+modules:
+  - name: kanji-dojo-app
+    buildsystem: simple
+    build-commands:
+      - chmod +x ./Kanji_Dojo-x86_64.AppImage
+      - ./Kanji_Dojo-x86_64.AppImage --appimage-extract
+      - mv ./squashfs-root/usr/* /app
+      - install -Dm644 app/linux/flatpak/io.github.syt0r.kanji-dojo.metainfo.xml -t /app/share/metainfo/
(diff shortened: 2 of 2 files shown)


io.github.syt0r.kanji-dojo.yml
```diff
@@ -0,0 +1,37 @@
...
+sdk: org.freedesktop.Sdk
+command: kanji-dojo.sh
+finish-args:
+  - --socket=x11
+  - --share=ipc
+  - --share=network
+  - --device=dri
+  - --socket=pulseaudio
+  - --filesystem=xdg-download
+  - --filesystem=~/.local/share/kanji-dojo/
```
hfiguiere, line 13: no. you shouldn't have to do that.

io.github.syt0r.kanji-dojo.yml
```diff
@@ -0,0 +1,37 @@
...
+modules:
+  - name: kanji-dojo-app
+    buildsystem: simple
+    build-commands:
+      - chmod +x ./Kanji_Dojo-x86_64.AppImage
+      - ./Kanji_Dojo-x86_64.AppImage --appimage-extract
```
hfiguiere, line 19: how will that work on aarch64?

why no building from source?

io.github.syt0r.kanji-dojo.yml
syt0r: Why not? I want it to use the same data as the AppImage version of my app

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+    "skip-arches": ["aarch64"]
```
hfiguiere, line 2: this is disappointing. Why not building from source instead?

io.github.syt0r.kanji-dojo.yml
hfiguiere: no

flathub.json
syt0r: I can't test it on aarch64 so that's why

Also it can probably be built from the sources, but I'm not sure how it would work if there's no network during build, it's a gradle java project that downloads all libraries and dependencies during build. I'm just using already built AppImage version that's already published by my github action because of this

hfiguiere: it works by adding everything to the manifest to be downloaded ahead of time.

Also since it is Java it shouldn'e be a problem to work cross-architecture. Afterall that the whole reason for Java (bloat) to exist.

hfiguiere: The generator for gradle
https://github.com/flatpak/flatpak-builder-tools/tree/master/gradle

syt0r: Does it means I'll have to publish these generated sources separately and use them during build?

PR #6166, opened 2025-02-13
Add org.libvips.nip4

jcupitt:
- [x] Please describe the application briefly. nip4 is an image processing spreadsheet. It's an update of nip2 for the gtk4 GUI toolkit, see https://github.com/libvips/nip2 This is a first test release -- I'd like to be able to point beta testers to a convenient linux binary. There should be a full release in a few months.

### DIFF
--- a/org.libvips.nip4.json
+++ b/org.libvips.nip4.json
@@ -334,7 +334,7 @@
                 {
                     "type" : "git",
                     "url" : "https://github.com/jcupitt/nip4.git",
-                    "tag" : "v9.0.0-11"
+                    "tag" : "v9.0.1"
                 }
             ]
         }
--- a/org.libvips.nip4.json
+++ b/org.libvips.nip4.json
@@ -334,7 +334,7 @@
                 {
                     "type" : "git",
                     "url" : "https://github.com/jcupitt/nip4.git",
-                    "tag" : "v9.0.1"
+                    "tag" : "v9.0.1-4"
                 }
             ]
         }
--- a/org.libvips.nip4.json
+++ b/org.libvips.nip4.json
@@ -11,18 +11,16 @@
         "--socket=wayland",
         "--talk-name=org.gtk.vfs.*",
         "--filesystem=xdg-run/gvfsd",
-        "--filesystem=home",
-        "--filesystem=/mnt",
-        "--filesystem=/srv"
+        "--filesystem=host"
     ],
     "cleanup" : [
         "/include",
         "/lib/libheif",
-        "/lib/x86_64-linux-gnu",
         "/lib/girepository-1.0",
         "/lib/pkgconfig",
         "/lib/cmake",
         "/lib/openjpeg-*",
+        "/lib/*.a",
         "/man",
         "/share/aclocal",
         "/share/doc",
@@ -39,31 +37,13 @@
         "/share/COPYING",
         "/share/RELEASE.txt",
         "/share/cmake",
-        "/bin/a*",
-        "/bin/b*",
-        "/bin/c*",
-        "/bin/d*",
-        "/bin/e*",
-        "/bin/f*",
-        "/bin/g*",
-        "/bin/h*",
-        "/bin/i*",
-        "/bin/l*",
-        "/bin/m*",
-        "/bin/o*",
-        "/bin/p*",
-        "/bin/r*",
-        "/bin/s*",
-        "/bin/t*",
-        "/bin/x*",
-        "/bin/vips",
-        "/bin/vipsheader",
+        "/bin/smem",
+        "/bin/fpack",
+        "/bin/funpack",
+        "/bin/speed",
         "/bin/vipsedit",
         "/bin/vipsprofile",
-        "/bin/vipsthumbnail",
-        "/bin/jxlinfo",
-        "*.la",
-        "*.a"
+        "/bin/vipsthumbnail"
     ],
     "modules" : [
 
@@ -73,7 +53,6 @@
             "buildsystem" : "cmake-ninja",
             "build-options" : {
                 "config-opts" : [
-                    "-DCMAKE_INSTALL_LIBDIR=lib",
                     "-DBUILD_TESTING=OFF",
                     "-DHWY_ENABLE_CONTRIB=OFF",
                     "-DHWY_ENABLE_EXAMPLES=OFF",
@@ -94,11 +73,6 @@
             "name" : "libdeflate",
             "builddir" : true,
             "buildsystem" : "cmake-ninja",
-            "build-options" : {
-                "config-opts" : [
-                    "-DCMAKE_INSTALL_LIBDIR=lib"
-                ]
-            },
             "sources" : [
                 {
                     "type" : "git",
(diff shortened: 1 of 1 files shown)


org.libvips.nip4.json
```diff
@@ -0,0 +1,478 @@
...
+                    "type" : "git",
+                    "url" : "https://github.com/ebiggers/libdeflate.git",
+                    "tag" : "v1.22",
+                    "commit" : "2335c047e91cac6fd04cb0fd2769380395149f15"
+                }
+            ]
+        },
+
+        {
+            "name" : "libjxl",
```
hfiguiere, line 113: 0.10.4 is in the runtime

org.libvips.nip4.json
```diff
@@ -0,0 +1,478 @@
...
+                    "type" : "git",
+                    "url" : "https://github.com/libjxl/libjxl.git",
+                    "tag" : "v0.11.0",
+                    "commit" : "4df1e9eccdf86b8df4c0c7c08f529263906f9c4f"
+                }
+            ]
+        },
+
+        {
+            "name" : "openjpeg",
```
hfiguiere, line 140: 2.3.0 is in the runtime

org.libvips.nip4.json
```diff
@@ -0,0 +1,478 @@
...
+                {
+                    "type" : "archive",
+                    "url" : "https://github.com/strukturag/libde265/releases/download/v1.0.15/libde265-1.0.15.tar.gz",
+                    "sha256" : "00251986c29d34d3af7117ed05874950c875dd9292d016be29d3b3762666511d"
+                }
+            ]
+        },
+
+        {
+            "name" : "libheif",
```
hfiguiere, line 394: libheif 1.18.2 is in the runtime

org.libvips.nip4.json
```diff
@@ -0,0 +1,478 @@
...
+    ],
+    "modules" : [
+
+        {
+            "name" : "highway",
+            "builddir" : true,
+            "buildsystem" : "cmake-ninja",
+            "build-options" : {
+                "config-opts" : [
+                    "-DCMAKE_INSTALL_LIBDIR=lib",
```
hfiguiere, line 76: you shouldn't need to set LIBDIR. make sure to have flatpak-builder 1.4.4 or the flatpak version

org.libvips.nip4.json
```diff
@@ -0,0 +1,478 @@
...
+                    "type" : "git",
+                    "url" : "https://github.com/randy408/libspng.git",
+                    "tag" : "v0.7.4",
+                    "commit" : "fb768002d4288590083a476af628e51c3f1d47cd"
+                }
+            ]
+        },
+
+        {
+            "name" : "libarchive",
```
hfiguiere, line 204: libarchive 3.7.4 is in the runtime

org.libvips.nip4.json
```diff
@@ -0,0 +1,478 @@
...
+            ]
+        },
+
+        {
+            "name" : "libdicom",
+            "builddir" : true,
+            "buildsystem" : "meson",
+            "config-opts" : [
+                "--buildtype=release",
+                "--libdir=lib",
```
hfiguiere, line 332: you shouldn't need to set meson lidir either. same as CMake.

org.libvips.nip4.json
```diff
@@ -0,0 +1,478 @@
...
+        "/bin/h*",
+        "/bin/i*",
+        "/bin/l*",
+        "/bin/m*",
+        "/bin/o*",
+        "/bin/p*",
+        "/bin/r*",
+        "/bin/s*",
+        "/bin/t*",
+        "/bin/x*",
```
bbhtt, line 58: set these in individual modules, this will break and cleanup unrelated things.

org.libvips.nip4.json
```diff
@@ -0,0 +1,478 @@
...
+    "finish-args" : [
+        "--device=dri",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--talk-name=org.gtk.vfs.*",
+        "--filesystem=xdg-run/gvfsd",
+        "--filesystem=home",
+        "--filesystem=/mnt",
+        "--filesystem=/srv"
```
bbhtt, line 16: this is `host`

jcupitt:
Thank you both very much for the review! The SDK has obviously seen a lot of new libraries since I last checked.

I'm travelling this week, I'll try to fix this next week when I'm back at my desk.

org.libvips.nip4.json
jcupitt: done, removed libde265 too

org.libvips.nip4.json
jcupitt: done!

org.libvips.nip4.json
jcupitt: I switched to builder 1.4.4 and indeed the `lib=` seems to be set by default now, thanks

jcupitt:
I included the review comments and built with builder 1.4.4 from flathub. It all seems to work, though I get `Unknown bus method ReloadConfig` at the end, I assume this is harmless.

org.libvips.nip4.json
hfiguiere: de265 isn't in the runtime because lawyers.

bbhtt: it's in the ffmpeg-full extension.

org.libvips.nip4.json
```diff
@@ -0,0 +1,361 @@
...
+                {
+                    "type" : "git",
+                    "url" : "https://github.com/ebiggers/libdeflate.git",
+                    "tag" : "v1.22",
+                    "commit" : "2335c047e91cac6fd04cb0fd2769380395149f15"
+                }
+            ],
+	    "cleanup" : [
+		"/bin/libdeflate-gunzip",
+		"/bin/libdeflate-gzip"
```
hfiguiere, line 86: shouldn't this just be `/bin`

org.libvips.nip4.json
```diff
@@ -0,0 +1,361 @@
...
+		"/bin/exrenvmap",
+		"/bin/exrheader",
+		"/bin/exrinfo",
+		"/bin/exrmakepreview",
+		"/bin/exrmaketiled",
+		"/bin/exrmanifest",
+		"/bin/exrmetrics",
+		"/bin/exrmultipart",
+		"/bin/exrmultiview",
+		"/bin/exrstdattr"
```
hfiguiere, line 209: shouldn't this just be `/bin`

jcupitt:
My vim added some hard tabs, I swapped them for spaces.

org.libvips.nip4.json
hfiguiere: that's how `cleanup` works.

jcupitt: I assumed I had to name them explicitly. I'll change this, thanks.

hfiguiere: from the man

>        cleanup (array of strings)
>           An array of file patterns that should be removed at the end. Patterns starting with / are taken to be full pathnames
>           (without the /app prefix), otherwise they just match the basename. Note that any patterns will only match files
>           installed by this module.

hfiguiere:
bot, build org.libvips.nip4

bbhtt:
> This is a first test release -- I'd like to be able to point beta testers to a convenient linux binary. There should be a full release in a few months.

So should this be merged to stable or do you intend to wait?

jcupitt:
I thought about this again -- I've updated the metainfo to stress that this is a test release, so I think this should be merged. 

I'm planning to get users in several institutions to try it out over the next few months, and an easy-to-install test build would be very helpful.

bbhtt:
bot, build org.libvips.nip4

jcupitt:
OK, let's leave this as a draft for now.

jcupitt:
I've spent some time polishing and I think this is ready for a first general release now.

org.libvips.nip4.json
```diff
@@ -0,0 +1,343 @@
...
+    "sdk" : "org.gnome.Sdk",
+    "command" : "nip4",
+    "finish-args" : [
+        "--device=dri",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--talk-name=org.gtk.vfs.*",
+        "--filesystem=xdg-run/gvfsd",
+        "--filesystem=host"
```
bbhtt, line 14: I guess I forgot to ask but why do you need `host`. Previously this was home+media+srv but why?

Does this not work with filechooser portals?

jcupitt: nip4 includes a macro language that lets you load files from any path, eg. `Image_file "/home/john/pics/x.jpg"`, meaning "Image constructor from a filename with path as argument". In turn this uses the libvips source system, which does not use portals.

jcupitt:
There have been a few small changes, mostly related to the windows build, so I updated the build file to the final one.

bbhtt:
bot, build org.libvips.nip4

PR #6177, opened 2025-02-16
Add io.github.sigmasd.VirtAudio

sigmaSd:

- [x] Please describe the application briefly. Stream Audio from remoete browsers to your computer as virtual microphones


io.github.sigmasd.VirtAudio.yml
```diff
@@ -0,0 +1,267 @@
+app-id: io.github.sigmasd.VirtAudio
+runtime: org.freedesktop.Platform
+runtime-version: "23.08"
+sdk: org.freedesktop.Sdk
+command: virtaudio
+finish-args:
+  # Display
+  - --socket=wayland
+  - --socket=x11
```
hfiguiere, line 9: it's either `x11` or `wayland` + `fallback-x11`

io.github.sigmasd.VirtAudio.yml
```diff
@@ -0,0 +1,267 @@
+app-id: io.github.sigmasd.VirtAudio
+runtime: org.freedesktop.Platform
+runtime-version: "23.08"
```
hfiguiere, line 3: current is 24.08

io.github.sigmasd.VirtAudio.yml
```diff
@@ -0,0 +1,267 @@
...
+  # Display
+  - --socket=wayland
+  - --socket=x11
+  - --share=ipc
+  # Audio access
+  - --socket=pulseaudio
+  # Network access for local server and WebSocket connections
+  - --share=network
+  # Needed for libinput
+  - --device=input
```
hfiguiere, line 16: this is no allowed, no one supports it.

sigmaSd: removed

io.github.sigmasd.VirtAudio.yml
sigmaSd: changed

io.github.sigmasd.VirtAudio.yml
sigmaSd: changed

hfiguiere:
bot, build io.github.sigmasd.VirtAudio

io.github.sigmasd.VirtAudio.yml
```diff
@@ -0,0 +1,265 @@
...
+  - --socket=x11
+  - --share=ipc
+  # Audio access
+  - --socket=pulseaudio
+  # Network access for local server and WebSocket connections
+  - --share=network
+  # Needed for mesa
+  - --device=dri
+  # Allow creating temporary files (for creating virtual mics)
+  - --filesystem=/tmp:create
```
bbhtt, line 18: there's /tmp inside the sandbox, why do you need host /tmp?

io.github.sigmasd.VirtAudio.yml
```diff
@@ -0,0 +1,265 @@
...
+        x-checker-data:
+          type: anitya
+          project-id: 20540
+          stable-only: true
+          url-template: https://www.freedesktop.org/software/libevdev/libevdev-$version.tar.xz
+    cleanup:
+      - /bin
+      - /include
+      - /lib/pkgconfig
+      - /share/man
```
bbhtt, line 39: add these to top level cleanup `*.a` too. no need to repeat.

sigmaSd: done

sigmaSd:
bot, build io.github.sigmasd.VirtAudio

sigmaSd:
bot, build io.github.sigmasd.VirtAudio

sigmaSd:
bot, build io.github.sigmasd.VirtAudio

sigmaSd:
The changes I did:
- make deno compile the app into a single executable, this require 2 patches, that I carry here
- vendor registry files, because these files are dynamic they get updated with new versions of packages, which don't care about in this case
- merge common cleanups and make them global

io.github.sigmasd.VirtAudio.yml
bbhtt: try setting `--env=TMPDIR=/var/tmp`.

registries/slint-ui-binary-darwin-arm64-registry.json
bbhtt, line 1: are these supposed to be here?

sigmaSd: Deno expects these when analyzing npm modules, I tried initially downloading them in flatpak, by specifing the url like this https://registry.npmjs.org/undici-types

This though is not possible because these urls are dynmaic, and they get updated from time to time (with new version of the package) so the sha256 will be different each time

Deno only wants a snapshot at a given time that include the used version

So the only idea I found is fetching and vendoring them here

sigmaSd: i add a commit that decrease their size a lot, and makes them just a couple of lines and human readable

sigmaSd:
bot, build io.github.sigmasd.VirtAudio

sigmaSd:
bot, build io.github.sigmasd.VirtAudio

io.github.sigmasd.VirtAudio.yml
sigmaSd: I just noticed this comment, no unfortunately changing to that flag results in the same problem, the audio is not passed to the microphone even though there are no errors when writing to it or errors from ffmpeg or errors in the logs

 I forgot to paste the whole snippet, this is the part that uses the fifo file
```ts
    const result = await new Deno.Command("pactl", {
      args: [
        "load-module",
        "module-pipe-source",
        `source_name=${micName}`,
        `file=${virtualMicPipePath}`,
        "format=s16le",
        "rate=48000",
        "channels=2",
      ],
    }).output();
```

sigmaSd: I'm not sure how to true is this, but asked claude:

>The filesystem permission is needed because PulseAudio needs to access the FIFO file from outside the Flatpak sandbox while your application needs to access it from inside the sandbox. This cross-sandbox communication is what necessitates the filesystem permission.

sigmaSd: I also tried changing the path to a location inside the sandbox but it have the same issue
```ts
    const virtualMicPipePath = `${
      Deno.env.get("XDG_RUNTIME_DIR")
    }/virtual_mic_pipe${this.#id}`;
```

bbhtt: can you check what file it is trying to access? Trying running strace on it. This is proabably influenced by some host puleseaudio configuration that places some socket in `/tmp` only.

sigmaSd: /tmp is not special, any path can work as long as I have permission to it from the sandbox, this also works for example
```ts
    const virtualMicPipePath = `${
      Deno.env.get("XDG_RUNTIME_DIR")
    }/mics/virtual_mic_pipe${this.#id}`;
```
```yml
  - --filesystem=xdg-run/mics:create
```

I think claude have the correct insight, this  cross read/write from inside and outside the sandbox requires this permission

bbhtt: that's better than `/tmp`.

bbhtt: does it need to be :create? You can create the directory in your code and give :ro only?

sigmaSd: Yes :ro works as well, though I don't know why it works, because I do write audio data to the virtual mics inside that directory

sigmaSd: is there a preference if I should choose /tmp or XDG_RUNTIME_DIR, or it doesn't matter I just choose one?

bbhtt: use `xdg-run/mics:ro`

sigmaSd: Turns out :create is needed, it only worked earlier because the folder is already created, I want to change the folder name to virtual_mics, and it turns out it can't create it even tough I do use mkdir  in my app

So I'll just change to xdg-run/virtual_mics:create

sigmaSd: I see now that the app create the directory inside the sandbox only, but the directory need to exist on the host for pulseaudio to work and this is achieved with :create flag

sigmaSd: done, changed to xdg-run/virtual_mics:create

sigmaSd:
bot, build io.github.sigmasd.VirtAudio

bbhtt:
can you try screen recording to show how it works?

sigmaSd:
I updated the app to slint nightly, which fixes wayland

@bbhtt I added a demo to the readme https://github.com/sigmaSd/virtAudio?tab=readme-ov-file#demo

sigmaSd:
https://github.com/sigmaSd/virtAudio?tab=readme-ov-file#demo

sigmaSd:
bot, build io.github.sigmasd.VirtAudio

sigmaSd:
bot, build io.github.sigmasd.VirtAudio

PR #6189, opened 2025-02-19
Add com.saunasim.saunasim

997R8V10:

[SaunaSim](https://saunasim.com) is a tool to be used on the [VATSIM](http://vatsim.net) network to help train virtual air traffic controllers. SaunaSim simulates aircraft performance, behavior, FMS (Flight Management System) and more to accurately replicate real aircraft behavior. The instructor controls the aircraft using commands. The aircraft positions are sent to the VATSIM FSD server so that they can show up on the student's radar display. It contains 2 parts: [sauna-ui](https://github.com/sauna-sim/sauna-ui) and [sauna-api](https://github.com/sauna-sim/sauna-api). The API is a ASP.NET Core 8.0 application that is bundled with the UI and handles the core simulation code as well as FSD interaction. The UI is built using Tauri and React to handle the user interface.

### DIFF
--- a/com.saunasim.saunasim.yml
+++ b/com.saunasim.saunasim.yml
@@ -32,7 +32,7 @@ modules:
         dest-filename: SaunaSim.deb
 
     build-commands:
-      - ar -x *.deb
+      - ar -x SaunaSim.deb
       - tar -xf data.tar.gz
       - 'install -Dm755 usr/bin/SaunaSim /app/bin/SaunaSim'
       - install -Dm644 usr/share/applications/SaunaSim.desktop /app/share/applications/SaunaSim.desktop
--- a/com.saunasim.saunasim.yml
+++ b/com.saunasim.saunasim.yml
@@ -1,0 +1,42 @@
+id: com.saunasim.saunasim
+
+runtime: org.gnome.Platform
+runtime-version: '46'
+sdk: org.gnome.Sdk
+
+rename-icon: SaunaSim
+rename-desktop-file: SaunaSim.desktop
+rename-appdata-file: SaunaSim.metainfo.xml
+
+command: SaunaSim
+finish-args:
+  - --socket=wayland # Permission needed to show the window
+  - --socket=fallback-x11 # Permission needed to show the window
+  - --device=dri # OpenGL, not necessary for all projects
+  - --share=ipc # To interface with Tauri Backend
+  - --share=network # To connect to VATSIM Network
+
+modules:
+  - name: binary
+    buildsystem: simple
+    sources:
+      - type: file
+        url: https://github.com/sauna-sim/sauna-ui/releases/download/v0.22.17/SaunaSim_0.22.17_amd64.deb
+        sha256: e1c271534eaa3b55b862fa91d22eb0d6d898ba7c2468d9629b3d0be7088b3f0d
+        only-arches: [x86_64]
+      - type: file
+        url: https://github.com/sauna-sim/sauna-ui/releases/download/v0.22.17/SaunaSim_0.22.17_arm64.deb
+        sha256: 1a833f99762f69d1a164c77098e9b4a181b4fa1eae60f11a6d733d5991cc7606
+        only-arches: [aarch64]
+
+    build-commands:
+      - ar -x *.deb
+      - tar -xf data.tar.gz
+      - 'install -Dm755 usr/bin/SaunaSim /app/bin/SaunaSim'
+      - install -Dm644 usr/share/applications/SaunaSim.desktop /app/share/applications/SaunaSim.desktop
+      - install -Dm644 usr/share/icons/hicolor/128x128/apps/SaunaSim.png /app/share/icons/hicolor/128x128/apps/SaunaSim.png
+      - install -Dm644 usr/share/icons/hicolor/32x32/apps/SaunaSim.png /app/share/icons/hicolor/32x32/apps/SaunaSim.png
+      - install -Dm644 usr/share/icons/hicolor/256x256@2/apps/SaunaSim.png /app/share/icons/hicolor/256x256@2/apps/SaunaSim.png
+      - install -Dm644 usr/share/metainfo/SaunaSim.metainfo.xml /app/share/metainfo/SaunaSim.metainfo.xml
+      - mkdir -p /app/lib/SaunaSim
+      - cp -r usr/lib/SaunaSim /app/lib
--- a/com.saunasim.saunasim.yml
+++ b/com.saunasim.saunasim.yml
@@ -1,7 +1,7 @@
 id: com.saunasim.saunasim
 
 runtime: org.gnome.Platform
-runtime-version: '46'
+runtime-version: '47'
 sdk: org.gnome.Sdk
 
 rename-icon: SaunaSim
@@ -17,7 +17,7 @@ finish-args:
   - --share=network # To connect to VATSIM Network
 
 modules:
-  - name: binary
+  - name: sauna-sim
     buildsystem: simple
     sources:
       - type: file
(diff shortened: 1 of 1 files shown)


997R8V10:
bot, build com.saunasim.saunasim

com.saunasim.saunasim.yml
```diff
@@ -0,0 +1,42 @@
+id: com.saunasim.saunasim
+
+runtime: org.gnome.Platform
+runtime-version: '46'
```
hfiguiere, line 4: current is 47

com.saunasim.saunasim.yml
```diff
@@ -0,0 +1,42 @@
...
+command: SaunaSim
+finish-args:
+  - --socket=wayland # Permission needed to show the window
+  - --socket=fallback-x11 # Permission needed to show the window
+  - --device=dri # OpenGL, not necessary for all projects
+  - --share=ipc # To interface with Tauri Backend
+  - --share=network # To connect to VATSIM Network
+
+modules:
+  - name: binary
```
hfiguiere, line 20: please use a less generic name.

com.saunasim.saunasim.yml
```diff
@@ -0,0 +1,42 @@
...
+    buildsystem: simple
+    sources:
+      - type: file
+        url: https://github.com/sauna-sim/sauna-ui/releases/download/v0.22.17/SaunaSim_0.22.17_amd64.deb
```
hfiguiere, line 24: why not building from source?

com.saunasim.saunasim.yml
hfiguiere: does it really need the GNOME runtime?

com.saunasim.saunasim.yml
997R8V10: The main code is open source however I am using certain API keys/secrets that are stored separately. Unfortunately, VATSIM (The network that the program interfaces with) requires some secret information such as API keys/etc not be made open source. Currently the application consumes an additional dynamic library at runtime that contains these API Keys. In order for them to be able to access the VATSIM network, the build must be built on the organization CI so it has access to the private API keys project.

The app will work without these keys but in a reduced capacity.

hfiguiere: so it's not open source?

997R8V10: The app itself is, the API keys are not.

com.saunasim.saunasim.yml
997R8V10: Done

com.saunasim.saunasim.yml
997R8V10: If the license for the FlatPak needs to be changed, I can change it to a more restrictive license.

hfiguiere:
bot, build com.saunasim.saunasim

com.saunasim.saunasim.yml
```diff
@@ -0,0 +1,42 @@
...
+        sha256: e1c271534eaa3b55b862fa91d22eb0d6d898ba7c2468d9629b3d0be7088b3f0d
+        only-arches: [x86_64]
+      - type: file
+        url: https://github.com/sauna-sim/sauna-ui/releases/download/v0.22.17/SaunaSim_0.22.17_arm64.deb
+        sha256: 1a833f99762f69d1a164c77098e9b4a181b4fa1eae60f11a6d733d5991cc7606
+        only-arches: [aarch64]
```
bbhtt, line 30:
```suggestion
        only-arches: [x86_64]
        dest-filename: SaunaSim.deb
      - type: file
        url: https://github.com/sauna-sim/sauna-ui/releases/download/v0.22.17/SaunaSim_0.22.17_arm64.deb
        sha256: 1a833f99762f69d1a164c77098e9b4a181b4fa1eae60f11a6d733d5991cc7606
        only-arches: [aarch64]
        dest-filename: SaunaSim.deb
```

com.saunasim.saunasim.yml
```diff
@@ -0,0 +1,42 @@
...
+
+    build-commands:
+      - ar -x *.deb
```
bbhtt, line 33:
```suggestion
      - ar -x SaunaSim.deb
```

bbhtt:
bot, build com.saunasim.saunasim

bbhtt:
can you provide a screenrecord of how the app works? I'm not really experienced about air traffic simulators.

bbhtt:
>  If not I can try to spool up a VM or just wait until I get home.

Sure, let's wait.

997R8V10:
> can you provide a screenrecord of how the app works? I'm not really experienced about air traffic simulators.

OK here's a quick ~5 min video I recorded using the Flatpak on Ubuntu to explain how the app is meant to be used.

The video was obviously way to big to upload onto GitHub so I have uploaded it onto Streamable temporarily.
[SaunaSim Demo Video on Streamable](https://streamable.com/tvkx75)

Let me know if you have any more questions and I'd be happy to explain further.

bbhtt:
Thanks. Can you also point where you are using the api keys? in the source code.

997R8V10:
> Thanks. Can you also point where you are using the api keys? in the source code.

It's in `sauna-api`. Here's the [link to the C# class that handles it](https://github.com/sauna-sim/sauna-api/blob/master/sauna-api/Utilities/PrivateInfoLoader.cs)

bbhtt:
Looks like this needs a private library `libsauna_vatsim_private`? Where is that coming from?

997R8V10:
That is a private small rust library I wrote that has the API keys slightly obfuscated at VATSIM and Navigraph's request. I wrote an algorithm there to try to store it in a way that didn't make it super easy to reverse engineer. That's the main reason I can't build from source. The app will still function if the library is missing at runtime and will just not allow connections to VATSIM specific servers or Navigraph. They can still connect to a custom server and use their own nav data. But for the production build the GitHub CI will bundle the private library as part of the final build.

bbhtt:
> That is a private small rust library I wrote that has the API keys slightly obfuscated at VATSIM and Navigraph's request.

I do believe that you can't mark the  license it as GPL 3.0-only in metainfo then. It requires that this private library is open source too.

997R8V10:
> I do believe that you can't mark the license it as GPL 3.0-only in metainfo then. It requires that this private library is open source and under a compatible license too.

That makes sense. I’m guessing that applies even if the library is only loaded during runtime and not required for compilation?

> Would it be possible to open source this? You can still keep the API keys private by loading them through GitHub secrets when building via your CI.

Personally, I couldn’t care less even if the API keys were stored in plaintext, but the owners of the actual APIs want some assurances that someone isn’t going to be able to acquire those API keys for their own use through some reverse engineering. Unfortunately, loading it through a secret during building would defeat the purpose of obfuscating it in the library as they could easily disassemble the private library. VATSIM had some issues in the past with other devs for clients storing their API keys in plaintext and it being reverse engineered. Of course there’s no real way to stop someone from definitively reverse engineering it anyways, but VATSIM seem happy with the current method I’m using.

If that’s not acceptable, I’ll just change the license in the metainfo to a more restrictive one since the app also has to pass VATSIM scrutineering and they weren’t happy when I was originally loading it as plaintext.

bbhtt:
> That makes sense. I’m guessing that applies even if the library is only loaded during runtime and not required for compilation?

I think runtime loading is a bit of gray area in the license and we can't do a full legal review here. I won't block on this. This is kind of like the case described here https://www.gnu.org/licenses/gpl-faq.html#GPLPlugins . Ideally you'd use a more permissive license than GPL but I see 3-4 people have contributions so relicensing is going to be a bit of a pain.

997R8V10:
> > That makes sense. I’m guessing that applies even if the library is only loaded during runtime and not required for compilation?
> 
> 
> 
> I think runtime loading is a bit of gray area in the license and we can't do a full legal review here. I won't block on this. This is kind of like the case described here https://www.gnu.org/licenses/gpl-faq.html#GPLPlugins . Ideally you'd use a more permissive license than GPL but I see 3-4 people have contributions so relicensing is going to be a bit of a pain.

I know all 4 of them personally so we will have a discussion on this and try to figure out the best way to go with the license to stay out of legal trouble.

PR #6202, opened 2025-02-22
Add io.github.Faithlife_Community.OuDedetai

ctrlaltf24:

- [X] Please describe the application briefly. An installer for Logos (bible study software) using wine as a wrapper. This software is a container around the actual logos software (3rd party)
- [X] I am an author/developer/upstream contributor to the project. If not, I contacted upstream developers about this submission. **Link:** https://github.com/FaithLife-Community/OuDedetai @n8marti and @thw26 should be added too


hfiguiere:
Packaging Windows app with Wine is not suitable for Flathub.

ctrlaltf24:
Why not may I ask? There's already a wine flatpak, this just builds on that. To the end user it's an application just like any other

ctrlaltf24:
Let me demonstrate how it works:
1. Flatpak is launched, opening our installer GUI (first party)
installer
2. When "Install" is hit, it configures the wine prefix, does all the registry edits it needs, etc (all without additional user input, they already told us to install). It also downloads the 3rd party Windows MSI installer from the 3rd-party's website
logos running
3. It then runs said MSI inside of the newly configured wine prefix (3rd party installer)
msi
4. After install is complete, user is presented with a "Run Logos" button which allows them to launch the windows EXE inside of the configured wine prefix
Run Logos


Known areas for improvement:
- I'd like to have a .desktop file for the installed app that installs if needed (prompting for EULA), then runs. We'd need to do some work on the 1st party app to make this happen, but it would feel a lot more "flatpak" to the end user and accurately expresses their intent. I understand if you want to hold off before we have this. To clarify this .desktop would do steps 2,3 and 4 resulting in the 3rd party application being run. Ideally a dialog would be shown showing progress while it installs before it's run.

bbhtt:
We can accept Windows app packaged through wine only if they are submitted and maintained officially by upstream. Is that going to be the case here?

appdata.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="console-application">
```
bbhtt, line 2: why is this console app?

appdata.xml
```diff
@@ -0,0 +1,31 @@
...
+    <id>io.github.Faithlife-Community.OuDedetai</id>
+
```
bbhtt, line 4: this is missing a lot of tags. You can go through and start adding them one by one https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines#id

appdata.xml
bbhtt, line 1: this should be at https://github.com/FaithLife-Community/OuDedetai

icons/oudedetai.png
bbhtt, line 1: icon should be at https://github.com/FaithLife-Community/OuDedetai

io.github.Faithlife_Community.OuDedetai.yml
```diff
@@ -0,0 +1,106 @@
...
+- org.freedesktop.Platform.GL32
+- org.freedesktop.Platform.GL32.Debug
+- org.freedesktop.Platform.VAAPI.Intel.i386
+- org.freedesktop.Platform.ffmpeg-full
+- org.freedesktop.Platform.ffmpeg_full.i386
+- org.winehq.Wine.gecko
+- org.winehq.Wine.mono
+- org.winehq.Wine.DLLs
+
+appstream-compose: false
```
bbhtt, line 21:
```suggestion

```

io.github.Faithlife_Community.OuDedetai.yml
```diff
@@ -0,0 +1,106 @@
...
+      - type: file
+        path: ./icons/verbum.png
+        dest-filename: verbum.png
+      - type: file
+        dest-filename: oudedetai.desktop
+        path: ./oudedetai.desktop
+      - type: file
+        dest-filename: appdata.xml
+        path: ./appdata.xml
+finish-args:
```
bbhtt, line 96: this should be before modules.

io.github.Faithlife_Community.OuDedetai.yml
```diff
@@ -0,0 +1,106 @@
...
+  - --share=ipc
+  - --socket=x11
+  - --socket=pulseaudio
+  - --share=network
+  - --allow=multiarch
+  - --allow=devel
+  - --system-talk-name=org.freedesktop.UDisks2
+  - --system-talk-name=org.freedesktop.NetworkManager
```
bbhtt, line 104: why are these two needed?

io.github.Faithlife_Community.OuDedetai.yml
```diff
@@ -0,0 +1,106 @@
...
+      - type: file
+        dest-filename: oudedetai.desktop
+        path: ./oudedetai.desktop
+      - type: file
+        dest-filename: appdata.xml
+        path: ./appdata.xml
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --socket=pulseaudio
```
bbhtt, line 99: why is pulseaudio needed?

io.github.Faithlife_Community.OuDedetai.yml
```diff
@@ -0,0 +1,106 @@
...
+  - --share=network
```
bbhtt, line 100: why is network needed?

oudedetai.desktop
bbhtt, line 1: desktop file should be upstream

oudedetai.desktop
```diff
@@ -0,0 +1,11 @@
+[Desktop Entry]
+Name=Ou Dedetai
+GenericName=FaithLife Wine App Installer
+Comment=Installs and manages either Logos or Verbum via Wine and Flatpak. Community supported.
+Exec=/app/bin/entrypoint.sh
+Icon=/app/share/icons/hicolor/512x512/apps/oudedetai.png
```
bbhtt, line 6: these should not be full paths.

bbhtt: also icon name must match Flatpak ID and without the png extension.

appdata.xml
ctrlaltf24: as in in that repo?

ctrlaltf24:
> We can accept Windows app packaged through wine only if they are submitted and maintained officially by upstream. Is that going to be the case here?

No, however if you notice there is no MSI here, technically this is a packaging of our third party installer (which we own) that sets up a Windows app within the sandbox.

May I ask the rational behind that policy? Mostly curious, and don't want to step on any toes, however the software that's packaged here is owned by us (the installer that is), so it may be a slightly different case than normal.

io.github.Faithlife_Community.OuDedetai.yml
ctrlaltf24: I'll try without some of these permissions, it started off with a copy/paste from the wine flatpak (with a couple I know I can remove)

appdata.xml
ctrlaltf24: Good catch

oudedetai.desktop
ctrlaltf24: How can this desktop file be in upstream if the icon path uses the flatpak id?

bbhtt: > if the icon path uses the flatpak id?

edit it with `desktop-file-edit` in the manifest.

bbhtt:
> May I ask the rational behind that policy? Mostly curious, and don't want to step on any toes, however the software that's packaged here is owned by us (the installer that is), so it may be a slightly different case than normal.

Because Flathub is primarily meant for Linux apps and windows apps through wine breaks often. Without upstream support it will turn into a poor packaging quickly once accepted.

appdata.xml
bbhtt: yes or wherever is upstream of the app / installer.

oudedetai.desktop
hfiguiere: > How can this desktop file be in upstream if the icon path uses the flatpak id?

You can use the same convention upstream. Or can use the original **icon name** and use `rename-icon` in the manifest that will deal with it (that mean icons are also installed with their original name). As said never a path and never the file extension

ctrlaltf24:
> Because Flathub is primarily meant for Linux apps and windows apps through wine breaks often. Without upstream support it will turn into a poor packaging quickly once accepted.

A reasonable concern, our community has been supporting this app since [Dec 2019](https://github.com/ferion11/LogosLinuxInstaller/commit/809a93189ffdc6a054092f9abd1e4f548fe0f0bc) and has a silver rating on [winehq's appdb](https://github.com/ferion11/LogosLinuxInstaller/commit/809a93189ffdc6a054092f9abd1e4f548fe0f0bc).

We've already been supporting this app via installing directly on the system with great success and hundreds of users, flatpak would enable us to have an even more consistent environment for an even better user experience (easier to test flatpak and publish on all system than spin up ~5-6 VMs and test in each as we're doing now).

If quality degrades, you'd be perfectly justified in yanking later. I see from your comments there's a fair bit of improvement I could make to this PR for flatpak packaging, and will get on that shortly, but the base app is well traveled and we have an active team resolving issues as they come up.

bbhtt:
I'd rather block the submission first pending official maintenance that going through the trouble of yanking it later - which is an unhappy outcome for existing users and maintainers too to see their package removed

hfiguiere:
so in the last 6 years you could have ported them properly to Linux and chose not to.

And now you'll put part of the burden onto the shoulder of volunteers that are already handling a large amount of software that is Linux native.

Understood.

PR #6203, opened 2025-02-22
Add eu.lucaweiss.lpa_gtk

Ferenc-:

- [x] Please describe the application briefly. A GTK eSIM card manager application
- [x] I am an author/developer/upstream contributor to the project. If not, I contacted upstream developers about this submission. **https://codeberg.org/lucaweiss/lpa-gtk/issues/8#issuecomment-2839776**


eu.lucaweiss.lpa-gtk.yaml
```diff
@@ -0,0 +1,82 @@
...
+runtime-version: "47"
+sdk: org.gnome.Sdk
+command: lpa-gtk
+finish-args:
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --socket=pcsc
+  - --share=ipc
+  - --device=dri
+cleanup: []
```
hfiguiere, line 12:
```suggestion
```

eu.lucaweiss.lpa-gtk.yaml
```diff
@@ -0,0 +1,82 @@
...
+modules:
+  - lib/blueprint-compiler.yaml
```
hfiguiere, line 14: this should be inline here.

desktopfile-icon.patch
hfiguiere, line 1: Just use the command `desktop-file-edit` to change the icon name.

metainfo-rename.patch
```diff
@@ -0,0 +1,29 @@
...
+rename to data/eu.lucaweiss.lpa-gtk.metainfo.xml
+index 15cbc53..bb67209 100644
+--- a/data/eu.lucaweiss.lpa_gtk.metainfo.xml
++++ b/data/eu.lucaweiss.lpa-gtk.metainfo.xml
+@@ -4,7 +4,7 @@ SPDX-FileCopyrightText: 2024 Luca Weiss
+ SPDX-License-Identifier: CC-BY-SA-4.0
+ -->
+ <component type="desktop-application">
+-    <id>eu.lucaweiss.lpa_gtk</id>
++    <id>eu.lucaweiss.lpa-gtk</id>
```
hfiguiere, line 13: why using a different id?

metainfo-rename.patch
```diff
@@ -0,0 +1,29 @@
...
+diff --git a/data/meson.build b/data/meson.build
+index 6afca15..6aaa738 100644
+--- a/data/meson.build
++++ b/data/meson.build
+@@ -14,7 +14,7 @@ if ascli_exe.found()
+   )
+ endif
+ 
+-install_data('eu.lucaweiss.lpa_gtk.metainfo.xml',
++install_data('eu.lucaweiss.lpa-gtk.metainfo.xml',
```
hfiguiere, line 26: you don't need to do that. You can put `rename-appdata-file` in the manifest if really needed. But then as mentionned prerviously, why using a different id?

lib/blueprint-compiler.yaml
```diff
@@ -0,0 +1,12 @@
...
+buildsystem: meson
+cleanup:
+  - '*'
+sources:
+  - type: git
+    url: https://gitlab.gnome.org/jwestman/blueprint-compiler
+    tag: v0.16.0
+    x-checker-data:
+      type: git
+      tag-pattern: ^v([\d.]+)$
```
hfiguiere, line 11: does this really need to be updated? Given that it is a build dependency and that it might actually break.

eu.lucaweiss.lpa-gtk.yaml
```diff
@@ -0,0 +1,82 @@
...
+    build-commands:
+      - for s in {32,64,128,256,512}; do ffmpeg -i eSTK_me.jpeg
+        -s "${s}x${s}" -update true "eSTK_me_${s}x${s}.png";
+        install -Dm644 "eSTK_me_${s}x${s}.png" "${FLATPAK_DEST}/share/icons/hicolor/${s}x${s}/apps/${FLATPAK_ID}.png";
+        done;
+  - name: lpa-gtk
+    sources:
+      - type: git
+        url: https://codeberg.org/lucaweiss/lpa-gtk
+        branch: main
```
hfiguiere, line 56: NEVER a git branch.

eu.lucaweiss.lpa-gtk.yaml
```diff
@@ -0,0 +1,82 @@
...
+    sources:
+      - type: archive
+        url: https://github.com/estkme-group/lpac/archive/refs/tags/v2.2.1.tar.gz
+        sha256: 3d87080a625b10430eebb82f89e2d24e16a84a8435a9c40b3718fd88c82028ba
+        x-checker-data:
+          type: anitya
+          project-id: 371628
+          stable-only: true
+          url-template: https://github.com/estkme-group/lpac/archive/refs/tags/v$version.tar.gz
+  - name: lpac-logo # As long as lpa-gtk doesn't have it's own
```
hfiguiere, line 40: this should be done upstream then.

eu.lucaweiss.lpa-gtk.yaml
```diff
@@ -0,0 +1,82 @@
...
+#         version-query: '.[0].name'
+      - type: patch
+        path: ./desktopfile-icon.patch
+      - type: patch
+        path: ./metainfo-compulsory-release.patch
+      - type: patch
+        path: ./metainfo-screenshot.patch
+      - type: patch
+        path: ./metainfo-rename.patch
+    buildsystem: simple
```
hfiguiere, line 77: no. this should be `buildsystem: meson`

eu.lucaweiss.lpa-gtk.yaml
Ferenc-: Removed

eu.lucaweiss.lpa-gtk.yaml
Ferenc-: Inlined

lib/blueprint-compiler.yaml
Ferenc-: Unpatched build dependencies can also get you pwned. À la  xz utils backdoor...
If a PR with a braking change is ever raised by the checker, then that obviusly won't be merged until upstream adapts.

eu.lucaweiss.lpa-gtk.yaml
Ferenc-: Done

lib/blueprint-compiler.yaml
hfiguiere: Much like a pwned one can be automatically grabbed by this unecessary update. That would have accelerated the xz backdoor.

Ferenc-: Nothing is grabbed automatically.
The notification and the pr is automatic. The decision to test, accept and merge is manual.

metainfo-rename.patch
z3ntu: I've pushed a commit to the source repo now to rename [...]lpa-gtk.desktop to [...]lpa_gtk.desktop due to this in the appstream spec:

> While hyphens are allowed for legacy compatibility, their usage is strongly discouraged to ensure interoperability of the AppStream ID with other tools such as D-Bus (and thereby making the ID more generic and useful).

z3ntu:
This is currently not functional:
```
$ flatpak run eu.lucaweiss.lpa-gtk

(eu.lucaweiss.lpa-gtk:2): Gtk-CRITICAL **: 09:56:36.712: gtk_editable_set_text: assertion 'text != NULL' failed
No APDU driver found
No APDU driver found
```

It's probably missing to build the qmi_qrtr driver for lpac. We also need access to the qrtr socket then, not sure if that's already handled in the permissions.

Also with https://codeberg.org/lucaweiss/lpa-gtk/issues/1 and https://codeberg.org/lucaweiss/lpa-gtk/issues/8 how can this handle access to /dev/tty* which we will need to communicate with those modems? Probably some privileged system helper will be needed there, and we also need to do something like `sudo mmcli -m any --inhibit` , or allow the user to call InhibitDevice as regular user, see https://codeberg.org/lucaweiss/lpa-gtk/issues/1#issuecomment-2066461

Ferenc-:
@z3ntu  Thanks for response here, an as always for the valued feedback!

> qmi_qrtr driver for lpac.

I will look into it, should be doable.

> We also need access to the qrtr socket then, not sure if that's already handled in the permissions.

Not yet, I will add it soon.

>  /dev/tty*

As long as there is no [serial port portal yet](https://github.com/flatpak/xdg-desktop-portal/issues/229), I think it will have to be `--dev=all` see [comment from the flatpak lead](https://github.com/flatpak/flatpak/issues/3959#issuecomment-727804466).

> and we also need to do something like sudo mmcli -m any --inhibit 

There are other flatpaks, which package `mmcli` like [satellite](https://github.com/flathub/page.codeberg.tpikonen.satellite/blob/master/page.codeberg.tpikonen.satellite.yaml#L19) so we should be able to do that as well, but I think that might not be needed. I have the hunch that the Calls flatpak probably  [does that too by simply talking to modem manager via dbus](https://github.com/flathub/org.gnome.Calls/blob/master/org.gnome.Calls.yaml#L12).

z3ntu:
> > We also need access to the qrtr socket then, not sure if that's already handled in the permissions.
> 
> Not yet, I will add it soon.

I actually don't think this is possible without changes in flatpak: https://github.com/flatpak/flatpak/blob/fa252b865a9305a5ff01564140a7273f05503ece/common/flatpak-run.c#L1957-L1971

> > /dev/tty*
> 
> As long as there is no [serial port portal yet](https://github.com/flatpak/xdg-desktop-portal/issues/229), I think it will have to be `--dev=all` see [comment from the flatpak lead](https://github.com/flatpak/flatpak/issues/3959#issuecomment-727804466).

I don't see how the unprivileged flatpak can access those devices which are owned by root though? Does flatpak have some kind of privilege escalation?

> > and we also need to do something like sudo mmcli -m any --inhibit
> 
> There are other flatpaks, which package `mmcli` like [satellite](https://github.com/flathub/page.codeberg.tpikonen.satellite/blob/master/page.codeberg.tpikonen.satellite.yaml#L19) so we should be able to do that as well, but I think that might not be needed. I have the hunch that the Calls flatpak probably [does that too by simply talking to modem manager via dbus](https://github.com/flathub/org.gnome.Calls/blob/master/org.gnome.Calls.yaml#L12).

Same, a regular user (to my knowledge) can't call this method, we either need to allow it through dbus system policy, or run it as root.

metainfo-rename.patch
Ferenc-: Removed.

eu.lucaweiss.lpa_gtk.yaml
```diff
@@ -0,0 +1,87 @@
...
+#         version-query: '.[0].name'
+      - type: patch
+        path: ./desktopfile-icon.patch
+      - type: patch
+        path: ./metainfo-compulsory-release.patch
+      - type: patch
+        path: ./metainfo-screenshot.patch
+    buildsystem: meson
+    build-commands:
+      - mkdir -p ${FLATPAK_DEST}/share/glib-2.0/schemas ${FLATPAK_DEST}/share/icons/hicolor
```
hfiguiere, line 87: why is this necessary?

eu.lucaweiss.lpa_gtk.yaml
```diff
@@ -0,0 +1,87 @@
...
+#         # version-query: .tag_name
+#         # As long as there are no releases:
+#         url: https://codeberg.org/api/v1/repos/lucaweiss/lpa-gtk/tags/
+#         version-query: '.[0].name'
+      - type: patch
+        path: ./desktopfile-icon.patch
+      - type: patch
+        path: ./metainfo-compulsory-release.patch
+      - type: patch
+        path: ./metainfo-screenshot.patch
```
hfiguiere, line 84:
```suggestion
      - type: patch
        paths: 
          - metainfo-compulsory-release.patch
          - metainfo-screenshot.patch
```

eu.lucaweiss.lpa_gtk.yaml
```diff
@@ -0,0 +1,87 @@
...
+    sources:
+      - type: archive
+        url: https://github.com/estkme-group/lpac/archive/refs/tags/v2.2.1.tar.gz
+        sha256: 3d87080a625b10430eebb82f89e2d24e16a84a8435a9c40b3718fd88c82028ba
+        x-checker-data:
+          type: anitya
+          project-id: 371628
+          stable-only: true
+          url-template: https://github.com/estkme-group/lpac/archive/refs/tags/v$version.tar.gz
+  - name: lpac-logo # As long as lpa-gtk doesn't have it's own
```
hfiguiere, line 50:
```suggestion
  - name: lpac-logo # As long as lpa-gtk doesn't have its own
```

desktopfile-icon.patch
hfiguiere: still unresolved

eu.lucaweiss.lpa_gtk.yaml
Ferenc-: Otherwise `meson` fails:
````
Running custom install script '/usr/bin/glib-compile-schemas /app/share/glib-2.0/schemas'
--- stdout ---

--- stderr ---
Error opening directory “/app/share/glib-2.0/schemas”: No such file or directory

FAILED: install script '/usr/bin/glib-compile-schemas /app/share/glib-2.0/schemas' failed with exit code 1.
FAILED: meson-internal__install 
````

File: desktopfile-icon.patch
Comment (Ferenc-) [reply desktopfile-icon.patch:1]: Done

File: eu.lucaweiss.lpa_gtk.yaml (hunk 9 above)
Comment (bbhtt) [reply eu.lucaweiss.lpa_gtk.yaml:86]: buildsystem and build-commands won't work together. If you use a named buildsystem, build-commands come from that.

File: eu.lucaweiss.lpa_gtk.yaml [hunk 12]
```diff
@@ -0,0 +1,86 @@
...
+    build-commands:
+      - for s in {32,64,128,256,512}; do ffmpeg -i eSTK_me.jpeg
+        -s "${s}x${s}" -update true "eSTK_me_${s}x${s}.png";
+        install -Dm644 "eSTK_me_${s}x${s}.png" "${FLATPAK_DEST}/share/icons/hicolor/${s}x${s}/apps/${FLATPAK_ID}.png";
+        done;
+  - name: lpa-gtk
+    sources:
+      - type: git
+        url: https://codeberg.org/lucaweiss/lpa-gtk
+        branch: main
```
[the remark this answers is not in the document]
hfiguiere: see https://github.com/flathub/flathub/pull/6203#discussion_r1966543814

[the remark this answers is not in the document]
Ferenc-: I fully agree. Currently waiting for upstream to tag the repo.

eu.lucaweiss.lpa_gtk.yaml
Ferenc-: OK, I have raised a PR upstream about the schema compilation.
The desktop-file-edit should not be needed once either https://gitlab.gnome.org/Teams/Design/app-icon-requests/-/issues/88 or an equivalent is done.

Ferenc-: > As for the desktop-file-edit it should be in post-install

Apparently `appstreamcli compose` runs before `post-install` and it fails with `gui-app-without-icon`, so trying to edit the icon property in `post-install` appears to be too late.

hfiguiere: > Apparently appstreamcli compose runs before post-install

no it doesn't.

eu.lucaweiss.lpa_gtk.yaml
```diff
@@ -0,0 +1,81 @@
...
+#       x-checker-data:
+#         type: json
+#         # Once there are releases:
+#         # url: https://codeberg.org/api/v1/repos/lucaweiss/lpa-gtk/releases/latest
+#         # version-query: .tag_name
+#         # As long as there are no releases:
+#         url: https://codeberg.org/api/v1/repos/lucaweiss/lpa-gtk/tags/
+#         version-query: '.[0].name'
+    buildsystem: meson
+    build-commands:
```
hfiguiere, line 80:
```suggestion
    post-install:
```

eu.lucaweiss.lpa_gtk.yaml
```diff
@@ -0,0 +1,81 @@
...
+      - desktop-file-edit --set-icon=${FLATPAK_ID} /run/build/lpa-gtk/data/eu.lucaweiss.lpa_gtk.desktop
```
hfiguiere, line 81: you are not changing the one that has been installed.

eu.lucaweiss.lpa_gtk.yaml
```diff
@@ -0,0 +1,81 @@
...
+          url-template: https://github.com/estkme-group/lpac/archive/refs/tags/v$version.tar.gz
+  - name: lpac-logo # As long as lpa-gtk doesn't have its own
+    sources:
+      - type: file
+        url: https://avatars.githubusercontent.com/u/151148370
+        sha256: e03fa7c49b70397458c0759c78e50db7a755af27c618683b444d589b7000a4a3
+        dest-filename: eSTK_me.jpeg
+    buildsystem: simple
+    build-commands:
+      - for s in {32,64,128,256,512}; do ffmpeg -i eSTK_me.jpeg
```
hfiguiere, line 58: you need only one size.

eu.lucaweiss.lpa_gtk.yaml
Ferenc-: > no it doesn't.

Right. Now I see.

> trying to edit the icon property in post-install appears to be too late.

What was indeed too late is to still edit the desktop file under `/run/build/`. In the `post-install` the desktop file is better edited under `/app/share/applications/`.

eu.lucaweiss.lpa_gtk.yaml
Ferenc-: Done.

eu.lucaweiss.lpa_gtk.yaml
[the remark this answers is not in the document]
z3ntu: I haven't tried but can't you just use `tag: f04e129c9fde05ad8c78e4f412cf239776526d57`, or otherwise pull the .tar.gz from a specified commit?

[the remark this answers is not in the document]
Ferenc-: > can't you just use tag:

Previously, we had the equivalent with `commit:`.

> otherwise pull the .tar.gz from a specified commit?

Building from an archive has some advantages over git, so I have switched back to archive for now anyway. 
Regardless I am not sure if this would be enough.

eu.lucaweiss.lpa-gtk.yaml
z3ntu: Yeah, definitely don't take the eSTK.me logo and put it on my app, I don't know about copyright but it might also be trademarked

Ferenc-: Done

bbhtt:
bot, build eu.lucaweiss.lpa_gtk

z3ntu:
0.2 is out, so you can update to that if you want.

But please don't forget about this comment: https://github.com/flathub/flathub/pull/6203#issuecomment-2677939392

Currently in flatpak this app is **not functional** apart from opening the UI and getting an error screen.

Ferenc-:
> Also with https://codeberg.org/lucaweiss/lpa-gtk/issues/1 and https://codeberg.org/lucaweiss/lpa-gtk/issues/8 how can this handle access to /dev/tty* 

So I have now ensured access to `/dev/tty*` for the flatpak, and I have actually tested the flatpak with [your initial AT patch](https://codeberg.org/lucaweiss/lpa-gtk/commit/6a92d46309f54a60065f945a3615d3bc98dae1f5) with an EG-25 modem, and it worked to the point that the eSIM adding dialog appeared: 
image

So I think we have reached at least potential usefulness with the flatpak.
The open questions are:
1. Should we instruct the user to do the modem inhibit manually or does it really have to be done automatically by the application?
2. What the plan is with the https://codeberg.org/lucaweiss/lpa-gtk/commit/6a92d46309f54a60065f945a3615d3bc98dae1f5 patch? I think it is already pretty useful.

Ferenc-:
@bbhtt Considering the currently applied labels on this PR. What changes or events do you expect from your point of view?

bbhtt:
Why are you asking me when:

> Currently in flatpak this app is not functional apart from opening the UI and getting an error screen.

and you just posed some "open questions".

> The open questions are:

Ferenc-:
> and you just posted some "open questions".

My apologies for the confusion, the open questions were for @z3ntu, and IMHO not directly impact the readiness of the flaptak. I will move that discussion to the upstream forge to avoid further noise here.

> It needs to be ready from your end.

@bbhtt  If you are OK, with the `--device=all` and the provided reasoning (in comments). Then from my end this is ready for submission. The remaining details of patching the EG-25 support etc, as mentioned I will discuss on the upstream forge.

bbhtt:
bot, build

z3ntu:
I'm not happy if the flatpak would ship without QRTR support which right now is the only supported target.

For AT support, I guess without udev rules on the host to chmod/chown the /dev node it also wouldn't be able to r/w on the device? Or is there some path to intentional privilege escalation somewhere?

Ferenc-:
> I'm not happy if the flatpak would ship without QRTR support which right now is the only supported target.

Could you give specifics what kind of test would make you happy regarding `QRTR`? Would an E2E on an `MSM8916` device suffice?

z3ntu:
> @z3ntu That is not needed. As I said now with [aa3a65c](https://github.com/flathub/flathub/commit/aa3a65c41c9bcfc74d10ac8440601f7f014281a6) the full end-to-end AT support (with the patch) is tested and working. The `/dev` node is usable in the inside of the flatpak just as much as on the host. You can also try it out if you don't believe it.

But you do have some udev rules on the host that change the permissions of the node in /dev? By default they will be owned by root:root and have permissions like `0620` if I'm not mistaken.

Ferenc-:
> But you do have some udev rules on the host that change the permissions of the node in /dev? By default they will be owned by root:root and have permissions like `0620` if I'm not mistaken.

[Systemd by default grants a group access](https://github.com/systemd/systemd/blob/main/rules.d/50-udev-default.rules.in#L44) on every `tty[A-Z]*[0-9]` node including `ttyUSB2` (which is the `EG25.AT`)
This group is by default the `dialout` but in some cases I have seen the `uucp` group.
As long as the user is in the right group on the host, it will have access inside the flatpak.

PR #6234, opened 2025-03-01
Add net.krafting.SemantiK.Lang.French

Krafting:

This adds the French language pack for semantik. While the app shipped with this built-in, we found that making it an addon would allow user to uninstall it if they don't need it. Saving space for the user. 

The work to remove the baked-in language file from the main app has been done, we just need this to be published to create a new release.


Krafting:
bot, build net.krafting.SemantiK.Lang.French

bbhtt:
How large are these language packs?

Krafting:
> How large are these language packs?

Depend, but French is around 130MB and english was bigger at 300MB iirc

net.krafting.SemantiK.Lang.French.yml
```diff
@@ -0,0 +1,30 @@
+app-id: net.krafting.SemantiK.Lang.French
+build-extension: true
+separate-locales: false
+runtime: net.krafting.SemantiK
+runtime-version: stable
+sdk: org.freedesktop.Sdk//23.08
```
hfiguiere, line 6: you should upgrade the runtime of the host app first.

Also you could build it directly from the main manifest.

Krafting: I'm sorry but i'm not sure what you mean by this ?

> you should upgrade the runtime of the host app first.

You mean, so the main app supports addon ? this is already done, as I already have the English language pack published on flathub.

> Also you could build it directly from the main manifest.

So, you mean, When building the main app, addon gets built as well and also published automatically on flathub  ? Do you have any docs about this somewhere ?

DodoLeDev: For the building part, [there is doc about bundled extensions](https://docs.flatpak.org/en/latest/extension.html#bundled-extensions), but for the publishing part, the [docs](https://docs.flathub.org/docs/for-app-authors/submission) are not clear...

DodoLeDev: > I'm sorry but i'm not sure what you mean by this ?

@Krafting by upgrading the runtime, @hfiguiere means that your manifest does not request the latest version of `org.freedesktop.Sdk` (which is 24.08)

bbhtt: You need an exception for bundled exceptions, but it is probably not worth switching it again since the English lang pack already exists as a separate extension.

Krafting: > > I'm sorry but i'm not sure what you mean by this ?
> 
> @Krafting by upgrading the runtime, @hfiguiere means that your manifest does not request the latest version of `org.freedesktop.Sdk` (which is 24.08)

Oh okay, I bumped the runtime version

Krafting: > You need an exception for bundled exceptions, but it is probably not worth switching it again since the English lang pack already exists as a separate extension.

Well, if we add more language pack later it might make sense to use the main app manifest instead of creating many PR ? for the exception it's a per-extension exception or per-domain (for example net.krafting.SemantiK.Lang.*) ?

Krafting:
bot, build net.krafting.SemantiK.Lang.French

net.krafting.SemantiK.Lang.French.yml
bbhtt: Sure, it's one exception for everything.

A seperate extension points make sense if you want third party people to submit there extensions under  `net.krafting.SemantiK.Lang`, if not then bundled is fine.

Krafting: well, let's just keep them separate then, I'd love for people to be able to create their own language pack as well!

PR #6274, opened 2025-03-11
Add io.welle.welle-gui

AlbrechtL:

- [X] Please describe the application briefly. A DAB/DAB+ Software Defined Radio (SDR).

### DIFF
--- a/io.welle.welle-gui.yml
+++ b/io.welle.welle-gui.yml
@@ -1,0 +1,103 @@
+app-id: io.welle.welle-gui
+runtime: org.kde.Platform
+runtime-version: '6.8'
+sdk: org.kde.Sdk
+command: welle-io
+finish-args:
+  - '--share=network'
+  - '--share=ipc'
+  - '--filesystem=xdg-documents'
+  - '--socket=pulseaudio'
+  - '--socket=wayland'
+  - '--socket=fallback-x11'
+  - '--device=dri'
+  - '--device=all'
+
+modules:
+  - name: fftw
+    buildsystem: autotools
+    config-opts:
+      - '--enable-shared'
+      - '--disable-static'
+      - '--enable-threads'
+      - '--enable-float'
+    sources:
+      - type: archive
+        url: 'http://fftw.org/fftw-3.3.10.tar.gz'
+        md5: 8ccbf6a5ea78a16dbc3e1306e234cc5c
+
+  - name: libfaad
+    buildsystem: cmake
+    sources:
+      - type: git
+        url: 'https://salsa.debian.org/multimedia-team/faad2.git'
+        tag: debian/2.11.2-1
+
+  - name: libusb
+    buildsystem: autotools
+    config-opts:
+      - '--disable-udev'
+    sources:
+      - type: git
+        url: 'https://github.com/libusb/libusb.git'
+        tag: v1.0.27
+
+  - name: librtlsdr
+    buildsystem: cmake
+    config-opts:
+      - '-Wno-dev'
+      - '-DDETACH_KERNEL_DRIVER=ON'
+    sources:
+      - type: git
+        url: 'https://gitea.osmocom.org/sdr/rtl-sdr.git'
+        tag: v2.0.2
+
+  - name: libairspy
+    buildsystem: cmake
+    config-opts:
+      - '-Wno-dev'
+    sources:
+      - type: git
+        url: 'https://github.com/airspy/airspyone_host.git'
+        tag: v1.0.10
+
+  - name: SoapySDR
+    buildsystem: cmake
+    config-opts:
+      - '-DCMAKE_BUILD_TYPE=Release'
+      - '-Wno-dev'
+    sources:
+      - type: git
+        url: 'https://github.com/pothosware/SoapySDR.git'
+        tag: soapy-sdr-0.8.1
+
+  - name: welle.io
+    buildsystem: qmake
+    build-commands:
+      # Copy welle.io binary
+      - cp src/welle-gui/welle-io /app/bin
+
+      # Copy icons
+      - install -Dm644 src/welle-gui/icons/16x16/welle-io.png /app/share/icons/hicolor/16x16/apps/io.welle.welle-gui.png
+      - install -Dm644 src/welle-gui/icons/32x32/welle-io.png /app/share/icons/hicolor/32x32/apps/io.welle.welle-gui.png
+      - install -Dm644 src/welle-gui/icons/48x48/welle-io.png /app/share/icons/hicolor/48x48/apps/io.welle.welle-gui.png
+      - install -Dm644 src/welle-gui/icons/128x128/welle-io.png /app/share/icons/hicolor/128x128/apps/io.welle.welle-gui.png
+      - install -Dm644 src/welle-gui/icons/256x256/welle-io.png /app/share/icons/hicolor/256x256/apps/io.welle.welle-gui.png
+      - install -Dm644 io.welle.welle-gui.desktop -t /app/share/applications
+
+      # Add current build date into appdata.xml
+      - sed -i 's/date="1970-01-01"/'date=\""$(date +%Y-%m-%d)"\"'/g' io.welle.welle-gui.appdata.xml
+      # Add app version and git hash into appdata.xml
+      - sed -i 's/version="0.0.0"/'version=\""$(cat src/welle-gui/_current_version)-$(git rev-parse --short HEAD)"\"'/g' io.welle.welle-gui.appdata.xml
+      - cat io.welle.welle-gui.appdata.xml
+      # Finally copy the appdata.xml
+      - install -Dm644 io.welle.welle-gui.appdata.xml -t /app/share/metainfo
+
+    config-opts:
+      - LIBS+=-L/app/lib
+      - LIBS+=-L/app/lib64
+      - 'QMAKE_LFLAGS+=-Wl,-rpath,/app/lib64'
+    sources:
+      - type: git
+        url: https://github.com/AlbrechtL/welle.io.git
+        commit: 1a4f14960071eb1bfe4bb757c522123ce5019470
--- a/README.md
+++ b/README.md
@@ -1,0 +1,1 @@
+Please see the upstream repository https://github.com/AlbrechtL/welle.io.


io.welle.welle-gui.yml
```diff
@@ -0,0 +1,103 @@
...
+command: welle-io
+finish-args:
+  - '--share=network'
+  - '--share=ipc'
+  - '--filesystem=xdg-documents'
+  - '--socket=pulseaudio'
+  - '--socket=wayland'
+  - '--socket=fallback-x11'
+  - '--device=dri'
+  - '--device=all'
```
hfiguiere, line 14: `all` includes `dri`, but why do you need `all`?

io.welle.welle-gui.yml
```diff
@@ -0,0 +1,103 @@
...
+
+modules:
+  - name: fftw
+    buildsystem: autotools
+    config-opts:
+      - '--enable-shared'
+      - '--disable-static'
+      - '--enable-threads'
+      - '--enable-float'
```
hfiguiere, line 23: please use the shared-module

https://github.com/flathub/shared-modules/blob/master/linux-audio/fftw3f.json

io.welle.welle-gui.yml
```diff
@@ -0,0 +1,103 @@
...
+        md5: 8ccbf6a5ea78a16dbc3e1306e234cc5c
+
+  - name: libfaad
+    buildsystem: cmake
+    sources:
+      - type: git
+        url: 'https://salsa.debian.org/multimedia-team/faad2.git'
+        tag: debian/2.11.2-1
+
+  - name: libusb
```
hfiguiere, line 36: please use the shared-modules too

io.welle.welle-gui.yml
```diff
@@ -0,0 +1,103 @@
...
+      - '-DCMAKE_BUILD_TYPE=Release'
+      - '-Wno-dev'
+    sources:
+      - type: git
+        url: 'https://github.com/pothosware/SoapySDR.git'
+        tag: soapy-sdr-0.8.1
+
+  - name: welle.io
+    buildsystem: qmake
+    build-commands:
```
hfiguiere, line 76: build-commands is only for when `buildsystem` is `simple`

io.welle.welle-gui.yml
```diff
@@ -0,0 +1,103 @@
...
+      # Copy welle.io binary
+      - cp src/welle-gui/welle-io /app/bin
```
hfiguiere, line 78:
```suggestion
      - install -Dm755 src/welle-gui/welle-io -t /app/bin
```

io.welle.welle-gui.yml
```diff
@@ -0,0 +1,103 @@
...
+      - install -Dm644 src/welle-gui/icons/48x48/welle-io.png /app/share/icons/hicolor/48x48/apps/io.welle.welle-gui.png
+      - install -Dm644 src/welle-gui/icons/128x128/welle-io.png /app/share/icons/hicolor/128x128/apps/io.welle.welle-gui.png
+      - install -Dm644 src/welle-gui/icons/256x256/welle-io.png /app/share/icons/hicolor/256x256/apps/io.welle.welle-gui.png
+      - install -Dm644 io.welle.welle-gui.desktop -t /app/share/applications
+
+      # Add current build date into appdata.xml
+      - sed -i 's/date="1970-01-01"/'date=\""$(date +%Y-%m-%d)"\"'/g' io.welle.welle-gui.appdata.xml
+      # Add app version and git hash into appdata.xml
+      - sed -i 's/version="0.0.0"/'version=\""$(cat src/welle-gui/_current_version)-$(git rev-parse --short HEAD)"\"'/g' io.welle.welle-gui.appdata.xml
+      - cat io.welle.welle-gui.appdata.xml
```
hfiguiere, line 92:
```suggestion
```

io.welle.welle-gui.yml
```diff
@@ -0,0 +1,103 @@
...
+      - install -Dm644 src/welle-gui/icons/32x32/welle-io.png /app/share/icons/hicolor/32x32/apps/io.welle.welle-gui.png
+      - install -Dm644 src/welle-gui/icons/48x48/welle-io.png /app/share/icons/hicolor/48x48/apps/io.welle.welle-gui.png
+      - install -Dm644 src/welle-gui/icons/128x128/welle-io.png /app/share/icons/hicolor/128x128/apps/io.welle.welle-gui.png
+      - install -Dm644 src/welle-gui/icons/256x256/welle-io.png /app/share/icons/hicolor/256x256/apps/io.welle.welle-gui.png
+      - install -Dm644 io.welle.welle-gui.desktop -t /app/share/applications
+
+      # Add current build date into appdata.xml
+      - sed -i 's/date="1970-01-01"/'date=\""$(date +%Y-%m-%d)"\"'/g' io.welle.welle-gui.appdata.xml
+      # Add app version and git hash into appdata.xml
+      - sed -i 's/version="0.0.0"/'version=\""$(cat src/welle-gui/_current_version)-$(git rev-parse --short HEAD)"\"'/g' io.welle.welle-gui.appdata.xml
```
hfiguiere, line 91: this should really be handled upstream as they are the one doing releases.

io.welle.welle-gui.yml
```diff
@@ -0,0 +1,103 @@
...
+      - '--disable-static'
+      - '--enable-threads'
+      - '--enable-float'
+    sources:
+      - type: archive
+        url: 'http://fftw.org/fftw-3.3.10.tar.gz'
+        md5: 8ccbf6a5ea78a16dbc3e1306e234cc5c
+
+  - name: libfaad
+    buildsystem: cmake
```
hfiguiere, line 30:
```suggestion
    buildsystem: cmake-ninja
```

io.welle.welle-gui.yml
```diff
@@ -0,0 +1,103 @@
...
+    config-opts:
+      - '-Wno-dev'
+      - '-DDETACH_KERNEL_DRIVER=ON'
+    sources:
+      - type: git
+        url: 'https://gitea.osmocom.org/sdr/rtl-sdr.git'
+        tag: v2.0.2
+
+  - name: libairspy
+    buildsystem: cmake
```
hfiguiere, line 56:
```suggestion
    buildsystem: cmake-ninja
```

io.welle.welle-gui.yml
```diff
@@ -0,0 +1,103 @@
...
+    config-opts:
+      - '-Wno-dev'
+    sources:
+      - type: git
+        url: 'https://github.com/airspy/airspyone_host.git'
+        tag: v1.0.10
+
+  - name: SoapySDR
+    buildsystem: cmake
```
hfiguiere, line 65:
```suggestion
    buildsystem: cmake-ninja
```

io.welle.welle-gui.yml
hfiguiere: also `sed` is terrible as it doesn't have any failure mode. use a patch instead.

io.welle.welle-gui.yml
```diff
@@ -0,0 +1,103 @@
...
+      # Copy welle.io binary
+      - cp src/welle-gui/welle-io /app/bin
+
+      # Copy icons
+      - install -Dm644 src/welle-gui/icons/16x16/welle-io.png /app/share/icons/hicolor/16x16/apps/io.welle.welle-gui.png
+      - install -Dm644 src/welle-gui/icons/32x32/welle-io.png /app/share/icons/hicolor/32x32/apps/io.welle.welle-gui.png
+      - install -Dm644 src/welle-gui/icons/48x48/welle-io.png /app/share/icons/hicolor/48x48/apps/io.welle.welle-gui.png
+      - install -Dm644 src/welle-gui/icons/128x128/welle-io.png /app/share/icons/hicolor/128x128/apps/io.welle.welle-gui.png
+      - install -Dm644 src/welle-gui/icons/256x256/welle-io.png /app/share/icons/hicolor/256x256/apps/io.welle.welle-gui.png
+      - install -Dm644 io.welle.welle-gui.desktop -t /app/share/applications
```
hfiguiere, line 86: these should also be installed with the upstream build system. Even qmake knows how to do it.

io.welle.welle-gui.yml
AlbrechtL: Do you have a succession how to put the version number and git hash into the manifest without doing a git commit? I would like to decrease the number of files that I have to change manually when doing a release.

Is your succession placing patches in this repo here?

io.welle.welle-gui.yml
AlbrechtL: welle.io needs access via `libusb` to USB dongles. During my tests, I had issues accessing USB. With `all` it was working. But maybe I miss something.

io.welle.welle-gui.yml
AlbrechtL: I don't find faad2 inside the shared-module repo.

io.welle.welle-gui.yml
AlbrechtL: Done. Thanks for the hint.

io.welle.welle-gui.yml
bbhtt: it's for libusb https://github.com/flathub/shared-modules/blob/master/libusb/libusb.json

io.welle.welle-gui.yml
```diff
@@ -0,0 +1,92 @@
...
+      - sed -i 's/date="1970-01-01"/'date=\""$(date +%Y-%m-%d)"\"'/g' io.welle.welle-gui.appdata.xml
+      # Add app version and git hash into appdata.xml
+      - sed -i 's/version="0.0.0"/'version=\""$(cat src/welle-gui/_current_version)-$(git rev-parse --short HEAD)"\"'/g' io.welle.welle-gui.appdata.xml
+      # Finally copy the appdata.xml
+      - install -Dm644 io.welle.welle-gui.appdata.xml -t /app/share/metainfo
+
+    config-opts:
+      - LIBS+=-L/app/lib
+      - LIBS+=-L/app/lib64
+      - 'QMAKE_LFLAGS+=-Wl,-rpath,/app/lib64'
```
bbhtt, line 88: /app/lib64 should not exist, use /app/lib

io.welle.welle-gui.yml
AlbrechtL: I did a test, with `--device=usb`. `flatpak` gives me the error `error: Unknown device type usb, valid types are: dri, all, kvm, shm`

I'm using Ubuntu 24.04 for development
```
$ flatpak --version
Flatpak 1.14.6
```

I think I have to stay with `--device=all`

Removing `--device=dri` done in https://github.com/AlbrechtL/flathub/commit/53e6d65eae558e47e1c468b8ed4b67c18881efc2.

io.welle.welle-gui.yml
AlbrechtL: Funny, when I was developing the flatpak a few months ago, I needed `/app/lib64`. Anyway, it seems that it is also working without it. Thanks for the hint.

Fixed in https://github.com/AlbrechtL/flathub/commit/c2fd82bc92b1a65e45e151641dc38a503e233e33.

io.welle.welle-gui.yml
AlbrechtL: Fixed in https://github.com/AlbrechtL/flathub/commit/399692c6c6273b50da796fa854a789f567555523.

io.welle.welle-gui.yml
AlbrechtL: Is this is a hard blocker? I mean, when I add this into the `qmake` .pro file, the files will be installed to `/app/..` for all users when they are doing `make install` in a local build. No matter if they are interested in flatpaks or not. Furthermore, this will mess up the users file system, or I'm wrong?

Do you have a reference Qt flatpak?

io.welle.welle-gui.yml
AlbrechtL: Understood. But I need this, currently (see https://github.com/flathub/flathub/pull/6274#discussion_r1991000223).

bbhtt: it won't work together with buildsystem: qmake, if you need it drop it and run qmake manually with the proper arguments.

hfiguiere: a quick search https://forum.qt.io/topic/8774/define-the-installation-prefix-using-qmake

io.welle.welle-gui.yml
hfiguiere: I wonder sometime if github show different things to different people:

image

this looked obvious to me.

io.welle.welle-gui.yml
hfiguiere: `org.kde.*` but they use cmake instead.

(my opinion of qmake is even lower than cmake)

AlbrechtL: Of course, I already checked the KDE apps. But it is a little bit hard to me to transfer the way how KDE apps are doing it to welle.io. Anyway, I will try to find a way to remove the extra install steps.

io.welle.welle-gui.yml
AlbrechtL: No github is showing the same to me
grafik

But I saw `faad2` first because it is the first highlighted line. But I already changed to the `libusb` shared module.

io.welle.welle-gui.yml
AlbrechtL: Ok, I will try to use the PREFIX.

io.welle.welle-gui.yml
AlbrechtL: Fixed in https://github.com/AlbrechtL/flathub/commit/fb66dbf0986effa9ed6f7e80d1cd6bede07ab510.

io.welle.welle-gui.yml
AlbrechtL: Fixed in https://github.com/AlbrechtL/flathub/commit/fb66dbf0986effa9ed6f7e80d1cd6bede07ab510

AlbrechtL:
@hfiguiere, @bbhtt Thanks for your review! I hope I did everything what you requested and suggested. Should I click to "Resolve conversation" for each comment or do you do it?

bbhtt:
bot, build io.welle.welle-gui

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "shared-modules"]
+	path = shared-modules
+	url = https://github.com/flathub/shared-modules.git
```
bbhtt, line 3: shared-modules isn't added properly

AlbrechtL: Sorry. It is my first time working with submodules. `git` created this file when I added the submodule. Unfortunately, I don't see what is wrong. I also checked https://github.com/flathub/org.octave.Octave/blob/master/.gitmodules but it looks identical. Do you know what is missing?

AlbrechtL:
> You likely need some cleanup, there's no cleanup apart from shared modules. https://docs.flatpak.org/en/latest/manifests.html#cleanup

Thanks for pointing out! Before I will add the cleanup, I would like to have a working flatpak. Do you agree?

bbhtt:
Yes that's fine

.gitmodules
bbhtt: You have to add and commit the shared-modules folder.

AlbrechtL: Sorry, that was my fault. I added the submodules folder to the `.gitignore` file :see_no_evil:. Fixed in https://github.com/AlbrechtL/flathub/commit/114d64d7ea2373d7322226c48fc042fe3ad638a1.

bbhtt:
bot, build io.welle.welle-gui

bbhtt:
Looks like it has a tray so you need `--talk-name=org.kde.StatusNotifierWatcher` but the tray icon is not showing up. You need to pass the appid as icon name there

AlbrechtL:
Thanks for reporting. Fixed in https://github.com/flathub/flathub/pull/6274/commits/867dd5d03133f8256b96b14a9de46e05f32e3420.

But I only need to add the line https://github.com/flathub/flathub/blob/867dd5d03133f8256b96b14a9de46e05f32e3420/io.welle.welle-gui.yml#L13

Now it is working!
grafik

AlbrechtL:
bot, build io.welle.welle-gui

bbhtt:
the icon doesn't work with that only on GNOME (with the tray extension). You should look into it.

PR #6284, opened 2025-03-15
Initial commit

jflatt:

MythTV, this is a DVR, and multimedia player with plugins.  Base for a home theater PC
https://mythtv.org
org.mythtv.MythTV


org.mythtv.MythTV.yaml
```diff
@@ -0,0 +1,1063 @@
+id: org.mythtv.MythTV
+runtime: org.kde.Platform
+runtime-version: '6.8'
+sdk: org.kde.Sdk
+command: mythfrontend
+finish-args:
+  - --device=all
+  - --persist=.mythtv
+  - --filesystem=home/.mythtv
```
bbhtt, line 9: why do you need both?

org.mythtv.MythTV.yaml
```diff
@@ -0,0 +1,1063 @@
...
+    build-commands:
+      - perl Makefile.PL
+      - make && make install
+      - find /app/lib/perl5 -type f -exec chmod u+w {} \;
+
+  - name: perl-HTTP-Message
+    sources:
+      - type: archive
+        url: https://cpan.metacpan.org/authors/id/O/OA/OALDERS/HTTP-Message-7.00.tar.gz
+        sha256: 5afa95eb6ed1c632e81656201a2738e2c1bc6cbfae2f6d82728e2bb0b519c1dc
```
bbhtt, line 633: but the perl sources in a seperate yaml and use the same build command for it. See https://github.com/flathub/com.gitlab.davem.ClamTk/blob/cf30cf7d7d0ec83e7e7fb88518caca135a857188/com.gitlab.davem.ClamTk.yaml#L62-L72 for inspiration

org.mythtv.MythTV.yaml
```diff
@@ -0,0 +1,1063 @@
...
+#        commit: 6a7c0bd85c0f4ac2861f96547ab11d0d469f38ca
+
+#  - name: dveo-asi
+#    sources:
+#      - type: git
+#        url: https://github.com/kierank/dveo-linux-master.git
+#        commit: 5d62458bb6e13822f59f85eb260e31bb2c7df3a2
+
+#MythBrowser requires QtWebkit.
+#MythNetvision requires the MythTV Python bindings (MythTV)
```
bbhtt, line 1021: drop these comments.

org.mythtv.MythTV.yaml
```diff
@@ -0,0 +1,1063 @@
...
+      - type: git
+        url: https://github.com/MythTV/mythtv.git
+        branch: fixes/35
+        commit: 30ba688773852d3f8ad44dd7ceb1b6634ca7ea4e
+      - type: file
+        path: mythtv128x128.png
+      - type: file
+        path: org.mythtv.MythTV.desktop
+      - type: file
+        path: org.mythtv.MythTV.metainfo.xml
```
bbhtt, line 1046: these should be in upstream repo not here.

org.mythtv.MythTV.yaml
```diff
@@ -0,0 +1,1063 @@
...
+      - type: patch
+        path: libva.patch
+    buildsystem: simple
+    build-options:
+      prepend-pkg-config-path: /app/lib64/pkgconfig
+      ldflags: -L/app/lib64
```
bbhtt, line 1052: lib64 is not supposed to exist it should be /app/lib, you were using an old flatpak-builder version.

org.mythtv.MythTV.yaml
```diff
@@ -0,0 +1,1063 @@
...
+      env:
+        PERL5LIB: /app/lib/perl5
+    build-commands:
+# --enable-bdjava --enable-libxvid
+      - cd mythtv && ./configure --prefix=/app --compile-type=release --enable-vulkan --enable-libmp3lame --enable-libx264 --enable-libx265 --enable-libvpx
```
bbhtt, line 1057: why not use autotools?

org.mythtv.MythTV.metainfo.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>org.mythtv.MythTV</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>GPL-2.0-or-later</project_license>
+  <name>MythTV</name>
+  <summary>
+    MythTV is a Free Open Source software digital video recorder (DVR) project distributed under the terms of the GNU GPL
+  </summary>
```
bbhtt, line 9: reduce the summary, drop the license and name no need to repeat. also summary should not have break `<summary>foobar</summary>`

org.mythtv.MythTV.metainfo.xml
```diff
@@ -0,0 +1,35 @@
...
+  </categories>
+  <url type="homepage">https://www.mythtv.org</url>
+  <launchable type="desktop-id">org.mythtv.MythTV.desktop</launchable>
+  <releases>
+    <release version="35.0" date="2025-02-24"/>
+  </releases>
+  <provides>
+    <binary>mythfrontend</binary>
+  </provides>
+  <content_rating type="oars-1.0" />
```
bbhtt, line 29:
```suggestion
  <content_rating type="oars-1.1" />
```

org.mythtv.MythTV.metainfo.xml
```diff
@@ -0,0 +1,35 @@
...
+  <screenshots>
+    <screenshot type="default">
+        <image type="source" width="170" height="130">https://www.mythtv.org/cache/tv_img_screenshots_mythtv_retro_theme_jpg_tv_black_l_a_e6e6f0.jpg</image>
```
bbhtt, line 32: this screenshot won't do, it's like an icon. Take a screenshot properly.

org.mythtv.MythTV.yaml
jflatt: > why do you need both?

Otherwise it can't initially create the directory on its own

org.mythtv.MythTV.yaml
jflatt: > but the perl sources in a seperate yaml and use the same build command for it. See https://github.com/flathub/com.gitlab.davem.ClamTk/blob/cf30cf7d7d0ec83e7e7fb88518caca135a857188/com.gitlab.davem.ClamTk.yaml#L62-L72 for inspiration

Not sure where that packages install.sh is coming from.  Some of these perl packages have different build processes, some use Makefile.PL, some call it Build.PL.  I wouldn't be able to cleanly iterate through them

org.mythtv.MythTV.yaml
bbhtt: this currently doesn't make sense it's either persist or the other one.

org.mythtv.MythTV.yaml
jflatt: Doesn't have one.  Also it's actually two projects, one depends on the other

org.mythtv.MythTV.yaml
bbhtt: Right now this seems unmaintainable

org.mythtv.MythTV.yaml
hfiguiere: > > why do you need both?
> 
> Otherwise it can't initially create the directory on its own

I doubt. These lead to two different directories. You should use `--persist`.

org.mythtv.MythTV.yaml
hfiguiere: >  I wouldn't be able to cleanly iterate through them

Didn't you use flatpak-builder-tools ? https://github.com/flatpak/flatpak-builder-tools/tree/master/cpan

If yes then these auto generated manifest are kept in a seperate files.
If no, then you should.

PR #6290, opened 2025-03-18
Add io.github.swordpuffin.wardrobe

SwordPuffin:

A graphical utility for easily installing and equipping community made Gnome Shell, Gtk3/4, icon, and cursor themes as well as wallpapers.

- [X] I am an author/developer/upstream contributor to the project. If not, I contacted upstream developers about this submission. [(repo)]( https://github.com/SwordPuffin/Wardrobe)

### DIFF
--- a/io.github.swordpuffin.wardrobe.json
+++ b/io.github.swordpuffin.wardrobe.json
@@ -1,0 +1,42 @@
+{
+    "id" : "io.github.swordpuffin.wardrobe",
+    "runtime" : "org.gnome.Platform",
+    "runtime-version" : "47",
+    "sdk" : "org.gnome.Sdk",
+    "command" : "wardrobe",
+    "finish-args" : [
+        "--share=network",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--device=dri",
+        "--socket=wayland",
+        "--talk-name=org.freedesktop.Flatpak",
+        "--filesystem=host:ro",
+        "--filesystem=home"
+    ],
+    "cleanup" : [
+        "/include",
+        "/lib/pkgconfig",
+        "/man",
+        "/share/doc",
+        "/share/gtk-doc",
+        "/share/man",
+        "/share/pkgconfig",
+        "*.la",
+        "*.a"
+    ],
+    "modules" : [
+        {
+            "name" : "wardrobe",
+            "builddir" : true,
+            "buildsystem" : "meson",
+            "sources" : [
+                {
+                    "type" : "git",
+                    "url" : "https://github.com/SwordPuffin/Wardrobe.git",
+                    "tag" : "v1.0.0"
+                }
+            ]
+        }
+    ]
+}
--- a/io.github.swordpuffin.wardrobe.json
+++ b/io.github.swordpuffin.wardrobe.json
@@ -5,14 +5,12 @@
     "sdk" : "org.gnome.Sdk",
     "command" : "wardrobe",
     "finish-args" : [
-        "--share=network",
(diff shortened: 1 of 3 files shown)


io.github.swordpuffin.wardrobe.json
```diff
@@ -0,0 +1,42 @@
...
+    "runtime-version" : "47",
+    "sdk" : "org.gnome.Sdk",
+    "command" : "wardrobe",
+    "finish-args" : [
+        "--share=network",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--device=dri",
+        "--socket=wayland",
+        "--talk-name=org.freedesktop.Flatpak",
```
bbhtt, line 13: there is no point in flatpaking something if every command requires spawning flatpak-spawn https://github.com/SwordPuffin/Wardrobe/blob/845c2667715850a856078e9fd139b68c83d354f0/src/window.py#L59-L63

bbhtt: also some of them are exposed inside the sandbox through settings portal

bbhtt: there's a portal to set the wallpaper too https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Wallpaper.html

SwordPuffin:
For the time being, I am just going to comment out the feature that required the commands, because I really can't figure out how to expose the org.gnome.desktop.interface and org.gnome.extensions.user-theme schemas to the sandbox.

The problem should be resolved.

bbhtt:
bot, build io.github.swordpuffin.wardrobe

bbhtt:
Where are the themes being downloaded? Also moving from shell themes to icon themes etc. feels a bit laggy.

bbhtt:
Looks like going to `~/.themes`? Don't use those paths, it's legacy, the correct location is `$XDG_DATA_HOME/themes` ie. `~/.local/share/themes` usually.

SwordPuffin:
The images are fairly slow to load, I don't know if that has to do with the ocs api or soup.

I'll change the paths to the more appropriate folders

SwordPuffin:
It's done now,

I used FLATPAK_USER_DIR instead because XDG_DATA_HOME was different in the sandbox. Also, I changed the path for wallpapers because I think it would not work as intended for non-english users.

bbhtt:
FLATPAK_USER_DIR seems wrong you want to use HOST_XDG_DATA_HOME https://docs.flatpak.org/en/latest/conventions.html#xdg-base-directories

hfiguiere:
Not even sure where you found that env variable, it surely isn't set in flatpak.

SwordPuffin:
I've made the changes, it uses HOST_XDG_DATA_HOME for the path.

I think I pulled FLATPAK_USER_DIR from Gnome Discourse or some random Flatpak issue from a few years ago. Looking back I don't know why I tried to used it.

io.github.swordpuffin.wardrobe.json
```diff
@@ -0,0 +1,40 @@
...
+    "modules" : [
+        {
+            "name" : "wardrobe",
+            "builddir" : true,
+            "buildsystem" : "meson",
+            "sources" : [
+                {
+                    "type" : "git",
+                    "url" : "https://github.com/SwordPuffin/Wardrobe.git",
+                    "tag" : "v1.0.1"
```
bbhtt, line 35: Don't retag, always create a new tag.

bbhtt:
bot, build io.github.swordpuffin.wardrobe

PR #6302, opened 2025-03-21
Add org.archivekeep.ArchiveKeep

kravemir:

- [x] Please describe the application briefly:

  Utility to facilitate archivation of personal files in multiple storages, with non-destructive incremental replication (synchronization).
  
  - proof type: [webpage](https://archivekeep.org/) links to [github repository](https://github.com/archivekeep/archivekeep), which contains sources for this application


org.archivekeep.ArchiveKeep.yaml
```diff
@@ -0,0 +1,66 @@
...
+modules:
+  - name: openjdk
+    buildsystem: simple
+    build-commands:
+      - /usr/lib/sdk/openjdk17/install.sh
+
+  - name: archivekeep-desktop
+    buildsystem: simple
+    only-arches:
+      - x86_64
```
hfiguiere, line 35: so what happen when building the aarch64 package?

kravemir: I have not tried, as I don't have ARM machine. However, looking at [Compatibility and versions﻿
 of Compose Multiplatform](https://www.jetbrains.com/help/kotlin-multiplatform-dev/compose-compatibility-and-versioning.html), it should be supported.

I thought, that I'd first verify, that it works fine, before claiming the support for the target. I don't have ARM machine with Linux, though.

org.archivekeep.ArchiveKeep.yaml
```diff
@@ -0,0 +1,66 @@
...
+    build-options:
+      append-path: "/usr/lib/sdk/openjdk17/bin"
+    build-commands:
+      - "sed -i s/distributionUrl.*/distributionUrl=gradle-bin.zip/ gradle/wrapper/gradle-wrapper.properties"
+      - ./gradlew --no-daemon --offline --console=plain createReleaseDistributable
+      - find
+      - mkdir /app/bin/
+      - mkdir /app/lib/
```
bbhtt, line 43:
```suggestion
      - mkdir -p /app/bin /app/lib
```

org.archivekeep.ArchiveKeep.yaml
```diff
@@ -0,0 +1,66 @@
...
+      - cp -r modules/app-desktop/build/compose/binaries/main-release/app/archivekeep-desktop/bin/* /app/bin/
+      - cp -r modules/app-desktop/build/compose/binaries/main-release/app/archivekeep-desktop/lib/* /app/lib/
+      - install -Dm644 modules/app-desktop/src/main/composeResources/drawable/ic_app.svg /app/share/icons/hicolor/scalable/apps/$FLATPAK_ID.svg
+      - install -Dm644 distribution/resources/$FLATPAK_ID.desktop /app/share/applications/$FLATPAK_ID.desktop
+      - install -Dm644 $FLATPAK_ID.appdata.xml /app/share/appdata/$FLATPAK_ID.appdata.xml
+      - find /app
```
bbhtt, line 49:
```suggestion

```

org.archivekeep.ArchiveKeep.yaml
```diff
@@ -0,0 +1,66 @@
...
+        sha256: "f397b287023acdba1e9f6fc5ea72d22dd63669d59ed4a289a29b1a76eee151c6"
+        dest: "gradle/wrapper"
+        dest-filename: "gradle-bin.zip"
+      - "gradle-sources-root.json"
+      - "gradle-sources-app-core.json"
+      - "gradle-sources-app-desktop.json"
+      - "gradle-sources-cli.json"
+      - "gradle-sources-files.json"
+      - type: file
+        path: org.archivekeep.ArchiveKeep.appdata.xml
```
bbhtt, line 66: this should be upstreamed to https://github.com/archivekeep/archivekeep

org.archivekeep.ArchiveKeep.appdata.xml
```diff
@@ -0,0 +1,48 @@
...
+    </provides>
+
+    <metadata_license>FSFAP</metadata_license>
+    <project_license>AGPL-3.0-or-later</project_license>
+
+    <developer id="org.kravemir">
+        <name>Miroslav Kravec</name>
+    </developer>
+
+    <content_rating type="oars-1.0" />
```
bbhtt, line 42:
```suggestion
    <content_rating type="oars-1.1" />
```

org.archivekeep.ArchiveKeep.yaml
kravemir: :heavy_check_mark: - cleaned up

org.archivekeep.ArchiveKeep.yaml
kravemir: :heavy_check_mark: - cleaned up

org.archivekeep.ArchiveKeep.yaml
kravemir: :heavy_check_mark: - upstreamed

org.archivekeep.ArchiveKeep.appdata.xml
kravemir: :heavy_check_mark: - applied to upstreamed appstream metainfo

org.archivekeep.ArchiveKeep.yaml
kravemir: @hfiguiere is it fine, if aarch64 is left out for now - until I'll get some ARM64 device with Linux? Or, do you expect some action from me?

kravemir:
@bbhtt thank you for reviewing. I've applied all the findings you shared in review comments.

And, I've also disabled fetching of git submodules. There's only a hugo theme, which is used only to build webdocs, but not flatpak application.

Could you please take another look?

org.archivekeep.ArchiveKeep.yaml
hfiguiere: right no an aarch64 package is still build. But what's the point to use a VM based language if it is to restrict architectures?

kravemir: > right no an aarch64 package is still build. 

I can't parse this. Can you please rephrase?

> But what's the point to use a VM based language if it is to restrict architectures?

The point of JVM/Kotlin is Jetpack Compose / Compose Multiplatform, which supports also Android. The Android app is planned as one of the next things. And, I'd like to try ARM64 Linux, too, eventually. I found [pine64 laptops, phones and tablets](https://pine64.org/devices/), and these look interesting, but I'm in between jobs currently, and therefore I'm being rather considerate with money spending right now. It's not that the plan is to restrict architecture, but that I'm not there yet with the implementation (testing).

Should I just not restrict it, and see what happens? It'd be totally blind delivery from my side. Or, maybe I could try QEMU, if that is a blocker.

hfiguiere: I mean "right now, an aarch64 package will still be built" and it will be broken.

Also not building the other arch mean it can't be tested.

kravemir: I thought this makes it not being built for aarch64. Looks like that's not the case, and therefore there's no reason to keep it there. I've removed the `only-arches` option.

If I understand it correctly now, then it's a module level control, and there can be one module for one arch, and another module for the other arch.

What defines, for which architectures will package be built for?

hfiguiere: https://docs.flathub.org/docs/for-app-authors/maintenance#flathubjson

hfiguiere: that will still be disappointing to limit the architecture for no reason.

kravemir: I see. You're probably right about not wanting to artificially limit the architecture.

I was just being cautious, but maybe that's not needed. And, maybe it would be better to just detect issue and fix, if there's any.

Do you happen to know a good tutorial for testing out aarch64 on x86_64 (AMD64), or do you have a good experience with some specific distro to do this?

kravemir: It fails due to missing dependencies, that require platform specific variants. The source lists for gradle builds were generated under x86-64.

But,.. the work in emulated aarch64 is quite slow. So, it will take mae some time to prepare a fix for this.

bbhtt:
bot, build org.archivekeep.ArchiveKeep

you can start the build

kravemir:
> Build [188601 failed](https://buildbot.flathub.org/#/builders/6/builds/188601)

I'm working on this - https://github.com/flathub/flathub/pull/6302#discussion_r2008877546.

org.archivekeep.ArchiveKeep.yaml
kravemir: The build works under aarch64, now.

kravemir:
I've fixed the `aarch64` build:

- generated source lists for gradle dependencies under `aarch64`:
  - nice and clean reproducibility of this manual step depends on https://github.com/jwharm/flatpak-gradle-generator/pull/7,
  - currently, the generation fails out of the box, because there's a configuration that won't resolve successfully, but actually isn't needed for the build itself. So, I generated the list by publishing plugin to maven local, adding an option to ignore the sourceset, and generated the list (and committed).
  - I'll incorporate the update in the upstream for the next release, once the PR for the plugin is merged.
- built using `flatpak run org.flatpak.Builder ...`,
- launched using `flaptk run org.archivekeep.ArchiveKeep`.

@bbhtt can you please take a look after this fix?

And, sorry I didn't edit/update my ask for re-review to mark it as invalid, after I found out, that it won't build for ARM64.

org.archivekeep.ArchiveKeep.yaml
kravemir: > so what happen when building the aarch64 package? ... that will still be disappointing to limit the architecture for no reason.

I remember now why I omitted it. Because I saw, that Jetpack Compose has platform specific dependencies, and the source lists were generated under `x86_64` platform. But, I forgot that.

It's fixed now, and build works for both architectures.

bbhtt:
bot, build org.archivekeep.ArchiveKeep

bbhtt:
How does this work? I'm trying to select a directory and it seems to be doing nothing?


https://github.com/user-attachments/assets/80cdbd31-5fa9-4d2b-9967-d9d55f0bab84

kravemir:
> How does this work? I'm trying to select a directory and it seems to be doing nothing?

:eyes: that's a bug, and in the very first step / experience :flushed: 

I don't know why, but it stopped working for Flatpak build on my machine also. I've rebuilt it locally with added print to console, and see that FileKit (`rememberDirectoryPickerLauncher`) now sends `null` for `directory` callback for some unknown reason. Will keep investigating.

kravemir:
@bbhtt very thank you for the catch.

I should have had tested the beginning / setup use-case myself (I was working mostly with existing setup archives / repositories).

The issue, that `org.freedesktop.portal.FileChooser` was not being opened, is fixed now. And, also some other issues that I found with further testing. I've released the new version. And this PR is ready for re-review.

Can you please review it again?

bbhtt:
bot, build org.archivekeep.ArchiveKeep

PR #6304, opened 2025-03-22
Add org.azahar_emu.Azahar

OpenSauce04:

**Azahar is a 3DS emulator which was created due to Lime3DS merging with another Citra fork known as PabloMK7's fork. This new project has replaced both of the prior projects.**
- [x] I am an author/developer/upstream contributor to the project. If not, I contacted upstream developers about this submission. **Co-owner**

### DIFF
--- a/org.azahar_emu.Azahar.json
+++ b/org.azahar_emu.Azahar.json
@@ -1,0 +1,133 @@
+{
+    "app-id": "org.azahar_emu.Azahar",
+    "runtime": "org.kde.Platform",
+    "runtime-version": "6.8",
+    "sdk": "org.kde.Sdk",
+    "sdk-extensions": [
+        "org.freedesktop.Sdk.Extension.llvm18"
+    ],
+    "command": "azahar-launcher",
+    "build-options": {
+        "append-path": "/usr/lib/sdk/llvm18/bin",
+        "prepend-ld-library-path": "/usr/lib/sdk/llvm18/lib",
+        "cflags": "-Wno-unused-command-line-argument",
+        "cxxflags": "-Wno-unused-command-line-argument"
+    },
+    "finish-args": [
+        "--device=all",
+        "--socket=wayland",
+        "--socket=fallback-x11",
+        "--socket=pulseaudio",
+        "--share=network",
+        "--share=ipc",
+        "--filesystem=host:ro",
+        "--filesystem=xdg-run/app/com.discordapp.Discord:ro",
+        "--filesystem=xdg-desktop:create",
+        "--filesystem=xdg-data/applications:create"
+    ],
+    "cleanup": [
+        "/include",
+        "/bin/glslangValidator",
+        "/bin/glslang",
+        "/bin/spirv-*",
+        "/bin/sdl2-config",
+        "/lib/pkgconfig",
+        "/lib/cmake",
+        "/share/doc",
+        "/share/man",
+        "/src",
+        "*.a",
+        "*.la"
+    ],
+    "modules": [
+        {
+            "name": "glslang",
+            "buildsystem": "cmake-ninja",
+            "config-opts": [
+                "-DCMAKE_BUILD_TYPE=Release"
+            ],
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://github.com/KhronosGroup/glslang/archive/refs/tags/15.1.0.zip",
+                    "sha256": "c4cc5083dd198640f3ad9d2014ac9b31433495a57cd3fb2420cd4e8ea8eb4b2b"
+                },
+                {
+                    "type": "archive",
+                    "url": "https://github.com/KhronosGroup/SPIRV-Tools/archive/refs/tags/vulkan-sdk-1.4.304.1.tar.gz",
+                    "sha256": "9fe736980d424c04f1303ae71b94b18bcc6046ae348909c393344a45e1bd7ff8",
+                    "dest": "External/spirv-tools"
+                },
+                {
+                    "type": "archive",
+                    "url": "https://github.com/KhronosGroup/SPIRV-Headers/archive/refs/tags/vulkan-sdk-1.4.304.1.tar.gz",
+                    "sha256": "66e6cec19e7433fc58ace8cdf4040be0d52bb5920e54109967df2dd9598a8d48",
+                    "dest": "External/spirv-tools/external/spirv-headers"
+                }
+            ]
+        },
+        {
+            "name": "rapidjson",
+            "buildsystem": "cmake-ninja",
+            "config-opts": [
+                "-DCMAKE_BUILD_TYPE=Release",
+                "-DRAPIDJSON_ENABLE_INSTRUMENTATION_OPT=OFF",
+                "-DRAPIDJSON_BUILD_DOC=OFF",
+                "-DRAPIDJSON_BUILD_EXAMPLES=OFF",
+                "-DRAPIDJSON_BUILD_TESTS=OFF",
+                "-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON"
+            ],
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://github.com/Tencent/rapidjson/archive/v1.1.0.tar.gz",
+                    "sha256": "bf7ced29704a1e696fbccf2a2b4ea068e7774fa37f6d7dd4039d0787f8bed98e",
+                    "x-checker-data": {
+                        "type": "anitya",
+                        "stable-only": true,
+                        "project-id": 7422,
+                        "url-template": "https://github.com/Tencent/rapidjson/archive/v$version.tar.gz"
+                    }
+                }
+            ]
+        },
+        {
+            "name": "azahar",
+            "buildsystem": "cmake-ninja",
+            "builddir": true,
+            "config-opts": [
+                "-DCMAKE_BUILD_TYPE=Release",
+                "-DCMAKE_C_COMPILER=clang",
+                "-DCMAKE_CXX_COMPILER=clang++",
+                "-DCMAKE_LINKER=lld",
+                "-DENABLE_QT_TRANSLATION=ON",
+                "-DUSE_DISCORD_PRESENCE=ON",
+                "-DUSE_SYSTEM_SDL2=ON"
+            ],
+            "cleanup": [
+                "/share/man",
+                "/share/pixmaps"
+            ],
+            "post-install": [
+                "install -Dm755 ../azahar-launcher.sh /app/bin/azahar-launcher",
+                "install -Dm644 ../org.azahar_emu.Azahar.metainfo.xml /app/share/appdata/org.azahar_emu.Azahar.metainfo.xml",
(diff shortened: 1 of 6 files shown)


OpenSauce04:
It's worth noting that as this manifest is based on the now-deprecated Lime3DS, which had an exception for `finish-args-unnecessary-xdg-data-applications-create-access` to allow a feature which creates game shortcuts to function, that exception will also be required for this app.

Additionally, once Azahar has been merged, I am going to update the Lime3DS manifest to include an EOL rebase.

OpenSauce04:
bot, build org.azahar_emu.Azahar

OpenSauce04:
Build is failing because of the aforementioned required exception

hfiguiere:
> Build is failing because of the aforementioned required exception

why do you need them?

org.azahar_emu.Azahar.metainfo.xml
hfiguiere, line 1: this file should be (submitted and managed) upstream. And installed as part of the standard build / install procedure.

org.azahar_emu.Azahar.json
```diff
@@ -0,0 +1,143 @@
...
+                "-DUSE_SYSTEM_SDL2=ON"
+            ],
+            "cleanup": [
+                "/share/man",
+                "/share/pixmaps"
+            ],
+            "post-install": [
+                "install -Dm755 ../azahar-launcher.sh /app/bin/azahar-launcher",
+                "install -Dm644 ../org.azahar_emu.Azahar.metainfo.xml /app/share/appdata/org.azahar_emu.Azahar.metainfo.xml",
+                "desktop-file-install --dir=/app/share/applications ../dist/azahar.desktop",
```
hfiguiere, line 121: why isn't this install by the build system?

org.azahar_emu.Azahar.json
```diff
@@ -0,0 +1,143 @@
...
+                "desktop-file-edit --set-key StartupWMClass --set-value azahar --set-key Exec --set-value azahar-launcher /app/share/applications/azahar.desktop",
+                "install -Dm644 ../dist/azahar.svg /app/share/icons/hicolor/scalable/apps/azahar.svg",
+                "install -Dm644 ../dist/icon.png /app/share/icons/hicolor/512x512/apps/azahar.png"
```
hfiguiere, line 124: why aren't the icons installed by the build system?

OpenSauce04:
> > Build is failing because of the aforementioned required exception
> 
> why do you need them?

I described it in the comment above, the emulator has a feature which allows you to create game shortcuts, and without this permission the function doesn't work.

hfiguiere:
> > > Build is failing because of the aforementioned required exception
> > 
> > 
> > why do you need them?
> 
> I described it in the comment above, the emulator has a feature which allows you to create game shortcuts, and without this permission the function doesn't work.

There is portal to create launchers dynamically.

org.azahar_emu.Azahar.metainfo.xml
OpenSauce04: Outside of the Flatpak build, we only distribute the Linux builds as an AppImage, meaning this file isn't relevant outside of a Flatpak context.

Additionally, the repository is organized in a way which doesn't require updating version numbers when we do a new release, and I have no interest in adding a file which will have to be updated before each time we do a release.

Upstreaming this file would also mean that in order to update the description or screenshots of the app, we would need to release a whole new update for the entire emulator, which objectively sucks.

hfiguiere: I hate to break it to you but the appstream file is NOT specific to flatpak. AppImage can also ship them. So do distro package that could be based of your upstream source distribution. 

> Upstreaming this file would also mean that in order to update the description or screenshots of the app, we would need to release a whole new update for the entire emulator, which objectively sucks.

No. It would just be that the metadata doesn't change unless there is a new release, which is somewhat relevant because there is no reason to change the metadata without a new release as that metadata is part of the release.

OpenSauce04: Your comment also doesn't address my largest issue, which is the need for the file to be updated each time.

Also, this seems weirdly opinionated considering that it doesn't affect Flathub at all. What we do with this file, it's being distributed via Flathub all the same.

hfiguiere: updating it must be. better do that upstream, no?

hfiguiere: The flathub repo should contain the minimum. Provisions for having things that should be upstream (ie everything but the manifest) are here because sometime it's proprietary, the project is unmaintained or upstream is uncooperative. The latter reflect on upstream.

OpenSauce04: > updating it must be. better do that upstream, no?

As I mentioned, there are no files in our repository that are tied to version numbers, and instead the version number is determined at compile-time using GitHub Actions environment variables.

Considering that whenever I update the emulator I have to update the Flatpak manifest as well to point to the new source archive, it's far more straightforward to simply update both the manifest and the metadata file in a single commit.

Upstreaming this file would meaningfully inconvenience me every single time I release an update in a completely avoidable way. We've been doing it like this for a while now with Lime3DS and it hasn't been problematic, and technically speaking this is basically just a large update for Lime3DS, so I don't think there should be an issue now.

org.azahar_emu.Azahar.json
OpenSauce04: It looks like it might be?
Screenshot_20250322_154144

This is left over from Citra's manifest, so I'm not sure what the logic behind it was. Does it seem safe to remove?

org.azahar_emu.Azahar.json
OpenSauce04: Same as the previous comment, but the 512x512 icon wasn't installed by the build system. I will add that upstream.

org.azahar_emu.Azahar.json
OpenSauce04: Yeah I tried removing this and the other thing from the Flatpak manifest and it seems to work just fine. Must be a relic of the past.

OpenSauce04:
@hfiguiere Using the dynamic launcher portal seems to require including GLib as a build dependency to access GIO, the library which provides the `GBytesIcon`, which is stated as a requirement for calling org.freedesktop.portal.DynamicLauncher.PrepareInstall: https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.DynamicLauncher.html#org-freedesktop-portal-dynamiclauncher-prepareinstall

I'd really rather avoid this, as Azahar is a Qt application, and including the entirety of GLib just for just this seems like major overkill. Is that exception still on the table?

hfiguiere:
Qt already link against glib....

OpenSauce04:
As far as I'm aware it doesn't? GLib is GTK territory. It looks like Qt *can* be used alongside GLib, but it isn't a dependency of Qt or anything.

hfiguiere:
you are aware wrong.

OpenSauce04:
> you are aware wrong.

I looked into it further, and while it seems that Qt doesn't hard-depend on GLib, it does have integrations with GLib which can be turned on at compile time and are enabled on basically every distro out there (not even Gentoo has a USE flag to disable it), so this shouldn't be an issue.

Constructive comments would be nice, I'm trying to be cooperative. You can correct me without the snark 🙃

JakobDev:
You don't need a `GBytesIcon`. The documentation is misleading. You need a structure with the string `strings` followed by the bytes of the icon as variant. You can find a [working implemntation in PyQt5 here](https://github.com/flatpak/xdg-desktop-portal/issues/797#issuecomment-1169610692). You just need to port it to C++ and you are good to go.

But please be aware that this API has some limitations:
- [You can't create shortcuts on the desktop](https://github.com/flatpak/xdg-desktop-portal/issues/827), only in the menu
- [The API doesn't work outside a Flatpak](https://github.com/flatpak/xdg-desktop-portal/issues/826)
- This API might not be implemented on every Desktop

org.azahar_emu.Azahar.json
```diff
@@ -0,0 +1,140 @@
+{
+    "app-id": "org.azahar_emu.Azahar",
+    "runtime": "org.kde.Platform",
+    "runtime-version": "6.8",
+    "sdk": "org.kde.Sdk",
+    "sdk-extensions": [
+        "org.freedesktop.Sdk.Extension.llvm18"
+    ],
+    "command": "azahar-launcher",
+    "rename-desktop-file": "azahar.desktop",
```
JakobDev, line 10: In azahar-emu/azah#582 you wrote

> To my knowledge this is not an upstream requirement

but now you are renaming the files. I highly recommend you to change the name upstream instead of here. Some things in Linux (e.g. the window icon on some Wayland compositors) are tied to the desktop file, so this would avoid problems.

OpenSauce04: Alright then, I'll update the desktop file name for the next minor update

OpenSauce04:
> But please be aware that this API has some limitations:
> 
>     * [You can't create shortcuts on the desktop](https://github.com/flatpak/xdg-desktop-portal/issues/827), only in the menu

Oh, then I don't think this portal is applicable anyway because the feature does allow desktop shortcuts to be created:
Screenshot_20250324_111606

OpenSauce04:
> > But please be aware that this API has some limitations:
> > 
> >  - [You can't create shortcuts on the desktop](https://github.com/flatpak/xdg-desktop-portal/issues/827), only in the menu
> 
> Oh, then I don't think this portal is applicable anyway because the feature does allow desktop shortcuts to be created:
> 
> Screenshot_20250324_111606
> 
> The other limitations aren't exactly encouraging either. Is our existing method fine then?

@JakobDev sorry to ping you, but I want to get everything required for the Flathub release into the next minor update, so figuring out the details is a priority.

JakobDev:
I'm not a Flathub Admin, so I dont decide if it's ready, but I can help you if you have a specific question.

org.azahar_emu.Azahar.json
JakobDev: Would be nice if you update the other names too

OpenSauce04: > Would be nice if you update the other names too

Other names? Like the launcher script?

org.azahar_emu.Azahar.metainfo.xml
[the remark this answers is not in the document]
bbhtt: ok didn't see the previous discussion

bbhtt:
bot, build org.azahar_emu.Azahar

OpenSauce04:
Build will be failing because of the `finish-args-unnecessary-xdg-data-applications-create-access`. As per the earlier conversation in this thread, the portal which allows creating shortcuts isn't fit for this feature because it doesn't allow creating desktop shortcuts.

bbhtt:
I didn't notice the previous conversation. Open a PR with the exception https://docs.flathub.org/docs/for-app-authors/linter#exceptions

OpenSauce04:
Sure thing, I'll report back here when it's merged.

OpenSauce04:
bot, build org.azahar_emu.Azahar

PR #6305, opened 2025-03-22
Add eu.codepoems.xl-converter

JacobDev1:

- [x] Please describe the application briefly. XL Converter is an image converter focused on modern formats.
- [x] I am an author/developer/upstream contributor to the project. If not, I contacted upstream developers about this submission. **Yes, I'm the author of this application.**


eu.codepoems.xl-converter.yaml
```diff
@@ -0,0 +1,336 @@
...
+finish-args:
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --filesystem=host
+  - --socket=pulseaudio
+  - --share=network
+modules:
+  
+  - name: libjxl
```
hfiguiere, line 26: this is in the runtime.

eu.codepoems.xl-converter.yaml
```diff
@@ -0,0 +1,336 @@
...
+      - oxipng_sources.json
+      - type: git
+        url: https://github.com/shssoichiro/oxipng.git
+        tag: v9.1.3
+        commit: e8e8309c2da99676d2be0859df66314fa4fd42c3
+        x-checker-data:
+          type: git
+          tag-pattern: ^v([\d.]+)$
+  
+  - name: libavif
```
hfiguiere, line 95: this is in the runtime.

eu.codepoems.xl-converter.yaml
```diff
@@ -0,0 +1,336 @@
...
+              - type: script
+                dest-filename: configure
+                commands:
+                  - exec ./configure.gnu $@
+
+  - name: xl-converter
+    buildsystem: simple
+    build-commands:
+      - install -Dm644 eu.codepoems.xl-converter.metainfo.xml /app/share/metainfo/eu.codepoems.xl-converter.metainfo.xml
+      - sed -i -E "s/^Icon=.*$/Icon=eu.codepoems.xl-converter/" xl-converter/misc/xl-converter.desktop
```
hfiguiere, line 314: don't use sed to change the desktop file. use `desktop-file-edit`.

eu.codepoems.xl-converter.metainfo.xml
hfiguiere, line 1: this should be (submitted and managed) upstream

eu.codepoems.xl-converter.yaml
JacobDev1: This application uses a custom-compiled version of libavif with SVT-AV1-PSY.

eu.codepoems.xl-converter.metainfo.xml
```diff
@@ -0,0 +1,94 @@
...
+            <li>Encode better quality JPEG images with JPEGLI.</li>
+            <li>Losslessly recompress your JPEG images.</li>
+            <li>Convert faster with parallel encoding.</li>
+            <li>Scale down images and preserve metadata with ExifTool.</li>
+            <li>QoL features available: drag 'n drop, dark theme, keep directory structure, copy smaller sources, and many more!</li>
+        </ul>
+    </description>
+    <launchable type="desktop-id">eu.codepoems.xl-converter.desktop</launchable>
+    <icon type="stock">eu.codepoems.xl-converter</icon>
+    <icon type="remote">https://raw.githubusercontent.com/JacobDev1/xl-converter/refs/tags/v1.2.0/assets/icons/logo.svg</icon>
```
hfiguiere, line 26:
```suggestion
```

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+    "only-arches": ["x86_64"]
```
hfiguiere, line 2: why only x86_64?

eu.codepoems.xl-converter.yaml
```diff
@@ -0,0 +1,336 @@
...
+    sources:
+      - type: git
+        url: https://github.com/ImageMagick/ImageMagick.git
+        tag: 7.1.1-43
+        commit: a2d96f40e707ba54b57e7d98c3277d3ea6611ace
+        x-checker-data:
+          type: git
+          tag-pattern: ^([\d.]+-\d+)$
+
+  - name: exiftool
```
hfiguiere, line 215: have you tested that exiftool works in the produced flatpak?

eu.codepoems.xl-converter.yaml
```diff
@@ -0,0 +1,336 @@
...
+        url: https://github.com/shssoichiro/oxipng.git
+        tag: v9.1.3
+        commit: e8e8309c2da99676d2be0859df66314fa4fd42c3
+        x-checker-data:
+          type: git
+          tag-pattern: ^v([\d.]+)$
+  
+  - name: libavif
+    buildsystem: simple
+    build-commands:
```
hfiguiere, line 97: this should be a `buildsystem: cmake-ninja`. You can also add a `shell` source to do stuff prior to that. (albeit using sed like that is unmaintainable)

hfiguiere: but see note about it being in the runtime already.

eu.codepoems.xl-converter.yaml
JacobDev1: Yes, all necessary binaries were tested and work correctly. In this case, I also added Brotli for JPEG XL support.

eu.codepoems.xl-converter.yaml
```diff
@@ -0,0 +1,336 @@
...
+    - BASEAPP_REMOVE_WEBENGINE=1
+    - BASEAPP_DISABLE_NUMPY=1
+  append-path: /usr/lib/sdk/rust-stable/bin
+cleanup-commands:
+  - /app/cleanup-BaseApp.sh
+finish-args:
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --filesystem=host
```
hfiguiere, line 21: why this overbroad file system permission?

flathub.json
JacobDev1: I have not yet tested other architectures. Traditionally, I have always provided x86_64 builds. Would you like me to test it on "aarch64"?

eu.codepoems.xl-converter.yaml
```diff
@@ -0,0 +1,336 @@
...
+      - -DJPEGXL_ENABLE_VIEWERS=OFF
+      - -DJPEGXL_ENABLE_PLUGINS=OFF
+      - -DJPEGXL_ENABLE_OPENEXR=OFF
+      - -DJPEGXL_ENABLE_DEVTOOLS=OFF
+    post-install:
+      - mkdir -p /app/xl-converter/bin/linux
+      - install -Dm755 /app/bin/cjxl /app/xl-converter/bin/linux/cjxl
+      - install -Dm755 /app/bin/djxl /app/xl-converter/bin/linux/djxl
+      - install -Dm755 /app/bin/jxlinfo /app/xl-converter/bin/linux/jxlinfo
+      - install -Dm755 /app/bin/cjpegli /app/xl-converter/bin/linux/cjpegli
```
hfiguiere, line 41: why do you duplicate these?

eu.codepoems.xl-converter.yaml
```diff
@@ -0,0 +1,336 @@
...
+        commit: 794a5dcf0d54f9f0b20d288a12e87afb91d20dfc
+        x-checker-data:
+          type: git
+          tag-pattern: ^v([\d.]+)$
+
+  - name: libjpeg-turbo
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DENABLE_STATIC=TRUE
```
hfiguiere, line 62: why static?
`jpegtran` is in the runtime as well and libjpegturbo

eu.codepoems.xl-converter.yaml
JacobDev1: They are removed by the `cleanup` routine so only one copy remains. The application relies on statically linked binaries placed inside its `bin` directory.

hfiguiere: don't use statically linked binaries either.

eu.codepoems.xl-converter.yaml
hfiguiere: ok, I missed perl being compiled deep down.

eu.codepoems.xl-converter.yaml
JacobDev1: The application relies heavily on drag 'n drop and is often used with files from all over the system. Perhaps, we can reduce it, but I'm unsure how would I handle asking for permission on drag 'n drop.

JacobDev1: This decision was made by referencing the HandBrake manifest with identical permissions.

eu.codepoems.xl-converter.yaml
JacobDev1: I will look into that.

eu.codepoems.xl-converter.yaml
hfiguiere: two wrongs don't make a right.

it doesn't even use the same toolkit.

AFAIK qt uses the portal for document drag and drop.

eu.codepoems.xl-converter.yaml
JacobDev1: For this program `libjxl` is its most crucial part. I would like to have the control over the version of `libjxl` used. If syntax or behavior changes between them, then the program will stop working correctly. There are many subtle differences, and having one common base for all builds is very important.

eu.codepoems.xl-converter.yaml
JacobDev1: Pardon me, in this case the binaries are dynamic. I'll look into reducing the reliance on statically linked binaries. However, I ran into issues when building `libavif` dynamically with custom codecs, which is why this library is static.

eu.codepoems.xl-converter.yaml
JacobDev1: I suppose the `DENABLE_STATIC=TRUE` is unnecessary here. I'll check if I can use the included `jpegtran`.

eu.codepoems.xl-converter.yaml
hfiguiere: the version will not change for the same runtime version.

also command line is NOT an ABI. Doing so is a self-inflicted problem.

JacobDev1: I am aware using CLI was a suboptimal decision. When I made it in 2023, I planned for this program to only parallelize `cjxl`. Still, It is fully-functional. Changing it would require rewriting the entire thing. I understand your point though.

I would like to keep the control over `libjxl` if possible. With each new release, I test it thoroughly. This is the cutting-edge and as such bugs happen. A lot of people rely on it for JPEG XL. It is a highly-specialized program and the control over its most crucial part is desirable.

JacobDev1: `cjxl` does not seem to be a part of the Sdk.

```bash
Running: which cjxl
which: no cjxl in (/run/ccache/bin:/app/bin:/usr/bin:/usr/lib/sdk/rust-stable/bin)
```

bbhtt: binaries aren't shipped those are usually used for debugging things, not supposed to be used by external software.

JacobDev1: This program relies on tool binaries like `cjxl`. Compiling `libjxl` is necessary to have them included.

eu.codepoems.xl-converter.yaml
JacobDev1: I reduced the permissions to `home` and `/media`. The former is because users store images not just in `xdg-pictures`, but all over the home dir. The latter is to read / write to mounted volumes like USB sticks, archival HDD, or other disks. Lossless JPEG Transcoding is commonly used for archival, so accessing `/media` is expected to work out of the box.

Reducing them further is not ideal. Drag 'n drop does not use portals in Qt ([see this issue](https://bugreports.qt.io/browse/QTBUG-91357)). This program is a batch file processor with advanced output options. As such, it will not work without filesystem permissions.

eu.codepoems.xl-converter.yaml
JacobDev1: > this is in the runtime.

`avifenc` (part of `libavif`) is not in the runtime. `libavif` needs to be compiled for `avifenc` be included. The program relies on it for AVIF encoding.

```
Running: which avifenc
which: no avifenc in (/run/ccache/bin:/app/bin:/usr/bin:/usr/lib/sdk/rust-stable/bin)
```

bbhtt: why aren't you using the libavif API for these?

JacobDev1: > why aren't you using the libavif API for these?

This approach:
- Made sense at the time when it was supposed to be a simple utility.
- Bypasses Python GIL and allows for easy parallelization.
- Avoids the need for custom bindings, which would be required to match the feature-set of `libjxl` and `libavif` apps.

This is not technically ideal, but fully functional. Swapping to the API would require rewriting the entire backend.

A similar example on Flathub is [Lossless Cut](https://github.com/flathub/no.mifi.losslesscut/blob/master/no.mifi.losslesscut.yaml) which calls FFmpeg in a similar way.

bbhtt: ffmpeg is a bit different, the API is not good

eu.codepoems.xl-converter.yaml
JacobDev1: I swapped `install` to `mv` + `chmod`. The end result is identical; only one copy of apps remain. `libjxl` is already compiled dynamically.

eu.codepoems.xl-converter.yaml
JacobDev1: I reworked the `libavif` module to use `buildsystem: cmake-ninja`, `shell` source, and runtime libraries where possible.

`avifenc` and `avifdec` are not included in the runtime so `libavif` must be compiled.

flathub.json
JacobDev1: I looked into building it for `aarch64`. Unfortunately, the Rust toolchain appears to be unavailable for this architecture.

```bash
flatpak run org.flatpak.Builder build --arch=aarch64 --user --ccache --install --force-clean eu.codepoems.xl-converter.yaml
error: Requested extension org.freedesktop.Sdk.Extension.rust-stable/aarch64/6.8 not installed
$ flatpak install org.freedesktop.Sdk.Extension.rust-stable/aarch64/6.8
Looking for matches…
error: No remote refs found similar to ‘org.freedesktop.Sdk.Extension.rust-stable/aarch64/6.8’
```

That is okay as I have not planned to support `aarch64`.

hfiguiere: install the rust sdk beforehand. it will work.

https://github.com/flatpak/flatpak-builder/issues/494

hfiguiere: the version is 24.08

JacobDev1:
Please review the changes I submitted.

The next tagged release of XL Converter will be published soon, which will move metainfo to the upstream. Would you like to propose any other changes before that happens so I can include them in the next release? Perhaps you might want to test the build? Let me know what else is needed.

JacobDev1:
I decided not to wait and move the metainfo now. All requested changes have been addressed.

eu.codepoems.xl-converter.yaml
```diff
@@ -0,0 +1,293 @@
...
+      - desktop-file-edit --set-icon=eu.codepoems.xl-converter xl-converter/misc/xl-converter.desktop
+      - install -Dm644 xl-converter/misc/xl-converter.desktop /app/share/applications/eu.codepoems.xl-converter.desktop
+      - install -Dm644 xl-converter/assets/icons/logo.svg /app/share/icons/hicolor/scalable/apps/eu.codepoems.xl-converter.svg
+      - install -Dm755 run.sh /app/bin/run.sh
+      - mkdir -p /app/xl-converter
+      - cp -r xl-converter/* /app/xl-converter/
+      - mkdir -p /app/xl-converter/bin/linux
+      - |
+        echo '#!/bin/bash
+        exec /usr/bin/jpegtran "$@"' > /app/xl-converter/bin/linux/jpegtran
```
bbhtt, line 280: use a script source for this https://docs.flatpak.org/en/latest/module-sources.html#shell-and-script-sources

eu.codepoems.xl-converter.yaml
```diff
@@ -0,0 +1,294 @@
...
+        dest: xl-converter
+        branch: stable
+        commit: f9588e507c04334c409255442b15332c3659b45b
+      - type: script
+        commands:
+          - python3 /app/xl-converter/main.py
+        dest-filename: run.sh
+      - type: script
+        commands:
+          - #!/bin/bash
```
bbhtt, line 291:
```suggestion
```

script adds this automatically.

bbhtt:
bot, build eu.codepoems.xl-converter

JacobDev1:
@bbhtt The test build is ready to be reviewed.

eu.codepoems.xl-converter.yaml
bbhtt: Qt doesn't support drag and drop portal yet.

eu.codepoems.xl-converter.yaml
```diff
@@ -0,0 +1,293 @@
...
+                  tag-pattern: ^v([\d.]+)$
+          - name: perl
+            buildsystem: simple
+            build-commands:
+              - ./Configure -des -Duseshrplib -Dman3dir=none -Dman1dir=none -Dprefix=/app
+              - make
+              - make install
+            cleanup:
+              - /bin/corelist
+              - /bin/cpan
```
bbhtt, line 218: is this everything in /bin that this installs or you are keeping something? Wondering why not just use `/bin/*`?

JacobDev1: This part is taken from the official [flatpak-cpan-generator](https://github.com/flatpak/flatpak-builder-tools/tree/master/cpan) README. I assumed there was a reason for it to look this way. The `perl` interpreter is excluded from the cleanup. This is needed to run ExifTool.

PR #6310, opened 2025-03-23
Add cc.cubocore.CoreTime

rahmanshaber:
This PR is to update the domain/ name org.cubocore.CoreTime to cc.cubocore.CoreTime . It also updates the version to the newest v5.0.0

- [X] Please describe the application briefly. <CoreTime is a very simple Clock/Calendar application which shows a clock, supports alarms, reminders, sports a stop watch and a timer>


cc.cubocore.CoreTime.appdata.xml
hfiguiere, line 1: this should be sumbitted / managed upstream. I'm pretty sure this was requested in the original submission too.

cc.cubocore.CoreTime.appdata.xml
```diff
@@ -0,0 +1,92 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2021 CuboCore -->
+
+<component type="desktop-application">
+    <provides>
+        <id>cc.cubocore.CoreTime</id>
+    </provides>
+    <replaces>
+        <id>org.cubocore.CoreTime</id>
+    </replaces>
```
hfiguiere, line 10: it's either provides or replace not both. Also it's missing the actual id.

cc.cubocore.CoreTime.appdata.xml
rahmanshaber: sorry, I didn't understood

cc.cubocore.CoreTime.yml
```diff
@@ -0,0 +1,44 @@
...
+runtime-version: '6.8'
+sdk: org.kde.Sdk
+command: coretime
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --share=network # Required to show weather info
+  - --socket=wayland
+  - --filesystem=home
+  - --filesystem=host
```
hfiguiere, line 12: these two filesystem permission are very broad and redundant (one includes the other)

cc.cubocore.CoreTime.yml
```diff
@@ -0,0 +1,44 @@
...
+  - --socket=pulseaudio # Required for ringing alarm
+  - --device=dri
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=org.freedesktop.Notifications
+  - --own-name=org.kde.* # For showing in the system tray
```
hfiguiere, line 17: the system tray doesn't work with this. Has never.

cc.cubocore.CoreTime.appdata.xml
hfiguiere: the file should be submitted and managed upstream.

https://github.com/flathub/flathub/pull/2257#issuecomment-820098777

cc.cubocore.CoreTime.yml
bbhtt: it's also blocked in linter, so you can't use this.

cc.cubocore.CoreTime.appdata.xml
```diff
@@ -0,0 +1,92 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2021 CuboCore -->
+
+<component type="desktop-application">
+    <provides>
```
bbhtt, line 5: where is the id tag here?

cc.cubocore.CoreTime.appdata.xml
```diff
@@ -0,0 +1,92 @@
...
+        <id>cc.cubocore.CoreTime</id>
```
bbhtt, line 6:
```suggestion
        <id>org.cubocore.CoreTime</id>
```

it cannot provide itself, that's a circular relation.

cc.cubocore.CoreTime.appdata.xml
```diff
@@ -0,0 +1,92 @@
...
+        <release version="4.5.0" date="2023-08-27" />
+        <release version="4.4.0" date="2022-08-02" />
+        <release version="4.3.0" date="2021-12-25" />
+    </releases>
+
+    <screenshots>
+        <screenshot type="default">
+            <caption>CoreTime</caption>
+            <image type="source">
+                https://gitlab.com/cubocore/coreapps/coretime/-/raw/master/coretime.png</image>
```
bbhtt, line 73:
```suggestion
            <image type="source">https://gitlab.com/cubocore/coreapps/coretime/-/raw/68ca8b0922afec80a31e0d8d269cc5599eed265a/coretime.png</image>
```

cc.cubocore.CoreTime.yml
```diff
@@ -0,0 +1,44 @@
...
+  - --share=ipc
+  - --socket=fallback-x11
+  - --share=network # Required to show weather info
+  - --socket=wayland
+  - --filesystem=home
+  - --filesystem=host
+  - --socket=pulseaudio # Required for ringing alarm
+  - --device=dri
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=org.freedesktop.Notifications
```
bbhtt, line 16: notifications does not go through the portal?

cc.cubocore.CoreTime.yml
```diff
@@ -0,0 +1,44 @@
...
+modules:
+  - name: libcprime
+    buildsystem: cmake-ninja
+    sources:
+      - type: git
+        url: https://gitlab.com/cubocore/libcprime.git
+        tag: v5.0.0
+  - name: coretime
+    post-install:
+      - install -Dp -m 644 org.cubocore.CoreTime.appdata.xml /app/share/appdata/org.cubocore.CoreTime.appdata.xml
```
bbhtt, line 36: this should be done by your buildsystem

cc.cubocore.CoreTime.yml
rahmanshaber: No, tested without it and it doesn't.

cc.cubocore.CoreTime.appdata.xml
rahmanshaber: pls can you explain, what's the difference?

rahmanshaber:
bot, build org.cubocore.CoreTime

cc.cubocore.CoreTime.appdata.xml
rahmanshaber: hi, as the appdata is removed, pls check if this is correct or not https://gitlab.com/cubocore/coreapps/coretime/-/blob/master/cc.cubocore.CoreTime.metainfo.xml?ref_type=heads#L6

rahmanshaber:
bot, build cc.cubocore.CoreTime

rahmanshaber:
bot, build cc.cubocore.CoreTime

cc.cubocore.CoreTime.yml
```diff
@@ -0,0 +1,38 @@
...
+  - /share/man
+  - /lib/*.la
+  - /lib/*.a
+modules:
+  - name: libcprime
+    buildsystem: cmake-ninja
+    sources:
+      - type: git
+        url: https://gitlab.com/cubocore/libcprime.git
+        tag: v5.0.0
```
hfiguiere, line 31: also add the commit here to protect against retagging,

cc.cubocore.CoreTime.yml
```diff
@@ -0,0 +1,38 @@
...
+  - name: coretime
+    buildsystem: cmake-ninja
+    sources:
+      - type: git
+        url: https://gitlab.com/cubocore/coreapps/coretime.git
+        tag: v5.0.1
```
hfiguiere, line 37: also add the commit here to protect against retagging,

rahmanshaber:
bot, build cc.cubocore.CoreTime

rahmanshaber:
all good i guess, and do i have to do anything with the package that this replacing?

cc.cubocore.CoreTime.appdata.xml
bbhtt: > hi, as the appdata is removed, pls check if this is correct or not https://gitlab.com/cubocore/coreapps/coretime/-/blob/master/cc.cubocore.CoreTime.metainfo.xml?ref_type=heads#L6

Add:

```
    <provides>
        <id>org.cubocore.CoreTime</id>
    </provides>
```

too

rahmanshaber:
bot, build cc.cubocore.CoreTime

cc.cubocore.CoreTime.yml
```diff
@@ -35,6 +35,6 @@ modules:
     sources:
       - type: git
         url: https://gitlab.com/cubocore/coreapps/coretime.git
-        commit: ed06e33cde531d49e277a10512e86bef34994108
+        commit: e5974464bb0934ed72db728676251e0eb96498f1
```
hfiguiere, line 38: I will note that retagging is a bad practice. (it's not specific to us)

rahmanshaber: I know, 
what's the problem with the commit number here? I updated it to the latest.

cc.cubocore.CoreTime.yml
```diff
@@ -0,0 +1,40 @@
...
+runtime-version: '6.8'
+sdk: org.kde.Sdk
+command: coretime
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --share=network # Required to show weather info
+  - --socket=wayland
+  - --filesystem=host
+  - --socket=pulseaudio # Required for ringing alarm
```
bbhtt, line 12: why does a "time related task manager" require access to host and sound?

rahmanshaber: host access to have a single settings file so all other apps in the suite act same, and sound access for ringing alarms as written in the comments

rahmanshaber: is this a blocker?

bbhtt: > host access to have a single settings file so all other apps

why would that need whole host access? where are you storing these?

rahmanshaber: at home/.config directory. just adding home dir access doesn't work, only host access works. Also if the alarm sound file in a drive other then home, app can't able to access them.

bbhtt: This doesn't make sense, if it's `home/.config` it should need that permission only, why would it need the rest of host?

rahmanshaber:
@bbhtt @hfiguiere Hi, can I get this merged, I have 16 more to go. 
This app needs the host access to get it working, these apps got bad reviews and it end up being apps can't access the files properly.

bbhtt:
Your explanation doesn't make sense much to me. Rather than submitting 16 more apps, it'd be better to figure out why they'd need `host`, considering config files are in `home/.config`.

rahmanshaber:
I'm not arguing about the config file part access with this one, as i don't have any clue what causes it. 
what I'm asking host access is for the external drive file access.

bbhtt:
External media access is fine.

rahmanshaber:
so, can we merge this, pls?

bbhtt:
That doesn't need full host. Also sound file in an external drive seems like an edge case. It's going to be a small file, it can be easily copied onto xdg-music or somewhere.

rahmanshaber:
it will be changed by the user, and they can chose it from an external media that's mounted to any other places that you mentioned.

rahmanshaber:
@bbhtt HI, are you still not satisfied with the answer?

bbhtt:
As I said, a task manager doesn't need host access. Either you should use portal filchooser or limit it to known xdg directories like xdg-music etc.

bbhtt:
bot, build cc.cubocore.CoreTime

bbhtt:
Also Flatpak apps aren't supposed to share config with host.

bbhtt:
Looks like it is already using portal filechoosers, the path returned by the portal stays stable for the session. Is that not enough?

rahmanshaber:
- First it's not a Task Manager, If you run the app it's a clock,weather,calendar app and based on time, weather one can set alarms and reminders, that's why we are calling it time related task manager. 
- The alarm sound file can be changed by the user using our settings/tweaks app (which is a separate app), and this app needs to retrieve that file to play the sound. We didn't used host in the v4.5 flatpak package and that broke these two functionality (settings and sound file). 
- We use common settings file for 16 of the apps, I know people might disagree with this but we won't change this. 
- We are using Flatpak/Flathub as it's easy to distribute packages, not for it's sandbox functionality. From our prospective, user knows what the app will get access to when they install it, it's clearly stated and the code is open source. 
- We already got many bad review and git issues by the user which is solely  related to how we did flatpak packaging, That's our fault, which I'm trying to fix.

Finally  if Flathub as a package distributor wants to control/ restrict how we make apps and it's functionality then it's unfortunate.

hfiguiere:
>  From our prospective, user knows what the app will get access to when they install it, 

No they won't.

rahmanshaber:
> > From our prospective, user knows what the app will get access to when they install it,
> 
> No they won't.

image

hfiguiere:
you wouldn't believe how many don't understand this.

rahmanshaber:
> you wouldn't believe how many don't understand this.

I understand this.

I explained what we think about the concerns raised here already. It's better for all our time if i get a clear answer, are you guys accepting the answer I gave and adding this and upcoming apps or not?

bbhtt:
If you are unlikely to change anything I'm also unlikely to accept this submission.

We absolutely will question and restrict permissions that don't make sense, it's being distributed through our platform and we are responsible for what is being added there.

PR #6327, opened 2025-03-29
Add org.vinegarhq.Sober

burnerphones:

Sober runs Roblox on Linux. The official Android Roblox app is downloaded at runtime, and linked with system libraries, similar to mcpe-launcher.

Even if the Android app is used, *all* Roblox features are available within Sober. So the usage of the Android app is moreso an implementation detail. We download the app at runtime, so there's no redistribution of Roblox IP here.

There's no emulators, virtual machines, or even really translation layers involved here. We have our own repository, where people have been testing Sober since August. We think it's worth a shot publishing to Flathub, though I understand this app may not be the right fit.

We estimate around 60k+ people used our app (derived from Automatic Downloads anonymous metrics, which aren't very accurate; we don't exactly do telemetry here).

We put a lot of effort into integrating with the Linux desktop, including using things like Wayland and libadwaita, staying on top of runtime updates, using the latest SDL3 releases, making sure Sober works great on all desktops.

For more info, I urge you to check out our website, online posts about our app, and of course, our metainfo description.

In terms of what Roblox thinks about us, they said we're not *not* allowed to do it. So you won't be liable for unfounded DMCAs if you accept this app.

To be clear, Sober's closed source because someone at Roblox suggested it'd reduce the likelihood of Sober getting abused for cheating, which would therefore allow the project to last longer without action from Roblox. We don't like proprietary software either, but I think allowing people to play games with friends on an otherwise FOSS operating system is nice. :)

Also, one more thing. Due to the nature of this app, we have to very closely follow Roblox's upstream, which sometimes does OTA config changes. For this reason, we need to roll out updates very quickly to users. I read online that Flathub intentionally adds a 3 hour delay to publishing updates, but given our manifest is very simple and only downloads a 2mb ZIP, perhaps we could be allow-listed to publish immediately? It's okay if not, I just wanted to ask.

- [X] I am an author/developer/upstream contributor to the project. If not, I contacted upstream developers about this submission. **Link:** I'm a developer of Sober since day 1, you can see me making commits to the official repository which is linked on the official website (https://sober.vinegarhq.org/): https://github.com/vinegarhq/sober/commits/gh-pages/

Thanks for considering my submission <3

Can you add @tunis4 as a maintainer as well? Thanks.

### DIFF
--- a/org.vinegarhq.Sober.yml
+++ b/org.vinegarhq.Sober.yml
@@ -13,8 +13,6 @@ finish-args:
   # SECURITY: shouldn't have much of an impact, although the user should disable this from Flatseal if controller support is unwanted
   # We don't ship this by default for backwards compat :( See above.
   # - --device=input
-  # DRI device. Required for graphics acceleration. Non-optional, since Roblox is a 3D game.
-  - --device=dri
   # Required by the game. Non-optional for now, sadly.
   # SECURITY: How bad is device=shm? should we try making a workaround?
   # SECURITY: ptrace access inside the sandbox should be no big deal, really :)
--- a/org.vinegarhq.Sober.yml
+++ b/org.vinegarhq.Sober.yml
@@ -34,9 +34,6 @@ finish-args:
   # SECURITY: Sober sees your Discord username and Discord ID due to how Discord IPC works, but it shouldn't be allowed to make actions on your behalf. Also, we obviously don't send that stuff to our server.
   - --filesystem=xdg-run/app/com.discordapp.Discord:create
   - --filesystem=xdg-run/discord-ipc-0
-  # Required for a workaround. Has no security or UX impact, since Sober is sadly English-only for now
-  # TODO: fix the bug and remove this
-  - --env=LC_ALL=en_US.UTF-8
 
 # We download prebuilt binaries from our infra.
 # We build them using GitHub Actions in a private GitHub repository, which pulls from our private source repo (not on GitHub), compiles the code, and uploads the artifact to our S3 bucket.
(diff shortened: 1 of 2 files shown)


org.vinegarhq.Sober.yml
```diff
@@ -0,0 +1,71 @@
...
+  # For controller support on old Flatpak versions. Optional.
+  # SECURITY: this allows for a sandbox escape
+  # TODO: don't ship this by default. Like, really.
+  - --device=all
+  # Controller support, though this only works on newer Flatpak versions. Optional.
+  # SECURITY: shouldn't have much of an impact, although the user should disable this from Flatseal if controller support is unwanted
+  # We don't ship this by default for backwards compat :( See above.
+  # - --device=input
+  # DRI device. Required for graphics acceleration. Non-optional, since Roblox is a 3D game.
+  - --device=dri
```
hfiguiere, line 17:
```suggestion
```

this is included with `all`.

org.vinegarhq.Sober.yml
```diff
@@ -0,0 +1,71 @@
...
+  # Required for game audio & microphone access when using the optional voice chat feature. Optional, but heavily recommended.
+  # SECURITY: allows unprompted microphone access. This is a limitation of the sandbox for now.
+  - --socket=pulseaudio
+  # Required to share what Roblox experience you're playing to your Discord servers & contacts. Optional
+  # SECURITY: Sober sees your Discord username and Discord ID due to how Discord IPC works, but it shouldn't be allowed to make actions on your behalf. Also, we obviously don't send that stuff to our server.
+  - --filesystem=xdg-run/app/com.discordapp.Discord:create
+  - --filesystem=xdg-run/discord-ipc-0
+  # Required for a workaround. Has no security or UX impact, since Sober is sadly English-only for now
+  # TODO: fix the bug and remove this
+  - --env=LC_ALL=en_US.UTF-8
```
hfiguiere, line 41: why? Seriously. Fix your software.

hfiguiere: bionic isn't glibc....

en_US isn't English. It's American English. You assume the user has the en_US locale installed. This assumption is incorrect.

bbhtt:
bot, build org.vinegarhq.Sober

bbhtt:
> There's no emulators, virtual machines, or even really translation layers involved here.

So how is it done?

PR #6349, opened 2025-04-06
Add io.github.ilya_zlobintsev.LACT

ilya-zlobintsev:

- [X] Please describe the application briefly. LACT is a configuration and monitoring tool for AMD, Nvidia and Intel GPUs.

LACT is a GPU control and monitoring tool. It consists of 2 parts:
- GUI: the UI is a standard GTK4 application. There isn't anything special regarding Flatpak packaging for the UI part.
- Daemon: to be able to write GPU settings, LACT needs a system daemon that runs as root. As this cannot be done from inside the Flatpak sandbox, the package comes with a helper script that sets up the daemon on the host if it is not detected. It is also possible to reject the daemon setup, in which case it will run inside of the UI process without root permissions. This mode can be useful for people who wish to use LACT only as an information/monitoring tool.

Communication between the GUI and daemon is performed over a unix socket at `/run/lactd.sock`.

While the daemon requirement falls under the `Submissions incompatible or restricted by Flatpak sandbox` policy, I am the upstream developer and want to support this kind of use case. The host service setup has been designed in a way that it will work on any system as long as it has systemd, regardless of host libraries.
Another way would be to publish only the GUI part as a Flatpak package, but IMO it would be misleading to users as they would also need to install a system package for the daemon in order to access most of the functionality.

More information about the Flatpak service setup can be found [here](https://github.com/ilya-zlobintsev/LACT/tree/master/flatpak).

I will also submit a lint exception PR for the `org.freedesktop.Flatpak` permission shortly.

### DIFF
--- a/io.github.ilya_zlobintsev.LACT.yaml
+++ b/io.github.ilya_zlobintsev.LACT.yaml
@@ -1,0 +1,126 @@
+app-id: io.github.ilya_zlobintsev.LACT
+runtime: org.gnome.Platform
+runtime-version: "48"
+sdk: org.gnome.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+  - org.freedesktop.Sdk.Extension.llvm20
+command: startup.sh
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --talk-name=org.freedesktop.Flatpak
+  - --filesystem=/run/lactd.sock
+cleanup:
+  - /include
+  - /share/vulkan/registry
+  - /lib/*.a
+  - /lib/pkgconfig
+modules:
+  - name: lact
+    buildsystem: simple 
+    sources:
+      - type: archive
+        url: https://github.com/ilya-zlobintsev/LACT/archive/refs/tags/v0.7.3.tar.gz
+        sha256: 212ac1eadcbc8d7112b7ee11176c6ab0aee04f53bda9cfb342679c2e1cf6a230
+        x-checker-data:
+          type: anitya
+          project-id: 372202
+          url-template: https://github.com/ilya-zlobintsev/LACT/archive/refs/tags/v$version.tar.gz
+      - generated-sources.json
+    build-options:
+      append-path: "/usr/lib/sdk/rust-stable/bin:/usr/lib/sdk/llvm20/bin"
+      prepend-ld-library-path: "/usr/lib/sdk/llvm20/lib"
+      env:
+        CARGO_HOME: "/run/build/lact/cargo"
+        PREFIX: "/app"
+    build-commands:
+      - cargo --offline fetch --manifest-path Cargo.toml
+      - cargo --offline build -p lact --features flatpak --release
+      - make install
+      - install -Dm755 flatpak/startup.sh /app/bin/startup.sh
+      - sed -i -E 's/lact gui/startup.sh/' /app/share/applications/io.github.ilya_zlobintsev.LACT.desktop
+    modules:
+      - name: yad
+        config-opts:
+          - --enable-standalone
+          - --disable-icon-browser
+          - --disable-tools
+        sources:
+          - type: archive
+            url: https://github.com/v1cont/yad/releases/download/v14.1/yad-14.1.tar.xz
+            sha256: dde047a915cd8d3892c32b6ba031876f5cda673e01882c99613f043867c88133
+            x-checker-data:
+              type: anitya
+              project-id: 5280
+              url-template: https://github.com/v1cont/yad/releases/download/v$version/yad-$version.tar.xz
+        cleanup:
+          - /share/icons
+        modules:
+          - name: intltool
+            cleanp:
+              - '*'
+            sources:
+              - type: archive
+                url: https://launchpad.net/intltool/trunk/0.51.0/+download/intltool-0.51.0.tar.gz
+                sha256: 67c74d94196b153b774ab9f89b2fa6c6ba79352407037c8c14d5aeb334e959cd
+                x-checker-data:
+                  type: anitya
+                  project-id: 1385
+                  url-template: https://launchpad.net/intltool/trunk/$version/+download/intltool-$version.tar.gz
+      - name: libdrm
+        buildsystem: meson
+        builddir: true
+        config-opts:
+          - -Dtests=false
+        sources:
+          - type: archive
+            url: https://gitlab.freedesktop.org/mesa/drm/-/archive/libdrm-2.4.124/drm-libdrm-2.4.124.tar.gz
+            sha256: 49c077f3938147e7c321fe89255eb189c1be68f6ed0aa36e29d38e3db0e84e08
+            x-checker-data:
+              type: anitya
+              project-id: 1596
+              url-template: https://gitlab.freedesktop.org/mesa/drm/-/archive/libdrm-$version/drm-libdrm-$version.tar.gz
+
+  - name: vulkan-tools
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DVULKAN_HEADERS_INSTALL_DIR=/app
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DBUILD_CUBE=OFF
+      - -DBUILD_ICD=OFF
+      - -DBUILD_VULKANINFO=ON
+    sources:
+      - type: archive
+        url: https://github.com/KhronosGroup/Vulkan-Tools/archive/v1.4.311/Vulkan-Tools-1.4.311.tar.gz
+        sha256: 7113bc0c746b45072e269fada0d684b4ae9de609c38d9e764b59793b14120a7b
+        x-checker-data:
+          type: anitya
+          project-id: 242111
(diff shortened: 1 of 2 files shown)


io.github.ilya_zlobintsev.LACT.yaml
```diff
@@ -0,0 +1,126 @@
...
+            url: https://github.com/v1cont/yad/releases/download/v14.1/yad-14.1.tar.xz
+            sha256: dde047a915cd8d3892c32b6ba031876f5cda673e01882c99613f043867c88133
+            x-checker-data:
+              type: anitya
+              project-id: 5280
+              url-template: https://github.com/v1cont/yad/releases/download/v$version/yad-$version.tar.xz
+        cleanup:
+          - /share/icons
+        modules:
+          - name: intltool
```
hfiguiere, line 62: please use intltool from shared-modules.

io.github.ilya_zlobintsev.LACT.yaml
```diff
@@ -0,0 +1,126 @@
...
+              - '*'
+            sources:
+              - type: archive
+                url: https://launchpad.net/intltool/trunk/0.51.0/+download/intltool-0.51.0.tar.gz
+                sha256: 67c74d94196b153b774ab9f89b2fa6c6ba79352407037c8c14d5aeb334e959cd
+                x-checker-data:
+                  type: anitya
+                  project-id: 1385
+                  url-template: https://launchpad.net/intltool/trunk/$version/+download/intltool-$version.tar.gz
+      - name: libdrm
```
hfiguiere, line 73: this is in the runtime already.

io.github.ilya_zlobintsev.LACT.yaml
```diff
@@ -0,0 +1,126 @@
...
+      prepend-ld-library-path: "/usr/lib/sdk/llvm20/lib"
+      env:
+        CARGO_HOME: "/run/build/lact/cargo"
+        PREFIX: "/app"
+    build-commands:
+      - cargo --offline fetch --manifest-path Cargo.toml
+      - cargo --offline build -p lact --features flatpak --release
+      - make install
+      - install -Dm755 flatpak/startup.sh /app/bin/startup.sh
+      - sed -i -E 's/lact gui/startup.sh/' /app/share/applications/io.github.ilya_zlobintsev.LACT.desktop
```
hfiguiere, line 44: use `desktop-file-edit`, not sed.

io.github.ilya_zlobintsev.LACT.yaml
ilya-zlobintsev: While `libdrm.so.2` is part of the base runtime, vendor-specific libraries such as `libdrm_amdgpu` and `libdrm_intel` are only provided via the `org.freedesktop.Platform.GL` extension.

The host service startup script works by running the daemon binary using libraries from the Flatpak runtime, the path for which is fetched from `/.flatpak-info` in the sandbox. The same info file does not directly provide paths for runtime extensions, so they are not loaded by the host daemon.

Including libdrm directly in the application directory ensures that the libraries are present in a known location, and also unbinds the library version from what's present in the runtime extension.

io.github.ilya_zlobintsev.LACT.yaml
ilya-zlobintsev: Done.

io.github.ilya_zlobintsev.LACT.yaml
ilya-zlobintsev: Done.

io.github.ilya_zlobintsev.LACT.yaml
```diff
@@ -0,0 +1,126 @@
...
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --talk-name=org.freedesktop.Flatpak
+  - --filesystem=/run/lactd.sock
+cleanup:
+  - /include
+  - /share/vulkan/registry
+  - /lib/*.a
```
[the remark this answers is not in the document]
ilya-zlobintsev: Done.

io.github.ilya_zlobintsev.LACT.yaml
bbhtt: Mesa in the runtime will badly break if you override runtime's libdrm. Don't do it.

ilya-zlobintsev: In that case I have a followup question: are runtime extensions guaranteed to be installed in the same location as the runtime itself? For example, I have the following data present in `/.flatpak-info`:
```
runtime-path=/var/lib/flatpak/runtime/org.gnome.Platform/x86_64/48/10943396e989593a17d9639a13927a190a3c6d292dacd0905ac3c3ec3371f2fe/files
runtime-commit=10943396e989593a17d9639a13927a190a3c6d292dacd0905ac3c3ec3371f2fe
runtime-extensions=org.gnome.Platform.Locale=f9452c93245819e8c2c240b24469f7d65f0d8d6bdafe5cd3253bf51aa9841494;org.freedesktop.Platform.GL.default=9f5bebaec7e2f20bd1b32e18c2629e876a88b2286e011835b5cf46d0680d1754;org.freedesktop.Platform.GL.nvidia-570-133-07=76a0755f4d05c576c52242ffdf381d2e9f37d0a92af4a408684d4133b3ee39ab;org.freedesktop.Platform.VulkanLayer.MangoHud=d3538035d8ca6168e2c828830247b0cef6ef7da47202075af7d47536aa8298b3;org.gtk.Gtk3theme.Adwaita-dark=7c055bbc0667524f34ca715abd67473480de93a176797a8fe4b7f7a71c4e3324;org.gtk.Gtk3theme.Breeze=931f423749ebe31620ce9eec710cc018ee1234fbf4b7c4a71973c9950b3ec1f3;org.gtk.Gtk3theme.adw-gtk3-dark=90431d86bb0470b72abc073b5c441d42fe8eea7fc352c73442c9c77b5e971582;org.freedesktop.Platform.openh264=0f52621e4540863ee86b1fe26216fff78fefa1096f367079692344139228e474;org.freedesktop.Platform.GL.default=de5d0e5d82d627f44f390875179c0117d369de727be152db92e75ccb6267bd20
```
If the runtime is at `/var/lib/flatpak/runtime/org.gnome.Platform/x86_64/48/10943396e989593a17d9639a13927a190a3c6d292dacd0905ac3c3ec3371f2fe/files`, does that mean that `org.freedesktop.Platform.GL.default` is at `/var/lib/flatpak/runtime/org.freedesktop.Platform.GL.default/`? 

Or could it be installed at a user level while the runtime is in a system path? (and if that's the case, can I somehow get the full extension path)

bbhtt: > In that case I have a followup question: are runtime extensions guaranteed to be installed in the same location as the runtime itself?

Yes for GL extensions, it should be by default

ilya-zlobintsev: Alternatively, I could look into linking libdrm statically, as done in [com.dec05eba.gpu_screen_recorder](https://github.com/flathub/com.dec05eba.gpu_screen_recorder/blob/master/com.dec05eba.gpu_screen_recorder.yml#L86)

ilya-zlobintsev: Would it be an acceptable solution if a custom libdrm was included in the sandbox, but installed at a custom location? This way it won't override the default libraries provided by GL extensions, and will only be explicitly loaded by the daemon on the host when requested.

ilya-zlobintsev: I've updated libdrm to be installed at a custom path, now it won't override the runtime-provided one and will only be used by the host daemon.

There is a source patch to accompany this that will be removed with the next application update.

bbhtt: Also I'm confused why the libdrm drivers from the mesa extension cannot be used. They are present in a known location inside the sandbox and outside.

ilya-zlobintsev: > Do you need only the libdrm drivers or the core libdrm library too?

Both are needed.

> They are present in a known location inside the sandbox and outside. As I said, there has to be a GL extension in runtime-path= base, unless someone manually removed it.

It turns out `/.flatpak-info` is also missing the extension version needed to get the full host path, e.g. it doesn't say that `org.gnome.Platform/x86_64/48` uses `org.freedesktop.Platform.GL.default/x86_64/24.08` or `23.08`. While this information is likely possible to query on the host, this would significantly complicate the launch script as the host path generator currently runs inside of the sandbox, where this information is missing. Installing the libraries in a known application path is a much simpler solution by comparison.

bbhtt: The version is in the metadata.

bbhtt:
bot, build io.github.ilya_zlobintsev.LACT

io.github.ilya_zlobintsev.LACT.yaml
ilya-zlobintsev: I've updated the startup script to read the extension version and use that for libdrm. The path discovery logic is more complicated than I'd like it to be, but I suppose it's not really intended for people to be using flatpak runtime libraries outside of flatpak.

In the future libdrm will be changed to be linked statically so the GL extension won't be needed for the daemon.

bbhtt:
bot, build io.github.ilya_zlobintsev.LACT

PR #6363, opened 2025-04-11
Add org.catacombing.kumo

chrisduerr:

- [X] Please describe the application briefly. A Wayland Mobile Browser
- [X] I am an author ~~/developer/upstream contributor~~ to the project. ~~If not, I contacted upstream developers about this submission. **Link:**~~


org.catacombing.kumo.json
```diff
@@ -0,0 +1,120 @@
...
+    "runtime-version": "24.08",
+    "sdk": "org.freedesktop.Sdk",
+    "sdk-extensions": [
+        "org.freedesktop.Sdk.Extension.rust-stable",
+        "org.freedesktop.Sdk.Extension.llvm20"
+    ],
+    "command": "kumo",
+    "finish-args": [
+        "--socket=pulseaudio",
+        "--socket=wayland",
```
chrisduerr, line 13: Not providing X11 support seems to cause a warning, however my application **intentionally** does not support X11 to simplify development. It was mentioned that some lints can be ignored, but I wanted to make clear that this was not omitted by accident.

org.catacombing.kumo.metainfo.xml
```diff
@@ -0,0 +1,56 @@
...
+            <image>https://raw.githubusercontent.com/catacombing/kumo/refs/heads/master/screenshots/tab_groups.png</image>
+            <caption>Tabs organized in a tab group</caption>
+        </screenshot>
+    </screenshots>
+    <releases>
+        <release version="0.0.0" date="2025-04-11">
+            <description>
+                <p>Preview release of Kumo</p>
+            </description>
+        </release>
```
chrisduerr, line 46: I've seen pre-releases made for other Flathub packages, so I'd assume this should be fine?

Technically no version has been "officially" released yet, since building WebKit is obviously an undertaking that is too much for more users. So I'd rather get an early "test" version out to get some more testers and refine things more before marking an official "stable" release.

That said the browser is functional and my sole browser on my phone right now. So despite the lack of a release, I do consider this app ready to be used by a broader audience.

org.catacombing.kumo.metainfo.xml
```diff
@@ -0,0 +1,56 @@
...
+    </releases>
+    <recommends>
+        <control>touch</control>
+        <control>keyboard</control>
+        <control>pointing</control>
+    </recommends>
```
chrisduerr, line 52: I'm not sure if the order of these matters. While keyboard and pointing are technically supported and should work just fine, touch is definitely the primary target audience.

From what I've seen essentially anything that isn't a hard requirement should just go into "recommends" and there isn't really such a thing as tiered platforms (just support or no support).

org.catacombing.kumo.metainfo.xml
hfiguiere, line 1: this file should be (submitted and managed) upstream

org.catacombing.kumo.json
```diff
@@ -0,0 +1,120 @@
...
+        }]
+    },{
+        "name": "kumo",
+        "buildsystem": "simple",
+        "build-commands": [
+            "cargo --offline fetch",
+            "cargo --offline build --release",
+            "install -Dm 755 ./target/release/kumo /app/bin/kumo",
+            "install -Dm 644 ./logo.svg /app/share/icons/hicolor/scalable/apps/org.catacombing.kumo.svg",
+            "sed -i 's/Name=Kumo/Name=org.catacombing.kumo/' Kumo.desktop",
```
hfiguiere, line 102: don't use sed. use `desktop-file-edit`.

org.catacombing.kumo.json
```diff
@@ -0,0 +1,120 @@
...
+            "-DENABLE_SPEECH_SYNTHESIS=OFF",
+            "-DENABLE_DOCUMENTATION=OFF",
+            "-DENABLE_MINIBROWSER=OFF"
+        ],
+        "build-options": {
+            "arch": {
+                "arm": {
+                    "cflags": "-g1",
+                    "cxxflags": "-g1"
+                },
```
hfiguiere, line 82:
```suggestion
```

org.catacombing.kumo.json
bbhtt: you need to submit an exception then https://docs.flathub.org/docs/for-app-authors/linter#exceptions

https://docs.flathub.org/docs/for-app-authors/linter#finish-args-only-wayland

org.catacombing.kumo.json
```diff
@@ -0,0 +1,112 @@
...
+            "-DENABLE_MINIBROWSER=OFF"
+        ],
+        "build-options": {
+            "arch": {
+                "aarch64": {
+                    "cflags": "-g1",
+                    "cxxflags": "-g1"
+                }
+            }
+        },
```
bbhtt, line 84: what are these for?

org.catacombing.kumo.json
bbhtt, line 1: can you format the json properly? https://0x0.st/8pTC.json

org.catacombing.kumo.json
chrisduerr: This produces minimal debug info, it's copied from the official wpewebkit flatpak that they ship on their own and was previously present in the archlinux pkgbuild before switching to clang.

It's probably fine to remove it if desired, but chances are this will make building this app lighter with no major downsides for most users. We're talking about ~4GB of debug data here according to Archlinux maintainers.

org.catacombing.kumo.json
chrisduerr: I was not aware there was a formatting standard in use by this repository.

org.catacombing.kumo.json
chrisduerr: For reference: https://github.com/Igalia/wpe-flatpak/blob/master/sdk/org.wpe.Sdk.json#L218-L229

This is probably not mandatory, but I don't think it will hurt.

hfiguiere: the default flags are different. you shouldn't need these. We also generate debug symbols.

bbhtt: default is -g which probably ends up generating gigantic debuginfo.

bbhtt:
bot, build org.catacombing.kumo

org.catacombing.kumo.json
chrisduerr: > default is -g which probably ends up generating gigantic debuginfo.

I believe this would be the result, yes. I can obviously remove this, but as the developer I don't really gain much from the additional symbols so it just seems wasteful and can actually run into build failures on some systems due to the amount of symbols generated.

bbhtt: why is this on aarch64 only though?

chrisduerr:
I apologize for that unnecessary failure, I've had this change made in my test environment but committed it to the wrong repo. Considering it was the last step and it matches my test environment, everything should hopefully work now.

org.catacombing.kumo.json
chrisduerr: That's a good point. Arch had it on x86 obviously, I'm not sure about flathub's builders but my aarch64 machine also struggles a bit more than my x86 one. So it's just for the platform that is most likely to fail due to symbol sizes. Since this was copied from the official wpe webkit repo, I'd assume their runners were struggling too.

I just added it for all platforms. I personally don't need it on either and helping the x86 runners a bit probably doesn't hurt.

bbhtt:
bot, build org.catacombing.kumo

bbhtt:
Why is the window not draggable on GNOME? and also missing any window bar.

chrisduerr:
> Why is the window not draggable on GNOME?

I've never used GNOME Wayland, I don't really optimize my software for it.

> missing any window bar.

Window decorations are just a waste of space on mobile device screen sizes. Putting in the effort to add decorations just to support desktop GNOME where this will never be used seems pointless to me.

Is supporting desktop GNOME well a requirement for flatpaks on flathub? It should certainly work, but I'll never be testing GNOME because I'm not really interested in the platform.

bbhtt:
Which platform does this support then?

bbhtt:
I can't get it to run on Plasma wayland either. It immediately crashes there.


```
2025-04-12T22:03:07.602187Z  INFO kumo: 84: Started Kumo
libEGL warning: egl: failed to create dri2 screen
MESA: error: ZINK: failed to choose pdev
libEGL warning: egl: failed to create dri2 screen

thread 'main' panicked at src/engine/webkit/mod.rs:321:64:
get render node
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

org.catacombing.kumo.json
```diff
@@ -0,0 +1,126 @@
...
+        {
+            "name": "kumo",
+            "buildsystem": "simple",
+            "build-commands": [
+                "cargo --offline fetch",
+                "cargo --offline build --release",
+                "install -Dm 755 ./target/release/kumo /app/bin/kumo",
+                "install -Dm 644 ./logo.svg /app/share/icons/hicolor/scalable/apps/org.catacombing.kumo.svg",
+                "desktop-file-edit --set-name=org.catacombing.kumo Kumo.desktop",
+                "desktop-file-edit --set-icon=org.catacombing.kumo Kumo.desktop",
```
bbhtt, line 112: ID as name

```
[Desktop Entry]
Type=Application
Name=org.catacombing.kumo
```

org.catacombing.kumo.json
```diff
@@ -0,0 +1,126 @@
...
+                "install -Dm 644 ./Kumo.desktop /app/share/applications/org.catacombing.kumo.desktop",
+                "install -Dm 644 ./org.catacombing.kumo.metainfo.xml -t /app/share/metainfo/"
```
bbhtt, line 114: if this is mobile only it should be marked in metainfo as such

image

https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines#mobile-only-apps

chrisduerr: > if this is mobile only it should be marked in metainfo as such

It technically supports non-mobile environments. It just works *better* on mobile and is optimized for it. I'm not sure that qualifies as "only on mobile"?

org.catacombing.kumo.json
chrisduerr: Right, seems like I confused myself there.

chrisduerr:
> Which platform does this support then?

I personally use my own compositor. But I do occasionally test it on Niri or Sway (though there is a bug in Sway right now I believe). I don't have any of the other mobile DEs available, which is one of the reasons I'd like to make it easier for other people to help with testing.

bbhtt:
it is crashing in one DE and impossible to test in another. I don't have time to set up DIY WMs to test this app.

org.catacombing.kumo.json
bbhtt: > It technically supports non-mobile environments. 

the question is whether you do / intended platform and if it is usable.

chrisduerr:
> it is crashing in one DE and impossible to test in another. I don't have time to set up DIY WMs to test this app.

Well as a Wayland compositor and client developer I would consider GNOME broken by design. Plasma not working is not intentional, but it's a platform I cannot test regularly because none of my devices run it.

The Plasma crash looks like a Mesa issue, so I'd be curious if that's the same device as the GNOME one? OpenGL ES 2+ support is required, so not sure if that was maybe some virtual machine.

I'd be willing to look into Plasma and try and get that working on one of my devices, but that's a lot of effort for no direct benefit to me. So before doing that I'd at least like to get confirmation that working on Plasma+Sway+Niri would be sufficient to get this into flathub? Or is GNOME and CSDs **mandatory**? Because I don't really want to waste my time on GNOME.

hfiguiere:
> I personally use my own compositor. But I do occasionally test it on Niri or Sway (though there is a bug in Sway right now I believe). I don't have any of the other mobile DEs available, which is one of the reasons I'd like to make it easier for other people to help with testing.

Flatpak is meant to provide applications independently from the host distribution. While they are sometime requirement beyond that, running a specific compositor surely is not one of them. Flathub is meant to distribute application for users, not experiments for the sake of it.

Maybe this is a sign that this is no ready for flathub.

For the record I have no idea what Niri is, and Sway always looked to be "for those who don't want a DE". And it crashes on the common ones. Yes GNOME Shell do exist for mobile.

chrisduerr:
> And it crashes on the common ones.

That's just not true though. I just tested it on Plasma mobile and it works just fine. I don't see how Sway being for people who don't want a DE makes it any less valid of a compositor. And from what I've heard it doesn't crash on GNOME either, it just doesn't support CSDs which is a GNOME bug they refuse to give way on.

So really this app works on multiple desktop+mobile compositors without major problems. It just doesn't work on GNOME due to choices made by GNOME (not this app). If GNOME is the only relevant compositor for publishing on flathub, it might make sense to document that in the requirements.

bbhtt:
> So really this app works on multiple desktop+mobile compositors without major problems. It just doesn't work on GNOME due to choices made by GNOME (not this app). 

So why does it crash on Plasma desktop wayland?


> If GNOME is the only relevant compositor for publishing on flathub, it might make sense to document that in the requirements.

It's not

bbhtt:
If it's really meant for mobile only and you are repeatedly mentioning mobile platforms and testing on them you should really mark this as mobile only in the metadata 

There's no point in saying it supports desktop when you don't test on them.

chrisduerr:
> There's no point in saying it supports desktop when you don't test on them.

That's the thing though. I test on Sway, which is desktop for small changes and for bigger things I run it on my phone to test.

I don't expect a lot of desktop users to be interested in this browser, so if you're saying that we can get this merged if I just declare it as mobile-only for now then I'm fine with that. But it's not actually accurate. If desktop users can still install mobile apps and it's just a notice on the flathub website, then that's probably fine.

bbhtt:
Apps on Flathub are supposed to work cross desktop. It's crashing in one and unusable in another (it's popping up on a random corner of the screen and I can't drag it or quit it)

chrisduerr:
> It's crashing in one and unusable in another

I mean GNOME is just unusable to me in general. But I can't really judge what you're experiencing since I don't have a GNOME desktop system to reproduce. Your plasma environment is spitting out Mesa warnings/errors talking about Zink (OpenGL shim for Vulkan-only devices), so that doesn't sound like a "normal" machine to me.

I've tested Niri (desktop), Sway (desktop), Phosh (mobile), Plasma Mobile, Catacomb (mobile), SXMO (Sway mobile) and they all worked just fine for me. I'm not quite sure where you draw the line at "work cross desktop", but it clearly works on various different platforms as far as I can tell.

bbhtt:
I don't, you have to share a screen record of how it works on some mobile DEs eg. Phosh and plasma mobile.

Then make it clear in the metainfo that this is mobile only (add the required metadata) and add a note in the description.

chrisduerr:
> I don't, you have to share a screen record using this Flatpak to show how it works on some mobile DEs eg. Phosh and plasma mobile.

Fair enough, let's just go with that then for the sake of actually being productive.

This is Kumo on phosh:

https://github.com/user-attachments/assets/4595522d-6bba-4bfc-bbd5-4b0b09e09932

If you want a recording that isn't painfully slow, I can also record it on my compositor. Note that it's much faster without recording even on posh, it's just not a very fast DE when recording the screen. It still shows up as "org.catacombing.kumo" in the app list since this was installed from the builder's flatpak output, that would be fixed with the latest version I'd imagine.

> Then make it clear in the metainfo that this is mobile only (add the required metadata) and add a note in the description.

The description and all project documentation already refer to it as a web browser focused on portrait-mode touchscreen mobile devices. I'd hope that's sufficient.

bbhtt:
bot, build org.catacombing.kumo

bbhtt:
This needs to wait until this works for me... this is the third DE I tried and it immediately crashes again.

This is Fedora 41 Sway inside GNOME Boxes VM.


https://github.com/user-attachments/assets/6fa41f5c-630f-4fe3-92b2-1f89024713e2

bbhtt:
There's obviously something going wrong which you should investigate.

chrisduerr:
@bbhtt You did enable 3d acceleration on the VM, right? As far as I know it has to be manually toggled on gnome boxes. I've never tried it with VirGL, but it should work in theory.

bbhtt:
Yes it's enabled.

chrisduerr:
Thanks for the quick reply. I think the issue is just that you have no render node (`/dev/dri/renderD128`) but only a drm device node (`/dev/dri/card0`). While technically WebKit expects a render node like `renderD128` I've confirmed that it technically does work with just `/dev/dri/card0`, which should make it work inside your VM if hardware acceleration is enabled. I'll have to double-check this with WebKit though to make sure there aren't some unexpected consequences, so I'll let you know when things should be working inside VMs.

chrisduerr:
The latest version should be capable of running without a render node path. I haven't tested it myself in a VM, but I'd appreciate if you could build it through @flathubbot and I'd be happy to test it myself first. I'd just like to avoid building it twice especially because it takes significantly longer on my personal hardware.

bbhtt:
You can start the builds 

bot, build org.catacombing.kumo

bbhtt:
Now it launches but crashes again when I try to go to a webpage


https://github.com/user-attachments/assets/3189c811-a950-4350-bd89-2d9baece560d

chrisduerr:
bot, build org.catacombing.kumo

chrisduerr:
Actually there's a very good chance that "app ID" isn't actually talking about the Wayland app Id but some glib thing instead. So there's a good chance this patch doesn't work.

chrisduerr:
Latest patch is somewhat of a hail mary that at least has a chance of working compared to the Wayland app ID one. If that doesn't do it I'll need to find a computer I can install Fedora 41 on.

bot, build org.catacombing.kumo

chrisduerr:
It appears that editing a message causes the bot to queue another build, I was not aware of this.

bbhtt:
Still seems to crash similarly.

chrisduerr:
Thanks for testing, sounds like I need to find a system with a matching flatpak version so I can actually test the changes myself. Ideally a newer version that actually prints the mismatched name.

chrisduerr:
While it took me forever to build on my old slow laptop, I have now confirmed that the latest version runs on Fedora 41 GNOME without any issues. I also did not have any issues dragging the application around by holding super.

chrisduerr:
> flatpak install --user https://dl.flathub.org/build-repo/177297/flathub.flatpakref

It seems like there's a bug in the new runner setup. This works correctly when changed to `org.catacombing.kumo.flatpakref` and I've double-checked this also works properly on Fedora Workstation 41 + GNOME.

PR #6426, opened 2025-04-25
Add app.ryujinx.Ryujinx

fpiesche:

- [x] Please describe the application briefly: This is a copy of the [`flathub/io.github.ryubing.Ryujinx`](https://github.com/flathub/io.github.ryubing.Ryujinx) package, updated to the most recent release version of the application and with an updated app ID. The github repo no longer exists and the source is now self-hosted. I've also added x-checker-data to ensure that update PRs are only created for new stable releases rather than on every commit to the upstream repo.

Please add @tancop to this repository as the original creator of the io.github.ryubing.Ryujinx package; from the situation over there I'm assuming he hasn't recently had the time to keep the repo up to date, hence my contribution here.


app.ryujinx.Ryujinx.metainfo.xml
hfiguiere, line 1: by now this should be (submitted and managed) upstream.

screenshots/01.jpg
hfiguiere, line 1: why are there screenshots here? They are not referenced anwywhere.

(question applies to all 3)

app.ryujinx.Ryujinx.metainfo.xml
fpiesche: There is no open submission process for upstream changes, annoyingly. I'll see what I can do, as including a metainfo.xml is also included for AppImages (which is the official upstream approach to Linux distribution).

hfiguiere: > There is no open submission process for upstream changes, annoyingly. I'll see what I can do, as including a metainfo.xml is also included for AppImages (which is the official upstream approach to Linux distribution).

then why isn't that one used?

hfiguiere:
Is a renaming really necessary?

bbhtt:
>  The github repo no longer exists and the source is now self-hosted. 

Is upstream going to verify it? If not then currently the new ID doesn't make sense.

screenshots/01.jpg
fpiesche: They are used by the metainfo; the `cdn.jsdelivr.net` URLs pull the files at the specified GitHub URL into a CDN specifically for open-source projects (as using GitHub as a CDN is not good practice). Bafflingly, upstream does not seem to host any official screenshots on their website or git repos.

app.ryujinx.Ryujinx.metainfo.xml
fpiesche: Sorry, I was unclear - there isn't an upstream metainfo.xml. I'll see if I can get access to their repo to submit this one, along with some other changes (their icon and desktop launchers don't use the app ID, for example).

screenshots/01.jpg
hfiguiere: so really no point renaming if upstream is unwilling to help making the submission official,.

fpiesche:
> > The github repo no longer exists and the source is now self-hosted.
> 
> Is upstream going to verify it? If not then currently the new ID doesn't make sense.

Upstream are not officially supporting Flatpak (as per the note in the metainfo on the current package) in favour of AppImage. The main reason I'm suggesting the rename is that the github repo referenced by the app ID no longer exists, so the app ID is no longer valid and could theoretically be hijacked (though at least for now the ryubing org still exists here).

fpiesche:
> > The github repo no longer exists and the source is now self-hosted.
> 
> Is upstream going to verify it? If not then currently the new ID doesn't make sense.
> 
> > The github repo no longer exists
> 
> Doesn't matter for existing apps, the org still exists

In which case ignore me; I'll submit the update to the original package only. Thanks for clarifying!

PR #6472, opened 2025-05-10
Add chat.delta.DeltaChatTauri

Simon-Laux:

- [x] Please describe the application briefly: deltachat desktop, but based on tauri instead of electron. deltachat is a messenger app.

- nicodh
- WofWca
- muelli
- missytake


Simon-Laux:
You may be asking yourself: why there are so many file in `generated/proxy-registry-cache-indices`?

We use pnpm.
To my knowledge it is not possible to build a pnpm cache with flathub-builder manifests, because it is not simply a folder with all packages, it is a content addressable store, that extracts all files of all packages and stores them by their hash in a de-duplicated manner.

So I came up with the workaround to record the pnpm installation (save all packages it downloads) and serve a proxy that fakes online installation for pnpm while it is installing inside of the flatpak build process.

The packages are written to the manifest at `generated/proxy-registry-cache-manifest.json`, but the indices for all packages are also required and they can not be fetched via manifest, because their content and (thus hash) changes all the time. So we download them to `generated/proxy-registry-cache-indices` and strip unnecessary information from them, so they only take 5mb instead of the >290mb they would take normally.

I'm open for alternative suggestions that are better than this workaround. (also for use in https://github.com/flathub/chat.delta.desktop, which faces the same challenge of using pnpm)

chat.delta.desktop.tauri.yml
```diff
@@ -0,0 +1,136 @@
+app-id: chat.delta.desktop.tauri
```
hfiguiere, line 1: this is not an acceptable app-id. "Tauri" is not the application name. (Nor "desktop", the `chat.delta.desktop` would be rejected today, for the record)

chat.delta.desktop.tauri.yml
```diff
@@ -0,0 +1,136 @@
...
+        CARGO_HOME: /run/build/deltachat-tauri/cargo
+
+        # The intention is to have a release build with debugging symbols.
+        # The flatpak build process should separate the debug symbols.
+        RUSTFLAGS: "-g"
+      append-path: /usr/lib/sdk/rust-stable/bin:/usr/lib/sdk/node20/bin:/run/build/deltachat-tauri/pnpm/bin
+    build-commands:
+      - pwd
+      - ls
+      - env
```
hfiguiere, line 56:
```suggestion
```

chat.delta.desktop.tauri.yml
```diff
@@ -0,0 +1,136 @@
...
+      - |
+        jq ".pnpm.supportedArchitectures.os = [\"linux\"] | .pnpm.supportedArchitectures.cpu = [\"x64\",\"arm64\"] | del(.pnpm.onlyBuiltDependencies[] | select(. == \"electron\"))" package.json > package.new.json
+        mv package.new.json package.json
+
+      # install deps from store
+      - |
+        node tool_replay.mjs &
+        REPLAY_PID=$!
+        env DEBUG="*" pnpm install --prefer-offline --frozen-lockfile
+        kill $REPLAY_PID
```
hfiguiere, line 77: why kill?  what kind of hack is this?

chat.delta.desktop.tauri.yml
```diff
@@ -0,0 +1,136 @@
...
+      # for some reason this is needed as soon as we enable the flatpak feature
+      - mv cargo/config cargo/config.toml
+      - sed -i 's/\\\\"/\"/g' cargo/vendor/tauri-plugin-single-instance/Cargo.toml
+      - sed -i 's/\\\\"/\"/g' cargo/vendor/tao/Cargo.toml
+
+      # build & package
+      - |
+        export CARGO_HOME=/run/build/deltachat-tauri/cargo
+        pnpm --filter @deltachat-desktop/target-tauri tauri build -b deb -- --offline --features=flatpak
+      - ls target/release/bundle
```
hfiguiere, line 89:
```suggestion
```

chat.delta.desktop.tauri.yml
```diff
@@ -0,0 +1,136 @@
...
+      - mkdir deb_data
+      - tar -xvzf data.tar.gz -C deb_data
+      - cp -r deb_data/usr/* /app
+
+      # icons
+      # - install -Dm 0644 deltachat-tauri.svg /app/share/icons/hicolor/scalable/apps/deltachat-tauri.svg
+      - install -Dm 0644 chat.delta.desktop.tauri.appdata.xml /app/share/metainfo/chat.delta.desktop.tauri.metainfo.xml
+      - install -Dm 0644 deltachat-tauri.svg /app/share/icons/hicolor/scalable/apps/chat.delta.desktop.tauri.svg
+
+      # rename files
```
hfiguiere, line 103: use  the `rename-icon` property in the manifest.

chat.delta.desktop.tauri.yml
```diff
@@ -0,0 +1,136 @@
...
+      - mv /app/share/icons/hicolor/256x256@2/apps/deltachat-tauri.png /app/share/icons/hicolor/256x256@2/apps/chat.delta.desktop.tauri.png
+      - mv /app/share/icons/hicolor/128x128/apps/deltachat-tauri.png /app/share/icons/hicolor/128x128/apps/chat.delta.desktop.tauri.png
+      - mv /app/share/icons/hicolor/32x32/apps/deltachat-tauri.png /app/share/icons/hicolor/32x32/apps/chat.delta.desktop.tauri.png
+      - mv /app/share/applications/deltachat-tauri.desktop /app/share/applications/chat.delta.desktop.tauri.desktop
+
+      # modify icon path in desktop file
+      - sed -i 's/Icon=deltachat-tauri/Icon=chat.delta.desktop.tauri/; s/Name=deltachat-tauri/Name=Delta Chat Tauri/' /app/share/applications/chat.delta.desktop.tauri.desktop
```
hfiguiere, line 110: don't use `sed` to change desktop files. Use the command `desktop-file-edit`

Also the icon would be fixed automatically when using `rename-icon` as mentionned above.

chat.delta.desktop.tauri.yml
```diff
@@ -0,0 +1,136 @@
...
+      - echo "StartupWMClass=deltachat-tauri" >> /app/share/applications/chat.delta.desktop.tauri.desktop
```
hfiguiere, line 111: again use `desktop-file-edit`.

chat.delta.desktop.tauri.yml
```diff
@@ -0,0 +1,136 @@
...
+      # icons
+      # - install -Dm 0644 deltachat-tauri.svg /app/share/icons/hicolor/scalable/apps/deltachat-tauri.svg
+      - install -Dm 0644 chat.delta.desktop.tauri.appdata.xml /app/share/metainfo/chat.delta.desktop.tauri.metainfo.xml
+      - install -Dm 0644 deltachat-tauri.svg /app/share/icons/hicolor/scalable/apps/chat.delta.desktop.tauri.svg
+
+      # rename files
+      - mv /app/share/icons/hicolor/256x256@2/apps/deltachat-tauri.png /app/share/icons/hicolor/256x256@2/apps/chat.delta.desktop.tauri.png
+      - mv /app/share/icons/hicolor/128x128/apps/deltachat-tauri.png /app/share/icons/hicolor/128x128/apps/chat.delta.desktop.tauri.png
+      - mv /app/share/icons/hicolor/32x32/apps/deltachat-tauri.png /app/share/icons/hicolor/32x32/apps/chat.delta.desktop.tauri.png
+      - mv /app/share/applications/deltachat-tauri.desktop /app/share/applications/chat.delta.desktop.tauri.desktop
```
hfiguiere, line 107: use `rename-desktop-file`

chat.delta.desktop.tauri.yml
```diff
@@ -0,0 +1,136 @@
...
+      - generated/desktop-git.json
+
+      # dependencies
+      - generated/proxy-registry-cache-manifest.json
+      - generated/sources-rust.json
+      - generated/pnpm.json
+
+      # assets
+      - type: file
+        path: chat.delta.desktop.tauri.256.png
```
hfiguiere, line 124: why is this needed?

chat.delta.desktop.tauri.yml
```diff
@@ -0,0 +1,136 @@
...
+        dest-filename: chat.delta.desktop.tauri.png
+      - type: file
+        path: deltachat-tauri.svg
+        dest-filename: deltachat-tauri.svg
```
hfiguiere, line 128: you don't need a SVG.

deltachat-tauri.48.svg
hfiguiere, line 1: never referenced.

flake.nix
hfiguiere, line 1: what is this for?

hfiguiere (changes requested):
why is there a million files in `generated`.

Simon-Laux:
> why is there a million files in generated.

easier for git on changes than one huge file, but we could also change it to put all that into a single file

flake.nix
Simon-Laux: so that you can easily get the dependencies for generating this with nix (for example on nixos).
This makes it possible that I can just run `nix develop` and then run `./generate.sh` to update the package without having to deal with dependency managment

chat.delta.desktop.tauri.yml
```diff
@@ -0,0 +1,136 @@
...
+        node tool_replay.mjs &
+        REPLAY_PID=$!
+        env DEBUG="*" pnpm install --prefer-offline --frozen-lockfile
+        kill $REPLAY_PID
+
+      # patching some cargo files
+      # for some reason this is needed as soon as we enable the flatpak feature
+      - mv cargo/config cargo/config.toml
+      - sed -i 's/\\\\"/\"/g' cargo/vendor/tauri-plugin-single-instance/Cargo.toml
+      - sed -i 's/\\\\"/\"/g' cargo/vendor/tao/Cargo.toml
```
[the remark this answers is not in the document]
Simon-Laux: what should I use instead?

chat.delta.desktop.tauri.yml
Simon-Laux: would `chat.delta.DeltaChatTauri` be good?

chat.delta.desktop.tauri.yml
Simon-Laux: the replay npm proxy server is not needed anymore so we end it, it does only reading, so killing it is fine.

chat.delta.desktop.tauri.yml
Simon-Laux: > Also the icon would be fixed automatically when using rename-icon as mentionned above.

so rename icon also edits the desktop file?

chat.delta.desktop.tauri.appdata.xml
hfiguiere, line 1: this file should be (submitted and managed) upstream.

chat.delta.desktop.tauri.appdata.xml
```diff
@@ -0,0 +1,120 @@
...
+  <screenshots>
+    <screenshot type="default" xml:lang="en">
+      <image>
+        https://raw.githubusercontent.com/deltachat/interface/main/screenshots/2024/desktop/linux/1.46_light_mode_beach_bg.png
+      </image>
+      <caption>Main View showcasing multiple accounts, chats and a group conversation</caption>
+    </screenshot>
+    <screenshot xml:lang="en">
+      <image>
+        https://raw.githubusercontent.com/deltachat/interface/main/screenshots/2024/desktop/linux/1.46_darkmode_beach_bg.png
```
hfiguiere, line 34: screenshot should never reference the tip of a branch as it's not immutable.

chat.delta.desktop.tauri.yml
Simon-Laux: does it hurt to have one? or is it never used anyways?

chat.delta.desktop.tauri.yml
Simon-Laux: good catch, it is not needed, I'll remove it

chat.delta.desktop.tauri.yml
hfiguiere: in general the name of the toolkit in the application name is bad taste. But that's your problem.

chat.delta.desktop.tauri.yml
```diff
@@ -0,0 +1,132 @@
...
+  - --talk-name=org.freedesktop.StatusNotifier.Watcher
+  - --talk-name=com.canonical.Unity.LauncherEntry
+  # do we need this?
+  - --talk-name=com.canonical.AppMenu.Registrar
+  - --share=network
+  - --env=WEBKIT_DISABLE_COMPOSITING_MODE=1 # without this it crashes on wayland with nvidia gpu - atleast on nixos
+
+cleanup:
+  # not used yet
+  # -
```
hfiguiere, line 29:
```suggestion
```

I'm pretty this cause a warning when running flatpak builder.

chat.delta.desktop.tauri.yml
hfiguiere: why does it need a proxy? Can't it find package in a set location that would be downloaded by flatpak-builder ? This looks like over engineered for no benefit.

chat.delta.desktop.tauri.yml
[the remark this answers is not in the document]
hfiguiere: 1. a patch
2. maybe fixing the cargo.toml? I have a hard time imagining what these sed thing does to the Cargo.toml.

flathub.json
hfiguiere, line 1: remove this it's not needed.

chat.delta.desktop.tauri.yml
```diff
@@ -0,0 +1,132 @@
...
+      - mv /app/share/icons/hicolor/32x32/apps/deltachat-tauri.png /app/share/icons/hicolor/32x32/apps/chat.delta.desktop.tauri.png
+      - mv /app/share/applications/deltachat-tauri.desktop /app/share/applications/chat.delta.desktop.tauri.desktop
+
+      # modify icon path in desktop file
+      - sed -i 's/Icon=deltachat-tauri/Icon=chat.delta.desktop.tauri/; s/Name=deltachat-tauri/Name=Delta Chat Tauri/' /app/share/applications/chat.delta.desktop.tauri.desktop
+      - echo "StartupWMClass=deltachat-tauri" >> /app/share/applications/chat.delta.desktop.tauri.desktop
+
+    sources:
+      # Delta Chat
+      - generated/desktop-git.json
```
hfiguiere, line 111: no the main source should be inline here. It's unmaintainable otherwise.

flake.nix
hfiguiere: so you are saying that for one to maintain this package, one has to have a specific distribution? That's really not helpful.

hfiguiere:
> > why is there a million files in generated.
> 
> easier for git on changes than one huge file, but we could also change it to put all that into a single file. For why this data is there at all, see my comment above: [#6472 (comment)](https://github.com/flathub/flathub/pull/6472#issuecomment-2869183043)

It's autogenerated. I don't see why there should be trillions of files. It overcomplicated already overcomplicated things.

chat.delta.desktop.tauri.yml
hfiguiere: if you really to have one, have it upstream. Otherwise you don't need it, not here.

chat.delta.desktop.tauri.yml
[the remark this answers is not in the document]
Simon-Laux: 2. would require fixing it upstream, which will take a long time until it lands here.

basically this is a fix for over escaping. for some reason it only complains when I give it the "flatpak" rust feature (which has special logic for flatpak, like using the background portal for autostart), before that I didn't need this fix... so that it did not complain before is weird.

flake.nix
Simon-Laux: no I'm saying that I'm on a special distro and having that file makes my life easier, it is optional.

chat.delta.desktop.tauri.yml
Simon-Laux: that would make the `generate.sh` script much more complex and you need to call that anyway to update the dependencies. or am I missing something here? why do you say it's "unmaintainable"? what is the usecase of having it inline? is there some manual process or tooling that relies on that?

bbhtt:
Host your pnpm cache elsewhere in a git repo, use that git repo in the manifest as a source, not here.

chat.delta.desktop.tauri.yml
Simon-Laux: it can not, unless pnpm has a hidden feature to import from existing npm cache that I don't know of.

flake.nix
Simon-Laux: if you are not on nixos then you need to install manually what generate.sh needs:

https://github.com/flathub/flathub/blob/f05f984ab9b029a5a6818e6305061404459c1ab6/generate.sh#L9-L13

Simon-Laux:
Now the stripped indices are inside of a single manifest file. the size is now only ~1.6 mb.

chat.delta.DeltaChatTauri.yml
```diff
@@ -0,0 +1,111 @@
...
+finish-args:
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=all # Webcam access for QR code scans and DRI for OpenGL
+  - --socket=pulseaudio # Record and play voice messages
+  # we use this for the moment, because the portal api also had some problems
+  # (https://github.com/deltachat/deltachat-desktop/pull/4929)
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=org.kde.StatusNotifierItem
```
bbhtt, line 19: is both needed? only `org.kde.StatusNotifierWatcher` should be needed nowadays.

chat.delta.DeltaChatTauri.yml
```diff
@@ -0,0 +1,111 @@
...
+  - --talk-name=org.freedesktop.StatusNotifier.Watcher
```
bbhtt, line 20: what implements this? afair none

tool_process.mjs
bbhtt, line 1: what are these `tool_*.mjs` for?

Simon-Laux: they are for the pnpm offline installation workaround, think of them like a flatpak-builder-tools/node, just for pnpm

chat.delta.desktop.tauri.yml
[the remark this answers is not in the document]
Simon-Laux: looks like this issue is present on upstream, so possibly a quirk/bug in cargo or the flatpak builder with git dependencies. I think it should go away once my changes were released and we can use upstream directly for those crates again.

Simon-Laux:
Thank you both very much for your very helpful feedback :heart: 

I made a pr to the electron edition flatpak to incorporate some of the points you mentioned there too: https://github.com/flathub/chat.delta.desktop/pull/169

chat.delta.desktop.tauri.yml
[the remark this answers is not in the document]
Simon-Laux: without these sed statements, we get this error:
```
error: failed to load source for dependency `tao`

Caused by:
  Unable to update https://github.com/Simon-Laux/tao?branch=hunt-down-open-url-crash-and-add-application_continue_user_activity#32049e49

Caused by:
  failed to update replaced source https://github.com/Simon-Laux/tao?branch=hunt-down-open-url-crash-and-add-application_continue_user_activity#32049e49

Caused by:
  failed to parse manifest at `/run/build/deltachat-tauri/cargo/vendor/tauri-plugin-single-instance/Cargo.toml`

Caused by:
  failed to parse `target_os = \"linux\"` as a cfg expression: unexpected character `\` in cfg, expected parens, a comma, an identifier, or a string
failed to build app: failed to build app
       Error failed to build app: failed to build app
/run/build/deltachat-tauri/packages/target-tauri:
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @deltachat-desktop/target-tauri@1.57.0 tauri: `tauri "build" "-b" "deb" "--" "--offline" "--features=flatpak"`
Exit status 1
Error: module deltachat-tauri: Child process exited with code 1
```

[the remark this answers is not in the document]
hfiguiere: As I mentionned in https://github.com/flathub/flathub/pull/6472#discussion_r2083339755

Use a patch. Not sed.

generated/desktop-git.json
```diff
@@ -0,0 +1,8 @@
+[
+    {
+        "type": "git",
+        "url": "https://github.com/deltachat/deltachat-desktop.git",
+        "commit": "e4c59ea7449944243926b052459f3f98b3a58967",
+        "dest": "."
```
bbhtt, line 6: why is this not in the main manifest?

bbhtt: also this is the same as https://github.com/flathub/chat.delta.desktop/blob/c89724971b5739de51e83bf12f53f905089362e1/generated/desktop-git.json#L4

why does two use the same source? What's the difference?

bbhtt: It's unsustainable to accept the same app with each new toolkit of the month

Simon-Laux: > is the only difference tauri vs electron?

- The backend is rewritten in rust instead of electron/nodejs javascript
- the chatmail core lib is directly compiled into the app as static library instead of being an extra executable
- tauri version supports autostart via background portal
- it does not need extra folder permissions since it uses the portal api
- webxdc mini apps start faster
- it uses a different folders for logs and userdata
- it uses webkitgtk instead of chromium
- it is smaller because it uses the webview of webkitgtk (which is part of the flatpak gnome platform), instead of shipping it's own browser and nodejs runtime (that's what electron does)
- and many more technical details

> It's unsustainable to accept the same app multiple times with each new toolkit/framework of the month

Of course, that wouldn't make sense. Though this can be thought of like a fork with a different name. Only difference to a classic fork is that it lives inside of the same repo under a different folder.

It will likely replace the electron version one day when it has proved itself in production for some time.

bbhtt: If you change the underlying toolkit there will be some technical differences, but is it essentially the same app?

bbhtt:
bot, build chat.delta.DeltaChatTauri

generated/desktop-git.json
Simon-Laux: > why is this not in the main manifest?

because it is generated by `generate.sh`together with the manifests for the dependencies and it would be more complex (and possibly error prone) to write directly to the main manifest.

I can change the generation script to write directly to the main manifest, but unless there is a good reason for why this should be in the main manifest, I'd rather not introduce that extra complexity.

@hfiguiere has also asked this above. There I also asked for reasons for having it in the main manifest, but got no answer yet: https://github.com/flathub/flathub/pull/6472#discussion_r2083366062

bbhtt: I can build and run it but if the UX, user side functionality is the same it is unlikely to be accepted.

This sets a poor precedent if we accept and allows everyone to submit Discord (Electron), Discord (tauri), Discord (insert random toolkit), which are essentially duplicates from the user side. The average user doesn't have to care about the underlying tech

chat.delta.DeltaChatTauri.yml
```diff
@@ -0,0 +1,113 @@
...
+runtime: org.gnome.Platform
+runtime-version: "48"
+sdk: org.gnome.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node20
+  - org.freedesktop.Sdk.Extension.rust-stable
+command: deltachat-tauri
+finish-args:
+  - --socket=fallback-x11
+  - --socket=wayland
```
bbhtt, line 12:
```suggestion
finish-args:
  - --share=ipc
  - --socket=fallback-x11
  - --socket=wayland
```

bbhtt:
bot, build chat.delta.DeltaChatTauri

generated/desktop-git.json
Simon-Laux: > If the app is distinct among different versions of your app on Flathub (e.g. Firefox Beta versus the standard Firefox), then using the full name including the edition may be acceptable.
> ~ https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/quality-guidelines#just-the-name

the front-end is the same. reading this line in the docs made me think there may be a chance for it to be accepted.

Otherwise what would be the minimum differences to be accepted into flathub?
- Other name? the name of the project of moving it to tauri is "DeltaTauri" (https://nlnet.nl/project/DeltaTauri/)
- Logo that is fundamentally different
- different ui themes

bbhtt: Name and icon doesn't matter in this case, you can always add `Name (insert toolkit)` to create new ones.

Different UI themes is a bit to minimal differences.

Are there any functionality and performance differences that are visible to the user? Are you planning to drop the Electron version and transition to this? (so for example tauri version will get new features etc. not the electron one)

bbhtt: If there is a transition plan and if this is more or less the same can the tauri version be offered through beta branch of https://github.com/flathub/chat.delta.desktop ?

Simon-Laux: > Are there any functionality and performance differences that are visible to the user?

- for webxdc mini apps yes, the startup on electron has a progress bar, on tauri it opens instantly.
- there are quite some people that dislike electron apps, so not electron is a feature for some.
- startup is twice as fast for tauri than electron edition on my machine.
- Less wayland bugs than with electron (which has issues with resizing windows on gnome, and either has a mini cursor or a blurry ui with fractional scaling)
- less permissions needed, which will is user visible on the flathub store page.

> Are you planning to drop the Electron version and transition to this? (so for example tauri version will get new features etc. not the electron one)

yes that is the idea. the tauri edition will be optimized to use portals, on electron we can not easily use native dependencies, so these improvements will only come to the tauri edition. Basically the tauri edition flatpak is in active development and the electron flatpak gets less of our attention.

So dc electron flatpak is current "stable" (with known bugs) and dc tauri flatpak is the upcoming replacement, which works fine for me, but it needs to proof itself with real users before we can make the full switch with all existing users to it.

bbhtt: Any reason to not use the beta branch?

bbhtt:
bot, build chat.delta.DeltaChatTauri

generated/desktop-git.json
Simon-Laux: > Any reason to not use the beta branch?

it uses different folder paths and there are no migrations, so using the beta branch is not appropriate as the data is not shared.

Ideally would be something like flathub or new flathub app that only has a beta branch if that is possible.

I'm changing the description to distinguish it from the electron edition:
````xml
<p>Delta Chat Tauri has the following advantages over the classic Delta Chat Desktop:</p>
    <ul>
        <li>Autostarts with your system (can be turned off)</li>
        <li>Faster and less resource hungry</li>
        <li>Smaller download</li>
        <li>Less permissions required on flatpak</li>
        <li>Based on Tauri instead of Electron</li>
    </ul>
```

bbhtt:
You can start the build if you fix it.

Simon-Laux:
bot, build chat.delta.DeltaChatTauri

Simon-Laux:
works locally, but not in CI :sob: 

anyway will look into this again later

Simon-Laux:
bot, build chat.delta.DeltaChatTauri

generated/desktop-git.json
hfiguiere: > it uses different folder paths and there are no migrations, so using the beta branch is not appropriate as the data is not shared.

I don't see how this is a problem for putting it on beta.

But definitely this should be "flathub-beta" instead.

nojisan:
bot, build chat.delta.DeltaChatTauri

Simon-Laux:
bot, build chat.delta.DeltaChatTauri

Simon-Laux:
bot, build chat.delta.DeltaChatTauri

Simon-Laux:
bot, build chat.delta.DeltaChatTauri

Simon-Laux:
bot, build chat.delta.DeltaChatTauri

Simon-Laux:
bot, build chat.delta.DeltaChatTauri

Simon-Laux:
rustc gets killed. maybe a problem of not enough RAM?

PR #6505, opened 2025-05-17
Add io.github.gameflorist.dunedynasty

codeflorist:
### Please confirm your submission meets all the criteria

- [X] Please describe the application briefly. Remaster/Enhancement of the classic real-time strategy game Dune II by Westwood Studios. Requires the original game data files to work.
- [X] The domain used for the application ID is [controlled by the application developer(s)][appid-domain] and the [application id guidelines][appid] are followed.
- [X] I have read and followed all the [Submission requirements][reqs] and the [Submission guide][reqs2] and I agree to them.
- [X] I am the maintainer of the fork at https://github.com/gameflorist/dunedynasty. The original Dune Dynasty hosted on [SourceForge](https://dunedynasty.sourceforge.net/) seems abandoned since 10 years and i was not able to contact the original developer David Wang.

[appid-domain]: https://docs.flathub.org/docs/for-app-authors/requirements/#control-over-domain-or-repository
[appid]: https://docs.flathub.org/docs/for-app-authors/requirements#application-id
[reqs]: https://docs.flathub.org/docs/for-app-authors/requirements
[reqs2]: https://docs.flathub.org/docs/for-app-authors/submission

PR #6536, opened 2025-05-24
Add io.github.Mrmayman.QuantumLauncher

Mrmayman:

- [x] Please describe the application briefly:

QuantumLauncher is a lightweight, fast and powerful Minecraft launcher which supports features like a built in (free) content store, fully independent Minecraft installations, and fixes/patches for old Minecraft versions.

- [x] Please include a screenrecord (10-20s) showcasing the application.

[Quantumlauncher showcase.webm](https://github.com/user-attachments/assets/bb04f855-21e0-404c-af8f-2d3bbc15015f)

# Note
- metainfo, icon, .desktop file are included in repo and are installed through `build-commands`. Please let me know if this is the wrong approach
- the latest commit name was a mistake, i meant "add captions to screenshots" but typed something else


io.github.Mrmayman.QuantumLauncher.yml
```diff
@@ -0,0 +1,54 @@
...
+      - type: archive
+        url: https://www.x.org/releases/individual/app/xrandr-1.5.3.tar.gz
+        sha256: 980befa14b48de2fa74dc07bbb22f746acfd12e1f9dbd539eab9a3c9b434d667
+  - name: quantum-launcher
+    buildsystem: simple
+    sources:
+      - type: git
+        url: https://github.com/Mrmayman/quantum-launcher
+        tag: v0.4.1
+        commit: acaa0e6eb0c58182b8184059dcbadb27b4ec85db
```
hfiguiere, line 33: tag and commit don't match.

io.github.Mrmayman.QuantumLauncher.yml
```diff
@@ -0,0 +1,54 @@
...
+  # Needs to talk to the network:
+  - --share=network
+  # Wayland access
+  - --socket=wayland
+  # X11 + XShm access
+  - --share=ipc
+  - --socket=fallback-x11
+  # Controller and other devices
+  - --device=all
+  - --device=shm
```
hfiguiere, line 18: why do you need shm?

io.github.Mrmayman.QuantumLauncher.yml
```diff
@@ -0,0 +1,54 @@
...
+          tag-pattern: ^v([\d.]+)$
+      - type: file
+        path: assets/freedesktop/io.github.Mrmayman.QuantumLauncher.desktop
+      - type: file
+        path: assets/freedesktop/quantum-launcher.metainfo.xml
+      - flathub-sources.json
+    build-commands:
+      - cargo --offline fetch --manifest-path Cargo.toml --verbose
+      - cargo --offline build --profile release-ql --verbose
+      - strip --strip-all target/release-ql/quantum_launcher
```
hfiguiere, line 45: why do you strip the binaries>

io.github.Mrmayman.QuantumLauncher.yml
Mrmayman: shm was added because @ApicalShark thought it was needed for x11.

It has been removed in the latest commit

io.github.Mrmayman.QuantumLauncher.yml
Mrmayman: it was originally for binary size (added quite a while ago) but now with the latest build profile it doesn't make a difference (didn't check earlier)

```sh
$ cp quantum_launcher quantum_launcher.bak 
$ strip --strip-all quantum_launcher
$ du quantum_launcher quantum_launcher.bak 
20156	quantum_launcher
20156	quantum_launcher.bak
```

it has been removed in the latest commit

io.github.Mrmayman.QuantumLauncher.yml
Mrmayman: fixed in latest commit, thanks for pointing it out

Mrmayman:
there are a few other issues that need fixing, will get back to you guys when done

io.github.Mrmayman.QuantumLauncher.yml
```diff
@@ -0,0 +1,54 @@
...
+        url: https://github.com/Mrmayman/quantum-launcher
+        tag: v0.4.1
+        commit: acaa0e6eb0c58182b8184059dcbadb27b4ec85db
+        x-checker-data:
+          type: git
+          tag-pattern: ^v([\d.]+)$
+      - type: file
+        path: assets/freedesktop/io.github.Mrmayman.QuantumLauncher.desktop
+      - type: file
+        path: assets/freedesktop/quantum-launcher.metainfo.xml
```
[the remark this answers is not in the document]
Mrmayman: now the latest commit might have fixed it (it works on my machine)

io.github.Mrmayman.QuantumLauncher.yml
```diff
@@ -0,0 +1,52 @@
...
+  - org.freedesktop.Sdk.Extension.rust-stable
+command: quantum-launcher
+finish-args:
+  # Needs to talk to the network:
+  - --share=network
+  # Wayland access
+  - --socket=wayland
+  # X11 + XShm access, fallback-x11 wont work in wayland
+  - --share=ipc
+  - --socket=x11
```
bbhtt, line 15: it's either wayland+fallback-x11 or only x11

hfiguiere: then make it `x11` only.

io.github.Mrmayman.QuantumLauncher.yml
```diff
@@ -0,0 +1,52 @@
...
+  # Controller and input devices
+  - --device=all
+  # GL access
+  - --device=dri
```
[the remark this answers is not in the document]
Mrmayman: Fixed in latest commit

bbhtt:
bot, build

Mrmayman:
I think I have fixed the crash. I was able to reproduce the same issue, after this fix it doesn't happen for me. Can you take a look now?

I fixed it by unsetting `WAYLAND_DISPLAY`. It's not a great fix, but I think it should work.

> your app is totally broken

I'm sorry for the weird bugs that are appearing here, I'm using the `winit` rust library for window management (and i'm using it in the completely default configuration aside from the flatpak sandbox stuff), `winit` is just super unreliable in general. Maybe these issues might also be caused by flatpak sandbox, in which case do let me know if something breaks. I have now tested it on my machine with both X11 and wayland, it works in both cases

Anyway I think this specific issue has been fixed, let me know if the fix isn't ideal, or if there's something else to be fixed

io.github.Mrmayman.QuantumLauncher.yml
```diff
@@ -0,0 +1,49 @@
...
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+command: quantum-launcher
+finish-args:
+  # Needs to talk to the network:
+  - --share=network
+  # X11 + XShm access, fallback-x11 wont work in wayland
+  - --share=ipc
+  - --socket=x11
+  - --env=WAYLAND_DISPLAY=
```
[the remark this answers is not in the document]
Mrmayman: this fix was based on a workaround documented in the Servo project
 https://book.servo.org/hacking/building-and-running-on-wsl.html

While that documentation is for WSL, the same error message appeared in this context, and the fix worked identically

I think that even though the Wayland socket was not exposed to the sandbox, the presence of the WAYLAND_DISPLAY variable probably misled winit (used internally by the iced GUI library) into trying to use Wayland (and failing)
I've seen similar issues with this library in the past

I don't know how the variable was visible inside the sandbox, but it clearly was, as removing (or running flatpak override) fixed the issue on my system

Unfortunately I’m using winit in its *default configuration* as part of iced, and I cannot trivially change or patch it, and switching GUI libraries isn't feasible.

All the wayland/x11 logic is done by the library and I have no control (aside from changing flatpak manifest to expose/hide any permissions)

If you know of any better way to fix this please do tell me, I'm a beginner at flatpak packaging, and this fix is the only one I am aware of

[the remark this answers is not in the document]
bbhtt: this is still an issue in winit, checking for the env var to select the backend is not enough, you have to check if the wayland socket is actually present. And then it should never crash.

hfiguiere, line 14: This should be `--unset-env=WAYLAND_DISPLAY` instead

Mrmayman: fixed in latest commit

Mrmayman:
in the latest commit i have also bundled metainfo and desktop files into this repo instead of leaving them in the app repo, because these have a few fixes/modifications not present in the v0.4.1 release commit's files

bbhtt:
Then remove the desktop and metainfo files from here

quantum-launcher.metainfo.xml
```diff
@@ -0,0 +1,97 @@
...
+        <control>touch</control>
+    </recommends>
+
+    <content_rating type="oars-1.1" />
+    <launchable type="desktop-id">quantum-launcher.desktop</launchable>
+
+    <screenshots>
+        <screenshot type="default">
+            <image
+            >https://raw.githubusercontent.com/Mrmayman/quantum-launcher/main/assets/screenshots/window.png</image>
```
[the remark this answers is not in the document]
Mrmayman: fixed in latest commit

Mrmayman:
> Then remove the desktop and metainfo files from here

Just to clarify, in the v0.4.1 tag commit (9d763e4b5358eb13e1f137fe8ed2beb546021c45), the .desktop and metainfo xml files in the app repository are missing a few last minute fixes:

- The <launchable> tag is broken.
- The screenshots are missing captions.

I fixed these issues only right after the release tag was created, but they're not present in the tagged commit that Flatpak builds from

Because of this, I included the corrected .desktop and metainfo.xml files directly in this PR instead of the app repository. This ensures that up to date metadata is present without requiring a new app release just for metadata

That said, I noticed other Flatpak PRs (e.g. #6505) have also included files in the PR instead of repo

Should I still remove the metainfo and desktop files from this PR and using the outdated versions in the repo, or is including them here acceptable in this case?

bbhtt:
Yes use the upstream ones in your repo and remove it from here. You can create a new tag or use a commit directly in the manifest

Mrmayman:
okay, i created a new tag which contains the updated desktop and metainfo in the app repo. I've removed the files from PR now

bbhtt:
bot, build

io.github.Mrmayman.QuantumLauncher.yml
```diff
@@ -0,0 +1,54 @@
...
+        commit: 715bb2a78ff2088f13516e0a413855ec15d42dff
+        x-checker-data:
+          type: git
+          tag-pattern: ^v([\d.]+)(?:-([a-zA-Z0-9]+))?$
+      - flathub-sources.json
+    build-commands:
+      - cargo --offline fetch --manifest-path Cargo.toml --verbose
+      - cargo --offline build --profile release-ql --verbose
+      - install -Dm755 target/release-ql/quantum_launcher /app/bin/quantum-launcher
+      - install -Dm644 assets/freedesktop/quantum-launcher.desktop /app/share/applications/io.github.Mrmayman.QuantumLauncher.desktop
```
hfiguiere: the problem here is because it's improperly named upstream the appstream file isn't updated. (that's the build failure). One easy way is to install the with the original name and use `rename-destkop-file` and `rename-appdata-file` in the manifest.

Or upstream use the proper names from the get go.

Mrmayman: I think it's fixed now, can you take a look?

bbhtt:
bot, build

bbhtt:
How does this work without a Mojang/Microsoft account? Minecraft requires a license.

bbhtt:
So this is using unlicensed copies? The this is blocked by the last point in https://docs.flathub.org/docs/for-app-authors/requirements#inclusion-policy and https://docs.flathub.org/docs/for-app-authors/requirements#license

https://github.com/Mrmayman/quantumlauncher/blob/17b49d8b0a84e8056939a6d0e1e6e7e1df6fbf4f/README.md?plain=1#L230-L232

> If you pirate the game, it's at your own risk. I am not responsible for any issues caused.
I recommend that you buy the game, but if you don't have the means, feel free to use this launcher.

Mrmayman:
For the full experience you will need to login with a paid account. The offline mode is limited and comes with many drawbacks such as mostly-non functional multiplayer, no skin support, no minecraft realms, some mods being blocked, etc. It's only intended for limited play

bbhtt:
But it does seem to allow single player and is against Mojang/Microsoft's ToS.

PR #6572, opened 2025-05-31
Add net.charlesschaefer.addictiontracker

charlesschaefer:

> Addiction Tracker - An app for tracking substance addiction and habits (made with Tauri)

https://github.com/user-attachments/assets/69f87912-5ae1-4b0b-9fa9-127e62394e23

- [x] The Flatpak ID follows all the rules listed in the [Application ID requirements][appid]. (Yes, net.charlesschaefer.addictiontracker).


flathub.json
bbhtt, line 1: drop this

Addiction Tracker.desktop
bbhtt, line 1: did you see the note in https://docs.flathub.org/docs/for-app-authors/requirements#required-metadata ?

libayatana-appindicator-gtk3.json.diff
bbhtt, line 1: this patch is not needed, it wasn't installing for you locally to /app/lib as you were using an older flatpak-builder. Use 1.4.4

net.charlesschaefer.addictiontracker.yml
```diff
@@ -0,0 +1,95 @@
+id: net.charlesschaefer.addictiontracker
+runtime: org.gnome.Platform
+runtime-version: '48'
+sdk: org.gnome.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node22
+  - org.freedesktop.Sdk.Extension.rust-stable
+command: addiction-tracker
+finish-args:
+  - --env=LD_LIBRARY_PATH=/app/lib
```
bbhtt, line 10:
```suggestion
```

net.charlesschaefer.addictiontracker.yml
```diff
@@ -0,0 +1,95 @@
...
+  - --env=FLATPAK_ENV=1
```
bbhtt, line 11: what is this for?

net.charlesschaefer.addictiontracker.yml
```diff
@@ -0,0 +1,95 @@
...
+  - --device=dri
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc
+  - --share=network # Permission needed to access the network
```
bbhtt, line 16:
```suggestion
  - --share=network
```

this is self explanatory

net.charlesschaefer.addictiontracker.yml
```diff
@@ -0,0 +1,95 @@
...
+        # Set the npm cache directory, used by npm to find its package metadata.
+        npm_config_cache: /run/build/addiction-tracker/flatpak-node/npm-cache
+        # Sets the directory where Node is located so that npm won't download the headers.
+        npm_config_nodedir: /usr/lib/sdk/node22
+    build-commands:
+      - npm --prefix=addiction-tracker install --offline &&
+        npm --prefix=addiction-tracker run tauri build -- --bundles deb --config src-tauri/flatpak-tauri.conf.json &&
+        VERSION=$(npm --prefix=addiction-tracker pkg get version | sed "s/\"//g") &&
+        cp "addiction-tracker/src-tauri/target/release/bundle/deb/Addiction Tracker_${VERSION}_amd64.deb" addictiontracker.deb &&
+        ls -al
```
bbhtt, line 69:
```suggestion
        cp "addiction-tracker/src-tauri/target/release/bundle/deb/Addiction Tracker_${VERSION}_amd64.deb" addictiontracker.deb
```

net.charlesschaefer.addictiontracker.yml
```diff
@@ -0,0 +1,95 @@
...
+      - npm --prefix=addiction-tracker install --offline &&
+        npm --prefix=addiction-tracker run tauri build -- --bundles deb --config src-tauri/flatpak-tauri.conf.json &&
+        VERSION=$(npm --prefix=addiction-tracker pkg get version | sed "s/\"//g") &&
+        cp "addiction-tracker/src-tauri/target/release/bundle/deb/Addiction Tracker_${VERSION}_amd64.deb" addictiontracker.deb &&
+        ls -al 
+      - ar -x addictiontracker.deb
+      - tar -xf data.tar.gz
+      - install -Dm755 usr/bin/addiction-tracker /app/bin/addiction-tracker
+      - install -Dm644 usr/share/applications/Addiction\ Tracker.desktop /app/share/applications/net.charlesschaefer.addictiontracker.desktop
+      - sed -i -e "s|Icon=addiction-tracker|Icon=net.charlesschaefer.addictiontracker|g" /app/share/applications/net.charlesschaefer.addictiontracker.desktop
```
bbhtt, line 74: use desktop-file-edit

net.charlesschaefer.addictiontracker.yml
```diff
@@ -0,0 +1,95 @@
...
+        url: "https://github.com/charlesschaefer/addiction-tracker.git"
+        tag: "app-v0.0.16"
+        #commit: "089aec3f53b47b10472041bc00edfbcd56563284"
+        dest: addiction-tracker
+      # - type: dir
+      #   path: ../
+      #   dest: addiction-tracker
+      - cargo-sources.json
+      - npm-sources.0.json
+      #- type: file
```
bbhtt, line 94:
```suggestion
        dest: addiction-tracker
      - cargo-sources.json
      - npm-sources.0.json
```

net.charlesschaefer.addictiontracker.yml
```diff
@@ -0,0 +1,95 @@
+id: net.charlesschaefer.addictiontracker
```
bbhtt, line 1: link your github account somewhere in https://charlesschaefer.net/contact.html so that I know you control the domain

bbhtt:
all metadata files need to be upstream

Addiction Tracker.desktop
charlesschaefer: Ouch, not really. I've included the desktop file just in case (I get it from the .deb bundling process that is executed during build). I did this because I saw in that page that it is required, but didn't notice it and other metadata can be obtained from upstream. Sometimes I struggle to understand what should be kept on the app repo (upstream) and on the flatpak repo.

I'll read all your comments, make the changes and review the files and build process again. Thanks.

libayatana-appindicator-gtk3.json.diff
charlesschaefer: From my trials it is needed indeed. When we use the runtime org.gnome.Platform//46 the libraries inside the shared modules repo will be placed in the container's "/app/lib". But using the version 47 and 48, the libayatana-appindicator's internal module called "ayatana-ido" starts placing the library files inside "/app/lib64". But "libayatana-indicator" module always tries to link it from the "/app/lib" dir. 

The easier way I found to stop having build errors where libayatana-indicator looks for ayatana-ido files in /app/lib while they are in /app/lib64 was to patch this file.

But if there is another way to go, I can change the approach.

net.charlesschaefer.addictiontracker.yml
charlesschaefer: To be honest I don't really know, I think I copied this part from the file generated by gnome builder for rust applications. But I don't think it is having any effects on the app. I'll trim it out.

net.charlesschaefer.addictiontracker.yml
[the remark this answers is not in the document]
charlesschaefer: The app is totally offline most of the time. But it allows the user to look for other instances of the app in the local network to synchronize data between devices.

net.charlesschaefer.addictiontracker.yml
```diff
@@ -0,0 +1,95 @@
...
+        dest: shared-modules
+        commit: a34a31e4dd50a0b38d1725f6881d1bedb0948b38
+      - type: patch
+        path: libayatana-appindicator-gtk3.json.diff
+        options:
+          - "-p1"
+          # - use-git: true
+    modules:
+      # - shared-modules/libayatana-appindicator/libayatana-appindicator-gtk3.json
+      - shared-modules/libappindicator/libappindicator-gtk3-12.10.json
```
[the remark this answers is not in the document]
charlesschaefer: Sorry, I don't understand why "obviously this wasn't build locally". You ask to send a video of the app working after being built locally, what I did. What is your claim? That I built it outside, and I'm trying to push the manifest just for sport?

To be honest, I've never spent so long to package an app like I had with flatpak. The easier approach to me would be to just bring the built .deb and repackage it as flatpak. But also read your advices to prefer building directly from the sources and thought this could be something vital to be approved (since I'm also the maintainer of the upstream).

I saw about the submodule on the docs. It was not clear to me if it should go on the upstream repo or flatpak repo. I read somewhere in the docs a warning to not pollute the flatpak repo with submodules. And I tested it in the upstream and didn't work. 

I couldn't find in the docs anything saying exactly where to put the submodule, and maybe it is something very natural for other people more experienced with Linux bundling options - not my case. So it was not very clear to me.

Finally, I saw somewhere (I'll research my history to check from where I got that) that the flatpak build servers would make the repo available automatically in every build when the repo is published.

So, in my tests I tried to keep the repo in the same folder as the manifest and, voilà, it finally worked. So it made sense to me the idea that the repo could be made available automatically by the servers.

All that said, this source part is something that I forgot to remove after finishing the local build. It really doesn't make sense here.

Sorry about that, there are also some comments and other craps here and there. I'll clean everything before submitting again.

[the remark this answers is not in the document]
hfiguiere: I don't see how patching a manifest like will have any effect. It's a first.

Also I don't see how `shared-modules/libappindicator/libappindicator-gtk3-12.10.json` is found. Maybe it works on your machine by pure coincidence, it surely won't work anywhere on the builders. Or one another developer system.

libayatana-appindicator-gtk3.json.diff
hfiguiere: > you were using an older flatpak-builder. Use 1.4.4

That's the way to go. If you distro doesn't have 1.4.4 complain to your distro. However the flathub documentation recommends using the flatpak version of flatpak-builder, which allow to work with distribution that actively don't support flatpak.

net.charlesschaefer.addictiontracker.yml
charlesschaefer: Done!

Addiction Tracker.desktop
charlesschaefer: I've changed the manifest to bring everything it needs from upstream: additional icons and metadata files. Desktop file is built during the compilation process and we won't need flathub.json anymore.

libayatana-appindicator-gtk3.json.diff
charlesschaefer: Yes, I was using the flatpak from the distro :-\

Changed to the one installed through flatpak and it worked without the patch. Thank you!

charlesschaefer:
The metafiles now are got from upstream. If you need to check them: 
Metainfo: https://github.com/charlesschaefer/addiction-tracker/blob/app-v0.0.16/bundle-scripts/net.charlesschaefer.addictiontracker.metainfo.xml
Desktop: it is generated at build time, but here's the version generated during my local flatpak's bundling tests: 

```
[Desktop Entry]
Categories=Science;
Comment=Addiction Tracker - An app for tracking substance addiction and habits
Exec=addiction-tracker
Icon=net.charlesschaefer.addictiontracker
Name=Addiction Tracker
GenericName=Opensource addiction and habits tracker
Terminal=false
Type=Application
MimeType=x-scheme-handler/addictiontracker
StartupNotify=true
Keywords=Health;Addiction;Substance Abuse;Smoking;Alcohol
SingleMainWindow=true
X-GNOME-UsesNotifications=true
X-Desktop-File-Install-Version=0.28
```

icons/hicolor/256x256@2/apps/addiction-tracker.png
bbhtt, line 1: drop the icon from here too?

net.charlesschaefer.addictiontracker.yml
```diff
@@ -0,0 +1,55 @@
...
+      env: 
+        CARGO_HOME: /run/build/addiction-tracker/cargo
+        XDG_CACHE_HOME: /run/build/addiction-tracker/flatpak-node/cache
+        npm_config_cache: /run/build/addiction-tracker/flatpak-node/npm-cache
+        npm_config_nodedir: /usr/lib/sdk/node22
+    build-commands:
+      - npm --prefix=addiction-tracker install --offline &&
+        npm --prefix=addiction-tracker run tauri build -- --bundles deb --config src-tauri/flatpak-tauri.conf.json &&
+        VERSION=$(npm --prefix=addiction-tracker pkg get version | sed "s/\"//g") &&
+        cp "addiction-tracker/src-tauri/target/release/bundle/deb/Addiction Tracker_${VERSION}_amd64.deb" addictiontracker.deb
```
bbhtt: `Addiction Tracker_${VERSION}_amd64.deb` won't work on arm64

icons/hicolor/256x256@2/apps/addiction-tracker.png
charlesschaefer: Ok, I thought, from the docs, you'd need an icon in the repo too. I'll remove it.

net.charlesschaefer.addictiontracker.yml
charlesschaefer: Yeah, totally forgot that. Thanks!

bbhtt:
bot, build

charlesschaefer:
Just tested the remote built package and it works as expected. 

Thank you for offering your time and knowledge to help us to publish our work ;-)

net.charlesschaefer.addictiontracker.yml
```diff
@@ -0,0 +1,56 @@
...
+  - --share=ipc
+  - --share=network
+
+modules:
+  - name: node22
+    buildsystem: simple
+    build-commands:
+      - /usr/lib/sdk/node22/install-sdk.sh
+
+  - shared-modules/libappindicator/libappindicator-gtk3-12.10.json
```
bbhtt: you have no permission for tray so what is this doing here?

net.charlesschaefer.addictiontracker.yml
```diff
@@ -0,0 +1,56 @@
...
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node22
+  - org.freedesktop.Sdk.Extension.rust-stable
+command: addiction-tracker
+finish-args:
+  - --device=dri
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc
+  - --share=network
```
bbhtt, line 13: what is network for?

bbhtt:
is the app a web wrapper?

net.charlesschaefer.addictiontracker.yml
charlesschaefer: What do you mean? Do I need to declare something to use the tray? Because it is working the way it is now. I can interact with the tray icon.

net.charlesschaefer.addictiontracker.yml
charlesschaefer: Despite the fact that the app works offline and don't access internet, the user can synchronize the data between other devices in their local network. That's why I need it.

charlesschaefer:
> is the app a web wrapper?

Nope. It is built with web tech, but just that. It uses the OS' webview to render the Angular frontend. But works totally offline.

net.charlesschaefer.addictiontracker.yml
charlesschaefer: > what is network for?
Here's a demonstration of the feature. The user can fetch their data from another device (discover through mDNS) by typing in the device receiving data the decrypting code it can see on the device sending data.

https://github.com/user-attachments/assets/a067de1c-f301-4f11-84bf-839b273ddd65

net.charlesschaefer.addictiontracker.yml
bbhtt: yes tray requires some permission like `--talk-name=org.kde.StatusNotifierWatcher`, or `com.cannonical.*` etc.

charlesschaefer: Yes, I've just updated the code with these permissions. 

I have a question. I'm finding it very hard to find the kind of permissions/portals we should use (I'm preparing another app to flatpak). There's some place where these permissions are documented? Sometimes even searching on google is hard to find the exact portal we should be using. 

Also, may I assume that if a service is exposed in DBus as org.some.Service.Method, I can just use --talk-name=org.some.Service and it will be enabled for the package?

bbhtt: some are granted by default, see https://docs.flatpak.org/en/latest/sandbox-permissions.html#d-bus-access

for others, see the "audit" part.

net.charlesschaefer.addictiontracker.yml
```diff
@@ -0,0 +1,60 @@
...
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=org.freedesktop.StatusNotifier.Watcher
```
bbhtt: `org.freedesktop.StatusNotifier.Watcher` does not exist, it's not implemented anywhere

net.charlesschaefer.addictiontracker.yml
```diff
@@ -0,0 +1,60 @@
...
+  - --talk-name=com.canonical.indicator
+  - --talk-name=org.ayatana.indicator # as suggested by https://ayatanaindicators.github.io/code/
```
bbhtt: can you check https://docs.flatpak.org/en/latest/debugging.html#audit-session-or-system-bus-traffic to see what is actually needed? not all of these should be required.

charlesschaefer: I just saw that, and I tried to use it. My point here, to put all these, is that I'll test only in one distribution (and looks like in this distro the permission is already open because it is working without this). And for what I saw in my research, each Desktop Manager has a different implementation for App Indicators. So, I'll test on Ubuntu with Gnome, I'll need org.ayatana.indicator. But, if I test on Ubuntu with Unity, it will need com.canonical.indicator and in Fedora with KDE, it will use org.kde.StatusNotifierWatcher. 

Does that make sense? Is that really the way it works?

bbhtt: no each DE doesn't have their own, only KDE implements the `StatusNotifierWatcher`, and the rest are `com.canonical.*` or `org.ayatana` depending on what you have used.

None of this is specific to any DE.

net.charlesschaefer.addictiontracker.yml
charlesschaefer: Oh, ok! I saw that somewhere and thought it would be better to add the most comprehensive list possible to target all DMs.

net.charlesschaefer.addictiontracker.yml
charlesschaefer: Ok, got it. But it is still plausible, then, to leave these 3 (kde, canonical and ayatana), right?

bbhtt: The interface needed depends on what library you have used to show the tray libappindicator, libayatana-indicator etc. Electron/chromium uses `org.kde.statusnotifierwatcher`.

bbhtt: > Ok, got it. But it is still plausible, then, to leave these 3 (kde, canonical and ayatana), right?

interact with the tray and post the dbus log that you got

charlesschaefer: Ah, ok. Got it! I'll leave only ayatana then. Thanks ;-)

charlesschaefer: > > Ok, got it. But it is still plausible, then, to leave these 3 (kde, canonical and ayatana), right?
> 
> interact with the tray and post the dbus log that you got

Ok, here's a part after I click to reopen the app when it is minimized to the tray. An app version built without these permissions. In the attachment are the full log entries. You'll note that there are two lines about org.kde.StatusNotifierWatcher 🤔 

[flatpak.log](https://github.com/user-attachments/files/20584991/flatpak.log)

```log
[clicked the tray icon and the menu entry "Show/Hide"]

B2081: <- :1.37 call org.gtk.Actions.DescribeAll at /org/gtk/Application/anonymous
B2085: <- :1.37 call org.freedesktop.DBus.Properties.GetAll at /org/gtk/Application/anonymous
C40: -> :1.37 return from B2081
C41: -> :1.37 return from B2085
B2088: <- :1.37 call org.gtk.Actions.DescribeAll at /org/gtk/Application/anonymous/window/2
C42: -> :1.37 return from B2088

```

charlesschaefer: Just asked to exit the app from the tray, and it logged this last line: 
```
C43: -> all signal org.kde.StatusNotifierItem.NewStatus at /org/ayatana/NotificationItem/tray_icon_tray_app
```

- all signal org.**kde**.StatusNotifierItem.NewStatus at /org/**ayatana**/NotificationItem/tray_icon_tray_app

I'm confused now hahaha

charlesschaefer: And here's the output when I try to build without some of the appindicator libraries: 

```log
    Finished `release` profile [optimized] target(s) in 57.00s
       Built application at: /run/build/addiction-tracker/addiction-tracker/src-tauri/target/release/addiction-tracker
thread '<unnamed>' panicked at crates/tauri-cli/src/interface/rust.rs:1503:17:


>>>>>>>>> Can't detect any appindicator library <<<<<<<<<


note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
FB: host_command_exited_cb 45003 34304

Error: module addiction-tracker: Child process exited with code 134
FB: Unmounting read-only fs: fusermount -uz /home/charles/Documents/Projetos/flathub/.flatpak-builder/rofiles/rofiles-j5KZwi
```

charlesschaefer: Hey, just tested on another distro (Ubuntu 24.04) and there the appindicator didn't work without the permission, as it did with Pop! OS. 

I needed to use the `org.kde.StatusNotifierWatcher` portal, even when using the libayatana-appindicator lib. 

So I think we're ok with only this permission.

bbhtt:
bot, build

bbhtt:
The tray icon does not work, and the tray menu seems to be in a different language?


https://github.com/user-attachments/assets/fdce85c8-0cb4-4814-9ed1-a198cebaf01e

charlesschaefer:
> The tray icon does not work, and the tray menu seems to be in a different language?
> Screencast.mp4

Yes, you're right. I just discovered that in my system it was getting the icon from my previous installation from .deb. And now it is even worse, it is getting the icon from another app 🤔 

I'll try to understand what is happening here and get back when I fix it. 

If you have any insight of why it is getting the icon from another app (also a tauri app but with totally different name and that I never built with flatpak), let me know.

bbhtt:
the icon name for the tray has to match the appid and you have to install it in the same place as your desktop file icon so that it gets exported outside the sandbox and the tray on host can access it.

charlesschaefer:
> the icon name for the tray has to match the appid and you have to install it in the same place as your desktop file icon so that it gets exported outside the sandbox and the tray on host can access it.

When you say "the same place as your desktop file", you mean `/app/share/icons`, right? Or some other structure under `/app/share/applications/`?

In the first case, I'm already placing the icons under `/app/share/icons/hicolor/{size}/apps/{app_id}.png`

bbhtt:
> In the first case, I'm already placing the icons under /app/share/icons/hicolor/{size}/apps/{app_id}.png

Yes the tray icon should be here too or can be under `scalable` folder too.

charlesschaefer:
> The tray icon does not work, and the tray menu seems to be in a different language?
> 
>  Screencast.mp4

I'm 

> > In the first case, I'm already placing the icons under /app/share/icons/hicolor/{size}/apps/{app_id}.png
> 
> Yes the tray icon should be here too or can be under `scalable` folder too.

I'm still no able to make this work. I have all the icons under `/app/share/icons/hicolor/{size|scalable}/apps/{app_id}.{png|svg}` and it doesn't work with Gnome :-\ 

But I'm still trying to solve. Then I update the PR ;-)

charlesschaefer:
Hey, I'm having a bad time trying to integrate this tray icon within flatpak. I've tracked the issue and now I see that gnome appindicator plugin tries to open the icon file using a path like `/run/user/1000/tray-icon/tray-icon-2-0.png`. It tries that both when running the app from a .deb directly or from the flatpak package. But, when running from flatpak, it can't read the file because it isn't there in the host machine. But, if you try `flatpak run --command=bash` in the package and checks that path, the image is there: 

```bash
# after running flatpak run net.charlesschaefer.testeflatpak in another terminal
ls /run/user/1000/tray-icon/tray-icon-2-0.png
ls: cannot access '/run/user/1000/tray-icon/tray-icon-2-0.png': No such file or directory

flatpak run --command=bash net.charlesschaefer.testeflatpak
[📦 net.charlesschaefer.testeflatpak ~]$ ls /run/user/1000/tray-icon/tray-icon-2-0.png
/run/user/1000/tray-icon/tray-icon-2-0.png
```

Also, if I run `xdg-open /run/user/1000/tray-icon/tray-icon-2-0.png` from inside the flatpak package, it opens the image for me. 

Despite all that, the image is not being shown in the tray icon. But, if I install the .deb on the host machine, run and close the app, this file is created in the host. Then, if I run the flatpak package, it will show the icon it finds in the host machine, instead of the one found in the flatpak package.

I think this happens because tauri uses libayatana and passes an icon path instead of an icon name when asking to render the icon. But there is differences between the flatpak exposed file system and the host filesystem. Do you know if there is any kind of configuration I need to do on the manifest to expose flatpak's `/run/user/1000/tray-icon/tray-icon-2-0.png` as the host's `/run/user/1000/tray-icon/tray-icon-2-0.png`?

bbhtt:
That's not a good location for the sandbox if Tauri is using that the sandbox has its own `$XDG_RUNTIME_DIR` which is inaccessible from the outside and vice-versa

Looks like you can set the path here https://docs.rs/tauri/2.0.0/tauri/tray/struct.TrayIconBuilder.html#method.temp_dir_path


Try using some path that is accessible from both inside the sandbox and outside, like `$XDG_CACHE_HOME/appname/trayicon.png`.

charlesschaefer:
Yeah, just after sending the message I had an insight and tried one thing that worked. I can set `--filesystem=xdg-run/tray-icon` and it works. Is that ok too?

bbhtt:
You can use it but ideally something like that should not require any extra filesystem permissions.

bbhtt:
bot, build

charlesschaefer:
Oh, finally I've finished that :-)

I've set a different path to the icon. Used the xdg-cache path, to which we already have access.

I had to change versions marks on the manifest to work with the most new code. I've run everything locally and it is running. The last commit I've pushed was lacking one of the updated dependencies manifest. That's why the build failed.

bbhtt:
bot, build

charlesschaefer:
Any idea of why we have this error? It is complaining that we don't have the npm package `@esbuild/linux-arm64` in cache, and it tries to download but hasn't network access. 

But the package is declared in the upstream package-lock.json and in this repo's [npm-sources.0.json](https://github.com/flathub/flathub/blob/b61fbd2c0a052c9f42e6a08c68ea1c853a09a1b0/npm-sources.0.json#L14). So, the package should be downloaded previously to be available in cache, no? 

Do you know if there is any known issue with the npm manifest generation for arm64?

bbhtt:
I don't know.

The JS ecosystem makes it as hard as possible to build offline.

charlesschaefer:
> I don't know.
> 
> The JS ecosystem makes it as hard as possible to build offline.

Yeah, you're right :-\

Anyway, I've just searched for some numbers about the arm64 user base, and it is so small that I think it isn't worthy to try to dive deep in that to solve the arm64 build. So I've created the flatpak.json again, restricting builds only to x86_64.

charlesschaefer:
bot, build

charlesschaefer:
bot, build

charlesschaefer:
Even with the `"only-arches"="x86_64"` configuration, the build still starts the aarch64 build. Since it fails because of NPM, it also cancels the x86_64 execution.

Any idea for how to solve that?

bbhtt:
The file is called flathub.json not flatpak.json

charlesschaefer:
bot, build

PR #6573, opened 2025-05-31
Add io.github.pieterdd.StretchBreak

pieterdd:

- [x] Please describe the application briefly. **Tool to help you take regular breaks from using your computer**
- [x] I am an _(please write whichever is applicable)_ **author** to the project.

https://github.com/user-attachments/assets/233896df-3e92-405a-a0e5-4ada0edbc717

(You're also seeing the accompanying GNOME Shell extension in action here, which I assume cannot be packaged as part of the Flatpak archive.)

### DIFF
--- a/io.github.pieterdd.StretchBreak.yaml
+++ b/io.github.pieterdd.StretchBreak.yaml
@@ -1,0 +1,40 @@
+id: io.github.pieterdd.StretchBreak
+runtime: org.gnome.Platform
+runtime-version: '48'
+sdk: org.gnome.Sdk
+sdk-extensions:
+- org.freedesktop.Sdk.Extension.rust-stable
+command: stretch-break
+finish-args:
+- "--share=ipc"
+- "--socket=fallback-x11"
+- "--socket=wayland"
+- "--socket=pulseaudio"
+- "--device=dri"
+- "--env=G_MESSAGES_DEBUG=none"
+- "--env=RUST_BACKTRACE=1"
+- "--talk-name=org.gnome.Mutter.IdleMonitor"
+- "--talk-name=org.freedesktop.ScreenSaver"
+- "--talk-name=org.kde.ScreenSaver"
+modules:
+  - name: stretch-break
+    buildsystem: simple
+    build-options:
+      append-path: "/usr/lib/sdk/rust-stable/bin"
+      env:
+        CARGO_HOME: "/run/build/stretch-break/cargo"
+    build-commands:
+    - cargo --offline fetch --manifest-path Cargo.toml --verbose
+    - cargo --offline build --release --verbose
+    - install -Dm755 ./target/release/stretch-break /app/bin/stretch-break
+    - install -Dm644 meta/logo-color-16x16.png /app/share/icons/hicolor/16x16/apps/io.github.pieterdd.StretchBreak.png
+    - install -Dm644 meta/logo-color-32x32.png /app/share/icons/hicolor/32x32/apps/io.github.pieterdd.StretchBreak.png
+    - install -Dm644 meta/logo-color-128x128.png /app/share/icons/hicolor/128x128/apps/io.github.pieterdd.StretchBreak.png
+    - install -Dm644 -t /app/share/applications/ meta/io.github.pieterdd.StretchBreak.desktop
+    - install -Dm644 -t /app/share/metainfo/ meta/io.github.pieterdd.StretchBreak.metainfo.xml
+    - install -Dm644 meta/io.github.pieterdd.StretchBreak.Core.service-flatpak /app/share/dbus-1/services/io.github.pieterdd.StretchBreak.Core.service
+    sources:
+    - type: git
+      url: https://github.com/pieterdd/StretchBreak
+      tag: initial-flatpak
+    - "cargo-sources.json"
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+/.flatpak-builder
+/repo
--- a/README.md
+++ b/README.md
@@ -1,0 +1,7 @@
+# Flatpak manifest for [Stretch Break](https://github.com/pieterdd/StretchBreak/)
+
+### Build instructions
+- In case of Rust dependency changes, get [flatpak-cargo-generator](https://pypi.org/project/flatpak-cargo-generator/), then execute `flatpak-cargo-generator /path/to/code/repo/Cargo.lock -o /path/to/manifest/repo/cargo-sources.json`.
+- Run `./flatpak-build.sh` to do a local test build. This script is not used by Flathub's build system.
+
+Note that the GNOME Shell extension will have to be installed separately.
--- a/cargo-sources.json
+++ b/cargo-sources.json
@@ -1,0 +1,4142 @@
+[
+    {
+        "type": "archive",
+        "archive-type": "tar-gzip",
+        "url": "https://static.crates.io/crates/CoreFoundation-sys/CoreFoundation-sys-0.1.4.crate",
+        "sha256": "d0e9889e6db118d49d88d84728d0e964d973a5680befb5f85f55141beea5c20b",
+        "dest": "cargo/vendor/CoreFoundation-sys-0.1.4"
+    },
+    {
+        "type": "inline",
+        "contents": "{\"package\": \"d0e9889e6db118d49d88d84728d0e964d973a5680befb5f85f55141beea5c20b\", \"files\": {}}",
+        "dest": "cargo/vendor/CoreFoundation-sys-0.1.4",
+        "dest-filename": ".cargo-checksum.json"
+    },
+    {
+        "type": "archive",
+        "archive-type": "tar-gzip",
+        "url": "https://static.crates.io/crates/IOKit-sys/IOKit-sys-0.1.5.crate",
+        "sha256": "99696c398cbaf669d2368076bdb3d627fb0ce51a26899d7c61228c5c0af3bf4a",
+        "dest": "cargo/vendor/IOKit-sys-0.1.5"
+    },
+    {
+        "type": "inline",
+        "contents": "{\"package\": \"99696c398cbaf669d2368076bdb3d627fb0ce51a26899d7c61228c5c0af3bf4a\", \"files\": {}}",
+        "dest": "cargo/vendor/IOKit-sys-0.1.5",
+        "dest-filename": ".cargo-checksum.json"
+    },
+    {
+        "type": "archive",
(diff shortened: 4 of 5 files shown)


io.github.pieterdd.StretchBreak.metainfo.xml
hfiguiere, line 1: This file should be (submitted and managed) upstream.

io.github.pieterdd.StretchBreak.metainfo.xml
```diff
@@ -0,0 +1,69 @@
...
+
+  <branding>
+    <color type="primary" scheme_preference="light">#ebd1d1</color>
+    <color type="primary" scheme_preference="dark">#420925</color>
+  </branding>
+
+  <screenshots>
+    <screenshot type="default">
+      <image>
+        https://raw.githubusercontent.com/pieterdd/StretchBreak/main/docs/mainWindow.png
```
hfiguiere, line 24: Screenshot URL should never reference the tip of a branch because it is no immutable.

io.github.pieterdd.StretchBreak.yaml
```diff
@@ -0,0 +1,45 @@
...
+finish-args:
+- "--share=ipc"
+- "--share=network"
+- "--socket=fallback-x11"
+- "--socket=wayland"
+- "--socket=pulseaudio"
+- "--device=dri"
+- "--env=G_MESSAGES_DEBUG=none"
+- "--env=RUST_BACKTRACE=1"
+- "--own-name=io.pieterdd.StretchBreak.Core"
```
hfiguiere, line 17: you should use upstream a name that matches the app-id. D-Bus as similar namespacing requirements where one doesn't simply take a random name. Then there is no need to `own-name`

io.github.pieterdd.StretchBreak.metainfo.xml
pieterdd: Moved it.

io.github.pieterdd.StretchBreak.metainfo.xml
pieterdd: Replaced with permalink.

io.github.pieterdd.StretchBreak.yaml
```diff
@@ -0,0 +1,45 @@
...
+  - name: stretch-break
+    buildsystem: simple
+    build-options:
+      append-path: "/usr/lib/sdk/rust-stable/bin"
+      env:
+        CARGO_HOME: "/run/build/stretch-break/cargo"
+    build-commands:
+    - cargo --offline fetch --manifest-path Cargo.toml --verbose
+    - cargo --offline build --release --verbose
+    - mkdir -p /app/bin
```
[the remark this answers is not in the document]
pieterdd: Fixed.

io.github.pieterdd.StretchBreak.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - install -Dm755 ./target/release/stretch-break /app/bin/stretch-break
+    - strip -s /app/bin/stretch-break
```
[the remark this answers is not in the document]
pieterdd: Fixed.

pieterdd:
Forgot to actually push, sorry.

hfiguiere:
bot, build

pieterdd:
Corrected a typo in the D-Bus service file, since it pointed to `/usr/bin/stretch-break` instead of `/app/bin/stretch-break`.

io.github.pieterdd.StretchBreak.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - cargo --offline fetch --manifest-path Cargo.toml --verbose
+    - cargo --offline build --release --verbose
+    - install -Dm755 ./target/release/stretch-break /app/bin/stretch-break
+    - install -Dm644 meta/logo-color-16x16.png /app/share/icons/hicolor/16x16/apps/io.github.pieterdd.StretchBreak.png
+    - install -Dm644 meta/logo-color-32x32.png /app/share/icons/hicolor/32x32/apps/io.github.pieterdd.StretchBreak.png
+    - install -Dm644 meta/logo-color-128x128.png /app/share/icons/hicolor/128x128/apps/io.github.pieterdd.StretchBreak.png
+    - install -Dm644 -t /app/share/applications/ meta/io.github.pieterdd.StretchBreak.desktop
+    - install -Dm644 -t /app/share/metainfo/ meta/io.github.pieterdd.StretchBreak.metainfo.xml
+    - install -Dm644 -t /app/share/dbus-1/services/ meta/io.github.pieterdd.StretchBreak.Core.service
+    - sed -i 's|/usr/bin/stretch-break|/app/bin/stretch-break|g' /app/share/dbus-1/services/io.github.pieterdd.StretchBreak.Core.service
```
hfiguiere, line 38: thing is this should be dealt with the upstream build system and even `/usr` is never the default prefix. This would include installing all the above. There are many GNOME application in Rust that build with meson.

io.github.pieterdd.StretchBreak.yaml
```diff
@@ -0,0 +1,43 @@
...
+- "--socket=fallback-x11"
+- "--socket=wayland"
+- "--socket=pulseaudio"
+- "--device=dri"
+- "--env=G_MESSAGES_DEBUG=none"
+- "--env=RUST_BACKTRACE=1"
+- "--talk-name=org.gnome.Mutter.IdleMonitor"
+- "--talk-name=org.freedesktop.ScreenSaver"
+- "--talk-name=org.kde.ScreenSaver"
+- "--talk-name=org.freedesktop.Notifications"
```
bbhtt: notifications don't go through the portal?

io.github.pieterdd.StretchBreak.yaml
```diff
@@ -0,0 +1,43 @@
...
+- "--share=network"
+- "--socket=fallback-x11"
+- "--socket=wayland"
+- "--socket=pulseaudio"
+- "--device=dri"
+- "--env=G_MESSAGES_DEBUG=none"
+- "--env=RUST_BACKTRACE=1"
+- "--talk-name=org.gnome.Mutter.IdleMonitor"
+- "--talk-name=org.freedesktop.ScreenSaver"
+- "--talk-name=org.kde.ScreenSaver"
```
bbhtt, line 18: what are these for?

io.github.pieterdd.StretchBreak.yaml
```diff
@@ -0,0 +1,43 @@
+id: io.github.pieterdd.StretchBreak
+runtime: org.gnome.Platform
+runtime-version: '48'
+sdk: org.gnome.Sdk
+sdk-extensions:
+- org.freedesktop.Sdk.Extension.rust-stable
+command: stretch-break
+finish-args:
+- "--share=ipc"
+- "--share=network"
```
bbhtt: what is network for?

io.github.pieterdd.StretchBreak.yaml
```diff
@@ -0,0 +1,43 @@
...
+- "--socket=fallback-x11"
+- "--socket=wayland"
+- "--socket=pulseaudio"
```
bbhtt, line 12: what is sound for?

io.github.pieterdd.StretchBreak.yaml
pieterdd: Meson is not a good fit for me, it's way too complex for what I need. If you really don't want this in the Flatpak build, I'll make a separate file with the Flatpak path in the upstream repo.

io.github.pieterdd.StretchBreak.yaml
pieterdd: Idle detection. The break countdown resets if you're idle long enough. The library that does the idle detection gets its information from `org.gnome.Mutter.IdleMonitor.GetIdletime`, `org.freedesktop.ScreenSaver.GetActiveTime` and a few other fallback places as shown [here](https://github.com/pieterdd/user-idle2-rs/blob/master/src/linux_dbus_impl.rs).

io.github.pieterdd.StretchBreak.yaml
pieterdd: Must've slipped in when copy/pasting from another manifest. I'll remove it.

io.github.pieterdd.StretchBreak.yaml
pieterdd: At the end of breaks, a notification sound is played since the user is probably not looking at their screen during the break.

io.github.pieterdd.StretchBreak.yaml
pieterdd: I call `org.freedesktop.Notifications.Notify` to create and `org.freedesktop.Notifications.CloseNotification` to revoke previously created notifications. When I remove this `--talk-name` line, I don't see any notification.

bbhtt: are you manually handling that or using some lib? something libnotify based would use the portal for example and would not need the permission

io.github.pieterdd.StretchBreak.yaml
bbhtt: is using the portal interface possible? https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Inhibit.html

io.github.pieterdd.StretchBreak.yaml
pieterdd: I had a problem with a library where it would not trigger notifications in certain conditions, so I started doing a `notify-send` call instead. Later I noticed a problem withdrawing the notification using CloseNotification when running from the Flatpak container, at which point I switched to doing `gdbus` calls for both Notify and CloseNotification.

I'll check if it works with libnotify.

io.github.pieterdd.StretchBreak.yaml
pieterdd: From what I can tell, it doesn't provide access to the same information. I'd need a way to retrieve the number of idle seconds since last keyboard/mouse input.

io.github.pieterdd.StretchBreak.yaml
pieterdd: It does. I've removed the now-obsolete permission.

bbhtt:
bot, build

bbhtt:
I don't see a LICENSE or COPYING file in https://github.com/pieterdd/StretchBreak the metainfo says it is GPL3+

pieterdd:
Confirmed GPLv3 license at https://github.com/pieterdd/StretchBreak/blob/main/LICENSE.

PR #6582, opened 2025-06-02
Add com.sciactive.QuickDAV

hperrin:

Allows you to transfer files between devices by running a temporary, secure WebDAV server on your system. You control what is shared and can see the logs from activity. It also supports gamepads when launched from Steam in big picture mode (great for transferring files to/from a Steam Deck).

https://github.com/user-attachments/assets/5d6a5f8a-a82c-4076-a68a-f4dea5c8143a

### DIFF
--- a/com.sciactive.QuickDAV.desktop
+++ b/com.sciactive.QuickDAV.desktop
@@ -1,0 +1,11 @@
+[Desktop Entry]
+Name=QuickDAV
+GenericName=QuickDAV
+Comment=QuickDAV
+Keywords=file share;sharing;network
+Exec=run.sh %f
+Icon=com.sciactive.QuickDAV
+Version=1.0
+Type=Application
+Terminal=false
+Categories=Utility
--- a/com.sciactive.QuickDAV.metainfo.xml
+++ b/com.sciactive.QuickDAV.metainfo.xml
@@ -1,0 +1,83 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.sciactive.QuickDAV</id>
+  <launchable type="desktop-id">com.sciactive.QuickDAV.desktop</launchable>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>Apache-2.0</project_license>
+
+  <name>QuickDAV</name>
+
+  <categories>
+    <category>Utility</category>
+  </categories>
+
+  <summary>Transfer files between devices</summary>
+
+  <description>
+    <p>Network file transfer app. Simple and easy upload, download, and management of your files
+      from any other device.</p>
+    <p>It works by creating a temporary, secure WebDAV server that you can connect to from a
+      remote device. If your device doesn't support WebDAV, you can connect with a web browser.</p>
+  </description>
+
+  <developer_name>Hunter Perrin</developer_name>
+
+  <releases>
+    <release version="3.4.0" date="2025-06-02">
+      <description>
+        <p>fix: improve performance of the log view when many log lines are present</p>
+        <p>chore: update packages and add fix for new version</p>
+      </description>
+    </release>
+  </releases>
+
+  <screenshots>
+    <screenshot type="default">
+      <image type="source">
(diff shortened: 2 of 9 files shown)


com.sciactive.QuickDAV.desktop
bbhtt, line 1: this should be upstream

com.sciactive.QuickDAV.metainfo.xml
bbhtt, line 1: this should be upstream

com.sciactive.QuickDAV.metainfo.xml
```diff
@@ -0,0 +1,83 @@
...
+      <description>
+        <p>fix: improve performance of the log view when many log lines are present</p>
+        <p>chore: update packages and add fix for new version</p>
+      </description>
+    </release>
+  </releases>
+
+  <screenshots>
+    <screenshot type="default">
+      <image type="source">
```
bbhtt: only one screenshot can be default

com.sciactive.QuickDAV.yml
```diff
@@ -0,0 +1,63 @@
...
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node22
+command: run.sh
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --share=network
+  - --device=dri
+  # Needed for game controller support in SteamOS.
+  - --device=input
```
bbhtt: this won't work on flatpak 1.14 which is still supported release of flatpak.

com.sciactive.QuickDAV.yml
```diff
@@ -0,0 +1,63 @@
...
+  - --filesystem=/run/udev:ro # overkill, but necessary: https://github.com/flathub/org.chromium.Chromium/issues/40
+  # Needed for sharing files.
+  - --filesystem=home
```
bbhtt, line 18: why is home needed? does it not work with filechooser portals?

com.sciactive.QuickDAV.yml
hperrin: It's required to get controllers working. Is there a way I can make it target only newer versions?

bbhtt: use `all`, majority of distros are using 1.14. 1.16 came out only recently.

com.sciactive.QuickDAV.yml
hperrin: It would, but then I'd need to open a file chooser immediately when the app opens, which is not a great experience, and is different than how it works on all other platforms.

com.sciactive.QuickDAV.yml
hperrin: Ah, ok. Thanks.

com.sciactive.QuickDAV.yml
bbhtt: maybe you can limit it to known xdg-download, document etc. directories?

bbhtt: or does it allow to share whole folders? from your description it seemed like a file sharing thing but the video shows a folder tab.

hperrin: It allows you to share whole folders. It sets up a WebDAV server with your home folder as the root share by default, then you can limit it further if you want. Also, if you start it on a Steam Deck, it will have your home folder and the SD card by default.

The reason I had it limited to the "quickdav" folder in the video is cause I didn't want to put the contents of my home folder in a public video.

hperrin: I updated the description in the metainfo to state that you can share folders. I should probably update that everywhere, because that's a really good point. I kinda forgot that most (easy) file sharing apps can only share files, not folders.

com.sciactive.QuickDAV.yml
```diff
@@ -0,0 +1,59 @@
...
+base: org.electronjs.Electron2.BaseApp
+base-version: "24.08"
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node22
+command: run.sh
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --share=network
+  - --device=dri
```
bbhtt:
```suggestion
```

dri is in all

bbhtt:
bot, build

PR #6587, opened 2025-06-03
Add app.katvan.Katvan

IgKh:

- [X] Please describe the application briefly: Katvan is a bare-bones editor for [Typst](https://typst.app) files, that has a strong focus on writing technical documents in Right-to-Left languages.
- [X] Please attach or link a screenrecord showcasing the application on Linux using the Flatpak:
  
[Screencast_20250603_222840.webm](https://github.com/user-attachments/assets/011275b0-02c0-4d2a-a4ef-1f6674f30019)

- [X] I am the **author** of the project.

### DIFF
--- a/app.katvan.Katvan.yml
+++ b/app.katvan.Katvan.yml
@@ -1,0 +1,74 @@
+id: app.katvan.Katvan
+runtime: org.kde.Platform
+runtime-version: '6.8'
+sdk: org.kde.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+command: katvan
+rename-desktop-file: katvan.desktop
+rename-icon: katvan
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --share=network
+  - --filesystem=xdg-data/typst/packages:ro
+build-options:
+  append-path: /usr/lib/sdk/rust-stable/bin
+
+modules:
+  - name: corrosion
+    buildsystem: cmake-ninja
+    builddir: true
+    cleanup:
+      - '*'
+    sources:
+      - type: git
+        url: https://github.com/corrosion-rs/corrosion.git
+        commit: b1fab721655c5c4b1b08a083d3cd29f163af75d0
+        tag: v0.5.1
+  
+  - name: cxxbridge
+    buildsystem: simple
+    build-commands:
+      - cargo install --root ${FLATPAK_DEST} --path .
+    build-options:
+      env:
+        CARGO_HOME: /run/build/cxxbridge/cargo
+    cleanup:
+      - '*'
+    sources:
+      - type: archive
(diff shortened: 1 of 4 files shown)


IgKh:
PR for the requisite exception: https://github.com/flathub-infra/flatpak-builder-lint/pull/657

app.katvan.Katvan.yml
```diff
@@ -0,0 +1,74 @@
+id: app.katvan.Katvan
+runtime: org.kde.Platform
+runtime-version: '6.8'
```
IgKh, line 3: This is not the latest KDE runtime - however the `6.9` branch has Qt 6.9.0, which unfortunately has a terrible regression under Wayland (https://bugreports.qt.io/browse/QTBUG-136110) causing crashes, and which is quite easy to hit while using this app. Qt 6.9.1 has a fix, but it is not yet updated in the runtime (and there doesn't seem to be a pending MR or something). As Flatpak makes working around the bug rather more difficult than it would otherwise be, I chose to submit with 6.8 and upgrade later.

bbhtt:
bot, build

app.katvan.Katvan.yml
```diff
@@ -0,0 +1,74 @@
...
+  - org.freedesktop.Sdk.Extension.rust-stable
+command: katvan
+rename-desktop-file: katvan.desktop
+rename-icon: katvan
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --share=network
```
bbhtt, line 15: what is network for?

IgKh: TL;DR: To download packages from the [Typst Universe](https://typst.app/universe/).

Bit of context: the convention in the Typst world is that external packages are automatically downloaded and cached on the first time the compiler sees an import statement from a namespace that represents a repository. For example - if you work on presentation slides, you'd have something like `#import "@preview/touying:0.6.1": *` at the top of the document; this will both automatically download that version of `touying` from the server associated with `@preview` (which is the current "official" central repository) if missing from the cache, and make the package's functions available for use.

This is different from how the TeX world traditionally does things, where the user has to download extra packages from CTAN ahead of times (or use a package manager included in their TeX distribution, which is distinct from the compiler/processor).

Users that like that way of working (or are developing their own package) can use an external Typst package manager (like typship) to download hand-picked packages to their machine, and then import from the `@local` namespace. Katvan has a toggle in settings to disable the `@preview` namespace for users such inclined, and it will not make any attempt to connect to the network. Such users can also revoke the network permission from the Flatpak with no loss of functionality (other than a maybe no so helpful error if they open a document that tries to import a package from `@preview`).

However the most common way of usage includes getting packages from the universe repository, so network access should be allowed by default. In principal Katvan will never attempt to do anything other than GET requests to https://packages.typst.org and collects no telemetry data, but there is no way to specify that on the manifest.

PR #6590, opened 2025-06-04
Add io.github.aginies.watermark

aginies:

- [X] Please describe the application briefly: 
 Python Application to Add a Watermark text to Images
https://www.youtube.com/watch?v=rNg0RGUvESI
- [X] I am author/developer/upstream contributor to the project.


watermark_app.py
hfiguiere, line 1: the flathub repository isn't meant to host the upstream source code.

hfiguiere:
This submission might not be acceptable: see point 3 of https://docs.flathub.org/docs/for-app-authors/requirements#inclusion-policy

io.github.aginies.watermark.xml
hfiguiere, line 1: this file is never referenced nor used.

io.github.aginies.watermark.yml
```diff
@@ -0,0 +1,38 @@
+app-id: io.github.aginies.watermark
+runtime: org.gnome.Platform
+runtime-version: '48'
+sdk: org.gnome.Sdk
+command: watermark_app
+tags:
+  - utility
+  - image-processing
+desktop-file-name-suffix: (Flatpak)
```
hfiguiere:
```suggestion
```

io.github.aginies.watermark.yml
```diff
@@ -0,0 +1,38 @@
+app-id: io.github.aginies.watermark
+runtime: org.gnome.Platform
```
hfiguiere: why do you need the GNOME platform?

io.github.aginies.watermark.xml
```diff
@@ -0,0 +1,35 @@
...
+    </p>
+  </description>
+  <url type="homepage">https://github.com/aginies/watermark</url>
+  <launchable type="desktop-id">io.github.aginies.watermark.desktop</launchable>
+  <provides>
+    <binary>watermark_app</binary>
+  </provides>
+  <screenshots>
+    <screenshot type="default">
+	    <image>https://github.com/aginies/watermark/blob/e9679ab2404849ea05e22e3730073c6df3ae95b1/watermark_app.jpg</image>
```
hfiguiere: this is not even a link to the image, but to the GitHub page for the file.

io.github.aginies.watermark.yml
```diff
@@ -0,0 +1,38 @@
...
+      - install -Dm644 io.github.aginies.watermark.png /app/share/icons/hicolor/256x256/apps/io.github.aginies.watermark.png
+      - install -Dm644 io.github.aginies.watermark.desktop /app/share/applications/io.github.aginies.watermark.desktop
+    sources:
+      - type: file
+        path: watermark_app.py
+      - type: file
+        path: io.github.aginies.watermark.png
+      - type: file
+        path: io.github.aginies.watermark.desktop
+  - name: python-dependencies
```
hfiguiere: 1. dependencies are always installed first.  It's logical.
2. build must be offline. This isn't.

python3-modules.json
hfiguiere, line 1: this file is never used / referenced.

watermark_app.py
hfiguiere: this problem trigger paragraph 3 of the rejection reasons.

aginies: now using the upstream from github tarball

io.github.aginies.watermark.xml
aginies: ok, i was not sure it was mandatory or not. I just removed it

io.github.aginies.watermark.yml
aginies: removed.

io.github.aginies.watermark.yml
aginies: i swith to freedesktop. Initially i had some issue with python while i was testing, this is no more the case.

io.github.aginies.watermark.xml
aginies: this file has been removed

python3-modules.json
aginies: it is now used to get pillow module

io.github.aginies.watermark.yml
aginies: this should be addressed now.

aginies:
> appstream file (never installed), icon and desktop file should all be in the same place as the source code: upstream.

Thanks for the review and and taking the time to comment to help me get this fixed. I hope I was able to fix all the issues properly.

io.github.aginies.watermark.yml
```diff
@@ -0,0 +1,35 @@
...
+      - tar -xvf watermark-2.3.tar.gz
+      # remove unwated files
+      - rm -rvf watermark-2.3/old watermark-2.3/windows watermark-2.3/DejaVuSans.ttf 
+      - rm -rvf watermark-2.3/Makefile watermark-2.3/test.jpg *.json
+      - install -Dm755 watermark-2.3/watermark_app.py /app/bin/watermark_app
+      - install -Dm644 watermark-2.3/io.github.aginies.watermark.png /app/share/icons/hicolor/256x256/apps/io.github.aginies.watermark.png
+      - install -Dm644 watermark-2.3/io.github.aginies.watermark.desktop /app/share/applications/io.github.aginies.watermark.desktop
+      - cp -a watermark-2.3/locale /app/share/
+    sources:
+      - type: file
```
hfiguiere: this should be type `archive`.

io.github.aginies.watermark.yml
```diff
@@ -0,0 +1,35 @@
...
+  - python3-modules.json
+  - name: watermark_app
+    no-autogen: true
+    no-make-install: true
+    buildsystem: simple
+    build-commands:
+      - tar -xvf watermark-2.3.tar.gz
+      # remove unwated files
+      - rm -rvf watermark-2.3/old watermark-2.3/windows watermark-2.3/DejaVuSans.ttf 
+      - rm -rvf watermark-2.3/Makefile watermark-2.3/test.jpg *.json
```
hfiguiere:
```suggestion
```

io.github.aginies.watermark.yml
```diff
@@ -0,0 +1,35 @@
...
+  - --socket=x11
+  - --device=dri
+  - --filesystem=host
+  - --talk-name=org.freedesktop.Notifications
+  - --share=network
+modules:
+  - python3-modules.json
+  - name: watermark_app
+    no-autogen: true
+    no-make-install: true
```
hfiguiere:
```suggestion
```

io.github.aginies.watermark.yml
```diff
@@ -0,0 +1,35 @@
...
+sdk: 'org.freedesktop.Sdk'
+command: watermark_app
+tags:
+  - utility
+  - image-processing
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --device=dri
+  - --filesystem=host
```
hfiguiere: why does this need `host` access?

io.github.aginies.watermark.yml
```diff
@@ -0,0 +1,35 @@
+app-id: io.github.aginies.watermark
+runtime: 'org.freedesktop.Platform'
+runtime-version: '24.08'
+sdk: 'org.freedesktop.Sdk'
+command: watermark_app
+tags:
+  - utility
+  - image-processing
```
hfiguiere:
```suggestion
```

io.github.aginies.watermark.yml
aginies: You need to select the image from your home directory, and you need to store the image with the watermark you created. So the tool needs an access to the host filesystem.

io.github.aginies.watermark.yml
aginies: this was to cleanup, but we can remove this rm commands

io.github.aginies.watermark.yml
aginies: ok done in last commit.

aginies:
I addressed all the issues. Thanks again for the review.

hfiguiere:
bot, build

io.github.aginies.watermark.yml
```diff
@@ -0,0 +1,27 @@
...
+      - tar -xvf watermark-2.3.tar.gz
+      - install -Dm755 watermark-2.3/watermark_app.py /app/bin/watermark_app
+      - install -Dm644 watermark-2.3/io.github.aginies.watermark.png /app/share/icons/hicolor/256x256/apps/io.github.aginies.watermark.png
+      - install -Dm644 watermark-2.3/io.github.aginies.watermark.desktop /app/share/applications/io.github.aginies.watermark.desktop
+      - cp -a watermark-2.3/locale /app/share/
+    sources:
+      - type: archive
+        url: https://github.com/aginies/watermark/archive/refs/tags/2.3.tar.gz
+        sha256: 1f569d7f8577dfa440145b43590f6f6cc9823be05f7c1c4f078eaa9bd9a83641
+        dest-filename: watermark-2.3.tar.gz
```
hfiguiere:
```suggestion
```

io.github.aginies.watermark.yml
```diff
@@ -0,0 +1,27 @@
...
+modules:
+  - python3-modules.json
+  - name: watermark_app
+    buildsystem: simple
+    build-commands:
+      - tar -xvf watermark-2.3.tar.gz
+      - install -Dm755 watermark-2.3/watermark_app.py /app/bin/watermark_app
+      - install -Dm644 watermark-2.3/io.github.aginies.watermark.png /app/share/icons/hicolor/256x256/apps/io.github.aginies.watermark.png
+      - install -Dm644 watermark-2.3/io.github.aginies.watermark.desktop /app/share/applications/io.github.aginies.watermark.desktop
+      - cp -a watermark-2.3/locale /app/share/
```
hfiguiere:
```suggestion
      - install -Dm755 watermark_app.py /app/bin/watermark_app
      - install -Dm644 io.github.aginies.watermark.png /app/share/icons/hicolor/256x256/apps/io.github.aginies.watermark.png
      - install -Dm644 io.github.aginies.watermark.desktop /app/share/applications/io.github.aginies.watermark.desktop
      - cp -a locale /app/share/
```

build will fail. It was more that just changing the type. But then you didn't check the build.

io.github.aginies.watermark.yml
aginies: Archive are automatically managed... ok

io.github.aginies.watermark.yml
aginies: Sorry, yes i didn't check the build... I was not aware that swithcing to archive will manage automatically the tar and the cd to the correct directory.

io.github.aginies.watermark.yml
bbhtt: This must use the portal, prompt a filechooser to select the file, prompt another one to save it.

Or limit it to xdg-pictures only.

io.github.aginies.watermark.yml
```diff
@@ -0,0 +1,25 @@
...
+runtime: 'org.freedesktop.Platform'
+runtime-version: '24.08'
+sdk: 'org.freedesktop.Sdk'
+command: watermark_app
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --device=dri
+  - --filesystem=host
+  - --talk-name=org.freedesktop.Notifications
```
bbhtt: use the notifications portal.

bbhtt:
This is currently blocked by point 3 here https://docs.flathub.org/docs/for-app-authors/requirements

bbhtt:
Why did you choose tk? It doesn't support wayland and is old. Why not GTK through pygobject or Qt through qtpy/pyqt/pyside?

Also Tcl/tk is going to be removed from the runtime from 25.08 version.

aginies:
> Why did you choose tk? It doesn't support wayland and is old. Why not GTK through pygobject or Qt through qtpy/pyqt/pyside?
> Also Tcl/tk is going to be removed from the runtime from 25.08 version.

Because it was mostly impossible to get the pygobject available on Windows platform. Initially it was written with pyobject then I switched to TK to be able to get an .exe versio for windows. I Will update my code with pyobject for the linux version and keep the tk one for windows.

bbhtt:
Some additional functionality will help, like choosing position of watermark, styles etc. or removing them depending on what your scope is.

io.github.aginies.watermark.yml
aginies: ok i will use org.freedesktop.portal.Notification.

io.github.aginies.watermark.yml
```diff
@@ -0,0 +1,25 @@
...
+runtime: 'org.freedesktop.Platform'
+runtime-version: '24.08'
+sdk: 'org.freedesktop.Sdk'
+command: watermark_app
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --device=dri
+  - --filesystem=host
+  - --talk-name=org.freedesktop.portal.Notification
```
bbhtt: you don't use the portal by adding permissions, you have to modify your code to use it...

eg. libnotify already supports it so if you use it through some binding then it should work automatically

aginies:
> Some additional functionality will help, like choosing position of watermark, styles etc. or removing them depending on what your scope is.

The idea was to create watermark in random position, so this is hard to remove. The usage it to put watermark on some document you send to official administration, adding for example: "INSURANCE CONTRACT" to your iban.  If people try to use this doc for anything else, then you can told them the doc was sent for a "insurance contract", nothing else.  This help against identity theft (especially for passport, IBAN etc..) . 
Of course this can be improved with some more options (like choosing style, color, size, size of the image watermarked, etc...), but the feature right now is enough for 90% of use cases.

bbhtt:
It's not enough for getting accepted here.

aginies:
version 3.1: 
- use python GTK3
- expert mode: font selection, font color choice or random, font transparency, text angle, density, optionnal Date + Hour, filename prefix, JPG compression level, Image resize, translation update
- Add filters on file selection to restrict access to images files
- flathub: Restrict filesystem to xdg-pictures
- bug fixes. 

screeenrecord: https://www.youtube.com/watch?v=rNg0RGUvESI

io.github.aginies.watermark.yml
aginies: This is now limited to xdg-pictures

io.github.aginies.watermark.yml
```diff
@@ -0,0 +1,24 @@
...
+runtime: 'org.freedesktop.Platform'
+runtime-version: '24.08'
+sdk: 'org.freedesktop.Sdk'
+command: watermark_app
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --device=dri
+  - --filesystem=xdg-pictures
+  - --share=network
```
bbhtt: what does it need network for?

io.github.aginies.watermark.yml
```diff
@@ -0,0 +1,24 @@
+app-id: io.github.aginies.watermark
+runtime: 'org.freedesktop.Platform'
+runtime-version: '24.08'
+sdk: 'org.freedesktop.Sdk'
+command: watermark_app
+finish-args:
+  - --share=ipc
+  - --socket=x11
```
bbhtt: if it is gtk3 this should be fallback-x11+wayland

io.github.aginies.watermark.yml
aginies: not needed, removed.

io.github.aginies.watermark.yml
aginies: done in latest PR. default wayland, fallback to x11.

aginies:
Thanks again for the review, I fixed latest issues.

io.github.aginies.watermark.yml
aginies: this should been addressed.

bbhtt:
bot, build

aginies:
Ok i will fix the missing metainfo.xml.

aginies:
I deleted the  wrong branch, sorry about that.

aginies:
I deleted the wrong branch, sorry about that.

bbhtt:
bot, build

bbhtt:
bot, build

bbhtt:
1. Drop the stock icon tag here https://github.com/aginies/watermark/blob/fa4cbbdacbfcac6e10ac5b05789ed76279703a31/io.github.aginies.watermark.metainfo.xml#L7
2. Add OARS tags https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines#open-age-ratings-service-oars

bbhtt:
bot, build

aginies:
for some strange reason the image screenshot doesnt appear in gnome-software. The image source are ok, something wrong in the metainfo.xml file?

https://github.com/aginies/watermark/edit/main/io.github.aginies.watermark.metainfo.xml

bbhtt:
I don't think it is working. I just tried "Test" as a watermark test, and it is adding some random characters and number.


https://github.com/user-attachments/assets/956b104f-8841-424b-b50a-ecb2b4361d97

bbhtt:
- Also for the filechooser try using gtkfilechoosernative https://lazka.github.io/pgi-docs/Gtk-3.0/classes/FileChooserNative.html that uses the portal and would allow to load images from anywhere.

- The default watermark is too light maybe you can make them slightly bolder?

> The image source are ok, something wrong in the metainfo.xml file?

Test builds don't have screenshots.

aginies:
> I don't think it is working. I just tried "Test" as a watermark test, and it is adding some random characters and number.
> 
>  Screencast.mp4

it add the test character + the date + hour as watermark on the image. This is expected. But in your case the text is not ok because the font can't be used. Try to choose another font, but the program shoudl test the default font is available and usable (Sans 20, /run/host/fonts/truetype/NotoSans-SemiBoldItalic.ttf)

aginies:
> * The default watermark is too light maybe you can make them slightly bolder?
> 
> > The image source are ok, something wrong in the metainfo.xml file?
> 
> Test builds don't have screenshots.

expert option, you can choose transparency, density etc... but i can get it more present by default.

aginies:
> * Also for the filechooser try using gtkfilechoosernative https://lazka.github.io/pgi-docs/Gtk-3.0/classes/FileChooserNative.html that uses the portal and would allow to load images from anywhere.

Will take a look.

> > The image source are ok, something wrong in the metainfo.xml file?
> Test builds don't have screenshots.

ah that's why :)

bbhtt:
> it add the test character

where are you seeing the test character? I'm seeing white boxes instead of it.


image

aginies:
> > it add the test character
> 
> where are you seeing the test character? I'm seeing white boxes instead of it.
> 
> image

Sounds like this could be a processing error of the font with PIL, or font missing (but the program should test this use case).

bbhtt:
bot, build

bbhtt:
it is still producing the white boxes with some fonts selected in the font chooser. what's the problem? because removing a default selection doesn't seem like a solution, it will break as soon as people choose a different font.

it works with liberation, dejavu among the ones i tested.

bbhtt:
And some fonts are non existent, something seems wrong with the way you are building the font list

image

aginies:
> it is still producing the white boxes with some fonts selected in the font chooser. what's the problem? because removing a default selection doesn't seem like a solution, it will break as soon as people choose a different font.
> it works with liberation, dejavu among the ones i tested.

I dont have any issue here with my fonts... and a I tried a lot. Getting the people selecting the font can test if the font path is present and registered correctly in their system. On my testing platform i Dont have any issue with the one selected by default "Sans 20". Even on windows this is ok with "Arial 20". If you select a font without an accessible font path then you can not use them with PIL to add some character (or there is maybe another way but with another py module). The problem with the font is that there is plenty of type, and plenty of encoding, and this is not possible to find the type and the encoding from the font itself, which leads to a situation where the font can not be used by PIL to add character on the image.

aginies:
> And some fonts are non existent, something seems wrong with the way you are building the font list
> 
> image

I am using GTkFontChooserDialog, if fc-list can't find the font selected I dont know how it was registered on the system.

aginies:
There is probably another way to get this fixed (finding fonts). About the "square" issue if you can give me the name of the font you are using i can try to reprodruce it here.
So its a problem with bitmap font VS scalable font, this is causing the main issue. So i need to restrict to trutype , or convert the bitmap one on the fly. will fix that

bbhtt:
The first one "Sans Regular" or something

aginies:
I fully rework the font selection code, only TTF font should be available now (no more bitmap one) as PIL is not capable of handling bitmap font. In the font chooser, if the preview show "squares" or "strange character", this will of course add square to the image and not the text. This is not possible to detect that the system can not handle that kind of font, as the system list them as usable. Its probably some specific font for some exotic country.

bbhtt:
bot, build

PR #6600, opened 2025-06-06
Add de.bschu.cashbox

bschubernd:
Yes my submission meets all the criteria

cashbox memorises the cost of articles and calculates the total price and change. It is intended for small clubs on a celebration, where members are not experienced in memorizing prices and doing mental arithmetic. 
- https://salsa.debian.org/api/v4/projects/debian%2Fcashbox/repository/files/screenshots%2Fpricelist.png/raw?ref=debian_version_0_3_4
- https://salsa.debian.org/api/v4/projects/debian%2Fcashbox/repository/files/screenshots%2Fsale.png/raw?ref=debian_version_0_3_4
- https://salsa.debian.org/api/v4/projects/debian%2Fcashbox/repository/files/screenshots%2Freceipt.png/raw?ref=debian_version_0_3_4
- [x] I am an author, developer and upstream contributor to the project.     

### DIFF
--- a/de.bschu.cashbox.yml
+++ b/de.bschu.cashbox.yml
@@ -14,9 +14,9 @@ modules:
   - name: cashbox
     buildsystem: simple
     build-commands:
-      - make install
+      - make prefix=$FLATPAK_DEST install
     sources:
       - type: git
         url: https://salsa.debian.org/debian/cashbox
-        commit: f2c5bec88d9c68ce1c253482cd732538eb4a9d56
-        tag: debian_version_0_3_4
+        commit: 99dfff2a0a901cdce894fa4c72b7489f32817bea
+        tag: debian_version_0_3_5


bschubernd:
if possible I would wish to have bbhtt as reviewer, because he/she has already reviewed my previous attempt.

hfiguiere:
You closed the previous PR removing all history.

de.bschu.cashbox.yml
```diff
@@ -0,0 +1,23 @@
...
+  - --share=ipc
+  - --socket=pulseaudio
+  - --filesystem=home
+  - --device=dri
+modules:
+  - name: cashbox
+    buildsystem: simple
+    build-commands:
+      - make install
+      - cp cashbox.py /app/bin/cashbox
```
hfiguiere: I don't know why you cp. The Makefile seems to have all that is needed. ~~Which I asked previously btw.~~

Edit: must be confusing with a different PR but given you closed the previous one....

bschubernd: I think the reason is, only to rename cashbox.py to cashbox (without .py).

bschubernd: But this is also done in the makefile. I will retest the flatpak without it. (Takes a while).

bschubernd: Yes you are right. I have deleted the cp. Thanks for the hint.

bschubernd:
The error was "install: invalid group 'flatbld'"
Maybe the group of the user flatbld is not flatbld but something else. (My Makefile assumes it is the same). If this is the case I can change it in the Makefile.
I will try to fix it in the Makefile. (takes a while)

bschubernd:
should work now, please try build again

bschubernd:
bot, build

bschubernd:
The test version seems to work. I have done a small test on librem5 (aarch64) and on my pc (x86_64).

bschubernd:
can the bot produce a final version (non-test)

hfiguiere:
there won't be any difference. it will happen when the repository is created (merging)

bbhtt:
> The error was "install: invalid group 'flatbld'"
Maybe the group of the user flatbld is not flatbld but something else. (My Makefile assumes it is the same). If this is the case I can change it in the Makefile.

What magic are you doing in makefile that requires username and group names? It shouldn't matter.

bbhtt:
>  Please attach or link a screenrecord showcasing the application on Linux using the Flatpak.

This said **screenrecord**, you linked a bunch of screenshots.

de.bschu.cashbox.yml
```diff
@@ -0,0 +1,22 @@
...
+  - --socket=fallback-x11
+  - --share=ipc
+  - --socket=pulseaudio
+  - --filesystem=home
+  - --device=dri
+modules:
+  - name: cashbox
+    buildsystem: simple
+    build-commands:
+      - make install
```
bbhtt: how does this work without a prefix set? Or did you hardcode `/app` somewhere?

de.bschu.cashbox.yml
```diff
@@ -0,0 +1,22 @@
...
+runtime: org.gnome.Platform
+runtime-version: '48'
+sdk: org.gnome.Sdk
+command: cashbox
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc
+  - --socket=pulseaudio
+  - --filesystem=home
```
bbhtt, line 11: why is home needed?

bschubernd:
> > The error was "install: invalid group 'flatbld'"
> > Maybe the group of the user flatbld is not flatbld but something else. (My Makefile assumes it is the same). If this is the case I can change it in the Makefile.
> 
> What magic are you doing in makefile that requires username and group names? It shouldn't matter.

I think the Makefile can not install files as root, when used from flatpak-builder. So I already added the line 
ROOT    = $(shell [ "$(FLATPAK_DEST)" ] && echo $(shell whoami) || echo root)
When your bot is running ROOT becomes "flatbld".
Now I also calculate the group of flatbld with the line in Makefile:
ROOTG   = $(shell [ "$(FLATPAK_DEST)" ] && echo $(shell id -gn) || echo root)
Now I can use "install -o $(ROOT) -g $(ROOT) ...."

Do you think it should work to install with "install -o root -g root ..." instead ?

hfiguiere:
> Do you think it should work to install with "install -o root -g root ..." instead ?

I think should never set `-o` and `-g` for install.  There is no reason to do so.

de.bschu.cashbox.yml
hfiguiere: It doesn't. It doesn't honour any sort of prefix either. The makefile check for `FLATPAK_DEST` and assume otherwise that it will install in `/usr`. It's over complicated for no reason.

bschubernd:
> > Please attach or link a screenrecord showcasing the application on Linux using the Flatpak.
> 
> This said **screenrecord**, you linked a bunch of screenshots.

https://www.bschu.de/html/readall/simplescreenrecorder-2025-06-06_13.16.20.mkv

bschubernd:
> s no reason to do so.

It is the same Makefile that creates the official debian package cashbox.deb.
And for this it is required to set the user and group and permissions. Many packages install binaries for different users, and even create this users in the postscript if needed.
In the case of flatpack it would be the same. But it is easier for me to calculate the current user and group and use it with -g and -o. If I would not do that, each install command would have to check in which environment it runs.

de.bschu.cashbox.yml
bschubernd: I have double checked the code. I do not use "/app" anywhere.
And if so, the debian version (running without flatpak) would fail.

But the Makefile checks for FLATPAK_DEST and uses this if it is set.
USR     = $(shell [ "$(FLATPAK_DEST)" ] && echo $(FLATPAK_DEST) || echo /usr)

de.bschu.cashbox.yml
bschubernd: home is needed, to give the user the chance to save and load his configuration in his home dir. 
(I have never checked without this option, because I think it will fail)

bbhtt:
That's incorrect you should not use particular users or groups. If someone wants to install as root they can use sudo make install.

And use PREFIX instead of FLATPAK_DEST or `/usr` fallback. It's the standard way.

de.bschu.cashbox.yml
hfiguiere: where is that configuration stored?

Don't say "home" because that's like telling "I live on planet Earth" if one ask where you live.

bschubernd: The user can decide where to store or load. You can see the selection box in the requested:
[ttps://www.bschu.de/html/readall/simplescreenrecorder-2025-06-06_13.16.20.mkv](https://www.bschu.de/html/readall/simplescreenrecorder-2025-06-06_13.16.20.mkv)

bschubernd:
> That's incorrect you should not use particular users or groups. If someone wants to install as root they can use sudo make install.
> 
> And use PREFIX instead of FLATPAK_DEST or `/usr` fallback. It's the standard way.

While building debian packages sudo is normally not available. 
Also I do not use particular users or groups. Instead I calculate the user or group, that would be used without option -u or -g and use them. So the result is the same as without the option -u or -g. But because if it is no flatpak I have to use -u and -g it is easier to do it as I do it.

I think if I use FLATPAK_DEST, every reader can assume that this ist flatpak specific. 
So I like it more than using PREFIX. Because PREFIX could also be used  (maybe in future) when building debian packages.

bbhtt:
Unless it gets fixed don't request a review.

bschubernd:
No Bug -> No Fix -> No Flatpak

de.bschu.cashbox.yml
```diff
@@ -18,5 +18,5 @@ modules:
     sources:
       - type: git
         url: https://salsa.debian.org/debian/cashbox
-        commit: 4e5c9b2ded413c627c007ea38b01fa782dcea093
+        commit: f2c5bec88d9c68ce1c253482cd732538eb4a9d56
         tag: debian_version_0_3_4
```
hfiguiere: retagging is bad.

bschubernd: retagging is bad after releasing, before that it is needed for testing.

for me debian/rules the world

And the debian release of cashbox 0.3.4 was Thu, 05 Jun 2025 13:41:52 +0200 (https://packages.debian.org/sid/cashbox  see Debian-Changelog).  The retagging of the source to 0.3.4 was before that.

Debian makes sure, that it is now only possible to release newer cashbox versions as 0.3.4.
Also my Makefile make sure, that the version number displayed in help/about of the running binary is the same version as the debian package.

bschubernd: It is a big success, that now exactly the same source can be used
(1) to build the debian version with all the many requirements and checks (like install with given user and group root) and
(2) that the same source can be used to build the flatpak.

Also the built flatpak version can handle any PREFIX. So if flatpak decides to change /app to something else, the build flatpak version of cashbox will run and for example find the default cashbox.css file. (if you want you can probably build a flatpak version with another PREFIX to verify that.)

So technically everything is ok from my point of view.

bschubernd:
Hi @bbhtt and @hfiguiere,
Thank you for reviewing. Could you please help with the issue described below:

###   current status

- The cashbox flatpak can be automatically built by your bot and I have tested it successfully on my pc and on my librem5 smartphone. I have provided a video link showing the cashbox flatpak running on my pc as requested.
- All your reviewer comments are implemented or answered and hence all reviews are marked solved. But there are labels “awaiting-changes” and “awaiting-upstream”. I do not understand why.
- As I understood you only have to remove the labels “awaiting-changes” and “awaiting-upstream”.

###    speculations
I do not understand what changes are awaited, or what I as the upstream author are expected to do. So I can only speculate:

- Does the provided video link not show what you want to see? Tell me what you want to see?

- Do you mislike, that the same sources are used to build the flatpak and the debian package? Do you want me to write a separate Makefile for the flatpak, where I could for example run install without options -g and -u?
- Is the Makefile to complicated (“wc -l Makefile” shows 98 lines)?
- Is one of my explanations unclear? If so, please unresolve the relevant review and write what you need. I could explain it in more detail or for example proof it by linking documentation to requirements when building debian packages. You can also try yourself to build the debian package with apt-get source and debuild. I could send you the needed config file and the exact commands to download (apt-get source) the sources from debian.org and build (debuild) the debian package.
- Is there a political problem? On https://docs.flathub.org/docs/for-app-authors/submission you write “App submissions are extremely welcome …”. But maybe what cashbox is doing, is unwanted? why? Or should a package provided as flatpak not in parallel exist as debian package?
- Do you have any other suggestions how to solve the problem?
- Is there a way to escalate this problem, so that somebody else unbiased can review it?

bbhtt:
There is no problem as I said the Makefile needs to be fixed first. Use prefix, drop setting user and group.

bbhtt:
Also we aren't corporations, if it is not good enough during review then it is going to sit. There's nowhere to escalate.

bschubernd:
@bbhtt ok boss I will do as you say. To make it as you want I must ask the following questions:

- Can I make a separate Makefile.flatpak which will be much shorter and only will do what you want to build the flatpak, that I will call from de.bschu.cashbox.yml like below?
```
    build-commands:
      - make -f Makefile.flatpak PREFIX=/app install
```
- I can also instead of PREFIX use prefix. You have mentioned both. Which do you prefer ?
- Is it true that I have to hardcode PREFIX=/app in the de.bschu.cashbox.yml manually? (I did find only examples that do set it manually (https://docs.flatpak.org/en/latest/manifests.html), instead of using the automatically by flatpak-builder set variable FLATPAK_DEST?
-  I will only call install in Makefile.flatpak without -g and -u but I need option -m. Is option -m ok?

bbhtt:
None of the changes I said requires maintaining a separate makefile. Prefix is the standard way to set the install root https://www.gnu.org/prep/standards/html_node/Directory-Variables.html#index-prefix instead you are defaulting to a Flatpak specific variable.

So if someone wants to install too `/sysroot/usr` then have to use FLATPAK_DEST which is weird outside Flatpak.

bbhtt:
bot, build

bbhtt:
why do you keep retagging? that's another thing that got justified by debian something before but it's wrong.

hfiguiere:
> why do you keep retagging? that's another thing that got justified by debian something before but it's wrong.

even Debian will tell you it's wrong.

bschubernd:
I do not plan to do any retagging.

bbhtt:
bot, build

bbhtt:
there is no icon here

image


also why are load, save etc. defaulting to XDG_DATA_HOME?

bschubernd:
Hi @bbhtt I see the picture of cashbox in your last post. Great !!!. So I assume you have also tried it. 
I hope it was easy to use and easy to understand. And I hope there is no unneeded information on the screen. This is very important for a small screen on smartphones like the librem5. In a real life situation, when you work for your club and sell drinks and food it must be easy to use. And it is important that you can instantly start. That means sometimes you have no time to enter prices before. So it must be possible to enter the prices while you sell things. Maybe you have time later, then you probably can enter all prices. And also you can save the list and give it to the next group of volunteers as staring point to continue to sell.

If you have any comments to this kind of question, I would be happy. And probably I could improve cashbox next time with your ideas for the users.

But I do not want to explain why I have used this or another variable, or why somewhere is no icon. This is what I will decide alone. This is my freedom of programming that I enjoy. Probably the most important point for me is freedom.

You can now decide if you want cashbox on flathub.

But if not then this is probably my last comment. I have already done things to the source that I do not like. For example to remove options -o and -g from install which only works, because this debian package installs all files as root. Maybe I will repair that if it is not needed to bring casbox to flathub.

So if you want cashbox on flathub you can stop to block the merging. If not this is your freedom.

bbhtt:
What are you saying? I reported two issues that need looking at by you

PR #6611, opened 2025-06-10
Add io.github.Faugus.faugus-launcher

Faugus:

- [X] Please describe the application briefly. A simple and lightweight app for running Windows games using UMU-Launcher

https://github.com/user-attachments/assets/16d8b67d-94b5-4aa8-a7b3-de83548b9868

- [X] I am the author and developer of the project.


modules/intltool-0.51.json
hfiguiere, line 1: this should be the shared-modules

modules/imagemagick.json
hfiguiere, line 1: this should be inline in the manifest: it's not autogenerated

modules/imagemagick.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "imagemagick",
+    "buildsystem": "simple",
```
hfiguiere: this should be buildsystem `autotools`, not `simple`

io.github.Faugus.faugus-launcher.yml
```diff
@@ -0,0 +1,130 @@
...
+    version: *runtime-version
+    versions: *runtime-version
+    autodelete: false
+    no-autodownload: true
+    add-ld-path: lib
+    download-if: have-intel-gpu
+    autoprune-unless: have-intel-gpu
+
+modules:
+  - name: metadata
```
hfiguiere: this should be merged with the "main" module

io.github.Faugus.faugus-launcher.yml
```diff
@@ -0,0 +1,130 @@
...
+        contents: |
+          /app/lib32
+          /app/lib/i386-linux-gnu
+
+  - name: faugus-launcher
+    buildsystem: simple
+    build-commands:
+      - chmod +x flatpak-patch.sh
+      - ./flatpak-patch.sh
+      - 'install -D faugus_launcher.py /app/bin/faugus-launcher'
```
hfiguiere: why is there no installation upstream?

io.github.Faugus.faugus-launcher.metainfo.xml
hfiguiere, line 1: this file should be (submitted and managed) upstream

io.github.Faugus.faugus-launcher.metainfo.xml
```diff
@@ -0,0 +1,44 @@
...
+  <url type="homepage">https://github.com/Faugus/faugus-launcher</url>
+  <url type="bugtracker">https://github.com/Faugus/faugus-launcher/issues</url>
+  <project_license>MIT</project_license>
+  <metadata_license>CC0-1.0</metadata_license>
+
+  <launchable type="desktop-id">io.github.Faugus.faugus-launcher.desktop</launchable>
+
+  <screenshots>
+    <screenshot type="default">
+      <image>https://raw.githubusercontent.com/Faugus/faugus-launcher/refs/heads/main/screenshots/main-list.png</image>
```
hfiguiere: screenshot URL should NEVER reference the tip of a branch.

modules/libayatana-appindicator-gtk3.json
hfiguiere, line 1: this also should be from the shared-modules

modules/libcanberra.json
hfiguiere, line 1: this also should be from the shared-modules

io.github.Faugus.faugus-launcher.yml
```diff
@@ -0,0 +1,130 @@
...
+        dest-filename: ld.so.conf
+        contents: |
+          /app/lib32
+          /app/lib/i386-linux-gnu
+
+  - name: faugus-launcher
+    buildsystem: simple
+    build-commands:
+      - chmod +x flatpak-patch.sh
+      - ./flatpak-patch.sh
```
hfiguiere: this look like a problem is search of a solution.

io.github.Faugus.faugus-launcher.yml
```diff
@@ -0,0 +1,130 @@
...
+            full_path="$(pwd)/${mo_path#./}"
+            lang=$(basename $(dirname "${full_path}"))
+            install -Dm644 "${full_path}" "/app/share/locale/${lang}/LC_MESSAGES/$(basename "${mo_path}")"
+          done
+        ' sh {} +
+      - 'chmod +x /app/bin/faugus-launcher'
+      - 'chmod +x /app/bin/faugus-run'
+      - 'chmod +x /app/bin/faugus-proton-manager'
+      - 'chmod +x /app/bin/faugus-components'
+      - 'chmod +x /app/bin/umu-run'
```
hfiguiere: `install` can set the mode properly making these 5 line unnecessary.

modules/libcanberra.json
Faugus: Thanks for the review.
I'm sorry for the many mistakes, this is my first time doing this.

How do I apply a patch to a shared-module?
I need to remove `/bin` from the `cleanup`, otherwise `canberra-gtk-play` won't work.

https://github.com/flathub/shared-modules/blob/a34a31e4dd50a0b38d1725f6881d1bedb0948b38/libcanberra/libcanberra.json#L1-L7

Faugus:
About the `finish-args` that the bot complained:
- If I don't set `filesystem=xdg-config/gtk-3.0:ro` the app won't use the system theme
- If I don't set `filesystem=xdg-data/applications:create` the app can't create a shortcut if the `~/.local/share/applications` doesn't exist, even the app being set to create if the folder doesn't exist (it works fine outside Flatpak)

bbhtt:
Using xdg-config/gtk-3.0 for themes is a broken behaviour. You should read the Flatpak docs on themes.

bbhtt:
Also the dynamic launcher portal exists for writing desktop files.

Faugus:
> You should read the Flatpak docs on themes.

I did. It says to use the theme provided by Flathub because it can't read theme in `/usr/share/themes`.
When I install my app, it downloads the correct theme, but it doesn't use it.

Faugus:
All changes are done.

io.github.Faugus.faugus-launcher.yml
```diff
@@ -0,0 +1,151 @@
+app-id: io.github.Faugus.faugus-launcher
+sdk: org.gnome.Sdk
+runtime: org.gnome.Platform
+runtime-version: &runtime-version '48'
+x-gl-version: &gl-version '1.4'
+x-gl-versions: &gl-versions 48;48-extra;1.4
```
bbhtt:
```suggestion
x-gl-versions: &gl-versions 24.08;24.08extra;1.4
```

GL version comes from the parent of gnome runtime -  fdsdk. usually it is `24.08-extra` but only for 24.08 it doesn't have the dash.

when you change to 25.08 use `25.08-extra`

io.github.Faugus.faugus-launcher.yml
```diff
@@ -0,0 +1,151 @@
...
+          /app/lib/i386-linux-gnu
+
+  - name: faugus-launcher
+    buildsystem: simple
+    build-commands:
+      - 'install -Dm644 flatpak/io.github.Faugus.faugus-launcher.metainfo.xml /app/share/metainfo/io.github.Faugus.faugus-launcher.metainfo.xml'
+      - 'install -Dm755 faugus_launcher.py /app/bin/faugus-launcher'
+      - 'install -Dm755 faugus_run.py /app/bin/faugus-run'
+      - 'install -Dm755 faugus_proton_manager.py /app/bin/faugus-proton-manager'
+      - 'install -Dm755 faugus_components.py /app/bin/faugus-components'
```
bbhtt: this badly needs a proper buildsystem...

io.github.Faugus.faugus-launcher.yml
```diff
@@ -0,0 +1,151 @@
...
+        url: https://github.com/Faugus/components/releases/download/v1.0.0/umu-run
+        sha256: 3069d5045ca356779a5ece30a54dfada0106954404d417a7d219633f533ac1ce
+
+    modules:
+      - name: imagemagick
+        buildsystem: autotools
+        build-commands:
+          - ./configure --prefix=/app --enable-shared --disable-static --disable-docs
+          - make
+          - make install
```
bbhtt: only simple can use manual build commands use config-opts.

io.github.Faugus.faugus-launcher.yml
```diff
@@ -0,0 +1,151 @@
...
+        sources:
+          - type: git
+            url: https://github.com/ImageMagick/ImageMagick.git
+            tag: 7.1.1-39
+
+      - name: libcanberra # patched, otherwise canberra-gtk-play doesn't work
```
bbhtt: what is the issue here?

Faugus: @bbhtt Using `libcanberra` from shared-modules, `canberra-gtk-play` doesn't work.
I need to remove `/bin` from the `cleanup` to make it work.

https://github.com/flathub/shared-modules/blob/a34a31e4dd50a0b38d1725f6881d1bedb0948b38/libcanberra/libcanberra.json#L1-L7

bbhtt: I dropped that cleanup, update the submodule.

Faugus: @bbhtt I updated it. It's trying to patch the file that was removed.
```
Downloading http://0pointer.de/lennart/projects/libcanberra/libcanberra-0.30.tar.xz
100  311k  100  311k    0     0   217k      0  0:00:01  0:00:01 --:--:--  217k
Failed to download sources: module libcanberra: Can't find file at 0001-gtk-Don-t-assume-all-GdkDisplays-are-GdkX11Displays-.patch
```

bbhtt: the patch file is there https://github.com/flathub/shared-modules/blob/master/libcanberra/0001-gtk-Don-t-assume-all-GdkDisplays-are-GdkX11Displays-.patch

Faugus:
Done

io.github.Faugus.faugus-launcher.yml
hfiguiere: it's probably a misinterpretation of https://github.com/flathub/flathub/pull/6611#discussion_r2138574027

io.github.Faugus.faugus-launcher.yml
```diff
@@ -0,0 +1,124 @@
...
+  - --device=all
+  - --allow=multiarch  # pressure-vessel
+  - --socket=pulseaudio # notification sound
+  - --talk-name=org.kde.StatusNotifierWatcher # system tray
+  - --env=PATH=/app/bin:/app/utils/bin:/usr/bin:/usr/lib/extensions/vulkan/MangoHud/bin:/usr/lib/extensions/vulkan/gamescope/bin:/usr/lib/extensions/vulkan/OBSVkCapture/bin # mangohud, gamescope, obsvkcapture
+  - --filesystem=home
+  - --filesystem=/mnt
+  - --filesystem=/run/media
+
+inherit-extensions:
```
user1-github, line 27: I know I'm probably not an authority to make suggestions here, but I noticed a couple of things that are worth mentioning, so I'll make some anyway.

Since you're using org.winehq.Wine as a base app, I also suggest adding org.winehq.Wine.gecko and org.winehq.Wine.mono under "inherit-extensions:" . This way you'll have only one installation of Wine gecko and mono that all Wine prefixes will use. Without these extensions Wine will install gecko and mono into the prefix itself, which means you'll have duplicate installations of gecko and mono if you're using multiple Wine prefixes.

io.github.Faugus.faugus-launcher.yml
```diff
@@ -0,0 +1,124 @@
...
+command: faugus-launcher
+separate-locales: false # translations
+base: org.winehq.Wine
+base-version: stable-24.08
+
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
```
user1-github, line 17: I also think --socket=fallback-x11 needs to be replaced with --socket=x11 because --socket=fallback-x11 means Wayland is going to be preferred over x11 and Wine's / Proton's Wayland driver is still pretty rough.

io.github.Faugus.faugus-launcher.yml
Faugus: Thanks!
I'm only using `org.winehq.Wine` because that was the only way it could inherit the `org.freedesktop.Platform.GL32`, which is needed for 32 bit .exe.
My app doesn't use Wine.

io.github.Faugus.faugus-launcher.yml
Faugus: That's what I did, but the bot complained for having both `--socket=x11` and `--socket=wayland`

io.github.Faugus.faugus-launcher.yml
user1-github: Oh, you mean it's not available if you try to choose it from the "Runner" drop down menu? If not, I think you can add the org.freedesktop.Platform.GL32 under "add-extensions:" instead of "inherit-extensions:" and then you won't need org.winehq.Wine.
[Here is an example of it](https://github.com/flathub/org.winehq.Wine/blob/branch/stable-24.08/org.winehq.Wine.yml#L43) in a manifest

Faugus: Exactly. It only uses Proton with umu-launcher.

Faugus: I tried that before, but the 32 bit .exe doesn't work.

io.github.Faugus.faugus-launcher.yml
bbhtt: you should only use `--socket=x11`

io.github.Faugus.faugus-launcher.yml
user1-github: If so, then I guess org.winehq.Wine is needed for something after all.

io.github.Faugus.faugus-launcher.yml
Faugus: @bbhtt after I set only `--socket=x11` my app started using the system theme. Why? (I'm on Wayland)

Faugus: @bbhtt without the `--socket=wayland` the wine wayland driver doesn't work.
But if I put `--socket=wayland` the app doesn't use the system theme.
Do you know any solution for this?

user1-github: Leave only `--socket=x11` in the manifest. For your own usage you can then override the default permissions of your app. If you also need Wayland access, type:
`flatpak override --user --socket=wayland io.github.Faugus.faugus-launcher`

Faugus: But the wine wayland driver is a feature of my app. So I need the `--socket=wayland`

user1-github: In that case I guess the only solution is to leave the permissions as is.

I gave this suggestion because at least from my experience the Wayland driver is still rough - some games crash when attempting to launch them, in others keyboard input doesn't work, other apps have broken unmovable windows, etc...
But that's on Wine 10.0. Maybe your app ships Proton version that are based on newer Wine versions where things are more stable with the Wayland driver?

Faugus:
Everything should be fine now

bbhtt:
bot, build

Faugus:
The log said it had problems installing `org.winehq.Wine`.
I managed to make it work without it.

Faugus:
bot, build

Faugus:
Without inheriting the extensions `org.freedesktop.Platform.GL32` and `org.freedesktop.Platform.Compat.i386` they are not automatically installed for the user.
They are necessary for the functioning of 32 bit games and launchers.

Faugus:
bot, build

Faugus:
It won't build when having `org.winehq.Wine` as base.
What should I do?

hfiguiere:
there is no wine for aarch64.

Faugus:
@hfiguiere I was taking a look at Lutris' manifest and it uses wine as base.
https://github.com/flathub/net.lutris.Lutris/blob/beta/net.lutris.Lutris.yml

hfiguiere:
and it doesn't build aarch64 either.

Faugus:
@hfiguiere how do I force it to build x86_64 only?

Faugus:
bot, build

Faugus:
Even inheriting `org.freedesktop.Platform.GL32` and `org.freedesktop.Platform.Compat.i386` they are not being installed with the app. Why?

io.github.Faugus.faugus-launcher.yml
```diff
@@ -0,0 +1,99 @@
+app-id: io.github.Faugus.faugus-launcher
+sdk: org.freedesktop.Sdk
+runtime: org.freedesktop.Platform
+runtime-version: &runtime-version '24.08'
+x-gl-version: &gl-version '1.4'
+x-gl-versions: &gl-versions 24.08;1.4
```
bbhtt:
```suggestion
x-gl-versions: &gl-versions 24.08;24.08extra;1.4
```

Faugus:
bot, build

Faugus:
Still won't install the extensions

bbhtt:
it will get installed once the app is on flathub.

Faugus:
Oh ok. I think that's it. Thank you very much for the reviews and the patience.

bbhtt:
the tray has no icon, make sure you are setting an icon that's named by appid and it is installed to where the desktopfile icon is installed


https://github.com/user-attachments/assets/877f682d-e148-48f5-8eea-709d5a309a4e

bbhtt:
looks like you are using the full path of the icon, try using just the name. the path is inside the sandbox and unavailable on host.

if that doesn't work you have to calculate the path on host.

bbhtt:
also there is no need to set different icon names for inside flatpak vs. outside. if `io.github.Faugus.faugus-launcher` is your chosen ID then everything follows that it doesn't matter if it is running in flatpak or packaged on distro.

Faugus:
It works for me
image

Faugus:
Tried on Ubuntu and it's really not showing. But why does it work on KDE?

Faugus:
bot, build

Faugus:
Should work now

bbhtt:
still not working, you did the opposite of what I said.

Faugus:
That was the only way it worked on Ubuntu.
I tried only the app id `io.github.Faugus.faugus-launcher` and it didn't work.
image

Faugus:
Testing on Linux Mint and it really doesn't work.
Tried every path possible and nothing.
All other icons in the app works, except the appindicator.
I have no idea why.

Faugus:
@bbhtt For some reason, on Linux Mint (Flatpak only) any of the paths work.
This must be a bug with Flatpak/Appindicator.

https://github.com/user-attachments/assets/bda93c33-ce69-4a63-ba25-13807764e0c5

bbhtt:
There is no bug in Flatpak

Faugus:
@bbhtt I have no idea what could be then.
The problem isn't the path, like I showed in the video.
I use the same path for other icons in the app.

It works on:
Arch KDE X11/Wayland
Fedora KDE X11/Wayland
Ubuntu X11/Wayland

But for some reason it doesn't work on Mint.
I thought Flatpak should work the same in all distros.

Faugus:
bot, build

Faugus:
@bbhtt Could you give it a try please? It should work now.

bbhtt:
Yes it works now. That's what I said to do.

Faugus:
I misunderstood. I was trying with the appid + file format.

PR #6631, opened 2025-06-17
Add io.github.cosmic_utils.minimon-applet

CarsonBurke:

An applet designed for Cosmic Desktop (but functional as an application on other DEs like Gnome and KDE Plasma) to showcase system resource utilization such as CPU, memory, VRAM and network usage.

image
*The Flatpak applet running in KDE Plasma as an app. On Cosmic, it sits in the panel by default*

### DIFF
--- a/io.github.cosmicUtils.cosmicAppletMinimon.json
+++ b/io.github.cosmicUtils.cosmicAppletMinimon.json
@@ -36,7 +36,7 @@
         {
            "type": "git",
            "url": "https://github.com/cosmic-utils/minimon-applet.git",
-           "commit": "0c30d29429cf0f0c49644893f0ffc7588c6c3dd8"
+           "commit": "e32508413410e79876a038fe58065820fa3ef4b2"
          },
         "cargo-sources.json"
       ]
--- a/io.github.cosmicUtils.cosmicAppletMinimon.json
+++ b/io.github.cosmicUtils.cosmicAppletMinimon.json
@@ -9,6 +9,7 @@
   "command": "cosmic-applet-minimon",
   "finish-args": [
     "--socket=wayland",
+    "--socket=fallback-x11",
     "--filesystem=xdg-config/cosmic:ro",
     "--talk-name=com.system76.CosmicSettingsDaemon",
     "--filesystem=/sys/class/hwmon:ro",
--- a/io.github.cosmicUtils.cosmicAppletMinimon.json
+++ b/io.github.cosmicUtils.cosmicAppletMinimon.json
@@ -1,0 +1,66 @@
+{
+  "id": "io.github.cosmicUtils.cosmicAppletMinimon",
+  "runtime": "org.freedesktop.Platform",
+  "runtime-version": "24.08",
+  "sdk": "org.freedesktop.Sdk",
+  "sdk-extensions": ["org.freedesktop.Sdk.Extension.rust-stable"],
+  "command": "cosmic-applet-minimon",
+  "finish-args": [
+    "--socket=wayland",
+    "--filesystem=xdg-config/cosmic:ro",
+    "--talk-name=com.system76.CosmicSettingsDaemon",
+    "--filesystem=/sys/class/hwmon:ro",
+    "--filesystem=/sys/class/drm:ro",
+    "--filesystem=/sys/devices:ro",
+    "--device=all"
+  ],
+  "build-options": {
+    "append-path": "/usr/lib/sdk/rust-stable/bin",
+    "build-args": ["--share=network"],
+    "env": {
+      "CARGO_HOME": "/run/build/cosmic-applet-minimon/cargo"
+    }
+  },
+  "modules": [
+    {
+      "name": "just",
+      "buildsystem": "simple",
+      "build-commands": ["install -Dm0755 just /app/bin/just"],
+      "sources": [
+        {
+          "type": "archive",
+          "strip-components": 0,
+          "dest-filename": "just.tar.gz",
+          "url": "https://github.com/casey/just/releases/download/1.25.2/just-1.25.2-x86_64-unknown-linux-musl.tar.gz",
+          "sha256": "ad56bf1a804243b51ef2c2eec4b3a8177f970ccc414cbbee219c2e5c06f7bcc9",
+          "only_arches": ["x86_64"]
+        },
+        {
+          "type": "archive",
+          "strip-components": 0,
+          "dest-filename": "just.tar.gz",
+          "url": "https://github.com/casey/just/releases/download/1.25.2/just-1.25.2-aarch64-unknown-linux-musl.tar.gz",
+          "sha256": "049d624255d150f18b0f3a1257a163366b0dc22e16526214773903125145e515",
+          "only_arches": ["aarch64"]
+        }
+      ]
+    },
+    {
+      "name": "cosmic-applet-minimon",
+      "buildsystem": "simple",
+      "build-commands": [
+        "just vendor",
+        "just build-vendored",
+        "just flatpak-install"
+      ],
+      "sources": [
+        {
+           "type": "git",
+           "url": "https://github.com/cosmic-utils/minimon-applet.git",
+           "commit": "0c30d29429cf0f0c49644893f0ffc7588c6c3dd8"
+         },
+        "cargo-sources.json"
+      ]
+    }
+  ]
+}
--- a/io.github.cosmicUtils.cosmicAppletMinimon.json
+++ b/io.github.cosmicUtils.cosmicAppletMinimon.json
@@ -22,35 +22,15 @@
   },
   "modules": [
     {
-      "name": "just",
-      "buildsystem": "simple",
-      "build-commands": ["install -Dm0755 just /app/bin/just"],
-      "sources": [
-        {
-          "type": "archive",
-          "strip-components": 0,
-          "dest-filename": "just.tar.gz",
-          "url": "https://github.com/casey/just/releases/download/1.25.2/just-1.25.2-x86_64-unknown-linux-musl.tar.gz",
-          "sha256": "ad56bf1a804243b51ef2c2eec4b3a8177f970ccc414cbbee219c2e5c06f7bcc9",
-          "only_arches": ["x86_64"]
-        },
-        {
-          "type": "archive",
-          "strip-components": 0,
-          "dest-filename": "just.tar.gz",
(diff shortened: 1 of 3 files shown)


io.github.cosmicUtils.cosmicAppletMinimon.json
```diff
@@ -0,0 +1,45 @@
+{
+  "id": "io.github.cosmicUtils.cosmicAppletMinimon",
+  "runtime": "org.freedesktop.Platform",
+  "runtime-version": "24.08",
+  "sdk": "org.freedesktop.Sdk",
```
bbhtt: use the cosmic baseapp.

io.github.cosmicUtils.cosmicAppletMinimon.json
```diff
@@ -0,0 +1,45 @@
...
+  "sdk-extensions": ["org.freedesktop.Sdk.Extension.rust-stable"],
+  "command": "cosmic-applet-minimon",
+  "finish-args": [
+    "--socket=wayland",
```
bbhtt: add fallback-x11

io.github.cosmicUtils.cosmicAppletMinimon.json
```diff
@@ -0,0 +1,45 @@
...
+    "--filesystem=xdg-config/cosmic:ro",
+    "--talk-name=com.system76.CosmicSettingsDaemon",
+    "--filesystem=/sys/class/hwmon:ro",
+    "--filesystem=/sys/class/drm:ro",
+    "--filesystem=/sys/devices:ro",
+    "--device=all"
```
bbhtt: why does it need device=all?

bbhtt:
is this an applet or an actual app with a tray applet?

applets are blocked by the inclusion policy https://docs.flathub.org/docs/for-app-authors/requirements#inclusion-policy

CarsonBurke:
>is this an applet or an actual app with a tray applet?

This is an applet that does work as an app, but is largely intended to be used as a tray applet. System76 developers reached out to me asking for this to be served as an applet, and they seem confident that this falls within the requirements - if that's worth anything.

io.github.cosmicUtils.cosmicAppletMinimon.json
CarsonBurke: Should be resolved now

io.github.cosmicUtils.cosmicAppletMinimon.json
CarsonBurke: Should be resolved now

io.github.cosmicUtils.cosmicAppletMinimon.json
```diff
@@ -0,0 +1,48 @@
+{
+  "id": "io.github.cosmicUtils.cosmicAppletMinimon",
```
bbhtt: the id does not follow the requirements

bbhtt:
> This is an applet that does work as an app

what do you mean by "works as an app"? This is why the PR checklist says to add a screenrecord, you added a screenshot.

CarsonBurke:
Ah apologies, I read that as "screenshot or screenrecord". I'll provide one in a sec.

CarsonBurke:
The applet being added to the panel on Cosmic Desktop
https://github.com/user-attachments/assets/1bdbe976-ea01-4331-b0ac-c930ec440326

The applet being launched and used like a window on KDE Plasma

https://github.com/user-attachments/assets/3b49626a-a39f-440d-b6a3-5aad5867ee98

Seems like it's not uploading those videos properly. I'll upload them elsewhere and link them.

bbhtt:
Ok, can you fix the last two comments? so that it builds and I can check it.

io.github.cosmicUtils.cosmicAppletMinimon.json
bbhtt: It should be `io.github.cosmic_utils.minimon-applet`, you have to update it in other places it is used.

io.github.cosmicUtils.cosmicAppletMinimon.json
CarsonBurke: I believe `device=all` is for reading some hardware information, probably for GPU usage. I think we tested and it didn't work without it. I can do another test to confirm if it is a blocker?

io.github.cosmicUtils.cosmicAppletMinimon.json
CarsonBurke: Got it, thanks. I'll do that.

CarsonBurke: Will be fixed/ready when [this](https://github.com/cosmic-utils/minimon-applet/pull/59) is merged. I'll let you know.

io.github.cosmicUtils.cosmicAppletMinimon.json
CarsonBurke: I'm not noticing any issues so I've removed it.

jackpot51:
@bbhtt is there background on the no extensions or tray applets rule? I have been unable to find a discussion and I want to know more.

bbhtt:
The reasoning is that we want user facing desktop apps in most cases, those aren't that and eg. a gnome shell extension zip in flatpak cannot be used

Tray applets etc. in most cases need to autostart/run as daemon etc. which also flatpak is not well suited for.

jackpot51:
I think we, the COSMIC team, will talk internally about how we want applets to be made available. They are not designed for other desktops, so it does not seem they would abide by the policy.

CarsonBurke:
Regardless of the Cosmic team's decision, I believe this applet can be published on Flathub as it's usable and useful on other desktop environments - it is decisively more than just a tray applet. I personally use it on KDE Plasma.

It functions like a normal app and does not need to autostart/run as a daemon.

bbhtt:
bot, build

io.github.cosmic_utils.minimon-applet.json
```diff
@@ -0,0 +1,47 @@
...
+  "id": "io.github.cosmic_utils.minimon-applet",
+  "runtime": "org.freedesktop.Platform",
+  "runtime-version": "24.08",
+  "sdk": "org.freedesktop.Sdk",
+  "sdk-extensions": ["org.freedesktop.Sdk.Extension.rust-stable"],
+  "base": "com.system76.Cosmic.BaseApp",
+  "base-version": "stable",
+  "command": "cosmic-applet-minimon",
+  "finish-args": [
+    "--socket=wayland",
```
bbhtt, line 11: Add --share=ipc

bbhtt:
the id for the exception needs to be updated.

CarsonBurke:
Ah oops, [pr](https://github.com/flathub-infra/flatpak-builder-lint/pull/667)

gmanskibiditoilet:
keep going, we have a pop os laptop and we want this applet as well!

bbhtt:
bot, build

bbhtt:
Why does the desktop file have nodisplay set? How are users expected to start it? (Going to the terminal is not an option for the non technical users)

git-f0x:
COSMIC applets are usually started by `cosmic-panel` (which can run on most DEs/WMs), and it's configured by config files or COSMIC Settings.
Since they generally aren't meant to be used without the panel, they have NoDisplay set to true.

bbhtt:
In this case it is supposed to be started independently on different environments according to the video showing KDE above.

git-f0x:
Yep, all applets can be launched independently (they're all just regular apps built in a certain way), it's just usually not an optimal experience (no window decorations, so they behave more like desktop widgets). Removing NoDisplay would make it suboptimal on COSMIC (or if cosmic-panel is used), but maybe a second desktop entry could work (that doesn't have NoDisplay, but has `NotShowIn=COSMIC`)?
In case it isn't appropriate for Flathub, a [COSMIC Flatpak](https://github.com/pop-os/cosmic-flatpak) repo has been set up.

bbhtt:
You can use NotShowIn in this one, why is a second desktop file needed?

The main desktop file matching the app id exactly needs to be visible.

bbhtt:
Feels like this entire architectured to be an applet and would get blocked by the rules. Maybe it works as an app... If you don't want to change that's ok. I am trying to get a build working to say the final verdict.

git-f0x:
>You can use NotShowIn in this one, why is a second desktop file needed?

Ah, right. I think COSMIC Settings doesn't care about NotShowIn, since it probably just searches for the applet key in the desktop file.

CarsonBurke:
I appreciate you giving this credence, I do want it to work as an app as well. Sorry for the NoDisplay thing - I'm not very familiar with specifications for .desktop files. I've made a [pr](https://github.com/cosmic-utils/minimon-applet/pull/62) to change it to `NotShowIn=COSMIC`

CarsonBurke:
Pr merged

bbhtt:
bot, build

bbhtt:
You can start the build when it is fixed.

gmanskibiditoilet:
almost there!

CarsonBurke:
Oops, didn't update the json after the pr merge mb.

CarsonBurke:
bot, build

CarsonBurke:
Turns out `devices=all` is needed for nvml (nvidia gpu reader library). Is that ok to add back in; does it need an exception?

bbhtt:
No you don't.

I tested it, but I don't think it works as an app. There seems to be no way to drag the panel thing or close it, which makes sense for an applet but not an app.

I'm also wondering what would be the use case for this as an app. I don't think people would want a floating mini window on the desktop work area overlapping other apps or going under them. Usually they go to the shell panel or taskbar etc.

CarsonBurke:
That's fair. I personally use it because I have many monitors and like the interface. But your argument makes sense. If you don't think it fits I don't have a strong counter-argument to give.

CarsonBurke:
The strongest thing I can provide is that with a few code additions, it can function similar to [discord overlay](https://flathub.org/apps/io.github.trigg.discover_overlay) as a non-interactible float-above overlay. I think it becomes quite usable on other DEs in that case, but 1. we don't have that functionality yet and 2. it's hard to say why you'd prefer that over something in the panel/dock.

git-f0x:
It would probably be best to have things like this in the COSMIC flatpak repo, since it isn't strictly an app (and works differently than other things on Flathub), and you can just add the repo and use the applet wherever you want.

mmstick:
Repository here: https://github.com/pop-os/cosmic-flatpak

CarsonBurke:
Will close this then, looks like cosmic has situated itself with their alternate Flatpak hosting solution. Thanks for your time bbhtt :)

PR #6650, opened 2025-06-20
Add io.github.teamclouday.AndroidMic

wiiznokes:

- [x] Please describe the application briefly. Use your Android phone as a microphone for your PC

https://github.com/user-attachments/assets/1e784336-717c-4e38-bc29-77c046b3c34b

- [x] I am an _(please keep whichever is applicable and remove the rest)_ upstream contributor to the project.
      If not, I contacted upstream developers about this submission. https://github.com/teamclouday/AndroidMic/issues/82

### DIFF
--- a/io.github.teamclouday.AndroidMic.json
+++ b/io.github.teamclouday.AndroidMic.json
@@ -1,5 +1,5 @@
 {
-  "id": "io.github.teamclouday.android-mic",
+  "id": "io.github.teamclouday.AndroidMic",
   "runtime": "org.freedesktop.Platform",
   "runtime-version": "24.08",
   "base": "com.system76.Cosmic.BaseApp",
@@ -63,7 +63,7 @@
         {
           "type": "git",
           "url": "https://github.com/wiiznokes/AndroidMic.git",
-          "commit": "f9803dd8ecadb23982b5896b3fca6404b1890605"
+          "commit": "cd56a77d2b3de1434f29b0ac05601782b1ddb8ae"
         },
         "cargo-sources.json"
       ]


wiiznokes:
I probably need a better solution to include adb. I'm open to sugestion

bbhtt:
screenshot is not video also you have to complete the checklist, tick the missing ones once it is done.

io.github.teamclouday.android-mic.json
```diff
@@ -0,0 +1,84 @@
...
+    {
+      "name": "adb",
+      "buildsystem": "simple",
+      "build-commands": ["install -Dm0755 platform-tools/adb /app/bin/adb"],
+      "sources": [
+        {
+          "type": "archive",
+          "strip-components": 0,
+          "dest-filename": "platform-tools-latest-linux.zip",
+          "url": "http://dl.google.com/android/repository/platform-tools-latest-linux.zip",
```
hfiguiere: how does that work on aarch64?

io.github.teamclouday.android-mic.json
```diff
@@ -0,0 +1,84 @@
...
+      "name": "protoc",
+      "buildsystem": "simple",
+      "build-commands": ["install -Dm0755 bin/protoc /app/bin/protoc"],
+      "cleanup": ["*"],
+      "sources": [
+        {
+          "type": "archive",
+          "strip-components": 0,
+          "dest-filename": "protoc.zip",
+          "url": "https://github.com/protocolbuffers/protobuf/releases/download/v29.3/protoc-29.3-linux-x86_64.zip",
```
hfiguiere: why don't you build from source?

io.github.teamclouday.android-mic.json
```diff
@@ -0,0 +1,84 @@
...
+  "sdk-extensions": ["org.freedesktop.Sdk.Extension.rust-stable"],
+  "command": "android-mic",
+  "finish-args": [
+    "--share=ipc",
+    "--socket=fallback-x11",
+    "--socket=wayland",
+    "--device=dri",
+    "--share=network",
+    "--talk-name=com.system76.CosmicSettingsDaemon",
+    "--filesystem=xdg-config/cosmic:ro",
```
bbhtt: needs exception

io.github.teamclouday.android-mic.json
```diff
@@ -0,0 +1,84 @@
...
+    "--device=usb",
```
bbhtt: This will not work on older flatpak versions

io.github.teamclouday.android-mic.json
```diff
@@ -0,0 +1,84 @@
...
+    "--socket=pulseaudio",
+    "--filesystem=xdg-run/pipewire-0 "
```
bbhtt:
```suggestion
    "--filesystem=xdg-run/pipewire-0"
```

this should be :ro?

io.github.teamclouday.android-mic.json
hfiguiere: device=usb isn't allowed as it is not compatible with most distro.

io.github.teamclouday.android-mic.json
wiiznokes: idk, afiak, adb is cross arch. But i can't test on aarch64

io.github.teamclouday.android-mic.json
wiiznokes: it's simpler this way i think

io.github.teamclouday.android-mic.json
hfiguiere: also this is an unversionned version which mean the build will break all the time. This should be a specific version. No "latest".

wiiznokes: i look over internet, and google doesn't provide link to download adb. I just found this link on this page https://developer.android.com/tools/releases/platform-tools#revisions

hfiguiere: > idk, afiak, adb is cross arch. But i can't test on aarch64

you don't know. it's your job here to know.

io.github.teamclouday.android-mic.json
hfiguiere: no it's not.

io.github.teamclouday.android-mic.json
wiiznokes: should i just disable the usb feature in the app? It wasn't working on my setup anyway when i tested

wiiznokes: > This will not work on older flatpak versions

you mean only the ""--device=usb"," line ?

bbhtt: yes

io.github.teamclouday.android-mic.json
wiiznokes: looks like arm isn't supported https://copr.fedorainfracloud.org/tmp/tmpp8kg92hp/niri.spec

io.github.teamclouday.android-mic.json
bbhtt: > should i just disable the usb feature in the app? 

if it makes sense. talk with upstream.

io.github.teamclouday.android-mic.json
wiiznokes: why?

io.github.teamclouday.android-mic.json
wiiznokes: done https://github.com/flathub-infra/flatpak-builder-lint/pull/670

io.github.teamclouday.android-mic.json
wiiznokes: i disabled adb option when on arm arch

io.github.teamclouday.android-mic.json
wiiznokes: i don't think it will be simpler to build from source. The project require a lot of build dependencies. I don't see why using the release on Github is not fine

io.github.teamclouday.android-mic.json
wiiznokes: i disabled the usb option. I'm sure upstream will be fine with that, he don't really care about flatpak

hfiguiere: you did find about `device=usb` in the documentation, right? Did you read that documentation?

wiiznokes: Yeah, but it doesn't works even on my system (with a up to date flatpak version)

io.github.teamclouday.android-mic.json
wiiznokes: > also this is an unversionned version which mean the build will break all the time. This should be a specific version. No "latest".

i honestly think it would be easier to fix the build at each update, because building adb is a pain, and release are not that frequent

hfiguiere: > i honestly think it would be easier to fix the build at each update

No it's not. Should be buildable at all times. Android have releases just pin one of them.

io.github.teamclouday.android-mic.json
hfiguiere: applications that need access to USB should use `device=all`.

don't know what not working since you are not telling.

io.github.teamclouday.android-mic.json
bbhtt: I think not building this from source is OK, as it is being cleaned up later. But also I don't think it requires a lot of dependencies, I see a bunch of apps building it without much. https://github.com/flathub/re.sonny.Workbench/blob/cd9237afb29ea9c6d5f356ba3a50751c294ab707/modules/libshumate.json#L14-L25

io.github.teamclouday.android-mic.json
bbhtt: some distros use this fork for abd https://github.com/nmeum/android-tools/ which is versioned and can be built from source. Probably also buildable on arm64

io.github.teamclouday.android-mic.json
```diff
@@ -0,0 +1,84 @@
...
+      "buildsystem": "simple",
+      "build-commands": ["install -Dm0755 platform-tools/adb /app/bin/adb"],
+      "sources": [
+        {
+          "type": "archive",
+          "strip-components": 0,
+          "dest-filename": "platform-tools-latest-linux.zip",
+          "url": "http://dl.google.com/android/repository/platform-tools-latest-linux.zip",
+          "sha256": "0ead642c943ffe79701fccca8f5f1c69c4ce4f43df2eefee553f6ccb27cbfbe8",
+          "only_arches": ["x86_64"]
```
bbhtt: what happens to the aarch64 build of the app without adb? I don't think it's a good idea to build x86_64 and arm64 slightly different.

io.github.teamclouday.android-mic.json
wiiznokes: it's indeed easier than i expected. I copy pasted the code to build from source

io.github.teamclouday.android-mic.json
wiiznokes: It was an error about accessory on the phone. I think the code for the usb option is not perfect currently, so it's fine to disable it for now utils it get fixed. I'm not even sure the error was due to flatpak, but more about detecting all phones

io.github.teamclouday.android-mic.json
wiiznokes: Yeah, i saw that in the fedora spec file. I'm currently trying to build it without success :/

wiiznokes:
oops, the appid should be `io.github.teamclouday.AndroidMic`

io.github.teamclouday.android-mic.json
hfiguiere: how do you access the device?

wiiznokes: with the Accessory protocol (https://github.com/teamclouday/AndroidMic/blob/main/RustApp/src/streamer/usb_streamer.rs)

hfiguiere: so basically you don't know. You are just winging it.

wiiznokes: yes and ? I'm just trying to add the app on flathub, not fixing every bug in the app. I'm not even the author of this part of the code. I already proposed a compromise by removing the option in the app, what do you want more

hfiguiere: so don't even read what I wrote. Nor the relevant bits in the documentation.

We didn't ask you to neuter USB support. We told you that `device=usb` didn't work. I even told you what to use instead. How does it access to the devices otherwise? I doubt this is even functional in the present state. But when I ask you how to access the device you link to some source code instead of telling me "it opens the USB devices" which with the current manifest won't work.

Part of the responsibility of packaging something is to be able to answer these questions. 

If the app is not functional there is no reason to submit it.

hfiguiere: The doc that would have answered all the questions https://docs.flatpak.org/en/latest/sandbox-permissions.html#device-access

io.github.teamclouday.android-mic.json
wiiznokes: closing as i'm planning to build it

io.github.teamclouday.android-mic.json
wiiznokes: Fyi, i already tested with `device=all` and it didn't work either. The adb one worked with no `device=all` or `device=usb` for some reason.

> If the app is not functional there is no reason to submit it.

Ofc, but in this case, it's just a feature of the app that don't work. Plus, the usb option can easilly be replaced by adb, so it not really important

gmanskibiditoilet:
> oops, the appid should be `io.github.teamclouday.AndroidMic`

great application, would u like to further test the flathub build?

wiiznokes:
> great application, would u like to further test the flathub build?

ty. I already know some bugs i need to fix. Notably the audio that make the app freeze 1/2 time. Maybe you can test if you have the same issue. I also didn't test the wifi option

PR #6667, opened 2025-06-26
Add com.fightcade.Fightcade.Steamos

Pobega:

This is an extension for `com.fightcade.Fightcade` that fixes things specifically for devices running SteamOS. Right now it includes the `min` browser since SteamOS gaming mode does not seem to work with URIs.

[Screencast From 2025-06-26 17-45-12.webm](https://github.com/user-attachments/assets/b020d3c3-9ffc-444d-a97d-a79a9542a056)

- [x] I am an _(please keep whichever is applicable and remove the rest)_ author/developer/upstream contributor to the project.

Not applicable here (see https://github.com/flathub/flathub/pull/6667#issuecomment-3016682178)

Considering this is an extension that adds quality of life for a Flathub application I already maintain, is this relevant? It is not providing a new individual application. Could someone please advise here?

### DIFF
--- a/com.fightcade.Fightcade.Steamos.yml
+++ b/com.fightcade.Fightcade.Steamos.yml
@@ -25,12 +25,10 @@ modules:
     sources:
       - type: archive
         url: https://download.gnome.org/sources/libsecret/0.21/libsecret-0.21.6.tar.xz
-        sha256: 747b8c175be108c880d3adfb9c3537ea66e520e4ad2dccf5dce58003aeeca090 
+        sha256: 747b8c175be108c880d3adfb9c3537ea66e520e4ad2dccf5dce58003aeeca090
   - name: min
     buildsystem: simple
     build-commands:
-      # Replace the placeholder PREFIX with the actual destination path in the run script
-      - sed -i "s|PREFIX|${FLATPAK_DEST}|g" run.sh
       # Install the run script for the Min browser
       - install -Dm755 run.sh ${FLATPAK_DEST}/bin/min-browser
       - install -Dm755 apply_extra ${FLATPAK_DEST}/bin/apply_extra
@@ -53,7 +51,7 @@ modules:
       - type: script
         dest-filename: run.sh
         commands:
-          - zypak-wrapper PREFIX/extra/min/min --no-gpu-sandbox --no-sandbox "$@" 2>&1
+          - zypak-wrapper /app/steamos/extra/min/min --no-gpu-sandbox --no-sandbox "$@" 2>&1
       - type: script
         dest-filename: apply_extra
         commands:
--- a/com.fightcade.Fightcade.Steamos.yml
+++ b/com.fightcade.Fightcade.Steamos.yml
@@ -34,8 +34,6 @@ modules:
       # Install the browser
       - bsdtar -xOf min.deb data.tar.xz | bsdtar -xJf -
       - mv opt/Min min
-      - rm -rf opt usr
-      - rm min.deb
       - mv min ${FLATPAK_DEST}/
     post-install:
       # Install the metainfo file for the Min browser
--- a/com.fightcade.Fightcade.Steamos.yml
+++ b/com.fightcade.Fightcade.Steamos.yml
@@ -48,8 +48,8 @@ modules:
         dest-filename: min.deb
         only-arches:
           - x86_64
-        url: https://github.com/minbrowser/min/releases/download/v1.35.0/min-1.35.0-amd64.deb
-        sha256: 6d066842e8d0f0fa9aad0bd7cb5f56a32a67937ac75e8710c3b229cbcfc89220
+        url: https://github.com/minbrowser/min/releases/download/v1.35.1/min-1.35.1-amd64.deb
+        sha256: 9ce204945a22cb88be860dba1141c896f927982bca0940e1f08f87333a9556db
       - type: file
         path: xdg-mime.sh
       - type: file
--- a/com.fightcade.Fightcade.Steamos.yml
+++ b/com.fightcade.Fightcade.Steamos.yml
@@ -10,7 +10,7 @@ modules:
     buildsystem: meson
     build-options:
       prefix: /app/steamos
-      lib: lib
+      libdir: lib
     config-opts:
       - -Dmanpage=false
       - -Dvapi=false
@@ -31,7 +31,12 @@ modules:
     build-commands:
       # Install the run script for the Min browser
       - install -Dm755 run.sh ${FLATPAK_DEST}/bin/min-browser
-      - install -Dm755 apply_extra ${FLATPAK_DEST}/bin/apply_extra
+      # Install the browser
+      - bsdtar -xOf min.deb data.tar.xz | bsdtar -xJf -
+      - mv opt/Min min
+      - rm -rf opt usr
+      - rm min.deb
+      - mv min ${FLATPAK_DEST}/
     post-install:
       # Install the metainfo file for the Min browser
       - >-
@@ -39,23 +44,15 @@ modules:
         -t ${FLATPAK_DEST}/share/metainfo/
         ${FLATPAK_ID}.metainfo.xml
     sources:
-      - type: extra-data
-        filename: min.deb
+      - type: file
+        dest-filename: min.deb
         only-arches:
           - x86_64
         url: https://github.com/minbrowser/min/releases/download/v1.35.0/min-1.35.0-amd64.deb
         sha256: 6d066842e8d0f0fa9aad0bd7cb5f56a32a67937ac75e8710c3b229cbcfc89220
-        size: 79754278
       - type: file
         path: com.fightcade.Fightcade.Steamos.metainfo.xml
       - type: script
         dest-filename: run.sh
         commands:
-          - zypak-wrapper /app/steamos/extra/min/min "$@" 2>&1
-      - type: script
-        dest-filename: apply_extra
-        commands:
-          - bsdtar -xOf min.deb data.tar.xz | bsdtar -xJf -
-          - mv opt/Min min
-          - rm -rf opt usr
-          - rm min.deb
+          - zypak-wrapper /app/steamos/min/min "$@" 2>&1
--- a/com.fightcade.Fightcade.Steamos.yml
+++ b/com.fightcade.Fightcade.Steamos.yml
@@ -51,7 +51,7 @@ modules:
       - type: script
         dest-filename: run.sh
         commands:
-          - zypak-wrapper /app/steamos/extra/min/min --no-gpu-sandbox --no-sandbox "$@" 2>&1
+          - zypak-wrapper /app/steamos/extra/min/min "$@" 2>&1
       - type: script
         dest-filename: apply_extra
         commands:
--- a/com.fightcade.Fightcade.Steamos.yml
+++ b/com.fightcade.Fightcade.Steamos.yml
@@ -33,12 +33,16 @@ modules:
       - install -Dm755 run.sh ${FLATPAK_DEST}/bin/min-browser
       # Install xdg-mime shim
       - install -Dm755 xdg-mime.sh ${FLATPAK_DEST}/bin/xdg-mime
-      # Install the browser
+      # extract debian package
       - bsdtar -xOf min.deb data.tar.xz | bsdtar -xJf -
       - mv opt/Min min
+      # make a copy of apache 2.0 license and install the NOTICE file
+      - install -Dm644 min/resources/app/LICENSE.txt ${FLATPAK_DEST}/share/licenses/LICENSE-APACHE-2.0.txt
+      - install -Dm644 -t ${FLATPAK_DEST}/share/licenses/ NOTICE.txt
+      # Install the browser
       - mv min ${FLATPAK_DEST}/
     post-install:
-      # Install the metainfo file for the Min browser
+      # Install the metainfo file
       - >-
         install -Dm644
         -t ${FLATPAK_DEST}/share/metainfo/
@@ -52,6 +56,8 @@ modules:
         sha256: 9ce204945a22cb88be860dba1141c896f927982bca0940e1f08f87333a9556db
       - type: file
         path: xdg-mime.sh
+      - type: file
+        path: NOTICE.txt
       - type: file
         path: com.fightcade.Fightcade.Steamos.metainfo.xml
       - type: script
--- a/com.fightcade.Fightcade.Steamos.yml
+++ b/com.fightcade.Fightcade.Steamos.yml
@@ -1,0 +1,63 @@
+id: com.fightcade.Fightcade.Steamos
+runtime: com.fightcade.Fightcade
+runtime-version: stable
+sdk: org.freedesktop.Sdk//24.08
+build-extension: true
+separate-locales: false
+
+modules:
+  - name: libsecret
+    buildsystem: meson
+    build-options:
+      prefix: /app/steamos
+      lib: lib
+    config-opts:
+      - -Dmanpage=false
+      - -Dvapi=false
+      - -Dgtk_doc=false
+      - -Dintrospection=false
+    cleanup:
+      - "/bin"
+      - "/include"
+      - "/lib/pkgconfig"
+      - "/share/gir-1.0"
+      - "/share/man"
+    sources:
+      - type: archive
+        url: https://download.gnome.org/sources/libsecret/0.21/libsecret-0.21.6.tar.xz
+        sha256: 747b8c175be108c880d3adfb9c3537ea66e520e4ad2dccf5dce58003aeeca090 
+  - name: min
+    buildsystem: simple
+    build-commands:
+      # Replace the placeholder PREFIX with the actual destination path in the run script
+      - sed -i "s|PREFIX|${FLATPAK_DEST}|g" run.sh
+      # Install the run script for the Min browser
+      - install -Dm755 run.sh ${FLATPAK_DEST}/bin/min-browser
+      - install -Dm755 apply_extra ${FLATPAK_DEST}/bin/apply_extra
+    post-install:
+      # Install the metainfo file for the Min browser
+      - >-
+        install -Dm644
+        -t ${FLATPAK_DEST}/share/metainfo/
+        ${FLATPAK_ID}.metainfo.xml
+    sources:
+      - type: extra-data
+        filename: min.deb
+        only-arches:
+          - x86_64
+        url: https://github.com/minbrowser/min/releases/download/v1.35.0/min-1.35.0-amd64.deb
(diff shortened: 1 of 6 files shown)


com.fightcade.Fightcade.Steamos.yml
```diff
@@ -0,0 +1,63 @@
...
+      - "/share/man"
+    sources:
+      - type: archive
+        url: https://download.gnome.org/sources/libsecret/0.21/libsecret-0.21.6.tar.xz
+        sha256: 747b8c175be108c880d3adfb9c3537ea66e520e4ad2dccf5dce58003aeeca090 
+  - name: min
+    buildsystem: simple
+    build-commands:
+      # Replace the placeholder PREFIX with the actual destination path in the run script
+      - sed -i "s|PREFIX|${FLATPAK_DEST}|g" run.sh
```
hfiguiere: why do you need to `sed`?

Pobega: If `${FLATPAK_DEST}` is placed directly into the script section below, it results in this:

```
[📦 com.fightcade.Fightcade bin]$ cat min-browser 
#!/bin/sh
zypak-wrapper ${FLATPAK_DEST}/extra/min/min --no-gpu-sandbox --no-sandbox "$@" 2>&1
```

Which is not a valid variable at runtime.

The `sed` command allows us to insert the path dynamically (should things change/move around in the future), but if it's preferred we could hard code it.

```
[📦 com.fightcade.Fightcade ~]$ cat /app/steamos/bin/min-browser 
#!/bin/sh
zypak-wrapper /app/steamos/extra/min/min --no-gpu-sandbox --no-sandbox "$@" 2>&1
```

Pobega:
Any help with the final check in the PR would be helpful; am I considered the upstream if this is an extension for a Flathub application I maintain? I'm unsure who the 'upstream' is in this scenario.

com.fightcade.Fightcade.Steamos.yml
hfiguiere: because it's not available at runtime. Just put the actual value. It's better than sed. There not need to be complicated.

Pobega: Done

com.fightcade.Fightcade.Steamos.yml
```diff
@@ -0,0 +1,61 @@
...
+      - >-
+        install -Dm644
+        -t ${FLATPAK_DEST}/share/metainfo/
+        ${FLATPAK_ID}.metainfo.xml
+    sources:
+      - type: extra-data
+        filename: min.deb
+        only-arches:
+          - x86_64
+        url: https://github.com/minbrowser/min/releases/download/v1.35.0/min-1.35.0-amd64.deb
```
bbhtt: why is this not being built from source?

com.fightcade.Fightcade.Steamos.yml
```diff
@@ -0,0 +1,61 @@
...
+        sha256: 6d066842e8d0f0fa9aad0bd7cb5f56a32a67937ac75e8710c3b229cbcfc89220
+        size: 79754278
+      - type: file
+        path: com.fightcade.Fightcade.Steamos.metainfo.xml
+      - type: script
+        dest-filename: run.sh
+        commands:
+          - zypak-wrapper /app/steamos/extra/min/min --no-gpu-sandbox --no-sandbox "$@" 2>&1
```
bbhtt: drop no-sandbox if you use zypak.

com.fightcade.Fightcade.Steamos.yml
bbhtt: ah i just noticed extra-data. but why is it extra-data, the licence is open source?

bbhtt:
> Any help with the final check in the PR would be helpful; am I considered the upstream if this is an extension for a Flathub application I maintain? I'm unsure who the 'upstream' is in this scenario.

https://github.com/minbrowser/min/


Why does this need to be an extension for fightcade? This seems like it'd be its own app submission.

com.fightcade.Fightcade.Steamos.yml
rezzafr33: Hi, I'm assisting @Pobega with this extension.
> why is this not being built from source?

The reason it's not building from source is that Min browser uses a forked version of Dragula, hosted on GitHub, instead of the npm version. This requires an active internet connection during the build process, which Flatpak Builder doesn't support.

We could switch to the [upstream npm version](https://github.com/rezzafr33/com.fightcade.Fightcade/blob/59f729631bb66e295fb6190f6fdb8d52b160befa/com.fightcade.Fightcade.yml#L302) of Dragula with minimal impact on [tab dragging](https://github.com/minbrowser/min/issues/735#issuecomment-1986700500), which honestly doesn't affect our use case.

> ah i just noticed extra-data. but why is it extra-data, the licence is open source?

The Apache 2.0 license for Min Browser likely necessitates a NOTICE file. I'm looking for input on whether to build it from source or use a prebuilt binary and package it directly without using extra data.

Pobega:
> Why does this need to be an extension for fightcade? This seems like it'd be its own app submission.

This is going to be a part of multiple enhancements for the Steam Deck/SteamOS that are exclusive to problems with running Fightcade on that platform. For example, we currently carry https://github.com/flathub/com.fightcade.Fightcade/pull/145 in the main Flatpak -- that would be another thing I'd prefer to just stick into an extension as to not interfere with the main Flatpak.

The idea is that you install this extension if you are on Deck, and if not everything continues as usual. It's to avoid us having to do runtime checks, since the Flatpak can't easily check if it's running on Deck or not -- just install this extension if you are on Deck.

The _actual_ solution would be for Valve to just fix URI handling for Flatpaks in gaming mode, but I doubt that is going to happen in any realistic time period.

bbhtt:
> The actual solution would be for Valve to just fix URI handling for Flatpaks in gaming mode

Can you provide more details on what is the problem here? What's broken on steam os?

com.fightcade.Fightcade.Steamos.yml
Pobega: Note that we actually needed to add `--no-sandbox` to work around a **long** outstanding issue with ReCaptcha not working in the Electron client: https://github.com/flathub/com.fightcade.Fightcade/pull/149

I don't mind dropping it for this, it was just habit to throw that in there.

Edit: And `--no-gpu-sandbox` (actually `--disable-gpu-sandbox`) was needed to work around the client not rendering on Nvidia GPUs.

Shouldn't be relevant here, until SteamOS devices with Nvidia GPUs become commonplace.

Pobega:
> Can you provide more details on what is the problem here? What's broken on steam os?

It's mentioned briefly in the OP, but in gaming mode for some reason URL/URI handling isn't processed correctly for Flatpaks. For example `file://` and `http://` URLs do nothing, they just never get propagated out to `xdg-open`

This doesn't seem to be an issue for regular applications, just Flatpaks.

com.fightcade.Fightcade.Steamos.yml
Pobega: [4f1bfb6](https://github.com/flathub/flathub/pull/6667/commits/4f1bfb64b314910e9220f1de60b4b740c05e6649)

Pobega:
> Why does this need to be an extension for fightcade? This seems like it'd be its own app submission.

Oh I also forgot to mention in my last comment, the big reason for wanting to split this into an extension rather than support it directly in the main Flatpak is that I don't want build issues related to SteamOS support to stop new versions of Fightcade from being disted.

Currently we use [x-checker-data](https://github.com/flathub/com.fightcade.Fightcade/blob/master/com.fightcade.Fightcade.yml#L142-L146) to fetch the newest release of Fightcade from a JSON file - when there's a new release, previous releases become unusable (which is also why we were allowed an [exception for a 0 hour publish-delay](https://github.com/flathub/com.fightcade.Fightcade/blob/master/flathub.json#L4).)

It would be a nightmare if something broken with, say, Min browser was completely blocking an upgrade for all users and rendering the Flatpak completely unusable, especially when the majority aren't even on SteamOS.

rezzafr33:
This is what we're trying to achieve, recorded with steamdeck builtin screen recorder in gamescope session/gaming mode:

https://github.com/user-attachments/assets/fd7133f5-d62d-4685-9595-84e658404c45
 
won't work without this extension because right now Steamos/steamdeck doesn't provide xdg-desktop-portal interfaces for Gaming mode.

Pobega:
For clarity on this above video, when clicking the URL to view recorded replays it opens [fightcade.com/replay](https://www.fightcade.com/replay) in a web browser (which doesn't work in SteamOS gaming mode due to the URL handling being broken, so we manually call out to a bundled `min`), and from the browser the replays are expressed as `fcade://` URLs (which also doesn't work without the `xdg-open` workaround in https://github.com/flathub/com.fightcade.Fightcade/pull/145)

barthalion:
Out of curiosity, what are the disadvantages of bundling it with the main app?

barthalion:
I mean other than this:

> Oh I also forgot to mention in my last comment, the big reason for wanting to split this into an extension rather than support it directly in the main Flatpak is that I don't want build issues related to SteamOS support to stop new versions of Fightcade from being disted.

Because I can't see it blocking anything if it repackages a binary artifact.

bbhtt:
Is this a bug in SteamOS or in some Flatpak related things? Did you investigate?

com.fightcade.Fightcade.Steamos.yml
bbhtt: if it doesn't build without network then leave it be.

rezzafr33:
> Is this a bug in SteamOS or in some Flatpak related things? Did you investigate?

This is a Steam Deck/SteamOS Gaming Mode limitation. Gaming Mode uses Gamescope, which currently lacks `xdg-desktop-portal` implementation. This issue doesn't occur in Desktop Mode, as it uses the KDE Plasma desktop.

edit: related issue https://github.com/Nexus-Mods/NexusMods.App/issues/3084

Pobega:
(re: the above, my SteamOS investigations were done 2+ years ago on this so apologies for not providing all the information upfront -- a lot of it is just from my foggy memory of trying to get it to work in 2023)

> Out of curiosity, what are the disadvantages of bundling it with the main app?

I wanted to split it into an extension to make it easy to just separate the SteamOS specific changes. I'm writing this kind of hastily so apologies if it's not well organized:

* Maintenance burden is one nice advantage of the extension. I'd prefer not to have to maintain codepaths for something that isn't my primary concern for support.
  * ~~While there are ways to detect if you're on SteamOS (IIRC there's a `/STEAMOS` file or something similar on the root partition -- though I can't find anything referencing that online right now) that requires reading files from outside of the Flatpak, which I'd prefer to avoid.~~  **Edit**: Actually I think this file does not exist -- I was conflating it with the `/.flatpak-info` file that exists in Flatpaks
    * Additionally if someone not on SteamOS wants these enhancements for any reason, it's easier to tell them to install the extension rather than create `/STEAMOS`
  * The other ways of detecting if running on Deck/SteamOS are through the SteamAPI, and since Fightcade is completely separate from Steam that is not ideal.
* While `min` is not built from source, `libsecret` is. Additionally if the `min` mirror ever disappears or the URL changes it will break our builds (which is easily fixable) but Fightcade **will** break in the meantime if the client is outdated.

The **primary** reason for wanting this in an extension is to keep the Fightcade Flatpak functional, though. As mentioned above we allow Flathubbot to automerge PRs to keep the Fightcade client up-to-date without any human interaction. Any build breakage will cause the Flatpak to become outdated and thus unusable, which is my highest priority as the maintainer.

From a technical perspective this is completely possible to do from inside `com.fightcade.Fightcade` (and how it was originally implemented.) Splitting it into an extension is more helpful to me as the maintainer than it is a technical advantage.

bbhtt:
Ok, can you switch to a regular source instead of extra-data? you can ship the notice inside the extension but as long as redistribution is allowed (which apache-2.0 allows) you do not need extra-data.

Pobega:
[0cb7d7a](https://github.com/flathub/flathub/pull/6667/commits/0cb7d7a4d1081203e713f4f8796498bd2ff89b22): Switched from `extra-data` to plain `file`

bbhtt:
bot, build

com.fightcade.Fightcade.Steamos.yml
```diff
@@ -0,0 +1,61 @@
...
+sdk: org.freedesktop.Sdk//24.08
+build-extension: true
+separate-locales: false
+
+modules:
+  - name: libsecret
+    buildsystem: meson
+    build-options:
+      prefix: /app/steamos
+      lib: lib
```
bbhtt:
```suggestion
      libdir: lib
```

com.fightcade.Fightcade.Steamos.yml
```diff
@@ -0,0 +1,61 @@
...
+        -t ${FLATPAK_DEST}/share/metainfo/
+        ${FLATPAK_ID}.metainfo.xml
+    sources:
+      - type: file
+        filename: min.deb
+        only-arches:
+          - x86_64
+        url: https://github.com/minbrowser/min/releases/download/v1.35.0/min-1.35.0-amd64.deb
+        sha256: 6d066842e8d0f0fa9aad0bd7cb5f56a32a67937ac75e8710c3b229cbcfc89220
+        size: 79754278
```
bbhtt:
```suggestion
```

com.fightcade.Fightcade.Steamos.yml
```diff
@@ -0,0 +1,61 @@
...
+      - install -Dm755 apply_extra ${FLATPAK_DEST}/bin/apply_extra
+    post-install:
+      # Install the metainfo file for the Min browser
+      - >-
+        install -Dm644
+        -t ${FLATPAK_DEST}/share/metainfo/
+        ${FLATPAK_ID}.metainfo.xml
+    sources:
+      - type: file
+        filename: min.deb
```
bbhtt:
```suggestion
        dest-filename: min.deb
```

com.fightcade.Fightcade.Steamos.yml
```diff
@@ -0,0 +1,61 @@
...
+        dest-filename: run.sh
+        commands:
+          - zypak-wrapper /app/steamos/extra/min/min "$@" 2>&1
+      - type: script
+        dest-filename: apply_extra
+        commands:
+          - bsdtar -xOf min.deb data.tar.xz | bsdtar -xJf -
+          - mv opt/Min min
+          - rm -rf opt usr
+          - rm min.deb
```
bbhtt:
```suggestion
```

com.fightcade.Fightcade.Steamos.yml
```diff
@@ -0,0 +1,61 @@
...
+    sources:
+      - type: archive
+        url: https://download.gnome.org/sources/libsecret/0.21/libsecret-0.21.6.tar.xz
+        sha256: 747b8c175be108c880d3adfb9c3537ea66e520e4ad2dccf5dce58003aeeca090
+  - name: min
+    buildsystem: simple
+    build-commands:
+      # Install the run script for the Min browser
+      - install -Dm755 run.sh ${FLATPAK_DEST}/bin/min-browser
+      - install -Dm755 apply_extra ${FLATPAK_DEST}/bin/apply_extra
```
bbhtt:
```suggestion
```

com.fightcade.Fightcade.Steamos.yml
```diff
@@ -0,0 +1,61 @@
...
+      - "/share/man"
+    sources:
+      - type: archive
+        url: https://download.gnome.org/sources/libsecret/0.21/libsecret-0.21.6.tar.xz
+        sha256: 747b8c175be108c880d3adfb9c3537ea66e520e4ad2dccf5dce58003aeeca090
+  - name: min
+    buildsystem: simple
+    build-commands:
+      # Install the run script for the Min browser
+      - install -Dm755 run.sh ${FLATPAK_DEST}/bin/min-browser
```
bbhtt, line 33: you have to unpack and install the deb here now

Pobega:
I hastily changed `extra-data` to `file` without really thinking of the side effects, apologies. I have updated the commit with [cc0a115](https://github.com/flathub/flathub/pull/6667/commits/cc0a115965581fe59ba4c9023c53469a65c8ee3b) which seems to work fine, and should resolve all the review comments above.

com.fightcade.Fightcade.Steamos.yml
```diff
@@ -0,0 +1,58 @@
...
+      # Install the browser
+      - bsdtar -xOf min.deb data.tar.xz | bsdtar -xJf -
+      - mv opt/Min min
+      - rm -rf opt usr
+      - rm min.deb
```
bbhtt:
```suggestion
```

These aren't installed, you don't need to clean these

Pobega: Oh right, I forgot that the things in the build stage aren't committed. 

[9ca37a9](https://github.com/flathub/flathub/pull/6667/commits/9ca37a98e811a55838235b4ff685a32eff95a7ea)

bbhtt:
bot, build

Pobega:
I never received a direct response, am I okay to mark the fifth checkmark in this scenario?

As for the NOTICE file, @rezzafr33 do you have any insight into what we need to put in it? And do we need to put it somewhere inside of the Flatpak itself, or on the `com.fightcade.Fightcade.Steamos` repository?

bbhtt:
You can mark it, it doesn't really apply

rezzafr33:
> I never received a direct response, am I okay to mark the fifth checkmark in this scenario?
> 
> As for the NOTICE file, @rezzafr33 do you have any insight into what we need to put in it? And do we need to put it somewhere inside of the Flatpak itself, or on the `com.fightcade.Fightcade.Steamos` repository?

It appears a NOTICE file isn't required, as the project doesn't explicitly provide one. A copy of the Apache 2.0 LICENSE.txt file alongside the min binary should be sufficient. Nevertheless, here's an example NOTICE file for reference::

```
Min Browser is distributed under the Apache License, Version 2.0.

A copy of this license is included in the LICENSE.txt file in this distribution.

The original work for Min Browser is copyrighted by:
Copyright 2016-2025 PalmerAL (and other contributors as noted in the Min Browser project)
You may obtain a copy of the Apache License, Version 2.0 at:
http://www.apache.org/licenses/LICENSE-2.0

Unless required by applicable law or agreed to in writing, software
distributed under the Apache License, Version 2.0 is distributed on an
"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
express or implied. See the Apache License, Version 2.0 for the
specific language governing permissions and limitations under the
License.
```

Pobega:
@rezzafr33 updated the repo for review, I believe it should be ready. See https://github.com/Pobega/flathub/pull/1

bbhtt:
bot, build

PR #6674, opened 2025-06-30
Add com.jetbrains.Toolbox

dreamyukii:

 This applications is for downloading and managing JetBrains IDE 

https://github.com/user-attachments/assets/b2169bc8-d506-45ed-83b9-c73308b8c5b0

- [ ] I am an _(please keep whichever is applicable and remove the rest)_ author/developer/upstream contributor to the project.


com.jetbrains.Toolbox.yml
```diff
@@ -0,0 +1,65 @@
...
+separate-locales: false
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --device=dri
+  - --socket=x11
+  - --socket=ssh-auth
+  - --share=network
+  - --socket=session-bus
+  - --talk-name=org.freedesktop.secrets
```
bbhtt: why is this needed?

jetbrains-toolbox.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+exec env XDG_CONFIG_HOME="$HOME/.config" XDG_DATA_HOME="$HOME/.local/share" XDG_CACHE_HOME="$HOME/.cache" /app/extra/toolbox/bin/jetbrains-toolbox "$@"
```
bbhtt: why are you changing these env vars?

com.jetbrains.Toolbox.yml
```diff
@@ -0,0 +1,65 @@
...
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --filesystem=home
```
bbhtt, line 16: why does it need entire home? where are they downloaded by default?

com.jetbrains.Toolbox.appdata.xml
```diff
@@ -0,0 +1,46 @@
...
+      NOTE: This wrapper is not verified by, affiliated with, or supported by JetBrains s.r.o.
+    </p>
+  </description>
+  <launchable type="desktop-id">com.jetbrains.Toolbox.desktop</launchable>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://www.jetbrains.com/toolbox-app/img/1_en-us.png</image>
+    </screenshot>
+  </screenshots>
+</component>
```
bbhtt, line 264: missing releases. make sure to validate.

com.jetbrains.Toolbox.yml
dreamyukii: Toolbox app will download the IDE at `$HOME/.local/share/JetBrains/Toolbox/apps`, and the config file for the IDE at `$HOME/.config/JetBrains`.

jetbrains-toolbox.sh
dreamyukii: This env is required for Toolbox working properly with flatpak, without this env vars Toolbox will download the IDE at `~/.var/app/com.jetbrains.Toolbox/data` and not in  `$HOME/.local/share/JetBrains/Toolbox/apps`. and the location of config file and cache will located at `~/.var/app/com.jetbrains.Toolbox/ config` and `~/.var/app/com.jetbrains.Toolbox/ cache`.

bbhtt: Config and cache files are supposed to be under ~/.var/app for Flatpak apps.

What is the problem when the IDE is downloaded to ~/.var?

Can you customise that path somehow, without changing this env var?

com.jetbrains.Toolbox.yml
bbhtt: So you need $HOME/.local/share/JetBrains only then. Config files are supposed to be in ~/.var/app for Flatpak apps

> config file for the IDE

The IDE is not getting launched under Flatpak is it? 

I thought this app was only for downloading and managing (uninstalling etc.)

hfiguiere:
I am not sure if this is suitable for flathub.

castrojo:
I'd like to justify why this should be on flathub. The problem with the jetbrains flatpaks is they don't really work well for how people expect jetbrains to work. Upstream developed this all-in-one app to manage of all their products. 

It's distro agnostic because it just installs everything in the user's home directory, which means when the user asks for developer stuff, python, build deps, etc, you name it, it's all managed by the Jetbrains IDE themselves. It downloads everything, there's no need for the IDEs to try to figure out how to talk to the flatpak runtimes.

It effectively gives up trying and decides to manage it itself outside of any linux distribution package. This also means that they IDEs Just Work right out of the gate, as opposted to the user figuring out how to set this up by hand. 

Therefore, my conclusion is, if it's all self contained anyway let's stick it in a flatpak so it's at least isolated and see how we get on. And if it delivers a better experience than the existing flatpaks, it gives us an opportunity to re-engage with Jetbrains for official support. I volunteer to talk to them!

com.jetbrains.Toolbox.appdata.xml
dreamyukii: done

com.jetbrains.Toolbox.yml
dreamyukii: sorry for the late response
ssh-auth is for ssh feature in this applications
network is needed for downloading IDE
session bus and secrets is not needed anymore

jetbrains-toolbox.sh
dreamyukii: sorry again for late response
yes you can change the location in the settings menu
i think i will remove the env from launch script

com.jetbrains.Toolbox.yml
```diff
@@ -0,0 +1,62 @@
...
+        commands:
+          - mkdir toolbox/
+          - tar -xzf toolbox.tar.gz --directory=toolbox/ --strip-components=1
+          - rm toolbox.tar.gz
+      - type: file
+        path: com.jetbrains.Toolbox.appdata.xml
+      - type: file
+        path: com.jetbrains.Toolbox.desktop
+      - type: file
+        path: com.jetbrains.Toolbox.svg
```
x80486: Alternatively, you could retrieve the SVG icon from [here](https://resources.jetbrains.com/storage/products/company/brand/logos/Toolbox_icon.svg). Full example here: https://github.com/flathub/com.jetbrains.DataGrip/blob/master/com.jetbrains.DataGrip.yaml#L93

jetbrains-toolbox.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+exec env XDG_DATA_HOME="$HOME/.local/share" /app/extra/toolbox/bin/jetbrains-toolbox "$@"
```
x80486, line 2: You shouldn't need to redefine this variable — or need this file. The Toolbox application, if it uses the `XDG Base Directory` swarm correctly, it will download everything into `~/.var/app/com.jetbrains.Toolbox/...`, and since the products aren't managed by Flatpak, each will create the required directories in the user's home.

Also, you shouldn't need `home` access, unless there is something wonky going on with the application.

com.jetbrains.Toolbox.yml
```diff
@@ -0,0 +1,62 @@
+id: com.jetbrains.Toolbox
+command: jetbrains-toolbox
+runtime: org.freedesktop.Platform
+runtime-version: "24.08"
+sdk: org.freedesktop.Sdk
+tags:
+  - proprietary
+separate-locales: false
+finish-args:
+  - --share=ipc
```
x80486, line 10: Last time I used Toolbox they required credentials.

You may need `--filesystem=xdg-run/keyring` and `--talk-name=org.freedesktop.secrets` — but give it a whirl without those initially...miracles happen 😅

jetbrains-toolbox.sh
dreamyukii: unfortunately Toolbox app need this workaround for the desktop file to function correctly

com.jetbrains.Toolbox.yml
dreamyukii: Thanks for the suggestion

PR #6677, opened 2025-07-01
Add net.bontal.Catalyst

pewriebontal:

- [X] Please describe the application briefly. < Catalyst is an intelligent task manager that uses AI to parse natural language into structured tasks. It automatically prioritizes them using the Eisenhower Matrix (Do, Schedule, Delegate, Eliminate). The app is designed to be privacy-first with local storage and features a clean, minimalist interface. >
- [X] Please attach a video showcasing the application on Linux using the Flatpak. <[Screencast_20250702_021800.webm](https://github.com/user-attachments/assets/bafc3551-2755-4142-811e-1d79e4d8d53b)>
- [X] The Flatpak ID follows all the rules listed in the [Application ID requirements][net.bontal.Catalyst].
- [X] I am an _(please keep whichever is applicable and remove the rest)_ author/developer/upstream contributor to the project.


net.bontal.Catalyst.yml
```diff
@@ -0,0 +1,59 @@
...
+      - install -Dm644 net.bontal.Catalyst.desktop /app/share/applications/net.bontal.Catalyst.desktop
+      - install -Dm644 net.bontal.Catalyst.metainfo.xml /app/share/metainfo/net.bontal.Catalyst.metainfo.xml
+      - install -Dm644 icon-256.png /app/share/icons/hicolor/256x256/apps/net.bontal.Catalyst.png
+      - install -Dm644 icon-128.png /app/share/icons/hicolor/128x128/apps/net.bontal.Catalyst.png
+      - install -Dm644 icon-64.png /app/share/icons/hicolor/64x64/apps/net.bontal.Catalyst.png
+      - desktop-file-validate /app/share/applications/net.bontal.Catalyst.desktop
+      - update-desktop-database /app/share/applications
+    sources:
+      - type: file
+        url: https://pkg.bontal.net/downloads/releases/catalyst-1.1.0-linux-x64.tar.gz
```
hfiguiere, line 51: what happen with the aarch64 package?

net.bontal.Catalyst.yml
```diff
@@ -0,0 +1,59 @@
+app-id: net.bontal.Catalyst
+runtime: org.gnome.Platform
+runtime-version: "48"
+sdk: org.gnome.Sdk
```
hfiguiere: why do you need the gnome platform?

icon-64.png
hfiguiere, line 1: you don't need this. remove it.

icon-128.png
hfiguiere, line 1: you don't need this. remove it.

net.bontal.Catalyst.yml
```diff
@@ -0,0 +1,59 @@
...
+      - mkdir -p /app/catalyst
+      - cp -r ./* /app/catalyst/
+      - mkdir -p /app/bin
+      - ln -s /app/catalyst/catalyst /app/bin/catalyst
+      - install -Dm644 net.bontal.Catalyst.desktop /app/share/applications/net.bontal.Catalyst.desktop
+      - install -Dm644 net.bontal.Catalyst.metainfo.xml /app/share/metainfo/net.bontal.Catalyst.metainfo.xml
+      - install -Dm644 icon-256.png /app/share/icons/hicolor/256x256/apps/net.bontal.Catalyst.png
+      - install -Dm644 icon-128.png /app/share/icons/hicolor/128x128/apps/net.bontal.Catalyst.png
+      - install -Dm644 icon-64.png /app/share/icons/hicolor/64x64/apps/net.bontal.Catalyst.png
+      - desktop-file-validate /app/share/applications/net.bontal.Catalyst.desktop
```
hfiguiere:
```suggestion
```

net.bontal.Catalyst.yml
```diff
@@ -0,0 +1,59 @@
...
+      - update-desktop-database /app/share/applications
```
hfiguiere:
```suggestion
```

net.bontal.Catalyst.desktop
hfiguiere, line 1: The upstream package doesn't have a .desktop ? It's a bug if it doesn't.

net.bontal.Catalyst.yml
pewriebontal: The org.gnome.Platform is required to build the zenity dependency, as it needs libadwaita. If there's a more minimal way to achieve this that you'd recommend, I'm happy to make the change.

icon-64.png
pewriebontal: ok

net.bontal.Catalyst.yml
hfiguiere: the freedesktop runtime includes zenity. Also you can build the older version of zenity that only require gtk3 which is in the freedesktop runtime, but since it's already there....

net.bontal.Catalyst.yml
pewriebontal: Only x86_64 binaries are available at this time, and the manifest now specifies only-arches: [x86_64].

net.bontal.Catalyst.yml
pewriebontal: Thanks for the info. The app now uses  freedesktop runtime and I am building `libsecret` as a module to handle its runtime dependency.

net.bontal.Catalyst.desktop
pewriebontal: noted

pewriebontal:
Hi @hfiguiere, thank you for all your guidance through the process.

I have pushed the final changes and believe I have addressed all the feedback:

*   **Platform:** The manifest now uses `org.freedesktop.Platform:24.08` and builds the required `libsecret` dependency, as `zenity` is already included in the runtime.
*   **Architecture:** The build is restricted to `x86_64` as aarch64 binaries are not yet available.
*   **Icons:** The smaller, redundant icon files and their corresponding install commands have been removed.
*   **.desktop File:** I've noted your feedback and added `.desktop` file directly in the upstream package.

I believe the manifest is now complete and ready for final approval. Please let me know if there is anything else I've missed.

net.bontal.Catalyst.yml
```diff
@@ -0,0 +1,65 @@
...
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --device=dri
+
+  - --share=network
+
+  - --talk-name=org.freedesktop.secrets
+
+  - --talk-name=org.freedesktop.GeoClue2
+  - --system-talk-name=org.freedesktop.GeoClue2
```
bbhtt: what are these 3 for?

net.bontal.Catalyst.yml
```diff
@@ -0,0 +1,65 @@
...
+sdk: org.freedesktop.Sdk
+command: catalyst
+
+finish-args:
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --device=dri
+
+  - --share=network
```
bbhtt, line 13: why do you need network?

net.bontal.Catalyst.metainfo.xml
```diff
@@ -0,0 +1,97 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>net.bontal.Catalyst</id>
+
+  <metadata_license>CC0-1.0</metadata_license>
+
+  <project_license>LicenseRef-proprietary=https://catalyst.bontal.net/license</project_license>
```
bbhtt: Image

bbhtt (changes requested):
Metainfo, desktopfile and icon must be inside the deb. These can be used by distributions too.

hfiguiere:
> **Architecture:** The build is restricted to `x86_64` as aarch64 binaries are not yet available.

No it's not. The build will just fail.

net.bontal.Catalyst.yml
pewriebontal: The network permission is for the Al task parsing feature that connects to online APls such as Gemini or user's chosen API.

The app also works completely offline.

net.bontal.Catalyst.yml
pewriebontal: The app uses the flutter_secure_storage plugin to handle sensitive data. The libsecret dependency is required for the app to function correctly.

Geolocation access allows users to add a location to tasks by choosing from a map, It is also required to autofill the location during setup because one of the Flutter plugins depends on Geoclue.

bbhtt: Why do you need both system and session for geolocation?

bbhtt: > flutter_secure_storage

Does this support the secrets portal?

net.bontal.Catalyst.metainfo.xml
pewriebontal: My apologies, I haven't updated the website yet, so the link is indeed still broken. I'll get the license page live by tonight SGT.

net.bontal.Catalyst.yml
pewriebontal: **GeoClue2 - Both permissions needed:**
I tested this and both are actually required. Without both session and system bus access, the location picker crashes when trying to get the user's current position. I believed some distros need one, some need the other, so I need both for compatibility.

**flutter_secure_storage portal support:**
Yes, I tested it and it does support the secrets portal! It works fine with `--talk-name=org.freedesktop.portal.secrets`.

bbhtt: > It works fine with --talk-name=org.freedesktop.portal.secrets.

You don't need that, it is granted by default.

pewriebontal: noted

bbhtt: Also Geoclue doesn't run in session not sure what you are testing. The same service can't be running as both system and session.

bbhtt: And how did you test secrets portal support? I can see no mention of it in the flutter_secure_storage repo.

pewriebontal:
> Metainfo, desktopfile and icon must be inside the deb. These can be used by distributions too.

Just to be perfectly clear about your feedback: if I include the metainfo.xml, desktopfile, and icon directly inside the application's release tarball (in their standard share/ locations), should I then:
1. Remove them as type: file entries from the Flatpak manifest's sources section?
2. And also remove those individual files from this Flatpak manifest repository?

I want to make sure I get this fully correct. Thank you!

bbhtt:
Yes correct.

pewriebontal:
Hi @hfiguiere and @bbhtt,

I have pushed a new commit that addresses all remaining feedback.
Thank you both for your patience and incredibly thorough guidance.

net.bontal.Catalyst.metainfo.xml
pewriebontal: updated the website also

bbhtt:
Is secret storing working now since you removed the permission? I didn't get an answer to https://github.com/flathub/flathub/pull/6677#discussion_r2179042502

bbhtt:
bot, build

pewriebontal:
> Is secret storing working now since you removed the permission? I didn't get an answer to [#6677 (comment)](https://github.com/flathub/flathub/pull/6677#discussion_r2179042502)

@bbhtt Yes, the app works perfectly fine.

PR #6679, opened 2025-07-01
Add com.bishwasaha.Koncentro

kun-codes:

Koncentro is a powerful productivity app combining Pomodoro technique, task management, and website blocking.

Features:
   1. Timeboxing: Set fixed durations for tasks to boost focus and efficiency.
   2. Website Blocker: Block distracting sites using customizable blocklists or allowlists.
   3. Workspaces: Separate work and personal tasks with dedicated Pomodoro timers, blockers, and task lists.
   4. Fluent Design: Built with Microsoft’s Fluent Design.
  
https://github.com/user-attachments/assets/cc746475-ca2d-49ff-b49b-6257b8bee547

### DIFF
--- a/com.bishwasaha.Koncentro.yaml
+++ b/com.bishwasaha.Koncentro.yaml
@@ -85,7 +85,7 @@ modules:
     -   name: mitmproxy
         buildsystem: simple
         build-commands:
-            - tar -xf mitmproxy/mitmproxy-12.1.1-linux-x86_64.tar.gz -C mitmproxy || tar -xf mitmproxy/mitmproxy-12.1.1-linux-aarch64.tar.gz -C mitmproxy
+            - tar -xf mitmproxy/mitmproxy.tar.gz -C mitmproxy
             - install -Dm755 mitmproxy/mitmdump ${FLATPAK_DEST}/koncentro/mitmdump
 
         sources:
@@ -93,6 +93,7 @@ modules:
                 url: https://downloads.mitmproxy.org/12.1.1/mitmproxy-12.1.1-linux-x86_64.tar.gz
                 sha256: cdbc5a92c87c63bb8fe13552b46d2d0afaa762193ddabe4c96df8f59690ffd76
                 dest: mitmproxy
+                dest-filename: mitmproxy.tar.gz
                 only-arches:
                     - x86_64
                 x-checker-data:
@@ -107,6 +108,7 @@ modules:
                 url: https://downloads.mitmproxy.org/12.1.1/mitmproxy-12.1.1-linux-aarch64.tar.gz
                 sha256: d1f485cea46a51917c6505b4e9bb9a3927daad80d86750e46320bd211af4e6a4
                 dest: mitmproxy
+                dest-filename: mitmproxy.tar.gz
                 only-arches:
                     - aarch64
                 x-checker-data:
--- a/com.bishwasaha.Koncentro.yaml
+++ b/com.bishwasaha.Koncentro.yaml
@@ -73,9 +73,9 @@ modules:
 
         sources:
             -   type: archive
-                url: https://github.com/kun-codes/Koncentro/archive/refs/tags/v0.7.2.tar.gz
+                url: https://github.com/kun-codes/Koncentro/archive/refs/tags/v0.8.0.tar.gz
                 dest: koncentro
-                sha256: e8870510f705feeeb027719612cc84cfa0f0ec0be26a1c2f247795a87a6e659a
+                sha256: 600392960042002b36a3d547a0681db732d9127af6bdf3edeb3d5580fb4bf00c
 
             -   type: script
                 commands:
--- a/com.bishwasaha.Koncentro.yaml
+++ b/com.bishwasaha.Koncentro.yaml
@@ -1,0 +1,118 @@
+app-id: com.bishwasaha.Koncentro
+runtime: org.kde.Platform
+runtime-version: '6.9'
+sdk: org.kde.Sdk
+base: io.qt.PySide.BaseApp
+base-version: '6.9'
+build-options:
+    env:
+        - BASEAPP_REMOVE_WEBENGINE=1
+        - BASEAPP_DISABLE_NUMPY=1
+command: run.sh
+rename-desktop-file: koncentro.desktop
+rename-icon: koncentro
+finish-args:
+    - --socket=fallback-x11
+    - --socket=wayland
+    - --share=network
+    - --socket=pulseaudio
+    - --share=ipc
+    - --device=dri
+    - --talk-name=org.kde.StatusNotifierWatcher
+    - --talk-name=org.freedesktop.Notifications  # for notifications
+    - --talk-name=org.freedesktop.Flatpak  # to run commands on host
+    - --env=PYTHONPATH=/app/koncentro/
+    - --filesystem=xdg-config/environment.d/:create  # to set environment variables on host
+    - --filesystem=home/.mitmproxy/:create  # mitmproxy stores its certificates here
+cleanup-commands:
+    - /app/cleanup-BaseApp.sh
+cleanup:
+    - '*.a'
+    - /include
+    - /lib/cmake
+    - /lib/pkgconfig
+    - /share/man
+
+modules:
+    - python3-certifi.yaml
+    - python3-loguru.yaml
+    - python3-psutil.yaml
+    - python3-pyside6-fluent-widgets.yaml
+    - python3-semver.yaml
+    - python3-sqlalchemy.yaml
+    - python3-uniproxy.yaml
+    - python3-validators.yaml
+    - python3-alembic.yaml
+
+    -   name: koncentro
+        buildsystem: simple
+        build-commands:
+            - python -m compileall .
+
+            - desktop-file-edit --set-key="Exec" --set-value="run.sh" "koncentro/installer-scripts/linux/common/koncentro.desktop"
+            - desktop-file-edit --set-key="Icon" --set-value="/usr/share/icons/hicolor/512x512/koncentro/koncentro" "koncentro/installer-scripts/linux/common/koncentro.desktop"
+            - desktop-file-edit --set-key="Categories" --set-value="Utility;Office;" "koncentro/installer-scripts/linux/common/koncentro.desktop"
+
+            - install -Dm644 "koncentro/installer-scripts/linux/common/koncentro.desktop" ${FLATPAK_DEST}/share/applications/koncentro.desktop
+            - install -Dm644 koncentro/installer-scripts/linux/common/com.bishwasaha.Koncentro.metainfo.xml ${FLATPAK_DEST}/share/metainfo/com.bishwasaha.Koncentro.metainfo.xml
+
+            - install -Dm644 koncentro/assets/logo_512x512.png ${FLATPAK_DEST}/share/icons/hicolor/512x512/apps/koncentro.png
+            - install -Dm644 koncentro/assets/logo_256x256.png ${FLATPAK_DEST}/share/icons/hicolor/256x256/apps/koncentro.png
+            - install -Dm644 koncentro/assets/logo_128x128.png ${FLATPAK_DEST}/share/icons/hicolor/128x128/apps/koncentro.png
+            - install -Dm644 koncentro/assets/logo_64x64.png ${FLATPAK_DEST}/share/icons/hicolor/64x64/apps/koncentro.png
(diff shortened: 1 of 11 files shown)


com.bishwasaha.Koncentro.yaml
```diff
@@ -0,0 +1,118 @@
+app-id: com.bishwasaha.Koncentro
+runtime: org.kde.Platform
+runtime-version: '6.9'
+sdk: org.kde.Sdk
+base: io.qt.PySide.BaseApp
+base-version: '6.9'
+build-options:
+    env:
+        - BASEAPP_REMOVE_WEBENGINE=1
+        - BASEAPP_DISABLE_NUMPY=1
```
bbhtt, line 10:
```suggestion
build-options:
  env:
    - BASEAPP_REMOVE_WEBENGINE=1
    - BASEAPP_DISABLE_NUMPY=1
```

com.bishwasaha.Koncentro.yaml
```diff
@@ -0,0 +1,118 @@
...
+rename-icon: koncentro
+finish-args:
+    - --socket=fallback-x11
+    - --socket=wayland
+    - --share=network
+    - --socket=pulseaudio
+    - --share=ipc
+    - --device=dri
+    - --talk-name=org.kde.StatusNotifierWatcher
+    - --talk-name=org.freedesktop.Notifications  # for notifications
```
bbhtt, line 22: this does not go through the portal?

com.bishwasaha.Koncentro.yaml
```diff
@@ -0,0 +1,118 @@
...
+    - --talk-name=org.freedesktop.Flatpak  # to run commands on host
```
bbhtt, line 23: what commands you need to run and why? this requires an [exception](https://docs.flathub.org/docs/for-app-authors/linter#exceptions).

com.bishwasaha.Koncentro.yaml
```diff
@@ -0,0 +1,118 @@
...
+    - --env=PYTHONPATH=/app/koncentro/
+    - --filesystem=xdg-config/environment.d/:create  # to set environment variables on host
```
bbhtt, line 25: what env vars do you have to set?

com.bishwasaha.Koncentro.yaml
```diff
@@ -0,0 +1,118 @@
...
+
+            -   type: script
+                commands:
+                    - python /app/koncentro/src/__main__.py "$@"
+                dest-filename: run.sh
+
+    -   name: mitmproxy
+        buildsystem: simple
+        build-commands:
+            - tar -xf mitmproxy/mitmproxy-12.1.1-linux-x86_64.tar.gz -C mitmproxy || tar -xf mitmproxy/mitmproxy-12.1.1-linux-aarch64.tar.gz -C mitmproxy
```
bbhtt: add a `dest-filename: mitmproxy.tar.gz` to the sources. drop this dual install comamnd.

com.bishwasaha.Koncentro.yaml
bbhtt: the file has some extra indentation and spaces going on in other places as well.

kun-codes: I have formatted the main file with 4 spaces of indentation. Its easier on the eyes for me

com.bishwasaha.Koncentro.yaml
kun-codes: nope, without it the notifications don't come through

com.bishwasaha.Koncentro.yaml
kun-codes: My app uses mitmproxy as a website blocker. Commands like `kwriteconfig6`, `kreadconfig6`and `gsettings` are needed to set the proxy url on the host machine via the desktop environment's settings. This can't be done within the flatpak sandbox

KDE uses `kwriteconfig6` and `kreadconfig6`, gnome and cinnamon use `gsettings`

com.bishwasaha.Koncentro.yaml
kun-codes: `HTTPS_PROXY`, `https_proxy`, `http_proxy` and `HTTP_PROXY` to ensure that the user doesn't bypass block domains throught the terminal as terminal tools use these environment variables for proxy settings

com.bishwasaha.Koncentro.yaml
kun-codes: done

kun-codes:
@bbhtt please have a look, I have responded to all your queries

com.bishwasaha.Koncentro.yaml
bbhtt: from https://github.com/flathub/flathub/pull/6679#discussion_r2179294239 and this it feels like this shouldn't be running inside a sandbox then.

kun-codes: that feature can be turned off for terminal apps then, but for the gui apps commnads like `kwriteconfig6`, `kreadconfig6` and `gsettings` are needed

Setting proxy manually everytime would be inconvenient for the user

com.bishwasaha.Koncentro.yaml
kun-codes: @bbhtt should I turn off setting environment variables?

bbhtt: this part is probably fine but needs an exception, the main concern is about flatpak-spawn.

kun-codes: These lines of code are from a library which I am using for my app
https://github.com/kun-codes/Uniproxy/blob/9d078ca0f877046d26333f878b46b58759f9d730/uniproxy/linux_proxy.py#L283
https://github.com/kun-codes/Uniproxy/blob/9d078ca0f877046d26333f878b46b58759f9d730/uniproxy/linux_proxy.py#L292

2nd line runs `kwriteconfig6`, `kreadconfig6` and `gsettings` to set proxy and 1st line runs a command to refresh environment variables. This library is written by me and has couple of thousand downloads: https://pypistats.org/packages/uniproxy

kun-codes:
@bbhtt should I request an [exception](https://docs.flathub.org/docs/for-app-authors/linter#exceptions) via a PR?

bbhtt:
As I said I don't think part of the features that require using gsettings etc. aren't a good fit for Flatpak. You are losing the most important benefit of Flatpak-ing it — the sandboxing.

As it is, it is exactly an unsandboxed package.

kun-codes:
I understand your concerns regarding the sandboxing principles of flatpak. But I request you to reconsider an exception.

I agree that sandboxing is core feature of flatpaks. But I have noticed that there are over 180 apps on flathub which make use of `flatpak spawn --host` which includes terminal emulators which can execute arbitary commands on host. My app only invokes few specific commands for the purpose of setting proxy configurations via the user's desktop environment.

Let me know your thoughts. I appreciate your work you do to maintain the quality of flathub.

bbhtt:
Many of those are "development tools", yours is not and we stopped accepting such development tools in some cases (see the inclusion policy).

Also just because there are some, doesn't mean we have to keep increasing the number though.

bbhtt:
Is flatpak-spawn entirely for website blocking or do other features depend on it?

How effective is that actually? If you are writing and storing configs in user's home directory someone can easily bypass them.

Something like `/etc/hosts` is slightly better in case of parental controls etc. but also not effective and not possible to do from a Flatpak.

bbhtt:
The main problem here is this starts out with needing to execute gsettings on host, systemctl on host etc. As more features are added later on, I suspect the dependency on things on host will likely increase and the sandbox at times will likely be an obstruction.

kun-codes:
Thanks for your response

- `flatpak-spawn` is used only for website blocking, not for any other functionality
- The approach is quite effective, as it uses a real proxy (via mitmproxy), making it system-wide and browser-agnostic. It doesn't rely on editing `/etc/hosts`, as it would require require root and isn’t feasible for a rootless app
- I’m not writing configs into the user’s home directory. The app stores its own config inside the flatpak directory at `~/.var/app/`. The only thing written outside that is temporary proxy-related environment variables in `~/environment.d/`, and even those are cleaned up automatically at the end of each work session or when the timer stops
- Regarding the usage of `flatpak-spawn` for other commands, I want to assure you that I do not intend to expand the use of `flatpak-spawn` beyond this use case. All the needed binaries (like `mitmdump`) are already bundled inside the flatpak


Please let me know if I should put in a warning in the flatpak metainfo regarding this so the user if more informed

bbhtt:
Apart from relying on host gsettings, there seems to be another problem https://github.com/kun-codes/Uniproxy/blob/9d078ca0f877046d26333f878b46b58759f9d730/uniproxy/linux_proxy.py#L17-L18

It only seems to support GNOME, KDE, and Cinnamon?

kun-codes:
Hi @bbhtt

The reason my library currently supports gnome, kde and cinnamon is that they provide dedicated terminal commands for setting proxy settings which gui applications pick up immediately without requiring a logout/login.

On the other hand, linux window managers don't have an api to set system wide proxy settings. While I can set http_proxy, https_proxy environment variables, and they will work for terminal applciations. Gui applications won't respect them unless the desktop environment itself propagates the settings

I am working towards adding support for other DEs like budgie, MATE and unity and they would be included before the first stable release of the app. Currently the app is in pre-release.

Logging out and logging in just to apply proxy settings would defeat the purpose of the productivity app

kun-codes:
@bbhtt I have added support for 4 more desktops which now has expanded the list of supported DEs to `kde`, `gnome`, `cinnamon`, `mate`, `budgie`, `unity` and `pantheon`

bbhtt:
bot, build

kun-codes:
It says `appstream-missing-screenshots` but the screenshots at https://github.com/kun-codes/Koncentro/blob/61393042c1beb700b8154ee837e11b7d262925d6/installer-scripts/linux/common/com.bishwasaha.Koncentro.metainfo.xml#L52 are reachable

bbhtt:
It needs to be direct links, raw.githubuser... not links to html pages.

kun-codes:
my bad, I will edit it right now, and release a patch fix release

kun-codes:
@bbhtt do you know what `appstream-failed-validation` means?

bbhtt:
Try running the linter locally it will show the error.

bbhtt:
It's supposed to output the detailed result of appstream validation but looks like it was lost when I switched it to print CI annotation instead of JSON. I added it back

bot, build

bbhtt:
also remove the x-checker from Python modules, these will end up generating a lot of unnecessary PRs.

bbhtt:
You are missing [releases](https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines#release) info in appstream.

kun-codes:
@bbhtt I have fixed the screenshot issue, I have also added release tag version

kun-codes:
> also remove the x-checker from Python modules, these will end up generating a lot of unnecessary PRs.

How would I get automated updates to my flatpak build then?

bbhtt:
Manually update them, your python deps should match the version you are developing against in poetry.lock.

bbhtt:
Also if you don't already have it, add a button in the setup screen and settings to "roll back" all the proxy changes ie. deleting ~/.config/environment.d/no-proxy.conf` etc.

kun-codes:
> Manually update them, your python deps should match the version you are developing against in poetry.lock.

My dependencies are always at the latest versions. Is it okay if I keep x-checker? I would have to manually approve pull requests anyways so there won't be any automatic merges

kun-codes:
> Also if you don't already have it, add a button in the setup screen and settings to "roll back" all the proxy changes ie. deleting ~/.config/environment.d/no-proxy.conf` etc.

The proxy variables are reset everytime the app closes. They are also reset once the setup is finished.

I would be adding a reset proxy settings button to the app settings

bbhtt:
> My dependencies are always at the latest versions

I doubt that policy works in real life, anyways.

kun-codes:
> > My dependencies are always at the latest versions
> 
> I doubt that policy works in real life, anyways.

if any dependency has to be pinned at a certain version in future, I would either close the PR or enter a static version number in the yaml

bbhtt:
you will need to remove the x-checker in that case otherwise it will keep sending the update.

kun-codes:
> you will need to remove the x-checker in that case otherwise it will keep sending the update.

yes, I would use [this](https://github.com/flathub-infra/flatpak-external-data-checker?tab=readme-ov-file#version-constraining)

kun-codes:
@bbhtt I have included a reset proxy button now

image

bot, build

PR #6688, opened 2025-07-05
dev.fredol.fredstalker

Fredolx:

- [ X] Please describe the application briefly. < IPTV app designed for stalker (a type of iptv service) >
- [ X] Please attach a video showcasing the application on Linux using the Flatpak. < [Screencast_20250705_033936.webm](https://github.com/user-attachments/assets/373b7443-962f-48c4-8a57-833930146c1e) >
- [ X] The Flatpak ID follows all the rules listed in the [Application ID requirements][appid].
- [ X] I have read and followed all the [Submission requirements][reqs] and the [Submission guide][reqs2] and I agree to them.
- [ X] I am the author of the project.

### DIFF
--- a/dev.fredol.fredstalker.yaml
+++ b/dev.fredol.fredstalker.yaml
@@ -1,0 +1,474 @@
+id: dev.fredol.fredstalker
+
+runtime: org.gnome.Platform
+runtime-version: "48"
+sdk: org.gnome.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.mono6
+build-options:
+  append-path: /usr/lib/sdk/flutter/bin
+
+command: fredstalker
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --device=dri
+  - --share=ipc
+  - --share=network
+  - --socket=pulseaudio
+  - --env=XCURSOR_PATH=/run/host/user-share/icons:/run/host/share/icons
+  - --env=WEBKIT_DISABLE_DMABUF_RENDERER=1
+  - --talk-name=org.gnome.SessionManager
+
+cleanup:
+  - "*.la"
+  - "*.a"
+  - "/include"
+  - "/lib/cmake"
+  - "/lib/pkgconfig"
+  - "/share/bash-completion"
+  - "/share/cmake"
+  - "/share/doc"
+  - "/share/man"
+  - "/share/zsh"
+
+modules:
+  - name: lmdb
+    sources:
+      - sha256: f3927859882eb608868c8c31586bb7eb84562a40a6bf5cc3e13b6b564641ea28
+        type: archive
+        url: https://github.com/LMDB/lmdb/archive/LMDB_0.9.22.tar.gz
+    make-install-args:
+      - prefix=/app
+    no-autogen: true
+    subdir: libraries/liblmdb
+
+  - name: libXmu
+    buildsystem: autotools
+    sources:
+      - type: git
+        url: https://gitlab.freedesktop.org/xorg/lib/libxmu.git
+        tag: libXmu-1.2.1
+        commit: 792f80402ee06ce69bca3a8f2a84295999c3a170
+
+  - name: xclip
+    buildsystem: autotools
+    sources:
+      - type: git
+        url: https://github.com/astrand/xclip.git
+        tag: "0.13"
+        commit: 9aa7090c3b8b437c6489edca32ae43d82e0c1281
+
+  - name: libXpresent
+    buildsystem: autotools
+    sources:
+      - type: git
+        url: https://gitlab.freedesktop.org/xorg/lib/libxpresent.git
+        tag: libXpresent-1.0.1
+        commit: 37507b5f44332accfb1064ee69a4f6a833994747
+
+  - name: luajit
+    no-autogen: true
+    cleanup:
+      - /bin
+      - /include
+      - /lib/pkgconfig
+      - /share/man
+    sources:
+      - type: git
+        url: https://github.com/LuaJIT/LuaJIT.git
+        mirror-urls:
+          - https://luajit.org/git/luajit.git
+        disable-shallow-clone: true
+        commit: 538a82133ad6fddfd0ca64de167c4aca3bc1a2da
+      - type: shell
+        commands:
+          - sed -i 's|/usr/local|/app|' ./Makefile
+
+  - name: yt-dlp
+    no-autogen: true
+    no-make-install: true
+    make-args:
+      - yt-dlp
+      - PYTHON=/usr/bin/python3
+    post-install:
+      - install yt-dlp /app/bin
+    sources:
+      - type: archive
+        url: https://github.com/yt-dlp/yt-dlp/releases/download/2025.03.27/yt-dlp.tar.gz
+        sha256: 2a269ae3e9a4510a09c48970e6152223b0bd086c47e55291befda39c070f3756
+
+  - name: uchardet
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DBUILD_STATIC=0
+    cleanup:
+      - /bin
+      - /include
+      - /lib/pkgconfig
+      - /share/man
+    sources:
+      - type: archive
+        url: https://www.freedesktop.org/software/uchardet/releases/uchardet-0.0.8.tar.xz
+        sha256: e97a60cfc00a1c147a674b097bb1422abd9fa78a2d9ce3f3fdcc2e78a34ac5f0
+
+  - name: libass
+    cleanup:
+      - /include
+      - /lib/pkgconfig
+    config-opts:
+      - --disable-static
+      - --enable-asm
+      - --enable-harfbuzz
+      - --enable-fontconfig
+    sources:
+      - type: git
+        url: https://github.com/libass/libass.git
+        tag: 0.17.3
+        commit: e46aedea0a0d17da4c4ef49d84b94a7994664ab5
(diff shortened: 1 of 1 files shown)


dev.fredol.fredstalker.yaml
```diff
@@ -0,0 +1,474 @@
...
+        url: https://github.com/Fredolx/fredstalker
+        commit: cebbbf2c3e639a295494063462322ee3640e4fda
+        tag: v1.0.0
+      - type: archive
+        url: https://github.com/Fredolx/FredStalker/releases/download/v1.0.0/fredstalker.tar.gz
+        sha256: 47946f67f4b564675e8a5d870c2686b1970c599284882109ecf846c84319afbf
+        dest: compiled
+    build-commands:
+      - install -Dm755 compiled/fredstalker /app/bin/fredstalker
+      - cp -r compiled/. /app/bin
```
hfiguiere: so what happen on the architecture this is not compiled for? why not building from source?

dev.fredol.fredstalker.yaml
```diff
@@ -0,0 +1,474 @@
+id: dev.fredol.fredstalker
+
+runtime: org.gnome.Platform
+runtime-version: "48"
+sdk: org.gnome.Sdk
```
hfiguiere: why does this need the GNOME runtime?

dev.fredol.fredstalker.yaml
```diff
@@ -0,0 +1,474 @@
...
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.mono6
```
hfiguiere: I don't see where it needs mono

dev.fredol.fredstalker.yaml
```diff
@@ -0,0 +1,474 @@
...
+      - -Dlibarchive=enabled
+      - -Dsdl2=enabled
+      - -Dvulkan=enabled
+      - -Dshaderc=disabled
+      # - -Dvulkan-interop=enabled
+    cleanup:
+      - /include
+      - /lib/pkgconfig
+    post-install:
+      - echo "screenshot-directory=~/.var/app/dev.fredol.fredstalker/Pictures/mpv" > /app/etc/mpv/mpv.conf
```
hfiguiere: this assumption is incorrect.

dev.fredol.fredstalker.yaml
Fredolx: I've looked at other manifests and they all don't build from source for flutter apps. I believe it's because it must be hard to compile all of flutter from source? I need to look if there is a tool like with rust and cargo to download all the packages offline.

dev.fredol.fredstalker.yaml
Fredolx: Pretty sure flutter uses gtk3

dev.fredol.fredstalker.yaml
Fredolx: I'll try removing it

dev.fredol.fredstalker.yaml
hfiguiere: The freedesktop SDK has gtk3.

dev.fredol.fredstalker.yaml
hfiguiere: Two wrongs don't make a right.

That doesn't even answer the first question.

Fredolx: I could make an archive for arm64 too.
I've never seen a single flutter app compiled from source and
doesn't have its package manager.

Spottube, a popular flutter app on linux, uses this approach.
https://github.com/flathub/com.github.KRTirtho.Spotube/blob/master/com.github.KRTirtho.Spotube.yml

It uses the gnome sdk and has two archives, one for x86 and one for arm.

bbhtt: >  and
doesn't have its package manager.

Look at the readme on that repo more closely.

bbhtt:
Fix the checklists in the PR description.

bbhtt:
If you need MPV use the MPV Flatpak as `base` and/or use the ffmpeg-full extension from runtime.

dev.fredol.fredstalker.yaml
```diff
@@ -0,0 +1,474 @@
...
+    cleanup:
+      - /bin
+      - /include
+      - /lib/pkgconfig
+      - /share/man
+    sources:
+      - type: git
+        url: https://github.com/LuaJIT/LuaJIT.git
+        mirror-urls:
+          - https://luajit.org/git/luajit.git
```
bbhtt: `mirror-urls` is only valid for archive sources. you inherited the mistakes from wherever these were copy pasted.

dev.fredol.fredstalker.yaml
```diff
@@ -0,0 +1,474 @@
...
+
+  - name: libXpresent
+    buildsystem: autotools
+    sources:
+      - type: git
+        url: https://gitlab.freedesktop.org/xorg/lib/libxpresent.git
+        tag: libXpresent-1.0.1
+        commit: 37507b5f44332accfb1064ee69a4f6a833994747
+
+  - name: luajit
```
bbhtt: this is in shared-modules https://github.com/flathub/shared-modules

dev.fredol.fredstalker.yaml
bbhtt: at the moment this is blocked on https://docs.flathub.org/docs/for-app-authors/requirements#building-from-source, as you have seen in the readme of the flatpak-flutter project there are quite a few flutter apps building from source on Flathub.

I recommend that you try first and if you can't get it to work due to specific issues, this can be reconsidered.

Fredolx:
@bbhtt @hfiguiere 
I applied all your recommendations to the manifest. Here's all the changes:
- Compiles entirely from source thanks to flatpak-flutter
- Uses mpv as base
- Uses the freedesktop sdk and extensions instead of gnome-sdk
- Mono was removed from deps

Fredolx:
bot, build

Fredolx:
bot, build

Fredolx:
bot, build

dev.fredol.fredstalker.yaml
```diff
@@ -0,0 +1,96 @@
...
+            dest: .pub-cache/hosted/pub.dev/sqlite3_flutter_libs-0.5.34
+      modules:
+          - flutter-sdk-3.32.5.json
+      build-commands:
+          - setup-flutter.sh -C .
+          - flutter build linux --release
+          - install -D $BUNDLE_PATH/fredstalker /app/bin/fredstalker
+          - cp -r $BUNDLE_PATH/lib /app/bin/lib
+          - cp -r $BUNDLE_PATH/data /app/bin/data
+          - install -Dm755 build/linux/x64/release/bundle/fredstalker /app/bin/fredstalker
```
bbhtt: BUNDLE_PATH

libmpv.pc
```diff
@@ -0,0 +1,10 @@
+prefix=/app
+exec_prefix=${prefix}
```
bbhtt, line 2: this should not be here

dev.fredol.fredstalker.yaml
```diff
@@ -0,0 +1,96 @@
+id: dev.fredol.fredstalker
+runtime: org.freedesktop.Platform
+runtime-version: "24.08"
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+    - org.freedesktop.Sdk.Extension.llvm19
+build-options:
```
bbhtt, line 7: you need to add ffmpeg-full for extra codecs.

bbhtt: https://docs.flatpak.org/en/latest/extension.html#loading-existing-extensions

libmpv.pc
Fredolx: are you talking about the whole file? Mpv base removes the pkgconfig and so my build fails without this custom libmpv.pc file

Fredolx:
```
Running: flutter build linux --release
CMake Error in flutter/ephemeral/.plugin_symlinks/media_kit_video/linux/CMakeLists.txt:
  Imported target "PkgConfig::mpv" includes non-existent path

    "/include"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.



CMake Generate step failed.  Build files cannot be regenerated correctly.
```

Fredolx:
I added the ffmpeg extension and fixed the bundle path for ARM

Fredolx:
bot, build

libmpv.pc
bbhtt: So your app requires MPV to compile? I didn't know that, then it is better to build MPV here.

Fredolx: Yes, it requires libmpv to build mediakit in flutter. 
Using mpv as base isn't a bad idea. It saves a ton of work from maintaining this flatpak and all we need is a .pc file to make it work.

bbhtt: it will get outdated quickly, you can ask the mpv maintainer to keep the pc file in.

Fredolx: I agree. Can we have this temporarily while we check with MPV's maintainer?

bbhtt: open an issue or a pr, all you need is remove the include and pkgconfig cleanups from there. it should be resolved before merge.

Fredolx: Ok sure, I'll make it

Fredolx: https://github.com/flathub/io.mpv.Mpv/issues/676

PR #6707, opened 2025-07-10
Add eu.retroshine.RetroShine

yellowdolphin:

- [X] Please describe the application briefly. An image batch processing tool that enhances old/scanned photos with automatic color correction
- [N/A] Please attach a video showcasing the application on Linux using the Flatpak. No video is available.


files/icon128.png
hfiguiere, line 1: remove this it's not needed.

eu.retroshine.RetroShine.yaml
```diff
@@ -0,0 +1,72 @@
...
+  - name: retroshine-icon
+    buildsystem: simple
+    build-commands:
+      - install -Dm644 icon128.png /app/share/icons/hicolor/128x128/apps/eu.retroshine.RetroShine.png
+      - install -Dm644 icon256.png /app/share/icons/hicolor/256x256/apps/eu.retroshine.RetroShine.png
+      - install -Dm644 /usr/share/icons/hicolor/index.theme /app/share/icons/hicolor/index.theme
+      - gtk-update-icon-cache -f /app/share/icons/hicolor
+    sources:
+      - type: file
+        path: files/icon128.png
```
hfiguiere:
```suggestion
```

eu.retroshine.RetroShine.yaml
```diff
@@ -0,0 +1,72 @@
...
+    buildsystem: simple
+    build-commands:
+      - install -Dm644 eu.retroshine.RetroShine.metainfo.xml /app/share/metainfo/eu.retroshine.RetroShine.metainfo.xml
+    sources:
+      - type: file
+        path: metainfo/eu.retroshine.RetroShine.metainfo.xml
+  - name: retroshine-icon
+    buildsystem: simple
+    build-commands:
+      - install -Dm644 icon128.png /app/share/icons/hicolor/128x128/apps/eu.retroshine.RetroShine.png
```
hfiguiere:
```suggestion
```

eu.retroshine.RetroShine.yaml
```diff
@@ -0,0 +1,72 @@
...
+      - install -Dm644 icon256.png /app/share/icons/hicolor/256x256/apps/eu.retroshine.RetroShine.png
+      - install -Dm644 /usr/share/icons/hicolor/index.theme /app/share/icons/hicolor/index.theme
+      - gtk-update-icon-cache -f /app/share/icons/hicolor
+    sources:
+      - type: file
+        path: files/icon128.png
+      - type: file
+        path: files/icon256.png
+  - name: retroshine-desktop
```
hfiguiere: this doesn't need to be a separate module.

eu.retroshine.RetroShine.yaml
```diff
@@ -0,0 +1,72 @@
...
+          #- chmod +x /app/bin/RetroShine
+
+  - name: metainfo
+    buildsystem: simple
+    build-commands:
+      - install -Dm644 eu.retroshine.RetroShine.metainfo.xml /app/share/metainfo/eu.retroshine.RetroShine.metainfo.xml
+    sources:
+      - type: file
+        path: metainfo/eu.retroshine.RetroShine.metainfo.xml
+  - name: retroshine-icon
```
hfiguiere: this doesn't need to be a separate module.

eu.retroshine.RetroShine.yaml
```diff
@@ -0,0 +1,72 @@
...
+          # this is current dir
+          # but FLATPAK_DEST is not set during install (only build)
+          # install with option -v to see output
+          - tar xfz RetroShine-1.1.1-x86_64.tar.gz
+          - ls -l RetroShine/RetroShine
+          # use softlink instead of this:
+          #- cp -r RetroShine/* /app/bin/
+          #- chmod +x /app/bin/RetroShine
+
+  - name: metainfo
```
hfiguiere: this doesn't need to be a separate module.

eu.retroshine.RetroShine.yaml
```diff
@@ -0,0 +1,72 @@
...
+      - install -Dm755 apply_extra ${FLATPAK_DEST}/bin/apply_extra
+    sources:
+      - type: extra-data
+        filename: RetroShine-1.1.1-x86_64.tar.gz
+        only-arches:
+          - x86_64
+        # Problem: "url" may cause connection error, "path" is not supported by type "extra-data", only by type "file".
+        #url: https://retroshine.eu/download/RetroShine-1.1.1-x86_64.tar.gz
+        #url: https://drive.usercontent.google.com/download?id=1sbMpnBxTvT8uHWpdxaJ580iGsqG4CbBZ&export=download&confirm
+        url: https://github.com/yellowdolphin/retroshine/releases/download/v1.1.1/RetroShine-1.1.1-x86_64.tar.gz
```
hfiguiere: this is not upstream. It's a problem.

eu.retroshine.RetroShine.yaml
```diff
@@ -0,0 +1,72 @@
...
+  - --socket=fallback-x11
+  - --share=ipc
+  - --device=dri
+  - --persist=RetroShine
+
+modules:
+  - name: retroshine-binary
+    buildsystem: simple
+    build-commands:
+      - mkdir ${FLATPAK_DEST}/bin
```
hfiguiere: if you install the apply_extra first you don't need to do that.

eu.retroshine.RetroShine.yaml
```diff
@@ -0,0 +1,72 @@
...
+        size: 222648883
+      - type: script
+        dest-filename: apply_extra
+        commands:
+          # where does the downloaded files go? -> /app/extra/
+          # this is current dir
+          # but FLATPAK_DEST is not set during install (only build)
+          # install with option -v to see output
+          - tar xfz RetroShine-1.1.1-x86_64.tar.gz
+          - ls -l RetroShine/RetroShine
```
hfiguiere:
```suggestion
```

eu.retroshine.RetroShine.yaml
```diff
@@ -0,0 +1,72 @@
...
+      - mkdir ${FLATPAK_DEST}/bin
+      # this fails: folder "bin" does not exist!
+      - ln -s ${FLATPAK_DEST}/extra/RetroShine/RetroShine ${FLATPAK_DEST}/bin/RetroShine
+      - install -Dm755 apply_extra ${FLATPAK_DEST}/bin/apply_extra
+    sources:
+      - type: extra-data
+        filename: RetroShine-1.1.1-x86_64.tar.gz
+        only-arches:
+          - x86_64
+        # Problem: "url" may cause connection error, "path" is not supported by type "extra-data", only by type "file".
```
hfiguiere: not sure what you think `path` would do, but I didn't think it does what you think it does.

eu.retroshine.RetroShine.yaml
```diff
@@ -0,0 +1,72 @@
...
+        #url: https://retroshine.eu/download/RetroShine-1.1.1-x86_64.tar.gz
```
hfiguiere: what's wrong with this url?

bbhtt:
> Please attach a video showcasing the application on Linux using the Flatpak. No video is available.

What does that mean? You should create a video.

files/icon128.png
yellowdolphin: done

eu.retroshine.RetroShine.yaml
yellowdolphin: done

eu.retroshine.RetroShine.yaml
yellowdolphin: done

eu.retroshine.RetroShine.yaml
yellowdolphin: done

eu.retroshine.RetroShine.yaml
yellowdolphin: done

eu.retroshine.RetroShine.yaml
yellowdolphin: done

eu.retroshine.RetroShine.yaml
yellowdolphin: Never mind, I replaced it by retroshine.eu (upstream)

eu.retroshine.RetroShine.yaml
yellowdolphin: yep, done.

eu.retroshine.RetroShine.yaml
yellowdolphin: done

eu.retroshine.RetroShine.yaml
yellowdolphin: For debugging you really need to avoid the download during install, hence I experimented with type file + path. Anyway, I purged the comments.

eu.retroshine.RetroShine.yaml
yellowdolphin: Nothing, now it is used.
(see commit)

yellowdolphin:
I bundled all requested changes in one commit and attached the requested video to the PR comment. I hope this is OK?

metainfo/eu.retroshine.RetroShine.metainfo.xml
```diff
@@ -0,0 +1,47 @@
...
+      <li>Simple interface, no editing experience needed</li>
+      <li>Advanced options for precise manual control</li>
+    </ul>
+  </description>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>LicenseRef-proprietary=https://retroshine.eu/LICENSE.txt</project_license>
+  <url type="homepage">https://retroshine.eu</url>
+  <url type="contact">https://retroshine.eu/contact.html</url>
+  <url type="faq">https://retroshine.eu/faq.html</url>
+  <url type="bugtracker">https://github.com/yellowdolphin/autolevels/issues</url>
```
bbhtt: why is this going to something unrelated?

eu.retroshine.RetroShine.yaml
```diff
@@ -0,0 +1,47 @@
...
+modules:
+  - name: retroshine-binary
+    buildsystem: simple
+    build-commands:
+      - install -Dm755 apply_extra ${FLATPAK_DEST}/bin/apply_extra
+      - ln -s ${FLATPAK_DEST}/extra/RetroShine/RetroShine ${FLATPAK_DEST}/bin/RetroShine
+      - install -Dm644 eu.retroshine.RetroShine.metainfo.xml /app/share/metainfo/eu.retroshine.RetroShine.metainfo.xml
+      - install -Dm644 icon256.png /app/share/icons/hicolor/256x256/apps/eu.retroshine.RetroShine.png
+      - install -Dm644 /usr/share/icons/hicolor/index.theme /app/share/icons/hicolor/index.theme
+      - gtk-update-icon-cache -f /app/share/icons/hicolor
```
bbhtt: these shouldn't be needed?

eu.retroshine.RetroShine.yaml
```diff
@@ -0,0 +1,47 @@
+app-id: eu.retroshine.RetroShine
+runtime: org.kde.Platform
+runtime-version: "6.9"
+sdk: org.kde.Sdk
+command: RetroShine
+finish-args:
+  - --share=network
```
bbhtt: why does it need network?

eu.retroshine.RetroShine.yaml
```diff
@@ -0,0 +1,47 @@
...
+  - --filesystem=home  # linter error: comment on PR
```
bbhtt, line 7: why does it need home? it doesn't use filechoosers / filechooser portals?

metainfo/eu.retroshine.RetroShine.metainfo.xml
yellowdolphin: Well spotted! RetroShine is a GUI for AutoLevels, I'm the author of both. However, I already created [an own repo for RetroShine](https://github.com/yellowdolphin/retroshine), with a [README](https://github.com/yellowdolphin/retroshine/blob/main/README.md) linking to retroshine.eu. 

So, option (A) would be to replace the url by https://github.com/yellowdolphin/retroshine/issues or (B) remove it.

What do you prefer?

bbhtt: use https://github.com/yellowdolphin/retroshine

bbhtt:
if you are the author why are you using extra-data?

eu.retroshine.RetroShine.yaml
yellowdolphin: I tried the portal, but it is not transparent as file paths change, which affects UX (users see FUSE-mounted path rather than the real filesystem location). It is critical for the user to understand where the output files of the batch-processing are written. Target users are "normies" who don't know what a terminal is.

I did not find a simple, clean solution for PyQt6 or python in general that maps all the paths (recommendations welcome). It would require major code changes and add new dependencies to GIO-related libs. I would like to support the portal in a future release, but for now it is not feasible.

eu.retroshine.RetroShine.yaml
yellowdolphin: Indeed, I removed it. Earlier it fixed "E: gui-app-without-icon", but I can't reproduce that.

eu.retroshine.RetroShine.yaml
yellowdolphin: I thought it necessary to allow links in help texts, but it seems these work just fine w/o it.
I removed network.

eu.retroshine.RetroShine.yaml
bbhtt: > users see FUSE-mounted path rather than the real filesystem location

There is an api to retrieve the actual host path

https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Documents.html#org-freedesktop-portal-documents-gethostpaths

yellowdolphin:
> if you are the author why are you using extra-data?

Good question, sorry if I got that wrong, but I was made believe that this was required for large binaries. If this is wrong, I surely would prefer to drop the extra-data and use source type archive, instead. 

Does this build code look reasonable?
```
modules:
  - name: retroshine
    buildsystem: simple
    build-commands:
        - install -Dm755 RetroShine ${FLATPAK_DEST}/bin/RetroShine
        - cp -r _internals ${FLATPAK_DEST}/bin/
    ...
```

hfiguiere:
>  but I was made believe that this was required for large binaries.

where is that even written?

yellowdolphin:
> > but I was made believe that this was required for large binaries.
> 
> where is that even written?

If you really wanna know... 
your docs: "Non-redistributable sources must use extra-data."
SOTA-GPT: "Include your PyInstaller‑built executable and all required files using extra-data since your app’s source is proprietary <citing your docs>."

Your docs are human-readable, but not AI-readable. But in fact, the term "source" can be confusing!

eu.retroshine.RetroShine.yaml
yellowdolphin: Yes, I saw that. But this is pretty recent. My app is based on PyQt6 and it does not support that API yet, I would need to implement it myself. Let me assess how difficult that would be...

bbhtt:
You are encouraged to not use LLMs.

eu.retroshine.RetroShine.yaml
yellowdolphin: Looks like the portal doesn't even implement that API yet. 
In the kde 6.9 runtime, the portal has interface version 4, neither xattr nor GetHostPaths are provided.

You can check that by running
    ```gdbus introspect --session --dest org.freedesktop.portal.Documents --object-path /org/freedesktop/portal/documents```
inside a
    `flatpak run --command=sh --devel org.kde.Sdk//6.9`

Same is true for my gnome host system.

yellowdolphin: Looks like the portal doesn't even implement that API yet. 
In the kde 6.9 runtime, the portal has interface version 4, neither xattr nor GetHostPaths are provided.

You can check that by running
    ```gdbus introspect --session --dest org.freedesktop.portal.Documents --object-path /org/freedesktop/portal/documents```
inside a
    `flatpak run --command=sh --devel org.kde.Sdk//6.9`

Same is true for my gnome host system.

yellowdolphin:
> You are encouraged to not use LLMs.

Replace "use" by "trust". No using them is a doomed position, I'm still too young for that.

eu.retroshine.RetroShine.yaml
yellowdolphin: xdg-desktop-portal version [1.19.0+](https://github.com/flatpak/xdg-desktop-portal/blob/main/NEWS.md) is needed.

Oh, the LLMs all tell me that I cannot update xdg-desktop-portal inside Flatpak app or SDK — it must be updated on the host system, because portals run outside the sandbox and serve multiple apps over D-Bus. Is that correct?

Then this cannot be done before most users have their next LTS version installed (my target user uses flathub and has no clue how to install any portal).

bbhtt: Then I won't accept this until this gets done.

Also maybe stop using LLMs and read some docs written by actual humans and gather some information about the platform you are developing and releaseing for. Half of what you wrote doesn't make sense.

bbhtt: Also, users seeing the fuse mounted path instead of the actual host path should really not be a blocker to using the portal.

Users are going to open the file through the filchooser — then they already know where the file is and the filechooser shows the actual path.

yellowdolphin: You make very strong assumptions about the potential users of my app, I wonder if you have any empirical evidence. My guess is that 90% of them seeing a FUSE path think it is a bug an loose trust in the app. But that's only a guess, based on daily non-representative experience.

Anyways, thanks a lot for your time, I learned a ton the last 2 days! I will continue the submission once the portal exposes the xattr on most user's machines.

bbhtt: pretty much most apps show the fuse mounted path, anyways you should limit to xdg-pictures and/or other xdg-* directories instead of full home. it's an image processing tool that should be a fair middleground. if you are concerned don't show the fuse mounted path front and center to the user.

PR #6713, opened 2025-07-11
Add com.infinipaint.infinipaint

ErrorAtLine0:

- [X] Please describe the application briefly. InfiniPaint is a vector graphics based infinite canvas program that allows for infinite zoom and online collaboration. It's good for large scale sketches, note taking, or use as a white board app.

https://github.com/user-attachments/assets/33982c26-e2c6-4def-8c52-ea23b1b7157f


com.erroratline0.infinipaint.metainfo.xml
```diff
@@ -0,0 +1,89 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2025 Yousef Khadadeh -->
+<component type="desktop-application">
+    <id>com.erroratline0.infinipaint</id>
```
hfiguiere: so this should be `com.infinipaint.infinipaint`

com.erroratline0.infinipaint.metainfo.xml
hfiguiere, line 1: this file should be (submitted and managed) upstream.

com.erroratline0.infinipaint.yml
```diff
@@ -0,0 +1,105 @@
...
+  - /Skia
+  - /include
+  - /lib/pkgconfig
+  - /lib/cmake
+  - /lib/cmake
+  - /lib/libupb.a
+  - /bin/proto*
+  - /lib/*.a
+        
+finish-args:
```
hfiguiere: `finish-args` should be above modules

com.erroratline0.infinipaint.yml
```diff
@@ -0,0 +1,105 @@
...
+        url: https://github.com/ErrorAtLine0/infinipaint.git
+        tag: v0.0.3
+  - name: infinipaint-flatpak-overwrite
+    buildsystem: simple
+    build-commands:
+      - mv com.erroratline0.infinipaint.metainfo.xml ${FLATPAK_DEST}/share/metainfo
+    sources:
+      - type: file
+        path: com.erroratline0.infinipaint.metainfo.xml
+cleanup:
```
hfiguiere: `cleanup` should be above modules

com.erroratline0.infinipaint.yml
```diff
@@ -0,0 +1,105 @@
...
+        - -DSKIA_UNICODE_ICU_LIB=/app/Skia/out/Static/libskunicode_core.a
+        - -DSKIA_UNICODE_CORE_LIB=/app/Skia/out/Static/libskunicode_icu.a
+        - -DSKIA_PARAGRAPH_LIB=/app/Skia/out/Static/libskparagraph.a
+        - -DCMAKE_BUILD_TYPE=Release
+        - -DTBB_NOT_ONETBB=ON
+    sources:
+      - type: git
+        url: https://github.com/ErrorAtLine0/infinipaint.git
+        tag: v0.0.3
+  - name: infinipaint-flatpak-overwrite
```
hfiguiere: you don't need a separate module for this

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+   "only-arches": ["x86_64"]
```
hfiguiere, line 2: I can't express my disappointment here.

com.erroratline0.infinipaint.yml
```diff
@@ -0,0 +1,105 @@
...
+      - mv src /app/Skia
+      - mv modules /app/Skia
+    cleanup:
+      - '*'
+    sources:
+      - type: archive
+        url: https://github.com/google/skia/archive/aefbd9403c1b3032ad4cd0281ef312ed262c7125.zip
+        sha256: e10c67a96828b9f36364e2459cdee3e3b381a3875ff78db2b30e7b6fe69509bb
+      - type: archive
+        url: https://infinipaint.com/downloads/skiabuilds/skia-aefbd940-linux-x86_84-flatpak-24.08.7z
```
hfiguiere: why isn't this built from source?

ErrorAtLine0: I build Skia from source in a flatpak manifest that I use to create single file bundles, but I can't do this here, because flathub doesn't let me connect to the network during build time. Skia's build system pulls its own build tools and libraries through a custom script, so I'm not sure if it's possible to build from source without an internet connection. Of course, I could dissect their build scripts so that I can pull everything they want through the manifest, but the library has a ton of dependencies (some of which are binaries, since as I said, it pulls its own build tools), so it's not that straight forward.
I looked at a [flatpak manifest written for Aseprite](https://github.com/Ketok4321/aseprite-flatpak/blob/main/aseprite.yaml), and in this manifest they used a prebuilt library. If there's an example of a flatpak manifest that builds skia from source, I'd like to see it for reference.
These prebuilt libraries are also why I locked this app to build only for x86_64 for now.

com.infinipaint.infinipaint.yml
```diff
@@ -0,0 +1,97 @@
...
+  - --socket=x11
+  - --device=all
+  - --share=network
+  - --filesystem=xdg-desktop
+  - --filesystem=xdg-documents
+  - --filesystem=xdg-download
+  - --filesystem=xdg-music
+  - --filesystem=xdg-pictures
+  - --filesystem=xdg-public-share
+  - --filesystem=xdg-videos
```
bbhtt: music, public-share and videos seems unnecessary for a graphics app.

com.infinipaint.infinipaint.yml
```diff
@@ -0,0 +1,94 @@
+id: com.infinipaint.infinipaint
+runtime: org.freedesktop.Platform
+runtime-version: '24.08'
+sdk: org.freedesktop.Sdk
+command: infinipaint
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --device=all
+  - --share=network
```
bbhtt, line 10: what are these two for?

ErrorAtLine0: --device=all is for OpenGL, and wacom tablet. Would the equivalent for this be --device=dri, --device=input, --device=usb? I could separate it into those three if that's good enough.

--share=network is because this app allows for online collaboration

bbhtt: no it's ok

bbhtt:
bot, build

PR #6749, opened 2025-07-26
Add com.fdossena.flogo

adolfintel:
  Create and run programs using flow charts. (Flogo is a teaching tool that I use with my students, it's a flow chart creator that also lets you run your programs.)
https://github.com/user-attachments/assets/c7426920-1ca3-42f6-9ce4-0467c94b6a64

**Important note:** I'm not sure about the sandbox permissions that I've used here. The application has to be able to read and save files, and registers the `.flogo` file extension (binary file with its own mime type). `home` would be enough, but I need my students to be able to read and save programs to flash drives, `host` seems to be the only option, unless there's special permissions for removable drives.


adolfintel:
bot, build

adolfintel:
If someone could shed some light on the host permission, it would be great 😅
I don't see many other options

com.fdossena.flogo.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Name=Flogo
+Comment=Create and run programs using flow charts
+Exec=/app/bin/run.sh %U
```
hfiguiere: this is never a full path.

com.fdossena.flogo.metainfo.xml
hfiguiere, line 1: this file should be (submitted and managed) upstream

com.fdossena.flogo.desktop
hfiguiere, line 1: this file should be (submitted and managed) upstream

com.fdossena.flogo.mime.xml
hfiguiere, line 1: this file should be (submitted and managed) upstream

flathub.json
hfiguiere, line 1: delete this file.

icon.png
hfiguiere, line 1: this file should be (submitted and managed) upstream

icon_file.png
hfiguiere, line 1: this file should be (submitted and managed) upstream

com.fdossena.flogo.yml
```diff
@@ -0,0 +1,52 @@
...
+        path: com.fdossena.flogo.metainfo.xml
+      - type: file
+        path: com.fdossena.flogo.mime.xml
+      - type: file
+        path: icon.png
+      - type: file
+        path: icon_file.png
+    build-commands:
+      - unzip flogo-*.zip -d /app/flogo
+      - install run.sh /app/bin/
```
hfiguiere:
```suggestion
      - install -Dm755 run.sh -t /app/bin/
```

com.fdossena.flogo.yml
```diff
@@ -0,0 +1,52 @@
...
+  - --device=dri
+  - --share=ipc
+  - --socket=x11
+  - --socket=pulseaudio
+  - --filesystem=host
+modules:
+  - name: flogo
+    buildsystem: simple
+    sources:
+      - type: file
```
hfiguiere: these should be "type: archive"

adolfintel:
@hfiguiere I did most of the changes that you suggested except one: it looks like I can't replace file with archive for the 2 zip files because it flattens the directory structure and makes a total mess of the application files.

I also have another question if you don't mind; you say that the metadata and desktop files should be upstream, so I moved them to my server, no problem, but I see virtually all other repos here having those files. Why?

com.fdossena.flogo.yml
hfiguiere: still todo

com.fdossena.flogo.yml
```diff
@@ -0,0 +1,57 @@
...
+      - unzip flogo-*.zip -d /app/flogo
+      - install -Dm755 run.sh -t /app/bin/
+      - install -Dm644 icon.png /app/share/icons/hicolor/256x256/apps/com.fdossena.flogo.png
+      - install -Dm644 icon_file.png /app/share/icons/hicolor/256x256/apps/mimetypes/com.fdossena.flogo-file.png
+      - install -Dm644 com.fdossena.flogo.desktop -t /app/share/applications/
+      - install -Dm644 com.fdossena.flogo.metainfo.xml -t /app/share/metainfo/
+      - install -Dm644 com.fdossena.flogo.mime.xml -t /app/share/mime/packages/
+      - patch-desktop-filename "${FLATPAK_DEST}"/flogo/resources/app.asar
+    post-install:
+      - update-mime-database /app/share/mime
```
hfiguiere:
```suggestion
```

this is not necessary at all.

com.fdossena.flogo.yml
adolfintel: As I said in my comment, it looks like I can't because it messes up the application files by flattening the directory structure. If there's some config to avoid it, I'd be happy to do it.

com.fdossena.flogo.yml
```diff
@@ -0,0 +1,57 @@
...
+        sha256: 1110b3eb8d90bb046ef771198484d47c178ecac58306cbb355c52b4e8460de77
+      - type: file
+        only-arches:
+          - aarch64
+        url: https://downloads.fdossena.com/Projects/Flogo/flatpak/flogo-1.1.0-aarch64.zip
+        sha256: 1b51f4feca95961a56b28626a76026f0bb6f30d88d72968d74f051ff331d08af
+      - type: file
+        path: run.sh
+      - type: file
+        url: https://downloads.fdossena.com/Projects/Flogo/flatpak/com.fdossena.flogo.desktop
```
hfiguiere: this and below: the URL isn't versioned. This mean it would break the build randomly when updating.

Why aren't they in the tarball instead?

com.fdossena.flogo.yml
hfiguiere: `strip-components` is well documented

But maybe the tarball should be fixed too to follow conventions.

com.fdossena.flogo.yml
adolfintel: To be honest I just copied what the FreeTube yml was doing, it's my first time 😅
I'll add them to the zip files

hfiguiere:
> I also have another question if you don't mind; you say that the metadata and desktop files should be upstream, so I moved them to my server, no problem, but I see virtually all other repos here having those files. Why?

Two wrongs don't make a right. It is clearly written in the requirement documentation that they should be managed / handled upstream. Since you are upstream there no "upstream doesn't want", "upstream refuse submission", "it's proprietary", "they don't care about free desktop", etc.

com.fdossena.flogo.yml
hfiguiere: vibe coding isn't good.

hfiguiere:
also https://docs.flathub.org/docs/for-app-authors/requirements#building-from-source

com.fdossena.flogo.yml
adolfintel: There's no need to be rude. I didn't vibe code anything, I'm just trying to figure out how flatpak works. I never used it before and it's quite complex.

com.fdossena.flogo.yml
adolfintel: Fixed

com.fdossena.flogo.yml
adolfintel: Removed

adolfintel:
Better now?

adolfintel:
> also https://docs.flathub.org/docs/for-app-authors/requirements#building-from-source

Correct me if I'm wrong, but electron apps aren't normally built from source here.

I saw the tool called `flatpak-yarn-generator` in your repos and it yells at me because electron zips don't have hashes, but the other dependencies are fine.

To get it to work I would basically have to remove electron from my package.json and have flatpak download the appropriate one, and I'd have to update it regularly or risk building against an old version. That seems insane and no other app here is doing that.

Is there something I'm missing?

bbhtt:
> I saw the tool called flatpak-yarn-generator 

Use Flatpak Node generator.

> Correct me if I'm wrong, but electron apps aren't normally built from source here.

You are wrong.

com.fdossena.flogo.yml
```diff
@@ -0,0 +1,44 @@
+app-id: com.fdossena.flogo
+runtime: org.freedesktop.Platform
+runtime-version: '24.08'
+branch: stable
```
bbhtt:
```suggestion
```

adolfintel:
@bbhtt I managed to make it build from source. Can you take a look at the updated files?

bbhtt:
bot, build

com.fdossena.flogo.yml
```diff
@@ -0,0 +1,45 @@
...
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node24
+command: run.sh
+separate-locales: false
+finish-args:
+  - --device=dri
+  - --share=ipc
+  - --socket=x11
+  - --socket=pulseaudio
+  - --filesystem=host
```
bbhtt: > I'm not sure about the sandbox permissions that I've used here. The application has to be able to read and save files

Does it not use filechoosers for these?

adolfintel: Yes, but file associations don't seem to work without this, at least not outside of folders that I explicitly allow.

I don't like it but I'm not sure how else it could be handled.

bbhtt: What do you mean by file associations don't work? What does that have to with this?

adolfintel: If I remove `--filesystem=host`, when I double click a file in my file manager, the application receives the path of the file, but it then fails to read it because it can't access that path

bbhtt: What does the desktop file of the app look like? That's why Flatpak adds `--file-forwarding`. That is not supposed to need extra permissions.

adolfintel: This is the current desktop file.
```
[Desktop Entry]
Name=Flogo
Comment=Create and run programs using flow charts
Exec=run.sh %U
Terminal=false
Type=Application
Icon=com.fdossena.flogo
MimeType=application/x-flogo
Categories=Education
StartupWMClass=Flogo
```

adolfintel: Hold on, the one it installed in the system does have the --file-forwarding option, it runs this command
`/usr/bin/flatpak run --branch=master --arch=x86_64 --command=run.sh --file-forwarding com.fdossena.flogo @@u %U @@`

bbhtt: sounds like a bug... maybe something is wrong with your portals or environment.

Can you remove it temporarily so that i can test a build?

adolfintel: Done. I can also add some debugging prints in the app if you want

bbhtt:
bot, build

PR #6762, opened 2025-07-28
Add io.github.heathcliff26.turbo-clicker

heathcliff26:
Add my new app Turbo Clicker to flathub.
It is currently at the first working iteration.
Further features follow when i have time.

Turbo Clicker is a GUI based auto-clicker for Linux.

https://github.com/user-attachments/assets/74e4974a-e766-4188-84d7-ca1a530764fc

### DIFF
--- a/io.github.heathcliff26.turbo-clicker.yaml
+++ b/io.github.heathcliff26.turbo-clicker.yaml
@@ -1,0 +1,38 @@
+app-id: io.github.heathcliff26.turbo-clicker
+runtime: org.freedesktop.Platform
+runtime-version: '24.08'
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+command: turbo-clicker
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc
+  - --device=all
+modules:
+  - name: turbo-clicker
+    buildsystem: simple
+    build-options:
+      append-path: /usr/lib/sdk/rust-stable/bin
+      env:
+        CARGO_HOME: /run/build/turbo-clicker/cargo
+        CARGO_NET_OFFLINE: 'true'
+    sources:
+      - type: git
+        url: https://github.com/heathcliff26/turbo-clicker
+        tag: v0.0.1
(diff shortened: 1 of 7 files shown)


io.github.heathcliff26.turbo-clicker.yaml
```diff
@@ -0,0 +1,38 @@
...
+runtime-version: '24.08'
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+command: turbo-clicker
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc
+  - --device=all
```
bbhtt, line 12: why does it need device=all?

bbhtt:
https://github.com/user-attachments/assets/1f34a08b-8cc5-4db4-ba1c-0a5c3d32d413

io.github.heathcliff26.turbo-clicker.yaml
heathcliff26: Because it uses `/dev/uinput` to create a virtual mouse.

heathcliff26:
Thank you for your review @bbhtt .
I will take a look at the error today after work.

heathcliff26:
Okay, it seems the problem here is that `/dev/uinput` has permission `0600`.
This can be temporarily solved by running:
```
sudo setfacl -m u:$(whoami):rw /dev/uinput
```
or permanently by creating a udev rule.

Funnily enough the setfacl is done on my system by something, i just don't know what.

@bbhtt Do you now of a way on changing/masking the file permission for the device for flatpak or run it as root? Because i can't find one and the app does not really make sense as a flatpak otherwise.

hfiguiere:
> Do you now of a way on changing/masking the file permission for the device for flatpak or run it as root? Because i can't find one and the app does not really make sense as a flatpak otherwise.

no this is not possible.

heathcliff26:
> > Do you now of a way on changing/masking the file permission for the device for flatpak or run it as root? Because i can't find one and the app does not really make sense as a flatpak otherwise.
> 
> no this is not possible.

Okay, in that case i'm closing this, because i don't think it makes sense as a flatpak if you need to manually make changes on the system. I will try packaging as an rpm instead.

Thank you two for your time.

PR #6771, opened 2025-08-01
Add org.kde.komodo

Akselmo:

- [x] Please describe the application briefly. KomoDo is a To-Do application that utilizes todo.txt syntax for it's To-Do items.
[Screencast_20250802_001557.webm](https://github.com/user-attachments/assets/877268de-138f-4665-9649-6e56a57c336c)


Akselmo:
Hi, never published to flathub before :eyes: 

Just wanted to mention that this first release does not have tarball yet like most KDE apps do, since I wanted to get this out there quickly. 

I am giving translators some time and then make a tarball, then I will likely modify the JSON file to use the tarball instead for future releases.

The app is definitely usable and ready for wider use though. It's just that I'm releasing it here before in the KDE tarballs, and I don't know if that's bad.

hfiguiere:
cc @flathub/kde

org.kde.komodo.json
```diff
@@ -0,0 +1,63 @@
...
+  "cleanup": [
+    "*.a",
+    "*.la",
+    "/doc",
+    "/docs",
+    "/include",
+    "/lib/cmake",
+    "/lib/pkgconfig",
+    "/lib/x86_64-linux-gnu/cmake",
+    "/lib/x86_64-linux-gnu/pkgconfig",
```
hfiguiere: do these two really exist? How about aarch64?

Akselmo: pinging @Justinzobel since he made the original manifest :) 

(i have no idea what im doing)

Justinzobel: I used a generic list of things we don't need, I realised after adding this one we should make it arch generic.

Justinzobel:
```suggestion
    "/lib/*/cmake",
    "/lib/*/pkgconfig",
```

org.kde.komodo.json
```diff
@@ -0,0 +1,63 @@
...
+      ]
+    },
+    {
+      "name": "komodo",
+      "buildsystem": "cmake-ninja",
+      "builddir": true,
+      "sources": [
+        {
+          "type": "git",
+          "url": "https://invent.kde.org/utilities/komodo.git",
```
tsdgeos: We do not build from git.

Please do a proper release with a tarball and build from there.

Akselmo: I'm now using archive

Akselmo: I'm still missing `x-checker-data` but I can't find anything about that anywhere.

Akselmo:
Now when I try to build the app on my system, i keep getting this:

```
> flatpak run --command=flathub-build org.flatpak.Builder --install org.kde.komodo.json

Running: flatpak-builder --verbose --force-clean --sandbox --keep-build-dirs --override-source-date-epoch 1321009871 --user --install-deps-from=flathub --ccache --mirror-screenshots-url=https://dl.flathub.org/media --repo=repo builddir --install org.kde.komodo.json
FB: Running: git config --get user.email
FB: Running: git config --get user.name
Can't initialize ccache use: Read-only file system
```

Akselmo:
bot, build

org.kde.komodo.json
tsdgeos: Please use 379835 as project id 

https://release-monitoring.org/project/379835/

Akselmo: Added!

tsdgeos:
Looks reasonable to me

bbhtt:
Looks like the "create todo list" allows saving without the `.txt` extension, but then "open todo list" cannot find the file as the filepicker is filtering on that one.


https://github.com/user-attachments/assets/74c597ca-eeeb-45ed-8ddb-41c088104b15



looks like it is also leaving behind a temporary file, but probably not a bug in the app.

org.kde.komodo.json
```diff
@@ -0,0 +1,72 @@
...
+    {
+      "name": "kirigamiaddons",
+      "config-opts": [
+        "-DBUILD_TESTING=OFF"
+      ],
+      "buildsystem": "cmake-ninja",
+      "sources": [
+        {
+          "type": "archive",
+          "url": "https://download.kde.org/stable/kirigami-addons/kirigami-addons-1.9.0.tar.xz",
```
Justinzobel:
```suggestion
          "url": "https://cdn.download.kde.org/stable/kirigami-addons/kirigami-addons-1.9.0.tar.xz",
```

Justinzobel (changes requested):

Akselmo:
> Looks like the "create todo list" allows saving without the `.txt` extension, but then "open todo list" cannot find the file as the filepicker is filtering on that one.
> 
> 
> https://github.com/user-attachments/assets/74c597ca-eeeb-45ed-8ddb-41c088104b15
> 
> 
> 
> looks like it is also leaving behind a temporary file, but probably not a bug in the app.

Huh, good catch, will have to fix that. Is this something that requires fixing before getting into flathub?

bbhtt:
preferably yes.

tsdgeos:
> preferably yes.

That's the weirdest take ever. Are we not accepting apps with bugs now? Will have to clear flathub completely of apps :D

org.kde.komodo.json
tsdgeos: Are you sure about this? We use download.kde.org links everywhere

Akselmo: Personally I'm fine with either

bbhtt:
That's not a weird take at all.

org.kde.komodo.json
Justinzobel: > Are you sure about this? We use download.kde.org links everywhere

```
❯ ds cdn.download.kde.org
1009236516.rsc.cdn77.org.
143.244.62.6
143.244.62.9

~ 
❯ ds download.kde.org
tinami.kde.org.
49.12.122.15
```

CDN goes to CDN77, without it goes to KDE infra.

tsdgeos:
> That's not a weird take at all.

Should we remove firefox from flathub because it has bugs?

bbhtt:
Where exactly did I say apps aren't allowed to have bugs? Please do not put words in my mouth.

If I find an obvious bug within 2 minutes of casual usage, and accept the app in that state it reflects badly on the app, the store and the review process.

If you don't have anything useful to add please don't interfere with the review process.

And this isn't a new thing at all it has been done on all submissions that needed it and on several kde app submissions before.

tsdgeos:
> Where exactly did I say apps aren't allowed to have bugs? Please do not put words in my mouth.

I said "That's the weirdest take ever. Are we not accepting apps with bugs now? Will have to clear flathub completely of apps :D"

and you said "That's not a weird take at all."

Please explain how reading those two sentences together does not mean you agree that we are not accepting apps with bugs and that we have to remove apps from flathub that have bugs.

bbhtt:
I have no idea what you want or why we are having this conversation but the answer to that is "yes I won't accept apps with obvious bugs that show up during review (until they are fixed)"

Justinzobel:
Can we please keep this on the topic of the app submission? If you'd like to discuss app submission process or guidelines, please create a separate issue.

travier:
I agree with @bbhtt, if we could fix this papercut before we publish the app that would be good so we don't immediately get users stumble upon it from the start. Thanks @Akselmo for working on the publishing here.

Akselmo:
> I agree with @bbhtt, if we could fix this papercut before we publish the app that would be good so we don't immediately get users stumble upon it from the start. Thanks @Akselmo for working on the publishing here.

I'll make a patch and add it here when I can, so I don't need to spin a new release.

Akselmo:
I added a patch, it was missing the default suffix from the save dialog. Now it should append ".txt" to the files per default and thus fix the issue.

One has to select "all files" if the suffix is missing of course.

Akselmo:
bot, build

bbhtt:
Something else broke, now it is only creating the xdp-*** temporary files


https://github.com/user-attachments/assets/6ddd683b-aaea-4d8d-a5b0-cadcb3f0ae0c

Akselmo:
> Something else broke, now it is only creating the xdp-*** temporary files
> Screencast.mp4

Uh. Wtf.

When I do the same, it shows up in Documents folder as expected. Am I missing some permission I may have set to myself globally?

Though when I tried to use flatpak settings to remove permissions, it still worked in Documents folder.

travier:
Testing that, I get both the `test.txt` file and a `.xdp-qt_temp.Gybtfa-fLkNvq` one.

Akselmo:
The temp file is expected: i added a feature that when saving the file, it creates a hidden copy of the file before saving, in case something goes wrong. It should always save to same temp file though, and it should be named after the saved file (.todo.txt)

I want to eventually add git support to this but for now that works as a "backup"

bbhtt:
It's working on KDE, not working on GNOME, could be a bug in the GNOME portal or the Qt/KDE component responsible for providing the API to the app.

Akselmo:
> It's working on KDE, not working on GNOME, could be a bug in the GNOME portal or the Qt/KDE component responsible for providing the API to the app.

Ah ok. I can remove the *.txt filters as a workaround for now, if that helps.

bbhtt:
I was looking to test another app on GNOME that uses the QtQuick Filedialog to see the status and seems Neochat is using it? https://github.com/search?q=repo%3AKDE%2Fneochat%20%22Dialogs.FileDialog%22&type=code

If so then the filedialog used by neochat can save it correctly on gnome for me. I clicked an image on the chat timeline > save image > chose a location outside `xdg-downloads` so that it goes through the portal.

bbhtt:
> Ah ok. I can remove the *.txt filters as a workaround for now, if that helps.

Does that break the suffix on KDE? If not, then that might be better, setting the defaultsuffix seems to cause more breakage on gnome at least.

bbhtt:
Looks like with defaultsuffix set, on gnome the file is correctly saved only if i type the extension as well


https://github.com/user-attachments/assets/9acdf044-6fa5-4b87-b12e-58280c48d172

bbhtt:
i guess an easy workaround instead of setting defaultsuffix might be to just prefill it with a default name + extension like `todo.txt`.

Akselmo:
I added a patch for removing the filetype filters, so at least things will show up no matter what they're named.

bot, build

bbhtt:
I guess that works too, seems ok to me.

PR #6818, opened 2025-08-13
Add club.haroohie.SerialLoops

jonko0493:

- [X] Please describe the application briefly. Serial Loops is an editor for the 2009 Nintendo DS visual novel game _Suzumiya Haruhi no Chokuretsu_ (_The Series of Haruhi Suzumiya_). It currently allows massive modification of the game (and will eventually allow editing nearly every game system) with the goal of enabling fans of the Haruhi series to create their own fangames as ROM hacks.
[Screencast_20250813_213341.webm](https://github.com/user-attachments/assets/27c4a010-19c7-4e18-97c4-2e6a92605428)

@WiIIiam278


club.haroohie.SerialLoops.yaml
```diff
@@ -0,0 +1,89 @@
...
+    buildsystem: simple
+    sources:
+      - type: git
+        url: https://github.com/ninja-build/ninja.git
+        tag: v1.13.1
+    build-commands:
+      - python3 ./configure.py --bootstrap
+      - mv ninja ${FLATPAK_DEST}/bin/ninja
+
+  - name: SDL
```
hfiguiere: 2.23.4 is in the runtime. isn't that enough?

club.haroohie.SerialLoops.yaml
```diff
@@ -0,0 +1,89 @@
...
+    buildsystem: simple
+    sources:
+      - type: git
+        url: https://github.com/libsdl-org/SDL.git
+        tag: release-2.32.8
+    build-commands:
+      - mkdir build
+      - cmake -S . -B build -DCMAKE_BUILD_TYPE=Release
```
hfiguiere: this should be `buildsystem: cmake-ninja` in any case.

club.haroohie.SerialLoops.yaml
```diff
@@ -0,0 +1,89 @@
...
+  prepend-pkg-config-path: "/usr/lib/sdk/dotnet8/lib/pkgconfig"
+
+command: SerialLoops
+
+finish-args:
+  - --device=dri
+  - --socket=x11
+  - --share=ipc
+  - --socket=pulseaudio
+  - --filesystem=~/SerialLoops:create
```
hfiguiere: this should be `--persist` instead.

club.haroohie.SerialLoops.yaml
jonko0493: Yep, that should be sufficient. Only reason we weren't already using it was because I wasn't aware it was available! Will switch now.

club.haroohie.SerialLoops.yaml
jonko0493: Ah, yes, you're correct. Thank you for the catch! Fixing now.

jonko0493: @hfiguiere from experimentation, it seems as though using `--persist`  creates a new directory within the flatpak filesystem rather than modifying the home directory. Is this the expected behavior or am I holding it wrong? If it is expected, I think `--filesystem` would be preferred in our case -- we want users upgrading from previous versions (or swapping between the AppImage or other distribution and the Flatpak for e.g. preview builds) to have access to their same projects. Is that reasonable?

club.haroohie.SerialLoops.yaml
jonko0493: The library we're using for SDL2 doesn't find it in the current location, so I've simply copied/symlinked the libs into the appdir. Hopefully this is acceptable.

club.haroohie.SerialLoops.yaml
```diff
@@ -0,0 +1,83 @@
...
+      - type: git
+        url: https://github.com/haroohie-club/SerialLoops.git
+        tag: 0.3.3
+      - ./nuget-sources.json
+      - type: file
+        path: club.haroohie.SerialLoops.metainfo.xml
+      - type: file
+        path: club.haroohie.SerialLoops.desktop
+      - type: file
+        path: AppIcon.svg
```
bbhtt: these should be coming from the upstream repo not here.

club.haroohie.SerialLoops.yaml
```diff
@@ -0,0 +1,83 @@
...
+        tag: v1.13.1
+    build-commands:
+      - python3 ./configure.py --bootstrap
+      - mv ninja ${FLATPAK_DEST}/bin/ninja
+
+    # Our SDL library expects libSDL to be in a specific spot, so we copy it here
+  - name: SDL
+    buildsystem: simple
+    build-commands:
+      - cp /usr/lib/x86_64-linux-gnu/libSDL* ${FLATPAK_DEST}/lib/
```
bbhtt: this won't work on aarch64

club.haroohie.SerialLoops.metainfo.xml
```diff
@@ -0,0 +1,83 @@
...
+    <content_attribute id="language-humor">none</content_attribute>
+    <content_attribute id="language-discrimination">none</content_attribute>
+    <content_attribute id="social-chat">none</content_attribute>
+    <content_attribute id="social-info">none</content_attribute>
+    <content_attribute id="social-audio">none</content_attribute>
+    <content_attribute id="social-location">none</content_attribute>
+    <content_attribute id="social-contacts">none</content_attribute>
+    <content_attribute id="money-purchasing">none</content_attribute>
+    <content_attribute id="money-gambling">none</content_attribute>
+  </content_rating>
```
bbhtt:
```suggestion
  <content_rating type="oars-1.1" />
```

club.haroohie.SerialLoops.metainfo.xml
```diff
@@ -0,0 +1,83 @@
...
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>GPL-3.0-or-later</project_license>
+  <name>Serial Loops</name>
+  <launchable type="desktop-id">club.haroohie.SerialLoops.desktop</launchable>
+  <summary>Editor for Suzumiya Haruhi no Chokuretsu</summary>
+  <description>
+    <p>
+      Serial Loops is a full-fledged editor for the Nintendo DS game Suzumiya Haruhi no Chokuretsu (The Series of Haruhi Suzumiya).
+      In its current state, it allows you to edit nearly everything about the game, and more is planned for the future. Please check
+      <a href="https://haroohie.club/chokuretsu/serial-loops/docs">our documentation</a> for more information on how to use it!
```
bbhtt: this isn't html.

club.haroohie.SerialLoops.yaml
```diff
@@ -0,0 +1,83 @@
...
+  - name: llvm
+    buildsystem: simple
+    build-commands:
+    - mkdir -p ${FLATPAK_DEST}/lib/llvm20/bin/
+    - mkdir -p ${FLATPAK_DEST}/lib/llvm20/lib/
+    - cp /usr/lib/sdk/llvm20/bin/clang ${FLATPAK_DEST}/lib/llvm20/bin
+    - cp /usr/lib/sdk/llvm20/bin/ld.lld ${FLATPAK_DEST}/lib/llvm20/bin
+    - cp /usr/lib/sdk/llvm20/bin/llvm-objcopy ${FLATPAK_DEST}/lib/llvm20/bin
+    - cp /usr/lib/sdk/llvm20/bin/llvm-objdump ${FLATPAK_DEST}/lib/llvm20/bin
+    - cp -r /usr/lib/sdk/llvm20/lib/* ${FLATPAK_DEST}/lib/llvm20/lib/
```
bbhtt:
```suggestion
    - cp -ra /usr/lib/sdk/llvm20/lib/* ${FLATPAK_DEST}/lib/llvm20/lib/
```

club.haroohie.SerialLoops.yaml
bbhtt:
```suggestion
      - cp -a /usr/lib/x86_64-linux-gnu/libSDL* ${FLATPAK_DEST}/lib/
```

club.haroohie.SerialLoops.yaml
hfiguiere: Then it should be fixed. `pkg-config` should find sdl2, and if they don't use that it's a bug.

club.haroohie.SerialLoops.yaml
hfiguiere: there is no reason to do it anyway. It's a broken build system at the core.

club.haroohie.SerialLoops.yaml
jonko0493: Okay, I'll fork it and publish my own version.

club.haroohie.SerialLoops.yaml
jonko0493: When you say these, are you just referring to the AppIcon?  Or also the desktop/nuget-sources? Most people seem to have their metainfo.xml checked in, but happy to move it if needed.

bbhtt: The comment is on line 71 to 76 it is very clear.

bbhtt: Even clearer if you read the requirements doc linked in PR.

club.haroohie.SerialLoops.yaml
hfiguiere: no. migratio isn't meant to be. It's sandboxed and should stay taht way. Persist is the way. People can still move stuff over.

club.haroohie.SerialLoops.yaml
hfiguiere: no you should NOT fork. Patch properly and submit that patch (and add it here in the mean time). It's a bug if it can't find sdl2.

jonko0493: Okay, I have looked into this more. The library does not contain a bug -- it is simply using the default .NET native library loader. See [here](https://learn.microsoft.com/en-us/dotnet/core/dependency-loading/default-probing) and [here](https://learn.microsoft.com/en-us/dotnet/standard/native-interop/native-library-loading) for more information, but suffice it to say that it is looking for `libSDL2.so` which does not appear to be present in the final flatpak sandbox, despite being present in `org.kde.Sdk`. Specifically, these two files are present in the x86_64 image:
```
/usr/lib/x86_64-linux-gnu/libSDL2-2.0.so.0
/usr/lib/x86_64-linux-gnu/libSDL2-2.0.so.0.3200.4
```
But `/usr/lib/x86_64-linux-gnu/libSDL2.so` is not present, so as a result, it's not resolving the library.

jonko0493: I wrote a custom DLL importer that replaces the search for `libSDL` with a specific search for `libSDL2-2.0.so.0` and the application now functions as intended. To me this seems like a bug in the app builder, but if this is intended behavior I can commit with the custom DLL importer.

jonko0493: @hfiguiere Do you have any thoughts on my findings here? If the lack of the `libSDL2.so` symlink is not an issue, then I will just commit the DLL importer and I will have addressed all feedback on this PR. Would like to avoid leaving this dangling.

hfiguiere: if it's looking for a `.so` then it's doing it wrong. Whoever implemented it doesn't understand .so versions. They are versioned and it should be looking for the versioned file.

I most Linux distro the `.so` is only part of the development pacakges. They are symlinks to the actually version you'll link against and the linker understand that.

jonko0493:
bot, build

jonko0493:
bot, build

club.haroohie.SerialLoops.yaml
jonko0493: Okay. I have put in a fix on the application side.

jonko0493:
@bbhtt @hfiguiere Hi folks -- build is passing and I've resolved all the feedback so far. Please let me know if there's anything else I should fix. Thanks!

PR #6822, opened 2025-08-15
Add io.github.mohfy.phoneme

mohfy:
I changed the App ID so I had to resubmit the app :)

[prev pr for more info](https://github.com/flathub/flathub/pull/6769)


mohfy:
bot, build

mohfy:
bot, build

bbhtt:
A change in the last component doesn't need a rename

icymi https://docs.flathub.org/docs/for-app-authors/requirements#renaming-flatpak-id

> A rename that does not change the domain portion of the app ID will not be accepted.

mohfy:
Then how else could I change the name of the app?

hfiguiere:
By changing the name but not the id.

PR #6834, opened 2025-08-17
Add net.ventoy.VentoyGUI

LorenzoIanotto:

- [x] Please describe the application briefly. < Ventoy is an open source tool to create bootable USB drive for ISO/WIM/IMG/VHD(x)/EFI files. With ventoy, you don't need to format the disk over and over, you just need to copy the ISO/WIM/IMG/VHD(x)/EFI files to the USB drive and boot them directly. >
- [x] Please attach a video showcasing the application on Linux using the Flatpak. < [https://youtu.be/Imm0gQanaNk](https://youtu.be/Imm0gQanaNk) >
- [x] contacted upstream developers about this submission. **Link:https://github.com/ventoy/Ventoy/pull/3319**

### DIFF
--- a/net.ventoy.VentoyGUI.yml
+++ b/net.ventoy.VentoyGUI.yml
@@ -92,7 +92,7 @@ modules:
             sources:
               - type: git
                 url: https://github.com/JingchengLi/libaio.git
-                commit: 1b18bfa
+                commit: 1b18bfafc6a2f7b9fa2c6be77a95afed8b7be448
             build-commands:
               - make prefix=/app install
           - name: lvm2
@@ -179,7 +179,7 @@ modules:
     sources:
       - type: git
         url: https://github.com/LorenzoIanotto/VentoyUDisks.git
-        commit: 17aa00e
+        commit: 17aa00e954db579eea26c19ace58631a9af8f391
       - type: script
         dest-filename: ventoy-gtk3.sh
         commands:
--- a/net.ventoy.VentoyGUI.yml
+++ b/net.ventoy.VentoyGUI.yml
@@ -1,0 +1,206 @@
+id: net.ventoy.VentoyGUI
+runtime: org.gnome.Platform
+runtime-version: '48'
+sdk: org.gnome.Sdk
+command: ventoy-gtk3.sh
+modules:
+  - name: udisks2
+    sources:
+      - type: git
+        url: https://github.com/storaged-project/udisks.git
+        tag: udisks-2.10.1
+    config-opts:
+      - "--enable-available-modules=no"
+      - "--enable-bcache=no"
+      - "--enable-btrfs=no"
+      - "--enable-introspection=no"
+      - "--enable-lvm2=no"
+      - "--enable-modules=no"
+      - "--enable-vdo=no"
+      - "--enable-zram=no"
+      - "--disable-smart"
+      - "--with-systemdsystemunitdir=no"
+      - "--with-tmpfilesdir=no"
+    modules:
+      - name: polkit
+        buildsystem: meson
+        sources:
+          - type: git
+            url: https://github.com/polkit-org/polkit.git
+            tag: "126"
+        config-opts:
+          - -Dlibs-only=true
+          - -Dintrospection=false
+        modules:
+          - name: pam
+            sources:
+              - type: archive
+                url: https://github.com/linux-pam/linux-pam/releases/download/v1.4.0/Linux-PAM-1.4.0.tar.xz
+                sha256: cd6d928c51e64139be3bdb38692c68183a509b83d4f2c221024ccd4bcddfd034
+            config-opts:
+              - "--includedir=/app/include/security"
+              - "--disable-doc"
+      - name: libnvme
+        buildsystem: meson
+        sources:
+          - type: git
+            url: https://github.com/linux-nvme/libnvme.git
+            tag: v1.15
+      - name: libblockdev
+        sources:
+          - type: git
+            url: https://github.com/storaged-project/libblockdev.git
+            tag: 3.3.1
+        config-opts:
+          - "--disable-tests"
+          - "--with-btrfs=no"
+          - "--with-dm=no"
+          - "--with-gtk-doc=no"
+          - "--with-lvm=no"
+          - "--with-lvm_dbus=no"
+          - "--with-mpath=no"
+          - "--with-nvdimm=no"
+          - "--with-escrow=no"
+          - "--with-tools=no"
+          - "--with-smart=no"
+          - "--with-smartmontools=no"
+        modules:
+          - name: libparted
+            cleanup:
+              - /bin
+              - /share
+            sources:
+              - type: archive
+                url: https://alpha.gnu.org/gnu/parted/parted-3.2.153.tar.xz
+                sha256: a1445d837ac4bc809ff9146ab3ac112ecc1b5639d23b7ea7b668d45aca058a2a
+            config-opts:
+              - --disable-device-mapper
+          - name: kmod
+            buildsystem: meson
+            sources:
+              - type: git
+                url: https://github.com/kmod-project/kmod.git
+                tag: v34
+            config-opts:
+              - -Dbashcompletiondir=no
+              - -Dzshcompletiondir=no
+              - -Dfishcompletiondir=no
+              - -Dtools=false
+              - -Dmanpages=false
+          - name: libaio
+            buildsystem: simple
+            sources:
+              - type: git
+                url: https://github.com/JingchengLi/libaio.git
+                commit: 1b18bfa
+            build-commands:
+              - make prefix=/app install
+          - name: lvm2
+            buildsystem: simple
+            sources:
+              - type: git
+                url: https://gitlab.com/lvmteam/lvm2.git
+                tag: v2_03_34
+            make-args:
+              - device-mapper
+            build-options:
+              no-debuginfo: true
+            build-commands:
+              - ./configure --prefix=/app --exec-prefix=/app --enable-pkgconfig
+              - make install_device-mapper
+              - cd libdm && make install_pkgconfig
+              - chmod 755 /app/lib/libdevmapper.so
+              - chmod 755 /app/sbin/dmsetup
+          - name: libcryptsetup
+            sources:
+              - type: git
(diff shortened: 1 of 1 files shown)


hfiguiere:
bot, build

net.ventoy.VentoyGUI.yml
```diff
@@ -0,0 +1,206 @@
...
+      - install -Dm644 ICON/logo_32.png /app/share/icons/hicolor/32x32/apps/net.ventoy.VentoyGUI.png
+      - install -Dm644 ICON/logo_48.png /app/share/icons/hicolor/48x48/apps/net.ventoy.VentoyGUI.png
+      - install -Dm644 ICON/logo_64.png /app/share/icons/hicolor/64x64/apps/net.ventoy.VentoyGUI.png
+      - install -Dm644 ICON/logo_72.png /app/share/icons/hicolor/72x72/apps/net.ventoy.VentoyGUI.png
+      - install -Dm644 ICON/logo_128.png /app/share/icons/hicolor/128x128/apps/net.ventoy.VentoyGUI.png
+      - install -Dm644 ICON/logo_256.png /app/share/icons/hicolor/256x256/apps/net.ventoy.VentoyGUI.png
+      - install -Dm644 ICON/logo_512.png /app/share/icons/hicolor/512x512/apps/net.ventoy.VentoyGUI.png
+      - install -Dm644 LinuxGUI/assets/net.ventoy.VentoyGUI.desktop /app/share/applications/net.ventoy.VentoyGUI.desktop
+      - install -Dm644 LinuxGUI/assets/net.ventoy.VentoyGUI.metainfo.xml /app/share/metainfo/net.ventoy.VentoyGUI.metainfo.xml
+finish-args:
```
hfiguiere: this should be above `modules`

net.ventoy.VentoyGUI.yml
```diff
@@ -0,0 +1,206 @@
...
+        buildsystem: autotools
+        sources:
+          - type: git
+            url: https://github.com/libatasmart/libatasmart.git
+            tag: v0.19
+  - name: ventoy_release
+    buildsystem: simple
+    sources:
+      - type: archive
+        url: https://github.com/ventoy/Ventoy/releases/download/v1.1.05/ventoy-1.1.05-linux.tar.gz
```
hfiguiere, line 177: so this binary, does it work on both aarch64 and x86_64? Why isn't this built from source?

net.ventoy.VentoyGUI.yml
```diff
@@ -0,0 +1,206 @@
...
+        url: https://github.com/LorenzoIanotto/VentoyUDisks.git
+        commit: 17aa00e
+      - type: script
+        dest-filename: ventoy-gtk3.sh
+        commands:
+          - cd /app
+          - VentoyGUI --xdg
+    build-commands:
+      - cd LinuxGUI && make gtk3
+      - install -Dm755 LinuxGUI/Ventoy2Disk.gtk3 /app/bin/VentoyGUI
```
hfiguiere, line 195: why isn't all of this part of the install rules?

net.ventoy.VentoyGUI.yml
ada-magicat: Apologies if this an intrusion, but I think this is relevant for this application submission. Even if this was built from source, the upstream "source" repository in question has a history on relying on blobs of unknown origin. See:
- https://github.com/ventoy/Ventoy/issues/132
- https://github.com/ventoy/Ventoy/issues/2795

I do not know if this falls into an exception and would still be allowed on flathub (at least reading the requirements it is not clear and proprietary applicationd are obviously allowed) but I think it is relevant context.

net.ventoy.VentoyGUI.yml
```diff
@@ -0,0 +1,206 @@
...
+        url: https://github.com/ventoy/Ventoy/releases/download/v1.1.05/ventoy-1.1.05-linux.tar.gz
+        sha256: 3379c99890359dcff55aab7f7b3286f87c988d1da2fd616e6a9e305fb0a1de9e
+        strip-components: 1
+    build-commands:
+      - cp -r ventoy-1.1.05/* /app/
+  - name: ventoy_gui
+    buildsystem: simple
+    sources:
+      - type: git
+        url: https://github.com/LorenzoIanotto/VentoyUDisks.git
```
hfiguiere, line 186: is this affiliated to ventoy? Because the location doesn't look like it is. Which mean the app-id isn't appropriate.

net.ventoy.VentoyGUI.yml
```diff
@@ -0,0 +1,206 @@
...
+              - -Dzshcompletiondir=no
+              - -Dfishcompletiondir=no
+              - -Dtools=false
+              - -Dmanpages=false
+          - name: libaio
+            buildsystem: simple
+            sources:
+              - type: git
+                url: https://github.com/JingchengLi/libaio.git
+                commit: 1b18bfa
```
hfiguiere: Please use full commit hash.

net.ventoy.VentoyGUI.yml
```diff
@@ -0,0 +1,206 @@
...
+        sha256: 3379c99890359dcff55aab7f7b3286f87c988d1da2fd616e6a9e305fb0a1de9e
+        strip-components: 1
+    build-commands:
+      - cp -r ventoy-1.1.05/* /app/
+  - name: ventoy_gui
+    buildsystem: simple
+    sources:
+      - type: git
+        url: https://github.com/LorenzoIanotto/VentoyUDisks.git
+        commit: 17aa00e
```
hfiguiere: Please use full commit hash.

net.ventoy.VentoyGUI.yml
bbhtt: then it should not be added to Flathub.

LorenzoIanotto: Sorry, I forgot to mention it early:
Ventoy consists of the program that is actually installed on the drive (which comes from grub2) and the GUI which installs it on the drive.
I am only compiling the GUI from source since it is the only code that runs on flatpak.
There is also another big problem: this thing is a packaging nightmare. It uses build scripts instead of an actual buildsystem and has a really strange way of handling dependencies.
The GUI which this app is about is compiled standalone: I have forked the project and converted the backend (which on upstream needs root access to manage dev files directly) to UDisks2, so it can be used by a non-root user. I am also moving on from the scripts and implemented a Makefile. 
I have contacted the developers but I don't know if they will answer since they did not in the past. I am willing to maintain this.

net.ventoy.VentoyGUI.yml
LorenzoIanotto: Please see https://github.com/flathub/flathub/pull/6834#discussion_r2281572567

net.ventoy.VentoyGUI.yml
LorenzoIanotto: The project doesn't have a build system, I just created a Makefile with the compile instructions. Should I expand it further?

hfiguiere: > The project doesn't have a build system

Then it's not ready to ship.

net.ventoy.VentoyGUI.yml
hfiguiere: Then it's a NO.

net.ventoy.VentoyGUI.yml
hfiguiere: > I am only compiling the GUI from source since it is the only code that runs on flatpak.

Then why is the rest in the flatpak? If you don't need it it has no business being here.

Right now, between the suspiciousnes on Ventoy and this I don't think this is a fit for Flathub.

> There is also another big problem: this thing is a packaging nightmare. It uses build scripts instead of an actual buildsystem and has a really strange way of handling dependencies.

That's a problem to take with upstream. But then, https://github.com/flathub/flathub/pull/6834#discussion_r2281033516

LorenzoIanotto: Yeah I don't really think upstream cares about solving this issues (https://github.com/ventoy/Ventoy/issues/3224). Maybe it is time to move on to something else until we have a software that is actually reproducible.
Sorry for that.

Arcitec:
First of all, thank you so much for making Ventoy compatible with Flatpaks by porting it to use udisks2.

Secondly, Ventoy is **not buildable from source without huge efforts,** because it's basically a Linux distro. Ventoy installs a Linux distro on the USB stick. That being the Linux kernel + BusyBox utilities + Fedora's SecureBoot shims and EFI binaries, etc. After installing the "Ventoy Linux Distro" on a USB stick, you can then boot from that stick and it will present a menu of ISO files that you can then chain-boot into. It's a multi-ISO boot stick tool. And it's a *pain* for users to install and update, since right now, it only exists as manually unzippable releases that people have to manually download. A Flatpak would be amazing.

You can see a list of all its required binaries in this document in section 5, "Binaries":

https://github.com/ventoy/Ventoy/blob/master/DOC/BuildVentoyFromSource.txt#L201

Ventoy was made by taking the trusted, pre-built binaries from those upstream, open-source projects and bundling them directly.

There is work being done to make Ventoy build more of the components itself, but it's a huge task and will not be done anytime soon: https://github.com/ventoy/Ventoy/issues/3224

Another issue is that Ventoy **necessarily** contains a SecureBoot MOK certificate and self-signed binaries. That is the only way to make SecureBoot work. It is the same as what you have to do to SecureBoot any other Linux distro (which as established, Ventoy basically is...). This SecureBoot support is necessary since there are ISOs that will refuse to boot if you don't have SecureBoot enabled.

Flathub already includes proprietary, commercial binary blobs of all kinds. Tons of apps here are re-bundled .deb packages where no source code is available at all. So including Ventoy which is a combination of popular open-source binaries is mild by comparison.

Should a binary exception be made for Ventoy too?

Arcitec:
Oh and I just saw that this Flatpak depends on the forked repo at https://github.com/LorenzoIanotto/VentoyUDisks.git.

That is a No for sure. Wait until Ventoy merges the Flatpak-compatible code changes instead, here:

https://github.com/ventoy/Ventoy/pull/3319

Arcitec:
I guess it's fine if this is closed by the bot for now since we're still waiting on reactions from Ventoy to upstream the UDisks support first.

PR #6851, opened 2025-08-21
Add io.github.archisman_panigrahi.screen-orientation-manager

archisman-panigrahi:

- [X] Please describe the application briefly. This application lets the user rotate the screen orientation, touchscreen input and the stylus input all in one go, using GUI, in X11 systems. While this is automatically done in GNOME or KDE in Wayland, everywhere else rotating the display does not simultaneously rotate the touch input. This app solves the problem there (many low-end devices such as ARM chromebooks are excellent tablets but are not powerful enough to run GNOME/KDE with Wayland but they fly with XFCE, where the app completely solves the issue of screen rotation). Also, this app works with GNOME and KDE with X11. It also supports quick rotation (like Android) with a system tray.
- [X] Please attach a video showcasing the application on Linux using the Flatpak. _Video updated._

https://github.com/user-attachments/assets/8663d639-3006-4f3b-a823-b265ba251d32

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "shared-modules"]
+	path = shared-modules
+	url = https://github.com/flathub/shared-modules.git


io.github.archisman_panigrahi.screen-orientation-manager.json
```diff
@@ -0,0 +1,50 @@
...
+  "command": "screen-orientation-manager",
+  "finish-args": [
+    "--share=ipc",
+    "--socket=fallback-x11",
+    "--socket=wayland",
+    "--filesystem=/sys/devices/virtual/dmi/id/product_name:ro",
+    "--filesystem=/sys/firmware/devicetree/base/model:ro"
+  ],
+  "modules": [
+    "shared-modules/libayatana-appindicator/libayatana-appindicator-gtk3.json",
```
hfiguiere, line 15: shared-modules are missing.

archisman-panigrahi:
bot, build

archisman-panigrahi:
bot, build

io.github.archisman_panigrahi.screen-orientation-manager.json
archisman-panigrahi: Thanks for pointing out. Totally forgot to add the submodule.

bbhtt:
This sounds like it should not be an app but something in system settings.

archisman-panigrahi:
It is not within my or anyone's capabilities to fix the issue on all desktop environments by adding such a functionality to each of their system settings. 

But this app runs in the background and works in  every desktop environment and supports all kind of devices. It is a very useful utility tool.

Also by the same logic, apps like https://flathub.org/apps/de.capypara.FieldMonitor or https://flathub.org/apps/io.github.realmazharhussain.GdmSettings or 
https://flathub.org/apps/app.polychromatic.controller should have been an inbuilt system settings rather than being a separate app.

But, each of these apps perform a very specific and useful task, and having more such utility tools in flathub will enrich the variety of available software in flathub and will also enrich the general linux user experience

hfiguiere:
Each time you say "but these other do it" you lose points in the argument. Two wrongs don't make a right. And you are exactly why making an exception is always a burden because there will be 10 people after to say "but..."

bbhtt:
I also have some concerns about the project. The app here seems to be a single 500 line Python file with some shell scripts?

archisman-panigrahi:
> The app here seems to be a single 500 line Python file with some shell scripts?

Yes, but could you tell me why is that concerning? The python file creates the GUI, and detects the device. It is also responsible for saving and restoring the settings. Then it invokes the appropriate shell script to call `xinput` and `xrandr`. And the python file is divided into separate functions that do individual things.

> Each time you say "but these other do it" you lose points in the argument.

I agree with your point about "but these other do it" when it comes to someone justifying their corruption or self harm. But this app is for the benefit of the users and so are the other apps. That is all I meant to say.

However I don't see why a GUI utility tool cannot be an app in flathub. This app satisfies all the requirements in https://docs.flathub.org/docs/for-app-authors/requirements#inclusion-policy. And it is not the same kind of system utility as sudo or xeyes. This is a full GUI app with touch screen support (please see the video above).

---

With the current situation in desktop Linux, this is the only possible solution I could come up with to support screen rotation of tablets (alternative is to ask every desktop environment make the same thing in their settings). 

While I initially made the app to solve my own problem (screen orientation of a touch screen tablet), later it turned out to be helpful for many other people (e.g. [ARM chromebooks](https://github.com/velvet-os/imagebuilder/issues/334) and [desktop automation](https://github.com/archisman-panigrahi/screen-orientation-manager/issues/3) projects). 

As I realized many people find this helpful, I wanted to publish it in flathub so that more people can find it easily and enjoy better capabilities of their touchscreen tablets and convertibles.

archisman-panigrahi:
I updated the video of the app.

archisman-panigrahi:
@bbhtt @hfiguiere Are you waiting for any changes? I think this app satisfies all the requirements in the submission policy, and including it will enrich flathub, as well as help users.

Please let me know if there is something I can do from my side.

hfiguiere:
https://github.com/flathub/flathub/pull/6851#issuecomment-3212832692 remains. No amount of impatience will tilt the balance in your way.

archisman-panigrahi:
Sorry, but I didn't understand the point of concern there. Please explain.

archisman-panigrahi:
> I also have some concerns about the project. The app here seems to be a single 500 line Python file with some shell scripts?

Which requirement of https://docs.flathub.org/docs/for-app-authors/requirements#inclusion-policy does it not satisfy?

PR #6860, opened 2025-08-24
Add io.github.ogoshen.AppRyzen

ogoshen:

- [x] Please describe the application briefly. Power management app for Ryzen processors.
[Screencast From 2025-08-23 20-33-54.webm](https://github.com/user-attachments/assets/2af51972-a0bd-4e09-b3c2-31221f726e6d)

- [x] I am an author contributor to the project.

### DIFF
--- a/io.github.ogoshen.AppRyzen.json
+++ b/io.github.ogoshen.AppRyzen.json
@@ -1,119 +1,109 @@
 {
-    "id": "io.github.ogoshen.AppRyzen",
-    "runtime": "org.gnome.Platform",
-    "runtime-version": "48",
-    "sdk": "org.gnome.Sdk",
-    "command": "appryzen",
-    "finish-args": [
-        "--share=network",
-        "--share=ipc",
-        "--socket=fallback-x11",
-        "--device=dri",
-        "--socket=wayland",
-        "--talk-name=org.freedesktop.Flatpak",
-        "--filesystem=/sys:ro"
-    ],
-    "build-options": {
-        "append-path": "/usr/lib/sdk/vala/bin",
-        "prepend-ld-library-path": "/usr/lib/sdk/vala/lib"
-    },
-    "cleanup": [
-        "/include",
-        "/lib/pkgconfig",
-        "/man",
-        "/share/doc",
-        "/share/gtk-doc",
-        "/share/man",
-        "/share/pkgconfig",
-        "/share/vala",
-        "*.la",
-        "*.a"
-    ],
-    "modules": [
-
-
-	{
-		"name": "ryzenadj",
-		"buildsystem": "cmake-ninja",
-		"builddir": true,
-		"config-opts": [
-			"-DCMAKE_BUILD_TYPE=Release",
-			"-DBUILD_SHARED_LIBS=OFF"
-		],
-		"sources": [
-			{
-				"type": "git",
-				"url": "https://github.com/FlyGoat/RyzenAdj.git",
-				"tag": "v0.17.0"
-			},
-			{
-				"type": "patch",
-				"path": "ryzenadj.patch"
-			}
-		]
-	},
-
-	{
-	  "name": "lm-sensors",
-	  "buildsystem": "simple",
-	  "builddir": true,
-	  "build-commands": [
-	    "make -j $FLATPAK_BUILDER_N_JOBS",
-	    "make install BUILD_STATIC_LIB=0 DEBUG=0 PREFIX=/ DESTDIR=$FLATPAK_DEST"
-	  ],
-	  "sources": [
-	    {
-	      "type": "git",
-	      "url": "https://github.com/lm-sensors/lm-sensors.git",
-	      "tag": "V3-6-0",
-	      "commit": "1667b850a1ce38151dae17156276f981be6fb557"
-	    }
-	  ]
-	},
-
-        {
-	    "name" : "live-chart",
-	    "builddir" : true,
-	    "buildsystem" : "meson",
-	    "sources" : [
+	"id": "io.github.ogoshen.AppRyzen",
+	"runtime": "org.gnome.Platform",
+	"runtime-version": "48",
+	"sdk": "org.gnome.Sdk",
+	"command": "appryzen",
+	"finish-args": [
+		"--share=network",
+		"--share=ipc",
+		"--socket=fallback-x11",
+		"--device=dri",
+		"--socket=wayland",
+		"--talk-name=org.freedesktop.Flatpak",
+		"--filesystem=/sys:ro"
+	],
+	"cleanup": [
+		"/include",
+		"/lib/pkgconfig",
+		"/man",
+		"/share/doc",
+		"/share/gtk-doc",
+		"/share/man",
+		"/share/pkgconfig",
+		"/share/vala",
+		"*.la",
+		"*.a"
+	],
+	"modules": [
 		{
-		    "type" : "git",
-		    "url" : "https://github.com/lcallarec/live-chart.git",
-		    "commit" : "9d34b48"
+			"name": "ryzenadj",
+			"buildsystem": "cmake-ninja",
+			"builddir": true,
+			"config-opts": [
+				"-DCMAKE_BUILD_TYPE=Release",
+				"-DBUILD_SHARED_LIBS=OFF"
+			],
+			"sources": [
+				{
+					"type": "git",
+					"url": "https://github.com/FlyGoat/RyzenAdj.git",
+					"tag": "v0.17.0"
+				},
+				{
+					"type": "patch",
+					"path": "ryzenadj.patch"
+				}
+			]
 		},
 		{
-		    "type" : "patch",
-		    "path" : "live-chart.patch"
-		}
-	    ]
-	},
-
-        {
-	    "name": "blueprint-compiler",
-	    "buildsystem": "meson",
-	    "cleanup": [
-		"*"
-	    ],
-	    "sources": [
+			"name": "lm-sensors",
+			"buildsystem": "simple",
+			"builddir": true,
+			"build-commands": [
+				"make -j $FLATPAK_BUILDER_N_JOBS",
+				"make install BUILD_STATIC_LIB=0 DEBUG=0 PREFIX=$FLATPAK_DEST"
+			],
+			"sources": [
+				{
+					"type": "git",
+					"url": "https://github.com/lm-sensors/lm-sensors.git",
+					"tag": "V3-6-0",
+					"commit": "1667b850a1ce38151dae17156276f981be6fb557"
+				}
+			]
+		},
 		{
-		    "type": "git",
-		    "url": "https://gitlab.gnome.org/jwestman/blueprint-compiler.git",
-		    "tag": "v0.16.0"
+			"name": "live-chart",
+			"builddir": true,
+			"buildsystem": "meson",
+			"sources": [
+				{
+					"type": "git",
+					"url": "https://github.com/lcallarec/live-chart.git",
+					"commit": "9d34b487602919675dd78f7c91e1f27c26eb874b"
+				},
+				{
+					"type": "patch",
+					"path": "live-chart.patch"
+				}
+			]
+		},
+		{
+			"name": "blueprint-compiler",
+			"buildsystem": "meson",
+			"cleanup": [
+				"*"
+			],
+			"sources": [
+				{
+					"type": "git",
+					"url": "https://gitlab.gnome.org/jwestman/blueprint-compiler.git",
+					"tag": "v0.16.0"
+				}
+			]
+		},
+		{
+			"name": "AppRyzen",
+			"builddir": true,
+			"buildsystem": "meson",
+			"sources": [
+				{
+					"type": "git",
+					"url": "https://github.com/ogoshen/AppRyzen.git",
+					"tag": "v1.0.1"
+				}
+			]
 		}
-	    ]
-	},
-
-        {
-            "name": "AppRyzen",
-            "builddir": true,
-            "buildsystem": "meson",
-            "sources": [
-                {
-                    "type": "git",
-                    "url": "https://github.com/ogoshen/AppRyzen.git",
-                    "tag": "v1.0.1"
-                }
-            ]
-        }
-    ]
+	]
 }


io.github.ogoshen.AppRyzen.json
```diff
@@ -0,0 +1,55 @@
...
+    ],
+    "command": "appryzen",
+    "finish-args": [
+        "--share=network",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--device=dri",
+        "--socket=wayland",
+        "--talk-name=org.freedesktop.Flatpak",
+        "--talk-name=io.github.ogoshen.AppRyzen.Server",
```
zefr0x:
```suggestion
```

This access is granted automatically by Flatpak and there is no need to specify manually.

io.github.ogoshen.AppRyzen.json
```diff
@@ -0,0 +1,55 @@
...
+        "org.freedesktop.Sdk.Extension.vala"
+    ],
+    "command": "appryzen",
+    "finish-args": [
+        "--share=network",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--device=dri",
+        "--socket=wayland",
+        "--talk-name=org.freedesktop.Flatpak",
```
zefr0x: If this was necessary, you need to file for an exception: https://docs.flathub.org/docs/for-app-authors/linter#exceptions

io.github.ogoshen.AppRyzen.json
```diff
@@ -0,0 +1,55 @@
...
+        "/share/pkgconfig",
+        "/share/vala",
+        "*.la",
+        "*.a"
+    ],
+    "modules": [
+        "ryzenadj.json",
+        "lm-sensors.json",
+        "live-chart.json",
+        "blueprint-compiler.json",
```
zefr0x: As these are not auto generated, they should be inline in the manifest. it's harder to maintain and review.

ryzenadj.json
```diff
@@ -0,0 +1,20 @@
+{
+	"name": "ryzenadj",
+	"buildsystem": "cmake-ninja",
+	"builddir": true,
+	"config-opts": [
+		"-DCMAKE_BUILD_TYPE=Release",
+		"-DBUILD_SHARED_LIBS=OFF"
```
zefr0x: Is it not possible to use shared libs? Otherwise, you should use dynamic linking

io.github.ogoshen.AppRyzen.json
```diff
@@ -0,0 +1,55 @@
...
+        "--device=dri",
+        "--socket=wayland",
+        "--talk-name=org.freedesktop.Flatpak",
+        "--talk-name=io.github.ogoshen.AppRyzen.Server",
+        "--filesystem=/sys:ro"
+    ],
+    "build-options": {
+        "append-path": "/usr/lib/sdk/vala/bin",
+        "prepend-ld-library-path": "/usr/lib/sdk/vala/lib",
+        "env": {}
```
zefr0x:
```suggestion
```

io.github.ogoshen.AppRyzen.json
```diff
@@ -0,0 +1,55 @@
+{
+    "id": "io.github.ogoshen.AppRyzen",
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "47",
```
zefr0x: Can't the latest `48` version be used?

live-chart.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name" : "live-chart",
+    "builddir" : true,
+    "buildsystem" : "meson",
+    "sources" : [
+        {
+            "type" : "git",
+            "url" : "https://github.com/lcallarec/live-chart.git",
+            "tag" : "master"
```
hfiguiere: `master` is not a tag. don't try to cheat to use branch. Add a real tag and/or a commit.

lm-sensors.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "lm-sensors",
+  "buildsystem": "simple",
+  "builddir": true,
+  "build-commands": [
+    "sed -i -e 's/ -g / -s /' Makefile",
+    "make -j $FLATPAK_BUILDER_N_JOBS",
+    "find . -type f | perl -lne 'print if -B and -x' | xargs strip",
```
hfiguiere: why strip? There is no reason to strip binaries.

io.github.ogoshen.AppRyzen.json
```diff
@@ -0,0 +1,55 @@
...
+    "sdk": "org.gnome.Sdk",
+    "sdk-extensions": [
+        "org.freedesktop.Sdk.Extension.vala"
```
hfiguiere: valac is in the GNOME SDK. Why do you need this?

io.github.ogoshen.AppRyzen.json
bbhtt: the first question is what is this permission needed for?

flathub.json
```diff
@@ -0,0 +1,4 @@
+{
+  "only-arches": ["x86_64"]
+}
+
```
bbhtt, line 4: why is it x86_64 only?

ogoshen: This is for x86 Ryzen CPUs exclusively

io.github.ogoshen.AppRyzen.json
ogoshen: Confirmed, compiles when removed.
IIRC was auto-added by Gnome builder

ryzenadj.json
ogoshen: Not cleanly, at least that I know of, I'll elaborate.

The RyzenAdj library is used to access the SMU in a protected part of memory either via libpci which requires root access, or a DKMS which users might not have, and flatpak can't install.

I use `flatpak-spawn` and `pkexec` to launch a d-bus daemon to handle that part.
`pkexec` will clear LD_LIBRARY_PATH as stated here https://linux.die.net/man/1/pkexec
And so, the program will actually not use flatpak's lib path and will try to load the host library.

Now there's a risk for mismatched library and Vala binding code.

io.github.ogoshen.AppRyzen.json
ogoshen: Nice, is it because the prefixes match?

io.github.ogoshen.AppRyzen.json
ogoshen: I'll file an exception then, needed to launch the d-bus server with root permission to access the SMU.

io.github.ogoshen.AppRyzen.json
zefr0x: > is it because the prefixes match?

yes

io.github.ogoshen.AppRyzen.json
bbhtt: can you explain some more?

ogoshen: Sure, the System Managment Unit in Ryzen processors manages power, current and thermals. 
For this app to actually be able to read and adjust CPU power limits, a d-bus server is (flatpak-)spawned on the host with root permissions requested through pkexec.
The server then uses [libryzenadj](https://github.com/FlyGoat/RyzenAdj) which interacts with the SMU over libpci and /dev/mem - which requires root access.

The app itself does not use root permissions, and can run in a crippled way without them given, with only sensor monitoring available.

io.github.ogoshen.AppRyzen.json
ogoshen: Really? that's a bit counter-intuitive.. 
Couldn't actually find any recommendation on what to use, so I used the lowest runtime that had the features I want.. 
By targeting 48 I'm not restricting users?

hfiguiere: no it's not. The GNOME 47 runtime will be EOL when GNOME 49 is released next month (September) or shortly thereafter.

> By targeting 48 I'm not restricting users?

This is a fundamental misunderstanding of what flatpak provides. https://docs.flatpak.org/en/latest/basic-concepts.html#basic-concepts

hfiguiere: > Couldn't actually find any recommendation on what to use, so I used the lowest runtime that had the features I want..

Like here:
https://docs.flathub.org/docs/for-app-authors/runtimes#currently-hosted-runtimes

And here:

https://docs.flatpak.org/en/latest/available-runtimes.html#check-software-available-in-runtimes

hfiguiere:
```suggestion
    "runtime-version": "48",
```

io.github.ogoshen.AppRyzen.json
bbhtt: > The app itself does not use root permissions, and can run in a crippled way without them given, with only sensor monitoring available.

sounds like it should not be flatpak-ed then.

io.github.ogoshen.AppRyzen.json
ogoshen: Thanks, clearer now.

> Major version releases of the runtime are synced with GNOME releases and are announced on GNOME Discourse. Usually a given branch of the runtime is supported for an year and EOL-ed upon the release of a newstable version.

io.github.ogoshen.AppRyzen.json
ogoshen: can you explain some more?

ogoshen:
bot, build

io.github.ogoshen.AppRyzen.json
ogoshen: Filed the exception PR which is blocked too, could we unblock this please, get compliance and deal with whether we should later?

io.github.ogoshen.AppRyzen.json
```diff
@@ -0,0 +1,119 @@
...
+        "--socket=fallback-x11",
+        "--device=dri",
+        "--socket=wayland",
+        "--talk-name=org.freedesktop.Flatpak",
+        "--filesystem=/sys:ro"
+    ],
+    "build-options": {
+        "append-path": "/usr/lib/sdk/vala/bin",
+        "prepend-ld-library-path": "/usr/lib/sdk/vala/lib"
+    },
```
hfiguiere: I don't see where this is needed.

io.github.ogoshen.AppRyzen.json
```diff
@@ -0,0 +1,119 @@
...
+
+        {
+	    "name" : "live-chart",
+	    "builddir" : true,
+	    "buildsystem" : "meson",
+	    "sources" : [
+		{
+		    "type" : "git",
+		    "url" : "https://github.com/lcallarec/live-chart.git",
+		    "commit" : "9d34b48"
```
hfiguiere: use full commits.

io.github.ogoshen.AppRyzen.json
```diff
@@ -0,0 +1,119 @@
...
+		]
+	},
+
+	{
+	  "name": "lm-sensors",
+	  "buildsystem": "simple",
+	  "builddir": true,
+	  "build-commands": [
+	    "make -j $FLATPAK_BUILDER_N_JOBS",
+	    "make install BUILD_STATIC_LIB=0 DEBUG=0 PREFIX=/ DESTDIR=$FLATPAK_DEST"
```
hfiguiere: `PREFIX` is incorrect. so is `DESTDIR`.

io.github.ogoshen.AppRyzen.json
```diff
@@ -0,0 +1,119 @@
...
+    "modules": [
+
+
+	{
+		"name": "ryzenadj",
+		"buildsystem": "cmake-ninja",
+		"builddir": true,
+		"config-opts": [
+			"-DCMAKE_BUILD_TYPE=Release",
+			"-DBUILD_SHARED_LIBS=OFF"
```
hfiguiere: this should build dynamic libraries.

hfiguiere:
manifest indentiation is all over the place

image

io.github.ogoshen.AppRyzen.json
ogoshen: Guess that's from gnome builder's default vala project template, I'll remove it gladly

io.github.ogoshen.AppRyzen.json
ogoshen: See https://github.com/flathub/flathub/pull/6860#discussion_r2310492620

ogoshen:
bot, build

io.github.ogoshen.AppRyzen.json
hfiguiere: so to break the sandbox. I guess this https://github.com/flathub/flathub/pull/6860#discussion_r2312191457 is still very relevant.

io.github.ogoshen.AppRyzen.json
bbhtt: From what you described, it seems like this is not really suited to be run inside a Flatpak.

ogoshen: 1. There are 91 instances of `"finish-args-flatpak-spawn-access": "the app predates this linter rule"` in the exception file.
2. Any of those are free to request elevated access, some of them do for a fact.
3. Most of the other entries with that permission, are indeed system tools, such as this app.
4. I agree, this isn't very well suited to be Flatpak'ed, and it's why I'm also pursuing other distribution methods.
5. But, I would still very much like this to be on Flathub. Handhelds, console-likes and other immutable Linux distros are a large portion of the target audience, and in that respect, flatpak is very much suitable.
6. This app is FOSS, anyone with concerns is free to inspect, modify and distribute.

I hope that's enough to reconsider, unblock the exception at https://github.com/ogoshen/flatpak-builder-lint/commit/a6df7a1dafc612be445d2ee9b116d9347777c569 or otherwise move this forward.

bbhtt: but you said the app won't be fully functional without root permissions and that will not work inside a Flatpak?

ogoshen: Not what I said, and makes no sense to submit a non working app.

App works fully with root permissions, obtained as described below, albeit the sandboxing exception.

Works, and shows a banner explaining the missing functions without root.

I'll post some screenshots later to make this clearer, I now realize I left out those parts from the video, sorry about that.

bbhtt: Inside the Flatpak sandbox you cannot obtain root, so I take that whatever uses root doesn't work and this is semi-functional?

ogoshen: Yes, when the daemon isn't running there's a banner shown, and Ryzen CPU monitoring is disabled, only generic libsensors is active.

Screenshot From 2025-10-16 20-06-48

Via flatpak-spawn, the pkexec dialog is shown either on startup, or when the banner button is clicked.

pkexec

Cancelling the dialog takes you back to the state in the first image, successfully authenticating launches the dbus daemon, which enables all app features as in the image below.

image

On second thought, the banner isn't very explanatory... I'll consider adding some text there.

ogoshen: Can I get a follow up please?

bbhtt: what followup do you want? you said you were going to change some things in the last conversation.

ogoshen: > On second thought, the banner isn't very explanatory... I'll consider adding some text there.

That was me on a reflection tangent, about changing a small app detail.

The followup is about moving on with the review process, unblocking the build pipeline and getting compliance .

If you're deliberately trying to get me to throw the towel here, you're doing a good job.

I ask you to be explicit in your next comment on what needs to happen to make this work, or be more convincing on why you think it can't.

bbhtt: With that reply this is never getting in.

In case it isn't explicit enough I will close it.

PR #6861, opened 2025-08-24
Add org.duckdns.bugfreeblog.mldonkeynext

carlonluca:
- [x] Please describe the application briefly. < GUI for the [mldonkey](https://en.wikipedia.org/wiki/MLDonkey) >
- [x] Please attach a video showcasing the application on Linux using the Flatpak. < [Screencast_20250824_194549.webm](https://github.com/user-attachments/assets/9e1ae317-2500-4cc4-ac09-45ddd2df9dea)
 >

### DIFF
--- a/org.duckdns.bugfreeblog.mldonkeynext.json
+++ b/org.duckdns.bugfreeblog.mldonkeynext.json
@@ -1,0 +1,48 @@
+{
+  "app-id": "org.duckdns.bugfreeblog.mldonkeynext",
+  "runtime": "org.kde.Platform",
+  "runtime-version": "6.9",
+  "sdk": "org.kde.Sdk",
+  "base": "io.qt.qtwebengine.BaseApp",
+  "base-version": "6.9",
+  "command": "mldonkey-next",
+  "finish-args": [
+    "--share=network",
+    "--socket=wayland",
+    "--device=dri",
+    "--share=ipc",
+    "--share=network",
+    "--socket=fallback-x11"
+  ],
+  "modules": [
+    {
+      "name": "mldonkey-next",
+      "buildsystem": "simple",
+      "build-commands": [
+        "./create_qt_resource_files.sh",
+        "cmake -DCMAKE_INSTALL_PREFIX=/app .",
+        "cmake --build .",
+        "cmake --install .",
+        "install -Dm644 flatpak/org.duckdns.bugfreeblog.mldonkeynext.desktop /app/share/applications/org.duckdns.bugfreeblog.mldonkeynext.desktop",
+        "install -Dm644 flatpak/org.duckdns.bugfreeblog.mldonkeynext.metainfo.xml /app/share/metainfo/org.duckdns.bugfreeblog.mldonkeynext.metainfo.xml",
+        "install -Dm644 flatpak/org.duckdns.bugfreeblog.mldonkeynext.svg /app/share/icons/hicolor/scalable/apps/org.duckdns.bugfreeblog.mldonkeynext.svg",
+        "install -Dm644 flatpak/org.duckdns.bugfreeblog.mldonkeynext_128.png /app/share/icons/hicolor/128x128/apps/org.duckdns.bugfreeblog.mldonkeynext.png",
+        "install -Dm644 flatpak/org.duckdns.bugfreeblog.mldonkeynext_512.png /app/share/icons/hicolor/512x512/apps/org.duckdns.bugfreeblog.mldonkeynext.png"
+      ],
+      "subdir": "app",
+      "sources": [
+        {
+          "type": "git",
+          "url": "https://github.com/carlonluca/mldonkey-next.git",
+          "tag": "v0.14.5"
+        },
+        {
+          "type": "archive",
+          "url": "https://github.com/carlonluca/mldonkey-next/releases/download/v0.14.5/webapp-v0.14.5.zip",
+          "dest": "app/webapp",
+          "sha256": "c401221ccd15982a7cdfa4eed31e5092979bfd49fdea39e50fe1e625b19b9b17"
+        }
+      ]
+    }
+  ]
+}


org.duckdns.bugfreeblog.mldonkeynext.json
```diff
@@ -0,0 +1,48 @@
...
+  ],
+  "modules": [
+    {
+      "name": "mldonkey-next",
+      "buildsystem": "simple",
+      "build-commands": [
+        "./create_qt_resource_files.sh",
+        "cmake -DCMAKE_INSTALL_PREFIX=/app .",
+        "cmake --build .",
+        "cmake --install .",
```
hfiguiere: this should be `buildsystem: cmake-ninja`

org.duckdns.bugfreeblog.mldonkeynext.json
```diff
@@ -0,0 +1,48 @@
...
+        "install -Dm644 flatpak/org.duckdns.bugfreeblog.mldonkeynext.desktop /app/share/applications/org.duckdns.bugfreeblog.mldonkeynext.desktop",
```
hfiguiere: why aren't these non flatpak specific files installed by the install rule of the build system?

org.duckdns.bugfreeblog.mldonkeynext.json
```diff
@@ -0,0 +1,48 @@
...
+    "--share=ipc",
+    "--share=network",
+    "--socket=fallback-x11"
+  ],
+  "modules": [
+    {
+      "name": "mldonkey-next",
+      "buildsystem": "simple",
+      "build-commands": [
+        "./create_qt_resource_files.sh",
```
hfiguiere: this should also be done by the buildsystem.

org.duckdns.bugfreeblog.mldonkeynext.json
carlonluca: This seemed like the proper way to do it, but if I do that, build-commands are not executed **before** building with cmake. I need to call the create_qt_resource_files.sh script before cmake runs. The only solution I could find is to build manually like this. Is there any other way to add custom commands before the cmake commands?

org.duckdns.bugfreeblog.mldonkeynext.json
carlonluca: Because the install rule was not instructed to do it. I only want the flatpak to use the icons.

org.duckdns.bugfreeblog.mldonkeynext.json
carlonluca: This is done manually by the user during the development. Not intended to be run every time cmake runs.

carlonluca:
bot, build

org.duckdns.bugfreeblog.mldonkeynext.json
hfiguiere: 1. build commands don't make sense with anything but `simple`
2. There are `shell` source if you really must. But this should be part of the build system.

org.duckdns.bugfreeblog.mldonkeynext.json
hfiguiere: why would the desktop file be different? This doesn't make sense.

org.duckdns.bugfreeblog.mldonkeynext.json
hfiguiere: this is what targets are for: only build when needed. If the must be run before the rest of the build then it should be in the build system.

org.duckdns.bugfreeblog.mldonkeynext.json
hfiguiere: same for the icons. Why would they be different?

bbhtt:
> The domain to be used for verification is bugfreeblog.duckdns.org. If you intend to verify this app, please confirm that you can upload https://bugfreeblog.duckdns.org/.well-known/org.flathub.VerifiedApps.txt. Otherwise, ignore this. Please comment if this incorrect.

Are you able to do this with these duckdns.org websites?

org.duckdns.bugfreeblog.mldonkeynext.json
carlonluca: There is no specific implementation for thr install rule for icons or desktop file and this is intended.

org.duckdns.bugfreeblog.mldonkeynext.json
carlonluca: Sorry but it is not intended to be run when cmake runs. It is intended to only run with other scripts (not needed in this flatpak) when the developer decides it is the time to run it.

org.duckdns.bugfreeblog.mldonkeynext.json
hfiguiere: that's a bug. The intent is incorrect.

carlonluca: If you consider this a showstopper, feel free to drop this PR.

carlonluca:
> > The domain to be used for verification is bugfreeblog.duckdns.org. If you intend to verify this app, please confirm that you can upload https://bugfreeblog.duckdns.org/.well-known/org.flathub.VerifiedApps.txt. Otherwise, ignore this. Please comment if this incorrect.
> 
> Are you able to do this with these duckdns.org websites?

Hello, yes, I added that file.

org.duckdns.bugfreeblog.mldonkeynext.json
hfiguiere: This is https://github.com/flathub/flathub/pull/6861#discussion_r2299566504

carlonluca: There is no request for a change in that comment. I only see two options.

carlonluca:
bot, build

org.duckdns.bugfreeblog.mldonkeynext.json
carlonluca: Updated.

org.duckdns.bugfreeblog.mldonkeynext.json
carlonluca: Updated.

org.duckdns.bugfreeblog.mldonkeynext.json
carlonluca: Updated.

PR #6877, opened 2025-08-28
Add io.github.vanillajonathan.pyrose

vanillajonathan:

- [x] Please describe the application briefly. < PyRose lets a user write some Python code and run it >

https://github.com/user-attachments/assets/cfb2191f-a034-4ca2-bf65-1eecc885d994

### DIFF
--- a/io.github.vanillajonathan.pyrose.json
+++ b/io.github.vanillajonathan.pyrose.json
@@ -1,0 +1,71 @@
+{
+    "id": "io.github.vanillajonathan.pyrose",
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "48",
+    "sdk": "org.gnome.Sdk",
+    "command": "pyrose",
+    "finish-args": [
+        "--share=ipc",
+        "--device=dri",
+        "--socket=fallback-x11",
+        "--socket=wayland"
+    ],
+    "cleanup": [
+        "/include",
+        "/lib/pkgconfig",
+        "/man",
+        "/share/doc",
+        "/share/gtk-doc",
+        "/share/man",
+        "/share/pkgconfig",
+        "*.la",
+        "*.a"
+    ],
+    "modules": [
+        {
+            "name": "vte",
+            "buildsystem": "meson",
+            "config-opts": [
+                "-Ddocs=false",
+                "-Dvapi=false"
+            ],
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://download.gnome.org/sources/vte/0.78/vte-0.78.5.tar.xz",
+                    "sha256": "bab93bff8d8acb05fc7c81a469e0d17c853da3851ce4be2a66d4d18772163a46",
+                    "x-checker-data": {
+                        "type": "gnome",
+                        "name": "vte",
+                        "stable-only": true
+                    }
+                }
+            ]
+        },
+        {
+            "name": "lsp-types",
+            "buildsystem": "simple",
+            "build-commands": [
+                "pip3 install --verbose --exists-action=i --no-index --find-links=\"file://${PWD}\" --prefix=${FLATPAK_DEST} \"lsp-types\""
+            ],
+            "sources": [
+                {
+                    "type": "file",
+                    "url": "https://files.pythonhosted.org/packages/f1/c8/374ef5d053af09c1e9c5889e75a9f9e419484a792b36cc600bc80e5ef27e/lsp_types-0.10.0-py3-none-any.whl",
+                    "sha256": "53fe04ed3807472ce165a51cbe6c7be487ccb8b0891e4008a2fc111acffc0283"
+                }
+            ]
+        },
+        {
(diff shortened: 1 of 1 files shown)


io.github.vanillajonathan.pyrose.json
```diff
@@ -0,0 +1,71 @@
...
+            ]
+        },
+        {
+            "name": "pyrose",
+            "builddir": true,
+            "buildsystem": "meson",
+            "sources": [
+                {
+                    "type": "dir",
+                    "path": "."
```
hfiguiere: how will it find the sources?

vanillajonathan: Oh, my  bad, I missed that. I updated it to point to the project's GitHub repository.

bbhtt:
what happens when python code by user starts requiring arbitrary modules from pypi? As you can't as freely install them inside Flatpak like on a distro. Is it a good idea to put this in a Flatpak?

vanillajonathan:
The application is intended to serve as a playground for the user to fiddle around and experiment with some code, it is not intended to be a IDE and hence does not have support for package management, unit tests, debugging, etc, it doesn't even let you open or save files so I don't see it as a big problem if users cant load arbitrary modules from pypi.

As a response to your concern I have added a preferences dialog which lets the user set the `PYTHONPATH` environment variable which is used to augment Python's default search path for module files.

bbhtt:
If it allows writing `import foobar` and running that then it will fail it doesn't need to intend to anything.

vanillajonathan:
Users can import packages by placing the packages in their `~/.local/lib/python3.12/site-packages` directory.

bbhtt:
That won't work, runtimes set PYTHONUSERBASE to a location in the sandbox so that it doesn't try to load python modules from host.

Mixing host modules inside the runtime will not work. They might be on different python versions.

vanillajonathan:
By setting the `PYTHONPATH` environment variable which augments Python's default search path for module it appears that I can indeed import modules from `~/.local/lib/python3.12/site-packages`.

I noted that there are other software on Flathub such as [Workbench](https://flathub.org/en/apps/re.sonny.Workbench) which lets user to write `import foobar` which fails.

I am open for any suggestions, ideas and proposals.

bbhtt:
Ok if it works, but not really great

vanillajonathan:
Perhaps I could use `flatpak-spawn` with the `--host` option to spawn the Python process on the host but that would require the the addition of a permission then all code would run outside the sandbox. I would prefer to run it inside the sandbox even if it means third-party packages are not available.

vanillajonathan:
I updated the runtime to GNOME 50. I would like it merged if anyone can take a look.

io.github.vanillajonathan.pyrose.json
```diff
@@ -0,0 +1,73 @@
...
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "50",
+    "sdk": "org.gnome.Sdk",
+    "command": "pyrose",
+    "finish-args": [
+        "--share=ipc",
+        "--device=dri",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--filesystem=~/.local:ro"
```
bbhtt, line 12: why is this needed?

io.github.vanillajonathan.pyrose.json
```diff
@@ -0,0 +1,73 @@
...
+        },
+        {
+            "name": "pyrose",
+            "builddir": true,
+            "buildsystem": "meson",
+            "sources": [
+                {
+                    "type": "git",
+                    "url": "https://github.com/vanillajonathan/pyrose",
+                    "tag": "v1.0.2"
```
bbhtt: you have to add commit after the tag

io.github.vanillajonathan.pyrose.json
vanillajonathan: It look for [`pyrefly`](https://pyrefly.org/) in `~/.local/bin/`.

bbhtt: why is it not included in the manifest here?

vanillajonathan: Because it has to be installed with the package manager [`uv`](https://docs.astral.sh/uv/). Pyrefly is a binary available for x86_64 and arm64.

bbhtt: i don't understand what do you mean by "it has to be installed with uv"? why is that a blocker to adding it to the manifest? it seems a regular python dependency to me.

vanillajonathan: No, Pyrefly is not a Python script, it is binary written in Rust.

bbhtt: then it is not buildable with cargo?

barthalion:
This seems acceptable but pyrefly binary should just be bundled with the app.

PR #6888, opened 2025-09-02
Add io.github.htkhiem.Euphonica

htkhiem:

- [x] Please describe the application briefly. Euphonica is a GUI client for Music Player Daemon (MPD) written in Rust and libadwaita. It offers features such as synced lyrics, automatic metadata & art fetching, audio visualiser, etc.
- [x] Please attach a video showcasing the application on Linux using the Flatpak. _Video below:_

https://github.com/user-attachments/assets/6431c5fe-ef0b-499d-84b6-152976b9e47e


flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64"]
```
hfiguiere: why?

htkhiem: Hi, thanks for the quick reply. It's there because I haven't tested building on aarch64 yet (I don't have access to an Arm system), _and_ the PipeWire backend of the visualiser might not be portable across architectures. It is currently only implemented for little-endian floats, though this might be available via on-the-fly conversion by PipeWire?

htkhiem: Question: I'm aware that dropping support for an arch later (after having released versions for them) will cause problems, but can one _add_ support for new archs safely at a later time, once sufficient refactoring & refactoring has been performed?

xamyl:
this looks so cool i want this so bad

flathub.json
hfiguiere: experience show it never happens. I don't see why you can't even try to build it on aarch64. It will never be tested if no one builds it.

htkhiem:
bot, build

flathub.json
htkhiem: I've just removed `flathub.json` & built it successfully on aarch64. Sorry for the noise.

io.github.htkhiem.Euphonica.json
```diff
@@ -0,0 +1,66 @@
...
+        "org.freedesktop.Sdk.Extension.llvm19"
+    ],
+    "command" : "euphonica",
+    "finish-args" : [
+        "--share=network",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--device=dri",
+        "--socket=wayland",
+        "--talk-name=org.freedesktop.secrets",
```
bbhtt: does it not support the secrets portal? do you use libsecret? libsecret uses the portal by default inside flatpak.

io.github.htkhiem.Euphonica.json
```diff
@@ -0,0 +1,66 @@
...
+        "--filesystem=xdg-run/pipewire-0"
```
bbhtt: this should be :ro?

io.github.htkhiem.Euphonica.json
htkhiem: During early development I wanted to keep the code portable so I used the [keyring](https://docs.rs/keyring/latest/keyring/) crate instead. It does use `secret-service` but the backing [dbus-secret-service](https://github.com/brotskydotcom/dbus-secret-service/blob/main/src/ss.rs) crate only supports `org.freedesktop.secrets`. As such for now Euphonica needs this talk-name.

I'll probably just use `libsecret` in a future update & drop this permission, as since then the app has taken on more Linux-specific features anyway.

io.github.htkhiem.Euphonica.json
htkhiem: Thanks, just added with 2437f10.

io.github.htkhiem.Euphonica.json
bbhtt: is that far away future or something soon that we can wait on?

htkhiem:
It'll be ready this weekend at the soonest. I already have a rough draft
running but it needs more testing, for example whether the app can still
default to the keyring given by org.freedesktop.Secrets but via the
sandboxed portal. If not, users will have to re-enter their passwords.

htkhiem:
v0.97.0 now uses `libsecret`. The `--talk-name=org.freedesktop.secrets` permission has been dropped.

bot, build

PR #6913, opened 2025-09-10
Add de.thejackimonster.KnockKnock

TheJackiMonster:

- [x] Please describe the application briefly. *-> The application is intended to control doors or other simple devices via SSH trigger commands via graphical interface.*
- [x] Please attach a video showcasing the application on Linux using the Flatpak. https://github.com/user-attachments/assets/d0740d86-0b10-41eb-9fb3-8bbfaed74351
- [x] I am an _(please keep whichever is applicable and remove the rest)_ author/developer/upstream contributor to the project.

### DIFF
--- a/de.thejackimonster.KnockKnock.json
+++ b/de.thejackimonster.KnockKnock.json
@@ -5,6 +5,7 @@
 	"sdk": "org.gnome.Sdk",
 	"command": "knock-knock",
 	"finish-args": [
+		"--device=dri",
 		"--socket=wayland",
 		"--socket=fallback-x11",
 		"--share=ipc",
@@ -15,9 +16,6 @@
 		{
 			"name": "iputils",
 			"buildsystem": "meson",
-			"post-install": [
-				"install -Dm0644 LICENSE -t ${FLATPAK_DEST}/share/licenses/$FLATPAK_ID/iputils"
-			],
 			"sources": [
 				{
 					"type": "archive",


hfiguiere:
>  Please attach a video showcasing the application on Linux using the Flatpak. -> N/A, the application is only usable inside a network with a fitting door/device.

Your application is not special.

de.thejackimonster.KnockKnock.json
```diff
@@ -0,0 +1,38 @@
...
+	"runtime-version": "48",
+	"sdk": "org.gnome.Sdk",
+	"command": "knock-knock",
+	"finish-args": [
+		"--socket=wayland",
+		"--socket=fallback-x11",
+		"--share=network",
+		"--socket=ssh-auth",
+		"--filesystem=host-os",
+		"--env=PATH=/app/bin:/usr/bin:/run/host/usr/bin:/run/host/bin"
```
hfiguiere: that's a no. 

There is absolutely no guarantee that anything in host is executable in the sandbox. It works? Play the lottery.

TheJackiMonster: That's the workaround I found online for running a ping subprocess inside the sandbox. Since ping is usually available to the host but it's not available inside the sandbox. If there's a runtime which solves that issue, I would change that.

TheJackiMonster:
> > Please attach a video showcasing the application on Linux using the Flatpak. -> N/A, the application is only usable inside a network with a fitting door/device.
> 
> Your application is not special.

So I can either record a video showcasing the application running without a proper device in local network. That will make it only behave as dummy for the most part... or I record a video with some device in local network and it may contain details which should not get published (like IP address or identifiers for SSH connections), so I would likely need to censor them in the video.

de.thejackimonster.KnockKnock.json
TheJackiMonster: Okay, so I've added iputils as module which works as well.

TheJackiMonster:
I've attached a video now, I added a module so that the flatpak does not require access to the ping binary from the host and I added instructions to properly install the license files.

de.thejackimonster.KnockKnock.json
```diff
@@ -0,0 +1,59 @@
+{
+	"app-id": "de.thejackimonster.KnockKnock",
+	"runtime": "org.gnome.Platform",
+	"runtime-version": "48",
```
bbhtt:
```suggestion
	"runtime-version": "49",
```

de.thejackimonster.KnockKnock.json
```diff
@@ -0,0 +1,59 @@
...
+	"sdk": "org.gnome.Sdk",
+	"command": "knock-knock",
+	"finish-args": [
+		"--socket=wayland",
+		"--socket=fallback-x11",
```
bbhtt, line 10:
```suggestion
		"--socket=fallback-x11",
		"--share=ipc",
```

de.thejackimonster.KnockKnock.json
```diff
@@ -0,0 +1,59 @@
...
+				"./scripts/install.sh /app"
+			],
+			"post-install": [
+				"install -Dm0644 LICENSE -t ${FLATPAK_DEST}/share/licenses/$FLATPAK_ID/knock-knock"
+			],
+			"sources": [
+				{
+					"type": "archive",
+					"url": "https://gitlab.com/TheJackiMonster/knock-knock/-/archive/v1.0.2/knock-knock-v1.0.2.tar.gz",
+					"sha512": "40ec6a0b0ddc5536971a86e3d239be33b5613eabe36375704835a120d5f3e949f3172de05a1dcefafdb1284fc276c1955a6af2d33ad9767f47f5b208304e4722",
```
bbhtt: So the app here is this file? https://gitlab.com/TheJackiMonster/knock-knock/-/blob/main/door.py?ref_type=heads

de.thejackimonster.KnockKnock.json
```diff
@@ -0,0 +1,59 @@
+{
+	"app-id": "de.thejackimonster.KnockKnock",
+	"runtime": "org.gnome.Platform",
```
bbhtt, line 3: why does it need gnome runtime?

de.thejackimonster.KnockKnock.json
TheJackiMonster: Yes, pretty much.

de.thejackimonster.KnockKnock.json
TheJackiMonster: It's using GTK and libhandy widgets.

de.thejackimonster.KnockKnock.json
bbhtt: https://docs.flathub.org/docs/for-app-authors/requirements#minimal-submissions

de.thejackimonster.KnockKnock.json
bbhtt: GTK3 or 4? GTK 3 is in freedesktop sdk and libhandy is easy to build.

TheJackiMonster: But doesn't the GNOME runtime already ship libhandy and GTK3? Is it really better to use freedesktop and add additional dependencies manually to maintain than relying on generally available runtimes that would increase overall redundant usage of them decreasing additional space required for flatpaks on the host?

I pretty much see arguments for both ways here.

de.thejackimonster.KnockKnock.json
TheJackiMonster: But in that case it doesn't make much sense, does it? Only because I don't split multiple classes into different source code files, doesn't mean it has reduced complexity. It's no fork and I also created appdata files as well as icon. Even the UI is separated in files from Glade.

So why would this apply here?

TheJackiMonster:
So what? Is this really getting closed because I don't build libhandy manually? I don't see how this is such a big deal.

de.thejackimonster.KnockKnock.json
TheJackiMonster: I've updated the application to use GTK4 and libadwaita now instead of GTK3 and libhandy.

TheJackiMonster:
So I guess, I'll reopen a duplicate then. What the hell?

bbhtt:
It's not going to be accepted. I forgot to mention why it was blocked, but see https://docs.flathub.org/docs/for-app-authors/requirements#minimal-submissions

https://docs.flathub.org/docs/for-app-authors/requirements#software-design-and-distribution-policy

TheJackiMonster:
> It's not going to be accepted. I forgot to mention why it was blocked, but see https://docs.flathub.org/docs/for-app-authors/requirements#minimal-submissions

I still don't understand this point. So would it be accepted when I split my code into multiple classes and functions? It's not even just one script file anymore. It splits UI setup into separate design files using Cambalache and I even have some CSS changes. Also the application allows custom commands via SSH to signal a door remotely, handles errors and shows whether the remote is available automatically. If the user does not want to mess with terminal a lot, it's not a simple wrapper.

> https://docs.flathub.org/docs/for-app-authors/requirements#software-design-and-distribution-policy

Does this mean that I'm using SSH or that I build ping for it via iputils? Because after my committed changes to the manifest, I don't see how it requests overly permissive access anymore. Which is exactly why I requested review afterwards.

So the only thing I can think of which is still open would be the argument about building Adwaita manually rather than using the GNOME runtime. But that's not a requirement looking at the documentation, it seems but flavor and I would think both options have advantages or disadvantages.

PR #6963, opened 2025-09-21
Add org.indii.mendingwall

lawmurray:

Mending Wall fixes common issues when using more than one desktop environment (DE), like broken themes and cluttered menus. These come about because different DEs tweak GTK and Qt themes to make apps look more at home in the DE. Mending Wall keeps such tweaks isolated between DEs so that they do not interfere with each other. The [website](https://mendingwall.indii.org/) has a lot more detail on how it works exactly.

https://github.com/user-attachments/assets/aba84612-edfa-40e1-a644-da8b9ff5d172

### DIFF
--- a/org.indii.mendingwall.yaml
+++ b/org.indii.mendingwall.yaml
@@ -74,16 +74,9 @@ modules:
   - name: mendingwall
     buildsystem: meson
     config-opts:
-      - "-Dc_args=-DBUILD_FOR_FLATPAK"
+      - "-Dbuildfor=flatpak"
     sources:
       - type: git
         url: https://github.com/lawmurray/mendingwall.git
-        tag: v0.3.3
-    post-install:
-      # for autostart files, which will launch from outside flatpak, wrap
-      # calls to mendingwalld with flatpak run...
-      - sed -i -E "s,/app/bin/mendingwalld,flatpak run --command=/app/bin/mendingwalld org.indii.mendingwall,g" /app/share/mendingwall/org.indii.mendingwall.watch.desktop
-
-      # ...and calls to mendingwallcli
-      - sed -i -E "s,/app/bin/mendingwallcli,flatpak run --command=/app/bin/mendingwallcli org.indii.mendingwall,g" /app/share/mendingwall/org.indii.mendingwall.restore.desktop /app/share/mendingwall/org.indii.mendingwall.restore.sh
+        tag: v0.3.4
 
--- a/org.indii.mendingwall.yaml
+++ b/org.indii.mendingwall.yaml
@@ -1,0 +1,89 @@
+id: org.indii.mendingwall
+runtime: org.gnome.Platform
+runtime-version: '49'
+sdk: org.gnome.Sdk
+command: mendingwall
+
+finish-args:
+  - --share=ipc
+  - --device=dri
+  - --socket=wayland
+  - --socket=fallback-x11
+
+  # directories for writing autostart files; the background portal is not used
+  # as it does not support the earlier init phases needed to restore themes
+  # before first use, specifically the X-GNOME-Autostart-Phase key and, for KDE,
+  # the ~/.config/plasma-workspace/env directory
+  - --filesystem=xdg-config/autostart:create
+  - --filesystem=xdg-config/plasma-workspace/env:create
+
+  # dconf backend for GSettings, so that can save and restore configuration of
+  # desktop environments
+  - --filesystem=xdg-run/dconf
+  - --filesystem=xdg-config/dconf:rw
+  - --talk-name=ca.desrt.dconf
+  - --env=GIO_EXTRA_MODULES=/app/lib/gio/modules/
+  - --env=GSETTINGS_BACKEND=dconf
+
+  # config files, so that can save and restore configuration of desktop
+  # environments
+  - --filesystem=xdg-config:rw
+  - --filesystem=xdg-config/gtk-3.0:create
+  - --filesystem=xdg-config/gtk-4.0:create
+  - --filesystem=xdg-config/xsettingsd:create
+
+  # application directories, to find .desktop files of installed applications
+  - --filesystem=xdg-data/flatpak/exports/share/applications:ro
+  - --filesystem=/var/lib/flatpak/exports/share/applications:ro
+  - --filesystem=/var/lib/snapd/desktop/applications:ro
+  - --filesystem=host-os:ro
+
+  # user application directory, to write updated .desktop files of installed
+  # applications
+  - --filesystem=xdg-data/applications:create
+
+cleanup:
+  - /include
+  - /lib/pkgconfig
+  - /man
+  - /share/doc
+  - /share/gtk-doc
+  - /share/man
+  - /share/pkgconfig
+  - '*.la'
+  - '*.a'
+
+modules:
+
+  - name: dconf
+    buildsystem: meson
+    config-opts:
+      - -Dbash_completion=false
+      - -Dman=false
+      - -Dsystemduserunitdir=' '
+    cleanup:
+      - /include
+      - /lib/pkgconfig
+      - /libexec
+      - /share/dbus-1
+    sources:
+      - type: git
+        url: https://gitlab.gnome.org/GNOME/dconf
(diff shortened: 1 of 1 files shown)


org.indii.mendingwall.yaml
```diff
@@ -0,0 +1,89 @@
...
+    config-opts:
+      - "-Dc_args=-DBUILD_FOR_FLATPAK"
+    sources:
+      - type: git
+        url: https://github.com/lawmurray/mendingwall.git
+        tag: v0.3.3
+    post-install:
+      # for autostart files, which will launch from outside flatpak, wrap
+      # calls to mendingwalld with flatpak run...
+      - sed -i -E "s,/app/bin/mendingwalld,flatpak run --command=/app/bin/mendingwalld org.indii.mendingwall,g" /app/share/mendingwall/org.indii.mendingwall.watch.desktop
```
hfiguiere: 1. DO NOT use sed to modify file. `desktop-file-edit` exist for .desktop files.

org.indii.mendingwall.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+      # ...and calls to mendingwallcli
+      - sed -i -E "s,/app/bin/mendingwallcli,flatpak run --command=/app/bin/mendingwallcli org.indii.mendingwall,g" /app/share/mendingwall/org.indii.mendingwall.restore.desktop /app/share/mendingwall/org.indii.mendingwall.restore.sh
```
hfiguiere: 1. DO NOT use sed to modify file. `desktop-file-edit` exist for .desktop files. For the rest, either upstream is fixed to behave properly or a patch is used.
2. .desktop files NEVER have any mention of flatpak.

org.indii.mendingwall.yaml
```diff
@@ -0,0 +1,89 @@
...
+  - --env=GSETTINGS_BACKEND=dconf
+
+  # config files, so that can save and restore configuration of desktop
+  # environments
+  - --filesystem=xdg-config:rw
+  - --filesystem=xdg-config/gtk-3.0:create
+  - --filesystem=xdg-config/gtk-4.0:create
+  - --filesystem=xdg-config/xsettingsd:create
+
+  # application directories, to find .desktop files of installed applications
```
hfiguiere, line 35: you assume so much about where flatpak is that it doesn't even feel like it's reliable nor properly tested.

org.indii.mendingwall.yaml
```diff
@@ -0,0 +1,89 @@
...
+  - --filesystem=xdg-data/flatpak/exports/share/applications:ro
+  - --filesystem=/var/lib/flatpak/exports/share/applications:ro
+  - --filesystem=/var/lib/snapd/desktop/applications:ro
+  - --filesystem=host-os:ro
+
+  # user application directory, to write updated .desktop files of installed
+  # applications
+  - --filesystem=xdg-data/applications:create
```
hfiguiere, line 43: there is a portal to install desktop files.

org.indii.mendingwall.yaml
lawmurray: Thanks. I've now pushed this upstream, and `post-install` is removed entirely.

org.indii.mendingwall.yaml
lawmurray: Thanks. I've now pushed this upstream, and `post-install` is removed entirely.

lawmurray:
> The domain to be used for verification is indii.org. If you intend to verify this app, please confirm that you can upload https://indii.org/.well-known/org.flathub.VerifiedApps.txt. Otherwise, ignore this. Please comment if this incorrect.

Confirming that I can. I've uploaded an empty file there for now to demonstrate.

org.indii.mendingwall.yaml
lawmurray: Yes, thanks for the heads up. The issue is that, as I understand it (using libportal):

1. [dynamic_launcher_install](https://libportal.org/method.Portal.dynamic_launcher_install.html) is for desktop files that start with the caller's app id.
2. [dynamic_launcher_request_install_token](https://libportal.org/method.Portal.dynamic_launcher_request_install_token.html) is for any desktop files, but requires the caller's app id to be in an allowlist (which Mending Wall is not).

Whereas here, the idea is to create desktop files for already-installed applications that override the `OnlyShowIn` or `NotShowIn` keys to organise them between desktop environments ([more info](https://mendingwall.indii.org/tidy-menus/)). I think this is outside the scope of the portal, and so `--filesystem=xdg-data/applications:create` is being requested to create the files directly. By way of comparison, the Flatpak manifests of [Libre Menu Editor](https://codeberg.org/libre-menu-editor/libre-menu-editor/src/branch/main/flatpak.yml) and [AppEditor](https://github.com/donadigo/appeditor/blob/master/com.github.donadigo.appeditor.json) request the same permission to edit application menus.

org.indii.mendingwall.yaml
lawmurray: It's not clear to me what you're referring to here. Do you mean the lines that follow with specific `applications` directories?

If so, let me explain the reasoning.

Outside of Flatpak, Mending Wall uses the `$XDG_DATA_DIRS` environment variable to determine where desktop files are installed on the host. Flatpak, of course, modifies `$XDG_DATA_DIRS` for the sandbox, and these are lost.

1. One option for the Flatpak is to basically hardcode the usual directories (`/usr/share/applications`, `/usr/local/share/applications`, etc). That's basically what I've done for the Flatpak build. I've tested that to work on Arch, Fedora, openSUSE, Debian and Ubuntu. These directories are quite standard after all. In the spirit of requesting only what is necessary I've enumerated those specific directories for the Flatpak permissions. [Libre Menu Editor](https://codeberg.org/libre-menu-editor/libre-menu-editor/src/branch/main/flatpak.yml) does similar, but is less specific.

2. A second option, which I'm very much trying to avoid, is requesting arbitrary permissions to escape the sandbox and get at the host's `$XDG_DATA_DIRS`.

The implication of omitting a directory that appears in the host's `$XDG_DATA_DIRS` is just that some desktop files will not be discovered, and not automatically adjusted with `OnlyShowIn` and `NotShowIn`. The user can still adjust them manually if they want. I would prefer that than broader sandbox permissions.

lawmurray:
I thought I'd make some longer videos to demonstrate what Mending Wall actually achieves.

### Before

1. We're on openSUSE Tumbleweed, with both GNOME and KDE installed. We're currently in GNOME. Notice how there are many KDE core applications in the menu. Also notice the default Adwaita icon theme in Files, and the default Adwaita cursor theme.
2. 0:24. We log out and log in to KDE. Notice how GNOME core applications appear in KDE, even though KDE has its own (not a problem per se, but it can be distracting).
3. 1:14. We log out and log in to GNOME again. Notice how the icon theme and cursor have changed from Adwaita to Breeze (the KDE default). Initially, there seems a be a cursor scaling issue too, although it is corrected after a click.

https://github.com/user-attachments/assets/ba584a67-ce72-4008-81cd-394b37225709

### After

1. We're in GNOME again. Notice, again, KDE core applications in the menu. We start up Mending Wall and enable its "Tidy Menus" feature first. Looking at the menus again, KDE core applications have been hidden.
2. 0:33. We go back to Mending Wall and enable its "Mend Themes" feature. Before logging out, we note that it's running in the background.
3. 0:54. Logging into KDE, GNOME core applications are also hidden from the menu.
4. 1:50. Logging out of KDE and back into GNOME, this time the cursor and icon theme are Adwaita, as we left them, not switched to Breeze as in the before video.

https://github.com/user-attachments/assets/8ea42e50-6abc-4480-91ca-4fd026d20ae2

I hope that helps in understanding what this is all about.

lawmurray:
bot, build

bbhtt:
You have to run the linter and ask for [exceptions](https://docs.flathub.org/docs/for-app-authors/linter#exceptions) for all of these.

lawmurray:
> You have to run the linter and ask for [exceptions](https://docs.flathub.org/docs/for-app-authors/linter#exceptions) for all of these.

Thanks @bbhtt, I'll do that.

lawmurray:
bot, build

PR #6972, opened 2025-09-24
Add io.github.bardiakz.Qrypt

bardiakz:

- [x] Please describe the application briefly. < Qrypt is a Flutter-based application for text encryption, compression, and obfuscation, featuring quantum-safe cryptography (ML-KEM, ML-DSA), multiple encryption methods (AES-CBC, AES-CTR, AES-GCM, RSA), and compression algorithms (GZip, LZ4, Brotli, Zstd). It enables secure message sharing across platforms like SMS, chat, or email, with a focus on user control and privacy. >
- [ ] Please attach a video showcasing the application on Linux using the Flatpak. < Please insert the video here >
- [x] I am a developer contributor to the project.


bardiakz:
bot, build

io.github.bardiakz.Qrypt.yaml
```diff
@@ -0,0 +1,113 @@
...
+      - mv git_repo/data/${FLATPAK_ID}.desktop ${FLATPAK_ID}.desktop
+      - desktop-file-edit ${FLATPAK_ID}.desktop --set-icon=${FLATPAK_ID}
+      - install -Dm644 -t "${FLATPAK_DEST}/share/applications/" ${FLATPAK_ID}.desktop
+      - install -Dm644 git_repo/assets/icon/qrypt_icon.png "${FLATPAK_DEST}/share/icons/hicolor/512x512/apps/${FLATPAK_ID}.png"
+      - install -Dm644 git_repo/data/${FLATPAK_ID}.metainfo.xml "${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.metainfo.xml"
+    sources:
+      - type: archive
+        only-arches: [x86_64]
+        dest: Qrypt
+        url: https://github.com/bardiakz/qrypt/releases/download/v4.4.9/qrypt-linux-x86_64.tar.gz
```
bbhtt, line 53: https://docs.flathub.org/docs/for-app-authors/requirements#building-from-source

io.github.bardiakz.Qrypt.yaml
```diff
@@ -0,0 +1,113 @@
...
+    buildsystem: cmake
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DCMAKE_INSTALL_PREFIX=/app
+      - -DBUILD_SHARED_LIBS=ON
+    sources:
+      - type: git
+        url: https://github.com/google/brotli.git
+        tag: v1.1.0
+  - name: zstd
```
bbhtt: these are in the runtime

io.github.bardiakz.Qrypt.yaml
```diff
@@ -0,0 +1,113 @@
...
+  - name: lz4
+    buildsystem: simple
+    build-commands:
+      - make -j$FLATPAK_BUILDER_N_JOBS
+      - make install PREFIX=/app
+    sources:
+      - type: git
+        url: https://github.com/lz4/lz4.git
+        tag: v1.10.0 # Latest stable
+  - name: brotli
```
bbhtt: ditto

io.github.bardiakz.Qrypt.yaml
```diff
@@ -0,0 +1,113 @@
...
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DOQS_BUILD_ONLY_LIB=ON
+      - -DCMAKE_INSTALL_PREFIX=/app
+      - -DBUILD_SHARED_LIBS=ON
+    sources:
+      - type: git
+        url: https://github.com/open-quantum-safe/liboqs.git
+        tag: 0.14.0
+  - name: lz4
```
bbhtt: ditto

io.github.bardiakz.Qrypt.yaml
```diff
@@ -0,0 +1,113 @@
...
+  - "*.la"
+  - /git_repo
+  - /io.github.bardiakz.Qrypt.desktop
+modules:
+  - name: liboqs
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DOQS_BUILD_ONLY_LIB=ON
+      - -DCMAKE_INSTALL_PREFIX=/app
```
hfiguiere:
```suggestion
```

This is never needed.

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64"]
```
hfiguiere, line 2: disappointing

io.github.bardiakz.Qrypt.yaml
bardiakz: thanks for the quick review
I submitted with pre built binaries because the app relies on a .env file for some default encryptions and obfuscation maps 
also flutter itself adds complexity, dependencies from pub.dev need to be bundled offline

bbhtt: you can look at https://github.com/TheAppgineer/flatpak-flutter

bardiakz: flatpak-flutter handles the flutter dependencies properly but the core issue remains, the app requires a .env file containing secret keys. I cant commit this .env to the public repo without exposing it which causes the build to fail.
is there a recommended way to handle this on flathub? or could an exception be made here?

bbhtt: secrets for what?

bardiakz: Some encryption method variants(like aes) use a static key by default (there is support for custom key but thats the default) and this can't be handled on server side since the whole point of this app is that it works with no internet

hfiguiere: So you are saying the your "encryption" is useless because you have a static key hidden in the binary for a symmetric algorithm.

bardiakz: It's just the default option for non technical users, custom keys are also supported

hfiguiere: Then we are gonna have to hold back this until you fix it. I you don't understand the basic of cryptography then maybe it is our responsibility to not let you ship this through flathub. Because it is reckless.

bardiakz: I understand your concern but in order for this to make sense i should explain why this app exists in the first place.

imagine this scenario, internet is blocked and your only ways of communication are a heavily monitored messenger designed by the government or sms which is also monitored 
now imagine you want to send a non sensitive data (in this example an xray vpn config) to your friend without being detected by regex or any other surveillance methods, in this case the default encryption with obfuscation would be enough
now if a more technical user wants to transfer a more sensitive data they can manually use one of the safe key transfer methods and use that custom key for aes

-the static key is not forced its just an option 

also hmac keys are stored in the .env too

hfiguiere: I read it as: 

This app exist to give a false sense of security and in the context you give endanger users. Extracting the "secret" for the symmetric encryption is trivial.

> -the static key is not forced its just an option

It should never even be considered as an option. This is just reckless.

bardiakz: I just gave you the example, adding a layer of aes even with a static key is arguably better than just plain obfuscation where its not meant to handle sensitive data
and false sense of security?? nowhere in the app or docs do I suggest or recommend using the static key option

that said, I will definitely add warnings in the ui to highlight the risks. thanks for pointing that out, a non technical user might indeed use it for sensitive 
but still removing the option entirely and not giving the choice doesnt seem like the best course. why would you bother manually exchanging keys with kem just to send a vless config?

PR #6975, opened 2025-09-25
Add net.pixieditor.PixiEditor

flabbet:

- [x] Please describe the application briefly.  PixiEditor is a node-based, universal 2D graphics editor
[Screencast From 2025-09-25 16-26-31.webm](https://github.com/user-attachments/assets/3a71bb66-153e-4604-bd7a-6509cb1793d7)
- [x] I am an author

### DIFF
--- a/flatpak-dotnet-generator.py
+++ b/flatpak-dotnet-generator.py
@@ -13,8 +13,7 @@ from pathlib import Path
 
 
 def main() -> None:
-    # Bump this to latest freedesktop runtime version.
-    freedesktop_default = "24.08"
+    freedesktop_default = "25.08"
     # Bump this to an LTS dotnet version.
     dotnet_default = "8"
 
--- a/net.pixieditor.PixiEditor.yml
+++ b/net.pixieditor.PixiEditor.yml
@@ -1,6 +1,6 @@
 id: net.pixieditor.PixiEditor
 runtime: org.freedesktop.Platform
-runtime-version: '24.08'
+runtime-version: '25.08'
 sdk: org.freedesktop.Sdk
 sdk-extensions:
   - org.freedesktop.Sdk.Extension.dotnet8
@@ -30,7 +30,7 @@ modules:
     build-commands:
     - /usr/lib/sdk/dotnet8/bin/install.sh
 
-  - name: binary
+  - name: PixiEditor binary
     buildsystem: simple
     sources:
       - output-sources.json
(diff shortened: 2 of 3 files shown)


net.pixieditor.PixiEditor.yml
```diff
@@ -0,0 +1,58 @@
...
+  - name: binary
+    buildsystem: simple
+    sources:
+      - output-sources.json
+      - type: file
+        path: appsettings.json
+      - type: file
+        path: pixieditor.sh
+      - type: archive
+        url: https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-27/wasi-sdk-27.0-x86_64-linux.tar.gz
```
hfiguiere, line 42: What happen on aarch64?

flatpak-dotnet-generator.py
```diff
@@ -0,0 +1,121 @@
...
+import concurrent.futures
+import json
+import subprocess
+import tempfile
+from pathlib import Path
+
+
+def main() -> None:
+    # Bump this to latest freedesktop runtime version.
+    freedesktop_default = "24.08"
```
hfiguiere: turns out latest is 25.08

net.pixieditor.PixiEditor.yml
```diff
@@ -0,0 +1,58 @@
...
+  - --filesystem=xdg-desktop
+  - --env=DOTNET_ROOT=/app/lib/dotnet
+
+modules:
+  - name: dotnet
+    buildsystem: simple
+    build-commands:
+    - /usr/lib/sdk/dotnet8/bin/install.sh
+
+  - name: binary
```
hfiguiere: use an explicit name.

net.pixieditor.PixiEditor.yml
flabbet: Ah right, fair point. Base app _should_ compile for arm (with appropriate changes in the manifest) but we don't officially support it yet so I suppose I should add 
```
{
  "skip-arches": ["aarch64"]
}
```

in that case?

hfiguiere: why not just building for aarch64 instead?

flabbet: It doesn't compile on my devices, we didn't publish arm builds for linux in other places (standalone, steam) and it's untested. There are bunch of preprocessors for arch in the codebase and even though I did try to account for ARM, I can't guarantee it will work.

We will eventually officially support arm for linux, but not now

PR #7037, opened 2025-10-14
Add org.kde.discover

renner0e:
- [X] Please describe the application briefly. Install or upgrade flatpaks and manage firmware upgrades

https://github.com/user-attachments/assets/2497f50b-6867-42c3-a22a-1c81e0a9acde

 
- [X] I contacted upstream developers about this submission.
I have contacted @travier, not everything works right now, we will iterate. This is why this is a draft.

linter exceptions: https://github.com/flathub-infra/flatpak-builder-lint/pull/736

@travier

### DIFF
--- a/org.kde.discover.yaml
+++ b/org.kde.discover.yaml
@@ -69,6 +69,9 @@ modules:
       # Warn about missing functionality
       - type: patch
         path: metainfo-warning.patch
+      # https://invent.kde.org/plasma/discover/-/merge_requests/1191
+      - type: patch
+        path: appdata-1191.patch
 
       - type: patch
         path: fix-appstream-id-mismatch.patch


travier:
CC @flathub/kde

org.kde.discover.yaml
```diff
@@ -0,0 +1,342 @@
...
+
+
+cleanup-commands:
+  # https://github.com/flathub/io.qt.qtwebengine.BaseApp?tab=readme-ov-file#cleanup
+  - /app/cleanup-BaseApp.sh
+  # is this missing from the script?
+  # TODO: upstream this
+  - rm -rfv ${FLATPAK_DEST}/lib/${FLATPAK_ARCH}-linux-gnu/pkgconfig
+
+  - rm -rf /app/lib/python*/
```
bbhtt: is this correct? it does not depend on anything python related? or what python modules are you cleaning up? they should probably be in the module level cleanup.

org.kde.discover.yaml
```diff
@@ -0,0 +1,342 @@
...
+  # qtwebengine
+  - /resources/*
+
+
+cleanup-commands:
+  # https://github.com/flathub/io.qt.qtwebengine.BaseApp?tab=readme-ov-file#cleanup
+  - /app/cleanup-BaseApp.sh
+  # is this missing from the script?
+  # TODO: upstream this
+  - rm -rfv ${FLATPAK_DEST}/lib/${FLATPAK_ARCH}-linux-gnu/pkgconfig
```
bbhtt: this should be in the baseapp cleanup instead indeed, you can open a pr there.

org.kde.discover.yaml
```diff
@@ -0,0 +1,342 @@
...
+    # needed for installing flatpaks
+    - --talk-name=org.freedesktop.Flatpak
+
+cleanup:
+  - /include
+  - /lib/pkgconfig
+  - /share/man
+  - /share/licenses
+  - "*.pc"
+  - "*.debug"
```
bbhtt: why?

org.kde.discover.yaml
```diff
@@ -0,0 +1,342 @@
...
+    # fwupd integration
+    - --system-talk-name=org.freedesktop.fwupd
+    # needed for installing flatpaks
+    - --talk-name=org.freedesktop.Flatpak
+
+cleanup:
+  - /include
+  - /lib/pkgconfig
+  - /share/man
+  - /share/licenses
```
bbhtt:
```suggestion
```

org.kde.discover.yaml
```diff
@@ -0,0 +1,342 @@
...
+            commit: 8054c66e0454a09a810f756996d1b280738594e5
+            tag: "v0.9"
+        x-checker-data:
+          type: git
+          tag-pattern: "v^([\\d.]+)$"
+
+    - name: fwupd
+      buildsystem: meson
+      config-opts:
+        - --libdir=/app/lib
```
bbhtt:
```suggestion
```

org.kde.discover.yaml
```diff
@@ -0,0 +1,342 @@
...
+        - -Dsystemd=disabled
+        - -Dpassim=disabled
+        - -Ddocs=disabled
+        - -Dtests=false
+        - -Dvendor_ids_dir=/app/share/hwdata
+        - -Dvalgrind=disabled
+        - -Dintrospection=disabled
+      build-options:
+        env:
+          PYTHONPATH: /app/utils/lib/python3.12/site-packages:/app/lib/python3.12/site-packages:/usr/lib/python3.12/site-packages
```
bbhtt: this will break on runtime updates.

bbhtt:
ah i didn't notice this was a draft.

anyway how working is this? doesn't discover have integration with native distro package managers etc.? Do those work?

renner0e:
> ah i didn't notice this was a draft.
> 
> anyway how working is this? doesn't discover have integration with native distro package managers etc.? Do those work?

I didn't try. I also think this is out of scope for a flatpak that is available everywhere. Tho that's just my opinion.

bbhtt:
if it is exposed in the UI people will try to use it and will be unhappy when it doesn't work. it also creates a difference between a system copy of discover and the flatpak copy

renner0e:
> if it is exposed in the UI people will try to use it and will be unhappy when it doesn't work. it also creates a difference between a system copy of discover and the flatpak copy

No the packagekit backend is not even compiled and won't be exposed to users at all. I put a little readme that shows what is not working.
I couldn't get the KNS backend to work and I think that's also kinda out of scope for something that could be used on other desktops.

Bazzite (when it used to ship discover) and Aurora have actually been shipping discover in this "neutered" form with only flatpak and fwupd support.
The drop-down menu on the install button will only show stuff that is actually available.

org.kde.discover.yaml
renner0e: https://github.com/flathub/io.qt.qtwebengine.BaseApp/pull/387
I hope this is the right place

Justinzobel:
I'm not sure that I like the idea of this. You'd have to test it with every distro to make sure package updates work, then you have to test with things like KDE Linux which have their own backend, and all sorts of things that can and will go wrong.

Discover (like Dolphin and Konsole) to me belong on the system, not as a Flatpak as they're too tightly intertwined with system functions.

org.kde.discover.yaml
renner0e: [`157d844` (#7037)](https://github.com/flathub/flathub/pull/7037/commits/157d844ef08e2cd0a3b54b46ec69ece15ac90d49)

hfiguiere:
I don't think this is suitable for flathub at all.

travier:
> I'm not sure that I like the idea of this. You'd have to test it with every distro to make sure package updates work, then you have to test with things like KDE Linux which have their own backend, and all sorts of things that can and will go wrong.

This would be to manage Flatpaks only, not packages or anything else on the system.

> Discover (like Dolphin and Konsole) to me belong on the system, not as a Flatpak as they're too tightly intertwined with system functions.

Yes, but we also have Dolphin and Konsole here on Flathub, as they are still useful even if they are incomplete.

travier:
The target for this Flatpak isn't to replace the Discover app already installed on some systems. It's for other systems that don't use Discover and where users might want to use it to manage their Flatpaks.

renner0e:
> The target for this Flatpak isn't to replace the Discover app already installed on some systems. It's for other systems that don't use Discover and where users might want to use it to manage their Flatpaks.

Yes exactly, I should've made my intentions clearer about this from the start.

I'm also not dead set on including the fwupd backend, if discover flatpak should only do flatpaks then I'm fine with that. But it should be appropriate considering gnome Firmware is on flathub as well.

Pointedstick:
> The target for this Flatpak isn't to replace the Discover app already installed on some systems.

Actually in KDE Linux, there's a good chance we do want to do that. Our current desire is to have all user-facing apps come from Flatpak, once their UX is good enough.

"…once their UX is good enough". If multiple backends don't work with the Flatpak version of this app, I'm not sure it's a good idea, because, yeah, to echo what others have said, if we make it available here, it has to work for everyone, and it has to work as well as the native-packaged version.

bbhtt:
IMO at best we can do Flathub beta. 

If Discover's only job was to manage Flatpak like Bazaar or Easyflatpak, then it'd make sense and at least that way, there is no difference between the native package and the Flatpak one, but that's not what happens in practice. 

A Flatpak-only Discover seems only useful for image based distros who use Flatpak for all their app needs?

Otherwise, on usual distros there's doesn't seem to be a need for a native discover and then a flatpak one on top? (Why would someone prefer the Flatpak copy in that case if their system supplied one already manages Flatpaks + extras)

org.kde.discover.yaml
```diff
@@ -0,0 +1,337 @@
...
+  # https://github.com/flathub/io.qt.qtwebengine.BaseApp?tab=readme-ov-file#cleanup
+  - /app/cleanup-BaseApp.sh
+  # is this missing from the script?
+  # TODO: upstream this
+  - rm -rfv ${FLATPAK_DEST}/lib/${FLATPAK_ARCH}-linux-gnu/pkgconfig
+
+  - rm -rf /app/lib/python*/
+
+modules:
+  - name: discover
```
bbhtt, line 43: You should add a note to the top of the metainfo that this can ONLY manage Flatpaks. Patch it I guess.

https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines#website-specific-formatting

org.kde.discover.yaml
bbhtt: Yes that's correct, I merged it. It should be available in 4-5 hours, then you can drop this.

Pointedstick:
> A Flatpak-only Discover seems only useful for image based distros who use Flatpak for all their app needs?

Several of these use Discover for system updates too. Kinoite, SteamOS, and KDE Linux spring to mind. Discover has dedicated backends for their system updating functionality.

travier:
> > The target for this Flatpak isn't to replace the Discover app already installed on some systems.
> 
> Actually in KDE Linux, there's a good chance we do want to do that. Our current desire is to have all user-facing apps come from Flatpak, once their UX is good enough.

Yes, in the end, but not right now. It's the same trade off that we make for Dolphin/Konsole/Filelight. They need significant work to fully function as Flatpaks so we bundle them in Image Based systems. But we still have them available as Flatpaks on Flathub even thought there are limitations.

But on Image Mode systems, the only way to get apps is via Flatpaks (or Snaps, etc.), so allowing the user to remove the primary interface used to install apps is a serious foot gun, so I would not recommend replacing a system Discover with this Flatpak until we implement some form of "recovery option".

> "…once their UX is good enough". If multiple backends don't work with the Flatpak version of this app, I'm not sure it's a good idea, because, yeah, to echo what others have said, if we make it available here, it has to work for everyone, and it has to work as well as the native-packaged version.

We don't have this high bar for other apps (Dolphin/Konsole/Filelight). Why would we treat Discover differently?

> Several of these use Discover for system updates too. Kinoite, SteamOS, and KDE Linux spring to mind. Discover has dedicated backends for their system updating functionality.

We might be able to make most of the system update functionality work with the Flatpak but the notification, KCM and automatic update parts won't work as Flatpak so they will have to remain on the host system for now.

> Otherwise, on usual distros there's doesn't seem to be a need for a native discover and then a flatpak one on top? (Why would someone prefer the Flatpak copy in that case if their system supplied one already manages Flatpaks + extras)

Indeed, using both the system packaged Discover with the Flatpak one makes no sense. You would use the Flatpak on a distribution that uses something else to update the system, where that something else may not support updating/managing Flatpaks.

I'm +1 with this being beta only for now and I agree that we need to make that clear in the README & the description.

Pointedstick:
> allowing the user to remove the primary interface used to install apps is a serious foot gun

For sure. You'd need to pin it or make it not uninstallable or something.

> We don't have this high bar for other apps (Dolphin/Konsole/Filelight). Why would we treat Discover differently?

Honestly, maybe we should have the same high bar.

I think Konsole and Filelight get over the bar, but IMO Dolphin definitely does not. Just look at https://invent.kde.org/kde-linux/kde-linux/-/issues/71#dolphin.

Were it up to me, I'd remove Dolphin from Flathub as I don't think it can shine as a product here in its current state.

org.kde.discover.yaml
renner0e: yeah it's only used for building, runs fine without it
[`3fc6d29` (#7037)](https://github.com/flathub/flathub/pull/7037/commits/3fc6d2905632dc59fb229ff5a23f4987f8aa3cae)

org.kde.discover.yaml
renner0e: Can you please elaborate what you mean by that? I'm not sure what is wrong with that.

hfiguiere:
```suggestion
```

Justinzobel:
> Yes, but we also have Dolphin and Konsole here on Flathub, as they are still useful even if they are incomplete.

I disagree. I think we should remove them. Incomplete or part-functioning apps do not belong on user's devices.

Justinzobel:
If this does get accepted, it must be beta only. Then we can refer anyone that complains about it to the README, which would indicate in big writing that it is severely limited.

org.kde.discover.yaml
```diff
@@ -0,0 +1,332 @@
...
+runtime-version: '6.9'
+sdk: org.kde.Sdk
+command: plasma-discover
+base: io.qt.qtwebengine.BaseApp
+base-version: '6.9'
+finish-args:
+    # access to system flatpaks, nothing works without it
+    - --filesystem=/var/lib/flatpak
+    # access to user flatpaks
+    - --filesystem=xdg-data/flatpak
```
JakobDev, line 11: How are Flatpaks handled that are nor in the default locations?

org.kde.discover.yaml
```diff
@@ -0,0 +1,332 @@
...
+          url: https://github.com/flatpak/flatpak/
+          commit: b676905d91aa3c4e524ca663784d4cf085c465d7
+          tag: 1.16.1
+          x-checker-data:
+            type: git
+            tag-pattern: "^([\\d.]+)$"
+
+      modules:
+         - shared-modules/intltool/intltool-0.51.json
+         - name: polkit
```
JakobDev, line 219: If you [disable the system helper](https://github.com/flatpak/flatpak/blob/bd468a89596e6f89a364b322e42a70948672f4e5/meson.build#L189) polkit is no loner needed

org.kde.discover.yaml
```diff
@@ -0,0 +1,332 @@
...
+               commit: 82185d9aaf0a79b100378f13a54c3d5d3763a971
+               x-checker-data:
+                 type: anitya
+                 project-id: 10899
+                 tag-template: v$version
+
+           modules:
+             - pyparsing.yaml
+
+         - name: bubblewrap
```
JakobDev, line 309: I wonder if this is actually needed. As far as I know bubblewrap is only needed to run

org.kde.discover.yaml
```diff
@@ -0,0 +1,332 @@
...
+           sources:
+             - type: archive
+               url: https://github.com/containers/bubblewrap/archive/refs/tags/v0.11.0.tar.gz
+               sha256: cfeeb15fcc47d177d195f06fdf0847e93ee3aa6bf46f6ac0a141fa142759e2c3
+               x-checker-data:
+                 type: anitya
+                 project-id: 10937
+                 url-template: https://github.com/containers/bubblewrap/archive/refs/tags/v$version.tar.gz
+
+         - name: xdg-dbus-proxy
```
JakobDev, line 328: Same as bubblewrap

JakobDev (commented):
I wonder if we can make PackageKit work. As far as I know (I'm not a expert) PackageKit is using D-Bus, so it might work if we add `--system-talk-name=org.freedesktop.PackageKit` and it's installed on the host.

org.kde.discover.yaml
renner0e: you mean those? yeah they don't work. Even if you give it permission to the host-etc and the path to the installation
https://docs.flatpak.org/en/latest/tips-and-tricks.html#adding-a-custom-installation

See:
[`60d6dc2` (#7037)](https://github.com/flathub/flathub/pull/7037/commits/60d6dc26cc7a66acc2d2893798fdd22474958f21)

org.kde.discover.yaml
renner0e: we need it
See:
[`745b2da` (#7037)](https://github.com/flathub/flathub/pull/7037/commits/745b2da4b0a6aa47d42dc5d98f4db95eb34df254)

org.kde.discover.yaml
renner0e: If I clean it up completely it still works, not sure if there would ever be a need for it but it doesn't work anyway because I get unprivileged namespace errors when I `flatpak run` inside the sandbox.
[`f45f261` (#7037)](https://github.com/flathub/flathub/pull/7037/commits/f45f261403443896c71de68808908154bdff57f2)

org.kde.discover.yaml
renner0e: [`987c8bb` (#7037)](https://github.com/flathub/flathub/pull/7037/commits/987c8bbe6ec1d46d270d541996461cc2b6c33a09)

org.kde.discover.yaml
JakobDev: I see what the problem is. Polkit is not needed by Flatpak itself (outside of a [fallback](https://github.com/flatpak/flatpak/blob/bd468a89596e6f89a364b322e42a70948672f4e5/app/flatpak-main.c#L621) which doesn't apply here) but by `flatpak-system-helper` which is a separate binary which runs with root permission. Flatpak talks to it using D-Bus. This binary together with polkit is not needed here and it's not really working. Flatpak is using the system helper of your system here.

If you build with disabled system helper, Flatpak won't use the existing system helper of your system. I can see the need of an additional build option here. I try to make a patch when I have time.

org.kde.discover.yaml
JakobDev: You can't run a Flatpak inside a Flatpak, so no need for this

org.kde.discover.yaml
JakobDev: This could be a problem. [I'm working on improving custom installations](https://github.com/flatpak/flatpak/pull/6343). I don't know when and in which form it will land but this problem could get bigger in the future.

A potential solution could be [the new host-root permisson](https://github.com/flatpak/flatpak/pull/6314). Another more sandbox friendly solution could be flatpak/xdg-desktop-portal#1737, but it's not going to be accepted.

org.kde.discover.yaml
JakobDev: Do you even need to build it?

renner0e: > Do you even need to build it?

yeah it's needed for flatpak to build, doesn't seem like you can turn it off

JakobDev: [It looks like you could just use any other executable that provides a --version](https://github.com/flatpak/flatpak/blob/bd468a89596e6f89a364b322e42a70948672f4e5/meson.build#L269). So maybe a shell script which just echo the latest version could be sued instead to save some build time.

org.kde.discover.yaml
bbhtt: we don't need to worry about non standard setups that < 1% of the people use. Your Flatpak PR is not relevant until it reaches a stable version

org.kde.discover.yaml
bbhtt: no keep it as is.

org.kde.discover.yaml
bbhtt: keep as is.

org.kde.discover.yaml
tulilirockz: This is not relevant, yeah

org.kde.discover.yaml
renner0e: [`b862f25` (#7037)](https://github.com/flathub/flathub/pull/7037/commits/b862f254104a79ce8aeef19b0b5fbc20e9e5dfea)

renner0e:
I removed qtwebengine in [`54a8b55` (#7037)](https://github.com/flathub/flathub/pull/7037/commits/54a8b55d80d3667812f1daca9c8eeecfdaa65b1c), we only need this as an optional build dependency.

As 99.9% of users won't ever authenticate to a flatpak repository (handled via browser AFAIU), but it will reduce it's size tremendously

README.md
```diff
@@ -0,0 +1,29 @@
...
+https://invent.kde.org/plasma/discover
+
+
+## Patches
+
+stolen from [gnome builder](https://github.com/flathub/org.gnome.Builder) and [bazaar](https://github.com/flathub/io.github.kolunmi.Bazaar)
+
+# broken stuff
+- launch button doesn't do anything
+- ~- no --user flatpaks~
```
bbhtt: why does this not work?

README.md
```diff
@@ -0,0 +1,29 @@
...
+- can't install .flatpak bundle files (needs permission to the path, inside the sandbox flatpak install --user /run/user/1000/doc/xxx/ works)
+- ~FWUPD backend~
```
bbhtt: drop org.freedesktop.fwupd then?

bbhtt:
https://github.com/flathub/flathub/pull/7037/files#r2430940100

README.md
renner0e: The launch button doesn't work because it can't find the .desktop files from the host. If I understand the output of `strace` correctly.
I assume I'd have to change something in regards to that. I think this is part of kservice.

README.md
renner0e: I can't verify that the firmware upgrades actually work but I could do that in the future when one of my devices has a firmware update or I could test it some other way, tho I'd have to look into that.

But it shows up in the UI and everything it *should* be functional I guess?

I could remove it for now if I don't know for certain that it works. I could just remove it if you prefer it that way.

renner0e:
bot, build

renner0e:
Seems like there is something wrong with the qcoro build from the runtime on ARM
https://github.com/flathub-infra/vorarbeiter/actions/runs/18575023892/job/52958264175#step:15:5486

renner0e:
bot, build

renner0e:
bot, build

README.md
renner0e: update on that:
It worked, but the button prompting you for the reboot in discover didn't, normal reboot was fine as well of course. 

### Before:
image

### After:
image

renner0e:
bot, build

README.md
JakobDev: By taking a [look at the source](https://invent.kde.org/plasma/discover/-/blob/93363d68711b2cc15f6121ed046997bee3f2202f/discover/DiscoverObject.cpp#L654-L687) you should give acces to `org.kde.LogoutPrompt`, `org.kde.Shutdown` and `org.freedesktop.login1`. I think I also found a bug: `org.freedesktop.login1` is on the system and not on the session bus.

bbhtt: what is org.freedesktop.login1 used for?

renner0e:
@bbhtt sorry for the ping, anything else you want me to do?

bbhtt:
was https://github.com/flathub/flathub/pull/7037#discussion_r2450604937 resolved?

also testing this to say what is working / not working would be good.

README.md
JakobDev: It looks like it's supposed to trigger a Reboot on non KDE Desktops. But it's not working anyway.

renner0e:
bot, build

README.md
renner0e: > By taking a [look at the source](https://invent.kde.org/plasma/discover/-/blob/93363d68711b2cc15f6121ed046997bee3f2202f/discover/DiscoverObject.cpp#L654-L687) you should give acces to `org.kde.LogoutPrompt`, `org.kde.Shutdown` and `org.freedesktop.login1`. I think I also found a bug: `org.freedesktop.login1` is on the system and not on the session bus.

yup the org.kde.* permissions indeed make it work.
Didn't test the other thing on non-KDE

renner0e: [`9922f06` (#7037)](https://github.com/flathub/flathub/pull/7037/commits/9922f069ad0789be940b629b5a8ebcb07fefdc7a)

renner0e:
> was [#7037 (comment)](https://github.com/flathub/flathub/pull/7037#discussion_r2450604937) resolved?
> 
> also testing this to say what is working / not working would be good.

it works now, documented more functionalities
https://github.com/flathub/flathub/pull/7037#discussion_r2473254903

renner0e:
bot, build

PR #7078, opened 2025-10-25
Add org.equicord.equibop

Creationsss:

- [x] Please describe the application briefly. < Equibop is a custom Discord desktop client based on Vesktop, with Equicord pre-installed. It provides better performance and additional features for Discord users on Linux. >
- [x] Please attach a video showcasing the application on Linux using the Flatpak. <https://atums.world/u/equibop_flathub.mp4>
- [x] I am an developer to the project.

### DIFF
--- a/org.equicord.equibop.yml
+++ b/org.equicord.equibop.yml
@@ -66,7 +66,7 @@ modules:
               desktop-file-edit --set-key=Exec --set-value="startequibop %U" build/org.equicord.equibop.desktop
               install -Dm644 build/org.equicord.equibop.desktop /app/share/applications/org.equicord.equibop.desktop
               install -Dm755 startequibop /app/bin/startequibop
-              install -D org.equicord.equibop.metainfo.xml -t /app/share/metainfo/
+              install -D build/org.equicord.equibop.metainfo.xml -t /app/share/metainfo/
 
               mv "$DIST_DIR" /app/bin/equibop
 


Creationsss:
bot, build

Creationsss:
bot, build

Creationsss:
bot, build

flathub.json
hfiguiere, line 1: remove this, it's unecessary

org.equicord.equibop.yml
```diff
@@ -0,0 +1,85 @@
+app-id: org.equicord.equibop
+
+runtime: org.freedesktop.Platform
+runtime-version: &runtime-version '24.08'
```
hfiguiere: current is 25.08

org.equicord.equibop.yml
```diff
@@ -0,0 +1,85 @@
...
+      - desktop-file-edit --set-key="Exec" --set-value="startequibop %U" --set-icon=$FLATPAK_ID
+        squashfs-root/equibop.desktop
+      - install -D squashfs-root/equibop.svg /app/share/icons/hicolor/scalable/apps/org.equicord.equibop.svg
+      - install -D squashfs-root/.DirIcon /app/share/icons/hicolor/256x256/apps/org.equicord.equibop.png
+      - install -D squashfs-root/equibop.desktop /app/share/applications/org.equicord.equibop.desktop
+      - install -Dm755 startequibop /app/bin/startequibop
+      - sed -i 's|<launchable type="desktop-id">equibop.desktop</launchable>|<launchable
+        type="desktop-id">org.equicord.equibop.desktop</launchable>|' org.equicord.equibop.metainfo.xml
+      - sed -i 's|</description>|</description>\n  <developer id="org.equicord">Equicord</developer>|'
+        org.equicord.equibop.metainfo.xml
```
hfiguiere: don't use sed to modify files. it has no failure mode and it's too obscure. For the desktop file look at `rename-desktop-file instead`. And for the icon `rename-icon`. or a patch.

Creationsss:
bot, build

org.equicord.equibop.yml
```diff
@@ -0,0 +1,81 @@
...
+  - --socket=wayland
+  - --share=ipc
+  - --share=network
+  - --talk-name=org.kde.StatusNotifierWatcher # Tray functionalities on KDE
+  - --talk-name=com.canonical.AppMenu.Registrar
+  - --device=all # Needed for GPU acceleration and the webcam
+  - --filesystem=xdg-videos:ro
+  - --filesystem=xdg-pictures:ro
+  - --filesystem=xdg-download # This and the above two are used for drag-n-drop, and download managing
+  - --filesystem=xdg-run/pipewire-0 # Pipewire interfacing
```
bbhtt:
```suggestion
  - --filesystem=xdg-run/pipewire-0:ro # Pipewire interfacing
```

org.equicord.equibop.yml
```diff
@@ -0,0 +1,81 @@
...
+  - --filesystem=xdg-run/speech-dispatcher # For TTS and VcNarrator
```
bbhtt:
```suggestion
  - --filesystem=xdg-run/speech-dispatcher:ro # For TTS and VcNarrator
```

org.equicord.equibop.yml
```diff
@@ -0,0 +1,81 @@
...
+  - --filesystem=~/.steam # Needed for SteamOS integration, as the XDG OpenURL portal doesn't work properly in Game Mode. We only need ~/.steam/steam.pipe but Flatpak can't correctly mount just that: 'File "/home/deck/.steam/steam.pipe" has unsupported type 0o10000'
```
bbhtt: do you need the whole folder or it can be limited to subfolders?

org.equicord.equibop.yml
```diff
@@ -0,0 +1,81 @@
...
+        sha512: c871d362f9a1652ee45e7f1c743f647698f262e7345bdae7fe14bbad770280a8f6b0a71822263bbb65bfd6879a2dca307a7502ffa45f28166bae2c9cd0ecbde5
+        x-checker-data:
+          type: json
+          url: https://api.github.com/repos/Equicord/Equibop/releases/latest
+          version-query: .tag_name | sub("v"; "")
+          url-query: .assets[] | select(.name == "org.equicord.equibop.metainfo.xml")
+            | .browser_download_url
+
+      - type: file
+        url: https://github.com/Equicord/Equibop/releases/download/v3.0.4/Equibop-3.0.4.AppImage
```
bbhtt: this is oss it should be built from source

bbhtt: https://docs.flathub.org/docs/for-app-authors/requirements#building-from-source

Creationsss: This is a fork of https://github.com/flathub/dev.vencord.Vesktop, which also packages pre-built AppImages rather than building from source. Since we're using the same approach as the upstream project that's already on Flathub, do we really need to change this to build from source?

bbhtt: Yes

org.equicord.equibop.yml
Creationsss: whole folder

Creationsss:
bot, build

Creationsss:
bot, build

Creationsss:
bot, build

Creationsss:
bot, buildbot, build

Creationsss:
bot, buildbot, build

org.equicord.equibop.yml
```diff
@@ -0,0 +1,195 @@
...
+    - --filesystem=xdg-videos:ro
+    - --filesystem=xdg-pictures:ro
+    - --filesystem=xdg-download # This and the above two are used for drag-n-drop, and download managing
+    - --filesystem=xdg-run/pipewire-0:ro # Pipewire interfacing
+    - --filesystem=xdg-run/speech-dispatcher:ro # For TTS and VcNarrator
+    - --filesystem=~/.steam # Needed for SteamOS integration, as the XDG OpenURL portal doesn't work properly in Game Mode. We only need ~/.steam/steam.pipe but Flatpak can't correctly mount just that: 'File "/home/deck/.steam/steam.pipe" has unsupported type 0o10000'
+    - --env=XCURSOR_PATH=/run/host/user-share/icons:/run/host/share/icons # This is used to show the correct cursor on Wayland
+
+modules:
+    - name: bun
```
bbhtt: this is build only dependency? if so then it should be cleaned up.

org.equicord.equibop.yml
```diff
@@ -0,0 +1,195 @@
...
+
+              # Install desktop file, startup script and metainfo
+              ls -la org.equicord.equibop.desktop || echo "Desktop file not found!"
+              install -Dm644 org.equicord.equibop.desktop /app/share/applications/org.equicord.equibop.desktop
+              install -Dm755 startequibop /app/bin/startequibop
+              install -D org.equicord.equibop.metainfo.xml -t /app/share/metainfo/
+
+              # Verify desktop file was installed
+              ls -la /app/share/applications/org.equicord.equibop.desktop || echo "Desktop file installation failed!"
+
```
bbhtt:
```suggestion
```

org.equicord.equibop.yml
```diff
@@ -0,0 +1,195 @@
...
+              else
+                echo "Error: Could not find unpacked directory"
+                exit 1
+              fi
+
+              # Install icons
+              install -Dm644 build/icon.svg /app/share/icons/hicolor/scalable/apps/org.equicord.equibop.svg
+
+              # Install desktop file, startup script and metainfo
+              ls -la org.equicord.equibop.desktop || echo "Desktop file not found!"
```
bbhtt:
```suggestion
```

org.equicord.equibop.desktop
bbhtt, line 1: desktop file should be upstream not here, you can modify the exec with desktop file edit in the manifest.

org.equicord.equibop.yml
Creationsss: no its needed at runtime, L92

org.equicord.equibop.yml
```diff
@@ -0,0 +1,195 @@
...
+
+              # Create version file
+              echo '{"version":"38.4.0"}' > node_modules/electron/dist/version
+
+              # Restore electron package files
+              cp electron-backup/package.json node_modules/electron/
+              cp electron-backup/path.txt node_modules/electron/
+
+          - patch -p1 < sandboxFix.patch
+          - patch -p1 < beforePack.patch
```
hfiguiere: this is not how you apply patches.

org.equicord.equibop.yml
```diff
@@ -0,0 +1,195 @@
...
+            sha512: 0080fad9b4f64f919f049c065081e1b0f7e02c6d42a5fd89a6a29d73b30645ba0b75d6fbe839d4a3dd8d212009e807167ebda75ed8e1cd37fedbaac46aaeaae6
+            strip-components: 1
+            x-checker-data:
+                type: json
+                url: https://api.github.com/repos/Equicord/Equibop/releases/latest
+                version-query: .tag_name | sub("v"; "")
+                url-query: .tarball_url
+                is-main-source: true
+
+          - type: file
```
hfiguiere, line 94: why isn't this an archive type?

org.equicord.equibop.yml
```diff
@@ -0,0 +1,195 @@
...
+            only-arches: [x86_64]
+            x-checker-data:
+                type: json
+                url: https://api.github.com/repos/Equicord/Equibop/releases/latest
+                version-query: .tag_name | sub("v"; "")
+                url-query:
+                    .assets[] | select(.name == "node_modules-x64.tar.gz")
+                    | .browser_download_url
+
+          - type: file
```
hfiguiere, line 107: same as above

org.equicord.equibop.yml
```diff
@@ -0,0 +1,195 @@
...
+    - --env=XCURSOR_PATH=/run/host/user-share/icons:/run/host/share/icons # This is used to show the correct cursor on Wayland
+
+modules:
+    - name: bun
+      buildsystem: simple
+      build-commands:
+          - install -Dm755 bun /app/bin/bun
+      sources:
+          - type: archive
+            url: https://github.com/oven-sh/bun/releases/download/bun-v1.3.1/bun-linux-x64.zip
```
hfiguiere: why isn't this build from source either?

org.equicord.equibop.yml
hfiguiere: that's not runtime. that's build time.

org.equicord.equibop.yml
Creationsss: These are `type: file` (not `type: archive`) because `type: archive` auto-extracts into the build directory. We need manual control to extract node_modules.tar.gz at line 60 and electron zips into node_modules/electron/dist at line 71.

org.equicord.equibop.yml
Creationsss: above

org.equicord.equibop.yml
Creationsss: Bun is copied into resources/bun/ at line 90, which gets included in the final app when dist/ is moved to /app/bin/equibop at line 114. The app needs bun at runtime.

org.equicord.equibop.yml
hfiguiere: and you get that manual control. It's all perfectly documented.
From the man page:
```
           dest (string)
               Directory inside the source dir where this source will be extracted.
```

org.equicord.equibop.yml
hfiguiere: see how wrong that is above.

org.equicord.equibop.yml
hfiguiere: There is also
```
           strip-components (integer)
               The number of initial pathname components to strip during extraction. Defaults to 1.
```

org.equicord.equibop.yml
hfiguiere: so there will be two copies at the end? Exactly the original point.

org.equicord.equibop.yml
Creationsss: should be fixed

org.equicord.equibop.yml
Creationsss: Building Bun from source requires Zig, LLVM, and WebKit, and would significantly increasing building, since its a build tool and a run dep and not system wide i don't think its necessary

org.equicord.equibop.yml
```diff
@@ -0,0 +1,167 @@
...
+
+modules:
+    - name: Equibop
+      buildsystem: simple
+      build-commands:
+          - tar xzf node_modules.tar.gz
+
+          - |
+              # Copy Bun binary to resources directory for runtime use
+              ARCH=$(uname -m)
```
hfiguiere: just use `$FLATPAK_ARCH` here

org.equicord.equibop.yml
Creationsss: should be fixed

Creationsss:
bot, build

org.equicord.equibop.yml
```diff
@@ -0,0 +1,164 @@
...
+
+              # Install the application
+              mv "$DIST_DIR" /app/bin/equibop
+
+      sources:
+          - type: file
+            path: startequibop
+
+          - type: file
+            url: https://github.com/Equicord/Equibop/releases/latest/download/org.equicord.equibop.metainfo.xml
```
hfiguiere: you should NEVER reference the tip of a branch.

hfiguiere: why isn't it in the tarball?

Creationsss: > why isn't it in the tarball?

https://github.com/Equicord/Equibop/blob/main/.github/workflows/meta.yml



> you should NEVER reference the tip of a branch.

fixed

Creationsss:
bot, build

org.equicord.equibop.yml
hfiguiere: > > why isn't it in the tarball?
> 
> https://github.com/Equicord/Equibop/blob/main/.github/workflows/meta.yml

So it's wrong. It's part of the release. It should be updated before the tag is done.

Creationsss: > > > why isn't it in the tarball?
> > 
> > 
> > https://github.com/Equicord/Equibop/blob/main/.github/workflows/meta.yml
> 
> So it's wrong. It's part of the release. It should be updated before the tag is done.

No

The metainfo.xml is generated after the release because it contains information about that specific release (version, date, changelog). It can't be committed before the tag because it doesn't exist yet - it needs the release to exist first in order to pull that information.

hfiguiere: as I said, it's incorrect. The appstream file is part of the release so it should be updated before the release is tagged. And be part of the tarball.

You don't put the cart before the horse.

Creationsss: The generateMeta script fetches release information from the GitHub API (version, date, changelog) and appends it to the existing metainfo.xml. It requires the release to already be published to pull this data. We can't generate it before tagging because the release data doesn't exist yet in the API.

See: https://github.com/Equicord/Equibop/blob/main/scripts/utils/generateMeta.mts#L39

If you really force this it would either require
 - Manually updating it each time i make a release
  - Rewriting generateMeta.mts
  - Maintaining a CHANGELOG.md file
  - Changing my release process
  
  this is really going off kilter just for flathhub as this is a fork of an already existing application on flathub

hfiguiere: > this is really going off kilter just for flathhub as this is a fork of an already existing application on flathub 

two wrongs don't make a right. you don't fix bugs in your app?

hfiguiere: and just to diffuse an incorrect assumption: appstream files are not flathub specific. RPM and DEB packages will also install them.

Creationsss:
bot, build

Creationsss:
bot, build

bbhtt:
@Creationsss I notice now Equibop was already on Flathub under the ID https://flathub.org/en/apps/io.github.equicord.equibop

So what was this submission about?

Creationsss:
> @Creationsss I notice now Equibop was already on Flathub under the ID https://flathub.org/en/apps/io.github.equicord.equibop
> 
> So what was this submission about?

The entire app was remade from scratch the old code had way too many bugs. The app ID was changed, and the developer/maintainer changed as well.

bbhtt:
So what should be done with the old one? We can't have two copies of the same app so one of them is getting dropped.

Creationsss:
> So what should be done with the old one? We can't have two copies of the same app so one of them is getting dropped.

drop https://flathub.org/en/apps/io.github.equicord.equibop

bbhtt:
can we [eol rebase](https://docs.flathub.org/docs/for-app-authors/maintenance#end-of-life-rebase) that to `org.equicord.equibop` so that people are automatically migrated by Flatpak to this?

Creationsss:
> can we [eol rebase](https://docs.flathub.org/docs/for-app-authors/maintenance#end-of-life-rebase) that to `org.equicord.equibop` so that people are automatically migrated by Flatpak to this?

i'll get to it when i get home from work

bbhtt:
you don't have to do it, we'll do it. I'm asking to see if it makes sense to migrate v2 equibop to v3 automatically.

also it'd be nice if you can verify this one, since the old one was verified.

Creationsss:
> you don't have to do it, we'll do it. I'm asking to see if it makes sense to migrate v2 equibop to v3 automatically.
> 
> 
> 
> also it'd be nice if you can verify this one, since the old one was verified.

yeah i think it would make sense, ill verify it

Creationsss:
app should be verified, or should be in the process? 
image

bbhtt:
yes thanks.

PR #7113, opened 2025-11-02
add org.fcast.Receiver

MoralCode:

- [X] Please describe the application briefly. This is the reference Receiver application for playing content received via FCast - an open source protocol that enables wireless streaming of audio and video content between devices, supporting various stream types such as DASH, HLS, and mp4.
- [X] Please attach a video showcasing the application on Linux using the Flatpak. https://youtu.be/ty3Ok6CpJ3k
- [X] I am an _(please keep whichever is applicable and remove the rest)_ None of the listed options.
      If not, I contacted upstream developers about this submission. I will do so privately via internal channels and have them confirm in this issue

Michael-Hollister
zvonimir-futo

### DIFF
--- a/org.fcast.Receiver.yaml
+++ b/org.fcast.Receiver.yaml
@@ -71,6 +71,9 @@ modules:
       - type: patch
         path: patches/nonetwork.patch
       
+      - type: patch
+        path: patches/captions.patch
+      
       - npm-sources.json
       
       - type: file
--- a/patches/captions.patch
+++ b/patches/captions.patch
@@ -1,0 +1,29 @@
+diff --git a/.flatpak/org.fcast.Receiver.metainfo.xml b/.flatpak/org.fcast.Receiver.metainfo.xml
+index 940e164..e87f2b6 100644
+--- a/.flatpak/org.fcast.Receiver.metainfo.xml
++++ b/.flatpak/org.fcast.Receiver.metainfo.xml
+@@ -52,19 +52,19 @@
+   <screenshots>
+     <screenshot type="default">
+       <image>https://gitlab.futo.org/videostreaming/fcast/-/raw/master/.flatpak/screenshots/org.fcast.Receiver/main.png</image>
++      <caption>The main window of the FCast Receiver showing the connection information and a loading spinner.</caption>
+     </screenshot>
+-    <caption>The main window of the FCast Receiver showing the connection information and a loading spinner.</caption>
+     <screenshot>
+       <image>https://gitlab.futo.org/videostreaming/fcast/-/raw/master/.flatpak/screenshots/org.fcast.Receiver/main_connected.png</image>
++      <caption>The main window of the FCast Receiver showing a device has been connected.</caption>
+     </screenshot>
+-    <caption>The main window of the FCast Receiver showing a device has been connected.</caption>
+     <screenshot>
+       <image>https://gitlab.futo.org/videostreaming/fcast/-/raw/master/.flatpak/screenshots/org.fcast.Receiver/player1.png</image>
++      <caption>The player window of the FCast Receiver showing some content playing with media controls at the bottom of the screen.</caption>
+     </screenshot>
+-    <caption>The player window of the FCast Receiver showing some content playing with media controls at the bottom of the screen.</caption>
+     <screenshot>
+       <image>https://gitlab.futo.org/videostreaming/fcast/-/raw/master/.flatpak/screenshots/org.fcast.Receiver/player2.png</image>
++      <caption>The player window of the FCast Receiver showing some content playing with media controls hidden.</caption>
+     </screenshot>
+-    <caption>The player window of the FCast Receiver showing some content playing with media controls hidden.</caption>
+   </screenshots>
+ </component>
+\ No newline at end of file


Michael-Hollister:
I am aware of this flatpak listing and approve of it being submitted to flathub.

org.fcast.Receiver.desktop
hfiguiere, line 1: This file should be (submitted and managed) upstream.

org.fcast.Receiver.metainfo.xml
hfiguiere, line 1: This file should be (submitted and managed) upstream.

Makefile
```diff
@@ -0,0 +1,25 @@
+validate-meta:
+	appstreamcli validate org.fcast.Receiver.metainfo.xml
```
hfiguiere: you should use the one in the flatpak as per the documentation. Otherwise it's pointless.

org.fcast.Receiver.yaml
```diff
@@ -0,0 +1,91 @@
...
+        npm_config_cache: /run/build/fcast-receiver/flatpak-node/npm-cache
+        npm_config_nodedir: /usr/lib/sdk/node22
+        npm_config_offline: 'true'
+    subdir: receivers/electron/
+    build-commands:
+      # Install npm dependencies
+      - npm install --offline
+      # run the build
+      - npm run build
+      - ../../deploy_flatpak.sh
```
hfiguiere: why is this a script?

org.fcast.Receiver.yaml
```diff
@@ -0,0 +1,91 @@
+app-id: org.fcast.Receiver
+runtime: org.freedesktop.Platform
+runtime-version: '24.08'
```
hfiguiere: current is 25.08

org.fcast.Receiver.yaml
```diff
@@ -0,0 +1,91 @@
...
+base-version: '24.08'
+
+command: run.sh
+finish-args:
+  - --share=network
+  - --socket=pulseaudio
+  - --socket=x11
+  - --share=ipc
+  - --device=dri
+  - --system-talk-name=org.fcast.Receiver
```
hfiguiere, line 16: why do you need system-talk?

org.fcast.Receiver.yaml
```diff
@@ -0,0 +1,91 @@
...
+  - --env=ELECTRON_TRASH=gio
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node22
+build-options:
+  append-path: /usr/lib/sdk/node22/bin
+  cflags: -O2 -g
+  cxxflags: -O2 -g
```
hfiguiere:
```suggestion
```

Makefile
MoralCode: What do you mean by this? Are you saying to use the `appstreamcli` binary available within the flatpak (i.e. change this to a `flatpak run <something> appstreamcli validate` command?)

org.fcast.Receiver.yaml
MoralCode: Because some if-then logic is needed to translate between the names that flatpak uses for the different architectures and what the build script for the application requires. I don't recall seeing a way to only run specific lines of the build steps on certain architectures the way you can limit i.e. file sources to only apply on certain architectures.

org.fcast.Receiver.yaml
MoralCode: will update, thanks!

org.fcast.Receiver.yaml
MoralCode: Leaving it out caused a bug (https://github.com/futo-org/fcast/issues/50?notification_referrer_id=NT_kwDOAQjwBbQyMDEwMzYwOTgzODoxNzM2Mjk0OQ#issuecomment-3478073782) where the system tray icon would not react to right click events, causing the app to not be able to be quit without using `flatpak kill`, especially on KDE desktops. Testing revealed that adding a talk name, even without the broad system bus access that a `--socket=system-bus` would allow, fixed the issue

Makefile
hfiguiere: https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines#validation

MoralCode: Oh right, i see the redundancy. I already have a `make lint` step in the makefile that runs things this way. i will adjust things to use that rather than split the linting across two make instructions

org.fcast.Receiver.yaml
MoralCode: oh i think i see a better way to do it

Makefile
MoralCode: moved this to the `make lint` section and rewrote it to use the linter feature for appstream files

org.fcast.Receiver.yaml
MoralCode: moved the actual build step back into the manifest and changed to `source`-ing the shell script to get the parameter needed to specify the architecture.
This follows the pattern set forth in the documentation [here](https://github.com/flatpak/flatpak-builder-tools/blob/ea92dc22ab7e4ab44133407b883c9a6792e54302/node/README.md?plain=1#L234-L240) except modified to use electron-forge style arguments instead of electron-builder.

org.fcast.Receiver.desktop
MoralCode: I have confirmed with @Michael-Hollister that this is something he is willing to do as the upstream maintainer. Because the version of the upstream project being packaged here does not currently contain these files, they have been added in the flatpak build process. I will work with Michael to have them upstreamed for the next release

org.fcast.Receiver.metainfo.xml
MoralCode: I have confirmed with @Michael-Hollister that this is something he is willing to do as the upstream maintainer. Because the version of the upstream project being packaged here does not currently contain these files, they have been added in the flatpak build process. I will work with Michael to have them upstreamed for the next release

MoralCode:
bot, build

org.fcast.Receiver.desktop
MoralCode: This has been upstreamed now. https://gitlab.futo.org/videostreaming/fcast/-/commit/458ea06cc414cbd0b476042250684be674c93bf5

org.fcast.Receiver.metainfo.xml
MoralCode: This has been upstreamed now. https://gitlab.futo.org/videostreaming/fcast/-/commit/458ea06cc414cbd0b476042250684be674c93bf5

MoralCode:
bot, build

MoralCode:
bot, build

PR #7129, opened 2025-11-05
Create issue template for slow connection reports

razzeee:
This now references https://github.com/flathub-infra/documentation/pull/527 so we might want to change that, if that page goes away


.github/ISSUE_TEMPLATE/3-slow-connection.yml
```diff
@@ -0,0 +1,118 @@
...
+      label: Flatpak Verbose Output
+      description: |
+        Attach the verbose output from a failing `flatpak update` or `flatpak install` run:
+        ```bash
+        flatpak update --verbose --ostree-verbose
+        ```
+
+        Or for a specific app:
+        ```bash
+        flatpak update --verbose --ostree-verbose org.gnome.Calculator
```
bbhtt: this will fail if they don't have this installed, better to change it to an install command that installs some small app

.github/ISSUE_TEMPLATE/3-slow-connection.yml
```diff
@@ -0,0 +1,118 @@
+name: Slow Connection to Flathub
+description: Report slow downloads from Flathub with diagnostic information
+title: "Slow downloads from Flathub (dl.flathub.org) — [Country/ISP]"
+labels: ["connectivity", "performance"]
```
bbhtt:
```suggestion
```

bbhtt:
I don't like issue templates with mandatory fields or so this should be something much simpler like https://github.com/flathub/flathub/blob/master/.github/ISSUE_TEMPLATE/1-quality-mod.md?plain=1 - one header for describing the problem and one header for the related logs/infos.

I'd reduce the logs to collect to 1. confirming the problem is isolated to only Flathub 2. `flatpak install --verbose --ostree-verbose $id`.

.github/ISSUE_TEMPLATE/3-slow-connection.yml
```diff
@@ -0,0 +1,118 @@
...
+      required: false
+
+  - type: input
+    id: isp
+    attributes:
+      label: ISP / Network
+      description: Your Internet Service Provider name, home/office network, NAT/CGNAT if known
+      placeholder: "e.g., Comcast, home network, behind CGNAT"
+    validations:
+      required: false
```
bbhtt:
```suggestion
```

.github/ISSUE_TEMPLATE/3-slow-connection.yml
razzeee: Changed

razzeee:
This is now a simple markdown template

PR #7130, opened 2025-11-05
Add com.actualbudget.actual

MikesGlitch:

- [x] Please describe the application briefly: Actual Budget is a privacy-focused app for managing finances
- [x] Please attach a video showcasing the application on Linux using the Flatpak. [demo.webm](https://github.com/user-attachments/assets/2441ccae-bb45-42c0-b501-4cddc5d2862f)
- [x] I am one of the core developers of the project

**Other core maintainers of Actual:**
@MatissJanis
@jfdoming
@youngcw
@joel-jeremy
@matt-fidd
@lelemm

Upstream repo: https://github.com/actualbudget/actual

### DIFF
--- a/com.actualbudget.actual.yml
+++ b/com.actualbudget.actual.yml
@@ -54,14 +54,14 @@ modules:
     sources:
       - type: file
         dest-filename: Actual.AppImage
-        url: https://github.com/actualbudget/actual/releases/download/v25.11.0/Actual-linux-x86_64.AppImage
-        sha256: 45d3a11d7997c742e2fe715872b4e7360040bf3e821570313f799c2bfdc3410d
+        url: https://github.com/actualbudget/actual/releases/download/v25.12.0/Actual-linux-x86_64.AppImage
+        sha256: a13ab7b10972a29628de1c14ece9844d4a258a810337f973dc53836015110068
         only-arches:
           - x86_64
 
       - type: file
         dest-filename: Actual.AppImage
-        url: https://github.com/actualbudget/actual/releases/download/v25.11.0/Actual-linux-arm64.AppImage
-        sha256: f4b0ae9db1b46949e4241584cb29c12043687d8b951d295e627518dae89788a1
+        url: https://github.com/actualbudget/actual/releases/download/v25.12.0/Actual-linux-arm64.AppImage
+        sha256: c15e5be1d8291f2d7351e86e3fb41c03838395cad873173ae15a45371542db7d
         only-arches:
           - aarch64


com.actualbudget.actual.yml
```diff
@@ -0,0 +1,68 @@
...
+    sources:
+      - type: script
+        dest-filename: actual-wrapper.sh
+        commands:
+          - exec zypak-wrapper /app/main/actual "$@"
+
+  - name: actual
+    buildsystem: simple
+    build-commands:
+      # Unable to build from source due to flatpak-node-generator not supporting yarn v4, so using AppImage extraction
```
MikesGlitch, line 45: I attempted to build from source, but the tooling doesn't seem to support yarn v4:

image

com.actualbudget.actual.yml
```diff
@@ -0,0 +1,68 @@
...
+      - chmod +x ./Actual.AppImage
+      - ./Actual.AppImage --appimage-extract
+      - mkdir -p /app/main
+      - cp -R squashfs-root/. /app/main
+      - desktop-file-edit --set-icon ${FLATPAK_ID} --set-key Exec --set-value 'actual %u' /app/main/actual.desktop
+      - install -Dm644 /app/main/actual.desktop /app/share/applications/${FLATPAK_ID}.desktop
+      - install -Dm644 /app/main/usr/share/icons/hicolor/256x256/apps/actual.png /app/share/icons/hicolor/256x256/apps/${FLATPAK_ID}.png
```
MikesGlitch, line 52: Ideally the image would be 512, but the last release of Actual didn't have that image available in the AppImage. We'll include the 512 image in the next release. 256x256 seems to render fine and Actual's maintainers would be happy to go with that until next release.

com.actualbudget.actual.yml
```diff
@@ -0,0 +1,68 @@
...
+
+  # Network access
+  - --share=network
+
+  # Use wayland but allow X11 fallback
+  - --socket=fallback-x11
+  - --socket=wayland
+
+  # Home directory & external drive access for budget backups (location is configurable in Actual)
+  - --filesystem=home
```
MikesGlitch: I got this when I ran the lint:

Screenshot from 2025-11-05 18-51-40

I'm not sure what to do here - we'd like to give the app home filesystem access so that the user can save their budgets wherever they want.

MikesGlitch:
> The domain to be used for verification is actualbudget.com. If you intend to verify this app, please confirm that you can upload https://actualbudget.com/.well-known/org.flathub.VerifiedApps.txt. Otherwise, ignore this. Please comment if this incorrect.

We're able to upload this file and would like to verify the app.

com.actualbudget.actual.yml
bbhtt: you can bypass the tooling by creating a vendored tarball of your node_modules.

com.actualbudget.actual.yml
bbhtt: would be nice if there was a default location scoped to some single directory. Does this "saving" go through a filechooser? Then if that supports the filechooser portal you won't need filesystem access.

com.actualbudget.actual.yml
```diff
@@ -0,0 +1,68 @@
...
+  - --filesystem=/media
+  - --filesystem=/run/media
+  - --filesystem=/mnt
```
bbhtt: these seem a bit unncessary.

flathub.json
bbhtt, line 1: drop this, it is not needed.

com.actualbudget.actual.yml
MikesGlitch: By default the app will save the budget to the documents directory under `~/Documents/Actual`. That all gets setup automatically on first launch. The app will write to that directory whenever anything is changed and it will store some backups in there as well. 

I assume it's common practice for users to adjust the directory permissions manually as they change them, so I have added `- --filesystem=xdg-documents` and removed the others.

com.actualbudget.actual.yml
MikesGlitch: Just tried that out, I've managed to get most of it building, but now I'm running into issues with @electron/rebuild + node-gyp. 

We have after-build hooks in our electron-builder config that ensure we're have the correctly built arch for some modules. That seems to be making network calls to download some headers.

image

I could try set it up, but this is a dependency of a dependency and feels like it would be flaky/hard to maintain - we do upgrades of these packages quite frequently. The AppImage extract seems like it would easier to maintain. 

Is it a hard requirement to somehow work around this?

bbhtt: Figuring out a non networked build solution would be good, if it gets too complicated i can think of exceptions.

MikesGlitch:
bot, build

MikesGlitch:
> ❌ [Test build](https://github.com/flathub-infra/vorarbeiter/actions/runs/19180699616) failed.
> 
> Help

Classic :man_facepalming: 
image

com.actualbudget.actual.yml
MikesGlitch: I'm talking it over with the other maintainers. The general feeling is we don't want to publish new release assets for these dependencies. We have a rule to only ever release something if the release is automated/has low maintenance burden. It might be worth waiting for yarn v4 support before progressing. 

I worked around it earlier today by manually doing what flatpak-node-generator does. Looking up the files, getting their sha256's and placing them in the right directories for them to be picked up by the process (preventing electron-builder from downloading it). It took a bit of effort, and we'd have to re-do it when we upgrade electron (happens multiple times a year - more maintenance). There's also the issue of the node versions potentially going out of sync between upstream pipeline's and this build agent - I think the prebuilds meant for node22 wouldn't work for node 23 for things like bettersqlite etc. Using flatpak-node-generator would solve that issue. 

Looks like this is going to be stale for a month either way - we release monthly and the metainfo file is invalid (it passed locally - i must have an old version) :man_facepalming:

MikesGlitch:
bot, build

com.actualbudget.actual.yml
MikesGlitch: Hi :wave: 

We gave building from source a shot without the flatpak-node-generator. It meant matching node versions and building two sets of node_modules (x64 and aarch64), publishing an artefact on the release etc. The other maintainers decided that it wasn't worth keeping.

What we've got in this PR is as far as we'll take it. If it's possible to progress with this we're willing, but if this isn't good enough for flathub we totally understand. There's a chance someone from the community could pick it up/it could be revisited at a later date.

bbhtt: can you explain how the current appimages are produced and tagged? link to ci yaml, pipelines etc.?

MikesGlitch: Our release process is documented here: https://actualbudget.org/docs/contributing/releasing

Essentially we create a tag and that triggers the release workflows. 

**This is our desktop app workflow:** https://github.com/actualbudget/actual/blob/master/.github/workflows/electron-master.yml

**It handles Window, Mac and Linux - for AppImages it:** 
- Sets up the environment (installs dependencies, clones a translations repo for in-app translations) [action](https://github.com/actualbudget/actual/blob/master/.github/actions/setup/action.yml)
- Builds the desktop app
  -  Includes building our yarn workspace packages [Line 63](https://github.com/actualbudget/actual/blob/cc4458c9ed22ef74b71a8420ff8a83841349e09f/.github/workflows/electron-master.yml#L63)
  -  Uses Electron builder to build the AppImage [build config](https://github.com/actualbudget/actual/blob/master/packages/desktop-electron/package.json)
 - Uploads the built AppImage to the release [Line 85](https://github.com/actualbudget/actual/blob/cc4458c9ed22ef74b71a8420ff8a83841349e09f/.github/workflows/electron-master.yml#L85)

**Here's our releases:** https://github.com/actualbudget/actual/releases

**Our desktop app release workflow runs can be found here:** https://github.com/actualbudget/actual/actions/workflows/electron-master.yml

**The last one was:** https://github.com/actualbudget/actual/actions/runs/19902268093

PR #7174, opened 2025-11-19
Add com.onclebob.Bob

pcarrier:
Only linting error is the filesystem access. Unfortunately it is needed today for the app to function with projects located anywhere on the filesystem. If and when this improves I'll restrict permissions further.

- [x] Please describe the application briefly. Start, build, preview, launch static websites without the terminal.
- [x] Please attach a video showcasing the application on Linux using the Flatpak. https://github.com/user-attachments/assets/d1cb1aee-0884-4269-b7d0-5661ea0177a3

### DIFF
--- a/com.onclebob.Bob.yml
+++ b/com.onclebob.Bob.yml
@@ -1,0 +1,47 @@
+# Generated from flatpak-flutter.yml, do not edit
+# Visit the flatpak-flutter project at https://github.com/TheAppgineer/flatpak-flutter
+app-id: com.onclebob.Bob
+runtime: org.freedesktop.Platform
+runtime-version: '25.08'
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.llvm21
+command: com.onclebob.Bob
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --filesystem=host
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --device=dri
+modules:
+  - name: Bob
+    buildsystem: simple
+    build-options:
+      arch:
+        x86_64:
+          env:
+            BUNDLE_PATH: build/linux/x64/release/bundle
+        aarch64:
+          env:
+            BUNDLE_PATH: build/linux/arm64/release/bundle
+      append-path: /usr/lib/sdk/llvm21/bin:/var/lib/flutter/bin
+      prepend-ld-library-path: /usr/lib/sdk/llvm21/lib
+      env:
+        PUB_CACHE: /run/build/Bob/.pub-cache
+    build-commands:
+      - setup-flutter.sh -C .
+      - flutter build linux --release --no-pub
+      - install -D $BUNDLE_PATH/bob /app/bin/com.onclebob.Bob
+      - cp -r $BUNDLE_PATH/lib /app/bin/lib
+      - cp -r $BUNDLE_PATH/data /app/bin/data
+      - install -Dm644 linux/com.onclebob.Bob.png /app/share/icons/hicolor/512x512/apps/com.onclebob.Bob.png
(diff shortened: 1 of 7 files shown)


dev.xmit.OncleBob.desktop
bbhtt, line 1: this should be upstream

bbhtt:
> Unfortunately it is needed today for the app to function with projects located anywhere on the filesystem. If and when this improves I'll restrict permissions further.

It won't improve unless you fix it. You can have a single directory like ~/Projects and configure the UI accordingly.

pcarrier:
> It won't improve unless you fix it. You can have a single directory like ~/Projects and configure the UI accordingly.

I could but that would be forcing my choice onto the user, and I'm not interested in doing that.
If anything because I use /src myself and I'm not about to change, but also because the next app will pick a different directory and we'll be incompatible. Also, we run arbitrary package.json scripts which might need all sorts of accesses.
By improving I meant something like filesystem bookmarks in macOS, implemented in a flutter library.
If that's blocking for flathub, let's drop this PR.
But given eg https://flathub.org/en/apps/com.axosoft.GitKraken gets full filesystem access, I think the case can be made we have the same needs for the same reasons.

pcarrier:
I can upload https://onclebob.xmit.dev/.well-known/org.flathub.VerifiedApps.txt and https://onclebob.com/.well-known/org.flathub.VerifiedApps.txt and would indeed like to verify.
Changed the bundle to com.onclebob.Bob.

dev.xmit.OncleBob.desktop
pcarrier: Done.

pcarrier:
OK I think it should be good to go modulo the perms issue. Please let me know if that's a blocker.

bot, build

pcarrier:
I confirm I can upload the file. Please choose the UUID.

bbhtt:
> I could but that would be forcing my choice onto the user,

It's not forcing the choice onto users but choosing between essentially unsandboxed and sandboxed. Also having good defaults is not "forcing"

pcarrier:
I already have a default. A default is something you can change, this wouldn't be that. It wouldn't be a default but a mandate, let's call a cat a cat.

I've mentioned other concerns, like arbitrary bun tasks needing to write elsewhere.

If sandboxing forces using one directory instead of another (and again it's _not_ a default if I cannot pick another value), that _is_ part of the decision. I'm made my mind, let me know if I should start my own repo and ship a ref file instead of linking to flathub. What I won't do is recommend a version that doesn't adjust to my or anyone else's workflow unless they adopt a particular file hierarchy (something other products in that category don't seem to have to do, not clear why Oncle Bob would be any different) and breaks for some projects without warning.

com.onclebob.Bob.yml
```diff
@@ -0,0 +1,46 @@
...
+      - install -D $BUNDLE_PATH/bob /app/bin/com.onclebob.Bob
+      - cp -r $BUNDLE_PATH/lib /app/bin/lib
+      - cp -r $BUNDLE_PATH/data /app/bin/data
+      - install -Dm644 linux/com.onclebob.Bob.png /app/share/icons/hicolor/512x512/apps/com.onclebob.Bob.png
+      - install -Dm644 linux/com.onclebob.Bob.desktop /app/share/applications/com.onclebob.Bob.desktop
+    sources:
+      - pubspec-sources.json
+      - type: git
+        url: https://github.com/xmit-co/bob.git
+        tag: v0.0.2-flatpak
```
hfiguiere, line 44:
```suggestion
        tag: v0.0.2-flatpak
        commit: 9de570d132aedbcf4f86c6e65a7dcede4665f834
```

bbhtt:
>  let me know if I should start my own repo and ship a ref file instead of linking to flathub.

Cool go ahead. Being so defensive from the very start leaves a bad taste anyway.

pcarrier:
I wouldn't call that the very start any more than a mandate a default, but OK.

PR #7186, opened 2025-11-20
add dev.louiscouture.path

Kibranoz:

- [x] Please describe the application briefly. This is an app that allows you to manages paths and aliases in a gui way
- [x] Please attach a video showcasing the application on Linux using the Flatpak. <

https://github.com/user-attachments/assets/67aa0447-d9dc-4663-b970-48ae93e393de

>
- [x] I am an _(please keep whichever is applicable and remove the rest)_ author/developer

### DIFF
--- a/dev.louiscouture.path.json
+++ b/dev.louiscouture.path.json
@@ -9,7 +9,8 @@
         "--socket=fallback-x11",
         "--device=dri",
         "--socket=wayland",
-        "--filesystem=home"
+        "--filesystem=~/.bashrc.d:create",
+        "--filesystem=~/.bashrc"
     ],
     "cleanup" : [
         "/include",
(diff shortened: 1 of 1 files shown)


Kibranoz:
bot, build

Kibranoz:
bot, build

bbhtt:
This does not fit Flathub use case. A GUI to modify PATH is not a standard thing to do.

Kibranoz:
> This does not fit Flathub use case. A GUI to modify PATH is not a standard thing to do.

This app will work across distributions if that is what you're asking.

The app only needs a .bashrc file on the user home, which is something nearly all distros have. If the distro does not have .bashrc.d, the app will add that directory on startup and add the script to source everything in the directory. If the user is on a distro that does not have a .bashrc by default, they can add one for the app to work. 

And people modify PATHS on Windows in a GUI . We need a way to do this in a GUI way on Linux if we want to raise Linux adoption.

Kibranoz:
bot, build

Kibranoz:
I updated the manifest to only allow the app to use the required file and folders instead of the whole home directory. This will make the app more secure.

bbhtt:
access to read bashrc is not "secure" at all. and what happens on other shells? I for example use zsh and store zshrc at `~/.config/zsh`.

PR #7188, opened 2025-11-21
Add ch.dragondreams.deigde

LordOfDragons:

- [X] Please describe the application briefly. Supports creating and modifying Drag[en]gine game projects.
- [X] Please attach a video showcasing the application on Linux using the Flatpak. https://www.youtube.com/watch?v=K_kJPcOmpBw

### DIFF
--- a/ch.dragondreams.deigde.yml
+++ b/ch.dragondreams.deigde.yml
@@ -10,7 +10,9 @@ finish-args:
   - --socket=pulseaudio
   - --socket=x11
   - --share=ipc
-  - --filesystem=host
+  #- --filesystem=host
+  - --filesystem=xdg-documents
+  - --filesystem=~/deprojects:create
   # required for input devices (gamepad) 
   - --device=all
 
--- a/ch.dragondreams.deigde.yml.in
+++ b/ch.dragondreams.deigde.yml.in
@@ -10,7 +10,9 @@ finish-args:
   - --socket=pulseaudio
   - --socket=x11
   - --share=ipc
-  - --filesystem=host
+  #- --filesystem=host
+  - --filesystem=xdg-documents
+  - --filesystem=~/deprojects:create
   # required for input devices (gamepad) 
   - --device=all
 


ch.dragondreams.deigde.yml
```diff
@@ -0,0 +1,168 @@
...
+sdk: org.freedesktop.Sdk
+command: deigde
+
+finish-args:
+  - --device=dri
+  - --share=network
+  - --socket=pulseaudio
+  - --socket=x11
+  - --share=ipc
+  - --filesystem=host
```
bbhtt: why does it need host?

LordOfDragons: Host is used because it is an editor where the user can store his files anywhere on his computer. Technically the user could be restricted to store his projects only inside his home directory to remove that permission but I'm not sure this is a nice thing to do.

bbhtt: host and home are too wide. A standard location would be better, like some xdg user dir or like ~/Projects etc.

Does it not use the portal?

LordOfDragons: It's using FOX toolkit which does not know about portals. Furthermore portals can only give permissions to a single file (IMAO) but for a game project it would be necessary to give full permission to an entire directory and all children files inside. So no, I don't think portal is an option at all.

The default directory proposed to store projects in is "~/DEProjects". This restricts users but would be fully valid as long as the editor has full permission to access any directory and files inside. The editor does not operate outside the game project base directory of the open game project.

bbhtt: no an entire directory works with portals as well.

> The default directory proposed to store projects in is "~/DEProjects"

So use `--filesystem=~/DEProjects:create` only?

LordOfDragons: Do you have some page documenting how directories can be used with portals? The flatpak portals documentation I found only describes support to request file access but nothing more: https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Documents.html .

LordOfDragons:
bot, build

ch.dragondreams.deigde.yml
bbhtt: you have to use it through a filechooser / filechooser portal

PR #7199, opened 2025-11-22
Add com.adamcake.ungoogled_cef.BaseApp

Adamcake:

- [X] Please describe the application briefly. This is a baseapp with a build of Chromium Embedded Framework (CEF), but fully source-based with the ungoogled-chromium patches applied to it, so no Google telemetry and no sketchy or outdated binaries. More detail in readme.
- [x] Please attach a video showcasing the application on Linux using the Flatpak. N/A. This is a baseapp, it doesn't do anything by itself.
- [X] I am an _(please keep whichever is applicable and remove the rest)_ developer to the project.

Additional maintainer: nmlynch94

By the way, this is going to need the large-build flag. Subsequently you can remove that flag from com.adamcake.Bolt.


hfiguiere:
what's the use case?

com.adamcake.ungoogled_cef.BaseApp.yml
```diff
@@ -0,0 +1,136 @@
...
+  org.freedesktop.Sdk.Extension.rust-stable:
+    version: *runtime-version
+    remove-after-build: true
+  org.freedesktop.Sdk.Extension.llvm21:
+    version: *runtime-version
+    remove-after-build: true
+
+modules:
+  - name: krb5
+    buildsystem: simple
```
hfiguiere, line 32: this should be `autotools`.

Adamcake:
The use-case is any software that uses libcef to open an embedded browser. Programs like electron are based on libcef, for example.

I would use it for [Bolt](https://github.com/flathub/com.adamcake.Bolt) - in fact I am already using it, but entirely in one manifest. Here I was told by bbhtt it would be better to turn it into a BaseApp: https://github.com/flathub/com.adamcake.Bolt/pull/30#issuecomment-2915493465

At least one other person has expressed interest: https://github.com/flathub/com.adamcake.Bolt/issues/29

com.adamcake.ungoogled_cef.BaseApp.yml
Adamcake: I tried that, it gives an error: `module krb5: Can't find autogen, autogen.sh or bootstrap`

hfiguiere: use the tarball like every other package.

hfiguiere: or a `shell` source to run `autoreconf`

hfiguiere:
so for one use case. Pointless.

Adamcake:
You asked me what "the use-case" is, your question was singular. There are as many use-cases as there are programs that need embedded browsers. That's how libraries work.

hfiguiere:
BaseApp are not for single use. So if the use case is one app it's pointless. Feel free to grammar check. That's not pleading the case.

Adamcake:
I didn't say it's just for me. In fact I said the opposite, I showed you that at least one other package maintainer is trying to use it for their "grayjay" package. I'd like to integrate this into other software that I use often, but one step at a time.

bbhtt:
You have put so many patches in here I have my doubts that another app other than from you will be willing to use this.

A baseapp needs to be generic enough so that multiple apps can share it.

We aren't going to have 3 different CEF baseapps.

Normally it would be published under the CEF upstream's namespace ID but due to the amount of patching that lost sense.

bbhtt:
You have put so many patches in here I have my doubts that another app other than from you will be willing to use this.

A baseapp needs to be generic enough so that multiple apps can share it.

We aren't going to have 3 different CEF baseapps.

Normally it would be published under the CEF upstream's namespace ID `io.github.chromiumembedded.cef` etc. and not yours, but due to the amount of patching that lost sense.

Adamcake:
This baseapp is a drop-in replacement for Google's official builds, so there's no issue with other CEF-based apps being willing or able to use it. The patches don't change that - all but one of them are to fix compatibility issues within the build system; some of them have been upstreamed, others aren't relevant to the upstream. The exception is the patch originating from org.chromium.Chromium which removes the setuid binary from Chromium's process sandbox, an issue which prevents official CEF builds from being able to work in flatpak at all.

As for the namespace ID, well, you can make it whatever you want, I don't mind. It doesn't seem to be an issue for `io.github.ungoogled_software.ungoogled_chromium` though.

Lastly, in your comment which I linked above, you told me:
> This is done on the condition that if in the future more apps start building libcef from source - then a baseapp needs to be created for it and bolt has to consume libcef from the baseapp. This redirection will be removed if that happens in the future.

I think I am complying exactly with that condition?

bbhtt:
Ungoogled chromium is a distinct project on GitHub and elsewhere, in contrast to this one, which extracts upstream CEF and applies patches as part of the PR.

Have you seen any other apps building their own libcef on Flathub? If so, can you open some issues and point them to this PR and ask if they want to switch to the baseapp in the future?

> I think I am complying exactly with that condition?

I didn't mean there would be app specific baseapps so a generic baseapp was the thought there even though I didn't mention it. That's the entire point of making a baseapp actually.

bbhtt:
I'm debating about the ID, if CEF upstream in the future wants to publish directly to Flathub they would obviously want to drop patches and keep only the "build fix" ones or even upstream them. So it can be problematic in the future if this one takes over the upstream namespace.

com.adamcake.ungoogled_cef.BaseApp.yml
```diff
@@ -0,0 +1,136 @@
...
+build_options:
+  # this is necessary for the build to succeed.
+  # flatpak-builder has a... bug? lovable quirk? anyway, build flags from previous builds are
+  # carried forward into subsequent builds unless you override with a non-empty set of new flags.
+  cflags: -O3
+  cflags-override: true
+  cppflags: -O3
+  cppflags-override: true
+  cxxflags: -O3
+  cxxflags-override: true
```
bbhtt, line 15: this removes all the hardening flags that come from the runtime.

generate-node-sources.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python
+
+import json
```
bbhtt, line 4: Add a licence to this if you wrote it

generate-submodule-sources.py
bbhtt, line 1: Add a licence to this if you wrote it

patch/fix-building-with-rust-1.90-or-later.patch
bbhtt, line 1: you don't build with rust nightly so is this needed?

patch/fix-wayland-and-disable-e2etests.patch
bbhtt, line 1: i see a bunch of patches just to disable tests. there are no build args to not build tests?

patch/flatpak-Add-initial-sandbox-support.patch
bbhtt, line 1: we need an ack from @refi64 as this patch introduces significant code.

README.md
```diff
@@ -0,0 +1,34 @@
+# ungoogled-cef
+This is a custom build of CEF (Chromium Embedded Framework) but with the ungoogled-chromium patchset applied to it, thus removing all the Google telemetry. Also see build.sh for a list of buildflag overrides; most features unnecessary for simple use-cases are turned off. The current version is 139.0.7258.139.
+
+Looking to use ungoogled-cef outside of flatpak? The build steps here are reproducible and in theory bootstrappable. Alternatively, I have pre-built debug and release distributions available [here](https://adamcake.com/cef).
+
+## Usage
+```yaml
+base: com.adamcake.ungoogled_cef.BaseApp
```
bbhtt, line 8: needs base-version too.

com.adamcake.ungoogled_cef.BaseApp.yml
```diff
@@ -0,0 +1,136 @@
...
+add-build-extensions:
+  org.freedesktop.Sdk.Extension.node22:
+    version: *runtime-version
+    remove-after-build: true
+  org.freedesktop.Sdk.Extension.rust-stable:
+    version: *runtime-version
+    remove-after-build: true
+  org.freedesktop.Sdk.Extension.llvm21:
+    version: *runtime-version
+    remove-after-build: true
```
bbhtt, line 28: juts use these as sdk-extensions?

bbhtt:
test builds on this should not be started. it will not work on the default runners for obvious reasons.

com.adamcake.ungoogled_cef.BaseApp.yml
nmlynch94: My understanding was this method would only include these extensions at build time, whereas sdk-extensions also include them at runtime.

bbhtt: your understanding is incorrect.

com.adamcake.ungoogled_cef.BaseApp.yml
bbhtt: you need to use `subdir: src`.

com.adamcake.ungoogled_cef.BaseApp.yml
nmlynch94: Got it. I found this option in the flatpak-manifest manpage. I just read through the docs on extensions and it expands on those options a lot more. Thanks for the correction.

patch/fix-building-with-rust-1.90-or-later.patch
Adamcake: Yes, it's needed. I first made (and named) this patch when I was building with rust nightly, but it turns out the issue was introduced somewhere between rustc 1.89 and 1.90, which in flathub's case was brought about by the update from sdk 24.08 to 25.08. There are various discussions about this issue here https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=289709

I've renamed it.

patch/fix-wayland-and-disable-e2etests.patch
Adamcake: I did read through all the buildflags and disabled all the ones that relate to tests. This one however isn't exposed as a buildflag.

If `chromium-remove-test-fonts.patch` is the other one you're referring to, then that also doesn't have a buildflag; chromium just expects all the listed fonts to be available on the system when building. We could have installed all of them in the sandbox instead of making a patch but it seemed pointless just for a test feature.

com.adamcake.ungoogled_cef.BaseApp.yml
Adamcake: The build fails without those, because CEF sets a lot of build flags for itself, which leads to duplicate definition errors of symbols like _FORTIFY_SOURCE. https://github.com/chromiumembedded/cef/blob/3479a74947aa82d049706c6e3509234bb6a6999f/cmake/cef_variables.cmake.in#L91-L192

What would you like us to do instead of this?

bbhtt: which flag is the problematic one? you can remove that and copy the rest here.

PR #7203, opened 2025-11-23
Add io.github.jmberesford.Retrom

JMBeresford:

      Retrom is a unified video game library management service for both native and emulated
      game libraries. Users can point Retrom at their existing library which will be indexed
      by Retrom. Users can also point Retrom to locally available emulator installations to
      facilitate launching their library items from the Retrom client.

https://github.com/user-attachments/assets/a1648e7b-5bc0-431f-881d-b0b0bd3756fb
      
### DIFF
--- a/io.github.jmberesford.Retrom.yml
+++ b/io.github.jmberesford.Retrom.yml
@@ -1,0 +1,103 @@
+id: io.github.jmberesford.Retrom
+
+runtime: org.gnome.Platform
+runtime-version: "49"
+sdk: org.gnome.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+  - org.freedesktop.Sdk.Extension.node22
+
+command: Retrom
+finish-args:
+  # Permissions required for launching user-configured emulators/launchers on host via flatpak-spawn
+  - --talk-name=org.freedesktop.Flatpak
+
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc # Needed for performance reasons if the app fallbacks to X11
+  - --socket=pulseaudio # Permissions required for emulator audio
+  - --device=all # Need 'input' for gamepad support, but ubuntu flatpak version is old so this is a workaround
+
+  # Permissions required for managing+launching user's game library and finding emulator executables on host
+  - --filesystem=host-os
+
+  # Permissions required for managing metadata (covers etc) from external providers, and for accessing optional remote dedicated Retrom server
+  - --share=network
+
+  - --env=FLATPAK=1
+
+build-options:
+  append-path: /usr/lib/sdk/node22/bin:/usr/lib/sdk/rust-stable/bin:/run/build/retrom/cargo/bin
+
+modules:
+  - name: retrom
+    buildsystem: simple
+    sources:
+      - type: git
+        url: https://github.com/JMBeresford/retrom
+        tag: "v0.7.51"
+        commit: "cb959286d27c99ed675d72db0129b10db0a39420"
+
+      # Static web assets
+      - type: archive
+        url: https://github.com/JMBeresford/retrom/releases/download/v0.7.51/desktop-frontend-dist_ubuntu-24.04.tar.gz
+        sha256: "1ce3862ee928933e1633b25989ac0eb658abb19e77059d08755a41619116390a"
+        strip-components: 0
+        only-arches:
+          - x86_64
+
+      - type: archive
+        url: https://github.com/JMBeresford/retrom/releases/download/v0.7.51/desktop-frontend-dist_ubuntu-24.04-arm.tar.gz
+        sha256: "f259959ca05a3c79e0df4f0c2169016bb3abcde07e4be9757654325124850ae7"
+        strip-components: 0
+        only-arches:
+          - aarch64
+
+      # Vendored cargo deps
+      - type: archive
+        url: https://github.com/JMBeresford/retrom/releases/download/v0.7.51/cargo-vendor_ubuntu-24.04.tar.gz
+        sha256: "e168116e1cfa791f86ac099ee07f33f349686af1c04e5dcdb7d860cdfd578bad"
+        strip-components: 0
+        only-arches:
+          - x86_64
+
+      - type: archive
+        url: https://github.com/JMBeresford/retrom/releases/download/v0.7.51/cargo-vendor_ubuntu-24.04-arm.tar.gz
+        sha256: "71dbc5f088f41ad07e93cd7695942c4c9e9dcb84f7bb276c7737d7b9b8a43a44"
+        strip-components: 0
+        only-arches:
+          - aarch64
+
+    build-options:
+      env:
+        CARGO_HOME: /run/build/retrom/cargo
+        XDG_CACHE_HOME: /run/build/retrom/flatpak-node/cache
+        npm_config_cache: /run/build/retrom/flatpak-node/npm-cache
+        npm_config_offline: "true"
+
+    build-commands:
+      - cargo fetch --offline
+        --config source.crates-io.replace-with=\"vendored-sources\"
+        --config source.vendored-sources.directory=\"vendor\"
+        --manifest-path packages/client/Cargo.toml
+
+      - cargo install --offline --path vendor/tauri-cli
+        --config source.crates-io.replace-with=\"vendored-sources\"
+        --config source.vendored-sources.directory=\"vendor\"
+
+      - cd packages/client && cargo --offline
+        tauri build --no-bundle --config ./tauri.build.conf.json --
+        --features "flatpak"
+        --offline
+        --config source.crates-io.replace-with=\"vendored-sources\"
+        --config source.vendored-sources.directory=\"../../vendor\"
+
+      - install -Dm644 -t /app/share/metainfo/ io.github.jmberesford.Retrom.metainfo.xml
+      - install -Dm644 -t /app/share/applications/ io.github.jmberesford.Retrom.desktop
+      - install -Dm755 -t /app/bin/ target/release/Retrom
+
+      - install -Dm644 packages/client/icons/32x32.png /app/share/icons/hicolor/32x32/apps/io.github.jmberesford.Retrom.png
+      - install -Dm644 packages/client/icons/128x128.png /app/share/icons/hicolor/128x128/apps/io.github.jmberesford.Retrom.png
+      - install -Dm644 packages/client/icons/128x128@2x.png /app/share/icons/hicolor/256x256/apps/io.github.jmberesford.Retrom.png
+      - install -Dm644 packages/client/icons/icon.png /app/share/icons/hicolor/256x256@2/apps/io.github.jmberesford.Retrom.png
+      - install -Dm644 LICENSE /app/share/licenses/io.github.jmberesford.Retrom/LICENSE


JMBeresford:
PR for adding linter exceptions: https://github.com/flathub-infra/flatpak-builder-lint/pull/767

JMBeresford:
I would also appreciate any guidance on how best to automate PRs for updating the flatpak once submitted. Retrom currently has fully automated releases. If I update the CI to build out a new manifest to reflect the update is the simplest way to manually create a PR in the CI action with the new manifest? I would very much like to keep the manifest maintained in the Retrom monorepo.

io.github.jmberesford.Retrom.yml
```diff
@@ -0,0 +1,99 @@
+id: io.github.jmberesford.Retrom
+
+runtime: org.gnome.Platform
+runtime-version: "48"
```
hfiguiere: current is 49

io.github.jmberesford.Retrom.yml
```diff
@@ -0,0 +1,99 @@
...
+sdk: org.gnome.Sdk//48
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+  - org.freedesktop.Sdk.Extension.node22
+
+command: Retrom
+finish-args:
+  - --socket=session-bus # Permissions required for launching user-configured emulators/launchers on host
```
hfiguiere: this is not allowed. I need to explicitely request the bus names you need.

io.github.jmberesford.Retrom.yml
```diff
@@ -0,0 +1,99 @@
...
+  - --socket=wayland # Permission needed to show the window on wayland
+  - --socket=fallback-x11 # Permission needed to fallback using X11 if wayland is not available
+  - --share=ipc # Needed for performance reasons if the app fallbacks to X11
+  - --socket=pulseaudio # Permissions required for emulator audio
+  - --device=dri # OpenGL/webGL used in web-view for built-in WASM emulators
+  - --device=all # Need 'input' for gamepad support, but ubuntu flatpak version is old so this is a workaround
```
hfiguiere:
```suggestion
  - --device=all # Need 'input' for gamepad support, but ubuntu flatpak version is old so this is a workaround
```

dri is included.

io.github.jmberesford.Retrom.yml
```diff
@@ -0,0 +1,99 @@
...
+        url: https://github.com/JMBeresford/retrom
+        tag: init-flatpak
+
+      # Static web assets
+      - type: archive
+        url: https://github.com/JMBeresford/retrom/releases/download/v0.7.43/desktop-frontend-dist_ubuntu-24.04.tar.gz
+        sha256: ed27a24e55a78be187a3d07d4fd074f9ab1560eef4e72aba7e5aa50e5e33a33e
+        strip-components: 0
+        only-arches:
+          - x86_64
```
hfiguiere, line 47: so what happen on `aarch64` ?

io.github.jmberesford.Retrom.yml
```diff
@@ -0,0 +1,99 @@
...
+
+      - type: archive
+        url: https://github.com/JMBeresford/retrom/releases/download/v0.7.43/desktop-frontend-dist_ubuntu-24.04-arm.tar.gz
```
hfiguiere: `arm` in the name but `only-arches:
          - x86_64` ?

io.github.jmberesford.Retrom.yml
```diff
@@ -0,0 +1,99 @@
...
+      # Vendored cargo deps
+      - type: archive
+        url: https://github.com/JMBeresford/retrom/releases/download/v0.7.43/cargo-vendor_ubuntu-24.04.tar.gz
+        sha256: 4e1f3fa48a4a22429fd58c0813c64d57ae8f9097c759675d087af8a860fbac14
+        strip-components: 0
+        only-arches:
+          - x86_64
+
+      - type: archive
+        url: https://github.com/JMBeresford/retrom/releases/download/v0.7.43/cargo-vendor_ubuntu-24.04-arm.tar.gz
```
hfiguiere: why would the vendored rust crates be architecture specific?

io.github.jmberesford.Retrom.yml
```diff
@@ -0,0 +1,99 @@
...
+      - install -Dm644 -t /app/share/metainfo/ io.github.jmberesford.Retrom.metainfo.xml
+      - install -Dm644 -t /app/share/applications/ io.github.jmberesford.Retrom.desktop
+      - install -Dm755 -t /app/bin/ target/release/Retrom
+
+      - install -Dm644 packages/client/icons/32x32.png /app/share/icons/hicolor/32x32/apps/io.github.jmberesford.Retrom.png
+      - install -Dm644 packages/client/icons/128x128.png /app/share/icons/hicolor/128x128/apps/io.github.jmberesford.Retrom.png
+      - install -Dm644 packages/client/icons/128x128@2x.png /app/share/icons/hicolor/256x256/apps/io.github.jmberesford.Retrom.png
+      - install -Dm644 packages/client/icons/icon.png /app/share/icons/hicolor/256x256@2/apps/io.github.jmberesford.Retrom.png
+
+    post-install:
```
hfiguiere:
```suggestion
```

io.github.jmberesford.Retrom.yml
```diff
@@ -0,0 +1,99 @@
+id: io.github.jmberesford.Retrom
+
+runtime: org.gnome.Platform
+runtime-version: "48"
+sdk: org.gnome.Sdk//48
```
hfiguiere:
```suggestion
sdk: org.gnome.Sdk
```

io.github.jmberesford.Retrom.yml
```diff
@@ -0,0 +1,99 @@
...
+build-options:
+  append-path: /usr/lib/sdk/node22/bin:/usr/lib/sdk/rust-stable/bin:/run/build/retrom/cargo/bin
+
+modules:
+  - name: retrom
+    buildsystem: simple
+    sources:
+      - type: git
+        url: https://github.com/JMBeresford/retrom
+        tag: init-flatpak
```
bbhtt: this is not a versioned tag and pin it to a commit or just use a commit only.

hfiguiere:
```suggestion
        tag: init-flatpak
        commit: 93e3726dd59dfc6093dc62aeb676b937b6a94008
```

io.github.jmberesford.Retrom.yml
hfiguiere: also https://github.com/flathub/flathub/pull/7203/files#r2553780853

io.github.jmberesford.Retrom.yml
JMBeresford: Updated to require only the ability to spawn on host

io.github.jmberesford.Retrom.yml
JMBeresford: The vendored crates are sourced from the build pipeline on release, which is run for each platform+architecture that Retrom supports (inclusive of windows and macOS). The vendored crates may be identical, but since they exist for both architectures it makes sense to me to be explicit here -- also to guard against some future **need** for adopting this pattern.

JMBeresford:
Updated this PR to address feedback, along with the PR for the lint exceptions linked above.

@bbhtt @hfiguiere  Would appreciate another look when either of you get the chance -- thanks for your time!

JMBeresford:
bot, build

JMBeresford:
bot, build

JMBeresford:
@bbhtt @hfiguiere Build passed, and test build confirmed working as expected -- thanks much for your patience and guidance! Anything else needed from me here?

bbhtt:
Launching from desktop file is broken:

```
io.github.jmberesford.Retrom.desktop[4284]: bwrap: execvp retrom: No such file or directory
```

and I can't start in standalone mode:

https://github.com/user-attachments/assets/709de78c-4b7e-4a93-a33d-9e2f27d21310

hfiguiere:
> Launching from desktop file is broken:

There is that warning in the build log no one reads:

image

Not sure if we could make this an error.

JMBeresford:
bot, build

JMBeresford:
@bbhtt Standalone mode is not currently supported in the Flatpak build -- there was a bug in showing the error notification however.

Both that and the desktop file issue are now resolved.


https://github.com/user-attachments/assets/d049743d-811d-49cd-a069-0ef057e52389

bbhtt:
So then you should hide that option if it is running inside Flatpak. And why is it not supported?

JMBeresford:
> So then you should hide that option if it is running inside Flatpak

@bbhtt Noted, agreed, and rest assured there are plans to improve this. **I assume this application-specific UX nit is not a blocker for this PR**, but will elaborate nonetheless:

This is currently non-trivial. The static web assets (i.e. the UI) is built in CI on release, and the resulting tarball is sourced during the flatpak build. In order for this part of the UI to be aware that it is in a flatpak environment it needs to communicate over the desktop layer's IPC. This is how the error/notification is sourced. This can of course be addressed, but I would really prefer not to block this PR on application-specific UX nits that can be iterated on later. **Each cycle of this review process takes orders of magnitude longer to iterate than any such development/UX fixes.** Adding a new IPC handler just for this, or spinning up flatpak-specific builds of the web assets will take time that would be better spent simply closing the gaps for standalone-mode support in the flatpak sandbox.

> And why is it not supported

It relies on a locally managed `postgres` service + DB. This includes installation+initialization of the DB. This process (which I plan to replace with a more holistic solution in the future) was already shaky at best w/o the added constraints/expectations of the flatpak sandbox. Similar to the above, I would prefer to spend the time it would take to get this support hacked in on a more robust solution for all environments.

---

**Note that standalone mode is a *secondary* use-case for the Retrom client. The primary use-case is to leverage a remote, dedicated Retrom server.**

> "Why then is the secondary use case presented as the primary one in the setup screen"?

Because the users that want to use standalone mode are the users that want it to 'just work' w/o any fuss. They get confused the second any kind of server/client relationship enters the equation -- even as a yes/no question. The users that have a dedicated server are savvy enough to understand they want to use the 'advanced' option.

Please let me know if there are any functional blockers here, thanks!

JMBeresford:
@bbhtt @hfiguiere any remaining feedback to address here?

bbhtt:
The previous issue I mentioned should be fixed first.

JMBeresford:
bot, build

JMBeresford:
@bbhtt Done, the mode selection step is now disabled in flatpak clients.

https://github.com/user-attachments/assets/f00bcedc-1ef9-4609-b620-8f2fe0ead76f

bbhtt:
how is this supposed to be used? where is the "retrom server"? If i click on localhost 5001 I get an error with "try again"

JMBeresford:
@bbhtt The client expects a dedicated Retrom server. This flatpak is _just_ the client. This was explained in my message above while explaining that standalone mode is a secondary use-case.

JMBeresford:
If inclined, you can spin up a server locally via docker fairly easily by following the docs: https://github.com/JMBeresford/retrom/wiki/Quick-Start#using-a-dedicated-server

bbhtt:
Is there a server I can test it with? This seems complicated just to test with as in the Flatpak is not self sufficient.

Can you mention in the metainfo that it requires a server or setting up one?

JMBeresford:
@bbhtt I've temporarily spun up a read-only instance locally that you can use:

`https://retrom-dev.volcaus.com` (use port 443 in the Retrom client)

Please let me know once you have finished testing so I can tear it down.

JMBeresford:
> Can you mention in the metainfo that it requires a server or setting up one?

Added

JMBeresford:
bot, build

bbhtt:
Yes i'm done using that server.

PR #7207, opened 2025-11-23
Add org.kde.hana

g-fb:

- [X] Please describe the application briefly. < Please insert the description here >
    Application to create thumbnail images from video files and join them in one image
    [hana showcase.webm](https://github.com/user-attachments/assets/70716e75-8133-4e8b-a0c3-0f0936aedd22)

### DIFF
--- a/org.kde.hana.yml
+++ b/org.kde.hana.yml
@@ -1,0 +1,23 @@
+# SPDX-FileCopyrightText: 2025 George Florea Bănuș <georgefb899@gmail.com>
+# SPDX-License-Identifier: CC0-1.0
+app-id: org.kde.hana
+runtime: org.kde.Platform
+runtime-version: 6.10
+sdk: org.kde.Sdk
+command: hana
+rename-icon: hana
+copy-icon: true
+finish-args:
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --device=dri
+  - --filesystem=home
+  - --talk-name=org.freedesktop.FileManager1
+modules:
+  - name: hana
+    buildsystem: cmake-ninja
+    sources:
+      - type: archive
+        url: https://download.kde.org/stable/hana/1.0.0/hana-1.0.0.tar.xz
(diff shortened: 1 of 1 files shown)


hfiguiere:
cc @flathub/kde

org.kde.hana.yml
```diff
@@ -0,0 +1,25 @@
...
+sdk: org.kde.Sdk
+command: hana
+rename-icon: hana
+copy-icon: true
+finish-args:
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --device=dri
+  - --filesystem=home
```
bbhtt, line 15: home is essentially sandbox escape. can this be limited to a few commom folders like xdg-pictures? or better go through the filechooser portal?

g-fb: > can this be limited to a few common folders like xdg-pictures?

The app saves images in the same folder as the video or in a user defined folder, so xdg-pictures is not enough.

> or better go through the file chooser portal?

Wouldn't that require the user to select an output folder each time they use the app?

bbhtt: > Wouldn't that require the user to select an output folder each time they use the app?

If it uses a separate folder each time, then yes, but a single folder once selected through the filechooser portal should be persistently made available by Flatpak.

usually these use a single location to save the files.

g-fb: > a single folder once selected through the filechooser portal should be persistently made available by Flatpak

That works for "saving in an user defined folder", but not for "saving in the same folder as the video", which is the default.

g-fb:
bot, build

org.kde.hana.yml
```diff
@@ -0,0 +1,25 @@
...
+  - --talk-name=org.freedesktop.FileManager1
+modules:
+  - name: hana
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
```
travier: Either RelWithDebInfo or nothing here.

tulilirockz: The buildsystem itself seems to take care of this..? I think

hfiguiere:
bot, build

PR #7334, opened 2025-12-16
Add eu.binjr.binjr

fthevenet:

> binjr is a time series browser; it renders time series produced by other applications as dynamically
            editable graphical representations and provides advanced features to navigate the data smoothly and efficiently
            (drag &amp; drop, zoom, history, detachable tabs, advanced time-range picker).
[Screencast_20251216_234441.webm](https://github.com/user-attachments/assets/ec1464c4-5fcb-4426-8d0a-9ebadf0e5242)

### DIFF
--- a/eu.binjr.binjr.yml
+++ b/eu.binjr.binjr.yml
@@ -11,7 +11,6 @@ finish-args:
   - --device=dri
   - --share=network
   - --filesystem=host
-  - --talk-name=org.gtk.vfs.*
 modules:
   - name: binjr
     buildsystem: simple


fthevenet:
Hi,
So the test build fails because I had to declare `--filesystem=host` as a finish arg in the manifest for the app to work as expected, which is to let users load and save files from any location they require via a file chooser dialog. 
My understanding is that ideally an application based on GTK3 should be able to use the XDG Desktop Portal API transparently (?)  to limit FS access only to the files/folders users picked using the FileChooser, but in the case of my app, which is written in Java, this does not seem to work.

While I understand how this is far from ideal and not consequence free, is this considered a blocker? From what I can see, there are many other applications on flathub on flathub that had to resort to giving very broad fs access.

Thanks!

eu.binjr.binjr.yml
```diff
@@ -0,0 +1,42 @@
...
+      - install -Dm644 icons/binjr_256.png /app/share/icons/hicolor/256x256/apps/eu.binjr.binjr.png
+      - install -Dm644 icons/binjr_512.png /app/share/icons/hicolor/512x512/apps/eu.binjr.binjr.png
+      - install -Dm644 metadata/eu.binjr.binjr.metainfo.xml /app/share/metainfo/eu.binjr.binjr.metainfo.xml
+      - install -Dm644 metadata/eu.binjr.binjr.desktop -t /app/share/applications
+    sources:
+      - type: git
+        url: https://github.com/binjr/binjr-flatpak-resources.git
+        tag: v3.26.2
+      - type: archive
+        url: https://github.com/binjr/binjr/releases/download/v3.26.2/binjr-3.26.2_linux-aarch64.rpm
```
bbhtt: this needs to be built from source. You are packaging for Flatpak, not for a distro.

eu.binjr.binjr.yml
```diff
@@ -0,0 +1,42 @@
...
+      - cp -ar opt/binjr/* /app/
+      - install -Dm644 icons/binjr_16.png /app/share/icons/hicolor/16x16/apps/eu.binjr.binjr.png
+      - install -Dm644 icons/binjr_32.png /app/share/icons/hicolor/32x32/apps/eu.binjr.binjr.png
+      - install -Dm644 icons/binjr_48.png /app/share/icons/hicolor/48x48/apps/eu.binjr.binjr.png
+      - install -Dm644 icons/binjr_64.png /app/share/icons/hicolor/64x64/apps/eu.binjr.binjr.png
+      - install -Dm644 icons/binjr_128.png /app/share/icons/hicolor/128x128/apps/eu.binjr.binjr.png
+      - install -Dm644 icons/binjr_256.png /app/share/icons/hicolor/256x256/apps/eu.binjr.binjr.png
+      - install -Dm644 icons/binjr_512.png /app/share/icons/hicolor/512x512/apps/eu.binjr.binjr.png
+      - install -Dm644 metadata/eu.binjr.binjr.metainfo.xml /app/share/metainfo/eu.binjr.binjr.metainfo.xml
+      - install -Dm644 metadata/eu.binjr.binjr.desktop -t /app/share/applications
```
hfiguiere: also why would any of these files not be in the rpm (which mean when building from source they would be there too)? There is nothing flatpak specific.

fthevenet: That's a good point; I just didn't think of this.

eu.binjr.binjr.yml
fthevenet: I understood that there was not network access during the build phase in the flathub build pipeline, which would make my gradle-based build (which pulls dependencies from the [Maven Central Repository](https://central.sonatype.com/)) quite complicated, so I thought it simpler to reuse a pre-built binary from one of the archive format my own build system produces (I picked the rpm, because it was the one needing the least modification, but I could have picked a plain tarball).

I saw plenty of existing Flathub publication doing the same thing (i.e. building by repackaging an archive) when looking for examples, so I figured this was a legitimate way of doing it.

eu.binjr.binjr.yml
fthevenet: Hmm.. scratch that. The rpm installs the bits into `/opt/binjr`, which means the `.desktop` file needs absolute paths for the launcher and icon (i.e `/opt/binjr/bin/binjr`), which won't work for the flatpak.
It's simpler to just keep the two separate, imo.

fthevenet:
Okay so I have spent a fair amount of time researching this and building from source would indeed require bending my build system backward to feed it all the dependencies it needs upfront (including the build tool itself, and likely the JDK too). 
There are scripts out there that could help, but they're all admittedly brittle and limited; getting this to work is going to be a lot of work on my part, and given that I do all this on my spare time, I'm not sure this is how I want to spent my holiday break. Especially since -- let's be honest -- front loading all dependencies and too chain as binary blobs to recompile what amount to about 10-30% of the final delivered bits does sound a lot like "package from a tarball with extra steps"...

Before I decide what to do next, I would please like  to get an answer to these questions:
* Is it or is it not allowed for my app to be sourced from an archive produced by my own build system? (it's all public, using github actions; a clear audit trail from the actual build logs to the Flathub package is easy enough for end users to establish if need be).
* Is it preferable that I extend my own build to produce a single-file bundle and submit this for publication?

Both this options would be far preferable from my end, so I need to know whether or not they are acceptable from the perspective of Flathub's maintainer before I decide to what to do.

Thanks for your time!

eu.binjr.binjr.yml
fthevenet: I'll work on building this from source.
I've converted the PR to draft while I work on that, and will request a review once it's ready.
Thanks

fthevenet:
bot, build

fthevenet:
bot, build

fthevenet:
bot, build

fthevenet:
bot, build

fthevenet:
bot, build

fthevenet:
bot, build

fthevenet:
bot, build

fthevenet:
I've completely reworked the original PR, so that now it is fully built from source.
This should now be ready for review.
Thanks for your time!

eu.binjr.binjr.yml
```diff
@@ -0,0 +1,51 @@
...
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.openjdk25
+command: binjr
+finish-args:
+  - --socket=x11
+  - --share=ipc
+  - --device=dri
+  - --share=network
+  - --filesystem=host
+  - --talk-name=org.gtk.vfs.*
```
bbhtt: what is this for?

fthevenet:
```
--device=dri
--share=network
--filesystem=host
```
Application uses HW accelerated rendering (though OpenGL), needs to access the network and  file system (underlying framework does not support XDG portals).

```
--talk-name=org.gtk.vfs.*
```
The (flatpak.org documentation)[https://docs.flatpak.org/en/latest/sandbox-permissions.html] states the following, which had me believe this is required to access USB and network shares:
> Typical GNOME and GTK applications should use:
> --talk-name=org.gtk.vfs.*
> --filesystem=xdg-run/gvfsd

While the app is definitely not a "typical GNOME application" it does use GTK. But to be fair, I am unsure wether or not this is needed.

fthevenet:
bot, build

eu.binjr.binjr.yml
fthevenet: I tested without `--talk-name=org.gtk.vfs.*` and it doesn't seem to be needed, so I removed it.

PR #7342, opened 2025-12-18
Add org.kde.kjournaldbrowser

cordlandwehr:

This PR is  related to a required linter exception: https://github.com/flathub-infra/flatpak-builder-lint/pull/798

- [X] Please describe the application briefly. < Please insert the description here >
KJournaldbrowser is a log viewer application for systemd's journald journal. It provides different filters and graphical highlights. The application is released as part of the KDE application releases.
- [X] Please attach a video showcasing the application on Linux using the Flatpak.  [kjournald_20251218_090912.webm](https://github.com/user-attachments/assets/0ea1f301-4a86-40d7-9880-51d50c3db68a)

### DIFF
--- a/org.kde.kjournaldbrowser.yml
+++ b/org.kde.kjournaldbrowser.yml
@@ -1,0 +1,27 @@
+id: org.kde.kjournaldbrowser
+runtime: org.kde.Platform
+runtime-version: '6.10'
+sdk: org.kde.Sdk
+command: kjournaldbrowser
+finish-args:
+  # X11
+  - --share=ipc
+  # Windowing
+  - --socket=fallback-x11
+  - --socket=wayland
+  # Hardware Acceleration
+  - --device=dri
+  # Browse the system journal
+  - --filesystem=/var/log/journal:ro
+modules:
+  - name: kjournald
+    buildsystem: cmake-ninja
+    sources:
+      - type: archive
+        url: https://download.kde.org/stable/release-service/26.04.3/src/kjournald-26.04.3.tar.xz
+        sha256: 6a97ac76c006dcd0ccdb33933ab71c611b79cc9ae20666deb8074a8ac43993ce
+        x-checker-data:
+          type: anitya
+          project-id: 8763
+          stable-only: true
+          url-template: https://download.kde.org/stable/release-service/$version/src/kjournald-$version.tar.xz


hfiguiere:
cc @flathub/kde

org.kde.kjournaldbrowser.yml
```diff
@@ -0,0 +1,31 @@
...
+        sha256: be88712aa1d07fcd6519fa9cf70141c80b26c9b2ddcb311b7f8175a098c95e09
+        x-checker-data:
+          type: anitya
+          project-id: 8763
+          stable-only: true
+          url-template: https://download.kde.org/stable/release-service/$version/src/kjournald-$version.tar.xz
+    post-install:
+      # Icon is included with breeze in runtime
+      - install -D /usr/share/icons/breeze/apps/32/utilities-log-viewer.svg ${FLATPAK_DEST}/share/icons/hicolor/scalable/apps/${FLATPAK_ID}.svg
+      - desktop-file-edit --set-key=Icon --set-value=${FLATPAK_ID} ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop
```
bbhtt: why is this not being done upstream?

cordlandwehr: Good point, I will create a MR to fix this directly in breeze. though it will take about a month for a next release to have it fixed there.
So, I can either remove it now here and icon is missing or prepare a PR after a next breeze release that provides the icon with correct ID.

cordlandwehr: Short update: the proper fix for the icon alias is now merged in breeze-icons and will be part of its next release in about 4 weeks.

org.kde.kjournaldbrowser.yml
```diff
@@ -0,0 +1,31 @@
...
+finish-args:
+  # X11
+  - --share=ipc
+  # Windowing
+  - --socket=fallback-x11
+  - --socket=wayland
+  # Hardware Acceleration
+  - --device=dri
+  # Browse the system journal
+  - --filesystem=/var/log/journal:ro
```
travier, line 16: This will likely only allow browsing the user journal and not the system one as the Flatpak does not keep the groups permissions from the user (i.e. `wheel`|`sudo` in this case).

travier:
Will need an exception for `/var` access or more realistically for full sandbox escape to truly be able to read the journal from the host.

org.kde.kjournaldbrowser.yml
cordlandwehr: Accessing the system journal requires the user to be in the systemd-journal group in order to access the datbase files in /var/log/journal/. On my tests this worked such that the test user is a member of that group and when running org.kde.kjournaldbrowser, also access to the system journal is possible.

cordlandwehr:
Could you elaborate why you expect access to full /var being needed?
With respect to the journald docs (https://www.freedesktop.org/software/systemd/man/latest/journald.conf.html) we can expect journald log data at /var/log/journal or (during early boot) at /run/log/journal, before a first --flush operation is performed. The early boot situation is negligible IMO because we already have the system booted into a state where eg. flatpak runner is available. So from the pure file system location point of view /var/log/journal should suffice

Justinzobel:
Ping to see where this is up to.

hfiguiere:
bot, build

org.kde.kjournaldbrowser.yml
hfiguiere: is this fix in the current runtime?

cordlandwehr: yes, the workaround is not needed anymore.

Justinzobel:
bot, build

Justinzobel:
```
Error: 'desktop-file-icon-key-wrong-value' error found in linter repo check. Details: Icon key in desktop file has wrong value: utilities-log-viewer
Error: 'appstream-missing-icon-file' error found in linter repo check. Details: No icon was generated by appstream. An icon needs to be installed. Allowed: 128, 256, 512px PNG or SVG
Error: 'no-exportable-icon-installed' error found in linter repo check. Details: No PNG or SVG icons named by org.kde.kjournaldbrowser were found in $FLATPAK_DEST/share/icons/hicolor/$size/apps or $FLATPAK_DEST/share/icons/hicolor/scalable/apps
Warning: 'runtime-update-available-to-org.kde.Platform-6.11' warning found in linter repo check. Details: Please consider updating to the latest runtime version
```

hfiguiere:
probably needs a `rename-icon: utilities-log-viewer`

Justinzobel:
> probably needs a `rename-icon: utilities-log-viewer`

@cordlandwehr are you able to test to see if this fixes it?

cordlandwehr:
Sure, but will take a few weeks. Currently I am an vacations with no laptop access. Will do it, when I am back.

cordlandwehr:
bot, build

cordlandwehr:
bot, build

cordlandwehr:
bot, build

org.kde.kjournaldbrowser.yml
```diff
@@ -0,0 +1,35 @@
...
+          type: anitya
+          project-id: 8763
+          stable-only: true
+          url-template: https://download.kde.org/stable/release-service/$version/src/kjournald-$version.tar.xz
+      - type: patch
+        # cherry picked from upstream LTS branch, will be part of release 26.08.2
+        path: 0001-Rename-desktop-file-icon-to-match-appstream.patch
+      - type: patch
+        # cherry picked from upstream LTS branch, will be part of release 26.08.2
+        path: 0002-Rename-application-icon-to-app-specific-name.patch
```
CodedOre: Are these patches still required? Other than that, I think this could also be done with the `rename-icon` manifest option, so I think having patches is a bit overkill.

cordlandwehr: I think I can get this solved without adding these patches by using rename-icon and then even drop rename-icon with the next patch release 26.08.02.

cordlandwehr:
bot, build

PR #7429, opened 2025-12-29
Add org.glabels.glabels-qt

p0358:
**WIP draft** for now, until the 4.0 is officially released as non-beta, for now to test builds/linting on Flathub infra until then.

  gLabels is a program for designing and printing various labels and business cards.
- [ ] Please attach a video showcasing the application on Linux using the Flatpak.
  < TODO >
      If not, I contacted upstream developers about this submission. **Link:** https://github.com/j-evins/glabels-qt/pull/255

### 💡 Please mention below the GitHub usernames of any additional maintainers needed (if any) 💡
- @j-evins (main upstream dev)
- @p0358 (me)

### TODO:
- video
- screenshots and version/release info in metafile
- release tag on upstream repo to pin to

### Notes:
- `--filesystem=~/.glabels:ro` is for `Directory for manually created product templates.` (users can put in template files there manually for the app to pick up; the app saves new templates created by users with the app to regular data folders on the other hand)

### DIFF
--- a/org.glabels.glabels-qt.yml
+++ b/org.glabels.glabels-qt.yml
@@ -12,7 +12,7 @@ finish-args:
   - --filesystem=~/.glabels:ro
 modules:
   - name: glabels-qt
-    buildsystem: cmake
+    buildsystem: cmake-ninja
     sources:
       - type: git
         url: https://github.com/j-evins/glabels-qt.git
--- a/org.glabels.glabels-qt.yml
+++ b/org.glabels.glabels-qt.yml
@@ -1,0 +1,20 @@
+id: org.glabels.glabels-qt
+runtime: org.kde.Platform
+runtime-version: '6.10'
+sdk: org.kde.Sdk
+command: glabels-qt
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc
+  - --device=dri
+  - --socket=cups
+  - --filesystem=~/.glabels:ro
+modules:
+  - name: glabels-qt
+    buildsystem: cmake
+    sources:
+      - type: git
+        url: https://github.com/j-evins/glabels-qt.git
+        #tag: 3.99-master602
+        commit: 82b264a7a057d8801577f49f9b46e33411f88ae5
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+/.flatpak-builder
+/repo
+/builddir


p0358:
bot, build

org.glabels.glabels-qt.yml
```diff
@@ -0,0 +1,20 @@
...
+runtime-version: '6.10'
+sdk: org.kde.Sdk
+command: glabels-qt
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc
+  - --device=dri
+  - --socket=cups
+  - --filesystem=~/.glabels:ro
```
hfiguiere, line 12: this should be `--persist` or use the proper XDG directories

org.glabels.glabels-qt.yml
```diff
@@ -0,0 +1,20 @@
...
+modules:
+  - name: glabels-qt
+    buildsystem: cmake
```
hfiguiere:
```suggestion
    buildsystem: cmake-ninja
```

org.glabels.glabels-qt.yml
p0358: The app does normally use proper XDG directories for its regular configuration, its main config ends up in `~/.var/app/org.glabels.glabels-qt/config/glabels.org/glabels-qt.conf` as expected (analog. of `~/.config/glabels.org/glabels-qt.conf` outside of Flatpak).

The `~/.glabels` on the other hand is meant to be *optional* and for the **user** to manually put predefined template files in there if desired. It's read-only, since the app doesn't put any files there itself, only reads if it exists.

I presume if we use `--persist` that it'd map `~/.glabels` from app's perspective into some data directory that'd have to be mentioned in upstream docs in case of Flatpak edition, right? I'm debating whether this would be better, or asking at upstream to support some other XDG path for that purpose too, not sure which'd be more intuitive to the users... (assuming `--filesystem=~/.glabels:ro` from actual home dir would not be acceptable in such scenario)

p0358:
Just so it doesn't get lost somewhere, there's also a matter of display name conflict with with the old `gLabels` at `org.gnome.glabels-3`, that I'll copy-paste my question comment about here too, so it doesn't get lost (from https://github.com/flathub/org.gnome.glabels-3/issues/12#issuecomment-3697475431):

> speaking of the new app, if I might ask here already: what would be the suggested/preferred naming for two apps once the new one is submitted? Currently they're both named just `gLabels` in the metafile.
> 
> The old one is unmaintained, so the new one will effectively "replace" it, but under a different ID. I agree the old one should stay as a separate app, mostly because the old app's files aren't being read 1:1 properly by the new one, so it wouldn't be nice to auto-migrate users of the old app.
> 
> But should they be renamed to `gLabels 3` and `gLabels 4`, or maybe something like `gLabels GTK` or so for old one and have (prettier) bare `gLabels` be taken by the current one, or something else?

PR #7474, opened 2026-01-03
Add it.astrogods.AstroGods

arcangelo7:

- [X] AstroGods is a modern, beautifully designed Astrology app focused on self-knowledge. It provides personalized birth chart interpretations, synastry analysis for relationships, and transit readings. The app is designed to make Astrology accessible as a tool for self-reflection and exploration. It uses the Swiss Ephemeris for precise astronomical calculations and follows the Egyptian Astrology methodology with equal-area houses (Vehlow).
- [X] A video showcasing the application on Linux using the Flatpak:

https://github.com/user-attachments/assets/27aaa5fc-46b9-4fa5-90dc-c4479c725b5e

### DIFF
--- a/it.astrogods.AstroGods.yml
+++ b/it.astrogods.AstroGods.yml
@@ -10,9 +10,9 @@ command: astrogods
 finish-args:
   - --share=ipc
   - --share=network
-  - --socket=x11
+  - --socket=wayland
+  - --socket=fallback-x11
   - --device=dri
-  - --env=GDK_BACKEND=x11
 modules:
   - name: astrogods
     buildsystem: simple


it.astrogods.app.json
```diff
@@ -0,0 +1,39 @@
+{
+  "app-id": "it.astrogods.app",
```
hfiguiere: `app` is NOT a proper application name for the purpose of the app id.

it.astrogods.app.json
```diff
@@ -0,0 +1,39 @@
...
+        "mkdir -p /app/bin",
+        "ln -s /app/astrogods/astrogods /app/bin/astrogods",
+        "install -Dm644 share/applications/it.astrogods.app.desktop /app/share/applications/it.astrogods.app.desktop",
+        "install -Dm644 share/metainfo/it.astrogods.app.metainfo.xml /app/share/metainfo/it.astrogods.app.metainfo.xml",
+        "cp -r share/icons /app/share/"
+      ],
+      "sources": [
+        {
+          "type": "archive",
+          "url": "https://github.com/arcangelo7/astrogods-linux/releases/download/4.4.0/linux-x64-4.4.0.tar.gz",
```
hfiguiere: This should be built from source.

hfiguiere: What happen on aarch64?

it.astrogods.app.json
```diff
@@ -0,0 +1,39 @@
+{
+  "app-id": "it.astrogods.app",
+  "runtime": "org.gnome.Platform",
+  "runtime-version": "48",
```
hfiguiere: current is 49

why does this need the GNOME runtime?

it.astrogods.app.json
arcangelo7: I'm sorry, I missed that line in the requirements. I've renamed the app ID to `it.astrogods.AstroGods`.

it.astrogods.app.json
arcangelo7: > This should be built from source.
> 
> https://docs.flathub.org/docs/for-app-authors/requirements#building-from-source

This is a closed-source, proprietary app. Building from source isn't possible since that would require making the source code public. According to Flathub requirements (https://docs.flathub.org/docs/for-app-authors/requirements), proprietary apps can be distributed through tarballs hosted on well-known hosts, which is what I'm doing with GitHub releases.

arcangelo7: > What happen on aarch64?

This app is built with Flutter. Theoretically, Flutter does support compiling for ARM (https://docs.flutter.dev/reference/supported-platforms), but I don't have an ARM device to test on at the moment. I've explicitly excluded aarch64 in the manifest for now with `"only-arches": ["x86_64"]`. I may add ARM support in the future once I can properly test it.

it.astrogods.app.json
arcangelo7: I chose version 48 thinking it was more stable than 49. As for the runtime, this is a Flutter app which requires GTK3 (see https://docs.flutter.dev/platform-integration/linux/building), so I went with GNOME Platform. I didn't know GTK3 was included in Freedesktop Platform as well (https://flatpak-docs.readthedocs.io/en/latest/available-runtimes.html). Fixed both, thanks for pointing this out.

it.astrogods.app.json
hfiguiere: that doesn't prevent an aarch64 package to be built.

it.astrogods.app.json
hfiguiere: > I chose version 48 thinking it was more stable than 49. 

We does the doc suggest this? So we can fix it.

arcangelo7: > > I chose version 48 thinking it was more stable than 49.
> 
> We does the doc suggest this? So we can fix it.

It wasn't from the docs, just my assumption that an older, established version would be more stable and less likely to have issues. No fix needed on your side!

it.astrogods.app.json
arcangelo7: > that doesn't prevent an aarch64 package to be built.

You're right, I misunderstood how `only-arches` works. I've now added a `flathub.json` file with `"only-arches": ["x86_64"]` to exclude aarch64 (https://docs.flathub.org/docs/for-app-authors/maintenance#limiting-the-set-of-architectures-to-build-on).

it.astrogods.AstroGods.json
```diff
@@ -0,0 +1,39 @@
...
+  "runtime": "org.freedesktop.Platform",
+  "runtime-version": "25.08",
+  "sdk": "org.freedesktop.Sdk",
+  "command": "astrogods",
+  "finish-args": [
+    "--share=ipc",
+    "--share=network",
+    "--socket=x11",
+    "--device=dri",
+    "--env=GDK_BACKEND=x11"
```
bbhtt: don't set this

it.astrogods.app.json
bbhtt: the question here is can you provide aarch64 builds?

arcangelo7: > the question here is can you provide aarch64 builds?

Yes, I can now provide aarch64 builds. To make this happen, I've open-sourced the app's frontend at https://github.com/arcangelo7/astrogods-app. This felt like the right move anyway, and it gives you full source transparency. There was also a practical reason: GitHub only offers free ARM runners for public repositories, so going open source was necessary to set up proper CI verification on both architectures. This also means the submission has changed significantly: the package is now built from source rather than precompiled binaries.

it.astrogods.AstroGods.json
arcangelo7: If I understand correctly, you're asking me to remove only the last line (--env=GDK_BACKEND=x11). I intentionally set it to force X11 because Flutter on Linux still has several documented issues when running on native Wayland. Running through XWayland provides a way more stable experience for now. That said, if this is a requirement for Flathub acceptance, I can remove it.

bbhtt: This is a debug env var that will effect unrelated things not just your app. If Flutter requires this then it is broken and should not be used. There are better toolkits out there with native wayland support.

bbhtt:
bot, build

it.astrogods.AstroGods.json
arcangelo7: I've updated the manifest to use Wayland. The GDK_BACKEND=x11 was added as a precaution, but Flutter has supported Wayland since [PR #66519](https://github.com/flutter/flutter/pull/66519). I've tested the app locally on Wayland and it works perfectly.

PR #7492, opened 2026-01-05
Add io.github.AmarBego.GitTop

AmarBego:

- [X] Please describe the application briefly. < A lightweight desktop client for GitHub notifications. Why spin up a browser just to check your GitHub notifications? >

https://github.com/user-attachments/assets/f5cb7680-a1c7-49e0-ba89-7520e6263906

- [x] I am an _author/developer_ to the project.


flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64"]
```
bbhtt: why?

io.github.AmarBego.GitTop.yml
```diff
@@ -0,0 +1,46 @@
...
+
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --device=dri
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.freedesktop.secrets
```
bbhtt, line 19: why does this not use the portal?

io.github.AmarBego.GitTop.yml
```diff
@@ -0,0 +1,46 @@
+app-id: io.github.AmarBego.GitTop
+runtime: org.freedesktop.Platform
+runtime-version: '24.08'
```
bbhtt: 25.08 is latest

io.github.AmarBego.GitTop.yml
```diff
@@ -0,0 +1,46 @@
...
+modules:
+  - name: gittop
+    buildsystem: simple
+
+    build-commands:
+      - mv cargo/config cargo/config.toml
+      - cargo --offline fetch --manifest-path Cargo.toml --verbose
+      - cargo build --release --offline
+      - install -Dm755 target/release/gittop /app/bin/gittop
+      - sed -i "s/^Icon=.*/Icon=io.github.AmarBego.GitTop/" src/platform/resources/gittop.desktop
```
bbhtt: fix this upstream

hfiguiere: Never use sed to change desktop files. It has no failure mode. Use the command `desktop-file-edit`.

io.github.AmarBego.GitTop.yml
AmarBego: while the portals exist for these, the standard rust dependencies (notify-rust and keyring) currently use the direct d-bus interfaces. rewriting the application's auth and notification systems to use different libraries (like ashpd) is out of scope for this PR.  these permissions are needed for the current code to function.

AmarBego:
good to check

io.github.AmarBego.GitTop.yml
bbhtt: you should have written it in the first place to use the portals.

AmarBego: @bbhtt I’ve reviewed both the Flathub packaging requirements and the Flatpak documentation, and I cannot find any rule stating that portals are a hard requirement for authentication or notifications. Flatpak explicitly allows scoped D-Bus access when necessary, provided permissions are minimal.
As stated [here](https://docs.flatpak.org/en/latest/sandbox-permissions.html#d-bus-access), Flatpak does not require portals for all functionality. The requirement is to avoid blanket session-bus/system-bus sockets and instead grant minimal --talk-name permissions when direct D-Bus communication is necessary. this is even stated in [flathub permission requirements ](https://docs.flathub.org/docs/for-app-authors/requirements#permissions)
The permissions used by this application are already the minimum scoped ones required by upstream:
--talk-name=org.freedesktop.Notifications
--talk-name=org.freedesktop.secrets

as i said before I understand the preference for portals, but preferences are not Flathub policy. The current submission is evaluated against Flatpak and Flathub requirements as they exist today. Scoped --talk-name permissions are explicitly documented as an acceptable mechanism for minimal D-Bus access. Until notify-rust and keyring provide native portal backends, these are objectively the minimum permissions possible.
If portal migration becomes a formal requirement in the future, the application can be updated then, but it is not grounds to block an otherwise compliant package.

Crates for this do not exist on a proper crossplatform level that upstream requires

bbhtt: You don't get to decide what gets accepted or not.

And the very first sentence of the permission guide you linked states to avoid static permissions as much as possible.

I see at least some that can be avoided by switching to use the portal.

AmarBego: ive never said i decide whats accepted or not? anyways the change makes no sense upstream and is fully compliant by flathubs own requirements as right now it is not possible to do with stable long standing backwards compatible crates and the whole reason of the program existing is for it to be as lean as possible and shipping a whole bridge with it defeats that purpose

bbhtt: > and is fully compliant by flathubs own requirements

It's not.

AmarBego: "should use [XDG Portals](https://flatpak.github.io/xdg-desktop-portal/docs/) and XDG standards whenever possible." keyword being whenever possible and it fully states to avoid as much as possible, which doesn't mean not allowed to use it
current scope of the project makes this not possible to switch to

bbhtt: It's possible you just don't want to. At least using the notification portal can be directly implemented in the app itself. It doesn't require a new library for it.

There are apps doing this out there.

AmarBego: it literally is not possible with rust-notify as it is strictly D-Bus only for now which ergo makes it impossible to communicate via XDG Portals. and what you're asking me to do is re-implement async DBus portal handling which im not a massive engineering team to make or maintain effectively

PR #7573, opened 2026-01-13
New app: org.superjmn.com.zafiro.avalonia.TestApp

SuperJMN:
This is an application that helps Avalonia developers explore the Zafiro.Avalonia library. It shows the components and utilities of the library.


SuperJMN:
Closing to resubmit with correct format

PR #7624, opened 2026-01-18
Update submission template

ghost:
Add a checkbox about AI-generated submissions

Since this is not a submission, I assume I don't need to fill out the template.


ghost:
@bbhtt FYI.

bbhtt:
What purpose does this serve? We are not reviewing every commit and lines of code to see who is writing them.

ghost:
Okay, I assumed it was forbidden to submit AI-generated apps to Flathub. Thanks.

hfiguiere:
Did you read the policy in the documentation?

I would have rather taken an update of the template so that the license be directly answered.

ghost:
My English is not as good as you think. It says

> **important**
> AI-generated **submissions** are not allowed as these are typically low-quality, often violate the submission requirements, and create unnecessary work for the reviewers. They also raise ethical and legal concerns such as potential licence infringement and resource wastage.
> We reserve the right to reject such submissions, without any review.

**What I understood from the text:**

Don't send AI generated code.

PR #7636, opened 2026-01-21
Add io.github.alyraffauf.Switchyard

alyraffauf:

- [x] Please describe the application briefly. Switchyard is a rules-based URL router and browser launcher that allows the user to configure which URLs open in which browsers.
- [x] Please attach a video showcasing the application on Linux using the Flatpak. [Screencast_20260120_194911.webm](https://github.com/user-attachments/assets/2e67f744-6d73-4e6f-b600-362ba8529f38)
- [x] I am an _(please keep whichever is applicable and remove the rest)_ author/developer to the project.
      If not, I contacted upstream developers about this submission. **Link:** N/A

### DIFF
--- a/io.github.alyraffauf.Switchyard.yml
+++ b/io.github.alyraffauf.Switchyard.yml
@@ -1,0 +1,50 @@
+app-id: io.github.alyraffauf.Switchyard
+runtime: org.gnome.Platform
+runtime-version: "49"
+sdk: org.gnome.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.golang
+command: switchyard
+
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --talk-name=org.freedesktop.Flatpak
+  - --filesystem=/var/lib/flatpak:ro
+  - --filesystem=/var/lib/snapd:ro
+  - --filesystem=host:ro
+  - --filesystem=xdg-data/applications:ro
+  - --filesystem=xdg-data/flatpak:ro
+
+modules:
+  - name: glib
+    buildsystem: meson
+    config-opts:
+      - -Dtests=false
+    sources:
+      - type: archive
+        url: https://download.gnome.org/sources/glib/2.86/glib-2.86.3.tar.xz
+        sha256: b3211d8d34b9df5dca05787ef0ad5d7ca75dec998b970e1aab0001d229977c65
+      - type: patch
+        path: build-aux/flatpak/glib-appinfo.patch
+
+  - name: switchyard
+    buildsystem: simple
+    build-options:
+      append-path: /usr/lib/sdk/golang/bin
+    build-commands:
+      - |
+        . /usr/lib/sdk/golang/enable.sh
+        go build -mod=vendor -trimpath -o switchyard-bin ./src
+      - install -Dm755 switchyard-bin /app/bin/switchyard-bin
+      - install -Dm755 data/switchyard-wrapper.sh /app/bin/switchyard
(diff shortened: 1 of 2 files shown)


io.github.alyraffauf.Switchyard.yml
```diff
@@ -0,0 +1,50 @@
...
+  - --device=dri
+  - --talk-name=org.freedesktop.Flatpak
+  - --filesystem=/var/lib/flatpak:ro
+  - --filesystem=/var/lib/snapd:ro
+  - --filesystem=host:ro
+  - --filesystem=xdg-data/applications:ro
+  - --filesystem=xdg-data/flatpak:ro
+
+modules:
+  - name: glib
```
hfiguiere, line 22: why do you need to patch glib?

alyraffauf: When we query for browsers, Glib will ignore them because we're in a sandboxed environment and the paths in Exec and Try_Exec don't exist. Effectively, we will not be able to detect any browsers at all without this small patch. Outside the sandbox, this is not an issue, because the binaries listed in Exec and Try_Exec provably exist, so Glib returns those desktop files as web browsers when we ask for them. It's the same patch used by https://github.com/flathub/re.sonny.Junction after they ran into the same problem.

hfiguiere: Two wrongs don't make a right.

alyraffauf: Apologies, was just pointing to the provenance of the patch/reasoning behind it. I would love for there to be an alternative here, but I'm not sure what it would look like without fixing the issue in Glib.

alyraffauf: If there's a more Flatpak/Flathub-native way to do this I'm happy to bring this submission back after working it out. So far the only other solutions I know of to account for the divergent and unexpected Glib behavior in and out of sandbox would be brittle (manually parsing .desktop files and finding executables, icons, etc. over various directories, etc) and require substantial Flatpak-specific codebase changes to implement.

barthalion: Isn't [Junction](https://flathub.org/en/apps/re.sonny.Junction) doing similar thing without patching glib?

alyraffauf: > Isn't [Junction](https://flathub.org/en/apps/re.sonny.Junction) doing similar thing without patching glib?

No, unfortunately, as stated above, junction in Flatpak patches glib. The patch in question is the same one used by Junction.

https://github.com/flathub/re.sonny.Junction/blob/41c3257927b0426f96a11d03fd320fad42c446f7/re.sonny.Junction.json#L33

barthalion: Oh, that sucks.

barthalion:
~~You will need to submit exceptions for these permissions to https://github.com/flathub-infra/flatpak-builder-lint~~ I noticed you did, ignore me!

io.github.alyraffauf.Switchyard.yml
hfiguiere: you could just copy & paste that bit of code. It's sonmething localized to a single function.

All of that look awfuly wrong.

alyraffauf: Do you mean the code in Glib itself? We would need to re-implement code to find the correct icons in addition, it's not just about the executables, so it would be quite a lot of code to manage that's specific to Flatpak. This is the blessed way to do so outside Flatpak, unpatched Glib just assumes you're on the same filesystem so it breaks in Flatpak.

hfiguiere: either it's something different that what glib intended or it's a bug in glib. But I don't think rebuilding glib with a patch is a solution unless the patch comes from upstream (because it the latter)

barthalion:
bot, build

Issue #7640, opened 2026-01-21
Rename com.github.marinm.songrec to re.fossplant.songrec


marin-m:
Hello,

I have renamed my `com.github.marinm.songrec` GTK+ app to `re.fossplant.songrec` upstream (the former name was invalid and could not be verified because a dash was missing in the Github username, now I own a https://songrec.fossplant.re/ domain, also it was long and dependant of my Github username) and this happens to require a change in the name of the package repository.

I was previously told (in issue #5652) that the install count for the app will be preserved + that the users will be redirected to the new package when doing the updates in GNOME Software thanks to the "EOL rebasing" procedure, can you please confirm that it will be the case?

See the current PR in the package repository:

* https://github.com/flathub/com.github.marinm.songrec/pull/10

See former issues for the track record:

* https://github.com/marin-m/SongRec/issues/180
* https://github.com/flathub/flathub/issues/5652

Regards,

barthalion:
You will need to submit it as a new app, then mark the old one with EOL flag: https://docs.flathub.org/docs/for-app-authors/maintenance#end-of-life

PR #7643, opened 2026-01-21
Add re.fossplant.songrec

marin-m:

- [x] Please describe the application briefly: SongRec is an open-source, unofficial Shazam client for Linux. The intent of this pull request is to replace the original [`com.github.marinm.songrec`](https://flathub.org/fr/apps/com.github.marinm.songrec) app ID with a new `re.fossplant.songrec` app ID for various reasons, see https://github.com/flathub/flathub/issues/7640
- [x] Please attach a video showcasing the application on Linux using the Flatpak. => https://github.com/user-attachments/assets/36f3b007-ac05-428b-a87d-c84308fa2104
- [x] I am an original author and maintainer of the project.

### DIFF
--- a/re.fossplant.songrec.json
+++ b/re.fossplant.songrec.json
@@ -36,7 +36,7 @@
                 {
                     "type": "archive",
                     "url": "https://github.com/marin-m/SongRec/releases/download/0.5.0/songrec_tarball_0.5.0_for_flathub_build.tar.gz",
-                    "sha256": "7fd5e9fb92c16fbccfb921fe304c316bac3d747b4cfb9a696674db0bb1a3c628"
+                    "sha256": "3b13a1fe06c129aeb4b80ee54bde65f8bd5572351da9487d9e6a2ddbc8dadd22"
                 }
             ]
         }


re.fossplant.songrec.json
```diff
@@ -0,0 +1,45 @@
...
+                "cargo --offline fetch --verbose",
+                "cargo --offline build --release --verbose",
+                "install -Dm755 ./target/release/songrec -t /app/bin/",
+                "cp -ra ./packaging/rootfs/usr/share /app/",
+                "cp -ra ./translations/locale /app/share/"
+            ],
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://github.com/marin-m/SongRec/releases/download/0.5.0/songrec_tarball_0.5.0_for_flathub_build.tar.gz",
```
hfiguiere, line 38: This should be built from source.

re.fossplant.songrec.json
```diff
@@ -0,0 +1,45 @@
...
+        "org.freedesktop.Sdk.Extension.rust-stable"
+    ],
+    "command": "songrec",
+    "finish-args": [
+        "--share=network",
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--device=all",
+        "--filesystem=home:ro",
```
hfiguiere: why does this need home access?

marin-m: This was deemed to be useful when invoking app through CLI, because audio files passed by users will then be loaded outside of a XDG portal, see:

* Discussion thread: https://github.com/flathub/flathub/pull/1914#discussion_r515678461
* Previous lgtm from Flathub maintainers: https://github.com/flathub/flathub/pull/1914#discussion_r515917318
* Associated commit: https://github.com/flathub/flathub/pull/1914/changes/cda9be6a69f6254a3639ae99ba7c936cc507a92f

re.fossplant.songrec.json
marin-m: The original PR (https://github.com/flathub/flathub/pull/1914) was created following instructions you can find here: https://belmoussaoui.com/blog/8-how-to-flatpak-a-rust-application/

The reason for using a tarball pre-generated from source, mentioned in the article is:

> You can notice that the application is built with network access from the build-args. Flatpak by default, downloads all the sources and don’t give network access to anything during the build process. Except we need Cargo to be able to download our dependencies. Such manifest is the way to go if you’re using it for CI. If your plan is to use it to distribute your application on a Flatpak based store like Flathub, all the dependencies have to be downloaded before starting the build.

The script used to generate the tarball from source is still: https://github.com/marin-m/SongRec/blob/master/packaging/flatpak/generate_dist_tarball_for_flathub.sh (it takes a Git tag as an argument)

Should be this way of doing considered totally deprecated? If way, is there a more up-to-date tutorial somewhere that can be followed?

hfiguiere: my bad I overlooked the build commands. (need more coffee)

it's that usually when tarball have "flathub" in their name it's because it's some prebuilt stuff.

a vendored tarball is not flathub specific so it's misleading.

No further questions

re.fossplant.songrec.json
hfiguiere: flathub is not for cli, so punching a hole in the snadbox for that use case is excessive.

marin-m: This is a dual GUI+CLI tool (but it can also be installed through other channels).

Which permission line should be used instead then if this is no more accepted?

barthalion: Flatpak has the `--file-forwading` switch for passing files through portals, e.g.

```
~/d/f/flatpak-builder-lint master | 1 ❱ flatpak run --file-forwarding --nofilesystem=home com.github.marinm.songrec audio-file-to-fingerprint @@ ~/Music/Nagisa\ Ni\ Te/Dream\ Sounds/02\ Anxiety.flac @@
data:audio/vnd.shazam.sig;base64,gCX+ykI4mcOsFAAAAJwRlAAAAAAAAA
```

So I'm inclined to agree that `home:ro` can be dropped this time around. I know this sounds annoying, but we've made the review process more thorough since Songrec was published 6 years ago; we "grandfather" apps when adding new rules but apply them on renames.

marin-m:
bot, build

marin-m:
bot, build

barthalion:
/merge head=d0352c6b5cd843541ddf104427e37d34ed27e348

Please remember to mark the old ID as EOL+EOL-rebase when this gets published. https://docs.flathub.org/docs/for-app-authors/maintenance#end-of-life

PR #7670, opened 2026-01-24
Add io.github.dicroce.Revere

dicroce:

- [X] Please describe the application briefly. Revere is on Onvif compatible video surveillance system. It contains an app for camera discovery and recording and app for viewing.
- [X] Please attach a video showcasing the application on Linux using the Flatpak. < https://youtu.be/aeviZqD3UBY >
- [X] The Flatpak ID follows all the rules listed in the [Application ID requirements][io.github.dicroce.Revere].
- [X] I am a developer.

### DIFF
--- a/io.github.dicroce.Revere.yaml
+++ b/io.github.dicroce.Revere.yaml
@@ -142,8 +142,8 @@ modules:
     sources:
       - type: git
         url: https://github.com/dicroce/revere.git
-        tag: v0.0.7
-        commit: f82304a3fd14ec8c478e6640b4fd42fdcae9b166
+        tag: v0.0.8
+        commit: 617c9a493253904c3466db7b56694dc37108542f
       # Pre-fetch FetchContent dependencies (pinned to commits for reproducibility)
       - type: git
         url: https://github.com/dicroce/nanots.git


io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,262 @@
+app-id: io.github.dicroce.Revere
+runtime: org.freedesktop.Platform
+runtime-version: '24.08'
```
bbhtt:
```suggestion
runtime-version: '25.08'
```

io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,262 @@
...
+
+  # Network access for RTSP cameras and ONVIF discovery
+  - --share=network
+
+  # Storage in ~/Documents/revere
+  - --filesystem=xdg-documents/revere:create
+
+  # System tray support (StatusNotifierItem protocol)
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=org.freedesktop.Notifications
```
bbhtt: notifications should use the portal

io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,262 @@
...
+
+  # Environment variables for GStreamer plugin discovery
+  - --env=GST_PLUGIN_PATH=/app/lib/gstreamer-1.0
+  # Library path for revere libs, OpenCV, gst-rtsp-server, and bundled ffmpeg
+  - --env=LD_LIBRARY_PATH=/app/revere:/app/lib64:/app/lib
```
bbhtt: no don't set this, all libraries should be installed to `/app/lib` instead

io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,262 @@
...
+
+build-options:
+  env:
+    PKG_CONFIG_PATH: /app/lib64/pkgconfig:/app/lib/pkgconfig:/app/lib/x86_64-linux-gnu/pkgconfig:/usr/lib/pkgconfig:/usr/lib/x86_64-linux-gnu/pkgconfig:/usr/share/pkgconfig
```
bbhtt:
```suggestion
```

io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,262 @@
...
+      - type: git
+        url: https://github.com/webmproject/libwebp.git
+        tag: v1.4.0
+
+  # ============================================
+  # Module 8: FFmpeg with full codec support
+  # Decode: H.264, H.265, AAC, ulaw, alaw
+  # Encode: H.264, AAC, JPEG, WebP
+  # ============================================
+  - name: ffmpeg
```
bbhtt: ffmpeg is in the runtime and in the codecs-extra extension why is this needed?

io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,262 @@
...
+
+  # ============================================
+  # Module 9: Revere Application
+  # ============================================
+  - name: revere
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DCMAKE_INSTALL_PREFIX=/app
```
bbhtt:
```suggestion
```

io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,262 @@
...
+        url: https://github.com/dicroce/traypp.git
+        commit: 13da575044ff10be98d832a9198204d36d2bbec4
+        dest: deps/traypp
+      - type: shell
+        commands:
+          # Patch traypp to use libappindicator instead of ayatana-appindicator
+          - sed -i 's/ayatana-appindicator3-0.1/appindicator3-0.1/g' deps/traypp/CMakeLists.txt
+          - sed -i 's/libayatana-appindicator/libappindicator/g' deps/traypp/CMakeLists.txt
+          # Patch all source files to use libappindicator includes
+          - find deps/traypp -type f \( -name "*.hpp" -o -name "*.cpp" -o -name "*.h" \) -exec sed -i 's|libayatana-appindicator/app-indicator.h|libappindicator/app-indicator.h|g' {} \;
```
bbhtt: use a patch

io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,262 @@
...
+      - -Dexamples=disabled
+      - -Dtests=disabled
+      - -Dintrospection=disabled
+    sources:
+      - type: archive
+        url: https://gstreamer.freedesktop.org/src/gst-rtsp-server/gst-rtsp-server-1.24.10.tar.xz
+        sha256: db21dfdd7bf2e718564d557378ada5358b411efe2a3e89e9f0f87a74537e2adc
+    post-install:
+      - ls -la /app/lib/pkgconfig/ || true
+      - find /app -name "*.pc" 2>/dev/null || true
```
bbhtt:
```suggestion
```

io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,262 @@
...
+
+  # ============================================
+  # Module 4: libdbusmenu (required by ayatana-appindicator)
+  # ============================================
+  # Shared modules for appindicator support (from flathub/shared-modules)
+  - shared-modules/intltool/intltool-0.51.json
```
hfiguiere, line 122: submodules are missing.

io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,262 @@
...
+        dest: deps/glfw
+    post-install:
+      # Create bin directory symlinks
+      - mkdir -p /app/bin
+      - ln -sf /app/revere/revere /app/bin/revere
+      - ln -sf /app/revere/vision /app/bin/vision
+
+      # Install desktop files
+      - install -Dm644 ../flatpak/io.github.dicroce.Revere.desktop /app/share/applications/io.github.dicroce.Revere.desktop
+      - install -Dm644 ../flatpak/io.github.dicroce.Vision.desktop /app/share/applications/io.github.dicroce.Vision.desktop
```
hfiguiere: this desktop file will never be exported.

io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,262 @@
...
+
+      # Install icons at multiple sizes
+      - install -Dm644 ../apps/revere/R.png /app/share/icons/hicolor/256x256/apps/io.github.dicroce.Revere.png
+      - install -Dm644 ../apps/vision/V.png /app/share/icons/hicolor/256x256/apps/io.github.dicroce.Vision.png
```
hfiguiere: nor this icon

io.github.dicroce.Revere.yaml
dicroce: done.

io.github.dicroce.Revere.yaml
dicroce: fixed.

io.github.dicroce.Revere.yaml
dicroce: done.

io.github.dicroce.Revere.yaml
dicroce: codecs-extra worked. changed it to that one.

io.github.dicroce.Revere.yaml
dicroce: done.

io.github.dicroce.Revere.yaml
dicroce: fixed.

io.github.dicroce.Revere.yaml
dicroce: fixed.

io.github.dicroce.Revere.yaml
dicroce: fixed

io.github.dicroce.Revere.yaml
dicroce: fixed

io.github.dicroce.Revere.yaml
dicroce: fixed.

io.github.dicroce.Revere.yaml
dicroce: fixed.

io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,210 @@
...
+  - --filesystem=xdg-documents/revere:create
+
+  # System tray support (StatusNotifierItem protocol)
+  - --talk-name=org.kde.StatusNotifierWatcher
+
+  # Environment variables for GStreamer plugin discovery
+  - --env=GST_PLUGIN_PATH=/app/lib/gstreamer-1.0
+
+add-extensions:
+  org.freedesktop.Platform.ffmpeg-full:
```
hfiguiere: this doesn't exist anymore.

io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,210 @@
...
+  # Module 4: libdbusmenu (required by ayatana-appindicator)
+  # ============================================
+  # Shared modules for appindicator support (from flathub/shared-modules)
+  - shared-modules/intltool/intltool-0.51.json
+  - shared-modules/libappindicator/libappindicator-gtk3-12.10.json
+
+  # ============================================
+  # Module 5: patchelf (for fixing RPATH)
+  # ============================================
+  - name: patchelf
```
hfiguiere: there should be a cleanup rule to remove it from the build.

io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,210 @@
...
+      - ln -sf /app/revere/revere /app/bin/revere
+      - ln -sf /app/revere/vision /app/bin/vision
+
+      # Move revere's shared libraries to /app/lib (flatpak standard location)
+      - mkdir -p /app/lib
+      - find /app/revere -maxdepth 1 -name '*.so*' -exec mv {} /app/lib/ \;
+
+      # Update RPATH so executables and libraries find dependencies in /app/lib and /app/lib64
+      - /app/bin/patchelf --set-rpath /app/lib:/app/lib64 /app/revere/revere
+      - /app/bin/patchelf --set-rpath /app/lib:/app/lib64 /app/revere/vision
```
hfiguiere: /app/lib64 doesn't exist.

hfiguiere: if you need to do that then your build system is totally broken.

io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,210 @@
...
+      # Tell CMake where to find pre-fetched FetchContent dependencies
+      - -DFETCHCONTENT_SOURCE_DIR_NANOTS=/run/build/revere/deps/nanots
+      - -DFETCHCONTENT_SOURCE_DIR_PUGIXML=/run/build/revere/deps/pugixml
+      - -DFETCHCONTENT_SOURCE_DIR_TRAYPP=/run/build/revere/deps/traypp
+      - -DFETCHCONTENT_SOURCE_DIR_GLFW3=/run/build/revere/deps/glfw
+    build-options:
+      env:
+        NCNN_TOP_DIR: /app
+        OPENCV_TOP_DIR: /app
+        FLATPAK_ID: io.github.dicroce.Revere
```
hfiguiere:
```suggestion
```

this is set

io.github.dicroce.Revere.yaml
dicroce: ffmpeg-full doesn't exist anymore?

In my initial submission I had ffmpeg building as part of the flatpak and it was suggested I use this.

io.github.dicroce.Revere.yaml
dicroce: what should be removed from the build?

io.github.dicroce.Revere.yaml
hfiguiere: it was using the older runtime. But I guess second guessing is part of the submission.

io.github.dicroce.Revere.yaml
hfiguiere: patchelf itself.

io.github.dicroce.Revere.yaml
dicroce: Hmmm.... I only have to do this for flatpak... The app sets its rpath to ./ for multiple reasons: 1) everything can be one directory tree. It won't step on any of your systems existing files. 2) it means the same install rules work on windows and linux 3) it definitely made macos support easier.

hfiguiere: so it's broken.

dicroce: Is this a deal breaker? I have to choose between a portable app and flatpak?

hfiguiere: you don't have to chose. Just have to do it properly. Not setting the rpath for example. `/app/lib64` doesn't exist either.

dicroce: Ok, so the whole idea of flatpak and snap (etc) is more portable apps. I love that, I think linux needs more of that. That's the whole reason I believe in setting rpath to ./   It lets you create self contained apps... It's also more compatible with other systems I care about too.... If it's a deal breaker please just say so and I'll give up on flatpak for my app.

hfiguiere: So you believe that breaking the principle of how software is built on a platform but not allowing to opt-out is correct?

patchelf is a tool to fix mis-compiled proprietary software. It's usually a testament to the low quality of said software.

Don't set the rpath and you are fine. This is not flatpak specific either.

dicroce: Would you be OK with me adding a build option to my project that I would use only for flatpak builds (perhaps -DFLATPAK_BUILD=ON) that doesn't set rpath? So the default source build would still make a portable app, but I could do a FLATPAK_BUILD=ON build to create binaries for flatpak?

dicroce: Also, lib64 is where opencv installs to by default. I can see if there is a way to change that, but I didn't create /lib64.

hfiguiere: > Also, lib64 is where opencv installs to by default.

No it's not. You probably are not using the recommended version of flatpak-builder that comes as a flatpak **as recommended in our documentation** and instead use the broken[1] one from your distro. 


[1] the breakage is caused by a combination of other tool like CMake breaking and wrong default chose eleswhere.

dicroce: Regarding lib64: I tested with flatpak run org.flatpak.Builder (the flatpak version, not the distro package) and OpenCV still installed to /app/lib64. This appears to be OpenCV's default behavior based on CMake's GNUInstallDirs detection, not a distro   
  flatpak-builder issue.                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                
  I've fixed it by explicitly passing -DCMAKE_INSTALL_LIBDIR=lib to the OpenCV build. Everything now installs to /app/lib.                                                                                                                                      
                                                                                                                                                                                                                                                                
  I've also removed all patchelf usage and wrapped the RPATH settings in the build system with if(NOT FLATPAK_BUILD), so RPATH is no longer set when building for flatpak.

io.github.dicroce.Revere.yaml
dicroce: fixed.

io.github.dicroce.Revere.yaml
dicroce: fixed

io.github.dicroce.Revere.yaml
dicroce: You're right - org.freedesktop.Platform.ffmpeg-full doesn't exist for 25.08. It's been renamed to org.freedesktop.Platform.codecs-extra with branch 25.08-extra.                                                                                              
                                                                                                                                                                                                                                                                
However, after testing, I found that the runtime handles this automatically - the codecs-extra extension mounts at /usr/lib/x86_64-linux-gnu/codecs-extra/ without needing an explicit add-extensions declaration. I've removed the add-extensions section entirely and the app works correctly.

dicroce:
bot, build

dicroce:
bot, build

io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,189 @@
...
+  # Module 2: OpenCV with Contrib Modules
+  # Must be built from source for contrib modules
+  # (bgsegm, tracking, optflow, etc.)
+  # ============================================
+  - name: opencv
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DCMAKE_INSTALL_LIBDIR=lib
```
hfiguiere:
```suggestion
```

this is set by flatpak-builder.

io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,189 @@
...
+        path: traypp-libappindicator.patch
+        options:
+          - -d
+          - deps/traypp
+      - type: git
+        url: https://github.com/glfw/glfw.git
+        commit: a74efa0d5628b74adc0426af4c5710e287fa7c2c
+        dest: deps/glfw
+    post-install:
+      # Create bin directory symlinks
```
hfiguiere: why doesn't the install target install all of this (the post-install)? This is probably the root of all the problems.

dicroce: You were right - the install target is now the proper solution. I've updated the CMake build system to detect Flatpak builds and install to standard FHS paths.                                                                                               
                                                                                                                                                                                                                                                                
  The build now detects FLATPAK_ID environment variable (set automatically by flatpak-builder) and sets FLATPAK_BUILD=TRUE. When building for Flatpak:                                                                                                          
  - Executables install to /app/bin                                                                                                                                                                                                                             
  - Libraries install to /app/lib                                                                                                                                                                                                                               
  - RPATH settings are skipped (not needed with standard paths)                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                
  For non-Flatpak builds, it preserves the portable app behavior (single directory with $ORIGIN RPATH).                                                                                                                                                         
                                                                                                                                                                                                                                                                
  The relevant changes are in settings.cmake:                                                                                                                                                                                                                   
  if(DEFINED ENV{FLATPAK_ID})                                                                                                                                                                                                                                   
      set(FLATPAK_BUILD TRUE)                                                                                                                                                                                                                                   
  endif()                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                
  if(FLATPAK_BUILD)                                                                                                                                                                                                                                             
      set(REVERE_INSTALL_BINDIR bin)                                                                                                                                                                                                                            
      set(REVERE_INSTALL_LIBDIR lib)                                                                                                                                                                                                                            
  else()                                                                                                                                                                                                                                                        
      set(REVERE_INSTALL_BINDIR revere)                                                                                                                                                                                                                         
      set(REVERE_INSTALL_LIBDIR revere)                                                                                                                                                                                                                         
  endif()                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                
  Updated to v0.0.4 which includes these changes. The post-install hacks are gone.

io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,180 @@
...
+        path: traypp-libappindicator.patch
+        options:
+          - -d
+          - deps/traypp
+      - type: git
+        url: https://github.com/glfw/glfw.git
+        commit: a74efa0d5628b74adc0426af4c5710e287fa7c2c
+        dest: deps/glfw
+    post-install:
+      # Install desktop files
```
hfiguiere: these two. hint. neither is even specific to flatpak. It's XDG standard.

dicroce: fixed.

dicroce:
bot, build

io.github.dicroce.Revere.yaml
dicroce: fixed.

dicroce:
bot, build

dicroce:
All comments are resolve right? Test build worked.

io.github.dicroce.Revere.yaml
```diff
@@ -0,0 +1,168 @@
...
+      - -DFETCHCONTENT_SOURCE_DIR_PUGIXML=/run/build/revere/deps/pugixml
+      - -DFETCHCONTENT_SOURCE_DIR_TRAYPP=/run/build/revere/deps/traypp
+      - -DFETCHCONTENT_SOURCE_DIR_GLFW3=/run/build/revere/deps/glfw
+    build-options:
+      env:
+        NCNN_TOP_DIR: /app
+        OPENCV_TOP_DIR: /app
+    sources:
+      - type: git
+        url: https://github.com/dicroce/revere.git
```
bbhtt, line 144: why do you have a 13 MB .c file embedding the entire roboto font or something?

bbhtt: also an entire copy of mbedTLS in the repo committed...

dicroce: Yes.
mbedTLS is needed because some camera manufacturers allow their cameras to be configured such that TLS is required for their onvif interface. mbedTLS is also nice because it provides a single clean TLS api that works on all the platforms I need (linux, windows & macos). Keeping the number of binary dependencies down is very helpful in a cross platform project... For some of the dependencies this would be impossible but for some recompiling is fast and straightforward.

bbhtt, line 144: tray icon is not working on gnome. it appears as 3 dots.

bbhtt:
I'm increasingly getting the feeling this was written with AI? 

External code being directly committed to the repo is not helping. I can see several modules committed to the repo which is not written by you and is under an entirely different licence

eg. https://github.com/dicroce/revere/blob/e20722ba599c749038f84f6506f33e826b241ec1/libs/imgui/include/imgui/ImGuiFileDialog.h#L4

mbedtls as I mentioned earlier.

Dependencies are not supposed to be used this way...

dicroce:
I started this project well before LLM's existed.... Nevertheless I won't deny that I use Claude now... My guess would be less than 10% of the code is Claude written. I've been a professional C++ dev since 1996, every line of Claude written code has been code reviewed by myself.

With regard to some of the codebases embedded herein: C++ dependencies are a pain in the ass, especially for cross platform projects... In fact many C++ dependencies are designed to be used in this way exactly (single header projects like nlohmann's json lib, or sqlite's amalgamated distribution)...

io.github.dicroce.Revere.yaml
dicroce: Fixed. Looks like on gnome I needed to pass the icon id instead of the path to the icon.

dicroce:
> I'm increasingly getting the feeling this was written with AI?
> 
> External code being directly committed to the repo is not helping. I can see several modules committed to the repo which is not written by you and is under an entirely different licence
> 
> eg. https://github.com/dicroce/revere/blob/e20722ba599c749038f84f6506f33e826b241ec1/libs/imgui/include/imgui/ImGuiFileDialog.h#L4
> 
> mbedtls as I mentioned earlier.
> 
> Dependencies are not supposed to be used this way...

My README.md has an acknowledgement for ImGuiFileDialog in the ## Technology Stack & Acknowledgements section

bbhtt:
I won't be accepting this submission. Blocked on https://docs.flathub.org/docs/for-app-authors/requirements#generative-ai-policy, https://docs.flathub.org/docs/for-app-authors/requirements#structure-and-organisation

dicroce:
Well that's disappointing. I was about to check in a change where mbedtls is pulled via FetchContent (since you seemed concerned with having it checked in).

PR #7673, opened 2026-01-25
Add es.gob.afirma

ivan-gj:

- [x] Please describe the application briefly. Autofirma is an electronic signature application from the Spanish Government
[autofirma-flatpak.webm](https://github.com/user-attachments/assets/3a16eb07-c1cf-415e-ae90-7b49eb2c1594)

- [x] I contacted upstream developers about this submission. **Link:** https://github.com/ctt-gob-es/clienteafirma/issues/484

I suggest **the following organization should have access to this repository**, as they are the upstream developers and should take over as maintainers at some point in time.
- [@adminctt](https://github.com/adminctt)

#### General notes

This application is focused on Spanish citizens, so it is **only available in Spanish**, but we hope you don't have a problem with this. We put as much metadata in English as possible. 

In terms of linting, when launching
```sh
flatpak run --command=flatpak-builder-lint org.flatpak.Builder manifest es.gob.afirma.yaml
```
or 
```sh
flatpak run --command=flatpak-builder-lint org.flatpak.Builder repo repo
```

we get the following errors.

#### finish-args-flatpak-appdata-folder-access

This is produced by `--filesystem=~/.var/app/org.mozilla.firefox/.mozilla/firefox:rw`. 
We think it's necessary in order to have the same experience as with the "native" package, which **automatically installs the corresponding certificates onto the user's Firefox database**, and in this case we are trying to apply the same logic for a hypothetical Firefox Flatpak, for example in an immutable environment.

This is of course not the most secure experience, and there's a possibility for a manual install by the user, but if this is not a deal-breaker, we'd rather have this and avoid too many complaints about the same issue.

But **please confirm with us** if we need to take a different route on this.

#### appid-url-not-reachable

This is a community package of a Spanish Government application that has historically always used `es.gob.afirma` as their app ID, and our goal is to **hand them the keys to the repository**, because the Flatpak is a much better solution than their current distribution methods. They do have control over the `gob.es` domain, but they don't have `https://afirma.gob.es` available. We feel this app ID is perfectly justified. 

#### finish-args-host-filesystem-access

The app needs host filesystem access because otherwise its internal file manager simply will not find any file, and it's a crucial feature for this app to digitally sign documents. It's a shame that it cannot use portals, but it's not up to us. 

The **importance** of this application for Spanish citizens makes us think this is a compromise worth making.

### Thank you

For maintaining one of the best, safest repositories of software in the world, if not the best.


es.gob.afirma.yaml
```diff
@@ -0,0 +1,154 @@
...
+runtime: org.freedesktop.Platform
+runtime-version: "25.08"
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.openjdk17
+command: autofirma
+
+finish-args:
+  - --filesystem=host
+  - --filesystem=~/.var/app/org.mozilla.firefox/.mozilla/firefox:rw
```
hfiguiere: why? You are basically giving permission to access all the firefox data.

es.gob.afirma.yaml
```diff
@@ -0,0 +1,154 @@
...
+      - "/bin/modutil"
+      - "/bin/signtool"
+
+    sources:
+      - type: archive
+        url: https://ftp.mozilla.org/pub/security/nss/releases/NSS_3_112_2_RTM/src/nss-3.112.2.tar.gz
+        sha256: 2019b8bd0bedeeae395fc360811af0eb51a19d047843059ec07ea435b90271d5
+
+  # Module for application store metadata (AppStream)
+  - name: metainfo
```
hfiguiere: this shouldn't be a separate module.

es.gob.afirma.yaml
```diff
@@ -0,0 +1,154 @@
...
+      # Ensure shared libraries (.so) are built
+      - --enable-shared
+      # Do not build static libraries (.a)
+      - --disable-static
+    sources:
+      - type: archive
+        url: https://ftp.mozilla.org/pub/nspr/releases/v4.37/src/nspr-4.37.tar.gz
+        sha256: 5f9344ed0e31855bd38f88b33c9d9ab94f70ce547ef3213e488d1520f61840fa
+
+  - name: nss
```
hfiguiere: nss is in the runtime

es.gob.afirma.yaml
ivan-gj: I explain it in the PR text, it's done so that the application can install a certificate on Firefox's database. Chromium-based browsers use a different system, somehow more "standardized".
The thing is that the file in this case is located right under that directory, and if I am not mistaken, filesystem permissions are for directories not files. Or is there a way?

es.gob.afirma.yaml
ivan-gj: I know, but the original scripts use `certutil`, which is part of `nss`, but somehow **it isn't available in the runtime**.

If this is trouble, I guess I can try and see if there is some kind of workaround.

es.gob.afirma.yaml
ivan-gj: Fixed, thank you very much.

es.gob.afirma.yaml
hfiguiere: the executable are in the sdk. just copy them over.

ivan-gj: Is that legal?? Man, if that's possible, then wow, things suddenly get much simpler. I'll try that very soon! Thanks!

assets/screenshots/settings.webp
bbhtt, line 1: why are these here?

patches/fix-desktop-file.patch
bbhtt, line 1: merge the desktop file patches into one

patches/fix-svg-icon-width.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/afirma-simple-installer/linux/instalador_deb/src/usr/share/AutoFirma/AutoFirma.svg b/afirma-simple-installer/linux/instalador_deb/src/usr/share/AutoFirma/AutoFirma.svg
+index 7c1c14a70..0a344b147 100755
+--- a/afirma-simple-installer/linux/instalador_deb/src/usr/share/AutoFirma/AutoFirma.svg
++++ b/afirma-simple-installer/linux/instalador_deb/src/usr/share/AutoFirma/AutoFirma.svg
+@@ -2,7 +2,7 @@
+ <!-- Created with Inkscape (http://www.inkscape.org/) -->
```
bbhtt: you have to drop the comment from the svg.

autofirma.bash
```diff
@@ -0,0 +1,170 @@
...
+_autofirma_ca="${_autofirma_dir}/AutoFirma_ROOT.cer"
+_autofirma_pfx="${_autofirma_dir}/autofirma.pfx"
+_cert_days="3650"
+_cert_cn="AutoFirma ROOT"
+_firefox_profiles_ini="${HOME}/.mozilla/firefox/profiles.ini"
+_firefox_flatpak_profiles_ini="${HOME}/.var/app/org.mozilla.firefox/.mozilla/firefox/profiles.ini"
+_nssdb="sql:${HOME}/.pki/nssdb"
+
+
+function _make_ca_config {
```
bbhtt: this should not be a script. if you need a config add the config file.

bbhtt:
I don't trust this at all considering it needs to modify trusted certificates and openssl configurations.

This should not be a community package.

If this is supposed to maintained by upstream then all the logic of the bash script should live in the application itself.

assets/screenshots/settings.webp
ivan-gj: Hello! I couldn't find a reliable place for this app's screenshots, so I figured why not the repo itself? I can take them out somewhere else, if you think that's better.

es.gob.afirma.yaml
ivan-gj: Now using `certutil` from SDK. Thank you.

patches/fix-svg-icon-width.patch
ivan-gj: Done!

patches/fix-desktop-file.patch
ivan-gj: Done!

autofirma.bash
ivan-gj: Done!

ivan-gj:
With this [last commit](https://github.com/flathub/flathub/pull/7673/commits/503119eead3a10d9991e5d9205ff2db0b522114f), we have greatly reduced downstream code in the launcher script. 

We originally based this script on [AUR's wrapper](https://aur.archlinux.org/cgit/aur.git/tree/autofirma?h=autofirma-bin), but @bbhtt was right to point out that very probably, upstream had a way to do this, and we found it, tested it, and it works perfectly fine.

ivan-gj:
Me and my colleagues are debating now whether it makes sense to keep support for the "auto-import" of CA into Firefox (Flatpak). It requires 
```yaml
  - --filesystem=~/.var/app/org.mozilla.firefox/.mozilla/firefox:rw
  ```
in the `finish-args` list, and also some "patching" in `autofirma.bash` and basically it is only trying to avoid having to manually import the certificate onto their browser; which the user would have to do anyway if they installed the browser **after** Autofirma is installed. 

We would love your input on this. 

--- 

## Update 2026-02-16

We are going with the simplified option, less downstream code, and just a bit more of manual action required from the user.

**Comment by trinukso** [comment]
Hi... sorry... still live?

**Comment by ivan-gj** [comment]
Yes! I was actually planning to work on this today. I've been very busy lately 😅, but discussions just ended and we have a clear path forward.

File: es.gob.afirma.yaml (hunk 1 above)
Comment (ivan-gj) [reply es.gob.afirma.yaml]: We got rid of the Flatpak specific parts, and switched to a "manual" action required.

**Comment by bbhtt** [comment]
Metainfo, patches and the bash script should be upstream. Also why are the troubleshooting docs here?

**Comment by ivan-gj** [comment]
We are trying to make upstream more compatible, but it's up to the Spanish government to allow that. And they do not move fast at all. 

Troubleshooting docs are here because we believe people will be looking up our repository for answers to their Flatpak issues. If you recommend a different paradigm, we can try and make it happen, but so far this seems like the easiest way for people to find. 

@bbhtt would you be okay if we patched their _metainfo_ and their _.deb postinstall script_ instead of simply making our own? We decided to go this route because there were simply too many changes required, but maybe you prefer it that way.

**Comment by bbhtt** [comment]
If there is already a metainfo in the .deb reusing that seems a good idea to me.

**Comment by ivan-gj** [comment]
We have now deleted our own custom files, and reused (and patched) the Debian installer files. 

Does it look better, now, @bbhtt ?

File: es.gob.afirma.yaml [hunk 6]
```diff
@@ -0,0 +1,97 @@
+app-id: es.gob.afirma
+runtime: org.freedesktop.Platform
+runtime-version: "25.08"
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.openjdk17
+command: autofirma
+
+finish-args:
+  - --filesystem=host
```
bbhtt, line 10: why is host needed? this needs an [exception](https://docs.flathub.org/docs/for-app-authors/linter#exceptions) https://docs.flathub.org/docs/for-app-authors/linter#finish-args-host-filesystem-access

es.gob.afirma.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - --socket=x11
+  - --share=ipc
+  - --share=network
+  - --socket=pcsc  # smart card support
+
+  - --env=PATH=/usr/bin:/app/bin:/app/jre/bin
+  # Likely needed to access the system keystore
```
bbhtt: is this needed or not?

es.gob.afirma.yaml
```diff
@@ -0,0 +1,97 @@
...
+    build-commands:
+      # Do not use ranged versions on bouncycastle deps. Fixes offline build
+      - sed -i -E 's|<version>\[([0-9]+\.[0-9]+).*?\)</version>|<version>\1</version>|g' .m2/repository/org/bouncycastle/bc*-jdk18on/*/bc*-jdk18on-*.pom
+      # Offline maven build (-o)
+      - mvn clean install -o -B -Denv=install
+
+      # Install main application files
+      - install -d /app/lib
+      - install -c -m 0755 afirma-simple/target/autofirma.jar /app/lib/
+      - install -Dm644 afirma-ui-simple-configurator/target/autofirmaConfigurador.jar /app/lib/autofirmaConfigurador.jar
```
bbhtt, line 49: replace the two step install commands with `-D` here and below it creates the directories.

es.gob.afirma.yaml
ivan-gj: It is required because the core purpose of Autofirma is to sign arbitrary files on the user's system. Since the files to be signed can be located anywhere on the filesystem, and Autofirma does not use any XDG portals for file access, we cannot restrict this to a narrower path.

Should we do the request for the exception now?

trinukso: I'm a user and I'm not familiar with the specifics of the host permission. Wouldn't access to the user's /home directory be enough, or just ‘Downloads’ or “Documents”?

ivan-gj: We could limit it to home, but people would probably freak out if they couldn't reach or save files on other directories. I think it's worth the "huge" warning on its future Flathub's page. We'll try to push the Spanish government to follow XDG standards, but it'll take months/years.

es.gob.afirma.yaml
ivan-gj: It was a remnant from original tweakings trying to make things work. Thanks!

es.gob.afirma.yaml
trinukso: In my opinion, “home” is enough. If any user gets angry about this, they should understand that it’s Flatpak and a poorly designed app that autofirma, and that there’s little to no effort to integrate it with the Linux desktop.

trinukso: I, for one, would limit it to Documents and Downloads. When I log in to the website, I don't really know where those files are. I honestly don't understand that part.

es.gob.afirma.yaml
ivan-gj: Fixed in [617e6ba](https://github.com/flathub/flathub/pull/7673/commits/617e6baaa42a9a7f5c77cf0768e4bab1f071dbdd)

es.gob.afirma.yaml
ivan-gj: Here's the PR for the required exceptions: https://github.com/flathub-infra/flatpak-builder-lint/pull/978

ivan-gj:
Hello!

Thank you for your patience and feedback. We have addressed the review comments and made a number of improvements:

- We found official upstream documentation covering the specific use cases, which allowed us to greatly simplify the repository — it now contains little more than a README linking to the upstream docs.
- We significantly reduced the amount of downstream code by reusing more of the original launcher.
- The manifest has been updated to use the requested `install` commands.
- We have submitted the [exceptions PR to flatpak-builder-lint](https://github.com/flathub-infra/flatpak-builder-lint/pull/978).

Please let us know if there is anything else you would like us to address — we are happy to make further changes.

bbhtt:
bot, build

PR #7717, opened 2026-02-01
Add io.github.qarmin.krokiet

qarmin:

- [X] Please describe the application briefly
Krokiet is an application, that finds duplicates, similar images, big files, similar videos and many more.
It is a new frontend for the Czkawka project, an alternative to the current Czkawka GTK (which is now in maintenance mode, app already is on Flathub: https://flathub.org/en/apps/com.github.qarmin.czkawka).

https://github.com/user-attachments/assets/aa74cccf-aa56-482a-b638-1cc728e437c2

### DIFF
--- a/io.github.qarmin.czkawka.krokiet.yml
+++ b/io.github.qarmin.czkawka.krokiet.yml
@@ -1,0 +1,41 @@
+app-id: io.github.qarmin.czkawka.krokiet
+runtime: org.freedesktop.Platform
+runtime-version: '25.08'
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+command: krokiet
+finish-args:
+- "--share=ipc"
+- "--socket=fallback-x11"
+- "--socket=wayland"
+- "--filesystem=host"
+- "--device=dri"
+build-options:
+  append-path: "/usr/lib/sdk/rust-stable/bin"
+  env:
+    CARGO_HOME: "/run/build/krokiet/cargo"
+modules:
+- name: krokiet
+  buildsystem: simple
+  build-commands:
+  - cargo --offline fetch --manifest-path Cargo.toml
+  # It appears that all GUI backends provided by slint have some issues:
+  # - winit_software - resizing the app breaks it, will be fixed in Slint 1.15 and Krokiet 11.0 - https://github.com/slint-ui/slint/issues/7663
+  # - winit_femtovg - blurry fonts - https://github.com/slint-ui/slint/issues/6215
+  # - winit_skia_opengl / winit_skia_vulkan - downloads files during the build - this may be a solution, but too hard for now, for me
+  # - femtovg_wgpu - does not support OpenGL, so it will not work on a large portion of computers
+  - cargo --offline build --release --features "heif, libavif, libraw, winit_femtovg" --no-default-features --bin krokiet
+  - install -Dm755 ./target/release/krokiet -t /app/bin/
+  - install -Dm644 ./krokiet/LICENSE_CC_BY_4_ICONS /app/share/licenses/io.github.qarmin.czkawka.krokiet/LICENSE_CC_BY_4_ICONS
+  - install -Dm644 ./krokiet/LICENSE_GPL_APP /app/share/licenses/io.github.qarmin.czkawka.krokiet/LICENSE_GPL_APP
+  - install -Dm644 ./krokiet/LICENSE_MIT_CODE /app/share/licenses/io.github.qarmin.czkawka.krokiet/LICENSE_MIT_CODE
+  - install -Dm644 ./data/icons/io.github.qarmin.czkawka.krokiet.svg -t /app/share/icons/hicolor/scalable/apps/
+  - install -Dm644 ./data/io.github.qarmin.czkawka.krokiet.desktop -t /app/share/applications/
+  - install -Dm644 ./data/io.github.qarmin.czkawka.krokiet.metainfo.xml -t /app/share/metainfo
+  sources:
+  - cargo-sources.json
+  - type: git
+    url: https://github.com/qarmin/czkawka.git
+    tag: 10.0.0_flatpak
+    commit: e33ada1a5f2ef1876f5086ecfc0367632e2bbece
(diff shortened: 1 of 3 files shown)


io.github.qarmin.krokiet.yml
```diff
@@ -0,0 +1,41 @@
...
+runtime-version: '25.08'
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+command: krokiet
+finish-args:
+- "--share=ipc"
+- "--socket=fallback-x11"
+- "--socket=wayland"
+- "--filesystem=host"
```
bbhtt, line 12: this needs an [exception](https://docs.flathub.org/docs/for-app-authors/linter#exceptions)

io.github.qarmin.czkawka.krokiet.yml
```diff
@@ -0,0 +1,41 @@
+app-id: io.github.qarmin.czkawka.krokiet
```
bbhtt: this is an invalid ID.

io.github.qarmin.krokiet.yml
```diff
@@ -0,0 +1,41 @@
...
+  - install -Dm644 ./krokiet/LICENSE_CC_BY_4_ICONS /app/share/licenses/io.github.qarmin.czkawka.krokiet/LICENSE_CC_BY_4_ICONS
+  - install -Dm644 ./krokiet/LICENSE_GPL_APP /app/share/licenses/io.github.qarmin.czkawka.krokiet/LICENSE_GPL_APP
+  - install -Dm644 ./krokiet/LICENSE_MIT_CODE /app/share/licenses/io.github.qarmin.czkawka.krokiet/LICENSE_MIT_CODE
+  - install -Dm644 ./data/icons/io.github.qarmin.czkawka.krokiet.svg -t /app/share/icons/hicolor/scalable/apps/
+  - install -Dm644 ./data/io.github.qarmin.czkawka.krokiet.desktop -t /app/share/applications/
+  - install -Dm644 ./data/io.github.qarmin.czkawka.krokiet.metainfo.xml -t /app/share/metainfo
+  sources:
+  - cargo-sources.json
+  - type: git
+    url: https://github.com/qarmin/czkawka.git
```
bbhtt, line 39: this and https://github.com/flathub/com.github.qarmin.czkawka/blob/81b2ebd01554e194392a71f4438e93ee2c2d00ef/com.github.qarmin.czkawka.yaml#L31 shares the same source

what is meant by "new frontend"? why should this be a separate submission and not an update over the existing one?

io.github.qarmin.czkawka.krokiet.yml
qarmin: According to this doc line `The ID must not exceed 255 characters and must have at least 3 components. Applications must not exceed 5 components`, I assumed that this ID was valid.

I initially tried `io.github.qarmin.krokiet`, but I encountered an issue where, during the build process (or a linter check), `https://github.com/qarmin/krokiet` was queried. Since this repository does not exist, the check/build failed. When I switched to a longer ID the build succeeded.

I changed id to shorter version and created an empty repository at that location, so with this changes, the compilation appears to work correctly.

io.github.qarmin.krokiet.yml
qarmin: The Czkawka repository on GitHub is a monorepo that contains applications (CLI, GTK GUI, and Krokiet) that are based on a core library and have more or less similar features, but different user interfaces.

From the beginning of the project, I focused on the GTK version (the Czkawka application available on Flathub). However, due to bugs and various issues, I decided to move from GTK to Slint.

Despite generally having fewer issues and over two years of development, I was not able to unify the feature lists between these two applications (partly due to my own decisions and lack of time, and partly due to limitations of Slint, which is a relatively new framework).

Currently, the user base is split, with some users using Czkawka and others using Krokiet. Unification remains my goal, but at this point replacing Czkawka with Krokiet would likely result in negative feedback from users (for example, those who rely on the image comparison feature), so I would not replace these two tools yet.

io.github.qarmin.czkawka.krokiet.yml
hfiguiere: selectively reading the documentation will sure work.

Because if you read this:

https://docs.flathub.org/docs/for-app-authors/requirements#code-hosting

You'd know what to use.

io.github.qarmin.krokiet.yml
bbhtt: sounds like this should wait for feature parity and be an update over the existing app instead of a new one.

This falls under https://docs.flathub.org/docs/for-app-authors/requirements#duplicate-submissions

qarmin: Feature parity may be achieved in the next 3 months or in 3 years, because I am not sure whether, or how, Slint will allow me to implement this, nor whether I will have enough time in the near future.

If only one option is possible, I would choose this one, but without name continuity(so new id, instead using flathub czkawka id). In that case, I would need to add a popup to the old version informing users that they need to migrate to the new application

qarmin: I added a warning to the Czkawka build, so I think it should now be safe to hide or remove `com.github.qarmin.czkawka` and replace it with this package.

Do I need to take any further action, or can this be handled directly in this PR?

bbhtt: why do you want to replace the existing working one with something that hasn't achieved feature parity to that? it sounds like a downgrade for users.

qarmin: Maybe for some users it could feel like a downgrade, but only in specific areas. The new GUI fixes several visual inconsistencies (e.g. light mode is actually fully light instead of partially dark due to limitations in styling certain components in gtk), includes more tools and does not freeze the entire interface during long operations such as deleting or copying files.

I still provide prebuilt binaries of the gtk version, and it remains available through other unofficial distribution channels, so that does not mean the app will not be available, it simply means the new gui is currently the supported and recommended version.

bbhtt: Ok we can accept this if you want to [EOL](https://docs.flathub.org/docs/for-app-authors/maintenance#end-of-life) the old one.

Is that going to be ok?

PR #7752, opened 2026-02-05
Add org.peergos.Peergos

ianopolous:

- [X] Please describe the application briefly. < Peergos is a file storage, sync, sharing and application protocol. >
- [X] Please attach a video showcasing the application on Linux using the Flatpak.  < https://peergos.net/secret/z59vuwzfFDovgun2sU9YF8LqJRVXaoVR39XAkvRR6sNp7CJnseecHhV/300913182#xk7vzJaNDAWy?open=true >
- [X] I am a developer to the project.

### DIFF
--- a/org.peergos.Peergos.yaml
+++ b/org.peergos.Peergos.yaml
@@ -20,7 +20,11 @@ finish-args:
   - --socket=pulseaudio
   - --socket=fallback-x11
   - --socket=wayland
+  - --env=LD_LIBRARY_PATH=/app/chromium/nonfree-codecs/lib
   - --env=PATH=/app/jre/bin:/app/bin:/usr/bin
+
+inherit-extensions:
+  - org.chromium.Chromium.Codecs
 
 build-options:
   append-path: /usr/lib/sdk/openjdk21/bin
@@ -32,6 +36,11 @@ cleanup:
   - "/app/bin/ant"
 
 modules:
+  - name: chromium-codecs-dir
+    buildsystem: simple
+    build-commands:
+      - mkdir -p /app/chromium/nonfree-codecs/lib
+
   - name: openjdk
     buildsystem: simple
     build-commands:


ianopolous:
The build is erroring with appstream-missing-screenshots, but the screenshots are in the metainfo file and the urls are public and accessible. I'm not sure what the problem is?

org.peergos.Peergos.desktop
hfiguiere, line 1: This file should be (submitted and managed) upstream.

org.peergos.Peergos.metainfo.xml
hfiguiere, line 1: This file should be (submitted and managed) upstream.

peergos.svg
hfiguiere, line 1: This file should be (submitted and managed) upstream.

org.peergos.Peergos.yaml
```diff
@@ -0,0 +1,79 @@
...
+  env:
+    JAVA_HOME: /usr/lib/sdk/openjdk21
+
+modules:
+  - name: openjdk
+    buildsystem: simple
+    build-commands:
+      - /usr/lib/sdk/openjdk21/install.sh
+
+  - name: ant
```
hfiguiere, line 49: if this is required at build time only then it should be cleaned up.

ianopolous: Yes it's only required at build time. Does that mean deleting it?

org.peergos.Peergos.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+  - name: peergos
+    buildsystem: simple
+    build-commands:
+      - export JAVA_HOME=/app/jdk
+      - export PATH=$JAVA_HOME/bin:$PATH/app/jdk
+
+      - ant dist
+      - rm -rf /app/ant
+      - rm /app/bin/ant
```
hfiguiere: this is not what I meant. Check `cleanup`

hfiguiere:
This doesn't have a UI. It just spawn a web browser. This is not suitable for flathub.

https://docs.flathub.org/docs/for-app-authors/requirements#impermissible-submissions

ianopolous:
The ui is to a localhost server. We could try and package a webview but that will be strictly less secure than the user's browser which is updated faster and likely more often.  Apart from serving the ui, the localhost server also handles the sync client and caching data for offline usage.

org.peergos.Peergos.yaml
ianopolous: Sorry, I don't understand. Is there a url I can read more?

ianopolous:
There are apps using the same technique of a browser for a ui in flatpak. For example https://flathub.org/en/apps/dev.lizardbyte.app.Sunshine

hfiguiere:
Two wrong don't make a right. "Somebody else does it" will NEVER plead in your favour. Quite the opposite.

ianopolous:
Sorry. I'm not trying to be antagonistic. I understand you don't want apps that just wrap a website - that was my reading of the docs. The other app I took as evidence that using the browser for a local UI is okay.

ianopolous:
I've updated the video link above to show the new UI with its own window. Hopefully this is ok?

org.peergos.Peergos.yaml
```diff
@@ -0,0 +1,75 @@
...
+      - type: git
+        url: http://github.com/peergos/web-ui.git
+        commit: 89131f546bbf229b4b5c647c23e542eb6c85f8dc
+
+      - type: file
+        path: peergos.sh
+
+cleanup:
+  - "/app/ant"
+  - "/app/bin/ant"
```
hfiguiere: I guess you never verified this work....

org.peergos.Peergos.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+    sources:
+      - type: git
+        url: http://github.com/peergos/web-ui.git
+        commit: 89131f546bbf229b4b5c647c23e542eb6c85f8dc
+
+      - type: file
+        path: peergos.sh
+
+cleanup:
```
hfiguiere: cleanup should be somewhere above modules.

ianopolous:
The linter says I'm not allowed to use the flatpak spawn access. I'm open to other ways to do this, but nothing I've tried has worked. I tried using chromium as a base, then i can run chromium --app=http://localhost:7777. That works, except it doesn't have the media codecs like h264 which I need. I also tried using com.google.Chrome as a base, but couldn't get that to run chrome at all.

hfiguiere:
This comment still applies

https://github.com/flathub/flathub/pull/7752#issuecomment-3853854564

ianopolous:
> This comment still applies
> 
> [#7752 (comment)](https://github.com/flathub/flathub/pull/7752#issuecomment-3853854564)

The docs say "simple web wrappers" are not allowed. This is not a simple web wrapper. The native app outside the browser is doing a lot: https://github.com/peergos/peergos

Are electron apps not allowed either?

rapenne-s:
> This comment still applies
> 
> [#7752 (comment)](https://github.com/flathub/flathub/pull/7752#issuecomment-3853854564)

Peergos web interface features local cache and embedded web server and assets, this allows users to not trust their peergos provider that could tamper with the login ui if you trust the peergos build you run locally. This is really more than a web browser showing a remote website.

this local web ui also permits users to configure directories sync with their peergos provider, otherwise there is only a complicated CLI interface for this synchronisation mechanism, it’s a must to have it easy to use on the web ui, only available when running peergos on the local machine.

ianopolous:
The way this app works is we run a localhost server that serves the user interface, and handles api calls and caches data locally, and runs the folder sync periodically. It then opens a webview to itself, such that if either the webview or the server process are killed the whole app stops. It has a desktop file and icon for a menu shortcut. It uses zenity to let user choose a host dir to sync when creating a new sync. After creating a sync, then the app (whilst it is running) keeps a local folder in sync with an encrypted folder on your Peergos server. As far as I can tell, with my newcomer understanding, this meets all the criteria for flathub. 

Can you explain to me why this doesn't meet the criteria? It seems to me to be semantically identical to an electron app for example, which seems to be explicitly supported by flathub, based on https://github.com/flathub/org.flathub.electron-sample-app.

PR #7791, opened 2026-02-10
Add app.yaak.Yaak

gschier:

- [x] Please describe the application briefly. An offline, Git friendly API Client (Postman alternative)
- [x] I am the author to the project.

There are some lint warnings due to permissions. I added comments in the manifest explaining what each is for. I went off Gitbutler as an example (as Yaak also integrates with Git).

There is also a pre-existing app ID `app.yaak.desktop` compiled into the binary that also caused a lint warning. I cannot change this, as it would break Yaak for existing users. However, I did change to app.yaak.Yaak for the flatpak metadata.

https://github.com/user-attachments/assets/991b5c14-b037-4974-91bb-ac87c3526937
      


app.yaak.Yaak.yml
```diff
@@ -0,0 +1,64 @@
...
+      - mv usr/bin/* /app/bin
+      - mv usr/lib/* /app/lib
+      - mv usr/share/* /app/share
+      - install -Dm644 app.yaak.Yaak.metainfo.xml /app/share/metainfo/app.yaak.Yaak.metainfo.xml
+      - install -Dm644 LICENSE /app/share/licenses/app.yaak.Yaak/LICENSE
+
+    sources:
+      - type: file
+        dest-filename: yaak.deb
+        url: https://github.com/mountain-loop/yaak/releases/download/v2026.1.2/yaak_2026.1.2_amd64.deb
```
hfiguiere: This should be built from source.

app.yaak.Yaak.yml
```diff
@@ -0,0 +1,64 @@
+id: app.yaak.Yaak
+runtime: org.gnome.Platform
+runtime-version: "48"
```
hfiguiere: current is 49

app.yaak.Yaak.metainfo.xml
hfiguiere, line 1: This file should be (submitted and managed) upstream.

app.yaak.Yaak.yml
gschier: Hmm, I tried getting it to work but it appears `flatpak-node-generator` doesn't support vanilla npm workspaces, which Yaak uses.

gschier: For context, it looks like conflicting module versions in workspace packages don't get the `resolve: <url>` property in the root `package-lock.json` so are silently skipped.

gschier: On further inspection, it might just be an [npm issue](https://github.com/npm/cli/issues/4460#issuecomment-2532984556). Have you heard of this before? Going to try patching with that CLI for now.

gschier: Took quite a bit of work but things are working now using build-from-source

app.yaak.Yaak.yml
gschier: Fixed

app.yaak.Yaak.metainfo.xml
gschier: Now upstream tracks this and it's pulled in during the build

app.yaak.Yaak.yml
```diff
@@ -83,7 +83,7 @@ modules:
       - type: git
         url: https://github.com/mountain-loop/yaak.git
         tag: v2026.2.0
-        commit: 9a1d6130346454a9eb07388e65f26076f2e0f59b
+        commit: 26aba6034fe43eb0f2f5bfde6e41896a71bbae88
```
hfiguiere, line 83: retagging is bad.

app.yaak.Yaak.yml
```diff
@@ -0,0 +1,136 @@
...
+
+      # Build scripts
+      - type: file
+        path: patch-lockfile.cjs
+      - type: file
+        path: generate-desktop.cjs
+
+      # Vendored Node.js binary (x86_64)
+      - type: archive
+        url: https://nodejs.org/download/release/v24.11.1/node-v24.11.1-linux-x64.tar.gz
```
hfiguiere: node is in the SDK (extension)

app.yaak.Yaak.yml
```diff
@@ -0,0 +1,136 @@
...
+        url: https://nodejs.org/download/release/v24.11.1/node-v24.11.1-linux-arm64.tar.gz
+        sha256: 0dc93ec5c798b0d347f068db6d205d03dea9a71765e6a53922b682b91265d71f
+        strip-components: 1
+        dest: vendored-node
+        only-arches:
+          - aarch64
+
+      # Vendored protoc binary and includes (x86_64)
+      - type: archive
+        url: https://github.com/protocolbuffers/protobuf/releases/download/v33.1/protoc-33.1-linux-x86_64.zip
```
hfiguiere: 1. it appear to be build only so it should be cleaned up.
2. why not building from source as well?

app.yaak.Yaak.yml
```diff
@@ -0,0 +1,136 @@
...
+      - type: archive
+        url: https://github.com/protocolbuffers/protobuf/releases/download/v33.1/protoc-33.1-linux-aarch_64.zip
+        sha256: 6018147740548e0e0f764408c87f4cd040e6e1c1203e13aeacaf811892b604f3
+        dest: protoc-bin
+        only-arches:
+          - aarch64
+
+      # License file
+      - type: file
+        path: LICENSE
```
hfiguiere: you don't need this, it will be picked up and automatically installed from the source tree.

app.yaak.Yaak.yml
```diff
@@ -0,0 +1,136 @@
...
+      - type: file
+        path: patch-lockfile.cjs
+      - type: file
+        path: generate-desktop.cjs
+
+      # Vendored Node.js binary (x86_64)
+      - type: archive
+        url: https://nodejs.org/download/release/v24.11.1/node-v24.11.1-linux-x64.tar.gz
+        sha256: 58a5ff5cc8f2200e458bea22e329d5c1994aa1b111d499ca46ec2411d58239ca
+        strip-components: 1
```
hfiguiere:
```suggestion
```

app.yaak.Yaak.yml
gschier: It's not build-only. Yaak includes it as a sidecar so it can do protobuf stuff for gRPC

app.yaak.Yaak.yml
gschier: Yaak bundles node as a sidecar to run its plugin runtime.

app.yaak.Yaak.yml
gschier: had to do it to get this PR ready to go. Won't be doing any retagging going forward.

app.yaak.Yaak.yml
gschier: Needed to make extraction easier for the node sidecar (otherwise the path will be different per arch)

hfiguiere: it's the default.

app.yaak.Yaak.yml
hfiguiere: build it from source.

app.yaak.Yaak.yml
```diff
@@ -0,0 +1,131 @@
...
+        dest: vendored-node
+        only-arches:
+          - x86_64
+
+      # Vendored Node.js binary (aarch64)
+      - type: archive
+        url: https://nodejs.org/download/release/v24.11.1/node-v24.11.1-linux-arm64.tar.gz
+        sha256: 0dc93ec5c798b0d347f068db6d205d03dea9a71765e6a53922b682b91265d71f
+        strip-components: 1
```
hfiguiere:
```suggestion
```

app.yaak.Yaak.yml
hfiguiere: and build it from source.

gschier: Building both Node and Protoc would be a significant amount of work. They are dependencies of Yaak, similar to how Electron apps bundle the Node+Chromium runtime.

They _are_ technically optional, however, so I already have a task on the roadmap to fetch these binaries lazily when the user requires them (similar to VSCode, Zed, etc). Should I prioritize that instead?

bbhtt: node24 is available as an sdk extension.

app.yaak.Yaak.yml
Keshav-writes-code: > looks like conflicting module versions in workspace packages do

i encountred simmilar situation but was able to fix it my generating the package-lock.json with this 
```sh
npm i --package-lock-only
```

app.yaak.Yaak.yml
gschier: Ended up removing `protoc` from the Flatpak build (too difficult to compile from source). Is there a way to have the user bring their own (point to installation path) if they want to use it?

app.yaak.Yaak.yml
gschier: I updated the manifest to vendor the node version from the sdk extension.

app.yaak.Yaak.yml
```diff
@@ -0,0 +1,95 @@
...
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc
+  - --device=dri
+  - --share=network
+  - --socket=pulseaudio # Preview audio responses
+  - --socket=ssh-auth # Git SSH remotes
+  - --socket=gpg-agent # Git commit signing
+  - --talk-name=org.freedesktop.secrets # Keyring for encryption
+  - --filesystem=home # Git repos, ~/.gitconfig, ~/.ssh, etc
```
bbhtt, line 24: why does a api client need access to git repos?

if only .gitconfig, .ssh access is needed it should be those only and in read-only mode

app.yaak.Yaak.yml
```diff
@@ -0,0 +1,95 @@
...
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc
+  - --device=dri
+  - --share=network
+  - --socket=pulseaudio # Preview audio responses
+  - --socket=ssh-auth # Git SSH remotes
+  - --socket=gpg-agent # Git commit signing
+  - --talk-name=org.freedesktop.secrets # Keyring for encryption
```
bbhtt, line 23: why does this not use the portal?

app.yaak.Yaak.yml
```diff
@@ -0,0 +1,95 @@
...
+
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --share=ipc
+  - --device=dri
+  - --share=network
+  - --socket=pulseaudio # Preview audio responses
+  - --socket=ssh-auth # Git SSH remotes
+  - --socket=gpg-agent # Git commit signing
```
bbhtt, line 22: i thought it is an API client why is it signing git commits?

lostfictions:
I would love to see this app on Flathub; I'm a little bummed that I poked around in issues to see if it was being submitted and found out that @gschier appears to have dropped this after being subjected to four (!) rounds of review. As a developer who had been planning to submit some apps to Flathub myself, it's pretty discouraging to see this.

(As a side note, as I understand Flathub rules have changed and some apps are grandfathered into more lax submission rules, but it's a little galling to check [the manifest for a similar app](https://github.com/flathub/com.usebruno.Bruno/blob/master/com.usebruno.Bruno.yml) and discover there's no issue that it's not built from source, nor that it has full home directory access.)

I'd be happy to help push this across the finish line if needed!

app.yaak.Yaak.yml
lostfictions: The app [integrates Git functionality](https://yaak.app/docs/collaboration/local-directory-git-sync#enabling-git-for-a-workspace) (this is also why the manifest builds Git from source). Agree that this should probably be `~/.gitconfig` and `~/.ssh:ro` though.

app.yaak.Yaak.yml
lostfictions: What would be the correct way to do this?

app.yaak.Yaak.yml
lostfictions: See above, the app integrates Git client functionality.

bbhtt:
We aren't basing descisions on what you think or don't think. If you want to work on it feel free to open a new PR, with the issues from here resolved. If you don't want to then the comment is noise in this PR.

app.yaak.Yaak.yml
bbhtt: Use the secrets portal.

KraXen72:
Hi! any progress on this? would love to see yaak on flatpak! thanks! :)

gschier:
I've decided to abandon this effort as I'm not able to justify the amount of work (and likely ongoing maintenance) required as solo, indie project maintainer. Thank you to the reviewers for putting in the time to review and answer questions.

PR #7806, opened 2026-02-12
Add net.froemling.bombsquad

Loup-Garou911XD:

Bombsquad is a party game u can play solo or with friends.

https://github.com/user-attachments/assets/24857755-c364-4fb3-8887-ea7a4c8e3bf3

@efroemling is the developer and maintainer of the [project](https://github.com/efroemling/ballistica)


Loup-Garou911XD:
bot, build

Loup-Garou911XD:
bot, build

Loup-Garou911XD:
sorry for the failing build, i didnt test it for aarch64, will update this soon

Loup-Garou911XD:
> The domain to be used for verification is froemling.net. If you intend to [verify](https://docs.flathub.org/docs/for-app-authors/verification) this submission, please confirm by uploading an empty `org.flathub.VerifiedApps.txt` file to https://froemling.net/.well-known/org.flathub.VerifiedApps.txt. Otherwise, ignore this. Please comment if this incorrect.

@efroemling please see this

net.froemling.bombsquad.appdata.xml
CodedOre, line 1: This file should be upstream.

Also, the extension `appdata.xml` is the legacy extension, you should use `metainfo.xml` instead.

net.froemling.bombsquad.desktop
CodedOre, line 1: This file should come from upstream.

net.froemling.bombsquad.releases.xml
CodedOre, line 1: This file should come from upstream.

net.froemling.bombsquad.yml
```diff
@@ -0,0 +1,81 @@
...
+sdk: org.freedesktop.Sdk
+command: bombsquad.sh
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=x11
+  - --socket=pulseaudio
+  - --device=dri
+#  - --device=input
+  - --filesystem=~/.ballisticakit:create
```
CodedOre: Why this filesystem permission and not a `persists`?

net.froemling.bombsquad.yml
```diff
@@ -0,0 +1,81 @@
...
+    build-commands:
+      - mkdir -p ${FLATPAK_DEST}/bin/bombsquad
+      - rm -f .venv/bin/python3.13
+      - ln -s /usr/bin/python3.13 .venv/bin/python3.13
+      - python3.13 -m venv --upgrade .venv
+      - make env-clean
+      - touch ballisticakit-windows/Generic/BallisticaKit.ico
+      - make cmake-build
+      - mv build/cmake/debug/ballisticakit build/cmake/debug/staged/
+      - cp -r build/cmake/debug/staged/* ${FLATPAK_DEST}/bin/bombsquad/
```
CodedOre: Are you building a debug build?
This should be a release build then.

net.froemling.bombsquad.yml
```diff
@@ -0,0 +1,81 @@
...
+        path: net.froemling.bombsquad.appdata.xml
+
+      - type: file
+        path: net.froemling.bombsquad.releases.xml
+
+      - type: script
+        dest-filename: bombsquad.sh
+        commands:
+          - cd /app/bin/bombsquad
+          - exec ./ballisticakit
```
CodedOre, line 59: Is there a reason you need this (very small) script and can't put `ballisticakit` as the Flatpak command directly?

net.froemling.bombsquad.yml
```diff
@@ -0,0 +1,81 @@
...
+      - type: archive
+        url: https://download.samba.org/pub/rsync/src/rsync-3.4.1.tar.gz
+        sha256: 2924bcb3a1ed8b551fc101f740b9f0fe0a202b115027647cf69850d65fd88c52
+    cleanup:
+      - /share/man
+      - /share/doc
+      - /bin
+
+  - name: bombsquad
+    buildsystem: simple
```
CodedOre, line 20: Why can't `buildsystem: cmake-ninja` be used directly, if you use CMake to build?

CodedOre (changes requested):
Your build system attempts to download files in the build stage.

This is not permitted on Flathub. All sources must be downloaded by Flatpak Builder ahead of build:

https://docs.flathub.org/docs/for-app-authors/requirements#no-network-access-during-build

net.froemling.bombsquad.releases.xml
```diff
@@ -0,0 +1,62 @@
+<?xml version='1.0' encoding='utf-8'?>
+<releases>
+    <release version="1.7.60" date="2026-02-11" urgency="low" type="stable">
+        <description>
+            <p>- Fixed a longstanding issue causing impact, roll, and skid sounds to not
+  function.
+- Fun easter-egg when clicking characters in inventory window (Thanks
+  EraOSBeta!)
```
bbhtt: Use li tags for these bullet points https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines#description

net.froemling.bombsquad.yml
```diff
@@ -0,0 +1,81 @@
...
+    build-options:
+      append-path: /usr/lib/sdk/llvm20/bin
+      prepend-ld-library-path: /usr/lib/sdk/llvm20/lib
+    build-commands:
+      - mkdir -p ${FLATPAK_DEST}/bin/bombsquad
+      - rm -f .venv/bin/python3.13
```
bbhtt, line 30: this is not supposed to exist beforehand so I don't see the point of the `rm -f`?

net.froemling.bombsquad.yml
```diff
@@ -0,0 +1,81 @@
...
+      - ln -s /usr/bin/python3.13 .venv/bin/python3.13
+      - python3.13 -m venv --upgrade .venv
```
bbhtt, line 32: what is the point of creating a venv? there is only one python in the runtime and set of dependencies you are going to use.

net.froemling.bombsquad.desktop
Loup-Garou911XD: it does come from upstream, i just use a PAT to push it to efroemling/flathub
does it explicitly need to be added as a source?

CodedOre: If the file is in this PR then it is not coming from the upstream source. 

It should come with the source code from the upstream repository.

hfiguiere: what's a PAT?

Yes the file doesn't belong here.

net.froemling.bombsquad.yml
```diff
@@ -0,0 +1,81 @@
...
+      - install -Dm755 bombsquad.sh ${FLATPAK_DEST}/bin/bombsquad.sh
+      - install -Dm644 ${FLATPAK_ID}.desktop ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop
+      - install -Dm644 ${FLATPAK_ID}.png ${FLATPAK_DEST}/share/icons/hicolor/512x512/apps/${FLATPAK_ID}.png
+      - install -Dm644 ${FLATPAK_ID}.appdata.xml ${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.metainfo.xml
+      - install -Dm644 ${FLATPAK_ID}.releases.xml ${FLATPAK_DEST}/share/metainfo/releases/${FLATPAK_ID}.releases.xml
+
+    sources:
+
+      - type: archive
+        url: https://github.com/efroemling/ballistica/releases/download/v1.7.60/linux_build_env.tar
```
hfiguiere: What happen on aarch64?

net.froemling.bombsquad.yml
```diff
@@ -0,0 +1,81 @@
...
+        sha256: 409727c39713b7382f2a6b0627b49f8923006659a911668f739e4272c48a21b1
+        strip-components: 0
+
+      - type: file
+        dest-filename: net.froemling.bombsquad.png
+        url: https://files.ballistica.net/bombsquad/promo/BombSquadIcon512.png
```
hfiguiere, line 52: why isn't this in the tarball?

net.froemling.bombsquad.yml
Loup-Garou911XD: > what is the point of creating a venv? there is only one python in the runtime and set of dependencies you are going to use.

well the project uses venv, and i just zip up the entire project dir to keep the dependencies, but then the venv gets broken because of wrong python path, so i just fix it, all the python dependencies are installed in a venv

net.froemling.bombsquad.yml
Loup-Garou911XD: U  mean why the build failed?
it needed a precompiled binary and i didnt test for aarch64 so i didnt know it was going to fail

net.froemling.bombsquad.yml
Loup-Garou911XD: it was easier this way, should i package it in there?

net.froemling.bombsquad.releases.xml
Loup-Garou911XD: yeah thanks for the docs, i tried br tag and that didnt work, will fix with li tag now

Loup-Garou911XD:
Just to clear up a few things, the https://github.com/efroemling/ballistica/releases/download/v1.7.60/linux_build_env.tar is the entire project dir with all the build dependencies, that includes assets for the game, binaries and python venv(which gets mangled up and thats is why i relink it to the python)

net.froemling.bombsquad.yml
Loup-Garou911XD: because bombsquad supports mods and we can put mods in ~/.ballisticakit/mods
and ~/.ballisticakit also contains replay files which are nice to have access to

hfiguiere:
so it's a case of "works on my machine". 

This fall into this case:

https://docs.flathub.org/docs/for-app-authors/requirements#unacceptable-software-design-practices

net.froemling.bombsquad.yml
Loup-Garou911XD: yeah uhh that is needed because ballisticakit needs to be run from cwd

net.froemling.bombsquad.yml
CodedOre: Archiving a build environment from one random system to then use it on another system sounds like massive hack to me, not like a good solution. 

If you need Python dependencies inside a Flathub build, use the [Flatpak Pip Generator](https://github.com/flatpak/flatpak-builder-tools/tree/master/pip).

Loup-Garou911XD:
> so it's a case of "works on my machine".
> 
> This fall into this case:
> 
> https://docs.flathub.org/docs/for-app-authors/requirements#unacceptable-software-design-practices

i am sorry if i didnt do anything correctly, but the tarball is missing just one binary for aarch64, which i fixed, just need to update the tarball

net.froemling.bombsquad.yml
CodedOre: Wait, what?

So, normally, this program could only be directly launched from the terminal? So, not from an app launcher with the desktop file, etc.?

This sounds like bad code design on the first look...

net.froemling.bombsquad.desktop
Loup-Garou911XD: > PAT

Personal access token, i used that to grant the upstream to push changes to efroemling/falthub
just to automate things

Loup-Garou911XD: https://github.com/efroemling/ballistica/blob/7771feb6ad4ee327b2256d8261a52a24c69be5ce/.github/workflows/release.yml#L535-L546
this is where the file gets pushed from upstream to this fork

hfiguiere: the file should be in the tarball. The only thing that should be here (in this repo) is the manifest. What you are basically doing is something over complicated to avoid doing the simple thing.

Loup-Garou911XD:
so from what i understand is packaging the venv is bad practice, is there a case where relinking the venv python wont work?
btw there are just build dependencies

net.froemling.bombsquad.desktop
Loup-Garou911XD: > the file should be in the tarball. The only thing that should be here (in this repo) is the manifest. What you are basically doing is something over complicated to avoid doing the simple thing.

alright, will fix that

net.froemling.bombsquad.yml
CodedOre: That does not justify a filesystem permission though. 

I can't think of a good reason why it can't use `persist=.ballisticakit` instead and store the data in it's own application data folder under `~/.var/app/net.froemling.Bombsquad/.ballisticakit`.

With the filesystem permission, you pollute the home folder of the user and the data can't be removed together with the application. 

Also, on a sidenote, but the XDG base directory specification has existed for quite some time now and applications should ideally use `XDG_DATA_HOME` for folders like this, rather than create dot-folders in the users home. 
But that's beside the point right now.

net.froemling.bombsquad.yml
Loup-Garou911XD: > Wait, what?
> 
> So, normally, this program could only be directly launched from the terminal? So, not from an app launcher with the desktop file, etc.?
> 
> This sounds like bad code design on the first look...

well it can, would just need a `Path=my_bombsquad_dir` in the .desktop file
its a upstream issue that i cant fix

Loup-Garou911XD: ig i could put that in the desktop file instead of having the the sh script do that

net.froemling.bombsquad.yml
Loup-Garou911XD: it would be a fair bit more complicated 😅 , the make target does some other stuff aswell

net.froemling.bombsquad.yml
Loup-Garou911XD: > I can't think of a good reason why it can't use `persist=.ballisticakit` instead and store the data in it's own application data folder under `~/.var/app/net.froemling.Bombsquad/.ballisticakit`.

Its just because so that user can conveniently access it from the filesystem 
i dont believe ~/.var/app/net.froemling.Bombsquad/.ballisticakit would be accessible from filesystem right?

Loup-Garou911XD: > Also, on a sidenote, but the XDG base directory specification has existed for quite some time now and applications should ideally use `XDG_DATA_HOME` for folders like this, rather than create dot-folders in the users home.
> But that's beside the point right now.

Thanks, will relay this upstream
although i notice its not set on my system, arch linux btw XDD

net.froemling.bombsquad.yml
```diff
@@ -0,0 +1,81 @@
...
+  - --device=dri
+#  - --device=input
+  - --filesystem=~/.ballisticakit:create
+
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.llvm20
+
+modules:
+
+  - name: rsync
```
hfiguiere: why is rsync needed? I don't see where it is used.

net.froemling.bombsquad.yml
```diff
@@ -0,0 +1,81 @@
...
+    buildsystem: autotools
+    config-opts:
+      - --prefix=${FLATPAK_DEST}
```
hfiguiere:
```suggestion
```

net.froemling.bombsquad.yml
CodedOre: > Its just because so that user can conveniently access it from the filesystem

Not enough of a reason to make a hole into the sandbox.

> i dont believe ~/.var/app/net.froemling.Bombsquad/.ballisticakit would be accessible from filesystem right?

Define "filesystem". For the users, they could access this path as it is mentioned from their filemanager. For your application, `~/.var/app/net.froemling.Bombsquad` is the application data folder and accessible without permissions. And with the `persist` option, the path would be also mounted in the sandbox as it would be at `~/.ballisticakit`.

> Thanks, will relay this upstream
> although i notice its not set on my system, arch linux btw XDD

Well, an unset variable meaning `~/.local/share` is also part of the specification...
But applications should use ideally use the variable when set or the default from the specification if not, not dot-folders in home.

Loup-Garou911XD: alright, added in my todo list

net.froemling.bombsquad.yml
CodedOre: So, the application follows the Windows filesystem model of putting everything into one folder...

I just ran a test-build on my system and found that all data files for the game are placed under `/app/bin/bombsquad/ba_data`, which is a huge no.
The data files should be under `/app/share/bombsquad`, not in `bin`. 

Ideally `ballisticakit` would understand this on its own. Otherwise you may actually need the script, so that the app works when run from `flatpak run` directly.

net.froemling.bombsquad.yml
hfiguiere: > it would be a fair bit more complicated

not sure that's true. It's more like the `makefile` is incomplete. It should with a single `make install` build from zero to installed binaries. Anything short of that is a bug.

Like why does the build steps need to `touch` or `cp -r`

Loup-Garou911XD: > > it would be a fair bit more complicated
> 
> not sure that's true. It's more like the `makefile` is incomplete. It should with a single `make install` build from zero to installed binaries. Anything short of that is a bug.
> 
> Like why does the build steps need to `touch` or `cp -r`

it doesnt have `make install`, but it does download all the required stuff, it was a shortcoming on my part

net.froemling.bombsquad.yml
Loup-Garou911XD: > Ideally `ballisticakit` would understand this on its own. Otherwise you may actually need the script, so that the app works when run from `flatpak run` directly.

well like i said, it expects to be run from cwd, because it needs the ba_data folder

Loup-Garou911XD: so yeah the sh script needs to stay if we need to be able to run it from cli

Loup-Garou911XD:
it seems like this requires a fair bit of cleanup
ill re-request a review when its done

net.froemling.bombsquad.yml
teamcons: That makes no sense

net.froemling.bombsquad.yml
teamcons: Everything the app needs should be in the sources

Loup-Garou911XD:
bot,build

Loup-Garou911XD:
bot, build

net.froemling.bombsquad.yml
```diff
@@ -0,0 +1,59 @@
...
+
+modules:
+
+  - name: bombsquad
+    buildsystem: simple
+    build-options:
+      append-path: /usr/lib/sdk/llvm20/bin
+      prepend-ld-library-path: /usr/lib/sdk/llvm20/lib
+      build-args:
+        - --share=network
```
CodedOre, line 25: You must not have internet access during the build:

https://docs.flathub.org/docs/for-app-authors/requirements#no-network-access-during-build

CodedOre (changes requested):

net.froemling.bombsquad.appdata.xml
CodedOre, line 1: If this file is now provided by the to the build by the upstream repository, then it should no longer be in this submission. Remove it.

net.froemling.bombsquad.desktop
CodedOre, line 1: As above, remove this file from the submission and use the one from the upstream repository.

net.froemling.bombsquad.metainfo.xml
CodedOre, line 1: As above, remove this file from the submission and use the one from the upstream repository.

net.froemling.bombsquad.releases.xml
CodedOre, line 1: As above, remove this file from the submission and use the one from the upstream repository.

net.froemling.bombsquad.yml
Loup-Garou911XD: > That makes no sense

the ballisticakit binary expects a ba_data folder which has all the data to be present in the cwd, it should ideally have a flag to tell the path of the ba_data but it doesnt and thats a limitation of the game currently

net.froemling.bombsquad.yml
Loup-Garou911XD: the flatpak pip generator seems to not handle a dependency, and it gives the format for how to add it manually but i cant do that everytime a dependency is bumped

```
Output saved to python3-requirements.json
Unresolved dependencies. Handle them manually
- ERROR: Failed to get zuban-0.6.2 source from https://pypi.org/pypi/zuban/0.6.2/json
  File "/home/loup/github_fork/ballistica/temp/flatpak-pip-generator.py", line 915, in <module>
    raise Exception(
        f"Not all dependencies can be determined. Handle them manually.\n{workaround}"
    )
Exception: Not all dependencies can be determined. Handle them manually.
Example on how to handle arch dependent wheels:
    - type: file
      url: https://files.pythonhosted.org/packages/79/ae/7e5b85136806f9dadf4878bf73cf223fe5c2636818ba3ab1c585d0403164/numpy-1.26.4-cp311-cp311-manylinux_2_17_aarch64.manylinux2014_aarch64.whl
      sha256: 7ab55401287bfec946ced39700c053796e7cc0e3acbef09993a9ad2adba6ca6e
      only-arches:
      - aarch64
    - type: file
      url: https://files.pythonhosted.org/packages/3a/d0/edc009c27b406c4f9cbc79274d6e46d634d139075492ad055e3d68445925/numpy-1.26.4-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
      sha256: 666dbfb6ec68962c033a450943ded891bed2d54e6755e35e5835d63f4f6931d5
      only-arches:
      - x86_64

```

net.froemling.bombsquad.yml
bbhtt: don't mark this as resolved, it's not

Loup-Garou911XD: well the build system is complex because this is not just a cpp game engine, the Makefile at project root is used for alot more that just building the cpp
<details>
<summary>running `make` in project root output</summary>
----------------------
Available Make Targets
----------------------

General
-------
help - List targets in this Makefile and basic descriptions for them.
env - Target that should be built before building almost any other target. This
env-pre-update - Set of prereqs safe to run if the project state is dirty.
env-clean
assets - Build all assets for all platforms.
assets-cmake - Build assets required for cmake builds (linux, mac).
assets-cmake-scripts - Build only script assets for cmake builds (linux, mac).
assets-server - Build assets required for server builds.
assets-windows - Build assets required for WINDOWS_PLATFORM windows builds.
assets-windows-Win32 - Build assets required for Win32 windows builds.
assets-windows-x64 - Build assets required for x64 windows builds.
assets-mac - Build assets required for mac xcode builds
assets-ios - Build assets required for ios.
assets-android - Build assets required for android.
assets-clean - Clean all assets.
resources - Build resources.
resources-clean - Clean resources.
meta - Build our generated sources.
meta-clean - Clean our generated sources.
clean - Remove ALL files and directories that aren't managed by git (except for a
clean-list - Show what clean would delete without actually deleting it.
dummymodules - Build/update dummy python modules.
dummymodules-clean
venv - Build the project's Python virtual environment. This should happen
venv-upgrade - Update pip requirements to latest versions.
venv-clean
docs - Generate all docs.
docs-cloud - Cloud version of docs
docs-clean
pcommandbatch_speed_test

Prefab
------
prefab-gui-debug - Assemble & run a gui debug build for this platform.
prefab-gui-release - Assemble & run a gui release build for this platform.
prefab-gui-debug-build - Assemble a debug build for this platform.
prefab-gui-release-build - Assemble a release build for this platform.
prefab-server-debug - Assemble & run a server debug build for this platform.
prefab-server-release - Assemble & run a server release build for this platform.
prefab-server-debug-build - Assemble a server debug build for this platform.
prefab-server-release-build - Assemble a server release build for this platform.
prefab-clean - Clean all prefab builds.
prefab-mac-x86-64-gui-debug
prefab-mac-arm64-gui-debug
prefab-mac-x86-64-gui-debug-build
prefab-mac-arm64-gui-debug-build
prefab-mac-x86-64-gui-release
prefab-mac-arm64-gui-release
prefab-mac-x86-64-gui-release-build
prefab-mac-arm64-gui-release-build
prefab-mac-x86-64-server-debug
prefab-mac-arm64-server-debug
prefab-mac-x86-64-server-debug-build
prefab-mac-arm64-server-debug-build
prefab-mac-x86-64-server-release
prefab-mac-arm64-server-release
prefab-mac-x86-64-server-release-build
prefab-mac-arm64-server-release-build
prefab-linux-x86-64-gui-debug
prefab-linux-arm64-gui-debug
prefab-linux-x86-64-gui-debug-build
prefab-linux-arm64-gui-debug-build
prefab-linux-x86-64-gui-release
prefab-linux-arm64-gui-release
prefab-linux-x86-64-gui-release-build
prefab-linux-arm64-gui-release-build
prefab-linux-x86-64-server-debug
prefab-linux-arm64-server-debug
prefab-linux-x86-64-server-debug-build
prefab-linux-arm64-server-debug-build
prefab-linux-x86-64-server-release
prefab-linux-arm64-server-release
prefab-linux-x86-64-server-release-build
prefab-linux-arm64-server-release-build
prefab-windows-x86-64-gui-debug
prefab-windows-x86-64-gui-debug-build
prefab-windows-x86-64-gui-release
prefab-windows-x86-64-gui-release-build
prefab-windows-x86-64-server-debug
prefab-windows-x86-64-server-debug-build
prefab-windows-x86-64-server-release
prefab-windows-x86-64-server-release-build

Spinoff
-------
spinoff-test - Run a given spinoff test.
spinoff-test-core - Build and check core feature set alone.
spinoff-test-base - Build and check base feature set alone.
spinoff-test-plus - Build and check plus feature set alone.
spinoff-test-classic - Build and check classic feature set alone.
spinoff-test-template_fs - Build and check template_fs feature set alone.
spinoff-test-ui_v1 - Build and check ui_v1 feature set alone.
spinoff-test-ui_v1_lib - Build and check ui_v1_lib feature set alone.
spinoff-test-scene_v1 - Build and check scene_v1 feature set alone.
spinoff-test-scene_v1_lib - Build and check scene_v1_lib feature set alone.
spinoff-test-clean - Blow away all spinoff-test builds.
spinoff-update - Grab the current parent project and sync it into ourself.
spinoff-upgrade - Upgrade to latest parent project and sync it into ourself.

Updating
--------
update - Update any project files that need it (does NOT build projects).
update-check - Don't update but fail if anything needs it.

Formatting
----------
format - Run formatting on all files in the project considered 'dirty'.
format-full - Same but always formats; ignores dirty state.
format-code - Run formatting for compiled code sources (.cc, .h, etc.).
format-code-full - Same but always formats; ignores dirty state.
format-scripts - Runs formatting for scripts (.py, etc).
format-scripts-full - Same but always formats; ignores dirty state.
format-makefile - Runs formatting on the project Makefile.

Checking
--------
check - Run all project checks. (static analysis)
check-full - Same as check but no caching (all files are checked).
check-ex - Same as 'check' plus optional/slow extra checks.
check-ex-full - Same as check-ex but no caching (all files are checked).
cpplint - Run Cpplint checks on all C/C++ code.
cpplint-full - Run Cpplint checks without caching (all files are checked).
pylint - Run Pylint checks on all Python Code.
pylint-full - Run Pylint checks without caching (all files are checked).
pylint-ex - Run Pylint checks on all Python Code (including extra slow ones).
pylint-ex-full - Run Pylint checks including extras without caching (all files are checked).
mypy - Run Mypy checks on all Python code.
zmypy - Run Mypy checks on all Python code.
mypy-full - Run Mypy checks without caching (all files are checked).
dmypy - Run Mypy checks on all Python code using daemon mode.
dmypy-stop - Stop the mypy daemon
pyright - Run Pyright checks on all Python code.
pycharm - Run PyCharm checks on all Python code.
pycharm-full - Run PyCharm checks without caching (all files are checked).
py_check_prepass - Build prerequisites needed for python checks.

Testing
-------
test - Run tests (live execution verification).
test-ex - Run tests (live execution verification). Includes extra slow ones.
test-ex-verbose
test-full - Run tests with any caching disabled.
test-ex-full - Run extended tests with any caching disabled.
test-message - Shortcut to test efro.message only.
test-dataclassio - Shortcut to test efro.dataclassio only.
test-rpc - Shortcut to test efro.rpc only.
test-threadpool - Shortcut to test efro.threadpool only.
test-restapi

Preflighting
------------
preflight - Format, update, check, & test the project. Do this before commits.
preflight-full - Same as 'preflight' without caching (all files are visited).
preflight-ex - Same as 'preflight' plus optional/slow extra checks.
preflight-ex-full - Same as 'preflight-ex' but without caching (all files visited).

Windows
-------
windows-staging - Stage assets and other files so a built binary will run.
windows-debug - Build and run a debug windows build (from WSL).
windows-release - Build and run a release windows build (from WSL).
windows-debug-build - Build a debug windows build (from WSL).
windows-debug-rebuild - Rebuild a debug windows build (from WSL).
windows-release-build - Build a release windows build (from WSL).
windows-release-rebuild - Rebuild a release windows build (from WSL).
windows-clean - Remove all non-git-managed files in windows subdir.
windows-clean-list - Show what would be cleaned.

CMake
-----
cmake - Build and run the cmake build.
cmake-gdb - Build and run the cmake build under the gdb debugger.
cmake-lldb - Build and run the cmake build under the lldb debugger.
cmake-build - Build but don't run it.
cmake-binary
cmake-clean
cmake-server
cmake-server-build
cmake-server-binary
cmake-server-clean
cmake-modular-build
cmake-modular
cmake-modular-binary
cmake-modular-clean
cmake-modular-server
cmake-modular-server-build
cmake-modular-server-binary
cmake-modular-server-clean
clion-staging - Stage assets for building/running within CLion.

Docker
------
docker-gui-release - Build the gui release docker image
docker-gui-debug - Build the gui debug docker image
docker-server-release - Build the server release docker image
docker-server-debug - Build the server debug docker image
docker-arm64-gui-release - Build the gui release docker image for arm64
docker-arm64-gui-debug - Build the gui debug docker image for arm64
docker-arm64-server-release - Build the server release docker image for arm64
docker-arm64-server-debug - Build the server debug docker image for arm64
docker-save - Save the bombsquad_server docker image to build/docker/bombsquad_server_docker.tar
docker-clean - Cleanup docker files

Flatpak
-------
flatpak-linux
flatpak-generate-flathub-manifest
flatpak-clean
</details>

As you can see there are all sorts of make and make being the single entrypoint for building or testing anything in the project
now when we run `make cmake` it downloads all the required assets and generates a new Makefile in `build/cmake/debug/Makefile` and runs it to build the cpp game engine and assemble the other bits and pieces like python files and assets in the same dir
So in our case the stuff is essentially wrapped by our custom Makefile and is not the cmakelist.txt Makefile
I hope this explanation is clear, if not then ask your doubts

CodedOre:
Why are you adding issues you already had resolved back into this PR, instead of resolving the remaining issues?

net.froemling.bombsquad.metainfo.xml
bbhtt: @Loup-Garou911XD don't mark these comments as resolved before the necessary change being pushed in the PR.

Loup-Garou911XD:
> Why are you adding issues you already had resolved back into this PR, instead of resolving the remaining issues?

It was auto commited by release in the upstream because I forgot to remove the workflow
I'll fix it

PR #7832, opened 2026-02-14
Add me.ahmetcetinkaya.whph

ahmet-cetinkaya:

- [x] Please describe the application briefly. **Work Hard Play Hard (WHPH)** is
      a productivity application designed for task management, habit tracking,
      and time optimization. It helps users balance their work and leisure time
      effectively.
- [x] Please attach a video showcasing the application on Linux using the
      Flatpak.
      https://github.com/user-attachments/assets/22365414-3407-43c6-b69f-ea21bc349058
- [x] The Flatpak ID follows all the rules listed in the [Application ID
      requirements][appid].
- [x] I have read and followed all the [Submission requirements][reqs] and the
      [Submission guide][reqs2] and I agree to them.

[appid]:
  https://docs.flathub.org/docs/for-app-authors/requirements#application-id

### DIFF
--- a/me.ahmetcetinkaya.whph.yaml
+++ b/me.ahmetcetinkaya.whph.yaml
@@ -15,9 +15,6 @@ finish-args:
   - --socket=pulseaudio
   - --device=dri
   - --talk-name=org.kde.StatusNotifierWatcher
-  - --talk-name=org.gnome.Shell
-  - --talk-name=org.kde.KWin
-  - --talk-name=org.kde.kwin.Scripting
 modules:
   - name: perl-xml-parser
     buildsystem: simple


me.ahmetcetinkaya.whph.metainfo.xml
bbhtt, line 1: this should be upstream

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,163 @@
+app-id: me.ahmetcetinkaya.whph
+runtime: org.freedesktop.Platform
+runtime-version: "25.08"
+sdk: org.freedesktop.Sdk
+command: whph
+separate-locales: false
+build-options:
+  env:
+    CPATH: /app/include:/app/include/libayatana-ido3-0.4:/app/include/libayatana-indicator3-0.4
```
bbhtt: you should not need this.

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,163 @@
...
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --socket=pulseaudio
+  - --device=dri
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=org.freedesktop.Notifications
```
bbhtt: what is this for and why does it not use the notifications portal?

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,163 @@
...
+  - --talk-name=org.gnome.Shell
+  - --talk-name=org.kde.KWin
+  - --talk-name=org.kde.kwin.Scripting
+  - --filesystem=xdg-config/kdeglobals:ro
+  - --env=GSETTINGS_BACKEND=dconf
+  - --talk-name=org.freedesktop.Settings
+  - --talk-name=org.freedesktop.Flatpak
```
bbhtt: what are these for? these are quite extensive and broad.

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,163 @@
...
+      - make
+      - make install
+    post-install:
+      - chmod -R u+w /app/lib/perl5
+    sources:
+      - type: archive
+        url: https://cpan.metacpan.org/authors/id/T/TO/TODDR/XML-Parser-2.44.tar.gz
+        sha256: 1ae9d07ee9c35326b3d9aad56eae71a6730a73a116b9fe9e8a4758b7cc033216
+
+  - name: intltool
```
bbhtt: use from shared-modules https://github.com/flathub/shared-modules/tree/master/intltool

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,163 @@
...
+        mkdir -p build/src
+        touch build/src/AyatanaIndicator3-0.4.gir
+        touch build/src/AyatanaIndicator3-0.4.typelib
+      - ninja -C build install
+    sources:
+      - type: archive
+        url: https://github.com/AyatanaIndicators/libayatana-indicator/archive/0.9.4.tar.gz
+        sha256: a18d3c682e29afd77db24366f8475b26bda22b0e16ff569a2ec71cd6eb4eac95
+
+  - name: libayatana-appindicator
```
bbhtt: use from shared-modules and the ones above too https://github.com/flathub/shared-modules/tree/master/libayatana-appindicator

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,163 @@
...
+      - install -D src/build/linux/x64/release/bundle/share/icons/hicolor/128x128/apps/whph.png /app/share/icons/hicolor/128x128/apps/me.ahmetcetinkaya.whph.png
+      - install -D src/build/linux/x64/release/bundle/share/icons/hicolor/64x64/apps/whph.png /app/share/icons/hicolor/64x64/apps/me.ahmetcetinkaya.whph.png
+      
+      # Metainfo
+      - install -D src/linux/packaging/flatpak/me.ahmetcetinkaya.whph.metainfo.xml /app/share/metainfo/me.ahmetcetinkaya.whph.metainfo.xml
+
+    sources:
+      - type: git
+        url: https://github.com/ahmet-cetinkaya/whph.git
+        branch: main
```
bbhtt: you need tag+commit.

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,163 @@
...
+      - install -D src/linux/packaging/flatpak/me.ahmetcetinkaya.whph.metainfo.xml /app/share/metainfo/me.ahmetcetinkaya.whph.metainfo.xml
+
+    sources:
+      - type: git
+        url: https://github.com/ahmet-cetinkaya/whph.git
+        branch: main 
+
+      # Flutter SDK
+      - type: archive
+        url: https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_3.24.3-stable.tar.xz
```
bbhtt: this has x86_64 libs in it. how does it work for aarch64?

me.ahmetcetinkaya.whph.metainfo.xml
```diff
@@ -0,0 +1,213 @@
...
+    <name>Ahmet Cetinkaya</name>
+  </developer>
+  <launchable type="desktop-id">me.ahmetcetinkaya.whph.desktop</launchable>
+  <url type="homepage">https://whph.ahmetcetinkaya.me</url>
+  <url type="bugtracker">https://github.com/ahmet-cetinkaya/whph/issues</url>
+  <url type="vcs-browser">https://github.com/ahmet-cetinkaya/whph</url>
+  <url type="donation">https://ahmetcetinkaya.me/donate</url>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://raw.githubusercontent.com/ahmet-cetinkaya/whph/master/fastlane/metadata/android/en-US/images/phoneScreenshots/1.png</image>
```
hfiguiere: URL should NEVER reference the tip of a branch. It's not immutable.

same below

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,163 @@
...
+  - --talk-name=org.gnome.Shell
+  - --talk-name=org.kde.KWin
+  - --talk-name=org.kde.kwin.Scripting
+  - --filesystem=xdg-config/kdeglobals:ro
+  - --env=GSETTINGS_BACKEND=dconf
+  - --talk-name=org.freedesktop.Settings
+  - --talk-name=org.freedesktop.Flatpak
+
+modules:
+  - name: perl-xml-parser
```
hfiguiere, line 19: since this is a build dependency, there should be a `"*"` cleanup rule to remove it all at the end.

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,163 @@
...
+  - name: whph
+    buildsystem: simple
+    build-commands:
+      # Build the application
+      - cd src && flutter build linux --release --verbose
+
+      # Install binary and assets
+      - install -D src/build/linux/x64/release/bundle/whph /app/bin/whph
+      - mkdir -p /app/bin/data /app/bin/lib
+      - cp -r src/build/linux/x64/release/bundle/data/flutter_assets /app/bin/data/flutter_assets
```
hfiguiere: What happen on aarch64?

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,163 @@
...
+      - cp -a src/build/linux/x64/release/bundle/lib/. /app/bin/lib/
```
hfiguiere: `$PREFIX/bin/lib` is never a valid thing according to filesystem standards.

hfiguiere:
this is not how you use shared-modules. There is a readme clearly saying how to.

ahmet-cetinkaya:
Thank you @hfiguiere, I'm still working on it.

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: The [`flutter_local_notifications`](https://pub.dev/packages/flutter_local_notifications) plugin implements the [Desktop Notifications Specification](https://specifications.freedesktop.org/notification/latest/) via DBus. Direct permission is retained as portal-based delivery is not yet supported by the plugin.

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: See responses below for detailed justifications of permissions and architecture support. Detailed documentation is also available in [`docs/packaging/FLATPAK_PACKAGING.md`](https://github.com/ahmet-cetinkaya/whph/blob/v0.22.1/docs/packaging/FLATPAK_PACKAGING.md).

me.ahmetcetinkaya.whph.metainfo.xml
ahmet-cetinkaya: Thank you for reviews, The screenshot URLs have been updated to point to the upstream repository using a specific release tag (`v0.22.1`) instead of the `main` branch.

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: I removed it.

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: Updated the source to use a specific tag and commit for immutability.

me.ahmetcetinkaya.whph.metainfo.xml
ahmet-cetinkaya: Thank you for reviews, The screenshot URLs have been updated to point to the upstream repository using a specific release tag (`v0.22.1`) instead of the `main` branch.

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: Fixed. I have updated the manifest to correctly use `shared-modules` by referencing `shared-modules/intltool/intltool-0.51.json` and `shared-modules/libayatana-appindicator/libayatana-appindicator-gtk3.json` directly as independent modules.

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: Fixed. I have updated the manifest to correctly use `shared-modules` by referencing `shared-modules/intltool/intltool-0.51.json` and `shared-modules/libayatana-appindicator/libayatana-appindicator-gtk3.json` directly as independent modules

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: The manifest now uses architecture-specific environment variables for the build output. Inside the `whph` module, `BUNDLE_PATH` is set to `src/build/linux/x64/release/bundle` for `x86_64` and `src/build/linux/arm64/release/bundle` for `aarch64`. This ensures that the correct architecture-specific binaries and libraries are bundled for each build. Additionally, the Flutter SDK module correctly handles `only-arches` for its binary dependencies.

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: The manifest now uses architecture-specific environment variables for the build output. Inside the `whph` module, `BUNDLE_PATH` is set to `src/build/linux/x64/release/bundle` for `x86_64` and `src/build/linux/arm64/release/bundle` for `aarch64`. This ensures that the correct architecture-specific binaries and libraries are bundled for each build. Additionally, the Flutter SDK module correctly handles `only-arches` for its binary dependencies.

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: Added.

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: Fixed. Restructured the installation to use `/app/lib/whph` for the application bundle and symlinked the executable to `/app/bin/whph`. This avoids the non-standard `bin/lib` path while ensuring the Flutter engine can correctly locate its AOT data and assets.

ahmet-cetinkaya:
Thank you so much @bbhtt and @hfiguiere for the thorough reviews! I've gone ahead and addressed all the feedback. As part of these updates, I have also significantly improved the upstream application and its packaging CI/CD workflows to better align with Flatpak packaging standards. 🙌🏻

me.ahmetcetinkaya.whph.metainfo.xml
bbhtt: i was taking about the entire file...

me.ahmetcetinkaya.whph.yaml
bbhtt: you can use the notifications portal directly in your app

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,101 @@
...
+  - --talk-name=org.gnome.Shell
+  - --talk-name=org.kde.KWin
+  - --talk-name=org.kde.kwin.Scripting
+  - --talk-name=org.freedesktop.Flatpak
+modules:
+  - name: prep-lib64
+    buildsystem: simple
+    build-commands:
+      - mkdir -p /app/lib
+      - ln -s lib /app/lib64
```
bbhtt:
```suggestion
```

update your flatpak-builder version instead this should not be required with newer Flatpak.

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,101 @@
...
+      - mkdir -p /app/bin
+      - ln -sf /app/lib/whph/whph /app/bin/whph
+      - install -D $BUNDLE_PATH/share/applications/whph.desktop /app/share/applications/me.ahmetcetinkaya.whph.desktop
+      - sed -i 's/Icon=whph/Icon=me.ahmetcetinkaya.whph/g' /app/share/applications/me.ahmetcetinkaya.whph.desktop
+      - sed -i 's/Version=@APP_VERSION@/Version=1.0/g' /app/share/applications/me.ahmetcetinkaya.whph.desktop
+      - sed -i 's|Exec=.*|Exec=/app/bin/whph %U|g' /app/share/applications/me.ahmetcetinkaya.whph.desktop
+      - sed -i 's|Categories=.*|Categories=Utility;|g' /app/share/applications/me.ahmetcetinkaya.whph.desktop
+      - sed -i '/StartupWMClass/d' /app/share/applications/me.ahmetcetinkaya.whph.desktop
+      - install -D $BUNDLE_PATH/share/dbus-1/services/whph.service /app/share/dbus-1/services/me.ahmetcetinkaya.whph.service
+      - sed -i 's|Exec=.*|Exec=/app/bin/whph|g' /app/share/dbus-1/services/me.ahmetcetinkaya.whph.service
```
bbhtt: use desktop-file-edit or fix your desktop file upstream

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,101 @@
...
+      - "for size in 16 22 24 32 48 64 128 256 512; do\n  install -D src/lib/core/domain/shared/assets/images/whph-512.png\
+        \ /app/share/icons/hicolor/${size}x${size}/apps/me.ahmetcetinkaya.whph.png\n\
+        \  install -D src/lib/core/domain/shared/assets/images/whph_logo_fg.png /app/share/icons/hicolor/${size}x${size}/apps/me.ahmetcetinkaya.whph.default.png\n\
+        \  install -D src/lib/core/domain/shared/assets/images/whph_logo_fg_play.png\
+        \ /app/share/icons/hicolor/${size}x${size}/apps/me.ahmetcetinkaya.whph.play.png\n\
+        \  install -D src/lib/core/domain/shared/assets/images/whph_logo_fg_pause.png\
+        \ /app/share/icons/hicolor/${size}x${size}/apps/me.ahmetcetinkaya.whph.pause.png\n\
+        done\n"
+      - gtk-update-icon-cache -f -t /app/share/icons/hicolor
```
bbhtt:
```suggestion
```

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,101 @@
...
+  - --socket=wayland
+  - --share=network
+  - --socket=pulseaudio
+  - --device=dri
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=org.freedesktop.Notifications
+  - --filesystem=xdg-config/kdeglobals:ro
+  - --filesystem=xdg-config/gtk-3.0:ro
+  - --filesystem=xdg-config/gtk-4.0:ro
+  - --env=GSETTINGS_BACKEND=dconf
```
bbhtt:
```suggestion
```

the upstream documentation you have about these is all nonsense. Themes and some other related settings are passed into the sandbox through the portal. These directories are never required.

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,101 @@
...
+  - --talk-name=org.freedesktop.Settings
```
bbhtt: where is this interface documented?

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,101 @@
...
+  - --talk-name=org.gnome.Shell
+  - --talk-name=org.kde.KWin
+  - --talk-name=org.kde.kwin.Scripting
```
bbhtt: how does active window detection work outside gnome or kde?

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,101 @@
...
+  - --talk-name=org.freedesktop.Flatpak
```
bbhtt: why is this needed for "active window detection" again? why aren't the above ones enough?

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,101 @@
...
+      - sed -i 's/Icon=whph/Icon=me.ahmetcetinkaya.whph/g' /app/share/applications/me.ahmetcetinkaya.whph.desktop
+      - sed -i 's/Version=@APP_VERSION@/Version=1.0/g' /app/share/applications/me.ahmetcetinkaya.whph.desktop
+      - sed -i 's|Exec=.*|Exec=/app/bin/whph %U|g' /app/share/applications/me.ahmetcetinkaya.whph.desktop
+      - sed -i 's|Categories=.*|Categories=Utility;|g' /app/share/applications/me.ahmetcetinkaya.whph.desktop
+      - sed -i '/StartupWMClass/d' /app/share/applications/me.ahmetcetinkaya.whph.desktop
+      - install -D $BUNDLE_PATH/share/dbus-1/services/whph.service /app/share/dbus-1/services/me.ahmetcetinkaya.whph.service
+      - sed -i 's|Exec=.*|Exec=/app/bin/whph|g' /app/share/dbus-1/services/me.ahmetcetinkaya.whph.service
+      - "for size in 16 22 24 32 48 64 128 256 512; do\n  install -D src/lib/core/domain/shared/assets/images/whph-512.png\
+        \ /app/share/icons/hicolor/${size}x${size}/apps/me.ahmetcetinkaya.whph.png\n\
+        \  install -D src/lib/core/domain/shared/assets/images/whph_logo_fg.png /app/share/icons/hicolor/${size}x${size}/apps/me.ahmetcetinkaya.whph.default.png\n\
```
hfiguiere: how can a PNG file be all these sizes at the same time?

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: Fixed. I've removed the prep-lib64 module from the manifest as suggested, as it is indeed no longer required with recent Flatpak/flatpak-builder versions.

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: @bbhtt I've updated the manifest to use desktop-file-edit instead of sed for desktop file modifications.

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,92 @@
...
+      - perl Makefile.PL PREFIX=/app
+      - make
+      - make install
+    post-install:
+      - chmod -R u+w /app/lib/perl5
+    sources:
+      - type: archive
+        url: https://cpan.metacpan.org/authors/id/T/TO/TODDR/XML-Parser-2.44.tar.gz
+        sha256: 1ae9d07ee9c35326b3d9aad56eae71a6730a73a116b9fe9e8a4758b7cc033216
+  - shared-modules/intltool/intltool-0.51.json
```
hfiguiere: this is built by `libayatana-appindicator-gtk3.json`

```suggestion
```

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,92 @@
...
+      - mkdir -p /app/lib/whph
+      - install -D $BUNDLE_PATH/whph /app/lib/whph/whph
+      - cp -r $BUNDLE_PATH/lib /app/lib/whph/
+      - mkdir -p /app/lib/whph/data
+      - cp -r $BUNDLE_PATH/data/flutter_assets /app/lib/whph/data/flutter_assets
+      - install -D $BUNDLE_PATH/data/icudtl.dat /app/lib/whph/data/icudtl.dat
+      - mkdir -p /app/bin
+      - ln -sf /app/lib/whph/whph /app/bin/whph
+      - install -D $BUNDLE_PATH/share/applications/whph.desktop /app/share/applications/me.ahmetcetinkaya.whph.desktop
+      - desktop-file-edit --set-icon=me.ahmetcetinkaya.whph --set-key=Exec --set-value='/app/bin/whph %U' --remove-key=Version /app/share/applications/me.ahmetcetinkaya.whph.desktop
```
hfiguiere: the `exec` is never a full path. You should fix this upstream. And the version. Whatever doc you read that say the version field is the application version is 100% wrong.

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: I've removed it.

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: You're absolutely correct. My original intention was to use the convert command to handle these dynamically, but I missed implementing the conversion step.

However, to keep the build process clean and avoid additional build-time dependencies, I've decided to move away from dynamic generation. I have now updated the logic to use pre-rendered static assets for each specific resolution.

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: I have updated the manifest to remove the redundant `intltool-0.51.json` inclusion, as it's already handled as a dependency within `libayatana-appindicator-gtk3.json`. I've also updated our local documentation to correctly reflect this shared-module structure.

Verified with a successful local build. 👍

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: Thank you for your feedback, @hfiguiere. I have addressed these concerns directly in the upstream build configuration to ensure the application generates a compliant desktop entry by default.

Key improvements include:
- **Executable Path:** Modified [CMakeLists.txt](https://github.com/ahmet-cetinkaya/whph/blob/main/src/linux/CMakeLists.txt) to utilize the relative binary name for the `Exec` key in production builds, avoiding absolute paths.
- **Desktop Entry Specification:** Removed the non-compliant `Version` key from [whph.desktop.in](https://github.com/ahmet-cetinkaya/whph/blob/main/src/linux/whph.desktop.in), as it incorrectly referenced the application version rather than the specification version.
- **Icon and Categories:** Standardized the `Icon` naming convention and refined the `Categories` list to align with FreeDesktop standards.

Consequently, I have removed the `desktop-file-edit` workarounds from the [Flatpak manifest](https://github.com/ahmet-cetinkaya/whph/blob/main/packaging/flatpak/flatpak-flutter.yaml), as they are no longer required. I would appreciate your further review of these updates.

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: @bbhtt; Outside of GNOME and KDE, the app utilizes native C++ fallback mechanisms to detect the active window, depending on the display server:
 
- **On Wayland:** Since there is no universal Wayland protocol for active window detection, we attempt to detect the compositor and use its specific tooling. We try:
  - **Sway:** Executing `swaymsg -t get_tree` to parse the focused window.
  - **wlroots (Hyprland, River, etc.):** Executing `hyprctl activewindow -j` or `wayinfo`.
  - **XWayland Fallback:** Executing `xprop -root _NET_ACTIVE_WINDOW` (via `flatpak-spawn --host`) since some Wayland implementations still support it for X11 clients.
  - **Process Heuristic Fallback:** If all else fails, it relies on a process-based heuristic by running `ps -eo comm` (via `flatpak-spawn --host`) to find the most recently active or highest CPU GUI process. While inaccurate, it acts as an absolute last resort.
   
- **On X11:** It natively communicates with the X server using `XGetWindowProperty(_NET_ACTIVE_WINDOW)` and related X11 atoms to reliably fetch the window class and name.

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: > @bbhtt; Outside of GNOME and KDE, the app utilizes native C++ fallback mechanisms to detect the active window, depending on the display server:
> 
>     * **On Wayland:** Since there is no universal Wayland protocol for active window detection, we attempt to detect the compositor and use its specific tooling. We try:
>       
>       * **Sway:** Executing `swaymsg -t get_tree` to parse the focused window.
>       * **wlroots (Hyprland, River, etc.):** Executing `hyprctl activewindow -j` or `wayinfo`.
>       * **XWayland Fallback:** Executing `xprop -root _NET_ACTIVE_WINDOW` (via `flatpak-spawn --host`) since some Wayland implementations still support it for X11 clients.
>       * **Process Heuristic Fallback:** If all else fails, it relies on a process-based heuristic by running `ps -eo comm` (via `flatpak-spawn --host`) to find the most recently active or highest CPU GUI process. While inaccurate, it acts as an absolute last resort.
> 
>     * **On X11:** It natively communicates with the X server using `XGetWindowProperty(_NET_ACTIVE_WINDOW)` and related X11 atoms to reliably fetch the window class and name.


In addition to above section for specific `org.freedesktop.Flatpak` permission.
The `--talk-name=org.freedesktop.Flatpak` permission is critical for allowing the app to use `flatpak-spawn --host` to execute commands on the host system. This is necessary for the "Active Window Detection" feature because:
 
1. **Executing Host Tools:** While GNOME can be queried purely via its D-Bus interface (which the other `--talk-name` permissions handle), our fallback detectors for KDE (`qdbus`, `journalctl`) and other compositors (`xprop`) rely on executing command-line utilities. These utilities exist on the host but are not available inside the strict Flatpak sandbox environment. By using `flatpak-spawn --host`, we can successfully run these tools from within the Flatpak.
2. **Accessing Host Processes:** For the absolute fallback heuristics outside of GNOME/KDE, the app runs `ps` with `flatpak-spawn --host` to inspect the host's process list. Running `ps` directly inside the sandbox would only show our isolated sandboxed processes, rendering the heuristic useless.
 
I explicitly outlined this necessity in our `FLATPAK_PACKAGING.md` under [Window Management](https://github.com/ahmet-cetinkaya/whph/blob/main/docs/packaging/FLATPAK_PACKAGING.md#window-management), detailing that this Host Access is a documented requirement for running arbitrary tools to determine the currently focused application.

hfiguiere:
really you don't need to use a chatbot to reply. It waste the reviewer time and burn the planet at the same time for only negative benefit.

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,89 @@
...
+      - install -D $BUNDLE_PATH/whph /app/lib/whph/whph
+      - cp -r $BUNDLE_PATH/lib /app/lib/whph/
+      - mkdir -p /app/lib/whph/data
+      - cp -r $BUNDLE_PATH/data/flutter_assets /app/lib/whph/data/flutter_assets
+      - install -D $BUNDLE_PATH/data/icudtl.dat /app/lib/whph/data/icudtl.dat
+      - mkdir -p /app/bin
+      - ln -sf /app/lib/whph/whph /app/bin/whph
+      - install -D $BUNDLE_PATH/share/applications/whph.desktop /app/share/applications/me.ahmetcetinkaya.whph.desktop
+      - install -D $BUNDLE_PATH/share/dbus-1/services/whph.service /app/share/dbus-1/services/me.ahmetcetinkaya.whph.service
+      - "for size in 16 22 24 32 48 64 128 256 512; do\n  install -D src/lib/core/domain/shared/assets/images/whph-${size}.png\
```
hfiguiere: this should be part of the build system.

me.ahmetcetinkaya.whph.metainfo.xml
ahmet-cetinkaya: Thank you @bbhtt for the clarification. I have now moved the entire `me.ahmetcetinkaya.whph.metainfo.xml` file to the upstream repository (under src/linux/share/metainfo/) and configured the build system (CMakeLists.txt) to install it into the bundle.

The Flatpak manifest has also been updated to copy the metadata from the application's build bundle instead of it being hosted directly in the Flathub repository.

Could you please confirm if this is the correct approach you were suggesting?

ahmet-cetinkaya:
@hfiguiere My native language is not English, so after writing the answers, I use an small AI model running on my local machine to improve my comments and correct any errors in meaning or grammar.

hfiguiere:
> My native language is not English, 

Neither is mine. 

But I guess maybe I should just stop reviewing. I'm not here to spend my free time to read AI slop.

ahmet-cetinkaya:
@hfiguiere My aim was to make your job easier by providing answers that are as clear and understandable as possible. If you find it difficult to read, I respect that, and I will take this into consideration on my part. 👍🏻

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: Thank you, I have moved the section to [CMAKE](https://github.com/ahmet-cetinkaya/whph/blob/main/src/linux/CMakeLists.txt) file.

hfiguiere:
> providing answers that are as clear and understandable as possible

Then don't burn the planet using a chatbot. AI slop is never clear and understandable.

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: I will try to add special implementation to flatpak builds for using flatpak notification portal.

ahmet-cetinkaya: @bbhtt, thanks! I've removed the static notification permission and implemented the portal service as suggested. Ready for another look.

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: I had added them because I was experiencing issues for auto theme, but after removing them and trying again, the auto theme now seems to be functioning correctly within the system. Therefore, I removed it. 👍🏻

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: https://github.com/ahmet-cetinkaya/whph/blob/main/docs/packaging/FLATPAK_PACKAGING.md#integration--theming

bbhtt: Did you understand the question? I do not want your explanation of it but the place where the interface is documented. 

Eg. `org.gnome.Shell` is [documented here](https://github.com/GNOME/gnome-shell/blob/main/data/dbus-interfaces/org.gnome.Shell.xml)

For one your explanation is again wrong as for one XDG settings portals do not require any extra permission and secondly, that's not even the correct name for it.

me.ahmetcetinkaya.whph.yaml
bbhtt: This does not seem suitable to be packaged inside Flatpak then.

ahmet-cetinkaya: @bbhtt; Hmm, do you have any suggestions for capturing an active window in Wayland?

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: Sorry, it didn't click right away.

I've removed it. After removing the last piece, I didn't notice any theme incompatibility in the application. This is good news, as it means a more minimal configuration.

This is a bit interesting to me because when preparing the Flatpak package, I was experiencing an issue where the application theme didn't match the system theme, and I clearly remember resolving it with permissions. But I suppose that wasn't the case after all; the issue was elsewhere, perhaps a system-specific problem.

Thank you.

flathub.json
hfiguiere, line 1: remove this file.

automerge will not be allowed.

me.ahmetcetinkaya.whph.yaml
ahmet-cetinkaya: @bbhtt, I removed the permission and added an information card to notify the user if the app cannot capture the active window.

me.ahmetcetinkaya.whph.yaml
```diff
@@ -0,0 +1,75 @@
+id: me.ahmetcetinkaya.whph
+runtime: org.freedesktop.Platform
+runtime-version: "25.08"
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.llvm20
+build-options:
+  prepend-pkg-config-path: /app/lib64/pkgconfig
```
hfiguiere:
```suggestion
```

there is no lib64. Use the latest flatpak-builder.

me.ahmetcetinkaya.whph.yaml
bbhtt: https://github.com/flathub/flathub/pull/7832#discussion_r2836899773 still stands it's either broken on other WMs (because you removed flatpak-spawn) or overly broad and depends on host components

as i said this doesn't seem suitable for Flatpak.

ahmet-cetinkaya: Ok, I removed the permissions @bbhtt.

PR #7869, opened 2026-02-19
Add io.github.serverket.cpugov

Serverket:

- [x] Please describe the application briefly. CPU Governor is a modern GTK4 utility for managing Linux CPU frequency scaling.

- [x] Please attach a video showcasing the application on Linux using the Flatpak. https://github.com/user-attachments/assets/60b59686-0818-4642-81ed-0b6f33d40371
- [x] The Flatpak ID follows all the rules listed in the [Application ID requirements][appid]. io.github.serverket.cpugov

### DIFF
--- a/io.github.serverket.cpugov.yml
+++ b/io.github.serverket.cpugov.yml
@@ -20,5 +20,5 @@ modules:
     sources:
       - type: git
         url: https://github.com/Serverket/cpugov.git
-        tag: v0.1.1
-        commit: f7808a437b9341c3ac2b2dcdfbc11327d5fa8ce6
+        tag: v0.1.2
+        commit: f10b4dd4e7e8599ae85d6b631ced5f1e345ff22a


io.github.serverket.cpugov.yml
```diff
@@ -0,0 +1,26 @@
+id: io.github.serverket.cpugov
+runtime: org.gnome.Platform
+runtime-version: '47'
```
hfiguiere: this runtime is EOL

Serverket: Thank you for the comment, I'll fix it.

Serverket:
bot, build

Serverket:
bot, build

io.github.serverket.cpugov.yml
```diff
@@ -21,4 +21,4 @@ modules:
       - type: git
         url: https://github.com/Serverket/cpugov.git
         tag: v0.1.0
-        commit: 197a5fffc907aabf52bf0bb5f07019cf38653280
+        commit: 0b9fbe23c67c39a3ef52a019e64c4d2a53f83a53
```
hfiguiere: retagging is bad.

Serverket: I'm using this method since it's the fastest way to apply changes on the upstream by amending and then changing the hash here...

I had to fix some build issues (meson options and read-only paths) for the initial submission.

hfiguiere: Retagging is bad - which is why we require commit hash along the tag. You don't need a tag to build.

Also you should try to build locally first.

Serverket: Got it, I appreciate the comment, I'll comply.

I've released v0.1.1 which includes fixes for the build system (meson options) and the read-only filesystem paths. Updated the manifest to use this new tag.

Serverket:
bot, build

Serverket:
bot, build

io.github.serverket.cpugov.yml
```diff
@@ -0,0 +1,24 @@
...
+sdk: org.gnome.Sdk
+command: cpugov-gtk
+
+finish-args:
+  # Display
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  # Talk to our D-Bus system daemon
+  - --system-talk-name=io.github.serverket.cpugov
```
bbhtt, line 13: how does this work? it needs to be installed separately on host or something?

Serverket: Yes, exactly! cpugov-daemon must be installed separately on the host system — it's the component that actually writes to /sys/.../scaling_governor as root. The Flatpak only ships the GTK4 GUI.

The sandboxed GUI to call methods on the daemon's D-Bus interface over the system bus. The daemon is distributed as a .deb (and installable from source via meson).

Serverket: I just wrote an install script to simplify this, one-command solution for version v.0.2.0

bbhtt: How does the workflow of the installing/configuring the host component look like? The Flatpak is supposed to be self-sufficient. Having host dependencies can be a blocker.

Serverket: I get it and I tried to architect it that way initially. However, modifying CPU governors requires writing directly to /sys/devices/system/cpu/... as root. A sandboxed Flatpak simply cannot (and should not) do this securely without breaking the sandbox entirely.

Because running an entire GTK4/Wayland GUI as root is highly discouraged and insecure, I use a privileged D-Bus system service.

1. The user installs the host component via .deb package or the automated install.sh script, which simply places a small Python script and a systemd service on the host.
2. The user installs the Flatpak GUI from Flathub.
3. When the Flatpak GUI launches, it communicates with the host over the system D-Bus.

There are three apps on Flathub that use this approach; Polychromatic and Solaar,.

bbhtt: Not going to be accepted. You just showed the app is not suitable for Flatpak. I suggest looking into alternative packaging methods.

Serverket: I respectfully disagree, a sandboxed GUI frontend communicating with a privileged host D-Bus daemon is an established arch pattern for hardware management tools on Flathub. For example, both Polychromatic (openrazer-daemon) and Solaar (udev rules) use this exact separated host-backend design to preserve sandbox compliance while managing hardware.

That said, is there a specific Flathub policy that prevents this currently? Do you suggest something in particular?

PR #7471, opened 2026-01-03
Add io.github.alper_han.crossmacro

alper-han:
CrossMacro - Mouse and keyboard macro automation tool for X11

Built with .NET 10 and Avalonia UI

### Please confirm your submission meets all the criteria

- [X] Please describe the application briefly. 
  CrossMacro is a powerful mouse and keyboard macro recording and playback tool for Linux. Features include: macro recording/replay, text expansion, global hotkeys, adjustable playback speed, and loop mode. Built with .NET 10 and Avalonia UI, supports X11 sessions.

- [X] Please attach a video showcasing the application on Linux using the Flatpak.

https://github.com/user-attachments/assets/23a0a835-4d49-4e33-958a-c9165d9ba8ee

- [X] The Flatpak ID follows all the rules listed in the [Application ID requirements][appid].

- [X] I have read and followed all the [Submission requirements][reqs] and the [Submission guide][reqs2] and I agree to them.

- [X] I am an author/developer to the project.

[appid]: https://docs.flathub.org/docs/for-app-authors/requirements#application-id
[reqs]: https://docs.flathub.org/docs/for-app-authors/requirements
[reqs2]: https://docs.flathub.org/docs/for-app-authors/submission

PR #7889, opened 2026-02-21
Add io.github.alper_han.crossmacro

alper-han:

    CrossMacro is a mouse and keyboard macro automation application for recording and replaying actions.
    It is built with .NET 10 and Avalonia UI.

    Main features include:
    - macro recording and playback
    - text expansion
    - global hotkeys
    - scheduling
    - loop mode and speed control
    - macro editor

    Linux support includes both X11 and Wayland:
    - On X11, CrossMacro uses native X11 APIs for input capture/simulation.
    - On Wayland, CrossMacro uses daemon mode when the daemon socket is available.
    - If daemon mode is not available, it falls back to direct mode. Direct mode requires low-level input device access (`/dev/
  uinput` and related devices). In practice, this usually means `input` group membership and proper `uinput` permissions/rules
  on the host.

    For Wayland mouse position:
    - Hyprland via IPC
    - KDE via D-Bus
    - GNOME via Shell extension
    - Relative fallback for other compositors

    This submission supersedes #7471 (which previously described X11-only support).

    - X11 (KDE)
    
    https://github.com/user-attachments/assets/23a0a835-4d49-4e33-958a-c9165d9ba8ee

    - Wayland (Hyprland)

    https://github.com/user-attachments/assets/f97880ac-0cb8-4c87-9642-1d5317ea2c44

### DIFF
--- a/io.github.alper_han.crossmacro.yml
+++ b/io.github.alper_han.crossmacro.yml
@@ -1,0 +1,98 @@
+app-id: io.github.alper_han.crossmacro
+runtime: org.freedesktop.Platform
+runtime-version: '25.08'
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.dotnet10
+command: crossmacro
+
+finish-args:
+  # Display
+  - --socket=x11
+  - --share=ipc
+
+  # Input devices (required for Wayland macro recording/playback)
+  - --device=all
+
+  # D-Bus: KDE Plasma
+  - --talk-name=org.kde.keyboard
+  - --talk-name=org.kde.KWin
+
+  # D-Bus: GNOME
+  - --talk-name=org.gnome.Shell
+
+  # Host setup helper (flatpak-spawn --host)
+  - --talk-name=org.freedesktop.Flatpak
+
+
+  # Filesystem: Compositor IPC
+  - --filesystem=xdg-run/hypr:ro
+
+  # Filesystem: Daemon socket (hybrid mode)
+  - --filesystem=/run/crossmacro:rw
+
+  # Filesystem: GNOME Shell extension (auto-install for Wayland support)
+  - --filesystem=~/.local/share/gnome-shell/extensions:create
+
+  # Environment
+  - --env=DOTNET_ROOT=/app/lib/dotnet
+  - --env=CROSSMACRO_FLATPAK=1
+
+build-options:
+  prepend-path: /usr/lib/sdk/dotnet10/bin
+  append-ld-library-path: /usr/lib/sdk/dotnet10/lib
+  prepend-pkg-config-path: /usr/lib/sdk/dotnet10/lib/pkgconfig
+  strip: true
+  env:
+    DOTNET_CLI_TELEMETRY_OPTOUT: 'true'
+    DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 'true'
+
+cleanup:
+  - '*.pdb'
+
+modules:
+  # .NET Runtime Installation (Framework-Dependent approach)
+  - name: dotnet
+    buildsystem: simple
+    build-commands:
+      - /usr/lib/sdk/dotnet10/bin/install.sh
+
+  - name: crossmacro
+    buildsystem: simple
+    build-commands:
+      # Publish as framework-dependent (supports both x86_64 and aarch64)
+      - dotnet publish src/CrossMacro.UI.Linux/CrossMacro.UI.Linux.csproj -c Release --no-self-contained -p:PublishReadyToRun=false -p:PublishTrimmed=false --source nuget-sources -o /app/lib/crossmacro
+
+      # Install launcher script
+      - install -Dm755 crossmacro.sh /app/bin/crossmacro
+
+      # Desktop entry
+      - install -Dm644 flatpak/io.github.alper_han.crossmacro.desktop /app/share/applications/io.github.alper_han.crossmacro.desktop
+
+      # Icons
+      - install -Dm644 src/CrossMacro.UI/Assets/icons/512x512/apps/crossmacro.png /app/share/icons/hicolor/512x512/apps/io.github.alper_han.crossmacro.png
+      - install -Dm644 src/CrossMacro.UI/Assets/icons/256x256/apps/crossmacro.png /app/share/icons/hicolor/256x256/apps/io.github.alper_han.crossmacro.png
+      - install -Dm644 src/CrossMacro.UI/Assets/icons/128x128/apps/crossmacro.png /app/share/icons/hicolor/128x128/apps/io.github.alper_han.crossmacro.png
+
+      # Metainfo
+      - install -Dm644 flatpak/io.github.alper_han.crossmacro.metainfo.xml /app/share/metainfo/io.github.alper_han.crossmacro.metainfo.xml
+
+      # License
+      - install -Dm644 LICENSE /app/share/licenses/io.github.alper_han.crossmacro/LICENSE
+
+    sources:
+      # Main source from GitHub
+      - type: git
+        url: https://github.com/alper-han/CrossMacro.git
+        tag: v0.9.8
+        commit: a1a6bff580f07583ca8ed50438ef7fdf7792bf61
+        x-checker-data:
+          type: git
+          tag-pattern: ^v([\d.]+)$
+
+      # Launcher script
+      - type: file
+        path: crossmacro.sh
+
+      # NuGet packages (offline)
+      - nuget-sources.json
--- a/crossmacro.sh
+++ b/crossmacro.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+# CrossMacro Flatpak Launcher (Hybrid Mode)
+
+export DOTNET_ROOT=/app/lib/dotnet
+
+# Check if CrossMacro daemon is available on host
+DAEMON_SOCKET="/run/crossmacro/crossmacro.sock"
+
+if [ -S "$DAEMON_SOCKET" ]; then
+    export CROSSMACRO_USE_DAEMON=1
+    echo "[CrossMacro] Using daemon mode (secure)" >&2
+else
+    export CROSSMACRO_USE_DAEMON=0
+    echo "[CrossMacro] Using direct mode (Flatpak permissions)" >&2
+fi
+
+exec /app/lib/dotnet/dotnet /app/lib/crossmacro/CrossMacro.UI.dll "$@"
--- a/nuget-sources.json
+++ b/nuget-sources.json
@@ -1,0 +1,275 @@
+[
+    {
+        "type": "file",
+        "url": "https://api.nuget.org/v3-flatcontainer/avalonia.buildservices/11.3.2/avalonia.buildservices.11.3.2.nupkg",
+        "sha512": "233f868e74ab91e56818004ad0aa7d0bb6e978044ddcf0061e373ada06a1846b17a82e94c777ab907ba484c1873d2212b47d3539d2694e89a415f1af047c3045",
+        "dest": "nuget-sources",
+        "dest-filename": "avalonia.buildservices.11.3.2.nupkg"
+    }
+,    {
+        "type": "file",
+        "url": "https://api.nuget.org/v3-flatcontainer/avalonia.controls.colorpicker/11.3.12/avalonia.controls.colorpicker.11.3.12.nupkg",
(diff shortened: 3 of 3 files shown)


alper-han:
I tried to keep permissions minimal and request only what CrossMacro currently needs.
I’m happy to justify each permission in the related review threads. I hope testing goes smoothly.
If you test direct fallback mode, please remember to log out and log back in after granting input/uinput access.
Thank you and have a great day.

hfiguiere:
> I reviewed several macro tools on Flathub and noticed that some of them request broad permissions.

We are not reviewing these apps, we are reviewing this one.

io.github.alper_han.crossmacro.yml
```diff
@@ -0,0 +1,104 @@
...
+  # Input devices (required for Wayland macro recording/playback)
+  - --device=all
+
+  # GPU acceleration
+  - --device=dri
+
+  # D-Bus: KDE Plasma
+  - --talk-name=org.kde.keyboard
+  - --talk-name=org.kde.KWin
+  - --own-name=org.crossmacro.Tracker
```
hfiguiere: no. Do you own `crossmacro.org` ?

io.github.alper_han.crossmacro.yml
```diff
@@ -0,0 +1,104 @@
...
+  prepend-path: /usr/lib/sdk/dotnet10/bin
+  append-ld-library-path: /usr/lib/sdk/dotnet10/lib
+  prepend-pkg-config-path: /usr/lib/sdk/dotnet10/lib/pkgconfig
+  strip: true
+  env:
+    DOTNET_CLI_TELEMETRY_OPTOUT: 'true'
+    DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 'true'
+
+cleanup:
+  - /lib/debug
```
hfiguiere:
```suggestion
```

NEVER

io.github.alper_han.crossmacro.yml
```diff
@@ -0,0 +1,104 @@
...
+
+  # Input devices (required for Wayland macro recording/playback)
+  - --device=all
+
+  # GPU acceleration
+  - --device=dri
+
+  # D-Bus: KDE Plasma
+  - --talk-name=org.kde.keyboard
+  - --talk-name=org.kde.KWin
```
bbhtt, line 19: what are these two for?

io.github.alper_han.crossmacro.yml
```diff
@@ -0,0 +1,104 @@
...
+  - --own-name=org.crossmacro.Tracker
+
+  # D-Bus: GNOME
+  - --talk-name=org.gnome.Shell
+  - --talk-name=org.gnome.Shell.Extensions
+
```
bbhtt, line 23: what are these for?

io.github.alper_han.crossmacro.yml
```diff
@@ -0,0 +1,104 @@
...
+  # D-Bus: Freedesktop
+  - --talk-name=org.freedesktop.Notifications
```
bbhtt: why does it not use the portal?

io.github.alper_han.crossmacro.yml
```diff
@@ -0,0 +1,104 @@
...
+
+
+  # Filesystem: Compositor IPC
+  - --filesystem=xdg-run/hypr:ro
```
bbhtt, line 29: how does this work? You are going to keep adding paths for every WM/DE out there?

io.github.alper_han.crossmacro.yml
```diff
@@ -0,0 +1,104 @@
...
+
+  # Filesystem: Daemon socket (hybrid mode)
+  - --filesystem=/run/crossmacro:rw
```
bbhtt, line 32: so it requires some host component to be installed separately?

io.github.alper_han.crossmacro.yml
```diff
@@ -0,0 +1,104 @@
...
+
+  # Filesystem: GNOME Shell extension (auto-install for Wayland support)
+  - --filesystem=~/.local/share/gnome-shell/extensions:create
```
bbhtt, line 35: how does this work outside of gnome?

io.github.alper_han.crossmacro.yml
alper-han: I've been thinking about changing this since very old versions.
Thanks for letting me know, I've fixed it now. I don't need it anymore.
I was using it for dbus.

io.github.alper_han.crossmacro.yml
alper-han: Removed /lib/debug from cleanup.
Thank you

io.github.alper_han.crossmacro.yml
alper-han: org.kde.keyboard is used to read the active keyboard layout on KDE for correct key mapping.

org.kde.KWin is used to load and run a KWin script that provides absolute cursor position and screen resolution on KDE-Wayland.

io.github.alper_han.crossmacro.yml
alper-han: org.gnome.Shell: To obtain screen resolution and mouse absolute position by loading a script on GNOME-Wayland

org.gnome.Shell.Extensions was redundant here (it is an interface on org.gnome.Shell, not a separate bus name), so I removed it.

io.github.alper_han.crossmacro.yml
alper-han: I removed it because I no longer needed it.

io.github.alper_han.crossmacro.yml
alper-han: No. We are not planning to add filesystem paths for every WM/DE. 
Absolute pointer support is currently implemented only for Hyprland , KDE  and GNOME
Other compositors use Relative mode.

io.github.alper_han.crossmacro.yml
alper-han: This is not mandatory; it is only for priority operation if installed in the system.

This is safer for Wayland users.
The input group permission does not need to be provided by the current user.

No permissions or tools are required for x11 users.

io.github.alper_han.crossmacro.yml
alper-han: This permission is only used on GNOME to auto-install our shell extension under ~/.local/share/gnome-shell/extensions.

Outside GNOME, that code path is not executed unused.

Gnome is a bit strict about plugin security.
Also, you need to restart the session when the plugin is installed for the first time.
A warning appears on the GUI.

hfiguiere:
bot, build

io.github.alper_han.crossmacro.yml
CodedOre: But you would create the folder on every system the application runs on, no matter if it runs GNOME or not. 

There is a DBus interface to install extensions over. Why does it not use it instead?

alper-han: The app checks whether the current session is GNOME (`XDG_CURRENT_DESKTOP` / `GDMSESSION`).
Only in GNOME sessions, it installs and enables the extension automatically; in non-GNOME sessions this code path is not executed.

https://github.com/alper-han/CrossMacro/blob/main/src/CrossMacro.Platform.Linux/DisplayServer/Wayland/GnomePositionProvider.cs#L112-L120
https://github.com/alper-han/CrossMacro/blob/main/src/CrossMacro.Platform.Linux/DisplayServer/Wayland/GnomePositionProvider.cs#L25-L81

This avoids manual setup and gives a smoother first-run experience for GNOME users.

I need to obtain the absolute position information and screen resolution.
I can remove this permission, but I don't think users can perform a manual installation.

If there is a better way, I don't know it yet.

io.github.alper_han.crossmacro.yml
bbhtt: most distros are wayland by default, so it becomes technically necessary.

io.github.alper_han.crossmacro.yml
bbhtt: where is the extension?

bbhtt:
what setups does this support? i take it gnome, kde and hyprland works? what about others?

io.github.alper_han.crossmacro.yml
alper-han: > where is the extension?

The extension is embedded in the app source (not as separate files in this Flathub PR).

  GNOME extension sources:
  - extension.js:
  https://github.com/alper-han/CrossMacro/blob/main/src/CrossMacro.Platform.Linux/DisplayServer/Wayland/GnomePositionProvider.cs#L25
  - metadata.json:
  https://github.com/alper-han/CrossMacro/blob/main/src/CrossMacro.Platform.Linux/DisplayServer/Wayland/GnomePositionProvider.cs#L83

  Runtime install path:
  ~/.local/share/gnome-shell/extensions/crossmacro@zynix.net/

  Install/enable logic:
  - install/write:
  https://github.com/alper-han/CrossMacro/blob/main/src/CrossMacro.Platform.Linux/DisplayServer/Wayland/GnomePositionProvider.cs#L128
  - enable/check:
  https://github.com/alper-han/CrossMacro/blob/main/src/CrossMacro.Platform.Linux/DisplayServer/Wayland/GnomePositionProvider.cs#L229

io.github.alper_han.crossmacro.yml
alper-han: > most distros are wayland by default, so it becomes technically necessary.

Yes, on Wayland it is effectively needed for full functionality when users do not grant host-level input/uinput access to their user account

On Wayland, this is required to capture keyboard/mouse events and to inject playback events

In short, there are two paths:
1. Preferred: use the system daemon/service setup.
2. Alternative: grant host input/uinput permissions directly.

X11 users do not need either of these.

alper-han:
> what setups does this support? i take it gnome, kde and hyprland works? what about others?

Yes, that’s correct.

Current support is:

- X11 sessions: fully supported across all desktop environments/window managers.
- Wayland:
  - GNOME, KDE Plasma, and Hyprland: support absolute + relative positioning.
  - Other Wayland compositors: currently relative positioning only.

For clarity:
- Absolute positioning = replay moves the pointer to exact screen coordinates (x,y), matching recorded positions.
- Relative positioning = replay applies movement deltas (move by dx,dy), so exact on-screen coordinates may differ depending on monitor/layout/starting position.

To reduce that drift in relative mode, I added an optional “reset to (0,0) before playback” behavior to improve initial position accuracy users can disable it if they prefer.

For other Wayland compositors, I have not found a reliable method yet for absolute positioning.

bbhtt:
bot, build

bbhtt:
So seems like it is not usable in wayalnd without usermod -aG or installing the daemon out of band

I also noticed that we rejected it last time on this reason and nothing seems to have changed in this regard. https://github.com/flathub/flathub/pull/7471#issuecomment-3732335903 The input group was mentioned there as well https://github.com/flathub/flathub/pull/7471#issuecomment-3707818146

alper-han:
> So seems like it is not usable in wayalnd without usermod -aG or installing the daemon out of band
> 
> I also noticed that we rejected it last time on this reason and nothing seems to have changed in this regard. [#7471 (comment)](https://github.com/flathub/flathub/pull/7471#issuecomment-3732335903) The input group was mentioned there as well [#7471 (comment)](https://github.com/flathub/flathub/pull/7471#issuecomment-3707818146)

A) I can add a setup warning and provide a copy-paste command example.
---
B) By using the org.freedesktop.Flatpak permission, I can make the setup screen trigger installation via pkexec.
---
That way, the daemon would be installed automatically.

What do you think about this approach? Do you have any better suggestions?

bbhtt:
It just means Flatpak is not a suitable method of packaging these type of applications.

bbhtt:
Only flatpak-spawn seems viable to me, given that this type of app is geared towards power users, I guess it is ok to have it.

alper-han:
It currently tries to run in three different ways:

1. If the daemon is already installed on the host, it uses daemon mode.
2. If the user already has input-group/device access, it uses direct mode.
3. If neither (1) nor (2) is available, it runs a Quick Setup step at startup via flatpak-spawn --host + pkexec to apply temporary device ACLs.

resim

Exit button
resim

I just realized the colors are wrong :)

---

I could have done step 1 entirely with pkexec, but that would likely become very complicated and distro-specific.
I could also have gone with step 2 only, but I did not consider it safe enough.

So I decided this approach is the most compatible and least error-prone option overall.


There are details in this PR about how it works, including the pkexec flow
https://github.com/flathub-infra/flatpak-builder-lint/pull/925

bbhtt:
bot, build

PR #7910, opened 2026-02-23
Add com.hdrsoft.PhotomatixLinux

mcgddd:

- Application to create high dynamic range photos in the style you want. Includes basic batch processing and presets for real estate photography. 
<

https://github.com/user-attachments/assets/23c8fc7e-c8c5-40d3-9884-cb9e551ba9a8

  >
- [X] I am a developer to the project.

### Explanation of permissions
This app has filesystem permission because it is an application that merges photos from the user's filesystem.  Home permission was not sufficient, because photographers often store their photos somewhere other than their home directory, due to the amount of space they take, including on NAS servers.
I did consider using portals, but the application offers batch processing, where not only a whole directory can be selected. but also individual files selected from that directory.  The results of the batch processing are by default written to a subdirectory of the directory where the files are stored.  This doesn't work with the portal system without significant reworking of the application, and/or making it significantly less usable.  I think the expectation for an application like this would be that it will be able to access the user's photos wherever they are, and write the results of the batch processing.

### Explanation of use of libjpeg-turbo8 package
The application is built as a .deb on Ubuntu 24, and the flatpak build process uses the .deb to create the flatpak.  There are some differences between the org.gnome.Sdk/48 environment and Ubuntu 24, particularly in that Ubuntu 24 uses a different JPEG library to org.gnome.Sdk/48, so I extract just the .so file and links for that library from the package for compatibility purposes.


mcgddd:
bot, build

com.hdrsoft.PhotomatixLinux.yaml
```diff
@@ -0,0 +1,108 @@
+app-id: com.hdrsoft.PhotomatixLinux
+runtime: org.gnome.Platform
+runtime-version: '48'
```
hfiguiere, line 3: 49 is current.

does this really need the GNOME runtime?

com.hdrsoft.PhotomatixLinux.yaml
```diff
@@ -0,0 +1,108 @@
...
+sdk: org.gnome.Sdk
+command: photomatix
+finish-args:
+  - '--share=ipc'
+  - '--socket=wayland'
+  - '--socket=fallback-x11'
+  - '--share=network'
+  - '--device=dri'
+  - '--filesystem=host'
+  - '--env=LD_LIBRARY_PATH=/app/lib:/app/lib/x86_64-linux-gnu'
```
hfiguiere, line 13: this shouldn't be needed.

com.hdrsoft.PhotomatixLinux.yaml
```diff
@@ -0,0 +1,108 @@
...
+build-options:
+  env:
+    PKG_CONFIG_PATH: /app/lib/pkgconfig:/app/share/pkgconfig:/usr/lib/pkgconfig:/usr/share/pkgconfig:/app/lib64/pkgconfig
```
hfiguiere, line 16:
```suggestion
```

com.hdrsoft.PhotomatixLinux.yaml
```diff
@@ -0,0 +1,108 @@
...
+modules:
+  - gtkmm.yaml
```
hfiguiere, line 18: this should be inline here. But a binary relying on gtkmm is a disaster waiting to happen because of ABI.

com.hdrsoft.PhotomatixLinux.yaml
```diff
@@ -0,0 +1,108 @@
...
+      - ar x PhotomatixLinux*.deb && tar xf data.tar.* && rm -f data.tar.* control.tar.* debian-binary
+      - cp -r usr/* /app/
+      - chmod +x /app/bin/photomatix
+      - mkdir -p /app/share/icons/hicolor/512x512/apps/
+      - cp /app/share/photomatix/icon.png /app/share/icons/hicolor/512x512/apps/com.hdrsoft.PhotomatixLinux.png
+      - mkdir -p /app/share/metainfo
+      - sed -e 's/<id>photomatix/<id>com.hdrsoft.PhotomatixLinux/' -e 's/photomatix.desktop/com.hdrsoft.PhotomatixLinux.desktop/' /app/share/metainfo/photomatix.metainfo.xml > /app/share/metainfo/com.hdrsoft.PhotomatixLinux.metainfo.xml
+      - rm /app/share/metainfo/photomatix.metainfo.xml
+      - mkdir -p /app/share/applications
+      - sed -e 's=/usr/share/photomatix/icon.png=com.hdrsoft.PhotomatixLinux=' -e 's=/usr/bin/photomatix=/app/bin/photomatix=' /app/share/applications/photomatix.desktop > /app/share/applications/com.hdrsoft.PhotomatixLinux.desktop
```
hfiguiere, line 94: Never use sed to change desktop files. It has no failure mode. Use the command `desktop-file-edit`.

com.hdrsoft.PhotomatixLinux.yaml
```diff
@@ -0,0 +1,108 @@
...
+      - rm /app/share/applications/photomatix.desktop
+      - mkdir -p /app/share/glib-2.0/schemas
+      - cp /app/share/photomatix/apps.photomatix.gschema.xml /app/share/glib-2.0/schemas
+      - glib-compile-schemas /app/share/glib-2.0/schemas/
+      - chmod a+r /app/share/glib-2.0/schemas/gschemas.compiled
+      - ar x libjpeg-turbo*.deb && tar -xf data.tar.* -C /app --wildcards --strip-components=2 ./usr/lib/x86_64-linux-gnu/libjpeg.so\* && rm -f data.tar.* control.tar.* debian-binary
+    sources: 
+      - type: file
+        url: https://www.hdrsoft.com/download/linux/ubuntu/PhotomatixLinux_3.0_Ubuntu23_24.deb
```
hfiguiere, line 103: is this redistributable?

com.hdrsoft.PhotomatixLinux.yaml
```diff
@@ -0,0 +1,108 @@
...
+        sha256: a2617a4585d0fbae8cb294b90d1dc5d338fc78d1641adb97ac6145d51dfce49e
+      # Needed for libjpeg-8.so because deb file for Photomatix is built on Ubuntu 24 and that uses a different libjpeg than the Gnome/48 platform.
```
hfiguiere, line 105: that mean binary is mis-compiled. Seeing this relies on Gtkmm too is another clue.

gtkmm.yaml
```diff
@@ -0,0 +1,93 @@
+name: gtkmm-stack
```
hfiguiere, line 1: since this is not autogenerated it shouldn't be a separate file.

gtkmm.yaml
```diff
@@ -0,0 +1,93 @@
...
+    sources:
+    - url: https://download.gnome.org/sources/mm-common/1.0/mm-common-1.0.3.tar.xz
+      sha256: e81596625899aacf1d0bf27ccc2fcc7f373405ec48735ca1c7273c0fbcdc1ef5
+      type: archive
+    cleanup:
+    - "*"
+  - name: sigc++-2
+    config-opts:
+    - "--disable-documentation"
+    - "--libdir=/app/lib"
```
hfiguiere, line 14: you NEVER need to set libdir. Make sure to use the latest flatpak-builder as per the flathub documentation

hfiguiere: same applies below.

com.hdrsoft.PhotomatixLinux.yaml
hfiguiere: and this is missing global cleanup rules to remove all the development only files.

com.hdrsoft.PhotomatixLinux.yaml
```diff
@@ -0,0 +1,108 @@
...
+        type: git
+        tag-pattern: "^(\\d+\\.\\d+[02468])$"
+      type: git
+    cleanup:
+    - "*.pod"
+    modules:
+    - name: perl
+      buildsystem: simple
+      sources:
+      - url: https://github.com/Perl/perl5.git
```
hfiguiere, line 34: use the tarballs instead.

com.hdrsoft.PhotomatixLinux.yaml
```diff
@@ -0,0 +1,108 @@
...
+build-options:
+  env:
+    PKG_CONFIG_PATH: /app/lib/pkgconfig:/app/share/pkgconfig:/usr/lib/pkgconfig:/usr/share/pkgconfig:/app/lib64/pkgconfig
+modules:
+  - gtkmm.yaml
+  - name: exiftool
+    buildsystem: simple
+    sources:
+    - url: https://github.com/exiftool/exiftool.git
+      commit: 0f63a780906abcccba796761fc2e66a0737e2f16
```
hfiguiere, line 23: add the tag here as well.

com.hdrsoft.PhotomatixLinux.yaml
```diff
@@ -0,0 +1,108 @@
...
+  - name: photomatix
+    buildsystem: simple
+    build-commands:
+      - ar x PhotomatixLinux*.deb && tar xf data.tar.* && rm -f data.tar.* control.tar.* debian-binary
+      - cp -r usr/* /app/
+      - chmod +x /app/bin/photomatix
+      - mkdir -p /app/share/icons/hicolor/512x512/apps/
+      - cp /app/share/photomatix/icon.png /app/share/icons/hicolor/512x512/apps/com.hdrsoft.PhotomatixLinux.png
+      - mkdir -p /app/share/metainfo
+      - sed -e 's/<id>photomatix/<id>com.hdrsoft.PhotomatixLinux/' -e 's/photomatix.desktop/com.hdrsoft.PhotomatixLinux.desktop/' /app/share/metainfo/photomatix.metainfo.xml > /app/share/metainfo/com.hdrsoft.PhotomatixLinux.metainfo.xml
```
hfiguiere, line 91: don't use sed here.

hfiguiere:
bot, build

com.hdrsoft.PhotomatixLinux.yaml
```diff
@@ -0,0 +1,108 @@
...
+runtime-version: '48'
+sdk: org.gnome.Sdk
+command: photomatix
+finish-args:
+  - '--share=ipc'
+  - '--socket=wayland'
+  - '--socket=fallback-x11'
+  - '--share=network'
+  - '--device=dri'
+  - '--filesystem=host'
```
bbhtt, line 12: why is host needed? it does not use the portals?

com.hdrsoft.PhotomatixLinux.yaml
mcgddd: It's a binary, and it uses gtkmm.  How would you suggest to do this better?  Is there a workable way that the product can be on flathub?  Our build process currently packages the flatpak from the .deb package that we create for Ubuntu 23/24.

com.hdrsoft.PhotomatixLinux.yaml
mcgddd: This was outlined in the submission.  Portals work for much of the filesystem access but not all.  We have a batch feature where people can choose a folder, or files within a folder, to process.  The output folder is by default in the folder of the files they chose.  This doesn't work with portals without requiring the users to go through extra user-unfriendly hoops.

PR #7960, opened 2026-03-01
Add io.github.bothlab.omerewriter

ximion:

  OMERewriter is a small tool that allows editing OME metadata for images generated by optical microscopes.

[20260301_101807.webm](https://github.com/user-attachments/assets/ddc24422-bb67-4953-b2e7-065de40cd46d)

### DIFF
--- a/io.github.bothlab.omerewriter.yaml
+++ b/io.github.bothlab.omerewriter.yaml
@@ -1,0 +1,124 @@
+id: io.github.bothlab.omerewriter
+runtime: org.kde.Platform
+runtime-version: '6.10'
+sdk: org.kde.Sdk
+command: OMERewriter
+rename-icon: omerewriter
+
+finish-args:
+  - --device=dri
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --filesystem=home
+  - --filesystem=/run/media
+  - --filesystem=/mnt
+  - --filesystem=/srv
+
+cleanup:
+  - /include
+  - /lib/pkgconfig
+  - "*.a"
+  - "*.la"
+
+modules:
+  - name: spdlog
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DBUILD_SHARED_LIBS=ON
+      - -DSPDLOG_BUILD_EXAMPLE=OFF
+      - -DSPDLOG_BUILD_TESTS=OFF
+      - -DSPDLOG_BUILD_BENCH=OFF
+      - -DSPDLOG_INSTALL=ON
+      - -DSPDLOG_FMT_EXTERNAL=ON
+    sources:
+      - type: archive
+        url: https://github.com/gabime/spdlog/archive/refs/tags/v1.17.0.tar.gz
+        sha512: 8df117055d19ff21c9c9951881c7bdf27cc0866ea3a4aa0614b2c3939cedceab94ac9abaa63dc4312b51562b27d708cb2f014c68c603fd1c1051d3ed5c1c3087
+
+  - name: gtest
+    buildsystem: cmake-ninja
+    sources:
+      - type: archive
+        url: https://github.com/google/googletest/archive/refs/tags/v1.17.0.tar.gz
+        sha256: 65fab701d9829d38cb77c14acdc431d2108bfdbf8979e40eb8ae567edf10b27c
+        x-checker-data:
+          type: anitya
+          project-id: 18290
+          url-template: https://github.com/google/googletest/archive/refs/tags/v$version.tar.gz
+    cleanup:
+      - '*'
+
+  - name: python3-six
+    buildsystem: simple
+    build-commands:
+    - pip3 install --verbose --exists-action=i --no-index --find-links="file://${PWD}"
+      --prefix=${FLATPAK_DEST} "six" --no-build-isolation
+    sources:
+    - type: file
+      url: https://files.pythonhosted.org/packages/b7/ce/149a00dd41f10bc29e5921b496af8b574d8413afcd5e30dfa0ed46c2cc5e/six-1.17.0-py2.py3-none-any.whl
+      sha256: 4721f391ed90541fddacab5acf947aa0d3dc7d27b2e1e8eda2be8970586c3274
+
+  - name: python3-Genshi
+    buildsystem: simple
+    build-commands:
+    - pip3 install --verbose --exists-action=i --no-index --find-links="file://${PWD}"
+      --prefix=${FLATPAK_DEST} "Genshi" --no-build-isolation
+    sources:
+    - type: file
+      url: https://files.pythonhosted.org/packages/27/69/259f8a1edc358a72c511b145c33b85a7667804a5c8ba04991aa5737d38a4/genshi-0.7.10-py3-none-any.whl
+      sha256: 1c45aa70c06f2fcab68137f040890e4fb105cc539f0121cb86d750656dafc37c
+
+  - name: boost
+    buildsystem: simple
+    build-commands:
+      - ./bootstrap.sh --prefix=/app
+      - ./b2 headers -j $FLATPAK_BUILDER_N_JOBS
+      - ./b2 --build-type=minimal -j $FLATPAK_BUILDER_N_JOBS install
+    sources:
+      - type: archive
+        url: https://archives.boost.io/release/1.90.0/source/boost_1_90_0.tar.bz2
+        sha256: 49551aff3b22cbc5c5a9ed3dbc92f0e23ea50a0f7325b0d198b705e8ee3fc305
+
+  - name: ome-common
+    buildsystem: cmake-ninja
+    builddir: true
+    sources:
+      - type: git
+        url: https://gitlab.com/codelibre/ome/ome-common-cpp.git
+        commit: fa4cf3dacc72bd14037f2a48d42fd31bf182da2e
+
+  - name: ome-model
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+      - -DBUILD_TESTS=OFF
+      - -Dxmldom=qt6
+    sources:
+      - type: git
+        url: https://gitlab.com/codelibre/ome/ome-model.git
+        commit: 99ab3fd25c27b3b625047bb0bc535201a611cd01
+
+  - name: ome-files
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+      - -DBUILD_TESTS=OFF
+      - -Dxmldom=qt6
+    sources:
+      - type: git
+        url: https://gitlab.com/codelibre/ome/ome-files-cpp.git
+        commit: d1abffaa9c1d2143735ce254754410e591882bf2
+      - type: patch
+        path: patches/omefiles_gcc15-build-fix.patch
+
+  - name: omerewriter
+    buildsystem: cmake-ninja
+    builddir: true
+    sources:
+      - type: git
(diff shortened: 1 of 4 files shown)


io.github.bothlab.omerewriter.yaml
```diff
@@ -0,0 +1,123 @@
...
+rename-icon: omerewriter
+
+finish-args:
+  - --device=dri
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --filesystem=home
+  - --filesystem=/media
+  - --filesystem=/mnt
+  - --filesystem=/srv
```
CodedOre, line 16: Why does the application not use the FileChooser portal?

io.github.bothlab.omerewriter.yaml
```diff
@@ -0,0 +1,123 @@
...
+        url: https://archives.boost.io/release/1.90.0/source/boost_1_90_0.tar.bz2
+        sha256: 49551aff3b22cbc5c5a9ed3dbc92f0e23ea50a0f7325b0d198b705e8ee3fc305
+
+  - name: ome-common
+    buildsystem: cmake-ninja
+    builddir: true
+    sources:
+      - type: git
+        url: https://gitlab.com/codelibre/ome/ome-common-cpp.git
+        tag: fa4cf3dacc72bd14037f2a48d42fd31bf182da2e
```
CodedOre: Use a proper tag with an additional commit reference.

io.github.bothlab.omerewriter.yaml
```diff
@@ -0,0 +1,123 @@
...
+  - name: ome-model
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+      - -DBUILD_TESTS=OFF
+      - -Dxmldom=qt6
+    sources:
+      - type: git
+        url: https://gitlab.com/codelibre/ome/ome-model.git
+        tag: 99ab3fd25c27b3b625047bb0bc535201a611cd01
```
CodedOre: Use a proper tag with an additional commit reference.

io.github.bothlab.omerewriter.yaml
```diff
@@ -0,0 +1,123 @@
...
+  - name: ome-files
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+      - -DBUILD_TESTS=OFF
+      - -Dxmldom=qt6
+    sources:
+      - type: git
+        url: https://gitlab.com/codelibre/ome/ome-files-cpp.git
+        tag: d1abffaa9c1d2143735ce254754410e591882bf2
```
CodedOre: Use a proper tag with an additional commit reference.

io.github.bothlab.omerewriter.yaml
```diff
@@ -0,0 +1,123 @@
...
+      - type: patch
+        path: patches/omefiles_gcc15-build-fix.patch
+
+  - name: omerewriter
+    buildsystem: cmake-ninja
+    builddir: true
+    sources:
+      - type: git
+        url: https://github.com/bothlab/omerewriter.git
+        tag: 357f2a01eb67ebe64387fba934026841fce11cf2
```
CodedOre: Use a proper tag with an additional commit reference.

io.github.bothlab.omerewriter.yaml
ximion: It needs access to the directory where the files are located to load referenced resources and write back data under a different name, and that's simply implemented using direct access.

io.github.bothlab.omerewriter.yaml
ximion: Cannot be done since this library does not have a release new enough for that and upstream also builds from this Git commit to get the latest round of fixes (ideally they'll make a release eventually, but there hasn't been one in years - everyone builds from Git).

io.github.bothlab.omerewriter.yaml
ximion: Cannot be done since this library does not have a release new enough for that and upstream also builds from this Git commit to get the latest round of fixes (ideally they'll make a release eventually, but there hasn't been one in years - everyone builds from Git).

io.github.bothlab.omerewriter.yaml
ximion: Cannot be done since this library does not have a release new enough for that and upstream also builds from this Git commit to get the latest round of fixes (ideally they'll make a release eventually, but there hasn't been one in years - everyone builds from Git).

io.github.bothlab.omerewriter.yaml
ximion: This is to include https://github.com/bothlab/omerewriter/commit/357f2a01eb67ebe64387fba934026841fce11cf2 so the Flatpak build actually works - will be switched to a proper tag once an actual upstream release is made. If it's absolutely necessary, I can make this a tag and include the patch as separate file, but that seemed like extra work for no benefit at the moment.

ximion:
Since this fails because the validator rejects building directly from Git commits, how am I supposed to use modules which can currently only be built if you use a direct Git commit? (the case for anything except for the "omerewriter" module itself, which I could change to become git tag + patch)

All I can think of is uploading a tarball of that code at a location I control and cheat the system that way, but that surely can't be the intended solution? Trying to get upstream to make releases is of course also an idea, but I don't know how successful that would be...

io.github.bothlab.omerewriter.yaml
CodedOre: Then you should use a `commit` field, not a `tag` one. 

Also, do note the [stable release requirement](https://docs.flathub.org/docs/for-app-authors/requirements#stable-releases).

io.github.bothlab.omerewriter.yaml
CodedOre: As above, use `commit` instead of `tag`.

io.github.bothlab.omerewriter.yaml
CodedOre: As above, use `commit` instead of `tag`.

io.github.bothlab.omerewriter.yaml
CodedOre: This will need a new release before the submission will be approved then. See the [Stable release requirement](https://docs.flathub.org/docs/for-app-authors/requirements#stable-releases).

CodedOre:
The reason the linter fails is because you used the wrong fields in the manifest. 

For your application code, you will need to create a new release, as per [stable release requirement](https://docs.flathub.org/docs/for-app-authors/requirements#stable-releases).

As for the dependencies, I can't give an authoritive answer. If possible, a new release from upstream should be used.

ximion:
Yeah, I was blind and didn't see the wrong field.

> For your application code, you will need to create a new release, as per [stable release requirement](https://docs.flathub.org/docs/for-app-authors/requirements#stable-releases).

Or I just include that one change as a patch (done now).

Those libraries are ancient at this point, and stable - if ever a new release is made, I will update this, of course.

bot, build

ximion:
`finish-args-home-filesystem-access` remains - OME-TIFF and this app aren't really set up for sandboxed filesystem access and very much care about having access to absolute filepaths.

CodedOre:
Then you need to request a linter exception. See the [documentation](https://docs.flathub.org/docs/for-app-authors/linter#exceptions) for details.

io.github.bothlab.omerewriter.yaml
hfiguiere: The standard path for mounted drives is `/run/media`. The legacy if `/media`. None of they others are.

io.github.bothlab.omerewriter.yaml
```diff
@@ -0,0 +1,127 @@
...
+
+finish-args:
+  - --device=dri
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --filesystem=home
+  - --filesystem=/media
+  - --filesystem=/mnt
+  - --filesystem=/srv
```
hfiguiere:
```suggestion
  - --filesystem=/media
  - --filesystem=/run/media
```

hfiguiere:
> Since this fails because the validator rejects building directly from Git commits,

No it doesn't. You put a commit in `tag` and when there is a tag we require a `commit` because people retag endlessly. So it protects against it. The messages are actually documented.

io.github.bothlab.omerewriter.yaml
ximion: This is a scientific application, and large amounts of its data are either on big network drives (mounted on the systems at `/mnt`) or on local disk arrays (often under `/srv`). I can drop `/media`, but dropping the others would render the tool pretty much useless for our lab.

This is a really common problem with Flatpak apps, maybe I should introduce a `--filesystem=data` permission upstream, finally... In the meantime, allowing those two locations fixes the issue.

CodedOre: You should consider that the default permissions should reflect an ordinary users computer. 
And, without special mount settings, drives are mounted to `/run/media` or `/media`.

Changing the default permission as suggested makes sense to me that way. 

For your lab, you can always provide an override to match how drives are mounted there.

ximion: It's not just my lab, this is the setup literally everywhere I have ever worked, and the failure mode of "file that I know exists is suddenly not found" is so confusing and bad (especially when drag & drop is involved, where the dragged file just vanishes) that I would much rather have this as default instead of receiving tons of bug reports upstream about the software not working (this is not theoretical, it *has* happened with other software until the default was changed).

This issue does not exist if a portal is used for file selection and the file is tunneled through, but that's not a scenario the OME-TIFF file format was ever designed for.

ximion:
> No it doesn't. You put a commit in `tag`

Yeah, I was completely blind and thought I had used `commit:` and that it was rejecting that tag. It's already fixed.

io.github.bothlab.omerewriter.yaml
```diff
@@ -0,0 +1,127 @@
...
+    config-opts:
+      - -DBUILD_SHARED_LIBS=ON
+      - -DSPDLOG_BUILD_EXAMPLE=OFF
+      - -DSPDLOG_BUILD_TESTS=OFF
+      - -DSPDLOG_BUILD_BENCH=OFF
+      - -DSPDLOG_INSTALL=ON
+      - -DSPDLOG_FMT_EXTERNAL=ON
+    cleanup:
+      - /include
+      - '*.a'
```
bbhtt:
```suggestion
```

dupe

io.github.bothlab.omerewriter.yaml
```diff
@@ -0,0 +1,127 @@
...
+  - --filesystem=home
+  - --filesystem=/run/media
+  - --filesystem=/mnt
+  - --filesystem=/srv
+
+cleanup:
+  - /include
+  - /lib/pkgconfig
+  - /lib/*.a
+  - /lib/*.la
```
bbhtt:
```suggestion
  - "*.a"
  - "*.la"
```

io.github.bothlab.omerewriter.yaml
bbhtt: can you explain more about "load referenced resources"? is this like a .md file referencing images situation?

ximion: Kind of, only that the referenced stuff isn't images, but other TIFFs and/or XML files that represent a coherent image.

bbhtt:
bot, build

ximion:
The build Flatpak works fine on my system :-)

PR #7981, opened 2026-03-03
Assign reviewers for eligible submissions

barthalion:
This is a very poor man's attempt to spread the load across reviewers and act upon submissions only when they are actually ready to be reviewed. For now, it means "not blocked by the pr-check", we can extend it to "it builds" in the future.

The list of usernames is hardcoded as the default token doesn't have permissions to list team members.

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -174,6 +174,25 @@ jobs:
               return 1
             }
 
+            assign_reviewer() {
+              local existing
+              existing=$(gh pr view "$PR_NUM" --json reviewRequests --jq '.reviewRequests | length' 2>/dev/null)
+              if [ "${existing:-0}" -gt 0 ]; then
+                echo "PR already has reviewers assigned, skipping"
+                return
+              fi
+              local pr_author
+              pr_author=$(gh pr view "$PR_NUM" --json author --jq '.author.login' 2>/dev/null)
+              local reviewer
+              reviewer=$(printf '%s\n' bbhtt barthalion hfiguiere \
+                | grep -v "^${pr_author}$" \
+                | shuf -n 1)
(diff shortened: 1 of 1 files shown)


barthalion:
Well, this is silly. @bbhtt @hfiguiere PTAL either way.

bbhtt:
I'm not sure what assigning is to a PR gives us? 

We already review every PR together and it doesn't necessarily mean someone is not going to review another person's assigned PR.

More eyes on a PR is better a single person could have missed something that another may catch. We do multiple passes each on a PR usually.

barthalion:
Ideally, we should stop spending time on submissions that don't check all initial requirements and later not building at all. I want to optimize the process for reducing the burden as we are all pretty burnt out at this point. This does not mean a PR can't have another reviewer, but at least clearly designates who should do the initial review while other admins can focus on different PRs.

Eventually, if we manage to have more reviewers than 3 (say, 5), we could assign pairs, or assign another reviewer only when the PR is marked as ready to enforce two-pass review.

bbhtt:
> Ideally, we should stop spending time on submissions that don't check all initial requirements

I already do that.

> but at least clearly designates who should do the initial review while other admins can focus on different PRs.

What does that do though? As it stands anyone can do the initial review (or any review). Why would someone need to be 'assigned' to do it?

I fear this will make the submitter unnecessarily ping the person when they see someone getting assigned. It's also an extra notifications for us who are already subscribed here.

hfiguiere:
> Ideally, we should stop spending time on submissions that don't check all initial requirements and later not building at all.

I already do that, or tend to. The thing is the bots don't run consistently in a timely fashion. Sometime it takes a very long time to get the pr-check, or the domain check or the first test build.

Assigning reviewers will make things worse as OP will feel even more entitled.

PR #7990, opened 2026-03-04
Add com.neosubhamoy.neodlp

neosubhamoy:

- [X] Please describe the application briefly. NeoDLP is a modern video/audio downloader based on yt-dlp with browser integration

https://github.com/user-attachments/assets/e8828595-0fdf-4799-9ae1-a57aa887ef79

- [X] I am the core developer/author of the project.

### DIFF
--- a/com.neosubhamoy.neodlp.yml
+++ b/com.neosubhamoy.neodlp.yml
@@ -30,7 +30,6 @@ modules:
   - name: aria2
     buildsystem: autotools
     config-opts:
-      - --prefix=/app
       - --with-ca-bundle=/etc/ssl/certs/ca-certificates.crt
     sources:
       - type: git


com.neosubhamoy.neodlp.yml
```diff
@@ -0,0 +1,97 @@
...
+
+command: neodlp
+finish-args:
+  - --socket=wayland # Show app window on wayland
+  - --socket=fallback-x11 # Fallback to X11 if wayland is not available
+  - --device=dri # Use OpenGL
+  - --share=ipc # Needed for good X11 performance
+  - --share=network # Allow network access
+  - --filesystem=xdg-download # Allow access to downloads folder for storing downloaded files
+  - --filesystem=~/.config/google-chrome/NativeMessagingHosts:create # Allow read-write (create) access to following browser native messaging host configs for neodlp browser extension manifest.json installation
```
bbhtt: I suggest dropping the browser extension this is not scalable at all.

com.neosubhamoy.neodlp.yml
```diff
@@ -0,0 +1,97 @@
...
+  - --filesystem=~/.config/chromium:ro
+  - --filesystem=~/.config/microsoft-edge:ro
+  - --filesystem=~/.mozilla/firefox:ro
+  - --filesystem=~/.config/mozilla/firefox:ro
+  - --filesystem=~/snap/firefox/common/.mozilla/firefox:ro
+  - --filesystem=~/.config/opera:ro
+  - --filesystem=~/.config/vivaldi:ro
+  - --filesystem=~/.config/naver-whale:ro
+  - --env=FLATPAK=1 # Used to determine flatpak environment on app side
+  - --talk-name=org.freedesktop.Notifications # Needed for showing desktop notifications
```
bbhtt: this should use the portal.

com.neosubhamoy.neodlp.yml
neosubhamoy: Well, browser integration is one of the core features of NeoDLP! I guess removing it would be a pretty big downside for users to use the Flatpak build! Are you sure I should remove it...?? :(

Also, I'm thinking about doing the browser integration via direct HTTP (if it's possible) instead of the NativeMessaging API (Then those filesystem accesses wouldn't be needed). But, for now, can we just keep it...??
I know that direct host filesystem read/write access is not encouraged in Flatpaks! But, sometimes it's just needed (I tried to already narrow down the filesystem scope as specifically as possible)

com.neosubhamoy.neodlp.yml
neosubhamoy: Actually, NeoDLP is built with Tauri v2! And, the app uses the [tauri-plugin-notification](https://v2.tauri.app/plugin/notification/) plugin to show desktop notifications! I followed this blogpost on how to package Tauri v2 apps for Flatpak : https://vincent.jousse.org/blog/en/packaging-tauri-v2-flatpak-snapcraft-elm/#base-configuration-file (from where I picked the talk permission). So, I don't really know the upstream implementation of tauri-plugin-notification and how it handles desktop notifications! (In short: I'm not really sure if using a portal is possible here)

com.neosubhamoy.neodlp.yml
CodedOre: But this way you will need to provide new exceptions for every possible browser your extension supports, meaning new exceptions to the linter, blocked builds in the meantime, browsers you don't explicitly support will not work right away...

In short, a maintaince nightmare. 

From what I can tell from your video the browser extension also only sends the YouTube link to the downloader?
In this case, you could simply use an custom URI schema. Set a mime type like `x-scheme-handler/my-app` in the desktop file, then your extension can open the `my-app://` uri to send data to your APP without any sandbox hole required.

sidecar-sources.json
CodedOre, line 1: Why aren't the programs in this file build from source?

sidecar-sources.json
```diff
@@ -0,0 +1,158 @@
+[
+    {
+        "type": "file",
+        "url": "https://github.com/yt-dlp/yt-dlp-nightly-builds/releases/download/2026.03.02.233544/yt-dlp_linux",
+        "sha256": "857325e8aeb7c627cf6969fb1a638fd0b73e5eaa62921fb48d2561a3fd5c26f9",
```
CodedOre: Using nightly builds is not a good idea. You will either have constant updates or you will have development software that is outdated as soon as it is downloaded. 

Both are not good ideas. 

Use an stable release of `yt-dlp`.

See also the [Stable Releases requirement](https://docs.flathub.org/docs/for-app-authors/requirements#stable-releases).

sidecar-sources.json
```diff
@@ -0,0 +1,158 @@
...
+        "sha256": "cca70eb486ce053d2f3c53dec73c6482795d6e487200ebb0ff4b365096e66d0f",
+        "dest": "src-tauri/binaries",
+        "dest-filename": "yt-dlp-aarch64-unknown-linux-gnu",
+        "only-arches": [
+            "aarch64"
+        ]
+    },
+    {
+        "type": "archive",
+        "url": "https://github.com/yt-dlp/FFmpeg-Builds/releases/download/autobuild-2026-03-02-14-51/ffmpeg-N-123108-g27cdf70086-linux64-gpl.tar.xz",
```
CodedOre: Why is it not possible to use the FFmpeg provided in the runtime?

Also, again the same point about using nightlies.

com.neosubhamoy.neodlp.yml
neosubhamoy: Got it! Thanks for your suggestion. Using URI scheme would be a good option. I will try to implement it later! But for now, I'm just removing the browser integration :(

sidecar-sources.json
neosubhamoy: Because building all those binaries from source would be a nightmare for me! (especially yt-dlp, deno, etc.) Also, I don't see that there's any hardened rule to build everything from source (if there are already pre-built binaries available). I mean, it's absolutely possible to build everything from source. But I don't see that the extra effort would be worth it!

sidecar-sources.json
neosubhamoy: Yah! It's because `yt-dlp` recommends using the `nightly` build in production these days! The stable version of yt-dlp is often released once every few months (which gets outdated pretty easily due to YouTube and other sites changing drastically fast these days)

sidecar-sources.json
neosubhamoy: This version of FFmpeg is patched by yt-dlp for the best compatibility (So it works the best with yt-dlp)
And, the same point I said about yt-dlp nightly also applies here!

sidecar-sources.json
CodedOre: > Also, I don't see that there's any hardened rule to build everything from source (if there are already pre-built binaries available).

Bildschirmfoto vom 2026-03-04 10-22-18

neosubhamoy: I see! But, so-called 'well-known' vendors are an exception, right? (and I guess yt-dlp, FFmpeg, Deno, etc. are well-known vendors...!! aren't they...??)
Also, I did it this way because I saw other similar apps like: 'Parabolic' is also doing it in this way (they also use pre-built binaries): https://github.com/flathub/org.nickvision.tubeconverter/blob/master/org.nickvision.tubeconverter.json
And all the external sidecar binaries I used are open-sourced and available on GitHub, so this should not be an issue! Correct me if I'm wrong....!!!

CodedOre: > Also, I did it this way because I saw other similar apps like […]

Just to note:
The requirements have changed over time. And submissions are reviewed based on the current requirements and guidelines. Other apps don't serve as a reason for exceptions. If anything, those should be updated to match new requirements if possible.

> I see! But, so-called 'well-known' vendors are an exception, right? (and I guess yt-dlp, FFmpeg, Deno, etc. are well-known vendors...!! aren't they...??)

As the requirements mention, the exceptions are for exceptional circumstances. They should not be taken as a base assumption.

Also, the "vendor" in question is currently you, as you want to package these dependencies with your app. And you are not "well-known".

> And all the external sidecar binaries I used are open-sourced and available on GitHub, so this should not be an issue! Correct me if I'm wrong....!!!

Being open-source on GitHub is not reason enough for an exception. Otherwise we'd have to give the exception to almost any submission.

sidecar-sources.json
CodedOre: But that would still imply you want to update your application every day at least.

I'd say it would be better to provide a stable version as a base line and then provide the nightlies as in-application update to the application cache, similar to how "Parabolic" does it.

neosubhamoy: Ok, fine! I will use yt-dlp stable branch. NeoDLP does have yt-dlp auto updates. But I currently disabled it for Flatpak. Since/app/bin is not writable, I will try to implement the app cache way of updating later!

sidecar-sources.json
CodedOre: > This version of FFmpeg is patched by yt-dlp for the best compatibility.

If I look at the [ReadMe of the repository](https://github.com/yt-dlp/FFmpeg-Builds?tab=readme-ov-file#patches-applied), I see this:

Bildschirmfoto vom 2026-03-04 11-16-22

So, is it really neccessary then?

neosubhamoy: Oh! I see...!! My bad! They might have recently revoked the patches. I guess the upstream FFmpeg has caught up (but as you can see, there used to be some patches historically). In that case, I guess we can use the flatpak ffmpeg runtime :)

sidecar-sources.json
neosubhamoy: Ok! So, you're just telling me to compile all these binaries: (yt-dlp, deno, aria2c, yt-dlp-pot-provider-rs) from source. And, to use the FFmpeg from the Flatpak runtime...!! Right...?? Kindly be specific...!! Also, currently, I don't have any idea how I will compile all of them! and, even if all of their build systems are compatible with offline builds or not....!! I know I will have a hard time figuring out all of these on my own...!! Any help is appreciated :(

neosubhamoy:
So, I'm proceeding with the following changes (as requested):
 - Disable browser integration features (and remove related filesystem permissions)
 - Disable desktop notification features (and remove related talk permission)
 - Use the default runtime `ffmpeg` shipped with Gnome 49 Sdk
 - Build stable `yt-dlp`, `aria2c`, and `neodlp-pot` (bgutil-ytdlp-pot-provider-rs) from source
 - Use the stable `deno` and its pre-compiled static `rusty_v8` library (as compiling it from source is a highly resource-intensive and complex task)

I guess these changes are enough...!! Kindly review these changes and do let me know if any additional changes are required...!! :)

neosubhamoy:
I have made the requested changes! Kindly review it :)

neosubhamoy:
bot, build

com.neosubhamoy.neodlp.yml
neosubhamoy: I disabled the desktop notification feature for now....!!

com.neosubhamoy.neodlp.yml
```diff
@@ -0,0 +1,165 @@
...
+  - --filesystem=~/.config/BraveSoftware/Brave-Browser:ro # Allow read-only access to following browser configs for yt-dlp cookie extraction
+  - --filesystem=~/.config/google-chrome:ro
+  - --filesystem=~/.config/chromium:ro
+  - --filesystem=~/.config/microsoft-edge:ro
+  - --filesystem=~/.mozilla/firefox:ro
+  - --filesystem=~/.config/mozilla/firefox:ro
+  - --filesystem=~/snap/firefox/common/.mozilla/firefox:ro
+  - --filesystem=~/.config/opera:ro
+  - --filesystem=~/.config/vivaldi:ro
+  - --filesystem=~/.config/naver-whale:ro
```
CodedOre: First, use `xdg-config` instead of `~/.config`. That way `XDG_CONFIG_HOME` is taken into account if set. 

Second, that are still quite a lot of permissions. Also a bit inconsistent, like you account for an Firefox Snap, but not an Firefox Flatpak.

com.neosubhamoy.neodlp.yml
```diff
@@ -0,0 +1,165 @@
...
+   
+  # yt-dlp stable python module -> python3 flatpak-pip-generator.py --runtime='org.gnome.Sdk//49' yt-dlp[default,curl-cffi]
+  - python3-yt-dlp.json
+
+  - name: aria2
+    buildsystem: simple
+    sources:
+      - type: git
+        url: https://github.com/aria2/aria2.git
+        commit: b4fd7cb1ca03e38ad9d7ab9308b8200cb1d41c25
```
CodedOre: This should, additionally, point to a release tag.

com.neosubhamoy.neodlp.yml
neosubhamoy: Ok, I will disable the cookies extraction feature too...!! Actually I took those permissions from Parabolic's implementation: https://github.com/flathub/org.nickvision.tubeconverter/blob/master/org.nickvision.tubeconverter.json

com.neosubhamoy.neodlp.yml
neosubhamoy: Go it..!! Will add the tag...!!

neosubhamoy:
bot, build

neosubhamoy:
@CodedOre Fixed those...!!! Anything else...??? :)

com.neosubhamoy.neodlp.yml
```diff
@@ -0,0 +1,156 @@
...
+  append-path: /usr/lib/sdk/node24/bin:/usr/lib/sdk/rust-stable/bin
+
+modules:
+  - shared-modules/libappindicator/libappindicator-gtk3-12.10.json
+   
+  # yt-dlp stable python module -> python3 flatpak-pip-generator.py --runtime='org.gnome.Sdk//49' yt-dlp[default,curl-cffi]
+  - python3-yt-dlp.json
+
+  - name: aria2
+    buildsystem: simple
```
bbhtt: use `autotools` and `config-opts`.

com.neosubhamoy.neodlp.yml
```diff
@@ -0,0 +1,156 @@
...
+        url: https://github.com/jim60105/bgutil-ytdlp-pot-provider-rs.git
+        tag: v0.7.2
+        commit: ed6fc38eb6ea6755979b7cba9507888ea167c4a2
+      # Rust package sources -> python3 flatpak-cargo-generator.py Cargo.lock -o flatpak/neodlp-pot-cargo-sources.json
+      - neodlp-pot-cargo-sources.json
+      # rusty_v8 static library sources
+      - type: file
+        url: https://github.com/denoland/rusty_v8/releases/download/v130.0.7/librusty_v8_release_x86_64-unknown-linux-gnu.a.gz
+        sha256: a6476cb94e9b0247081c465424eb793d77732b8db80848132d78cbb50f34b75d
+        dest-filename: rusty_v8-x86_64.a.gz
```
bbhtt: what is this and why is it not built from source?

com.neosubhamoy.neodlp.yml
```diff
@@ -0,0 +1,156 @@
...
+      # Install Rust packages offline
+      - cargo --offline fetch --manifest-path Cargo.toml --verbose
+      # Copy rusty_v8 static library to proper cargo registry location for offline build
+      - mkdir -p /run/build/neodlp-pot/cargo/.rusty_v8
+      - |
+        if [ "$(uname -m)" = "x86_64" ]; then
+          cp rusty_v8-x86_64.a.gz /run/build/neodlp-pot/cargo/.rusty_v8/https___github_com_denoland_rusty_v8_releases_download_v130_0_7_librusty_v8_release_x86_64_unknown_linux_gnu_a_gz
+        elif [ "$(uname -m)" = "aarch64" ]; then
+          cp rusty_v8-aarch64.a.gz /run/build/neodlp-pot/cargo/.rusty_v8/https___github_com_denoland_rusty_v8_releases_download_v130_0_7_librusty_v8_release_aarch64_unknown_linux_gnu_a_gz
+        fi
```
bbhtt: why is this needed? seems the target copy path is too complicated.

com.neosubhamoy.neodlp.yml
neosubhamoy: Ok!

com.neosubhamoy.neodlp.yml
neosubhamoy: `rusty_v8` is the Rust bindings for Chrome's JavaScript V8 Engine (built by `deno`). see here: https://github.com/denoland/rusty_v8 , the [project](https://github.com/jim60105/bgutil-ytdlp-pot-provider-rs) NeoDLP relies on to generate YouTube PO Tokens depends on it (in short: one of the crates used on that project downloads the rusty_v8 engine on buildtime, and if I don't place it into the proper place the compilation fails as flatpak builds doesn't have network access, that's why I have to pre-download it and place it to the corrent location for that crate to find it out). And, it is not compiled from source because of that same reason: The crate expects the pre-compiled binary in that location, and also due to the fact that compiling the V8 engine from source is a highly resource-intensive and complex task. Anyways, I guess I will remove the PO Token feature too...!! :(

com.neosubhamoy.neodlp.yml
neosubhamoy: As, I said before:

> the [project](https://github.com/jim60105/bgutil-ytdlp-pot-provider-rs?rgh-link-date=2026-03-07T03%3A50%3A20Z) NeoDLP relies on to generate YouTube PO Tokens depends on it (in short: one of the crates used on that project downloads the rusty_v8 engine on buildtime, and if I don't place it into the proper place the compilation fails as flatpak builds doesn't have network access, that's why I have to pre-download it and place it to the corrent location for that crate to find it out)

For that reason: The path is kinda complicated...!! Anyways, I'm removing it...!! So, it doesn't matter anymore :(

com.neosubhamoy.neodlp.yml
```diff
@@ -0,0 +1,110 @@
...
+modules:
+  - shared-modules/libappindicator/libappindicator-gtk3-12.10.json
+   
+  # yt-dlp stable python module -> python3 flatpak-pip-generator.py --runtime='org.gnome.Sdk//49' yt-dlp[default,curl-cffi]
+  - python3-yt-dlp.json
+
+  - name: aria2
+    buildsystem: autotools
+    config-opts:
+      - --prefix=/app
```
hfiguiere:
```suggestion
```

neosubhamoy: Thanks, for suggesting...!! Removing it...!! :)

neosubhamoy:
Requested changes are done...!! Kindly review...!! :)

neosubhamoy:
bot, build

PR #8048, opened 2026-03-10
Add io.github.ostapkonst.HashVerifier

ostapkonst:

- [x] Please describe the application briefly. Description: a cross-platform application for generating and validating file checksums using multiple cryptographic hash algorithms.

https://github.com/user-attachments/assets/14073a51-836e-4155-8a45-07f0677ad787

### DIFF
--- a/io.github.ostapkonst.HashVerifier.yml
+++ b/io.github.ostapkonst.HashVerifier.yml
@@ -12,7 +12,6 @@ finish-args:
   - --device=dri
   - --filesystem=xdg-desktop
   - --filesystem=xdg-documents
-  - --filesystem=~/.config/hashverifier
 modules:
   - name: hashverifier
     buildsystem: simple


ostapkonst:
bot, build

ostapkonst:
Received 'appid-url-not-reachable'. Did I understand correctly that to get the PR through, I need either to reopen the PR with this App ID: io.github.ostapkonst.hash-verifier, or rename the repository to HashVerifier?

io.github.ostapkonst.HashVerifier.yml
```diff
@@ -0,0 +1,130 @@
...
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.golang
+command: hashverifier
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  # The program's main task is to read files and dirs to calculate the hashes
+  - --filesystem=host
```
teamcons: Why not use portals?

teamcons:
> Received 'appid-url-not-reachable'. Did I understand correctly that to get the PR through, I need either to reopen the PR with this App ID: io.github.ostapkonst.hash-verifier, or rename the repository to HashVerifier?

You dont need to open any new PR

teamcons:
Please lint and build locally first, and fix your mistakes there, do not abuse flathub infra

ostapkonst:
> > Received 'appid-url-not-reachable'. Did I understand correctly that to get the PR through, I need either to reopen the PR with this App ID: io.github.ostapkonst.hash-verifier, or rename the repository to HashVerifier?
> 
> You dont need to open any new PR

Renaming the upstream repository resolved the 'appid-url-not-reachable' error.

ostapkonst:
> Please lint and build locally first, and fix your mistakes there, do not abuse flathub infra

No problem. I launched it manually once.

io.github.ostapkonst.HashVerifier.yml
ostapkonst: It's difficult to apply the portal system here for the following reasons:

- The checksum files generated by the program can contain both relative paths (if the checksum file is saved next to the directory being hashed) and absolute paths (if the checksum file is saved in an arbitrary location). Therefore, it's important to be able to obtain correct **persistent absolute paths**. When using portals, instead of persistent paths, we would get paths like: `/run/user/<user-id>/doc/<random>/<folder-name>` in the checksum files. This doesn't suit us.

- Following symlinks is enabled by default; according to [this](https://github.com/flatpak/flatpak/issues/5227) ticket, for example, symlink support within portals is limited.

- The current implementation uses non-native dialogs, specifically `gtk_file_chooser_dialog_new`. These do not support portals, unlike the native `gtk_file_chooser_native_new`. I had to abandon `gtk_file_chooser_native_new` due to cross-platform compatibility (because of Windows), if I remember correctly.

bbhtt: > t persistent absolute paths. When using portals, instead of persistent paths, we would get paths like:

GetHostPaths exists in portal for this

> Following symlinks is enabled by default;

Disable it

> he current implementation uses non-native dialogs,

Switch it to use the portals.

ostapkonst: With this approach, we would lose some functionality — like following symlinks and adjusting the codebase based on how the app is distributed. Manual path input would also probably have to go, replaced entirely with dialog boxes.

Since apps shouldn't ask for extra permissions, I'd suggest other options:

- give read-only access to the `host` + read-write to `home`

- give read-only or read-write to `/mnt` + read-write to `home`

Maybe `home` could be replaced with something more specific.

Or another idea: show a dialog when the app starts, asking the user to enter commands to grant filesystem access — though this is less desirable than the options above.

ostapkonst: Since it’s clear from the discussion that restrictions can’t be avoided, I’ve limited file system access to the level required for the app to remain as functional.

ostapkonst:
bot, build

io.github.ostapkonst.HashVerifier.yml
ostapkonst: In the end, I limited access to just two dirs:
- --filesystem=xdg-desktop
- --filesystem=xdg-documents

ostapkonst: It could even be reduced to just `xdg-desktop`, as long as the user finds the app useful.

If even these permissions seem excessive from your perspective, then we could grant no filesystem access at all and let the user manually set permissions after installing the app. As I mentioned above, I can notify the user about these restrictions with a dialog box when the application starts.

PR #8049, opened 2026-03-10
Add io.github.q962.PictureWindow

q962:

  Loop display images

https://github.com/user-attachments/assets/73eb939b-3187-4372-be1d-2babf8b762c2

- [X] I am an  author contributor to the project.

### DIFF
--- a/io.github.q962.PictureWindow.yml
+++ b/io.github.q962.PictureWindow.yml
@@ -3,17 +3,12 @@ runtime: org.gnome.Platform
 runtime-version: "49"
 sdk: org.gnome.Sdk
 command: PictureWindow
-sdk-extensions:
-  - org.freedesktop.Sdk.Extension.rust-stable
 
 finish-args:
   - --socket=wayland
   - --socket=fallback-x11
   - --device=dri
   - --share=ipc
-
-build-options:
-  append-path: "/usr/lib/sdk/rust-stable/bin"
 
 modules:
   - name: xmake
@@ -26,51 +21,19 @@ modules:
     cleanup:
       - "*"
 
-  - name: grass
-    buildsystem: simple
-    build-options:
-      env:
-        CARGO_HOME: /run/build/grass/cargo
-    build-commands:
-      - cargo build --release --locked
-      - install -Dm 755 ./target/release/grass /app/bin/grass
-    cleanup:
-      - '*'
-    sources:
-      - type: git
-        url: https://github.com/connorskees/grass
-        tag: 0.13.4
-        commit: e0bb9e2eabfc3a58e42b03089cd7b22c68d09d0b
-      - grass-cargo-sources.json
-
-  - name: sass-wrapper
-    buildsystem: simple
-    build-commands:
-      - cp sass /app/bin/sass
-    sources:
-      - type: script
-        commands:
-          - grass $2 $3 # skip --no-source-map
-        dest-filename: sass
-    cleanup:
-      - '*'
-
   - name: PictureWindow
     buildsystem: simple
     build-options:
       env:
-        XMAKE_GLOBALDIR: /run/temp/xmake
+        XMAKE_GLOBALDIR: xmake-global
         XMAKE_ROOT: y
     build-commands:
       - xmake g --network=private
-      - xmake f --mode=releasedbg
-      - xmake b
+      - xmake f --mode=debug
+      - xmake b -vD
       - xmake install -o /app
-      - eu-unstrip /app/bin/PictureWindow /app/bin/PictureWindow.sym -o PictureWindow
-      - mv PictureWindow /app/bin
-      - rm /app/bin/*.sym
     sources:
       - type: git
         url: https://github.com/q962/PictureWindow
-        tag: v1.0.1
-        commit: 8d49346394956aaebd653e0ad49de7badf44e632
+        tag: v1.0.2
+        commit: 4f475c28e0cfadd355da3e50c6703f83fc6c51a6


q962:
bot, build

io.github.q962.PictureWindow.yml
```diff
@@ -0,0 +1,63 @@
...
+    sources:
+      - type: git
+        url: https://github.com/xmake-io/xmake.git
+        tag: v3.0.5
+        commit: f20295f255d967af9d91de6f1ed5eeadf17d30a7
+    cleanup:
+      - "*"
+
+  - name: sass
+    buildsystem: simple
```
CodedOre: Why is this not build from source?

Also, the GNOME SDK already contains `libsass` and `sassc`. Why is a `dart-sass` required for your project?

io.github.q962.PictureWindow.yml
```diff
@@ -0,0 +1,63 @@
...
+      - "*"
+
+  - name: PictureWindow
+    buildsystem: simple
+    build-commands:
+      - |
+        set -e
+
+        export XMAKE_GLOBALDIR=/run/temp/xmake
+        export XMAKE_ROOT=y
```
CodedOre: These should be set using `build-options`.

io.github.q962.PictureWindow.yml
```diff
@@ -0,0 +1,63 @@
...
+        xmake g --network=private
+
+        xmake f --mode=releasedbg
+        xmake b
+        xmake install -o /app
+
+        eu-unstrip /app/bin/PictureWindow /app/bin/PictureWindow.sym -o PictureWindow
+        mv PictureWindow /app/bin
+        rm /app/bin/*.sym
```
CodedOre: Debug symbols should not be part of the normal package. If placed at the correct location, Flatpak Builder should create an Debug extension for them automatically.

q962: My build tool strips the symbol files into a separate .sym file.

This results in my application lacking symbol information.

Therefore, I need to merge the code and let Flatpak handle the symbol file stripping.

q962:
bot, build

io.github.q962.PictureWindow.yml
q962: Why? I know `build-options` applies to all scripts. But I only have one script.

 If I split the script line by line, I will choose to use `build-options`.

io.github.q962.PictureWindow.yml
q962: Removing the sass module caused a compilation error. I need to investigate why.

q962: The GNOME SDK does not include Sass.

```
$ pkg-config --list-all | grep sass
$ find /usr | grep sass
```

I can't find Sass

q962: > Why is this not build from source?

LibSass has reached its end of life. 

Therefore, it needs to be migrated to dart-Sass.

I don't know how to compile dart-sass.

If doing so is not allowed, I will upload the compiled files.

q962:
Your answer will guide my next steps.

io.github.q962.PictureWindow.yml
CodedOre: > The GNOME SDK does not include Sass.

Right. I assumed `sassc` should be available, since it is a build dependency for libadwaita.

> LibSass has reached its end of life.
> Therefore, it needs to be migrated to dart-Sass.
> I don't know how to compile dart-sass.
> If doing so is not allowed, I will upload the compiled files.

Don't take this as a definitive answer, but [the Flatpak for Fractal](https://github.com/flathub/org.gnome.Fractal/blob/master/org.gnome.Fractal.json#L22-L45) uses [grass](https://github.com/connorskees/grass) as a SASS compiler instead, which as a Rust project can be build from source. Maybe this is a alternative?

q962:
bot, build

io.github.q962.PictureWindow.yml
```diff
@@ -0,0 +1,84 @@
...
+    buildsystem: simple
+    build-commands:
+      - cp sass /app/bin/sass
+    sources:
+      - type: script
+        commands:
+          - |
+            #!env bash
+            grass $2 $3 # skip --no-source-map
+        dest-filename: sass
```
CodedOre: Why not update the buildsystem so it can find and use `grass` as well as `sass`? \
Then this not needed.

io.github.q962.PictureWindow.yml
```diff
@@ -0,0 +1,84 @@
...
+
+  - name: grass
+    buildsystem: simple
+    build-options:
+      env:
+        CARGO_HOME: /run/build/grass/cargo
+    build-commands:
+      - cargo build --release --locked
+      - mkdir -p /app/bin
+      - install -D ./target/release/grass /app/bin/
```
CodedOre:
```suggestion
    build-commands:
      - cargo build --release --locked
      - install -Dm 755 ./target/release/grass /app/bin/grass
```

io.github.q962.PictureWindow.yml
q962: I won't choose to use Grass. Grass offers me no advantage; it lacks the `watch` command. 
So I'll choose Dart-Sass.

I can't use Dart-Sass here, so I need some way to simulate its functionality.

Dart-Sass generates mapping files by default, but I need to disable this feature.

Although these mapping files will not be packaged.

They appear on my system during local development, but I don't want to see them.

CodedOre: I didn't meant to say you need to switch to `grass` completely. Just that the buildsystem can use it when available, so that the script mapping is not needed here. 

If you say you want to keep this here, then at least move the `sass` script you have here into the `grass` module. 

Also, remove the shebang from the script. Flatpak Builder creates one for script sources automatically.

q962: I believe these are two modules, so I didn't write them together.

The parameters for grass and dart-sass are different.

q962: > Also, remove the shebang from the script. Flatpak Builder creates one for script sources automatically.

This is a subconscious behavior.

io.github.q962.PictureWindow.yml
hfiguiere: > My build tool strips the symbol files into a separate .sym file.

Then they should be fixed.

io.github.q962.PictureWindow.yml
```diff
@@ -0,0 +1,81 @@
...
+
+        xmake g --network=private
+
+        xmake f --mode=releasedbg
+        xmake b
+        xmake install -o /app
+
+        eu-unstrip /app/bin/PictureWindow /app/bin/PictureWindow.sym -o PictureWindow
+        mv PictureWindow /app/bin
+        rm /app/bin/*.sym
```
hfiguiere:
```suggestion
      - xmake g --network=private
      - xmake f --mode=releasedbg
      - xmake b
      - xmake install -o /app
      - eu-unstrip /app/bin/PictureWindow /app/bin/PictureWindow.sym -o PictureWindow
      - mv PictureWindow /app/bin
      - rm /app/bin/*.sym
```

io.github.q962.PictureWindow.yml
q962: > Then they should be fixed.

What do you mean by 'fix'? What exactly needs to be fixed? BuildTool?  

If flatpak supports separate debug files, then I wouldn't need to do this.

However, this approach is not currently supported.

The build tool I'm using has its own reasons for doing this.

I don't mind, because eu-unstrip can solve this problem.

hfiguiere: > What do you mean by 'fix'? What exactly needs to be fixed? BuildTool?

Not generating the .sym. This is NOT standard behaviour.

q962: What standard?

hfiguiere: the practice in use for 3 decades?

q962: I don't understand what you want me to do.

During the build process, the buildtool stores symbol information in the `.sym` file.

This results in my app lacking symbol information. `flatpak` cannot create the corresponding `.debug` file, instead creating a strange file `.sym.debug` ( a symbol file for symbol files :)   ).

`eu-unstrip` can fix this problem. It allows the `.sym` file to be incorporated into the app, so `flatpak` can generate the corresponding `.debug` file for my application.

If this isn't possible, what's the point of `eu-unstrip`?

bbhtt:
bot, build

PR #8175, opened 2026-03-22
Add it.andreafontana.hideout

trikko:

- [X] Please describe the application briefly. Hideout is a minimal and secure desktop application for file encryption and decryption, powered by [GnuPG](https://gnupg.org/).
- [X] Please attach a video showcasing the application on Linux using the Flatpak.  https://github.com/user-attachments/assets/557c2173-2987-41ee-a486-ad85268f7a78
- [X] I am an author contributor to the project.
    

### DIFF
--- a/it.andreafontana.hideout.yml
+++ b/it.andreafontana.hideout.yml
@@ -23,7 +23,7 @@ modules:
     buildsystem: simple
     build-commands:
       - rm -f dub.selections.json
-      - find deps/gid -name "dub.sdl" -o -name "dub.json" | while read f; do dub add-local --system "$(dirname "$f")" "0.9.9"; done
+      - GID_VERSION=$(grep "gid:adw1" dub.sdl | sed -n 's/.*version="[^0-9.]*\([0-9.]*\)".*/\1/p') && dub add-local --system "deps/gid" "$GID_VERSION"
       - dub build --build=release --d-version=Flatpak --skip-registry=all
       - install -D hideout /app/bin/hideout
       - install -D hideout_icon.svg /app/share/icons/hicolor/scalable/apps/it.andreafontana.hideout.svg
@@ -32,8 +32,8 @@ modules:
       - install -D LICENSE /app/share/licenses/it.andreafontana.hideout/LICENSE
     sources:
       - type: archive
-        url: https://github.com/trikko/hideout/archive/refs/tags/v0.9.10.tar.gz
-        sha256: d4d8f35890e99aeeb12c969e71749f870e729a12ef6f16946df565bbe39c4381
+        url: https://github.com/trikko/hideout/archive/refs/tags/v0.9.11.tar.gz
+        sha256: b075302a8900ecaf530f25ad8d506b93e245346acd9ca7cf2c83dc61cce8137f
       - type: archive
         url: https://github.com/Kymorphia/gid/archive/refs/tags/v0.9.9.tar.gz
         sha256: 5b177d8e02bba4096d5c24fc38bb3b39b99f23c8ce04736a3b93e2927eeb835e
--- a/it.andreafontana.hideout.yml
+++ b/it.andreafontana.hideout.yml
@@ -1,7 +1,7 @@
 app-id: it.andreafontana.hideout
 runtime: org.gnome.Platform
 runtime-version: '50'
-sdk: org.freedesktop.Sdk//25.08
+sdk: org.gnome.Sdk
 command: hideout
 finish-args:
   - --share=ipc
@@ -32,8 +32,8 @@ modules:
       - install -D LICENSE /app/share/licenses/it.andreafontana.hideout/LICENSE
     sources:
       - type: archive
-        url: https://github.com/trikko/hideout/archive/refs/tags/v0.9.9.tar.gz
-        sha256: a21e6eba8d73942f4d3722fcdbff4f4e22f5b62b11e064d4d798b5f7a2030e55
+        url: https://github.com/trikko/hideout/archive/refs/tags/v0.9.10.tar.gz
+        sha256: d4d8f35890e99aeeb12c969e71749f870e729a12ef6f16946df565bbe39c4381
       - type: archive
         url: https://github.com/Kymorphia/gid/archive/refs/tags/v0.9.9.tar.gz
         sha256: 5b177d8e02bba4096d5c24fc38bb3b39b99f23c8ce04736a3b93e2927eeb835e
--- a/it.andreafontana.hideout.yml
+++ b/it.andreafontana.hideout.yml
@@ -10,8 +10,6 @@ finish-args:
   - --device=dri
   - --filesystem=host
   - --filesystem=/tmp
-  - --talk-name=org.freedesktop.secrets
-  - --talk-name=org.gnupg.gpg-agent
 
 sdk-extensions:
   - org.freedesktop.Sdk.Extension.ldc
--- a/it.andreafontana.hideout.yml
+++ b/it.andreafontana.hideout.yml
@@ -1,0 +1,52 @@
+app-id: it.andreafontana.hideout
+runtime: org.gnome.Platform
+runtime-version: '50'
+sdk: org.gnome.Sdk
+command: hideout
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --filesystem=host
+  - --filesystem=/tmp
+  - --talk-name=org.freedesktop.secrets
+  - --talk-name=org.gnupg.gpg-agent
+
+modules:
+  - name: ldc
+    buildsystem: simple
+    sources:
+      - type: archive
+        url: https://github.com/ldc-developers/ldc/releases/download/v1.42.0/ldc2-1.42.0-linux-x86_64.tar.xz
+        sha256: a7bc9c956138f558cadf9c962352f59d41c80df6eb3ae3f8039f25be14a69303
+        only-arches: [x86_64]
+        strip-components: 1
+      - type: archive
+        url: https://github.com/ldc-developers/ldc/releases/download/v1.42.0/ldc2-1.42.0-linux-aarch64.tar.xz
+        sha256: 687707c3e20ff910528eb2d92f27a98cb0960284de3b026e6bf20284ac1c8511
+        only-arches: [aarch64]
+        strip-components: 1
+    build-commands:
+      - cp -r * /app/
+
+  - name: hideout
+    buildsystem: simple
+    build-commands:
+      - rm -f dub.selections.json
+      - find deps/gid -name "dub.sdl" -o -name "dub.json" | while read f; do /app/bin/dub add-local --system "$(dirname "$f")" "0.9.9"; done
+      - /app/bin/dub build --compiler=/app/bin/ldc2 --build=release --d-version=Flatpak --skip-registry=all
+      - install -D hideout /app/bin/hideout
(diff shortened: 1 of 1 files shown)


it.andreafontana.hideout.yml
```diff
@@ -0,0 +1,52 @@
...
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --filesystem=host
+  - --filesystem=/tmp
+  - --talk-name=org.freedesktop.secrets
+  - --talk-name=org.gnupg.gpg-agent
+
+modules:
+  - name: ldc
```
hfiguiere: there is a ldc SDK. Use that.

trikko:
Thanks @hfiguiere! I wasn't aware of that. Sorry, it's my first time submitting to Flathub.

So I should change the SDK like this?

trikko:
1. Domain/URL issue
My domain provider doesn't let me host files directly on the bare domain andreafontana.it, it automatically redirects everything to www.andreafontana.it. I’ve set up the redirect for the .well-known path as well, so it should be reachable that way. Is there a way to make the linter accept the redirect?

2. Filesystem access
The app is an encryption tool. Just like a file archiver (7zip/Zip), users expect to encrypt/decrypt files exactly where they are. That's why I added host and host-tmp access. The app itself doesn't touch anything in tmp or anywhere else on the system unless it's explicitly processing a file for encryption/decryption. It won't interfere with other apps or system files.

trikko:
Hey @hfiguiere, I've updated the manifest to use the ldc and llvm20 SDK extensions as requested.

Could you also give me some feedback on the two points I mentioned earlier? Any guidance on how to proceed would be greatly appreciated!

it.andreafontana.hideout.yml
```diff
@@ -0,0 +1,44 @@
+app-id: it.andreafontana.hideout
+runtime: org.gnome.Platform
+runtime-version: '50'
+sdk: org.freedesktop.Sdk//25.08
```
bbhtt:
```suggestion
sdk: org.gnome.Sdk
```

it.andreafontana.hideout.yml
```diff
@@ -0,0 +1,44 @@
...
+command: hideout
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --filesystem=host
+  - --filesystem=/tmp
```
bbhtt: why is this needed?

it.andreafontana.hideout.yml
```diff
@@ -0,0 +1,44 @@
...
+  - --talk-name=org.freedesktop.secrets
```
bbhtt: this should use the portal

it.andreafontana.hideout.yml
```diff
@@ -0,0 +1,44 @@
...
+  - --talk-name=org.gnupg.gpg-agent
```
bbhtt: what is this and where does the implementation exist for this bus name?

bbhtt:
> My domain provider doesn't let me host files directly on the bare domain andreafontana.it, it automatically redirects everything to [www.andreafontana.it](http://www.andreafontana.it/). I’ve set up the redirect for the .well-known path as well, so it should be reachable that way. Is there a way to make the linter accept the redirect?

No that's your problem to solve. If it doesn't work use a different appid

it.andreafontana.hideout.yml
trikko: When a user selects a file to encrypt/decrypt, the expected behavior is that the resulting file is created in the same directory as the source if not set. Without directory-level access (which Portals don't provide after a single file selection), the app can't write the output file to the same path automatically.
 
Forcing the user to go through an extra "Save As" portal every single time (to manually pick the same folder they just opened) adds significant friction and a redundant step to the process. For a tool meant for frequent file processing, this "double-dialogue" UX is quite poor.

Is there a way to keep this workflow while satisfying the sandbox requirements, or is the "Save As" portal the only path considered acceptable?

it.andreafontana.hideout.yml
trikko: I switched from org.gnome.Sdk to org.freedesktop.Sdk//25.08 because I was having trouble getting the required extensions (required by hfiguiere in a previous review) to work with the GNOME SDK. Since I'm not a Flatpak expert, I might be missing something on how to correctly make these extensions work with a GNOME runtime. Any guidance here would be appreciated.

trikko:
> No that's your problem to solve. If it doesn't work use a different appid

I understand the policy, though a bit more guidance on how the linter handles redirects would have been helpful. That said, I'm already in contact with my domain provider to see if I can resolve the redirect issue on my end.

it.andreafontana.hideout.yml
trikko: Sorry, the --talk-name stuff are leftover from some previous tests. Removed.

it.andreafontana.hideout.yml
trikko: https://github.com/flathub/flathub/pull/8175#discussion_r2979587537

trikko:
bot, build

> > No that's your problem to solve. If it doesn't work use a different appid
> 
> I understand the policy, though a bit more guidance on how the linter handles redirects would have been helpful. That said, I'm already in contact with my domain provider to see if I can resolve the redirect issue on my end.

Ok, moved away from their hosting for naked domain so their auto-redirect doesn't apply.

trikko:
bot, build

trikko:
Would the app still be rejected if I limited permissions to `--filesystem=home`? 

This could be a viable compromise: if the source file is within the user's home directory (or a subdirectory), I can automatically pre-fill the output path (e.g., filename + .gpg). If the file is outside the home directory, I would clear the destination field and force a manual path selection via the file picker.

While this isn't an ideal user experience, it seems like a necessary trade-off for sandboxing if I can't use `--filesystem=host`. I will make sure to clarify to my users that this specific limitation is a requirement for the Flathub ecosystem.

it.andreafontana.hideout.yml
bbhtt: i meant `org.freedesktop.secrets`

it.andreafontana.hideout.yml
bbhtt: why do you need `/tmp`?

bbhtt:
>  though a bit more guidance on how the linter handles redirects would have been helpful. 

i don't know what there is to clarify about linter's URL check, it does not allow redirects as you can see.

it.andreafontana.hideout.yml
bbhtt: you have to manually install the correct branch, gnome 50 is based on fdsdk 25.08 so the extension branch to install is 25.08

and you cannot use gnome-platform and a different sdk

it.andreafontana.hideout.yml
trikko: Yes don't worry leftovers I've already removed, as said

bbhtt:
you should reduce to home and ask for an [exception](https://docs.flathub.org/docs/for-app-authors/linter#exceptions) https://docs.flathub.org/docs/for-app-authors/linter#finish-args-home-filesystem-access

trikko:
> > though a bit more guidance on how the linter handles redirects would have been helpful.
> 
> i don't know what there is to clarify about linter's URL check, it does not allow redirects as you can see.

I fixed. Moved hosting for naked domain on another provider. Now it works.

it.andreafontana.hideout.yml
trikko: Same reason. BTW let's move to /home permission, limiting UX if that sounds better for you.

trikko:
> you should reduce to home and ask for an [exception](https://docs.flathub.org/docs/for-app-authors/linter#exceptions) https://docs.flathub.org/docs/for-app-authors/linter#finish-args-home-filesystem-access

Thanks, done.

trikko:
bot, build

it.andreafontana.hideout.yml
trikko: Fixed. And yet, I was pretty sure I had tried that and it didn't work.

bbhtt:
bot, build

trikko:
bot, build

trikko:
All done! 🎉

it.andreafontana.hideout.yml
```diff
@@ -0,0 +1,41 @@
...
+      - GID_VERSION=$(grep "gid:adw1" dub.sdl | sed -n 's/.*version="[^0-9.]*\([0-9.]*\)".*/\1/p') && dub add-local --system "deps/gid" "$GID_VERSION"
+      - dub build --build=release --d-version=Flatpak --skip-registry=all
+      - install -D hideout /app/bin/hideout
+      - install -D hideout_icon.svg /app/share/icons/hicolor/scalable/apps/it.andreafontana.hideout.svg
+      - install -D it.andreafontana.hideout.desktop /app/share/applications/it.andreafontana.hideout.desktop
+      - install -D data/it.andreafontana.hideout.metainfo.xml /app/share/metainfo/it.andreafontana.hideout.metainfo.xml
+      - install -D LICENSE /app/share/licenses/it.andreafontana.hideout/LICENSE
+    sources:
+      - type: archive
+        url: https://github.com/trikko/hideout/archive/refs/tags/v0.9.11.tar.gz
```
bbhtt: show button should open the filemanager instead of the filechooser and when decrypting i am not able to change the output file name

trikko: 1. Could you please clarify which "show" button you are referring to? If it's the one in the final screen, it is intended to open the file manager (Gtk.FileLauncher.openContainingFolder()) and highlight the generated file. Is it opening a file chooser instead for you? 

2. Regarding the output filename: in the Flatpak version, the direct path entry is restricted due to sandbox limitations. If a user manually types a path outside of the allowed directories (/home/user) the operation would save the file inside the sandbox, making it "invisible" to the user. To avoid this, I've designed the workflow to require a file picker selection, which ensures the app has the necessary permissions to write the file in the chosen location. I'm open to suggestions if there's a more 'Flatpak-native' way to handle this. (this friction and the need to avoid such UX compromises is precisely why I originally requested broader filesystem permissions)

bbhtt: then you should make that field greyed out.


> Is it opening a file chooser instead for you?

yes

trikko: Ooooops, good catch! It should be grayed out indeed. Easy fix.

Regarding the file chooser: thank you for the confirmation. That's very unexpected for an openContainingFolder call. To help me investigate this behavior, could you tell me which Linux distribution and desktop environment you are using to test?

trikko: Now that I've applied the change, I'm reconsidering if graying out the field is the best approach for the user experience, compared to just making it non-editable. However, if having the field enabled but non-editable is a blocker for the app's approval, I'll keep the current modification to ensure compliance.

trikko: @bbhtt I'm unable to reproduce the bug you mentioned. On my end, it correctly opens the file manager instead of a file picker. Could you possibly provide a short screen recording showing the entire workflow? This would help me reproduce the issue step-by-step and identify why it is behaving this way on your system.

bbhtt: no it worked now somehow, bump the commit here for the other fix.

trikko:
bot, build

it.andreafontana.hideout.yml
trikko: To avoid making multiple small releases for every fix, I've temporarily updated the yml to download from the main branch directly. Once we've addressed all the feedback and the app is ready for final approval, I'll create a new stable release with the proper tag and update the yml to point to it one last time.

PR #8176, opened 2026-03-22
Add eu.grimripper.GrimRipper

tanuva:

  GrimRipper is a CD ripping application that can fetch metadata from the internet. It is a fork of asunder which has existed for a while. The original author was not interested in porting asunder to newer GTK versions, but accepted forks that do the same. See [the discussion in the asunder bugtracker](https://littlesvr.ca/bugs/show_bug.cgi?id=59). (Requires log in with `visitor`/`visitor` lately to fend off bots.)
  https://github.com/user-attachments/assets/404f7a4f-0b02-4ff5-8b40-5240a96e4f97
- [x] I am an author/developer of the project.

### DIFF
--- a/eu.grimripper.grimripper.yaml
+++ b/eu.grimripper.grimripper.yaml
@@ -54,6 +54,23 @@ modules:
         path: patches/cdparanoia.meson
         dest-filename: meson.build
 
+  - name: vorbis-tools
+    buildsystem: autotools
+    no-autogen: true
+    config-opts:
+      - --disable-ogg123
+      - --disable-oggdec
+      - --disable-ogginfo
+      - --disable-vcut
+      - --disable-vorbiscomment
+      - --disable-oggtest
+      - --disable-vorbistest
+      - --disable-curltest
+    sources:
+      - type: archive
+        url: https://downloads.xiph.org/releases/vorbis/vorbis-tools-1.4.3.tar.gz
+        sha256: a1fe3ddc6777bdcebf6b797e7edfe0437954b24756ffcc8c6b816b63e0460dde
+
   - name: grimripper
     buildsystem: meson
     builddir: true
--- a/eu.grimripper.grimripper.yaml
+++ b/eu.grimripper.grimripper.yaml
@@ -60,4 +60,4 @@ modules:
     sources:
       - type: git
         url: https://gitlab.gnome.org/tanuva/grimripper.git
-        commit: e448db9bc5d14bccf4af9c769f319d7da23e6b5f
+        commit: 00262ba6ce1f67530aa90bf410b1c7b2d90748dd
--- a/eu.grimripper.grimripper.yaml
+++ b/eu.grimripper.grimripper.yaml
@@ -75,6 +75,25 @@ modules:
         url: https://downloads.xiph.org/releases/vorbis/vorbis-tools-1.4.3.tar.gz
         sha256: a1fe3ddc6777bdcebf6b797e7edfe0437954b24756ffcc8c6b816b63e0460dde
 
+  - name: lame
+    buildsystem: autotools
+    no-autogen: true
+    config-opts:
+      - --disable-shared
+      - --disable-dependency-tracking
+      - --disable-gtktest
+      - --disable-decoder
+    sources:
+      - type: archive
+        url: https://downloads.sourceforge.net/project/lame/lame/3.100/lame-3.100.tar.gz
+        sha256: ddfe36cab873794038ae2c1210557ad34857a4b6bdc515785d1da9e175b1da1e
+      - type: patch
+        path: patches/lame-3.99.5-tinfo.patch
+        strip-components: 0
+      - type: shell
+        commands:
+          - autoconf
+
   - name: grimripper
     buildsystem: meson
     builddir: true
--- a/eu.grimripper.grimripper.yaml
+++ b/eu.grimripper.grimripper.yaml
@@ -1,7 +1,7 @@
 id: eu.grimripper.grimripper
 runtime: org.gnome.Platform
 sdk: org.gnome.Sdk
-runtime-version: "48"
+runtime-version: "50"
 command: grimripper
 finish-args:
   - --socket=wayland
--- a/eu.grimripper.grimripper.yaml
+++ b/eu.grimripper.grimripper.yaml
@@ -81,4 +81,4 @@ modules:
     sources:
       - type: git
         url: https://gitlab.gnome.org/tanuva/grimripper.git
-        commit: fe52b8852f791a57e8dc2821d63643d71b15bcbe
+        commit: f48970c8151c37ada021983a39c4f0fa31033787
--- a/eu.grimripper.grimripper.yaml
+++ b/eu.grimripper.grimripper.yaml
@@ -46,8 +46,6 @@ modules:
   - name: cdparanoia
     buildsystem: meson
     builddir: true
-    config-opts:
-      - --prefix=/app
     sources:
       - type: archive
         url: http://downloads.xiph.org/releases/cdparanoia/cdparanoia-III-10.2.src.tgz
@@ -59,8 +57,6 @@ modules:
   - name: grimripper
     buildsystem: meson
     builddir: true
-    config-opts:
-      - --prefix=/app
     sources:
       - type: git
         url: https://gitlab.gnome.org/tanuva/grimripper.git
--- a/eu.grimripper.grimripper.yaml
+++ b/eu.grimripper.grimripper.yaml
@@ -1,0 +1,71 @@
+id: eu.grimripper.grimripper
+runtime: org.gnome.Platform
+sdk: org.gnome.Sdk
+runtime-version: "48"
+command: grimripper
+finish-args:
+- --socket=wayland
+- --socket=fallback-x11
+
+# From Flatpak docs:
+# This is not necessarily required, but without it the X11 shared memory
+# extension will not work, which is very bad for X11 performance.
+- --share=ipc
+
+# Required for CD drive access
+# https://github.com/flatpak/flatpak/issues/12
+- --device=all
+
+# Required to load music metadata
+- --share=network
+
+# Required for input-output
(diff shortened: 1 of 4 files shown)


eu.grimripper.grimripper.yaml
```diff
@@ -0,0 +1,71 @@
+id: eu.grimripper.grimripper
+runtime: org.gnome.Platform
+sdk: org.gnome.Sdk
+runtime-version: "48"
```
hfiguiere: this version will be EOL any day now.

eu.grimripper.grimripper.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+# Required for CD drive access
+# https://github.com/flatpak/flatpak/issues/12
+- --device=all
+
+# Required to load music metadata
+- --share=network
+
+# Required for input-output
+- --env=TMPDIR=/var/tmp
```
hfiguiere: not sure why you need this.

tanuva: A leftover from the first manifest author that I never questioned, sorry. :)

tanuva:
I am a bit puzzled by the build failure:
> Failed to download sources: module libcddb: Can't find file at (null)

After clearing the working directories I can reproduce that locally with `flathub-build`, but not when I use `flatpak-builder` with
```sh
flatpak-builder --user --force-clean --install flatpak-build-devel path/to/eu.grimripper.grimripper.yaml
```
Running `flatpak run -vv --command=flathub-build ...` (verbose) doesn't yield additional information. It looks like the libcddb archive is being downloaded successfully. I can see it in `.flatpak-builder/downloads`. It has the correct sha256 sum, but for some reason it is not being accepted. Do you coincidentally have an idea about that?

eu.grimripper.grimripper.yaml
```diff
@@ -0,0 +1,67 @@
...
+    buildsystem: autotools
+    sources:
+      - type: archive
+        url: http://downloads.sourceforge.net/project/libcddb/libcddb/1.3.2/libcddb-1.3.2.tar.bz2
+        sha256: 35ce0ee1741ea38def304ddfe84a958901413aa829698357f0bee5bb8f0a223b
+      - type: shell
+        commands:
+          - cp -p /usr/share/automake-*/config.{sub,guess} .
+      - type: patch
+        path: ../../patches/libcddb-pointer-types.patch
```
hfiguiere: where do you think this patch reside????

eu.grimripper.grimripper.yaml
```diff
@@ -0,0 +1,67 @@
...
+        sha256: 005db45ef4ee017f5c32ec124f913a0546e77014266c6a1c50df902a55fe64df
+      - type: file
+        path: cdparanoia.meson
+        dest-filename: meson.build
+
+  - name: grimripper
+    buildsystem: meson
+    builddir: true
+    config-opts:
+      - --prefix=/app
```
hfiguiere:
```suggestion
```

eu.grimripper.grimripper.yaml
```diff
@@ -0,0 +1,67 @@
...
+    sources:
+      - type: git
+        url: https://gitlab.gnome.org/tanuva/grimripper.git
+        commit: e448db9bc5d1
```
hfiguiere: please use full hash

eu.grimripper.grimripper.yaml
hfiguiere: it's a case of "works on my machine".

tanuva: > where do you think this patch reside????

I'm sorry? This is the first time I do this, sorry for my ignorance. From the error message it wasn't exactly obvious that the builder is talking about the patch file.

> it's a case of "works on my machine".

No. I said I can reproduce it, but from the message I was unable to guess that the patch file is at fault. I interpreted that it had already failed when getting the libcddb archive while the error was talking about "(null)".

tanuva:
bot, build

patches/cdparanoia.meson
bbhtt, line 1: why is this here and where is it from? if upstream doesn't support meson just use the buildsystem they support?

eu.grimripper.grimripper.yaml
```diff
@@ -0,0 +1,63 @@
...
+      - type: file
+        path: patches/cdparanoia.meson
+        dest-filename: meson.build
+
+  - name: grimripper
+    buildsystem: meson
+    builddir: true
+    sources:
+      - type: git
+        url: https://gitlab.gnome.org/tanuva/grimripper.git
```
bbhtt, line 102: why is this itself a fork of another repo called GrimRipper?

patches/cdparanoia.meson
tanuva: cdparanoia has seen the last release in 2008. I would be glad if there was an upstream that supported things. :) The Fedora maintainers have 7 patches in their cdparanoia package by now (see `build.log` links in the [package build information](https://koji.fedoraproject.org/koji/buildinfo?buildID=2753713)).

Building cdparanoia with meson has proven stable for me so far while the same with autotools lead to obscure errors (ELF sections too large, missing symbols). So far I fail to explain what is going on there apart from meson clearly handling things better. If you insist, I would investigate autotools further and try adding relevant Fedora patches here.

eu.grimripper.grimripper.yaml
tanuva: As can be seen in the discussion I linked in the PR description, there have been multiple people over the years who started to modernize asunder. Instead of starting from scratch, I picked up one of the seemingly abandoned forks and polished it further to get it into a releaseable shape. I plan to continue working on it towards GTK4 and libadwaita.

bbhtt:
```
flatpak run eu.grimripper.grimripper
Warning: could not load config file: No such file or directory
Couldn't remove config file in wrong location...
error: command data missing
libcddb: error: command data missing
```

tanuva:
> ```
> flatpak run eu.grimripper.grimripper
> Warning: could not load config file: No such file or directory
> Couldn't remove config file in wrong location...
> error: command data missing
> libcddb: error: command data missing
> ```

Curious. The first two aren't beautiful but expected for the moment. The second two are unexpected. I assume the main application window did not open for you?

I tried to reproduce this on a fresh Fedora 43 installation, but GrimRipper started up fine. Did you maybe have asunder installed before and the configuration file still in place? I wonder whether something in the settings migration may have gone wrong. (Then again, that shouldn't happen in the sandboxed environment.)

What distro are you on?

bbhtt:
No it was fresh install on Fedora 43 inside gnome boxes.

tanuva:
I have a hard time seeing in the code how a libcddb error would prevent the main window from opening. (Or reproducing the problem...) I enabled logging (to stderr) by default now. If you run the new build on a clean machine that did not have GrimRipper settings before then it should be more verbose on startup.

bot, build

bbhtt:
```
 flatpak run eu.grimripper.grimripper
Loading configuration
Warning: could not load config file: No such file or directory
Couldn't remove config file in wrong location...
Starting GrimRipper 3.1.0
using completion file name: /home/test/.var/app/eu.grimripper.grimripper/cache/grimripper/grimripper_album_artist
Reading completion data
using completion file name: /home/test/.var/app/eu.grimripper.grimripper/cache/grimripper/grimripper_album_title
Reading completion data
using completion file name: /home/test/.var/app/eu.grimripper.grimripper/cache/grimripper/grimripper_album_genre
Reading completion data
starting track: 0, ending track: 0
read_disc: discid=00000001
error: command data missing
libcddb: error: command data missing
Found 0 CDDB matchesupdate_tracklist() disk '00007fd5' '(null)' '(null)' '(null)'
test@fedora:~/Pictures$ 
```

tanuva:
> ```
> starting track: 0, ending track: 0
> read_disc: discid=00000001
> ```

Looks like it finds an audio (or mixed mode) CD in the drive. That CD turns out to have no tracks in the TOC. Maybe that part is confusing things... I could try and add a blind bail-out for that - however: Since it doesn't seem to be the Fedora installation ISO, which CD does your VM have in the drive? Is that an ISO I can get from somewhere, too?

bbhtt:
There is no CD attached to the VM.

image

tanuva:
Very curious. I added a workaround to GrimRipper now that tries to detect the invalid disc (zero total length or only zero-length tracks). Very strange though that the kernel even allows for opening `/dev/cdrom` and reading the TOC, apparently.

bot, build

bbhtt:
Clicking preferences shows oggenc was not found in your path.

tanuva:
Good to see the workaround is effective.

Tbh I had accepted the `oggenc` thing for now, but indeed the default setup should roughly work. So I added libogg, libvorbis and vorbis-tools to the package now so that GrimRipper can produce ogg files.

During testing I noticed that the default output directory (`$HOME`) of course does not get a sandbox portal. That should ideally be solved through an onboarding dialog, but I'd rather do that after moving to GTK4. Thus I added a sandbox access to the user's home directory for now.

Would you object to including lame in the flatpak as well? I suspect that'll still be the most frequently used output format...

bot, build

tanuva:
Of course, the linter doesn't like home dir access... Maybe the default must change to xdg-music instead. :thinking:

bbhtt:
you can add any dependency needed, the output directory can be hardcoded to xdg-music or its subdirectory or such for now.

tanuva:
Alright. lame needed a bit of help but here's another iteration.

* output directory is xdg-music, resulting in reduced sandbox access
* `lame` and `oggenc` are included now, enabling the (I assume) most popular formats to work out of the box. (Other encoders can be picked up from the host system's PATH as far as I observed.)

bot, build

hfiguiere:
> So I added libogg, libvorbis and vorbis-tools to the package now so that GrimRipper can produce ogg files.

I don't understand why you need to build libogg and libvorbis since they are in the runtime.

tanuva:
> I don't understand why you need to build libogg and libvorbis since they are in the runtime.

Oops. As can be seen it didn't occur to me to check that first. Sorry!

bot, build

PR #8183, opened 2026-03-23
Add io.github.forkgram.tdesktop

ilya-fedin:

- [x] Please describe the application briefly. A Telegram Desktop fork with a few UI changes (such as square avatars) and productivity features.
- [x] Please attach a video showcasing the application on Linux using the Flatpak. https://github.com/flathub/flathub/pull/8183#issuecomment-4119039657

@23rd (the author)

(the repo is pretty much the copy of the https://github.com/flathub/org.telegram.desktop repo)

### DIFF
--- a/flathub.json
+++ b/flathub.json
@@ -1,0 +1,3 @@
+{
+    "automerge-flathubbot-prs": true
+}
--- a/io.github.forkgram.tdesktop.yml
+++ b/io.github.forkgram.tdesktop.yml
@@ -1,0 +1,376 @@
+id: io.github.forkgram.tdesktop
+runtime: org.gnome.Platform
+runtime-version: '50'
+sdk: org.gnome.Sdk
+command: Forkgram
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --device=all
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.gnome.Mutter.IdleMonitor
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=com.canonical.AppMenu.Registrar
+  - --talk-name=com.canonical.indicator.application
+  - --talk-name=org.ayatana.indicator.application
+  - --talk-name=org.sigxcpu.Feedback
+  - --unset-env=QT_PLUGIN_PATH
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.llvm22
+build-options:
+  append-path: /usr/lib/sdk/llvm22/bin
+  append-ld-library-path: /usr/lib/sdk/llvm22/lib
+  cflags: -g1
+  cxxflags: -g1
+  env:
+    CMAKE_BUILD_TYPE: None
+    CMAKE_PREFIX_PATH: /app
+modules:
+  - name: pkg-config
+    buildsystem: simple
+    build-commands:
+      - install -d $FLATPAK_DEST/bin
+      - |
+        cat <<EOF > $FLATPAK_DEST/bin/pkg-config
+        #!/bin/sh
+        unset PKG_CONFIG_ALLOW_SYSTEM_LIBS
+        exec /usr/bin/pkg-config "\$@"
+        EOF
+      - chmod +x $FLATPAK_DEST/bin/pkg-config
+    cleanup:
+      - '*'
+
+  - name: implib
+    buildsystem: simple
+    build-commands:
+      - |
+        install -d $FLATPAK_DEST/lib
+        mkdir build
+        cd build
+        implib() {
+          LIBFILE=$(basename $1)
+          LIBNAME=$(basename $1 .so)
+          ../implib-gen.py -q $1
+          gcc $CFLAGS -c -o $LIBFILE.tramp.o $LIBFILE.tramp.S
+          gcc $CFLAGS -c -o $LIBFILE.init.o $LIBFILE.init.c
+          ar rcs $FLATPAK_DEST/lib/$LIBNAME.a $LIBFILE.tramp.o $LIBFILE.init.o
+        }
+        implib /usr/lib/$FLATPAK_ARCH-linux-gnu/libgtk-3.so
+        implib /usr/lib/$FLATPAK_ARCH-linux-gnu/libgdk-3.so
+    sources:
+      - type: git
+        url: https://github.com/yugr/Implib.so.git
+        commit: ecf7bb51a92a0fb16834c5b698570ab25f9f1d21
+    cleanup:
+      - '*'
+
+  - name: jinja2
+    buildsystem: simple
+    build-commands:
+      - pip3 install --exists-action=i --no-index --find-links="file://${PWD}" --prefix=${FLATPAK_DEST}
+        flit-core --no-build-isolation
+      - pip3 install --exists-action=i --no-index --find-links="file://${PWD}" --prefix=${FLATPAK_DEST}
+        "Jinja2" --no-build-isolation
+    sources:
+      - type: file
+        url: https://files.pythonhosted.org/packages/69/59/b6fc2188dfc7ea4f936cd12b49d707f66a1cb7a1d2c16172963534db741b/flit_core-3.12.0.tar.gz
+        sha256: 18f63100d6f94385c6ed57a72073443e1a71a4acb4339491615d0f16d6ff01b2
+        x-checker-data:
+          type: pypi
+          name: flit-core
+      - type: file
+        url: https://files.pythonhosted.org/packages/7e/99/7690b6d4034fffd95959cbe0c02de8deb3098cc577c67bb6a24fe5d7caa7/markupsafe-3.0.3.tar.gz
+        sha256: 722695808f4b6457b320fdc131280796bdceb04ab50fe1795cd540799ebe1698
+        x-checker-data:
+          type: pypi
+          name: MarkupSafe
+      - type: file
+        url: https://files.pythonhosted.org/packages/df/bf/f7da0350254c0ed7c72f3e33cef02e048281fec7ecec5f032d4aac52226b/jinja2-3.1.6.tar.gz
+        sha256: 0137fb05990d35f1275a587e9aee6d56da821fc83491a0fb838183be43f66d6d
+        x-checker-data:
+          type: pypi
+          name: Jinja2
+    cleanup:
+      - '*'
+
+  - name: systemd
+    buildsystem: meson
+    builddir: true
+    config-opts:
+      - --buildtype=plain
+    make-args:
+      - systemd-detect-virt
+    no-make-install: true
+    post-install:
+      - install -Dt "$FLATPAK_DEST/bin" systemd-detect-virt
+    sources:
+      - type: git
+        url: https://github.com/systemd/systemd.git
+        commit: c0a5a2516d28601fb3afc1a77d7b42fcfe38fced
+        tag: v260.1
+        x-checker-data:
+          type: json
+          url: https://api.github.com/repos/systemd/systemd/releases/latest
+          tag-query: .tag_name
+          timestamp-query: .published_at
+          version-query: $tag | sub("^[vV]"; "")
+      - type: shell
+        commands:
+          - >-
+            sed -i "s/'link_with' : libshared/'link_with' : [libshared_static, libsystemd_static,
+            libbasic_static]/" meson.build
+
+  - name: boost
+    buildsystem: simple
+    build-commands:
+      - ./bootstrap.sh --prefix=${FLATPAK_DEST} --with-libraries=regex
+      - ./b2 variant=release link=static runtime-link=shared cxxflags="$CXXFLAGS"
+        linkflags="$LDFLAGS" -j $FLATPAK_BUILDER_N_JOBS install
+    sources:
+      - type: archive
+        url: https://archives.boost.io/release/1.90.0/source/boost_1_90_0.tar.bz2
+        sha256: 49551aff3b22cbc5c5a9ed3dbc92f0e23ea50a0f7325b0d198b705e8ee3fc305
+        x-checker-data:
+          type: anitya
+          project-id: 6845
+          stable-only: true
+          url-template: https://archives.boost.io/release/$version/source/boost_${major}_${minor}_${patch}.tar.bz2
+    cleanup:
+      - '*'
+
+  - name: ada
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+      - -DBUILD_TESTING=OFF
+      - -DADA_TOOLS=OFF
+    sources:
+      - type: git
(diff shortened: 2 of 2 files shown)


hfiguiere:
Likely falls under https://docs.flathub.org/docs/for-app-authors/requirements#duplicate-submissions

because of

> (the repo is pretty much the copy of the https://github.com/flathub/org.telegram.desktop repo)

ilya-fedin:
> because of

It's logical to reuse already existing manifest when the app has pretty much the same dependencies, no?

Also, another Telegram fork, which is also a copy of org.telegram.desktop's repo, already exists on flatpak: https://github.com/flathub/io.github.kukuruzka165.materialgram. How was it accepted if this makes it fail under duplicate submission?

hfiguiere:
The fact that a telegram fork is already on Flathub is both irrelevant (it's this package we review) and relevant as there is already too many of these.

Even more since they rebuild a large part of the stack for no reason.

ilya-fedin:
> Even more since they rebuild a large part of the stack for no reason.

I assume you mean Qt by "a large part of the stack". The app uses both webkitgtk and Qt. Since there's no runtime which includes both webkitgtk and Qt, either webkitgtk or Qt has to be rebuilt. And Qt is smaller than webkitgtk.

ilya-fedin:
Can we continue without a video? I don't have one and making it is troublesome as it means revealing my contacts.

bbhtt:
> Can we continue without a video? 

No

> How was it accepted if this makes it fail under duplicate submission?

That submission predates the policy and we don't retroactively apply new policies to existing submissions that predate.

> A Telegram Desktop fork with a few UI changes (such as square avatars) and productivity features.

Where is the full set of changes? Changes like square avatars does not constitute grounds for accepting or reviewing a new submission.

23rd:
Hello.

The main focus of both my Android and desktop forks has been on a powerful forwarding system. While the official version allows you to forward messages without the “sender's name,” my fork allows for fine-grained manipulation of already downloaded files: quick forwarding with your own caption, merging scattered media from different messages into a single album, rearranging media in an already sent album, and you can even “remember” media and replace any other media you’ve sent with it. All of this was my personal optimization so that people using my fork wouldn’t have to resend a file that was sent “to the wrong place.” All of this is achieved using the official Telegram API, which official clients do not utilize to its full potential.

And here's a simple list of other changes.

### Chat List Enhancements
- Last Seen Time in Dialog List — shows each user's last seen time in the chat list
- Improved Online Status Timestamps — "Just now" (<1 min), "X minutes ago" (<1 hour), full date/time (>1 hour)
- Raise Unmuted & Unread Dialogs Above Muted
- Hide "All Chats" Tab
- Disable Global Search
- Open Telegram URL from Clipboard — button to open tg:// link from clipboard in search panel

### Message & Media Features
- Calculate Duration of Selected Media
- Send OGG as Voice Message — with random waveform
- Copy Selected Text Depersonalized — sender names replaced with "Person 1", "Person 2", etc.

### Bot / Mini App Features
- Skip Share Box from App Bots
- Save Chosen Platform per Bot
- Refresh Button in Bot Panel

### Keyboard Shortcuts
- Bind Dialogs to Ctrl+6 through Ctrl+9
- Shortcut to Open Main Menu
- Simplified Media Viewer Zoom — +/-/Up/Down without Ctrl, mouse wheel reversed from upstream
- Middle/Right-Click on Hamburger Menu — left: menu, middle: Archive, right: Saved Messages

### Information Display
- Date in Forwarded Message Header
- Separated Admin Log Entries for Media Edits
- "Go to First Message" — jump to first message in channel/chat

### Other Features
- Auto-Replace "..." with Unicode Ellipsis
- Open Links with Custom URI Scheme — configurable protocol handler (e.g., mpv://, vlc://)
- Search Selected Text via Context Menu — configurable search engine (default: dgg.gg)
- Audio Fade Toggle — enable/disable gradual volume fade
- Auto-Submit Local Passcode — submits once enough characters entered
- Hide Archived Stories
- Forward and Delete Button — forwards then deletes in one action
- Auto-Copy Incoming Login Codes
- Removed Account Limit — unlimited accounts (upstream: 3)



There were probably other things done as well, but a lot has changed over the 9 years I’ve been maintaining this fork, and I can’t remember everything anymore.
If this is enough, I can try recording a video for you later.
If this isn’t enough, then thank you for your time, and let’s just close this request.

bbhtt:
the video is not optional it is part of the PR template.

have you thought of making a baseapp for the shared parts of all these telegram clients?

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+    "automerge-flathubbot-prs": true
```
bbhtt, line 2: this is only allowed if the app is going to be verified, and it needs an exception

io.github.forkgram.tdesktop.yml
```diff
@@ -0,0 +1,367 @@
+id: io.github.forkgram.tdesktop
+runtime: org.gnome.Platform
+runtime-version: '49'
```
bbhtt: 50 is latest

io.github.forkgram.tdesktop.yml
```diff
@@ -0,0 +1,367 @@
...
+sdk: org.gnome.Sdk
+command: Forkgram
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --device=all
+  - --talk-name=org.freedesktop.Notifications
```
bbhtt, line 13: why doesn't it use the portal?

io.github.forkgram.tdesktop.yml
```diff
@@ -0,0 +1,367 @@
...
+  - --talk-name=org.gnome.Mutter.IdleMonitor
```
bbhtt, line 14: what is this for?

io.github.forkgram.tdesktop.yml
ilya-fedin: Because no library provides support for portal v2 with sounds

io.github.forkgram.tdesktop.yml
ilya-fedin: Telegram has a feature of in-app pass code which gets activated by a user-configurable in-app timer. With the permission, it won't lock as long as the user does any input on the machine (as intended). Without it, it can only update user input timestamp when the input is for the application window and will lock even if the user is using some other app on the machine (which is inconvenient and annoying).

flathub.json
ilya-fedin: I guess there will be no problems with verification through github.io? It will be a big pity to not have this, as it will lead to update delays (which are every month) only due to the need to click merge button manually each time.

ilya-fedin:
> have you thought of making a baseapp for the shared parts of all these telegram clients?

I'm not familiar with the process and not sure whether that would be actually useful given that none of telegram-specific libraries have backward/forward compatibility guarantees, Telegram uses private Qt APIs (which means it has to be rebuilt with each patch Qt update) and e.g. the aforementioned materialgram is two feature releases behind upstream and forkgram.

io.github.forkgram.tdesktop.yml
ilya-fedin: It doesn't build with 50

hfiguiere: then we can wait until it does.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "desktop-app-patches"]
+	path = desktop-app-patches
+	url = https://github.com/desktop-app/patches.git
```
hfiguiere: there is also something about using a submodule not in the flathub organisation. This make it impossible to fix the build if needed.

But the linter will tell you

https://docs.flathub.org/docs/for-app-authors/linter#external-gitmodule-url-found

io.github.forkgram.tdesktop.yml
```diff
@@ -0,0 +1,367 @@
...
+        x-checker-data:
+          type: json
+          url: https://api.github.com/repos/protocolbuffers/protobuf/releases/latest
+          tag-query: .tag_name
+          timestamp-query: .published_at
+          version-query: $tag | sub("^[vV]"; "")
+    cleanup:
+      - '*'
+
+  - name: jpegli
```
hfiguiere: `libjxl` is in the runtime.

.gitmodules
ilya-fedin: > This make it impossible to fix the build if needed.

Not sure what you mean? Those patches are exactly for hot-patching Qt after severe regressions which happen frequently and get fixed only after months upstream (or never).

io.github.forkgram.tdesktop.yml
ilya-fedin: You could notice by the module name that it's jpegli, not libjxl, which isn't in the runtime.

io.github.forkgram.tdesktop.yml
```diff
@@ -0,0 +1,367 @@
...
+        commit: 332feb17d17311c748445f7ee75c4fb55cc38530
+        tag: v0.11.2
+        x-checker-data:
+          type: json
+          url: https://api.github.com/repos/libjxl/libjxl/releases/latest
+          tag-query: .tag_name
+          timestamp-query: .published_at
+          version-query: $tag | sub("^[vV]"; "")
+      - type: file
+        url: https://github.com/libjxl/libjxl/commit/ee3955b1553bcc10304d45b85dfef9afa9349d72.patch
```
hfiguiere: how are you gonna reconcile this patch with the auto update above?

io.github.forkgram.tdesktop.yml
```diff
@@ -0,0 +1,367 @@
...
+        sha256: 5f8e3aa4fec3812ba17bbbf387a8781785a802642c31b1485d6758959d43f334
+      - type: shell
+        commands:
+          - sed -i 's/offset + t/offset + i/' ee3955b1553bcc10304d45b85dfef9afa9349d72.patch
```
hfiguiere: why instead isn't this a local patch instead of obscure unmaintainable dark magic?

io.github.forkgram.tdesktop.yml
```diff
@@ -0,0 +1,367 @@
...
+    buildsystem: meson
+    builddir: true
+    config-opts:
+      - --buildtype=plain
+    make-args:
+      - systemd-detect-virt
+    no-make-install: true
+    post-install:
+      - install -d "$FLATPAK_DEST/bin"
+      - install -D systemd-detect-virt "$FLATPAK_DEST/bin"
```
hfiguiere:
```suggestion
      - install -Dm755 systemd-detect-virt -t "$FLATPAK_DEST/bin"
```

io.github.forkgram.tdesktop.yml
ilya-fedin: I will just remove this when flathub bot will open PR with update to the version with this commit

io.github.forkgram.tdesktop.yml
ilya-fedin: Because this was simpler to me... I didn't want to add patches to be removed when they release 0.12 anyway when I can add just three lines in the time being.

23rd:
> * Please attach a video showcasing the application on Linux using the Flatpak. < Please insert the video here >

<details>
<summary>Video</summary>

https://github.com/user-attachments/assets/f46db101-c598-4e9b-9db1-5a513a426f06

</details>

ilya-fedin:
Can we get exceptions for the checks?

bbhtt:
Only for the json automerge if it is going to be verified.

ilya-fedin:
> Only for the json automerge if it is going to be verified.

Why couldn't commits from upstream deps file be tracked?

bbhtt:
the other two tracks-commits one can be given if those are going to be manually updated and not by automated workflows. i don't remember the discussion but according to the exceptions file it was given under the same condition to telegram.

the external-gitmodule you have to change.

ilya-fedin:
> the external-gitmodule you have to change.

Could the patches be fetched another way to avoid their double maintenance between upstream and flathub?

ilya-fedin:
I attempted to do it the same way as tg_owt and tde2e fetched but not sure whether I changed x-checker-data for Qt right... Documentation on the checker is quite limited.

bbhtt:
You can create a tarball from that repo where the patches live instead of a submodule. Github autogenerates tarballs from every commit or tag or branch of a repo.

ilya-fedin:
> You can create a tarball from that repo where the patches live instead of a submodule.

Isn't it a worse option than submodule given that tarball is a binary file?

bbhtt:
No why? Just use it as a source archive and apply the patches with the patch command.

ilya-fedin:
> No why? Just use it as a source archive and apply the patches with the patch command.

Well I already attempted to do like that (with git though), it will need the same exception as for tg_owt/tde2e I guess and I'm not sure whether I changed x-checker-data for Qt the right way to synchronize Qt updates with the patches...

ilya-fedin:
Oh, I can launch it locally... And it indeed fails
```
jq: error: $major is not defined at <top-level>, line 1, column 38:
    "https://download.qt.io/archive/qt/\($major).\($minor)/\($version)/single/qt-everywhere-src-\($version).tar.xz"
                                         ^^^^^^
jq: error: $minor is not defined at <top-level>, line 1, column 48:
    "https://download.qt.io/archive/qt/\($major).\($minor)/\($version)/single/qt-everywhere-src-\($version).tar.xz"
                                                   ^^^^^^
jq: 2 compile errors
```
I wonder how can I get major and minor with json checker...

ilya-fedin:
bot, build

ilya-fedin:
What's next? Should I create a PR at https://github.com/flathub-infra/flatpak-builder-lint?

23rd:
Hello.
How is it going?

ilya-fedin:
> What's next? Should I create a PR at https://github.com/flathub-infra/flatpak-builder-lint?

@bbhtt @hfiguiere could you please answer? i'm not sure whether i should create a pr there or is there anyhting else left to do.

bbhtt:
I already said which exceptions will be granted and under what conditions above. What more is needed here?

ilya-fedin:
> What more is needed here?

It is the first time I push an app with exceptions from beginning and I don't quite understand how it should requested:
* Should I create a PR at https://github.com/flathub-infra/flatpak-builder-lint or that doesn't work for packages not having a repo yet and it's you folks should do something internally?
* How to verify the app not having a repo? IIRC it's pushing a button on flathub.org for already existing apps but how that will work before the package has a repo?

CodedOre:
> * Should I create a PR at https://github.com/flathub-infra/flatpak-builder-lint or that doesn't work for packages not having a repo yet and it's you folks should do something internally?

The linter does not require the application to already have an repository in the Flathub organization. But it does require the exceptions to be in its exception file (as far as permissible for them to be added).

ilya-fedin:
https://github.com/flathub-infra/flatpak-builder-lint/pull/982

ilya-fedin:
bot, build

ilya-fedin:
bot, build

PR #8191, opened 2026-03-24
Add dev.j7126.combat_tracker

j7126:

- [X] Please describe the application briefly. A simple combat tracker app for ttrpg games.
[Screencast_20260324_231144.webm](https://github.com/user-attachments/assets/8b0184c5-64f6-4c46-839c-b6baa70518d9)


.gitignore
CodedOre, line 1: This file does not need to be part of the submission.

dev.j7126.combat_tracker.yml
```diff
@@ -0,0 +1,53 @@
...
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.llvm20
+command: combat_tracker
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --filesystem=host
```
CodedOre: Why does this application need full unrestricted access to the entire host filesystem?

There seems to be absolutely no reason for this at all.

Use the portals.

dev.j7126.combat_tracker.yml
```diff
@@ -0,0 +1,53 @@
...
+  - --filesystem=xdg-config/kdeglobals:ro
+  - --filesystem=xdg-config/gtk-3.0:ro
+  - --filesystem=xdg-config/gtk-4.0:ro
```
CodedOre: I don't see any good reason for this permission.

flatpak-flutter.yml
```diff
@@ -0,0 +1,51 @@
...
+            BUNDLE_PATH: build/linux/arm64/release/bundle
+      append-path: /usr/lib/sdk/llvm20/bin:/run/build/combat_tracker/flutter/bin
+      prepend-ld-library-path: /usr/lib/sdk/llvm20/lib
+      env:
+        PUB_CACHE: /run/build/combat_tracker/.pub-cache
+    build-commands:
+      - flutter build linux --release --no-pub
+      - install -D $BUNDLE_PATH/combat_tracker /app/bin/combat_tracker
+      - cp -r $BUNDLE_PATH/lib /app/bin/lib
+      - cp -r $BUNDLE_PATH/data /app/bin/data
```
CodedOre: `/app/bin` is no place for libraries and data files. 
The Filesystem Hierarchy Standard exists for a reason.

.gitignore
hfiguiere: .gitignore are fine.

generated/modules/flutter-sdk-3.41.4.json
hfiguiere, line 1: why is there two versions of this?

dev.j7126.combat_tracker.yml
j7126: Originally I was having an issue, that because the application displays the path of the user's recent files, it was showing the internal `/run/user/....` instead of the host path. Initially I just set filesystem host to fix this display issue.
However I have just learned about `xattr::user.document-portal.host-path`. I'll implement that path display using that, and then remove the host filesystem permission.

flatpak-flutter.yml
j7126: That path was generated by https://github.com/TheAppgineer/flatpak-flutter
Most of the other apps in the list on the bottom of that readme, that are already published on flathub are also already using that path.
What would you suggest to use instead?

CodedOre: If it would fully follow the Linux Filesystem Hierarchy standard, the libraries would be in `/app/lib` (depending on the situation even a subfolder in it) and the contents of `data` in `/app/share/$APP_NAME`.\
Files in `/app/bin` are expected to be executable binaries by the system.

If this is done by flatpak-flutter, it should be investigated if they could resolve this. 

And other apps doing it wrong is not really an argument. Just because it slipped under the radar before doesn't mean it should be done.

j7126: After a bit further investigation, it looks like flutter may actually require those directories to be located relative to the executable exactly like that...

CodedOre: Even if Flutter follows the "throw everything into one folder" filesystem model used by Windows, `/app/bin` is the wrong folder for it. You can't execute an image file, but the standard expects everything in `/app/bin` should be executable.

j7126: would just `/app` be considered correct?

j7126: When I change it to be just `/app` it does not work of course because the binary is not in bin. The error: `Error: Command 'combat_tracker' not found`

What's the expected solution here?

j7126:
bot, build

flatpak-flutter.yml
CodedOre: > would just `/app` be considered correct?

It's better than `/app/bin`. But it's also not correct. 

To my best knowledge, correct would be storing shared libraries under `/app/lib`, architecture-dependent, but app-specific files under `/app/lib/$APP_NAME`, and architecture-independent application data under `/app/share/$APP_NAME`.

> When I change it to be just `/app` it does not work of course because the binary is not in bin. The error: `Error: Command 'combat_tracker' not found`
> 
> What's the expected solution here?

Well, because the binary itself is of course expected to be found in the folder for binaries, `/app/bin`. If its not in there, it can't be found. 

You can use a simple script that launches the binary from it's wrong path if really needed. But ideally the binary would understand that it should find its libraries and data files from the right paths.

j7126:
bot, build

flatpak-flutter.yml
j7126: It seems that there are very many things hard-coded to look for that data directory next to the binary - both in flutter and in external packages.
I did however successfully figure out how to change the location of the libraries.

Would the following be considered an acceptable solution:

- the data directory located at `/app/share/dev.j7126.combat_tracker/data`
- the main binary located in `/app/share/dev.j7126.combat_tracker/`
- app-specific libraries located at `/app/lib/dev.j7126.combat_tracker/`
- `/app/bin/combat_tracker` would be a shell script that launches the main binary

Are there any other tweaks I would need to apply to this plan?

j7126:
bot, build

PR #8192, opened 2026-03-24
Add org.freedesktop.Sdk.Extension.qt611

da2ce7:
Qt 6.11.0 SDK extension for org.freedesktop.Sdk 25.08, providing 39 Qt submodules built from source as a composable SDK extension.

Depends on org.freedesktop.Sdk.Extension.llvm22 for shader toolchain (qtshadertools) and other LLVM-dependent modules.

QtWebEngine is excluded due to extreme build size (Chromium).

Builds for x86_64 and aarch64.

      This is an SDK extension (not an application). It provides Qt 6.11.0
      libraries and development tools for building Qt-based Flatpak
      applications on the org.freedesktop.Sdk 25.08 runtime. 39 Qt submodules
      are built from source (cmake-ninja), with all source archives fetched
      from download.qt.io with SHA-512 verification. QtWebEngine is excluded.
- [X] Please attach a video showcasing the application on Linux using the
      Flatpak. N/A — This is an SDK extension, not a GUI application.
      It has no user-facing interface.
- [X] The Flatpak ID follows all the rules listed in the Application ID
      requirements. The ID `org.freedesktop.Sdk.Extension.qt611` follows the
      org.freedesktop.Sdk extension point naming convention, which is exempt
      from the reverse-DNS domain ownership requirement.
- [X] I have read and followed all the Submission requirements and the
      Submission guide and I agree to them.
- [X] I am a packager/maintainer of this SDK extension manifest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,59 @@
+# SDK Extension for Qt 6.11
+
+This extension contains the [Qt](https://www.qt.io/) 6.11 libraries and development tools, providing a comprehensive cross-platform application framework for building modern desktop and embedded applications.
+
+Included modules: QtBase, QtDeclarative (QML/Quick), QtWayland, QtMultimedia, QtSvg, QtImageFormats, QtTools, QtShaderTools, Qt5Compat, QtWebSockets, QtWebChannel, QtPositioning, QtLocation, QtConnectivity, QtSerialPort, QtSerialBus, QtSensors, QtSCXML, QtRemoteObjects, QtCharts, QtNetworkAuth, QtHttpServer, QtQuickTimeline, QtQuick3D, QtQuick3DPhysics, QtGrpc, QtSpeech, QtVirtualKeyboard, QtLottie, Qt3D, QtGraphs, and more.
+
+**QtWebEngine is NOT included** due to its extreme build size (Chromium). Use a separate extension or the system WebEngine if needed.
+
+## Usage
+
+### Build
+
+In order to build your app with Qt 6.11 provided by this extension you have to set the following variables in your app manifest:
+
+```json
+{
+  "sdk-extensions": ["org.freedesktop.Sdk.Extension.qt611"],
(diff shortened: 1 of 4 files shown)


hfiguiere:
There is no point of this. The KDE 6.11 SDK will be here soon enough.

Erick555:
> This is an SDK extension (not an application). It provides Qt 6.11.0
> libraries and development tools for building Qt-based Flatpak
> applications on the org.freedesktop.Sdk 25.08 runtime

Those libraries won't be in Platform though so how compiled apps are supposed to run? Chasing manually every dependency and coping into app?

da2ce7:
@Erick555 The purpose of a SDK isn't only to compile apps for a platform. But also to develop the apps themselves; for example, I use SDK's to supply my VSCodium flatpak environment with development tools.

The KDE 6.1 SDK dose not integrate with my development environment in a happy way. Since I am developing on the Free desktop Stack.

There is precedence for this: @bbhtt  thankfully maintains https://github.com/flathub/org.freedesktop.Sdk.Extension.rust-nightly/ that I use for development.

This is very useful for development even if it is not included in any platform itself.

hfiguiere:
Qt is not Rust. the Rust sdk is used to build Rust software on flathub.

da2ce7:
@hfiguiere  I need *both* rust and Qt;

please consider: https://github.com/KDAB/cxx-qt

Of course, rust gets out the way once it is compiled, so this is only for development; for production I can use the KDE runtime.

Erick555:
> The purpose of a SDK isn't only to compile apps for a platform

Not _only_ but not _at all_. You could use extension to develop app but not create working flatpak app out if it which would be rather disappointing and limited use case not matching what description says.

> This is very useful for development even if it is not included in any platform itself.

Rust isn't needed in Platform for rust apps to work, Qt is needed.

PR #8256, opened 2026-03-30
Add org.sorobanexam.anzan

solsticedhiver:

- [X] Please describe the application briefly. An app to practice anzan (mental calculation while visualizing a soroban or not) by adding/subtracting number flashing on screen or the soroban

https://github.com/user-attachments/assets/dc8a248d-d6be-41c9-acec-21b09a71baf2

Some frames were skipped by OBS Studio. I don't know why. So some numbers were not showed on video. But you got the idea.

The app can also be used to exercise on a real soroban, with higher number of digits. Demoing here with a virtual soroban. With or without text-to-speech enabled.

the same but better video and audio with older version of app, demoing more things
https://www.youtube.com/watch?v=d9wdYUsNvTY
https://www.youtube.com/watch?v=twLG65S61XM


yakushabb:
> **Telemetry**
> 
> If telemetry is allowed, a unique random ID is generated, and used along side each TTS request to our server sorobanexam.org. It is kept for all next runs of the app. This helps us to keep track of the number of users of our app.

Source: https://github.com/solsticedhiver/anzan?tab=readme-ov-file#telemetry

solsticedhiver:
If you could say what you don't like in a few words, that would help. I guess you don't like it

The option was first called telemetry because I was planning on to collect user behavior, but I gave up on that. I should have rename it as "telemetry" is not a well loved word in opensource ;-)

The plan is to include other TTS API for the inclusion on f-droid. WIP
Another idea would be to include a virtual soroban inside the app. But no work done on that.

org.sorobanexam.anzan.yml
```diff
@@ -0,0 +1,55 @@
...
+  - org.freedesktop.Sdk.Extension.llvm20
+command: anzan
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --socket=pulseaudio
+  - --share=network
+  - --talk-name=org.freedesktop.ScreenSaver
```
bbhtt, line 17: what do you need this for? if it is idle-inhibition there is a portal for it

org.sorobanexam.anzan.yml
```diff
@@ -0,0 +1,55 @@
...
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.llvm20
+command: anzan
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --socket=pulseaudio
+  - --share=network
```
bbhtt, line 16: what is network for?

org.sorobanexam.anzan.yml
```diff
@@ -0,0 +1,55 @@
...
+    build-commands:
+      - setup-flutter.sh
+      - flutter build linux --release --no-pub
+      - install -D $BUNDLE_PATH/anzan /app/bin/anzan
+      - cp -r $BUNDLE_PATH/lib /app/bin/lib
+      - cp -r $BUNDLE_PATH/data /app/bin/data
+      - mkdir -p /app/share/{metainfo,applications,icons/hicolor/256x256/apps}
+      - cp org.sorobanexam.anzan.png /app/share/icons/hicolor/256x256/apps
+      - cp org.sorobanexam.anzan.desktop /app/share/applications/
+      - cp org.sorobanexam.anzan.metainfo.xml /app/share/metainfo/
```
bbhtt: install -Dm0644, reduces it to 3 steps

org.sorobanexam.anzan.yml
```diff
@@ -0,0 +1,55 @@
...
+    sources:
+      - type: git
+        url: https://github.com/solsticedhiver/anzan.git
+        commit: dd0cafab4de500c725f454f49721f30bf3ab157c
+      - type: file
+        path: org.sorobanexam.anzan.png
+      - type: file
+        path: org.sorobanexam.anzan.metainfo.xml
+      - type: file
+        path: org.sorobanexam.anzan.desktop
```
bbhtt: these files are not here?

org.sorobanexam.anzan.yml
```diff
@@ -0,0 +1,55 @@
...
+      - generated/sources/pubspec.json
+    modules:
+      - generated/modules/flutter-sdk-3.41.6.json
```
bbhtt, line 52: neither are these two generated modules

bbhtt:
telemetry is not a blocker, just make sure have it in OARS metadata as such. it's in the metainfo docs.

org.sorobanexam.anzan.yml
solsticedhiver: If the TTS (text-to-speech) feature is enabled, it retrieves mp3 sounds of the numbers

org.sorobanexam.anzan.yml
solsticedhiver: Numbers are displayed flashed. if the screensaver blanks the screen in the middle of a series, that's bad. the flutter package showed an error about org.freedesktop.ScreenSaver because it couldn't access it so I used the doc to enable it.
Not sure, if using the portal will change anything, because the flutter package don't use it

org.sorobanexam.anzan.yml
solsticedhiver: Do I need to add it in this repo/PR? The doc let me think it was not needed

org.sorobanexam.anzan.yml
solsticedhiver: Do I need to add it in this repo/PR? The doc let me think it was not needed

solsticedhiver:
bot, build

org.sorobanexam.anzan.yml
CodedOre: These files should indeed not be in this PR.
However, if you use a file source, Flatpak Builder will expect the files to be present in the manifest folder, or in this case this PR. So it will fail as it can't find them. 

Have the files in your repository and install them from the source for the application itself.

org.sorobanexam.anzan.yml
CodedOre: These files would fall under what the requirements call "dependency manifests", so these should be in this PR.

org.sorobanexam.anzan.desktop
CodedOre, line 1: This file should come with the source code from the upstream repository.

org.sorobanexam.anzan.metainfo.xml
CodedOre, line 1: This file should come with the source code from the upstream repository.

org.sorobanexam.anzan.png
CodedOre, line 1: This file should come with the source code from the upstream repository.

org.sorobanexam.anzan.metainfo.xml
```diff
@@ -0,0 +1,65 @@
...
+  </description>
+
+  <launchable type="desktop-id">org.sorobanexam.anzan.desktop</launchable>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://cdn.githubraw.com/solsticedhiver/anzan/main/flatpak/2025-04-11T23-21.png</image>
+      <caption>The main screen while flashing a number on screen</caption>
+    </screenshot>
+    <screenshot>
+      <image>https://cdn.githubraw.com/solsticedhiver/anzan/main/flatpak/2025-04-15T23-47.png</image>
```
hfiguiere: URL should NEVER reference the tip of a branch. It's not immutable.

(same above)

solsticedhiver: Would something like that be better? https://github.com/solsticedhiver/anzan/blob/23478ed18917746deebe46650b504428ddf40b60/flatpak/2025-04-15T23-47.png

or https://raw.githubusercontent.com/solsticedhiver/anzan/23478ed18917746deebe46650b504428ddf40b60/flatpak/2025-04-11T23-21.png

solsticedhiver:
bot, build

solsticedhiver:
bot, cancel

bbhtt:
why did you mark a bunch of comments as resolved without the changes pushed here? Don't do it, it wastes my time..

solsticedhiver:
Then it's you that is supposed to mark them as resolved ? I have commented on them most of the time. And pushed change, if need be

bbhtt:
No you can mark them resolved but only when you have pushed the change here. None of the changes about removing metainfo desktop file from here and upstreaming them are solved.

solsticedhiver:
Then how am I supposed to do that? if the manifest wants dep in this repo, how can I remove them here ? The files are already on the source repo. I just don't know how to let the manifest use them

CodedOre:
> Then how am I supposed to do that? if the manifest wants dep in this repo, how can I remove them here ? The files are already on the source repo. I just don't know how to let the manifest use them

Just remove the file sources from the manifest? If they are in the source code repository and you added that as a got source, the build system can access the files from there. You don't need to manually reference them as a source in the manifest.

solsticedhiver:
bot, build

org.sorobanexam.anzan.yml
solsticedhiver: So ? I can mark that as resolved ?

org.sorobanexam.anzan.yml
```diff
@@ -0,0 +1,52 @@
...
+    build-commands:
+      # update commit version to latest HEAD
+      - |
+        commit=`git rev-parse --short HEAD`
+        sed -i "/.*commit/s/HEAD/$commit/" lib/config.dart
+      - setup-flutter.sh
+      - flutter build linux --release --no-pub
+      - install -D $BUNDLE_PATH/anzan /app/bin/anzan
+      - cp -r $BUNDLE_PATH/lib /app/bin/lib
+      - cp -r $BUNDLE_PATH/data /app/bin/data
```
CodedOre, line 42: `/app/bin` is only meant for binaries, not libraries and data files. 

Follow the Linux Filesystem Hierarchy standard. 

If this is automatically done by the generator you use, please file an issue with them, as they should fix this.

org.sorobanexam.anzan.yml
```diff
@@ -0,0 +1,52 @@
...
+      - install -Dm0644 flatpak/org.sorobanexam.anzan.png /app/share/icons/hicolor/256x256/apps/org.sorobanexam.anzan.png
+      - install -Dm0644 flatpak/org.sorobanexam.anzan.desktop /app/share/applications/org.sorobanexam.anzan.desktop
+      - install -Dm0644 flatpak/org.sorobanexam.anzan.metainfo.xml /app/share/metainfo/org.sorobanexam.anzan.metainfo.xml
```
CodedOre, line 45: None of these files are Flatpak specific, so the folder name is misleading.

org.sorobanexam.anzan.yml
```diff
@@ -0,0 +1,52 @@
...
+            BUNDLE_PATH: build/linux/arm64/release/bundle
+      append-path: /usr/lib/sdk/llvm20/bin:/var/lib/flutter/bin
+      prepend-ld-library-path: /usr/lib/sdk/llvm20/lib
+      env:
+        PUB_CACHE: /run/build/anzan/.pub-cache
+    build-commands:
+      # update commit version to latest HEAD
+      - |
+        commit=`git rev-parse --short HEAD`
+        sed -i "/.*commit/s/HEAD/$commit/" lib/config.dart
```
CodedOre, line 37: Flathub is not for nightly software. 

Use an proper release with an actual version number. 

Also, avoid using sed in the build commands.

org.sorobanexam.anzan.yml
```diff
@@ -0,0 +1,52 @@
...
+      - install -D $BUNDLE_PATH/anzan /app/bin/anzan
+      - cp -r $BUNDLE_PATH/lib /app/bin/lib
+      - cp -r $BUNDLE_PATH/data /app/bin/data
+      - install -Dm0644 flatpak/org.sorobanexam.anzan.png /app/share/icons/hicolor/256x256/apps/org.sorobanexam.anzan.png
+      - install -Dm0644 flatpak/org.sorobanexam.anzan.desktop /app/share/applications/org.sorobanexam.anzan.desktop
+      - install -Dm0644 flatpak/org.sorobanexam.anzan.metainfo.xml /app/share/metainfo/org.sorobanexam.anzan.metainfo.xml
+    sources:
+      - type: git
+        url: https://github.com/solsticedhiver/anzan.git
+        commit: 140443ee4a8e5dd73ae2fbecb2d731cf07c5853d
```
CodedOre, line 49: Add a tag additional to the commit.

org.sorobanexam.anzan.yml
solsticedhiver: `flutter build` creates a bundle directory with an executable and alongside it, a  lib and data library.
If you know where to put /data and /lib and make it work, do tell me.
If you are not happy with it, please fill a bug report at flutter repo.

org.sorobanexam.anzan.yml
solsticedhiver: I have made a metainfo info to create the flatpak. so it is specific to the flatpak thing. I am not free to organize my repo source directory like I want anymore ?

Why does it matter? It's not the source directory name that matter but the destination directory!

org.sorobanexam.anzan.yml
solsticedhiver: this is what other flutter app published on flathub are doing.

org.sorobanexam.anzan.yml
solsticedhiver: I am seeing other app published on flathub having no problem with even weirder name

org.sorobanexam.anzan.yml
CodedOre: The standard is quite clear:

- Binaries that are supposed to be executable by the user should be in `/$PREFIX/bin`
- Libraries shared multiple applications should be in `/$PREFIX/lib` 
- Libraries used only by a specific application should be in `/$PREFIX/lib/$APPNAME` 
- Architecture-independent data of an application should be in `/$PREFIX/share/$APPNAME`.

On the host system, `$PREFIX` is `/usr`. Flatpak extends the specification by providing an separate prefix for the application under `/app`.

So, the lib folder should be under `/app/lib`, the data folder under `/app/share/$APPNAME`. 

And just to note:
It's the responsibility of the developer to follow established standard, not of us reviewers. \
If your tooling does not follow standards, you're the one who should report or fix it.

CodedOre: > this is what other flutter app published on flathub are doing.

I said it before in another submission:

Just because it slipped under the radar before doesn't meant it isn't wrong.

org.sorobanexam.anzan.yml
solsticedhiver: This is just to be complete, and as information.

If you had tested the app, you would have seen that it is shown with a release tag, but only in the about dialog, there is the commit.

org.sorobanexam.anzan.yml
CodedOre: > I have made a metainfo info to create the flatpak. so it is specific to the flatpak thing.

The Appstream metainfo file predates Flatpak and is used generally to provide information about an application on Linux. 

Flathub just happens to enforce its presence, unlike other packaging formats.

CodedOre: > I am not free to organize my repo source directory like I want anymore ?
> 
> Why does it matter? It's not the source directory name that matter but the destination directory!

The data files for desktop integration should not be specific to one package format, which is why I pointed this out.

PR #8259, opened 2026-03-31
Add io.github.kyuyrii.zordeer

Kyuyrii:

- [X] Please describe the application briefly. < A Qt launcher for games that run via Wine/Proton, with support for the UMU launcher. >
- [X] Please attach a video showcasing the application on Linux using the Flatpak. < https://github.com/user-attachments/assets/23bfae07-4a53-46aa-81a5-b00b1a0250f5 >


Kyuyrii:
I'm opening this pull request because barthalion asked me to.

Kyuyrii:
bot, build

io.github.kyuyrii.zordeer.yaml
```diff
@@ -0,0 +1,170 @@
...
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio # notification sound
+  - --allow=devel # ubisoft connect and other launchers
+  - --allow=multiarch
+  - --allow=per-app-dev-shm  # pressure-vessel
+  - --device=all
+  - --system-talk-name=org.freedesktop.UDisks2
+  - --talk-name=org.kde.StatusNotifierWatcher # system tray
+  - --env=PATH=/app/bin:/usr/bin
```
CodedOre:
```suggestion
```

This should already be set by Flatpak.

io.github.kyuyrii.zordeer.yaml
```diff
@@ -0,0 +1,170 @@
...
+  - --share=network
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio # notification sound
+  - --allow=devel # ubisoft connect and other launchers
+  - --allow=multiarch
+  - --allow=per-app-dev-shm  # pressure-vessel
+  - --device=all
+  - --system-talk-name=org.freedesktop.UDisks2
```
CodedOre, line 20: Why the permission for UDisk?

io.github.kyuyrii.zordeer.yaml
```diff
@@ -0,0 +1,170 @@
...
+      - mkdir -p /app/i386/nvidia-vaapi-driver
+      - mkdir -p /app/x86_64/nvidia-vaapi-driver
+      - install -Dm644 ld.so.conf /app/etc/ld.so.conf
+    sources:
+      - type: inline
+        dest-filename: ld.so.conf
+        contents: |
+          /app/lib
+          /app/lib32
+          /app/lib64
```
hfiguiere:
```suggestion
```

there is no lib64

io.github.kyuyrii.zordeer.yaml
```diff
@@ -0,0 +1,170 @@
...
+        url: https://github.com/Kyuyrii/Zordeer.git
+        tag: '3.7'
+        commit: 508b6a17a24b5882b265503a4bcb99cd6e1455d2
+    build-commands:
+      - install -Dm755 zordeer /app/bin/zordeer
+      - install -Dm644 io.github.kyuyrii.zordeer.desktop /app/share/applications/io.github.kyuyrii.zordeer.desktop
+      - install -Dm644 io.github.kyuyrii.zordeer.svg /app/share/icons/hicolor/scalable/apps/io.github.kyuyrii.zordeer.svg
+      - install -Dm644 metainfo/io.github.kyuyrii.zordeer.metainfo.xml /app/share/metainfo/io.github.kyuyrii.zordeer.metainfo.xml
+      - install -Dm644 translation/*.toml -t /app/share/Zordeer/translation
+      - desktop-file-edit --set-name="Zordeer [Flatpak]" /app/share/applications/io.github.kyuyrii.zordeer.desktop
```
hfiguiere: why does this need to be changed?

modules/python3-pyqt6.json
hfiguiere, line 1: this is clearly not generated by the flatpak-pip-generator tool.

modules/python3-tomlkit.json
hfiguiere, line 1: nor is this one.

io.github.kyuyrii.zordeer.yaml
Kyuyrii: Errors were appearing that only disappeared after granting permission to UDisks2.

Zordeer also has a function to automatically mount .iso files, but it doesn't currently work with Flatpak.

io.github.kyuyrii.zordeer.yaml
Kyuyrii: To make it clear to the user which version they are using, avoiding confusion.

io.github.kyuyrii.zordeer.yaml
Kyuyrii: Correcting the comment, those weren't errors preventing execution, but I suppose they weren't supposed to appear.

Regarding mounting the .iso file, it's a function that mounts the .iso file and searches for the .exe file inside it so it can be executed.

Since I haven't yet made it work on Flatpak, I don't know if that's the only permission required.

modules/python3-pyqt6.json
Kyuyrii: I just tried using flatpak-pip-generator with PyQt6, and this appeared:
```
PyQt packages are not supported by flapak-pip-generator
However, there is a BaseApp for PyQt available, that you should use
Visit https://github.com/flathub/com.riverbankcomputing.PyQt.BaseApp for more information
```

io.github.kyuyrii.zordeer.yaml
CodedOre: Two points about that:

First, the name and id of the desktop file is still the same. So it will likely not work as you expect on every desktop. It will just use the file from the last defined directory. 
It may work on some desktops who implement special logic, but I'm quite certain it will not work for example on GNOME. 

Second, you seem to expect a user to install multiple versions of the same application. But that does not sound like a very likely scenario for most users. So, the "[Flatpak]" suffix wouldn't give a great benefit there.

Kyuyrii: Gnome Software and Plasma Discover can display different versions of the same app; I've seen Plasma Discover glitch and only show one version of an app without specifying which version it is.

And I've also seen users confused a few times about which version of an app they downloaded when using the app store that came with the distro.

modules/python3-tomlkit.json
Kyuyrii: I modified this file, but unfortunately I don't know what to do with the PyQt6 file.

io.github.kyuyrii.zordeer.yaml
Kyuyrii: I managed to get mounting .iso files working on Flatpak, but since I saw that UDisks2 doesn't necessarily only use /media, I added access to /run/media as well. I'm just not sure if /mnt is also used.

modules/python3-pyqt6.json
teamcons: ok

io.github.kyuyrii.zordeer.yaml
CodedOre: So, UDisk is used to mount iso images?

Do the iso images need to be mounted though? Or would it be possible to use an in-sandbox tool to read or extract the content of the iso instead?

CodedOre (commented):
A question: Does your application uses the wine from the Wine BaseApp, or does it allows to download other Wine versions in app?

If it is the latter, would it make sense to use the PyQt BaseApp instead?

Kyuyrii:
Zordeer allows using Wine from `$PATH`, using external Wine, using Proton without UMU, and using Proton with UMU.

That's why, unfortunately, I can't use BaseApp PyQt.

io.github.kyuyrii.zordeer.yaml
Kyuyrii: The problem with extracting a file instead of mounting it is that instead of simply accessing the containing file, the content will be duplicated.

Kyuyrii:
I was testing mounting .iso files, but unfortunately I can't mount an .iso file accessed via Document Portal.

When I try to mount an .iso file located in the Downloads folder, it mounts immediately because Zordeer has `--filesystem=xdg-download` in its manifest.

However, when I try to mount an .iso file located elsewhere in the $HOME folder using Document Portal, I get errors like this:
`org.freedesktop.UDisks2.Error.Failed: Error waiting for loop object after creating '/dev/loop47': Timed out waiting for object`

Kyuyrii:
I requested the review again because I added information about the Zordeer situation on Flatpak.

Honestly, I don't know if there's a specific portal for mounting .iso files that are accessed via Document Portal.

But I imagine this error makes the problem of using Document Portal clearer, since it's clearly not the fault of `UDisks2`.

Kyuyrii:
I've added access to some folders that may contain .exe files and Wine prefixes. Is it okay to have these permissions, or should I remove them?

To clarify why I added these permissions:

~/App
This is the previous Zordeer folder. Wine prefixes may exist within this folder. If someone has used another version of Zordeer, when migrating to the Flatpak version, it will not be necessary to move the prefixes, only the files containing the information.
The issue wouldn't just be moving the Wine prefixes, but also having to edit the configuration files.

~/Games
It is common to find Wine prefixes or games in this folder, especially if the user has used another launcher before or UMU directly.

~/GOG Games
Folder used for GOG games. Both Wine prefixes and installers may exist within it.

io.github.kyuyrii.zordeer.yaml
CodedOre: Having access to the system service that manages all drives on your computer is a big permission though. 

And considering it also does not work well with the paths from the sandbox (which isn't really that surprising), I would rather tend to not approve of it. 

I do understand that other methods have their own drawbacks. But access to core system services is not something granted on a whim.

io.github.kyuyrii.zordeer.yaml
```diff
@@ -0,0 +1,174 @@
...
+  - --talk-name=org.kde.StatusNotifierWatcher # system tray
+  - --env=QT_PLUGIN_PATH=/usr/share/runtime/lib/plugins:/usr/lib/plugins
+  - --env=GST_PLUGIN_SYSTEM_PATH=/app/lib/gstreamer-1.0:/app/lib32/gstreamer-1.0:/usr/lib/extensions/gstreamer-1.0:/usr/lib/x86_64-linux-gnu/codecs-extra/lib/gstreamer-1.0:/usr/lib/i386-linux-gnu/codecs-extra/lib/gstreamer-1.0:/usr/lib/x86_64-linux-gnu/gstreamer-1.0:/usr/lib/i386-linux-gnu/gstreamer-1.0
+  - --filesystem=xdg-download
+  - --filesystem=~/App
+  - --filesystem=~/Games
+  - --filesystem=~/GOG Games
+  - --filesystem=/media
+  - --filesystem=/mnt
+  - --filesystem=/run/media
```
CodedOre: You've heard it before, but you should really use the FileChooser portal instead. 

That there "might" be files users "might" want to use in the application is not that good of an argument.

Kyuyrii: The problem is that there's really no way to access it via the Document Portal. It will cause problems when using Wine/Proton/UMU.

Zordeer can even run installers accessed via the Document Portal using the "Run in prefix" function, but when it comes to Wine prefixes, unfortunately there's nothing I can do.

It's common for .exe files and prefixes to be in those folders.

It's like finding image files in the xdg-pictures folder.

For example, when running the UMU launcher without specifying a prefix, it creates the default prefix inside the $HOME/Games folder.

Kyuyrii: Regarding /media, /run/media, and /mnt, are they also considered a problem?

If Zordeer has so little access to the folders, it will be necessary to modify parts of Zordeer to inform the user that they are using the Document Portal and that they should copy the files into the Zordeer folder, or install Flatseal to allow access to the necessary folders.

Ultimately, this will allow access to the same folders, or even more, because the user will want Wine/Proton/UMU to function properly.

Kyuyrii: Or, the user will simply give up on using it, because they have to do so much just to run a simple .exe file correctly.

io.github.kyuyrii.zordeer.yaml
Kyuyrii: In this case, UDisks2 is only used if the user selects an .iso file instead of an .exe/.msi file and attempts to run it.

Using UDisks2 was a decision made only because it could correctly mount the .iso files and would already be available on the system.

Kyuyrii: The Flatpak package org.winehq.Wine also has access to UDisks2, and Zordeer uses org.winehq.Wine as a base.

So apparently this is a requirement for Wine to function correctly.

io.github.kyuyrii.zordeer.yaml
Kyuyrii: I'm basically asking for the bare minimum to deliver a decent experience.

Zordeer is a very simple app; it would be easy to find out if it's doing something it shouldn't.

It doesn't even access files from those locations automatically, and only transfers files into its own folder, specifically community-made versions of Wine/Proton (Proton-GE, Proton-Sarek, Kron4ek, etc.), the UMU launcher, and SteamGridDB images.

Kyuyrii:
I recorded a video demonstrating the difference between using the Document Portal and accessing files and folders directly.

This error is not from Zordeer, it's from UMU.

https://github.com/user-attachments/assets/c992a5cb-62a4-41d4-af77-e558d9ab395c

io.github.kyuyrii.zordeer.yaml
Kyuyrii: Actually, I'm considering an ideal situation.

If the user, for example, has a folder called "Jogos" instead of "Games", the `--filesystem=~/Games` will not be useful for that user.

From what I've seen, `--filesystem=xdg-games` doesn't exist.

CodedOre: I mean, if it really is not possible to use the Document portal and no clear standard folder exists, you might as well use `home` and [request a linter exception](https://docs.flathub.org/docs/for-app-authors/linter#exceptions).

But, just to note this because of previous experience:
The home permission should only be used at the direction of the user. 

You should still store your own application data in the sandboxed folder.

Kyuyrii: Zordeer Flatpak uses XDG_CONFIG_HOME and XDG_DATA_HOME by default.

I opened a pull request:

https://github.com/flathub-infra/flatpak-builder-lint/pull/971

io.github.kyuyrii.zordeer.yaml
Kyuyrii: I removed that part from the manifest and added that information to the about window.

Kyuyrii:
I think it should be good enough to test now. I'll ask the bot to build Zordeer again.

Kyuyrii:
bot, build

Kyuyrii:
For some reason, the i386 extensions are not being required, only the x86_64 ones.

However, this doesn't happen with org.winehq.Wine, so I imagine that's the expected behavior.

Kyuyrii:
bot, build

Kyuyrii:
bot, build

io.github.kyuyrii.zordeer.yaml
```diff
@@ -0,0 +1,178 @@
...
+
+inherit-extensions:
+  - org.winehq.Wine.gecko
+  - org.winehq.Wine.mono
+
+add-extensions:
+# The x86_64 extensions are listed because UMU was unable to find the .so files correctly.
```
barthalion: This has to be sorted out differently. These extensions are inherited from the runtime and defining these explicitly changes nothing in terms of visibility within the sandbox.

barthalion (changes requested):

io.github.kyuyrii.zordeer.yaml
Kyuyrii: The problem is that if I remove `org.freedesktop.Platform.VAAPI.Intel` from the manifest, this appears:

`x86_64-linux-gnu-capsule-capture-libs: warning: Dependencies of /usr/lib/x86_64-linux-gnu/dri/intel-vaapi-driver/iHD_drv_video.so not found, ignoring: Missing dependencies: Could not find "libigdgmm.so.12" in LD_LIBRARY_PATH "", ld.so.cache, DT_RUNPATH or fallback /lib/x86_64-linux-gnu:/usr/lib/x86_64-linux-gnu:/usr/x86_64-pc-linux-gnu/lib:/lib64:/usr/lib64:/lib:/usr/lib
`

Kyuyrii: I imagine the alternative would be to change the LD_LIBRARY_PATH, but I suppose that's considered a worse solution.

barthalion: Is this actually breaking anything or just displaying a warning?

Kyuyrii: In the games I'm testing, it doesn't seem to break anything, but there are many games with different needs, so ignoring this could be a problem later.

From what I understand, VAAPI is used in videos that are played in games.

barthalion: This isn't a generic VA-API error, it's strictly about the driver for Intel iGPU users. Unless you are one, you don't need that extension for anything.

Kyuyrii: I use an Intel iGPU, the UHD Graphics 610.

barthalion: And so for you it would be auto-installed. If test builds of this app are your only app, Flatpak won't auto-install extensions across remotes for security. And test builds are separate remotes, just to be clear.

Kyuyrii: The problem isn't with the installation; apparently, it's with the path where the files are located.

Even though I already have this extension installed, the warning only disappears when I add it to the manifest.

barthalion: Adding it to the manifest doesn't change the path it's exposed at.

Kyuyrii: So something very strange is happening, because adding this to the manifest is making the warning disappear. Could it be a Flatpak bug?

Kyuyrii:
I'll have to test something here, because the situation is a bit confusing.

Kyuyrii:
bot, build

io.github.kyuyrii.zordeer.yaml
Kyuyrii: I managed to solve the problem.

barthalion:
image

This doesn't seem correct on my en_IE system.

Kyuyrii:
> This doesn't seem correct on my en_IE system.

I changed the path where the translation files are placed, can you test again?

This problem doesn't occur here. In my case, I only run `flatpak run io.github.kyuyrii.zordeer`.

io.github.kyuyrii.zordeer.yaml
```diff
@@ -0,0 +1,83 @@
...
+      - type: git
+        url: https://github.com/Kyuyrii/Zordeer.git
+        tag: '4.0'
+        commit: dd15957830fb711319436d3d8a56a4c012e9add6
+    build-commands:
+      - install -Dm755 zordeer /app/bin/zordeer
+      - install -Dm644 io.github.kyuyrii.zordeer.desktop /app/share/applications/io.github.kyuyrii.zordeer.desktop
+      - install -Dm644 io.github.kyuyrii.zordeer.svg /app/share/icons/hicolor/scalable/apps/io.github.kyuyrii.zordeer.svg
+      - install -Dm644 metainfo/io.github.kyuyrii.zordeer.metainfo.xml /app/share/metainfo/io.github.kyuyrii.zordeer.metainfo.xml
+      - install -Dm644 translation/*.toml -t /app/bin/translation
```
CodedOre: `/app/bin` is strictly for executable binaries. It's not the correct place for translation files.

Kyuyrii:
I managed to reproduce the error here, I'll try to fix it and then release it in a new version.

io.github.kyuyrii.zordeer.yaml
Kyuyrii: I'll revert this along with a new version; I'm trying to understand what's causing the translation problem.

Kyuyrii: I reverted that change.

Kyuyrii:
Apparently I managed to fix the problem with the translations; there was a missing fallback when Zordeer didn't have a file for the system's language.

I added MangoHud to the manifest; this link I created is so that the mangohud command can be found without needing to change the PATH.

Kyuyrii:
bot, build

Kyuyrii:
Thanks for testing Zordeer, I really didn't know that problem you showed existed.

Zordeer should be working correctly now.

barthalion:
bot, build

PR #610, opened 2018-09-07
Add org.freedesktop.Piper

whot:
A disclaimer here: Piper needs ratbagd which is a system-installation and not easily flatpaked, on its own it's a bit pointless since it won't have access to devices.

PR #8278, opened 2026-04-03
Add  io.github.niltonperimneto.Twister

niltonperimneto:

- [X] Please describe the application briefly. <This is a GUI for configuring gaming mice that uses libratbag as the backend just like Piper, but using Tauri+Svelte instead of GTK3 for a more modern and desktop-agnostic look and feel>
- [X] Please attach a video showcasing the application on Linux using the Flatpak. <https://github.com/user-attachments/assets/a1910973-5a60-4e75-b741-e1e43971e6ac>
- [X] I am an author contributor to the project.


io.github.niltonperimneto.Twister.metainfo.xml
hfiguiere, line 1: This file should be (submitted and managed) upstream.

io.github.niltonperimneto.Twister.metainfo.xml
```diff
@@ -0,0 +1,96 @@
...
+
+  <branding>
+    <color type="primary" scheme_preference="light">#e0e0e0</color>
+    <color type="primary" scheme_preference="dark">#1a1a2e</color>
+  </branding>
+
+  <screenshots>
+    <screenshot type="default">
+      <caption>Device overview with DPI editor</caption>
+      <image>https://raw.githubusercontent.com/niltonperimneto/twister/master/public/screenshot_2.png</image>
```
hfiguiere: URL should NEVER reference the tip of a branch. It's not immutable.

same below

io.github.niltonperimneto.Twister.metainfo.xml
hfiguiere: This file is never used or referenced

niltonperimneto: I think I fixed it. I moved the metainfo file upstream

io.github.niltonperimneto.Twister.metainfo.xml
niltonperimneto: I replaced screenshot URLs with immutable links

io.github.niltonperimneto.Twister.yml
```diff
@@ -0,0 +1,54 @@
...
+  - org.freedesktop.Sdk.Extension.node20
+command: twister
+
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --device=dri
+  - --socket=fallback-x11
+  - --system-talk-name=org.freedesktop.ratbag1
```
CodedOre, line 16: How does the application deal with the situation when ratbag is not installed?

io.github.niltonperimneto.Twister.yml
```diff
@@ -0,0 +1,54 @@
...
+    build-options:
+      append-path: /usr/lib/sdk/rust-stable/bin:/usr/lib/sdk/node20/bin
+      env:
+        CARGO_HOME: /run/build/twister/cargo
+        npm_config_cache: /run/build/twister/flatpak-node/npm-cache
+        CARGO_NET_OFFLINE: 'true'
+    build-commands:
+      - npm ci --offline
+      - npm run build
+      - sed -i 's|tauri = { version = "2.10", features = \[\] }|tauri = { version = "2.10", features = ["custom-protocol"] }|' src-tauri/Cargo.toml
```
CodedOre: Why is this needed?

io.github.niltonperimneto.Twister.yml
```diff
@@ -0,0 +1,54 @@
...
+      - cargo build --release --manifest-path src-tauri/Cargo.toml
+      - sed -i 's|https://raw.githubusercontent.com/niltonperimneto/twister/master/public/|https://raw.githubusercontent.com/niltonperimneto/twister/3736152b4bb9d34012e278e9f0d5af6a35716e10/public/|g' io.github.niltonperimneto.Twister.metainfo.xml
```
CodedOre: Why can't the upstream metainfo file not use permalinks?

You will need to update the metainfo anyway with an update, to add the new release notes.

io.github.niltonperimneto.Twister.yml
niltonperimneto: It is similar to how Piper, which is already on Flathub, does; however, I do provide information on what to do if it is installed but not enabled:

Screenshot From 2026-04-03 16-44-02

If you want me to provide further information, such as code snippets from the dbus code and the Tauri frontend I'll happily provide it

io.github.niltonperimneto.Twister.yml
niltonperimneto: It is there because my upstream repo builds Tauri with features = [], so the Flatpak may try devUrl (localhost) at runtime.
This patch enables custom-protocol, forcing the app to load bundled frontend files instead of a dev server, which prevents Connection refused in production. The other way that I can handle this is to include the custom-protocol on the src-tauri toml. I can do whichever way you guys think is better

CodedOre: The thing with `sed` is that it is not failsafe AFAIK.

If you don't want to permanently enable this Tauri feature, you could add a feature to your crate that would enable this feature and then set this when building the app with cargo as a feature flag.

niltonperimneto: > The thing with `sed` is that it is not failsafe AFAIK.
> 
> If you don't want to permanently enable this Tauri feature, you could add a feature to your crate that would enable this feature and then set this when building the app with cargo as a feature flag.

Done

io.github.niltonperimneto.Twister.yml
bbhtt: https://docs.flathub.org/docs/for-app-authors/requirements#host-dependent-applications

niltonperimneto: > https://docs.flathub.org/docs/for-app-authors/requirements#host-dependent-applications

How about Piper, which is already on Flathub and relies on the same host dependency, which, in fact, is made by them? What is fundamentally different between both frontends?

Link: https://github.com/flathub/flathub/pull/610

teamcons: That was eight years ago. Standards change. Old stuff dont get re reviewed.

niltonperimneto: Fair enough. Just a bit weird that my issue was in part being too late

niltonperimneto: So thare is no path that I could have to get my app on Flathub unless some portal is introduced to allow ratbagd on flatpak?

teamcons: I dont know a think about ratbagd, but wouldnt it work including it in the build, and add permission to access usb peripherals?

The expectation is to have flatpaks be self contained

niltonperimneto: > I dont know a think about ratbagd, but wouldnt it work including it in the build, and add permission to access usb peripherals?
> 
> The expectation is to have flatpaks be self contained

Unfortunately, I can't imagine a way, and it is because ratbagd needs to be a system-level daemon because it requires persistent udev rules for device detection, system D-Bus activation so it starts on demand when any client connects, write access to HID device nodes under /dev, and a systemd service file registered on the host...

I think the only way is to have a portal proposed to solve this issue or if the precedent of Piper justified it, because even though it is a host dependency, as the one @bbhtt mentioned, there is precedent for it, not only with Piper that happens to use the same daemon, but countless other examples that require another dependency such as Pipewire, NetworkManager and transmission-daemon... so on and so forth...

If you guys think that the libratbag dependency is unacceptable I think the only way is to reject my submission until something happens...

PR #8289, opened 2026-04-05
Add org.lyricapp.sofar

RedyAu:
Initial submission for the Sófár Hangoló desktop app. Source tarball is built from the tagged lyric release and ships the desktop metadata from the upstream repository.


bbhtt:
PRs not following the template will be closed.

PR #8362, opened 2026-04-12
Add fr.emersion.goguma

patwid:

- [X] Please describe the application briefly. Goguma is a modern IRC client designed for mobile devices, offering features like support for IRC bouncers, an easy-to-use interface, and offline access to past conversations. It is available on platforms like Android and iOS, and is lightweight to conserve device resources.
- [X] Please attach a video showcasing the application on Linux using the Flatpak. https://github.com/user-attachments/assets/fbb8c85f-a8d5-4cfa-a745-7e469f48deb6
- [X] I am an _(please keep whichever is applicable and remove the rest)_ author/developer/upstream contributor to the project.
      If not, I contacted upstream developers about this submission. **I contacted them on their irc channel #emersion on libera.chat on 2026-04-11 17:43 UTC and they approved it**

Since it is a flutter app with dependencies, we use https://github.com/TheAppgineer/flatpak-flutter to make the build from source work, without network access inside a sandbox. 

For sqlite to work (sqflite looks for libsqlite.so but actual name is libsqlite.0.so) we build it from source.


fr.emersion.goguma.yml
```diff
@@ -0,0 +1,68 @@
+# Generated by flatpak-flutter v0.14.1 from flatpak-flutter.yml, do not edit
+# Visit the project at https://github.com/TheAppgineer/flatpak-flutter
+app-id: fr.emersion.goguma
+runtime: org.freedesktop.Platform
+runtime-version: '22.08'
```
hfiguiere: This runtime has been EOL for almost 2 years being almost 4 years old.

fr.emersion.goguma.yml
```diff
@@ -0,0 +1,68 @@
...
+      - "for cache_dir in .pub-cache/git/cache/*/; do\n  [ -d \"$cache_dir\" ] ||\
+        \ continue\n  pkg_name=$(basename \"$cache_dir\" | sed 's/-[a-f0-9]*$//')\n\
+        \  local_clone=$(find .pub-cache/git -maxdepth 1 -type d -name \"${pkg_name}-*\"\
+        \ ! -path \"*/cache/*\" | head -1)\n  if [ -n \"$local_clone\" ]; then\n \
+        \   git --git-dir=\"$cache_dir\" remote set-url origin \"$PWD/$local_clone\"\
+        \n  fi\ndone\n"
+      - setup-flutter.sh
+      - flutter config --no-analytics
+      - setup-flutter.sh
+      - 'sed -i ''1i set(CMAKE_C_COMPILER "gcc")\nset(CMAKE_CXX_COMPILER "g++")''
```
hfiguiere: never use sed.

fr.emersion.goguma.yml
patwid: Updated runtime to 25.08

fr.emersion.goguma.yml
patwid: Replaced with tmp file approach

fr.emersion.goguma.desktop
hfiguiere, line 1: This file should be (submitted and managed) upstream.

fr.emersion.goguma.metainfo.xml
hfiguiere, line 1: This file should be (submitted and managed) upstream.

fr.emersion.goguma.yml
```diff
@@ -0,0 +1,77 @@
...
+command: goguma
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --talk-name=org.freedesktop.Notifications
+modules:
+  - name: sqlite3
```
hfiguiere, line 18: this exist in the runtime already

fr.emersion.goguma.yml
```diff
@@ -0,0 +1,77 @@
...
+      - "for cache_dir in .pub-cache/git/cache/*/; do\n  [ -d \"$cache_dir\" ] ||\
+        \ continue\n  pkg_name=$(basename \"$cache_dir\")\n  pkg_name=${pkg_name%-*}\n\
+        \  local_clone=$(find .pub-cache/git -maxdepth 1 -type d -name \"${pkg_name}-*\"\
+        \ ! -path \"*/cache/*\" | head -1)\n  if [ -n \"$local_clone\" ]; then\n \
+        \   git --git-dir=\"$cache_dir\" remote set-url origin \"$PWD/$local_clone\"\
+        \n  fi\ndone\n"
+      - setup-flutter.sh
+      - flutter config --no-analytics
+      - setup-flutter.sh
+      - '{ printf ''set(CMAKE_C_COMPILER "gcc")\nset(CMAKE_CXX_COMPILER "g++")\n''
```
hfiguiere: this is as bad as sed. I don't understand why you need to do that?

fr.emersion.goguma.yml
patwid: I updated the description. Sqflite looks for libsqlite.so, but the actual name is libsqlite.0.so. I tried to use a symlink approach, but couldn't get it to work.

hfiguiere: > Sqflite looks for libsqlite.so,

this is so wrong. Whoever wrote this has no clue how DSO versioning works and shouldn't be allowed to write that code.

fr.emersion.goguma.yml
patwid: Flutter hardcodes Clang as C/C++ compiler, but the SDK only has gcc.

CodedOre: Clang is in the LLVM SDK extension.

disable-titlebar.patch
CodedOre, line 1: While I do understand that the application "is meant for mobile devices", by putting it on Flathub people will have the chance to install it on their desktops. 

How will it work on there without titlebar?

fr.emersion.goguma.yml
patwid: Nice, the extension works!

disable-titlebar.patch
patwid: Yes, at least for users on desktop environments, this is probably not working properly. But I argue, that this is a reasonable tradeoff, since it is mainly focused for mobile devices. Maybe we could have two separate flatpaks, if needed?

fr.emersion.goguma.desktop
patwid: I opened a PR to upstream the resources.

fr.emersion.goguma.metainfo.xml
patwid: I opened a PR to upstream the resources.

disable-titlebar.patch
hfiguiere: mobile is really not the majority case for Flathub. So the best fix is to detect it instead of doing it unconditionally.

Two package will definitely not be accepted.

hfiguiere: > The detection must happen at build time, though or am I missing something?

How would it detect what it runs on at build time?

>  I meant the majority case for Goguma would be mobile users.

Then it is not suitable for Flathub.

emersion: I don't think Flathub should be shipping custom patches like this one. I would prefer to solve this particular issue upstream. See the discussion at https://codeberg.org/emersion/goguma/pulls/354

fr.emersion.goguma.desktop
emersion: Thanks for that! For future reference, the PR is at https://codeberg.org/emersion/goguma/pulls/353

fr.emersion.goguma.yml
emersion: I think I've fixed it upstream by running `flutter pub add sqlite3_flutter_libs`, which uses bundled libraries instead of the runtime's. Not sure that would be preferred, but it's an option.

disable-titlebar.patch
patwid: Dropped the patch

fr.emersion.goguma.desktop
patwid: Merged!

fr.emersion.goguma.metainfo.xml
patwid: Merged!

fr.emersion.goguma.yml
patwid: @hfiguiere What do you think?

hfiguiere: that 1. it should use the system libs, 2. that if the code is looking for a .so then the code should be fixed.

emersion: (2) sounds unreasonable since this is deeply nested in libraries and it will break all other platforms.

Moreover, other Flutter apps hosted on Flathub have been merged without this requirement:

https://github.com/flathub/im.fluffychat.Fluffychat/blob/8a659a352fc7dd735324dee08521478fcca21357/im.fluffychat.Fluffychat.json#L65

hfiguiere: 1. it's not that app we are reviewing.
2. at least they are not rebuilding sqlite3
3. what's unreasonable is ignoring the problem. "since this is deeply nested in libraries". How about you submit a patch to fix it?

emersion: I don't have any interest in "fixing" what you think is "wrong". If Goguma isn't shipped on Flathub because Flathub wants to block it based on this issue, not my problem.

fr.emersion.goguma.yml
```diff
@@ -0,0 +1,69 @@
...
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.llvm22
+command: goguma
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --talk-name=org.freedesktop.Notifications
```
bbhtt, line 16: why does this not use the portal?

fr.emersion.goguma.yml
```diff
@@ -0,0 +1,69 @@
...
+        aarch64:
+          env:
+            BUNDLE_PATH: build/linux/arm64/release/bundle
+        x86_64:
+          env:
+            BUNDLE_PATH: build/linux/x64/release/bundle
+      env:
+        PUB_CACHE: /run/build/goguma/.pub-cache
+    build-commands:
+      - "for cache_dir in .pub-cache/git/cache/*/; do\n  [ -d \"$cache_dir\" ] ||\
```
bbhtt, line 39: this looks horrible, you can use `- |`

bbhtt: also why does it have -d? it either exists in the build or doesn't in which case the build fails. the find command is already filtering to dirs.

fr.emersion.goguma.yml
emersion: The Flutter library used for notifications doesn't support the portal yet: https://github.com/MaikuB/flutter_local_notifications/issues/1757

See also:
https://github.com/flathub/im.fluffychat.Fluffychat/blob/8a659a352fc7dd735324dee08521478fcca21357/im.fluffychat.Fluffychat.json#L16

CodedOre: > See also:
> https://github.com/flathub/im.fluffychat.Fluffychat/blob/8a659a352fc7dd735324dee08521478fcca21357/im.fluffychat.Fluffychat.json#L16

Other apps don't serve as precedent. 

The requirements ands exceptions have been updated over the time for new submissions. 

And the notification portal is not new either. It can be expected from a toolkit for Linux to support this nowadays.

emersion: I respect your decision to reject this new submission, but I will note that it feels quite unfair from a submitter point-of-view. (Applying the new rules to all existing apps with a grace period would be more fair, in my opinion. That's what many other distributors are doing for what it's worth, e.g. Google Play. But that's your decision.)

bbhtt: What grace period and what "new" rules? 

The rules are listed in the requirements page and they existed before this submission was made.

emersion: My comment is a suggestion to make the rules more fair. You should feel free to ignore the suggestion if you aren't interested.

My suggestion is: each time rules for new submissions are updated, notify existing app maintainers about the changes (e.g. via a mailing list or feed or thread or similar), give them a few months to comply, and drop apps that haven't been updated. That way, all apps need to follow the same rules regardless of historical inclusion.

bbhtt: We grandfather existing apps.

And this discussion is off topic you don't get to decide this.

emersion: It is off-topic, and as I've said multiple times above, it is up to you. That's completely fine. I'm just letting you know how I feel.

PR #8402, opened 2026-04-17
Add it.mijorus.simba

mijorus:

- [x] Please describe the application briefly. < This application is design to manage the Samba server on the current machine with a simple user interface >
- [x] Please attach a video showcasing the application on Linux using the Flatpak. < https://github.com/mijorus/simba/blob/add-to-flathub/docs/Screencast%20From%202026-04-17%2013-54-08.mp4  >
- [x] The Flatpak ID follows all the rules listed in the [Application ID requirements][appid] (I own that domain).
- [x] I am an author  to the project.


it.mijorus.simba.json
```diff
@@ -0,0 +1,40 @@
+{
+    "id": "it.mijorus.simba",
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "49",
```
bbhtt, line 4: 50 is latest

it.mijorus.simba.json
```diff
@@ -0,0 +1,40 @@
...
+    "sdk": "org.gnome.Sdk",
+    "command": "simba",
+    "finish-args": [
+        "--share=ipc",
+        "--socket=wayland",
+        "--socket=fallback-x11",
+        "--device=dri",
+        "--filesystem=host",
+        "--talk-name=org.freedesktop.Flatpak"
```
bbhtt, line 13: why are these two needed?

mijorus: Sure.

Since the app is interacting with systems, these permissions are required to start and stop the samba server, as well as enabling users management and modifying the SMB.conf file with pkexec

bbhtt: https://docs.flathub.org/docs/for-app-authors/requirements#host-dependent-applications

mijorus: Can I apply for an exception, right?

bbhtt: It means we can't accept this app. It's blocked by the inclusion criteria. I linked to the specific criteria.

mijorus: Just asking... If I invite the user to install a deamon and communicate with that daemon over dbus, would that be ok?

Otherwise if there is no way out of this, I'll close this PR and distribute the app differently

bbhtt: No it still means the app will be non functional as installed from Flathub. You may look for alternate distribution mechanisms.

mijorus:
I'm a bit disappointed tbh, since if I scroll through the list of recently published apps I see a lot of them with this exception.

But anyway...

PR #8433, opened 2026-04-20
Add org.koderoots.musik

DenysMb:

- [X] Please describe the application briefly. **Simple audio player for KDE Plasma**

[Screencast_20260420_201052.webm](https://github.com/user-attachments/assets/dbbbd65f-9d2d-4b94-b199-93649ac7e0a1)

### DIFF
--- a/org.koderoots.musik.json
+++ b/org.koderoots.musik.json
@@ -9,9 +9,7 @@
     "--socket=fallback-x11",
     "--socket=wayland",
     "--socket=pulseaudio",
-    "--device=dri",
-    "--filesystem=xdg-music:ro",
-    "--filesystem=xdg-download:ro"
+    "--device=dri"
   ],
   "modules": [
     {
@@ -23,8 +21,8 @@
         {
           "type": "git",
           "url": "https://github.com/KodeRoots/Musik.git",
-          "tag": "0.1.1",
-          "commit": "776a30b002fb94330dfaf42e90d7601c353ccc99"
+          "tag": "0.1.2",
+          "commit": "e037e2119f6dd52700241cddda16311509e4db82"
         }
       ]
     }
--- a/org.koderoots.musik.json
+++ b/org.koderoots.musik.json
@@ -1,0 +1,34 @@
+{
+  "app-id": "org.koderoots.musik",
+  "runtime": "org.kde.Platform",
+  "runtime-version": "6.10",
+  "sdk": "org.kde.Sdk",
+  "command": "musik",
+  "finish-args": [
+    "--share=ipc",
+    "--socket=fallback-x11",
+    "--socket=wayland",
+    "--socket=pulseaudio",
+    "--device=dri",
+    "--filesystem=xdg-music:ro",
+    "--filesystem=xdg-download:ro",
+    "--talk-name=org.mpris.MediaPlayer2",
+    "--own-name=org.mpris.MediaPlayer2.musik"
+  ],
+  "modules": [
+    {
+      "name": "musik",
+      "buildsystem": "cmake-ninja",
+      "config-opts": ["-DCMAKE_BUILD_TYPE=Release"],
+      "builddir": true,
+      "sources": [
+        {
+          "type": "git",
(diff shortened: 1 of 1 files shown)


org.koderoots.musik.json
```diff
@@ -0,0 +1,34 @@
...
+  "finish-args": [
+    "--share=ipc",
+    "--socket=fallback-x11",
+    "--socket=wayland",
+    "--socket=pulseaudio",
+    "--device=dri",
+    "--filesystem=xdg-music:ro",
+    "--filesystem=xdg-download:ro",
+    "--talk-name=org.mpris.MediaPlayer2",
+    "--own-name=org.mpris.MediaPlayer2.musik"
```
bbhtt: this should use the appid as subname then the permission is not needed

org.koderoots.musik.json
```diff
@@ -0,0 +1,34 @@
...
+  "command": "musik",
+  "finish-args": [
+    "--share=ipc",
+    "--socket=fallback-x11",
+    "--socket=wayland",
+    "--socket=pulseaudio",
+    "--device=dri",
+    "--filesystem=xdg-music:ro",
+    "--filesystem=xdg-download:ro",
+    "--talk-name=org.mpris.MediaPlayer2",
```
bbhtt: why do you need this?

org.koderoots.musik.json
```diff
@@ -0,0 +1,34 @@
...
+  "sdk": "org.kde.Sdk",
+  "command": "musik",
+  "finish-args": [
+    "--share=ipc",
+    "--socket=fallback-x11",
+    "--socket=wayland",
+    "--socket=pulseaudio",
+    "--device=dri",
+    "--filesystem=xdg-music:ro",
+    "--filesystem=xdg-download:ro",
```
bbhtt: why are these needed?

org.koderoots.musik.json
```diff
@@ -0,0 +1,34 @@
...
+  "modules": [
+    {
+      "name": "musik",
+      "buildsystem": "cmake-ninja",
+      "config-opts": ["-DCMAKE_BUILD_TYPE=Release"],
+      "builddir": true,
+      "sources": [
+        {
+          "type": "git",
+          "url": "https://github.com/KodeRoots/Musik.git",
```
bbhtt, line 23: no licence file in repo root.

org.koderoots.musik.json
DenysMb: Yeah, it's not needed. Removed.

org.koderoots.musik.json
DenysMb: Removed.

org.koderoots.musik.json
DenysMb: If the user opens Dolphin, for example, and go to Music folder and click on the song, this is needed to the application to be opened with the selected audio file. I added both directories because they are the most common directories where audio files are stored.

Folders that are not `Music` or `Download`, when the user tries to open a song file clicking on it, it will open the application but without the audio file selected. So they need to open from the application (or drag and drop the audio to the application).

If we remove this, then the user won't be able to open songs like this from anywhere else.

In the ideal world, the permission here should be `home:ro`, so all audio files in user's home folder would be able to be opened like this, but since we need to avoid adding this permission, I added only these two folders, because it's better than nothing.

[Screencast_20260421_130001.webm](https://github.com/user-attachments/assets/90629208-42fd-4fac-a293-dff10891ee5e)

org.koderoots.musik.json
DenysMb: Added.

org.koderoots.musik.json
bbhtt: no that's what file-forwarding is for which Flatpak writes in desktop files automatically

DenysMb: Yeah, the problem is that I was using `%U` instead of `%f` in the desktop file. Makes sense. Removed.

bbhtt:
bot, build

PR #5529, opened 2024-08-23
Add surf.deckr.deckr

sgehrman:
- [x] Please describe the application briefly.
 This app works with the deckr chrome extension.  This app already exists, but I had to change the app id to get the verification working.  see com.cocoatech.deckr

- [x] The domain used for the application ID is [controlled by the application developers][appid-domain] and the [application id guidelines][appid] are followed.
- [x] I have read the and followed all the [Submission and licence requirements][reqs].
- [x] I have [built][build] and tested the submission locally.
- [x] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about this submission. **Link:**

[appid-domain]: https://docs.flathub.org/docs/for-app-authors/requirements/#control-over-domain-or-repository
[appid]: https://docs.flathub.org/docs/for-app-authors/requirements#application-id
[reqs]: https://docs.flathub.org/docs/for-app-authors/requirements
[build]: https://docs.flathub.org/docs/for-app-authors/submission/#before-submission

PR #8455, opened 2026-04-22
Add com.getgrist.grist

wvengen:

- [x] Please describe the application briefly: _Grist is a modern relational spreadsheet. It combines the flexibility of a spreadsheet with the robustness of a database._
- [x] Please attach a video showcasing the application on Linux using the Flatpak: _[video](https://github.com/user-attachments/assets/53a58b38-4518-43d2-a977-a982f1a3bbf7)_
- [x] I am a contributor for the Flathub submission, with the app authors agreeing to my submission in https://github.com/gristlabs/grist-desktop/issues/105#issuecomment-4247121197

Additional maintainers: @paulfitz @dsagal 

Some notes on the application / manifest:
- The application's main application is in Javascript, running in Electron.
- Formulas / code (Python) run in a [Pyodide](https://pyodide.org/) sandbox by default. This needs Python modules compiled with Pyodide / emscripten. Upstream does this manually (using a Docker container), and hosts these compiled Python modules on AWS S3. To allow the modules to build from source, some changes to the build process were necessary.
-  Getting pyodide and its dependencies to fully build from source is quite a job, it requires specific commits of emscripten, llvm, and others. It seems most people use [pyodide-build](https://pyodide.org/en/stable/development/building-packages.html) to download precompiled versions that work together and are tested. I've stuck to that approach, otherwise this part would become more and more complex. If there would be multiple applications on Flathub using emscripten, it may be sensible to look into making an Sdk extension, but I think that is a bit too much right now.
- Some small changes to the build process were necessary, to make it work without network access.
- There is a desire to allow users run without sandboxing as well, so the Python modules are also installed inside the Flatpak itself as regular Python modules.
- Next upstream version will include metainfo, desktop, mime in the source tarball, instead of downloading them separately - https://github.com/gristlabs/grist-desktop/pull/113
- Next upstream version will include the disable-update-patch, instead of the patch included here - https://github.com/gristlabs/grist-desktop/pull/112
- ~~I have not tested on ARM, that would be interesting, but may need adaptations, esp. regarding pyodide (but perhaps also precompiled node packages).~~ added


wvengen:
The build is failing with error `appid-url-not-reachable`, because the user-agent `python-requests/2.33.0` seems to be blocked on the website, which is webhoster [WPX](https://wpx.net/). Is reachability with this user-agent a requirement?

Looks similar to https://github.com/flathub/flathub/pull/5529#issuecomment-2325455509

hfiguiere:
If you decide to break the web, then I'm not sure how we can fix it.

wvengen:
Thanks for looking at this. It's certainly not so nice, but I'm afraid it will not be so easy to change (getting the hoster to change their settings or moving the website to a different hoster is both a challenge). I will try to see what is possible.

If this fails, could you consider adding this app to the [linter exceptions](https://github.com/flathub-infra/flatpak-builder-lint/blob/master/flatpak_builder_lint/staticfiles/exceptions.json) (`surf.deckr.deckr` has the same reason).

**update** the hoster now accepts the user-agent (for this website); that was easier than I expected - thanks for suggestion to fix it properly!

bbhtt:
Yes it's fine, you can open a PR for exception.

pyodide-worker-yarn.lock
bbhtt, line 1: this should not be here

pyodide-worker-package.json
bbhtt, line 1: this should not be here

patches/disable-update.patch
bbhtt, line 1: some of these patches should be upstreamed.

wvengen: The disable-update patch has already been merged upstream https://github.com/gristlabs/grist-desktop/pull/112 and will be part of the next release (then this patch can be dropped).

The skip-fetch-pyodide-packages patch is to allow building the pyodide packages locally (instead of fetching precompiled packages by the Grist developers separately from the regular build process).
The relax-astroid-build-deps patch is to allow building the astroid dependency with pyodide.

pyodide-worker-package.json
wvengen: It's indeed cleaner to get these files properly upstream. I'll open a PR there. Then for the next release of this app, these files can be dropped.

wvengen:
bot, build

wvengen:
bot, build

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "only-arches": ["x86_64"]
```
hfiguiere: why?

wvengen: To compile Python packages to WASM, [pyodide](https://pyodide.org/) is used, which uses [emsdk](https://emscripten.org/docs/tools_reference/emsdk.html), which uses various precompiled components for compiling to WASM, hence is architecture-specific. It is currently only worked out for x86_64. See the [PR description](https://github.com/flathub/flathub/pull/8455#issue-4309203866) for a bit more information.

wvengen: To honour the build-from-source approach, I have chosen to build the pyodide Python packages from source. This brings in quite some complexity, and currently restricts the architecture to x86_64. If you think it would be better to use precompiled pyodide Python packages from a gristlabs-provided S3 bucket, that would be an option (though, perhaps, setting an undesired example). That would also make the build work on `aarch64` right away (most probably). Curious to hear your thoughts on this.

wvengen:
bot, build

flathub.json
wvengen: I was lucky, it required only minor changes. I hope I can find someone to test the build!

As each new arch needs additional work, I've left the `flathub.json` with `x86_64` and `aarch64`.

flathub.json
hfiguiere, line 1: remove this file

pyodide-worker-package.json
bbhtt: you have to remove it before I consider accepting the submission. we can wait for the next release.

wvengen: Thank you for the note!

This is work in progress in https://github.com/gristlabs/grist-core/pull/2297. Once this is accepted, I could download both files from the grist-core repository instead of including them here. In the next release, these will be part of the source tarball.

wvengen:
bot, build

wvengen:
Updated after new upstream version. Now the local source patch, yarn lockfiles and desktop metadata are used from the upstream archive instead, and I think all issues have been resolved. Please let me know if there's anything to improve here.

Note that this app upgraded to Electron 41, and even though [Flatpak documentation recommends](https://docs.flatpak.org/en/latest/electron.html#sandbox-permissions) to use Xwayland, it seems that Wayland [now is the default](https://www.electronjs.org/blog/tech-talk-wayland) (the app crashed when not adding wayland permissions when running in a Wayland desktop). This worked fine with manual testing, so leaving it with Wayland + X11-fallback.

com.getgrist.grist.yml
```diff
@@ -0,0 +1,177 @@
...
+command: grist-desktop
+separate-locales: false
+finish-args:
+  - --share=ipc
+  - --device=dri
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --share=network
+  - --filesystem=~/.grist
```
hfiguiere: this should be `persist` instead

wvengen: This is is a standard location, where people can put e.g. local plugins. It could be persist, but that would require the user to know whether app is installed as Flatpak from Flathub, or e.g. AppImage. So I think it is user-friendly to keep this path as filesystem. What do you think?

wvengen: Note that this is not used by the app to store state data (that uses `electron.app.getPath("documents")` which ends up in `~/.var/app/com.getgrist.grist/config`).

hfiguiere: > This is is a standard location

not it's not standard since this doesn't follow the XDG standard.

hfiguiere: What I think that this location is incorrect and your assumption is incorrect. What if the plugins require things that are not in the sandbox. are for a different version or even architecture. etc.

wvengen: A Grist plugin is not a shared library or binary, but basically a web page that can be embedded for custom functionality. Usually, this is done by using an external URL ([documentation on custom widgets](https://support.getgrist.com/widget-custom/#custom-widget-linking)). But it is also possible, to provide an offline custom widget. That is called a plugin. It is self-contained and does not need anything from the Flatpak environment (other than the provided API for crossing the embedding boundary, but that is all in-browser).

I could set this to something else easily with an environment variable, e.g. in the app-state directory. But that would require an extra section in any documentation on how to use this feature, explaining that the path is different for the Flatpak.

wvengen: > it it's not standard

Agree - there is no real standard for this. It's just introducing different behaviour across different installations.

I now see that almost no Flatpak has something like this. Sorry for not noticing earlier. I'll consider where to put it.

wvengen:
bot, build

com.getgrist.grist.yml
wvengen: We've decided to move the (default for this) path somewhere else in the future, and I've updated the manifest to reflect that with the current version. Thanks for setting this discussion in motion!

com.getgrist.grist.yml
```diff
@@ -0,0 +1,178 @@
...
+finish-args:
+  - --share=ipc
+  - --device=dri
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --share=network
+  - --env=ELECTRON_TRASH=gio
+  - --env=XCURSOR_PATH=/run/host/user-share/icons:/run/host/share/icons
+  - --env=TMPDIR=/tmp
```
bbhtt: why is this needed?

com.getgrist.grist.yml
```diff
@@ -0,0 +1,178 @@
...
+  - org.freedesktop.Sdk.Extension.node24
+command: grist-desktop
+separate-locales: false
+finish-args:
+  - --share=ipc
+  - --device=dri
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --share=network
```
bbhtt, line 17: why is this needed?

com.getgrist.grist.yml
```diff
@@ -0,0 +1,178 @@
...
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node24
+command: grist-desktop
+separate-locales: false
+finish-args:
+  - --share=ipc
+  - --device=dri
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
```
bbhtt, line 16: why is this needed?

com.getgrist.grist.yml
```diff
@@ -0,0 +1,178 @@
...
+cleanup:
+  - ".yarnrc" # offline config during build
+  - "*.musl.node" # we use glibc, not musl
+  - "*-musl.node" # we use glibc, not musl
+  - "*.exe" # and ship even windows binaries
+cleanup-commands:
+  # remove temporary build directories
+  - find /app -name 'build-tmp-*' -o -name obj.target | xargs rm -Rf
+  # remove more npm binaries for other platforms
+  - find /app -name darwin-x64 | xargs rm -Rf
```
bbhtt: in the top before modules:

com.getgrist.grist.yml
wvengen: - Grist has the option to add custom widgets from remote locations (html / css / js).
- These custom widgets may show e.g. a map from an online source.
- One may want to show e.g. images or movies from remote URLs in a table.

com.getgrist.grist.yml
wvengen: it isn't, thank you

com.getgrist.grist.yml
wvengen: - there is an introduction movie that can be played, which contains sound
- a user may want to play an movie using the [video widget](https://support.getgrist.com/widget-custom/#grist-video-player)

bbhtt:
bot, build

bbhtt:
This looks very much like a web service wrapped in Electron that is not a desktop application. It still has the default Electron menubar, half of the dropdowns are not relevant, accounts / logins etc.

This would violate the web wrapper policy.

wvengen:
This submission is an Electron version of an application that exists as an online web service. It is meant to exist in both ways: online for groups to work together, and have features like webhooks and such; and as a desktop application, that can open the same files (sqlite-based), but can be used locally and also offline.
Perhaps it can be compared to other office applications, like OpenOffice/LibreOffice existing both as a desktop and online version ([Collabora Online](https://www.collaboraonline.com/) and other derivatives), or Microsoft Office, that also exists as native applications and is available on the web.
Yes, there are some parts where the parts relevant for online, kind of leak to the desktop. This has been mitigated in various cases, but there are some edges. If there is enough interest in the desktop version, I think this will be able to be improved over time. Also, there is an issue with the help menu not always functioning, which you may have encountered.

But do note that this is not a wrapper of an online service: all the code runs locally. This desktop application is a way to run the server-side component locally without needing to install it on a server (and have it more integrated into the desktop). One could verify this by removing the network permission, and see that the application still works as intended.

From the [Flathub requirements](https://docs.flathub.org/docs/for-app-authors/requirements#no-trademark-violations):
> Simple web wrapper applications that embed a web engine and load a remote website without adding significant functionality or meaningful desktop integration will not be accepted.

Please let me know if I can clarify this more. Technically, it is a web wrapper for a service bundled within the Flatpak. It is not a web wrapper for a remote service. As to my understanding, this submission would look like it qualifies for inclusion, but please let me know if I misunderstood this.

In any case, thank you for guarding the quality of Flathub.

**updated** added link to Collabora Online

hfiguiere:
> Perhaps it can be compared to other office applications, like OpenOffice existing both as a desktop and online version

That's actually false.

Spoffy:
Hey all, I'm one the engineers at Gristlabs that maintains Grist Desktop.

As @wvengen has described, Grist Desktop is a fully offline, portable version of the hosted Grist service. It does use Electron to allow us to share the same UI, but doesn't (and technically can't) access a hosted Grist service at all.

The intention of Grist Desktop is to allow someone to work with Grist documents without needing any hosting knowledge or to use our hosted SaaS. This might be someone at home, or more frequently someone traveling without any network connectivity, such as on public transport or in remote areas.

.grist files themselves are fully portable and self-contained (the same as any .ods or .xlsx file), and can be downloaded and moved between Grist instances. Grist Desktop loads and saves them from the filesystem as in any typical office software. 

It definitely has a few rough edges, such as the default Electron toolbar, or the login system. The login system is bypassed entirely in Grist Desktop to use a single, automatically logged in, local-only account. It'd be great to adjust the interface to hide that, but it's tricky to dedicate time right now to something which doesn't have a substantial impact on the quality of life or feature set available to users. 

One hope we've got is that Flathub will make Grist Desktop more accessible to a larger audience, as currently it's only hosted on our Github repository. That would help us to dedicate more resources to improving the level of polish it has as a desktop app.

Hope that helps clear things up!

bbhtt:
I can't accept it if it violates the inclusion policy, if you can fix the issues it can be considered.

wvengen:
Thanks for your reply.
Could you indicate exactly which part of the policy this application currently violates?

I can find in [Flathub requirements](https://docs.flathub.org/docs/for-app-authors/requirements#no-trademark-violations):

> Simple web wrapper applications that embed a web engine and load a remote website without adding significant functionality or meaningful desktop integration will not be accepted.

This application does not violate this requirement, as a remote website is not loaded.

bbhtt:
You seem to be more interested on finding your way out of the policies than fixing the issues mentioned that you yourself acknowledged were 'issues'.

This is starting to feel like a scheme to dump the app here. If you see at the top it mentions accepting high quality properly integrated apps multiple times which this is not.

It looks exactly like the previous comment described, that the hosted service was wrapped in Electron without much attention to detail.

paulfitz:
> This is starting to feel like a scheme to dump the app here.

CTO of Grist Labs here. Grist actually began life as a desktop app, many years ago, then migrated to the web. In my spare time, and building on some community work, I made a desktop version that eventually the company started to maintain. It definitely isn't a priority (there's no money in it), and that does show in the lack of polish. If it isn't good enough for flathub, that's totally fine, I just feel bad for @wvengen (thanks for the time you put into this!!).

A few data points if it makes any difference:
 * There's apparently been an AUR package since 2024 - https://aur.archlinux.org/packages/grist-desktop-bin
 * A pretty typical user story - https://community.getgrist.com/t/run-grist-desktop-as-a-local-lan-server-on-macos-script/13176 (offline needed, or install too nerdy otherwise)
 * There's maybe a winget package? idk, just came across it searching https://github.com/microsoft/winget-pkgs/pull/159490
 * Despite not being able to justify much work on it, grist-desktop DOES get regular updates since its core is identical to all our Grist flavors (server, desktop, static, enterprise) meaning the spreadsheet stuff users care about (as opposed to that sad menu bar) is in really excellent shape (try it!) and will remain so.

wvengen:
Closing, until Grist Desktop is more adapted to desktop use.
I hope the work done here will be of use then.

PR #8496, opened 2026-04-27
Add systems.fracture.launcher

parker-bryce-andrew:

Fracture windows then resize, modify, and use shaders with the fractured selection. 

This application creates fractured views of other applications and desktops. A fractured view is just a mirrored view of a region cut (or cropped) from the original application. After a fractured view is created, it provides the ability to modify it by resizing, scaling, adding transparency, removing colors ranges, adding colors, and by using shaders on it. 

The repository is [here](https://github.com/parker-bryce-andrew/fracture).  

The repository still states it's made for Gnome, but it's not specific to Gnome. If it's not on Gnome, it uses a portal for window or screen selection. 

https://github.com/user-attachments/assets/150d306a-2786-4609-8be5-b5b24662b32b

- [x] I am the author and only developer to the project.


img/icon/128x128.png
teamcons, line 1: This should be upstream

img/icon/256x256.png
teamcons, line 1: This should be upstream

img/icon/32x32.png
teamcons, line 1: All the images should be upstream yo

flathub.json
teamcons, line 1: Why? 
If because you cant test: not valid

fracture.desktop
teamcons, line 1: This should be upstream
No icon and wrong exec path

fracture.metainfo.xml
teamcons, line 1: This should be upstream

LICENSE-APACHE
teamcons, line 1: This should be upstream

img/icon/32x32.png
parker-bryce-andrew: In the repository? Where do I add them?

fracture.desktop
parker-bryce-andrew: The icon path is added in the .yml file [here](https://github.com/parker-andrew-forks/flathub/blob/my-app-submission/systems.fracture.launcher.yml#L25). It does correctly add the icon. When I wrote it, I originally was going to make it so I could more easily change the app id for testing, but then never went back to complete the rest.

flathub.json
parker-bryce-andrew: I tried to compile for aarch64 and received errors. I think was a crate that couldn't compile for aarch64 because the method used i8 on aarch and u8 on x64. I didn't want to return to fix it yet because I didn't know how much more involved it was going to be to finish all of it.

fracture.desktop
teamcons: You need a proper buildsystem. This is like fixing a spaceship with bra straps and bubblegum

img/icon/32x32.png
teamcons: The most common ive seen is to have a /data to put desktop, metainfo, icons, screenshots... You can take example on gnome apps directory structure.

Following what is done elsewhere makes it easier for new contributors to jump in

parker-bryce-andrew: I made a [new commit](https://github.com/flathub/flathub/pull/8496/commits/8f6deb4097463e571b1c36e31f14778c170bcfbd). These are all installed from the repository directly now. I also simplified the .desktop file to just include the id directly for now. The images and other data is removed in the new commit.

parker-bryce-andrew:
I made a [new commit](https://github.com/flathub/flathub/pull/8496/commits/8f6deb4097463e571b1c36e31f14778c170bcfbd). 

It resolves
* Issues with the linter. The `sdk-extensions` were defined incorrectly. They should be corrected now, and when I tested building, it seemed to correctly install the right versions of the extensions if they were missing. It also removed the `commands` line that caused `manifest-unknown-properties.`
* The .desktop file is simplified with directly defining the application id.
* The images and other metadata are installed directly from one of the modules repositories instead of in this pull requests repository.

I also
* Added the https://fracture.systems/.well-known/org.flathub.VerifiedApps.txt as requested earlier by the bot.

parker-bryce-andrew:
bot, build systems.fracture.launcher

systems.fracture.launcher.yml
```diff
@@ -0,0 +1,84 @@
+id: systems.fracture.launcher
+# The GNOME runtime and sdk includes gtk
```
bbhtt:
```suggestion
```

systems.fracture.launcher.yml
```diff
@@ -0,0 +1,84 @@
...
+runtime: org.gnome.Platform
+sdk: org.gnome.Sdk
+sdk-extensions:
+  [
+    org.freedesktop.Sdk.Extension.llvm20,
+    org.freedesktop.Sdk.Extension.rust-stable,
+  ]
```
bbhtt:
```suggestion
sdk-extensions:
  - org.freedesktop.Sdk.Extension.llvm20
  - org.freedesktop.Sdk.Extension.rust-stable
```

systems.fracture.launcher.yml
```diff
@@ -0,0 +1,84 @@
...
+      - cargo --offline fetch --manifest-path Cargo.toml --verbose
+      - cargo build --offline --release --all-features
+      - install -Dm0755 target/release/gsc-launcher /app/bin/launcher
+    sources:
+      - type: git
+        url: https://github.com/parker-bryce-andrew/fracture-launcher.git
+        commit: d40424ff57f2a99e3171fe8e0ce7801e13df1660
+      - cargo-launcher-sources.json
+
+finish-args:
```
bbhtt: this should be on top after sdk-extensions

systems.fracture.launcher.yml
```diff
@@ -0,0 +1,84 @@
...
+  - --socket=wayland
+  # This uses WebGPU and DmaBuffers with pipewire
```
bbhtt:
```suggestion
```

systems.fracture.launcher.yml
```diff
@@ -0,0 +1,84 @@
...
+  - --device=dri
+  - --talk-name=org.freedesktop.Screencast
+  - --filesystem=xdg-run/pipewire-0
+  #
+  # The rest is used only on GNOME
+  #
+  # It uses dconf to add a keyboard shortcut for Gnome users
```
bbhtt:
```suggestion
```

systems.fracture.launcher.yml
```diff
@@ -0,0 +1,84 @@
...
+      - type: git
+        url: https://github.com/parker-bryce-andrew/fracture-launcher.git
+        commit: d40424ff57f2a99e3171fe8e0ce7801e13df1660
+      - cargo-launcher-sources.json
+
+finish-args:
+  - --socket=wayland
+  # This uses WebGPU and DmaBuffers with pipewire
+  - --device=dri
+  - --talk-name=org.freedesktop.Screencast
```
bbhtt: where does this exist?

systems.fracture.launcher.yml
```diff
@@ -0,0 +1,84 @@
...
+  # It uses dconf to add a keyboard shortcut for Gnome users
+  - --talk-name=org.freedesktop.Flatpak
+  # GNOME users can optionally enable extension support allowing
+  # them to easily launch on the active window.
+  - --talk-name=org.gnome.Shell
+  - --talk-name=org.gnome.Shell.Extensions
+  - --talk-name=org.gnome.Mutter.ScreenCast
+  # If the screencast fails with GNOME's version, it sends a Notification
+  # and fails back to freedesktop.Screencast
+  - --talk-name=org.freedesktop.Notifications
```
bbhtt: this should use the portal.

systems.fracture.launcher.yml
```diff
@@ -0,0 +1,84 @@
...
+  - --socket=wayland
+  # This uses WebGPU and DmaBuffers with pipewire
+  - --device=dri
+  - --talk-name=org.freedesktop.Screencast
+  - --filesystem=xdg-run/pipewire-0
+  #
+  # The rest is used only on GNOME
+  #
+  # It uses dconf to add a keyboard shortcut for Gnome users
+  - --talk-name=org.freedesktop.Flatpak
```
bbhtt: is this optional too? having a sandbox escape for optional features is not a good idea.

also why can't you do this by accessing the dconf db instead of asking for flatpak-spawn?

systems.fracture.launcher.yml
parker-bryce-andrew: Why are you removing comments? It's just defining why the runtime was used

systems.fracture.launcher.yml
parker-bryce-andrew: This was not working. This was how it was originally defined, it was wrong. It fails validation and it doesn't install the extensions if they're missing.

systems.fracture.launcher.yml
```diff
@@ -0,0 +1,84 @@
...
+      env:
+        CARGO_HOME: /run/build/launcher-build/cargo
+        CARGO_NET_OFFLINE: "true"
+    build-commands:
+      - cargo --offline fetch --manifest-path Cargo.toml --verbose
+      - cargo build --offline --release --all-features
+      - install -Dm0755 target/release/gsc-launcher /app/bin/launcher
+    sources:
+      - type: git
+        url: https://github.com/parker-bryce-andrew/fracture-launcher.git
```
bbhtt, line 77: so this is a script written in rust where you hardcoded the path https://github.com/parker-bryce-andrew/fracture-launcher/blob/d40424ff57f2a99e3171fe8e0ce7801e13df1660/src/main.rs#L5? why isn't this part of the above?

systems.fracture.launcher.yml
bbhtt: it's saying the most obvious thing.

systems.fracture.launcher.yml
bbhtt: there's no difference just a change of syntax.

parker-bryce-andrew:
```
sdk-extensions: 
  org.freedesktop.Sdk.Extension.llvm20/x86_64/25.08
  org.freedesktop.Sdk.Extension.rust-stable/x86_64/25.08
```
I originally had that in the submission version. This builds if I have those extensions installed already, but if they're missing it doesn't build. I think I tried the other version too. Those specifically failed the linter by the bot.

I tested the version I submitted though, and it successfully installs if they're missing.

systems.fracture.launcher.yml
parker-bryce-andrew: I don't know how. I would have avoided it if I could find a way. I spent a lot of time searching too. I think it's only used to check keyboard shortcuts and to define a new one.

systems.fracture.launcher.yml
parker-bryce-andrew: In some environments, like VirtualBox, it will use DmaBuffers but panic when importing them. I don't know what's wrong. I don't think this problem is unique to this project, I tried using OBS in VirtualBox and it also crashed.

The launcher handles these panics and restarts in a safe mode that will hopefully not also crash.

systems.fracture.launcher.yml
parker-bryce-andrew: I don't think it's being used either. Maybe I thought ashpd called into it? I don't know, the portal for screencasting seems to work with it removed.

systems.fracture.launcher.yml
parker-bryce-andrew: I'm `notify-rust` which I think calls this directly [here](https://docs.rs/crate/notify-rust/latest/source/src/xdg/mod.rs#46)

systems.fracture.launcher.yml
parker-bryce-andrew: It's now removed [here](https://github.com/flathub/flathub/pull/8496/commits/82ffeb8180c7ebbdb1f23b3d6c59dcc392961f4f)

systems.fracture.launcher.yml
parker-bryce-andrew: I moved it to the top in the latest commit [here](https://github.com/flathub/flathub/pull/8496/commits/82ffeb8180c7ebbdb1f23b3d6c59dcc392961f4f).

systems.fracture.launcher.yml
parker-bryce-andrew: After comitting the suggestion, I went and tested it after uninstalling. It worked without issue. Thanks for the feedback.

parker-bryce-andrew:
I left the talk for Notification because I think it's used by `notify-rust`, but removed the extra Screencast talk.

There's still the 
```
- --talk-name=org.freedesktop.Flatpak
```
Which is used to run `dconf` to check keyboard shortcuts and add a keyboard shortcut. The Flathub docs mention using dconf [here](https://docs.flatpak.org/en/latest/sandbox-permissions.html#dconf-access) but I couldn't get the configuration to work with these permissions.

parker-bryce-andrew:
If it's really needed, I can remove the talk. I think users of this application might not have issues setting the shortcut themselves. Though, I don't want to remove it even with the extra warnings on the flatpak.

This is what happens to my application when it's removed

Screenshot From 2026-04-28 00-02-17

systems.fracture.launcher.yml
```diff
@@ -0,0 +1,73 @@
+id: systems.fracture.launcher
+runtime: org.gnome.Platform
+sdk: org.gnome.Sdk
```
bbhtt, line 4:
```suggestion
runtime-version: "50"
sdk: org.gnome.Sdk
```

systems.fracture.launcher.yml
```diff
@@ -0,0 +1,73 @@
...
+  - --talk-name=org.freedesktop.Notifications
+  # GNOME users can optionally enable extension support allowing
+  # them to easily launch on the active window.
+  - --talk-name=org.gnome.Shell
+  - --talk-name=org.gnome.Shell.Extensions
+  - --talk-name=org.gnome.Mutter.ScreenCast
+build-options:
+  append-path: /usr/lib/sdk/rust-stable/bin:/usr/lib/sdk/llvm20/bin
+  prepend-ld-library-path: /usr/lib/sdk/llvm20/lib
+runtime-version: "50"
```
bbhtt:
```suggestion
```

systems.fracture.launcher.yml
bbhtt: how are you setting it? there's Gromit-MPX on Flathub which uses a dconf hole to set shortcuts instead of flatpak-spawn.

systems.fracture.launcher.yml
bbhtt: i'm saying it doesn't need to be a separate thing, it could be part of the main project which is above.

parker-bryce-andrew: They both have different sets of crates and different sources. I don't have an opinion on how it should be defined in the .yml file, this was just the result of failing to define them together already. The original repository [does have them together](https://github.com/parker-bryce-andrew/fracture/tree/main/flatpak/launcher-crate), I never went back and removed it because the launcher is still very simple.

systems.fracture.launcher.yml
parker-bryce-andrew: I tried following the instructions [here](https://docs.flatpak.org/en/latest/sandbox-permissions.html#dconf-access). When that failed I tried copying the [Dconf Editor](https://github.com/GNOME/dconf-editor/blob/master/build-aux/ca.desrt.dconf-editor.json) flatpak, then realized it had a talk to `org.freedesktop.Flatpak` and gave up and just defined the talk.

I didn't know of other flatpaks that had solved it when I was writing it, and everything else I tried seemed to fail. It seems like Gromit also has a specific [dconf binary](https://github.com/bk138/gromit-mpx/blob/72eab92b6a9c92b14523de77ff5d39815ad5d2e1/flatpak/net.christianbeier.Gromit-MPX.yml#L30) that it uses. I might be able to make it work, but I don't immediately know what to change.

systems.fracture.launcher.yml
hfiguiere: 1. so I'm reading "this is to work around bugs"
2. this could be a shell script.

flathub.json
hfiguiere: it's not like you are trying. we can wait for it to be resolved.

systems.fracture.launcher.yml
parker-bryce-andrew: 1. It is, but I think correctly fixing it would be very complicated when I don't even know if people will really want to use this software. Like I have a MacBook Pro from 2014 that I tested with it, and it works, but the the screen flickers when using the DmaBuffers. The graphics driver warns that it's incomplete in the terminal. And if I try to do lots with it I can get it to panic. 

And like I had already mentioned, I had an experience of trying to use OBS in VirtualBox and also having it crash. This catches crashes from VirtualBox and restarts it with a CPU based rendering that doesn't crash.

2. It could be, but this seemed easier for me to reliably maintain. I might extend it more, but I don't know if anyone will really use this software yet.

flathub.json
parker-bryce-andrew: That's not true. I just feel like I'm increasing my workload without purpose.

Screenshot From 2026-04-28 01-46-02

hfiguiere: The crate is a fork of yours.

parker-bryce-andrew:
Is it possible to just drop the talk for to Flatpak for dconf? It's not very important to my application. I do want to support setting the keyboard shortcut later. I also think just having permission to all of dconf isn't ideal either for a single keyboard shortcut.

I also don't want to support aarch64 at the moment. I don't own a device that runs it. It wasn't in the docs as required. I do want my application supported everywhere, it's written with WebGPU, but I didn't plan on writing support for Windows and MacOS either at the moment.

Is this OK?

bbhtt:
If you want to support setting shortcuts I suggest implementing the "editing dconf" way first or checking if it is at least possible. There's no point in having the dconf permission but not the feature in the app.

parker-bryce-andrew:
The permissions for dconf seem to be working after my [latest pull request](https://github.com/flathub/flathub/pull/8496/commits/06b464c1f0370a31bd63fc21f71789726c254391). 

It also
* Removes the Flatpak dbus call that was too permissive
* At least repairs the errors that were reported for wgpu-bridge in my comment [here](https://github.com/flathub/flathub/pull/8496#discussion_r3152089408) for aarch64
* No longer prevents aarch64 from building. I don't know if it will build though.

parker-bryce-andrew:
bot, build systems.fracture.launcher

flathub.json
parker-bryce-andrew: It is. I went ahead and changed it in the latest commit. The compiler made it all the way to linking and then failed. It might work, I don't know, I started to use cross for rust then had missing dependencies in my docker image. So I just comitted it and removed the arch restriction. 

This fork, or crate, I don't think it's the right way to be creating an instance that will be used for wgpu. I think I'm supposed to call into `init_with_callback` from [here](https://github.com/gfx-rs/wgpu/blob/trunk/wgpu-hal/src/vulkan/instance.rs#L562)  to correctly initialize a wgpu instance with additional extensions. 

This is to say I really don't want to move too far into repairing something that I need to rewrite, but I don't know if this is the only issue either.

My latest commit is [here](https://github.com/flathub/flathub/pull/8496/commits/06b464c1f0370a31bd63fc21f71789726c254391).

parker-bryce-andrew:
It is successfully building and running for me on `aarch64` using the `flatpak-builder`. 

Screenshot From 2026-04-28 15-56-17

It's amazing to me that this all works

systems.fracture.launcher.yml
bbhtt: notification portal can be implemented in the app itself.

parker-bryce-andrew: The code is much less clean, I have to use an async runtime on top of all of these changes and rebuild the dependencies for the launcher I think

Screenshot From 2026-05-01 01-36-52

I'll start on this, but is there really a difference in support level? Is this really an issue?

parker-bryce-andrew:
bot, build

systems.fracture.launcher.yml
parker-bryce-andrew: The ashpd demo is working for me, but the code example is failing. 

https://github.com/user-attachments/assets/cb01bcbf-ee15-4ac0-b0dd-6792a1ab177b

The crate I'm using has almost 100 million downloads. 

Screenshot From 2026-05-01 03-26-10

Am I just wrong again? I'll have to try to determine later why I'm failing to use example code. 

The notifications also really aren't important for running my application, it just provides the user information on how to stop the first crash on start. I think there's only 2 places in the code where it's used.

parker-bryce-andrew: I was going to comment about the support level seeming good on the notifications crate, maybe even better than xdg-desktop-portals, but I forgot to write that part. 

Screenshot From 2026-05-01 03-34-30

parker-bryce-andrew: I mixed up the crate. 8 million not 100 million 

Screenshot From 2026-05-01 03-40-11

parker-bryce-andrew: I think I'm just wrong again. Maybe I'm forgetting about the sandbox for flatpak, and that I'm using a runtime that includes GTK which provides support to the portals. I'm just not thinking it through very clearly, not spending the time to understand the problem. 

I think I mostly just want to get my application published. I'm more so expecting there to not be a way for the application to work if the users aren't receiving the notification, while users that do receive it are just provided information on how to stop it from failing to safe mode on start.

parker-bryce-andrew:
bot, build

parker-bryce-andrew:
@bbhtt So I've been thinking that maybe we should include both notification options for extra compatibility to catch those X11 environments without xdg-desktop-portals. 

I think that's everything now. It was just user error, I think ashpd's notification portal only works inside of flatpaks.

parker-bryce-andrew:
I actually think this may have failed. When I tested the build and forced an error it failed to restart. I think I need to fix it.

Screenshot From 2026-05-01 17-35-57

parker-bryce-andrew:
bot, build

parker-bryce-andrew:
It's working now.

Screenshot From 2026-05-01 18-20-39

PR #8507, opened 2026-04-30
New pr

johnpetersa19:
A GTK4 application to manage and install mainline Linux kernels.  
This version is adapted for Arch Linux and allows users to browse and download kernels.

- [ ] Please attach a video showcasing the application on Linux using the Flatpak.  
N/A - A video will be provided if requested. The application builds and runs locally.

- [X] The Flatpak ID follows all the rules listed in the Application ID requirements.

- [X] I have read and followed all the Submission requirements and the Submission guide and I agree to them.

- [X] I am an developer to the project.


CodedOre:
So, you have:

- Merged the `master` branch into your submission
- Not given all information requested in the template
- Used an end-of-life runtime

… all not great looks in themselves. 

But, based on the description, it seems the application would only work on Arch Linux and probably needs deep system integration, both of which are not permissible for Flathub. 

And if all that were not enough, the manifest does not request the permissions it would need to interact with the system. Just to be clear, its not like you would be permitted to use these permissions. But it shows to me that you didn't even test if the Flatpak actually works. 

Stop trying to submit slop.

PR #8542, opened 2026-05-02
Add org.community.CopilotWebApp

psbanerjee:
CopilotWebApp is a community-maintained desktop application that provides a dedicated window for Microsoft Copilot by running it in Chromium app mode. It seamlessly integrates with the Linux desktop, offering easy installation via Flathub. This is an unofficial package and not affiliated with Microsoft.


CodedOre:

Also, your "application" is just a shortcut to launch Chromium. \
That's not an application, why are you trying to submit that?

We don't appreciate joke PRs.

bbhtt:
Blocked for repeatedly doing slop PRs

PR #8557, opened 2026-05-04
Add app.deepink.Deepink

vitonsky:

[Deepink](https://deepink.app/) is an open-source, privacy-first note-taking application focused on security and user experience. Everything lives inside a single encrypted vault — your notes, files, attachments, and metadata — all unreadable to anyone but you.

https://github.com/user-attachments/assets/87391504-6888-4335-970c-9b885dfbf6c4

Maintainers to include
- vitonsky
- katsyuta


app.deepink.Deepink.yaml
```diff
@@ -0,0 +1,91 @@
...
+      # Install AppStream metadata
+      - install -Dm644 app.deepink.Deepink.metainfo.xml
+        /app/share/metainfo/app.deepink.Deepink.metainfo.xml
+
+    sources:
+      # ── Pre-built release artifact from GitHub ───────────────────────────
+      # flatpak-builder downloads and verifies the sha256 before building.
+      # Update both url and sha256 on every new release.
+      - type: archive
+        url: https://github.com/DeepinkApp/deepink/releases/download/v0.1.3/deepink-linux-x64-0.1.3.zip
```
hfiguiere, line 79: What happen on aarch64?

This should be built from source.

app.deepink.Deepink.desktop
hfiguiere, line 1: This file should be (submitted and managed) upstream.

app.deepink.Deepink.metainfo.xml
hfiguiere, line 1: This file should be (submitted and managed) upstream.

app.deepink.Deepink.yaml
vitonsky: @hfiguiere only x86_64 arch is currently supported. Should we update something? I have no experience in flatpak building.

---

We have no resources to setup the build from sources. It is difficult because we have step with build the WASM code for twofish encryption: https://github.com/DeepinkApp/deepink/tree/master/packages/twofish

We currently build all that stuff in docker, and cannot to upload artifacts to NPM because we do not trust the NPM infrastructure. We build both WASM and bindings manually or in our CI.

As i understand we cannot to run the docker via manifest. In case we would try to build WASM with no docker, it would require clang and all deps for that. I don't know how to test that.

If somebody in community (maybe you, @hfiguiere) are opened to help with the setup we could build in manifest.

If there are no volunteers, I would ask to approve the use CI artifacts in build as we do right now. If any additional security measures is needed, I'm open to implement that.

vitonsky:
@flathubbot @hfiguiere i see the error `finish-args-home-filesystem-access` in linter log https://github.com/flathub-infra/vorarbeiter/actions/runs/25324356389

We use file system to let user choose the vault directory, the sync state directory, and to import/export notes.

What is your recommendations about this error? Should I add some exception for linter or update the permissions?

CodedOre:
Use the FileChooser portal. Then you don't need a static permission.

app.deepink.Deepink.yaml
hfiguiere: > If somebody in community (maybe you, @hfiguiere) are opened to help with the setup we could build in manifest.

Why would I spend my free time doing something I don't care about using a tech stack I would never have chosen to use?

If the build system is so complicated that's a you problem, not a me problem.

vitonsky:
@CodedOre what is it and how it does work?

## The context

We use the file chooser dialog in our electron app when user choose where to save the exported file, and to pick the directory/archive to import.

User also can pick the directory with vault, the app will remember it, and will use that directory with no user interactions in future. In future we will add the support for the templates dir that will work the same way.

app.deepink.Deepink.yaml
vitonsky: @hfiguiere you don't have to. We have no problem with our build system.
We will not waste our time to configure build in flatpak.
Instead we will continue to use the CI artifacts.

vitonsky:
I've published the file https://deepink.app/.well-known/org.flathub.VerifiedApps.txt

CodedOre:
Using the FileChooser portal means to ask the system to provide the file picker, which is then run by the system outside the sandbox. The system then gives the application access to only the picked file or folder. 

In most frameworks, like GTK4, this should just happen transparently. And Electron should support it out of the box AFAIK.

app.deepink.Deepink.yaml
CodedOre: Just to remind you:

> All source available submissions **must** be built entirely from source code.

This is a documented requirement for all new submissions.

vitonsky:
@CodedOre i think that is not our case. That would work for a file choosing in user session, but as I understand it would not work in scenario where
- We show the file chooser to let user select file/directory
- User choose
- We remember that path and will use that directory/file in background now and in future sessions with no additional interactions with user

CodedOre:
> We remember that path and will use that directory/file in background now and in future sessions with no additional interactions with user

That is fully possible with the FileChooser portal, so that should not be grounds for an exception.

app.deepink.Deepink.yaml
vitonsky: @CodedOre as I said, we are opened to contributions. If anybody in community could help us to setup the build from source, we will accept that changes.

Otherwise we will not to build from source, because have no bandwidth for that.

If that is critical for Flathub, then maybe the Flathub does not fit for our app, and will not available be via Flathub. It is undesirable, because we want to make our app available for users on all platforms. But we are not ready to invest dozens of hours just to make it available on Flathub.

Maybe the compromise can be found here? We could rename our app for flathub and pretend we are not source available. Or maybe we can set some flag that our app is proprietary?

The mentioned rule are confuse me. Could you explain why source available submissions must be built from source code, but priprietary software like Spotify may be uploaded as a blob?

vitonsky:
@CodedOre is there any docs about how it does work?

CodedOre:
I assume the question is about the persistence of a file shared over the FileChooser portal?

The short version:
- You call the FileChooser portal
- The system shows the user the file picker
- Flatpak will then create a path under `/run/user/$UID/$DOCID/$FILE` and mount the selected file/folder in this path
- The portal returns that path to the application

You can check that files granted over the FileChooser portal are persistent with `flatpak documents --columns=all $APP_ID`.

The FileChooser portal has an [documentation](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.FileChooser.html), but your framework should already implement that transparently.

vitonsky:
@CodedOre I've just tried to build with no filesystem permission, it seems all general use cases works fine, so I delete the permission. Thanks for elaboration.

vitonsky:
bot, build

app.deepink.Deepink.yaml
CodedOre: > The mentioned rule are confuse me. Could you explain why source available submissions must be built from source code, but priprietary software like Spotify may be uploaded as a blob?

The rule is already there for some years, but AFAIK this was done to improve transparency and to give better scrutiny about what actually is run in a Flatpak. Think that was partly in wake of the xz story back then. 

Also, it ensures your application is actually build against the libraries present in the runtime. 

As for proprietary applications: \
It's different for them because we don't have much options there.

Building from source with a verifiable manifest is the rule now.

Also, while Hub may use strong words, he is right. We are volunteering reviewers with enough on our plate. We don't have the time to write or maintain your manifests for you.\
We can answer question in the Matrix rooms and on the Flathub Discourse to help, but if you don't want to put in the effort to build your own application according to the rules of our platform, why should we?

vitonsky:
I see the error `Error: 'manifest-unknown-properties' error found in linter manifest check. Details: Invalid property 'only-arches' in context 'manifest'`

Our app currently support only x86 & x64 arch. How I can claim it in manifest?
Is there some reference/docs for a flatpak manifest scheme?

vitonsky:
I found the config reference docs https://docs.flatpak.org/en/latest/flatpak-builder-command-reference.html
As I understand, we declare the supported arches just on module level

vitonsky:
bot, build

CodedOre:
> Is there some reference/docs for a flatpak manifest scheme?

[`man flatpak-manifest`](https://docs.flatpak.org/en/latest/flatpak-builder-command-reference.html#flatpak-manifest)

> Our app currently support only x86 & x64 arch. How I can claim it in manifest?

That would be done with the [`flathub.json` file](https://docs.flathub.org/docs/for-app-authors/maintenance#flathubjson).

**That being said: An build for aarch64 should exist, unless there is a very good reason. And since you should build from source anyway, I don't immediately see one.**

app.deepink.Deepink.yaml
vitonsky: @CodedOre i understand you. As I said above, we have the decision on our side - we will not invest time to build from source for Flathub.

In case our package will be rejected due to this rule, we will setup our own repo and will upload the flatpak package there.

As I said previously, it is undesirable outcome of events, because it would hurt the UX. User would need to add custom repo. It would be **a step added by Flathub platform**, not by us, and we would directly say about that to our users.

@CodedOre Could you invite some decision maker here to understand what compromise can be found here. Who are the person who make final decision if publish or not to publish on Flathub?

As I know the [Flathub have a Spotify](https://flathub.org/en/apps/com.spotify.Client) that is obviously not an open source. Let's just add the "Proprietary" flag and will publish Deepink on Flathub.

vitonsky:
@CodedOre 
> An build for aarch64 should exist, unless there is a very good reason

We have no builds for aarch64. How I can configure manifest to make bot respect that fact and do not try to build aarch64?

app.deepink.Deepink.yaml
teamcons: > @CodedOre Could you invite some decision maker here to understand what compromise can be found here. Who are the person who make final decision if publish or not to publish on Flathub?

You are already talking with them. They are the manager. And the managers boss.


> As I know the [Flathub have a Spotify](https://flathub.org/en/apps/com.spotify.Client) that is obviously not an open source. Let's just add the "Proprietary" flag and will publish Deepink on Flathub.

It shouldnt be an issue for an open source app to be built from sources.

Note also that flathub has a lot of old submissions done when policies were more lax

teamcons:
> We have no builds for aarch64. How I can configure manifest to make bot respect that fact and do not try to build aarch64?

Thats in a flathub.json
CodedOre linked to the documentation. In the message before.

app.deepink.Deepink.yaml
```diff
@@ -0,0 +1,89 @@
...
+  - --device=dri
+
+  # ── Network ──────────────────────────────────────────────────────────────
+  # Needed for the built-in update-checker and any future sync features.
+  - --share=network
+
+  # ── D-Bus ────────────────────────────────────────────────────────────────
+  # Send desktop notifications (used by the Reminders feature).
+  # Also used to listen for the screen lock events and lock vault automatically
+  - --talk-name=org.freedesktop.Notifications
```
teamcons, line 32: Use the portal for notifications you do not need this

app.deepink.Deepink.yaml
teamcons: > @CodedOre i understand you. As I said above, we have the decision on our side - we will not invest time to build from source for Flathub.
>

The flathub requirements are only good practice requirements. Building from a verifiable source is a good requirement and in line with doing apps people can trust

Which is what you want if you sell your app as privacy focused.

The same argument for portals and strict static permissions.

bbhtt: I simplified it for you by closing and thus rejecting it. You can publish wherever you want outside of here.

In the future, if you can't follow the laid out requirements, then don't open the submission PR. This argument is a waste of time on every end.

PR #8559, opened 2026-05-04
Symbolical app submission

pracedru:
### Please confirm your submission meets all the criteria

- [X] Please describe the application briefly. Symbolical is a math document application inspired by MathCAD and SMath.
- [X] Please attach a video showcasing the application on Linux using the Flatpak. 

https://github.com/user-attachments/assets/7d372086-f91d-4198-995c-5e0399069ee3

- [X] The Flatpak ID follows all the rules listed in the [Application ID requirements][appid].
- [X] I have read and followed all the [Submission requirements][reqs] and the [Submission guide][reqs2] and I agree to them.
- [X] I am a developer and contributor to the project.

[appid]: https://docs.flathub.org/docs/for-app-authors/requirements#application-id
[reqs]: https://docs.flathub.org/docs/for-app-authors/requirements
[reqs2]: https://docs.flathub.org/docs/for-app-authors/submission

PR #8560, opened 2026-05-04
Add dk.pracedru.Symbolical

pracedru:

- [X] Please describe the application briefly. Symbolical is a math document application inspired by MathCAD and SMath.
https://github.com/user-attachments/assets/149000c7-cd3e-4fa4-a8d8-6ee3b950ef92
- [X] I am an  author contributor to the project.


hfiguiere:
dupe of #8559

pracedru:
I fixed the PR name though.

pracedru:
> dupe of #8559

The manifest is also located correctly I think.
I fixed the pr name like it said. How is that spam?

dk.pracedru.Symbolical/python3-setuptools.json
```diff
@@ -0,0 +1,6 @@
+{
```
ghost: Empty, redundant.

dk.pracedru.Symbolical/dk.pracedru.Symbolical.yml
```diff
@@ -0,0 +1,62 @@
...
+      - install -Dm644 dk.pracedru.Symbolical.metainfo.xml /app/share/metainfo/dk.pracedru.Symbolical.metainfo.xml
+    sources:
+      - type: file
+        path: dk.pracedru.Symbolical.desktop
+      - type: file
+        path: dk.pracedru.Symbolical.xml
+      - type: file
+        path: icon-256.png
+      - type: file
+        path: NotoSerif.ttf
```
ghost: Missing license file.

pracedru: I'll fix that, and I'll also put the one for pyqt

dk.pracedru.Symbolical/python3-setuptools.json
pracedru: Yeah Ill remove it

teamcons:
> > dupe of #8559
> 
> The manifest is also located correctly I think. I fixed the pr name like it said. How is that spam?

The submission guide says explicitly it isnt necessary to open a new PR instead of just fixing it. So it isnt super liked

dk.pracedru.Symbolical/.gitignore
teamcons, line 1: Why is this here?

dk.pracedru.Symbolical/app_icon.svg
teamcons, line 1: This should be upstream

dk.pracedru.Symbolical/dk.pracedru.Symbolical.desktop
teamcons, line 1: This should be upstream

dk.pracedru.Symbolical/dk.pracedru.Symbolical.metainfo.xml
teamcons, line 1: This should be upstream

dk.pracedru.Symbolical/dk.pracedru.Symbolical.xml
teamcons, line 1: This should be upstream

dk.pracedru.Symbolical/dk.pracedru.Symbolical.yml
```diff
@@ -0,0 +1,62 @@
...
+base: com.riverbankcomputing.PyQt.BaseApp
+base-version: '6.10'
+command: symbolical
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --filesystem=xdg-documents
+  - --filesystem=xdg-pictures
```
teamcons: Why do yiu need both filesystem permissions instead of using portals?

dk.pracedru.Symbolical/index.theme
teamcons, line 1: This should be upstream

dk.pracedru.Symbolical/launcher.sh
teamcons, line 1: Okay you get the idea. All the stuff should come with the sources. It makes no sense to have half the repo follow the manifest instead of just getting upstream and calling a buildsystem

pracedru:
> > > dupe of #8559
> > 
> > 
> > The manifest is also located correctly I think. I fixed the pr name like it said. How is that spam?
> 
> The submission guide says explicitly it isnt necessary to open a new PR instead of just fixing it. So it isnt super liked

Oh. my bad. It says Merging is blocked, Cannot change this locked branch. I thought i had to close the pr to free this branch from being locked. This is my first attempt at publishing to flathub. Sorry about that.

dk.pracedru.Symbolical/.gitignore
pracedru: I was testing the build of the flatpak and i didn't the output form that to be added to this repo.

dk.pracedru.Symbolical/app_icon.svg
pracedru: So, does this mean that this should be in my project repo?

dk.pracedru.Symbolical/dk.pracedru.Symbolical.yml
pracedru: I don't know. This is the first time I am trying to publish to flathub. Are Portals better?

dk.pracedru.Symbolical/launcher.sh
pracedru: aahh. ok. So this repo just needs the manifest file and everything else should be in my project repo. I will work on that.. 
So .. I don't close this pr right?
I just git rm the files from this branch, put them in my project repo, correct the manifest accordingly and push these changes?

pracedru:
> This pull request is temporarily marked as blocked as some automated checks failed on it. Please make sure the following items are done:
> 
> * Flatpak manifest is at toplevel
> * The [requirements](https://docs.flathub.org/docs/for-app-authors/requirements) and [submission process](https://docs.flathub.org/docs/for-app-authors/submission) have been followed

What does this mean "Flatpak manifest is at toplevel" ?
Isn't it supposed to be in the folder ?

bbhtt:
It's supposed to be in the folder when you are slopping your way through the PR. Don't bother with a second slop submission not following requirements

dk.pracedru.Symbolical/dk.pracedru.Symbolical.yml
CodedOre: Portals are always better than static permissions.

dk.pracedru.Symbolical/dk.pracedru.Symbolical.yml
```diff
@@ -0,0 +1,62 @@
...
+  - --socket=pulseaudio
+  - --env=QT_AUDIO_BACKEND=pulseaudio
+  - --socket=cups
+modules:
+  - name: desktop-entry
+    buildsystem: simple
```
CodedOre: This module should be merged with the application module.

dk.pracedru.Symbolical/dk.pracedru.Symbolical.yml
```diff
@@ -0,0 +1,62 @@
...
+      - type: file
+        path: app_icon.svg
+      - type: file
+        path: index.theme
+      - type: file 
+        path: dk.pracedru.Symbolical.metainfo.xml
+  - name: mime-update
+    buildsystem: simple
+    build-commands:
+      - update-mime-database /app/share/mime
```
CodedOre: This should not be necessary. 
And even if that be the case, it should be an `post-install` instead of its own module.

dk.pracedru.Symbolical/dk.pracedru.Symbolical.yml
```diff
@@ -0,0 +1,62 @@
...
+  - name: symbolical
+    buildsystem: simple
+    build-commands:
+      - install -d /app/symbolical
+      - cp -r * /app/symbolical/
+      - install -Dm755 launcher.sh /app/bin/symbolical
```
CodedOre: Please use something like meson or a setup.py to have an consistent buildsystem instead of putting all commands in here.

bbhtt:
Why was this reopened?

CodedOre:
@bbhtt, see https://discourse.flathub.org/t/publishing-my-app-to-flathub/12094/2

dk.pracedru.Symbolical.yml
```diff
@@ -0,0 +1,39 @@
...
+base-version: '6.10'
+command: symbolical
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --socket=pulseaudio
+  - --env=QT_AUDIO_BACKEND=pulseaudio
+  - --socket=cups
```
teamcons, line 15: why do you need cups permission? Is the portal not enough?

hfiguiere:
> @bbhtt, see https://discourse.flathub.org/t/publishing-my-app-to-flathub/12094/2

Well, no.

Whining doesn't bring any favour.

barthalion:
@hfiguiere @bbhtt The checklist followed requirements, as for everything else, maybe have some more empathy for people doing it for the first time. If there are no obvious signs of LLM misuse, just unsubscribe if you are not interested in reviewing, this sort of abrasive behaviour is not doing any favours to Flathub as a project.

bbhtt:
There were absolutely signs of LLM (mis)usage that weren't coming from a legit "first timer". I am empathetic for the first attempt, not when the same problem gets repeated. 

Allowing AI slop producers to roam free and making sure everyone gets burnt out, parotting the exact same thing to them, every day, as they fail to repeatedly read documentation and follow the most basic requirements, is not going to do any good to the project either. I can apply the exact same "just stop caring" advice to you too.

The discourse post has little value to me, as everyone in the history of producing AI slop has always said the exact same thing, asked for forgiveness and/or cried injustice.

pracedru:
> The domain to be used for verification is pracedru.dk. If you intend to [verify](https://docs.flathub.org/docs/for-app-authors/verification) this submission, please confirm by uploading an empty `org.flathub.VerifiedApps.txt` file to https://pracedru.dk/.well-known/org.flathub.VerifiedApps.txt. Otherwise, ignore this. Please comment if this incorrect.

I have cleared the file in https://pracedru.dk/.well-known/org.flathub.VerifiedApps.txt

dk.pracedru.Symbolical/dk.pracedru.Symbolical.yml
pracedru: > Portals are always better than static permissions.

I have tested this and while runs there is some funky behavior of the save file dialog in some distros, there is some actual non functional behavior on my Arch Gnome box. Everything works as intended when i run it outside the flatpak container, but inside the container, The Save As dialog looks like this:
image

The intended default folder ends up in the filename. If i try going into the documents folder and save a document with some name, nothing happens.
If i run it outside the container i get a different saveas dialog
image
Its still a GTK dialog but here it works ok.
I think maybe i did something wrong in my code
```python
	def on_save_as_clicked(self):
		file_ext = "Symbolical Files (*.sbl)"
		dir = str(self._document_view.document.path)

		dl = QStandardPaths.StandardLocation.DocumentsLocation
		dir = QStandardPaths.standardLocations(dl)[0]
		fd = QFileDialog(self)
		name = fd.getSaveFileName(self, 'Save File', filter=file_ext, directory=dir)[0]
		if name == "":
			return
		fi = QFileInfo(name)
		self._document.name = fi.fileName()
		self._document.path = fi.path()
		self.parent().update_title()
		self._controller.save(self._document)
		return True
```
But I looked in the documentation on that file dialog and it looks right to me:
[Dokumentation Here](https://doc.qt.io/qt-6/qfiledialog.html#getSaveFileName)

pracedru: This also worked fine with the solution with:
  - --filesystem=xdg-documents
  - --filesystem=xdg-pictures

dk.pracedru.Symbolical.yml
pracedru: I didn't know that the portal would give printer access.. 
Since i have issues with the save dialog on portals, ill just do some thorough testing of different distros and desktop environments.

dk.pracedru.Symbolical/dk.pracedru.Symbolical.yml
pracedru: I have made a meson solution for all of this.

dk.pracedru.Symbolical/index.theme
pracedru: done

dk.pracedru.Symbolical/.gitignore
pracedru: i removed it

dk.pracedru.Symbolical/dk.pracedru.Symbolical.yml
pracedru: Licenses has been added both in the app about dialog and in the file system under
image

dk.pracedru.Symbolical/python3-setuptools.json
pracedru: This file has been removed

dk.pracedru.Symbolical/dk.pracedru.Symbolical.yml
pracedru: I have also tested on Fedora Workstation. Here the result is the same as on my Arch Gnome VM
On Fedora Plasma 44 i looks like this :
image
The behavior is the same, but the save dialog works and a document is serialized to a file.

dk.pracedru.Symbolical/dk.pracedru.Symbolical.yml
pracedru: I have added it all to a meson build script.

pracedru:
> ✅ [Test build succeeded](https://github.com/flathub-infra/vorarbeiter/actions/runs/25415466975). To test this build, install it from the testing repository:
> 
> ```
> flatpak install --user https://dl.flathub.org/build-repo/281454/dk.pracedru.Symbolical.flatpakref
> ```
> 
> _Built for aarch64 and x86_64 architectures._

So far i think the only issue is with the save dialog through the portals. Everything else seems to work.

dk.pracedru.Symbolical/dk.pracedru.Symbolical.yml
pracedru: On LMDE the behavior is the same and no file is saved.
image

I suppose this is a bug in the QT Portals abstraction for file dialogs maybe... 
Not an issue that is easy for me to fix here. Maybe i can go with the :
  - --filesystem=xdg-documents
  - --filesystem=xdg-pictures
  for now and when this bug is resolved i can go back to Portals.

pracedru:
Oh the issue with the portals is marked as oudated now.
https://github.com/flathub/flathub/pull/8560#discussion_r3185379126
I put the documentation of the portals issue  there.. Do i need to copy it up here?

dk.pracedru.Symbolical/dk.pracedru.Symbolical.yml
CodedOre: You should at least try to find out where the issue exactly is. If it truly is an bug with Qt or the portals, there should be an upstream ticket about it. 

But you should take another look at it anyways. Right now, you don't recommend an filename and extension to the portal as well.

dk.pracedru.Symbolical.yml
```diff
@@ -0,0 +1,22 @@
...
+modules:
```
CodedOre, line 18: You will need to add the cleanup command for the base app.

dk.pracedru.Symbolical/dk.pracedru.Symbolical.yml
pracedru: I will look further into this. Create a ticket if it needs be.

pracedru: Ok I made a small test with Qt and C++ just to be better able to debug it and find the cause.. But this actually worked. My build env has Qt 6.11 though. So maybe this issue has been solved between 6.10 and 6.11.
image

pracedru: Oh well .. I tried building this on an VM with LMDE 7. It has an older Qt version 6.8.
But it also worked fine there. So maybe this is an issue with PyQt. I will keep digging.

pracedru: Ok so i tested this with PyQt on my local machine with forcing Qt to use Portals for file dialog and it also worked. 
image
My current Pythen environment which is just the system env has PyQt 6.10 just like the flatpak package.
So.. It seems like this issue is somehow related to the flatpak env.. 
Is this something you guys can help me debug?
I know there is a way to enter into a shell in the flatpak container.

dk.pracedru.Symbolical.yml
pracedru: I have tested without the  - --socket=cups and there is no printers in the list of the print dialog.

pracedru:
I added the document name to the saveas dialog as you mentioned here https://github.com/flathub/flathub/pull/8560#discussion_r3195030256  and strangely the behavior changed.. But it might explain something.
image
Its in danish sorry about that. But it says it can't find the folder /app/symbolical/src. 
Its a bit strange because i am not supplying that folder to the dialog. However I am creating the dialog before i submit the path. 
Maybe the dialog tries to resolve the flatpak path before showing the window.

pracedru:
Ok. I think i have finally figured it out.
My code now looks like this:

``` 
	def on_save_as_clicked(self):
		file_ext = "Symbolical Files (*.sbl)"
		dir = str(self._document_view.document.path)
		if dir == "":
			dl = QStandardPaths.StandardLocation.DocumentsLocation
			dir = QStandardPaths.standardLocations(dl)[0]

		fd = QFileDialog(self)
		fd.setAcceptMode(QFileDialog.AcceptMode.AcceptSave)
		fd.setNameFilter(file_ext)
		fd.setDirectory(dir)
		fd.selectFile(self._document.name)
		fd.setWindowTitle("Save File")

		if fd.exec() != QDialog.DialogCode.Accepted:
			return False
		fi = QFileInfo(fd.selectedFiles()[0])
		self._document.name = fi.fileName()
		self._document.path = fi.path()
		self.parent().update_title()
		self._controller.save(self._document)
		return True
```
This also works outside flatpak but inside flatpak i get this:

image

Inside the Flatpak container Qt resolves QStandardPaths.StandardLocation.DocumentsLocation to an english locale version of the user documents folder. Since this folder doesn't exist on my computer it fails.. Why exactly it would then try the folder of the executable i can't really explain. Does Flatpak not expose locale information to its apps? Do i need to set something in the manifest to have the locale settings evailable for the app or?

pracedru:
yes the portal does not give QT access to:
~/.config/user-dirs.dirs
this is why Qt resolves those paths wrongly.

Can i add add a static permission to this particular file?

I've tested that this  works:
finish-args:
  - --filesystem=xdg-config/user-dirs.dirs
  
  But is it ok?

dk.pracedru.Symbolical.yml
pracedru: Also i added the cleanup command.

pracedru:
ok lets try this:

bot, build

pracedru:
ok the bot fails the package on the:
--filesystem=xdg-config/user-dirs.dirs

What is the recommended action to take to resolve this problem?

CodedOre:
No, adding access to `xdg-config/user-dirs.dirs` is a workaround, not a solution. And access to the users config folder is not permitted for good reason. 

Maybe ask other Qt application developers, like in the KDE Discourse, if they know a solution. \
I can point out the issues, but Qt is not my expertise.

pracedru:
> No, adding access to `xdg-config/user-dirs.dirs` is a workaround, not a solution. And access to the users config folder is not permitted for good reason.
> 
> Maybe ask other Qt application developers, like in the KDE Discourse, if they know a solution. I can point out the issues, but Qt is not my expertise.

I could make a workaround my self if i knew a way to get these Locale specifik informations inside the flatpak container.. Surely there must be a documented way to get this information for applications in Flatpak containers. With all the apps on Flathub there has got to be someone else with this particular problem.

pracedru:
I can see that this app io.gitlab.luiggircardoso.Draft got access to --filesystem=xdg-documents 
He doesn't even have a saveas or save dialog.
Why is it ok for that app to have that access?

I am searching for solutions to my problem that other people have made for document based applications. And I just find that they get access to the folders. Here is another one:

https://github.com/ilshat-ishdavletov/flathub/blob/add-manuscript/io.gitlab.ilshat_apps.manuscript.yml

CodedOre:
**Other apps don't serve as precedent.** Each submission is judged on their own merit. 

But, just for the sake of argument: \
I assume the case for Draft is because it is an specific access without file picker that got the permission. And the other is a Markdown app, which needs access to neighboring files which is not possible with portals alone. 

Your app has single files the user may choose to store where they like, so a static filesystem permission will not circumvent the issues we've seen here, unless you go full sandbox-break with `home` or `host`.

The application already works with portal access alone. So, I'd rather have you fix this remaining issue rather than open up the sandbox as a workaround.

pracedru:
> io.gitlab.luiggircardoso.Draft

I would love to fix the issue. How do i get access to the data in xdg-config/user-dirs.dirs so i can know what the path to the  documents folder is?

pracedru:
What about --filesystem=xdg-config/user-dirs.dirs:ro?
That is a very limited access to the data i need.

CodedOre:
Just a question: Why the need to have the documents folder in particular as the starting folder?

The chances are very likely the user will store it in another place or in a subfolder anyways. You can't even necessarily assume that the documents folder has not been moved to another disk, or even network storage anyways. May not always be the norm, but its something that the XDG user specification does allow. 

I personally wouldn't do many assumptions here and just let the user decide where they want to store the file. Most applications AFAIK just originally put the user out on the home directory. \
If you want to be fancy you could use the `GetHostPaths` API from the documents portal to get the latest used folder for subsequent runs.

pracedru:
> That is a very limited access to the data i need.

This is a completely standardized way to handle file dialogs in document view applications.. 
If i open Writer and press save i get a save file dialog like this:
image

pracedru:
In the case of the app io.gitlab.luiggircardoso.Draft you granted access to the entire documents folder. Read and Write. Only to write to a single file ~/Documents/draft.txt. 
So talk about opening up more than is needed. I just need read only access to one single file ~/.config/user-dirs.dirs. That seems much more reasonable to me.. 

> No, adding access to `xdg-config/user-dirs.dirs` is a workaround, not a solution. And access to the users config folder is not permitted for good reason.
 
I beg to differ. Qt expects that file to be there.  I am not asking for permission to any folder, just read only access to one single file.

This is not even uncommon like here TeamSpeak gets access to an entire folder in .config/TeamSpeak:

image

pracedru:
Is this a bug or missing feature in flatpak?
I see that flatpak populates this folder 
/home/magnus/.var/app/dk.pracedru.Symbolical/config
with an empty file named user-dirs.dirs

And that that file has the path information in it, if one gets access to say xdg-documents or xdg-pictures.
Is this why you are so reluctant to do this?
Because that file is supposed to be filled with the folders from the portal at some point?

CodedOre:
The main point is: Any access to a host config folder would require a linter exception to be added. 

And I don't see that the benefit of starting in Documents rather than in home is that great of a benefit that it would warrant an exception. \
Sandbox-escape and exceptions should be for cases where not having these would restrict a important functionality. 

Maybe discussion about a better solution could be done. But this is not really actionable for this submission then and should be better discussed elsewhere.

pracedru:
> The main point is: Any access to a host config folder would require a linter exception to be added.
> 
> And I don't see that the benefit of starting in Documents rather than in home is that great of a benefit that it would warrant an exception. Sandbox-escape and exceptions should be for cases where not having these would restrict a important functionality.
> 
> Maybe discussion about a better solution could be done. But this is not really actionable for this submission then and should be better discussed elsewhere.

I'm sorry i got side tracked on this. 
Its not just the starting location actually .. One can't even save a document without access to xdg-documents.

CodedOre:
> Its not just the starting location actually .. One can't even save a document without access to xdg-documents.

I can't reproduce your claim on my test system though. It does save the file here.

Drop the directory change from the save file picker and use the working directory. If you then also not change that working directory in your launch script (you should not need `cd /app-dir; exec python app`, just `exec python /app-dir/app` also works), the FileChooser portal will always start in a valid home folder.

Anyways, unless you can give verifiable issues that are known upstream, I don't see why a filesystem permission is needed.

pracedru:
> > Its not just the starting location actually .. One can't even save a document without access to xdg-documents.
> 
> I can't reproduce your claim on my test system though. It does save the file here.
> 
> Drop the directory change from the save file picker and use the working directory. If you then also not change that working directory in your launch script (you should not need `cd /app-dir; exec python app`, just `exec python /app-dir/app` also works), the FileChooser portal will always start in a valid home folder.
> 
> Anyways, unless you can give verifiable issues that are known upstream, I don't see why a filesystem permission is needed.

Ok.
I will make those changes to the launcher.
I will also accept that the xdg information of the user folders isn't available.
I found that on my system it actually does save a new file, but the filename is garbled up a bit. I think maybe its because I ensure that the file extension is added to the filename even if the user doesn't write it in the saveas dialog. I will see if i can get the dialog through the portal to enforce the file extension.

pracedru:
I thing this should do it:

bot, build

pracedru:
This application should work on any architecture that has com.riverbankcomputing.PyQt.BaseApp. 
Do i need to add something to the manifest to get it on say Arm also?

pracedru:
Unfortunately during testing i found a regression in the about dialog. so:

bot, build

dk.pracedru.Symbolical.yml
```diff
@@ -0,0 +1,24 @@
...
+  - --socket=cups
+cleanup-commands:
+  - /app/cleanup-BaseApp.sh
+modules:
+  - name: symbolical
+    buildsystem: meson
+    sources:
+      - type: git
+        url: https://gitlab.com/symbolical/symbolical.git
+        commit: 0d406b86
```
CodedOre: It would be better to use the full commit hash.

CodedOre:
> This application should work on any architecture that has com.riverbankcomputing.PyQt.BaseApp. 
> Do i need to add something to the manifest to get it on say Arm also?

Flathub does not provide builds for 32-bit architectures anymore. Your app is already build for all supported architectures, which are x86_64 and aarch64 (which is 64-bit ARM btw).

dk.pracedru.Symbolical.yml
pracedru: Done

pracedru:
AAHH.. I am just testing it now on my laptop with F44 Workstation. And it can't save files :(
It works if the user manually writes the file name with the extension. Apparently the file dialog portal on F44 doesn't support enforcing the file extensions on the filename. 

I don't really understand it because my desktop runs Arch Gnome and my laptop runs Fedora Gnome and they are both v50. So the portal should be the same right? Also.. They (the save dialog) look very similar.

no ok. it also fails on my desktop.. 
I had setup my debug env to force qt to use portals.. And that actually worked. In the debug env i get the portal dialog (gtk4) which is different from the normal dialog (gtk3). But apparently the file creation is slightly different in my debug env compared to the flatpak container env .

pracedru:
No wait.. It does save the file it still just garbles up the filename 
image

Despite that it actually gets this name from the portal dialog:
/run/user/1000/doc/ec6cc0c/New Document.sbl

So. I don't know what the solution to this could be.

CodedOre:
I can't test this right now, but I had the thought it might be related to you renaming the file you get from the portal. But I can't verify that for now. 

Maybe ask in the [XDG Desktop Portal Matrix room](https://matrix.to/#/!FEQDCDpkIktOBQNdgv:matrix.org?via=matrix.org&via=gnome.org&via=kde.org) if someone there knows what causes this. 

In any case, it would be better to suggest a proper filename with extension to the portal instead of trying to rename it afterwards.

pracedru:
I changed that. I now set the dialog to enforce the file extension. And like you can see from the message above the file name from the file dialog is with extension.

I can confirm that it works on Fedora 44 with Plasma though.
So i suspect that it might be some difference in how the GTK and KDE portals work maybe.

pracedru:
This looks like it could be a bug in Qt's portal abstraction code for dialogs. Even if I find the cause it might take a long time before the fix released for the pyqt libs that my app use.

pracedru:
Ok talking to the people on the Desktop Portals Matrix room, i realized that i have a viable, though not very nice, temporary solution to the problem until the upstream fixes are available.
I can make sure that the initial filename has the correct extiension in the dialog. So the users will see it. Then i will remove the extension enforcement. Then if the user types in a new name without the right extension and presses save, I can halt the process and show an error message with : "the filename must include the extension .sbl"
This is a terrible UX but at least my app wont fail silently.

pracedru:
There is done. Everything should be ok now.
ok so now i made it so if the user types a name without extension they get this:
image
it then opens a new save dialog with 
image
And they can press save.
Pretty clonky solution.. But it works.
Then eventually when all this is solved upstream and lands in the flatpak repos.. Ill remove it.

bot, build

CodedOre:
> There is done. Everything should be ok now.
> ok so now i made it so if the user types a name without extension they get this:
> image

I think your error message tells it the wrong way around...

pracedru:
turns out this issue has been reported 4 years ago:
https://qt-project.atlassian.net/browse/QTBUG-91091

pracedru:
> I think your error message tells it the wrong way around...

Oh yeah i noticed.
That image is outdated. My bad.
its fixed already.
image

pracedru:
So there are three bugs related to all this.
https://github.com/flatpak/flatpak/issues/6651
https://github.com/flatpak/xdg-desktop-portal/issues/2000
https://qt-project.atlassian.net/browse/QTBUG-91091

Until these are fixed and land in flatpak, this above solution will be good i think.
you can try it out in the app..
You have to consciously delete the extension to even trigger this.

CodedOre:
Yes, the system as it is now works for the time. 
Let's hope that the Qt bug can be resolved upstream soon.

CodedOre:
bot, build

dk.pracedru.Symbolical.yml
```diff
@@ -0,0 +1,24 @@
...
+  - --socket=cups
+cleanup-commands:
+  - /app/cleanup-BaseApp.sh
+modules:
+  - name: symbolical
+    buildsystem: meson
+    sources:
+      - type: git
+        url: https://gitlab.com/symbolical/symbolical.git
+        commit: 1fb8b7e655a202ccb603c533efc3b063e00cb655
```
CodedOre, line 24: This commit does not track a release currently, which it should according to the stable releases policy. 

Also, while optional, having the Tag additionally would be good as well in my view. The commit hash is mandatory.

pracedru:
Tag v0.5.0.5 added.

bot, build

PR #8564, opened 2026-05-05
Add io.github.kyuyrii.meganimus

Kyuyrii:

- [x] Please describe the application briefly. < A shortcut creator for games >
- [x] Please attach a video showcasing the application on Linux using the Flatpak. < https://github.com/user-attachments/assets/7268b2b7-7527-44f8-906e-c1596e4fc465 >

### DIFF
--- a/io.github.kyuyrii.meganimus.yaml
+++ b/io.github.kyuyrii.meganimus.yaml
@@ -21,7 +21,6 @@ finish-args:
 
 modules:
   - name: Meganimus
-    builddir: true
     buildsystem: simple
     sources:
       - type: git
--- a/io.github.kyuyrii.meganimus.yaml
+++ b/io.github.kyuyrii.meganimus.yaml
@@ -5,6 +5,11 @@ runtime-version: '6.10'
 command: meganimus
 base: com.riverbankcomputing.PyQt.BaseApp
 base-version: '6.10'
+cleanup-commands:
+  - /app/cleanup-BaseApp.sh
+build-options:
+  env:
+    - BASEAPP_REMOVE_WEBENGINE=1
 
 finish-args:
   - --share=network
--- a/io.github.kyuyrii.meganimus.yaml
+++ b/io.github.kyuyrii.meganimus.yaml
@@ -3,7 +3,6 @@ sdk: org.kde.Sdk
 runtime: org.kde.Platform
 runtime-version: '6.10'
 command: meganimus
-separate-locales: false # translations
 base: com.riverbankcomputing.PyQt.BaseApp
 base-version: '6.10'
 
--- a/io.github.kyuyrii.meganimus.yaml
+++ b/io.github.kyuyrii.meganimus.yaml
@@ -25,8 +25,8 @@ modules:
     sources:
       - type: git
         url: https://github.com/Kyuyrii/Meganimus.git
-        tag: '4.1'
-        commit: b8e88043a1232342dda089f1a1498a25c859a773
+        tag: '4.2'
+        commit: 0cc25a694abd3894f3b2494bd2110128cbcc25b4
     build-commands:
       - install -Dm755 meganimus /app/bin/meganimus
       - install -Dm644 io.github.kyuyrii.meganimus.desktop /app/share/applications/io.github.kyuyrii.meganimus.desktop
--- a/io.github.kyuyrii.meganimus.yaml
+++ b/io.github.kyuyrii.meganimus.yaml
@@ -29,8 +29,8 @@ modules:
     sources:
       - type: git
         url: https://github.com/Kyuyrii/Meganimus.git
-        tag: '4.2'
-        commit: 0cc25a694abd3894f3b2494bd2110128cbcc25b4
+        tag: '4.3'
+        commit: 6dfe638890f32db8498df4dba49abbe549523ca1
     build-commands:
       - install -Dm755 meganimus /app/bin/meganimus
       - install -Dm644 io.github.kyuyrii.meganimus.desktop /app/share/applications/io.github.kyuyrii.meganimus.desktop
--- a/io.github.kyuyrii.meganimus.yaml
+++ b/io.github.kyuyrii.meganimus.yaml
@@ -26,8 +26,8 @@ modules:
     sources:
       - type: git
         url: https://github.com/Kyuyrii/Meganimus.git
-        tag: '4.0'
-        commit: 53a202d2ec5f51a7aac9b594258875aa94a9ffe4
+        tag: '4.1'
+        commit: b8e88043a1232342dda089f1a1498a25c859a773
     build-commands:
       - install -Dm755 meganimus /app/bin/meganimus
       - install -Dm644 io.github.kyuyrii.meganimus.desktop /app/share/applications/io.github.kyuyrii.meganimus.desktop
--- a/io.github.kyuyrii.meganimus.yaml
+++ b/io.github.kyuyrii.meganimus.yaml
@@ -1,0 +1,39 @@
+app-id: io.github.kyuyrii.meganimus
+sdk: org.kde.Sdk
+runtime: org.kde.Platform
+runtime-version: '6.10'
+command: meganimus
+separate-locales: false # translations
+base: com.riverbankcomputing.PyQt.BaseApp
+base-version: '6.10'
+
+finish-args:
+  - --share=network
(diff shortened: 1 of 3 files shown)


io.github.kyuyrii.meganimus.yaml
```diff
@@ -0,0 +1,39 @@
+app-id: io.github.kyuyrii.meganimus
+sdk: org.kde.Sdk
+runtime: org.kde.Platform
+runtime-version: '6.10'
+command: meganimus
+separate-locales: false # translations
```
hfiguiere: why?

io.github.kyuyrii.meganimus.yaml
```diff
@@ -0,0 +1,39 @@
...
+  - --device=dri
+  - --filesystem=home
+  - --filesystem=/media
+  - --filesystem=/mnt
+  - --filesystem=/run/media
+  - --filesystem=~/.var/app/org.winehq.Wine
+
+modules:
+  - name: Meganimus
+    builddir: true
```
hfiguiere:
```suggestion
```

if there was a real buildsystem it could be useful.

io.github.kyuyrii.meganimus.yaml
Kyuyrii: I had forgotten to remove that part.

io.github.kyuyrii.meganimus.yaml
```diff
@@ -0,0 +1,37 @@
+app-id: io.github.kyuyrii.meganimus
+sdk: org.kde.Sdk
+runtime: org.kde.Platform
+runtime-version: '6.10'
+command: meganimus
+base: com.riverbankcomputing.PyQt.BaseApp
+base-version: '6.10'
+
```
hfiguiere, line 13: this is missing the baseapp cleanup https://github.com/flathub/com.riverbankcomputing.PyQt.BaseApp#example-pyqt-application

io.github.kyuyrii.meganimus.yaml
```diff
@@ -0,0 +1,37 @@
...
+  - --share=network
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --device=dri
+  - --filesystem=home
+  - --filesystem=/media
+  - --filesystem=/mnt
+  - --filesystem=/run/media
+  - --filesystem=~/.var/app/org.winehq.Wine
```
hfiguiere, line 24: why does it needs to access wine data?

Kyuyrii: For some reason, Wine Flatpak only respects the WINEPREFIX variable in the manifest, so the default Wine prefix is ​​inside the Wine Flatpak folder. If the user chooses an .exe file that is inside the Wine Flatpak folder, the Document Portal is used, making the obtained path different from the actual path.

io.github.kyuyrii.meganimus.yaml
Kyuyrii: I've now added cleanup-commands.

io.github.kyuyrii.meganimus.yaml
Kyuyrii: The permissions listed are only because .desktop files need to use the actual file and folder paths.

Document Portal, in addition to providing different paths, isolates the selected file, which completely breaks the execution of a game.

Kyuyrii:
I recorded this video to demonstrate the reason for the `--filesystem=` permissions.

As you can see, when selecting a file with Document Portal, the path obtained is different from the actual path.

In the case of a native game, this caused the script that starts the game to fail to find the game files, since the script is isolated in a path created by Document Portal.

In the case of a game running via Wine, Wine Flatpak couldn't find the path created by Document Portal, so it didn't even try to execute the .exe file from that path. The .exe file in question is inside the Wine Flatpak folder, meaning it has access, but it needs the actual path of the .exe file.

https://github.com/user-attachments/assets/ddbeef22-a445-480d-8a90-66b85f26d22e

Kyuyrii:
I created a pull request to add the necessary exceptions for Meganimus.

https://github.com/flathub-infra/flatpak-builder-lint/pull/1016

io.github.kyuyrii.meganimus.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+modules:
+  - name: Meganimus
+    buildsystem: simple
+    sources:
+      - type: git
+        url: https://github.com/Kyuyrii/Meganimus.git
```
bbhtt, line 31: It's blocked per https://docs.flathub.org/docs/for-app-authors/requirements#software-design-and-distribution-policy

Unless you are going to factor in the 4000 lines file into readable and manageable chunks so that I can sanely read and understand what is is about, it is going to be blocked.

You are already using LLMs to produce this, maybe it's time to do some of it by yourself.

Kyuyrii: Zordeer is larger than Meganimus, and this hasn't caused any maintenance problems; Zordeer has been around for months.

The code is like this because it's simpler that way.

Any code will be difficult to understand at first if whoever is modifying it isn't involved in the development.

For example, Lutris and Bottles are divided into many .py files with little code in each, dozens of files like that. This doesn't make them seem more organized to me; in fact, it makes it much more complicated for me to understand how to contribute.

bbhtt: > The code is like this because it's simpler that way.

It's like that because you let an LLM run through it.

Kyuyrii: > It's like that because you let an LLM run through it.

I test my apps, I've made several changes to them because I found problems, and they've been around for months.

But the code being a single file has never been a problem.

bbhtt: It has never been a problem because you are letting an LLM generate it. It's not sustainable and as I said I can't read through it.

I have to read some of it to know if the app is malicious insecure doing incorrect things, if I can't do that I can't accept it.

bbhtt: also the repository is missing the most basic expectations of a software project: having CI, most basic expectation of a Python project: having at least a linter and formatter in CI, some dependency lockfile or specification file, a buildsystem

None of this is a problem to you because an LLM is generating it.

Kyuyrii: Zordeer is available on Flathub and is larger than Meganimus, with the permissions `--filesystem=`home, /media, /mnt, and /run/media.

Zordeer is my app, just like Meganimus; neither is malicious. Zordeer was accepted on Flathub last month.

Meganimus accesses fewer files than Zordeer; it doesn't even try to run games or modify Wine prefixes, it just gets the path.

bbhtt: It's not my problem if some people are being overly empathetic to you and accepting submissions that are breaking the requirements. It doesn't change what is written there and what I am following.

It breaks every one of the points here:

> Projects must demonstrate reasonable structure, organization, maintainability, 

The criteria for each of those being what other established projects around your ecosystem are doing or what someone reasonable would think is reasonable.

Kyuyrii: > Projects must demonstrate reasonable structure, organization, maintainability,

I can understand the structure and organization, and it's simple for me to maintain this app.

Kyuyrii: Meganimus isn't malicious, based on what you said. It's possible to obtain the actual paths of files and folders without needing the `--filesystem=` permissions, which means that if I manage to add that to Meganimus, it wouldn't even be capable of doing anything malicious, since it wouldn't have free access to the user's files and folders.

I didn't try adding that because I didn't even know it was possible; I thought part of the idea of ​​the Document Portal was that apps shouldn't even have information about the actual paths of files and folders.

bbhtt: If you aren't going to solve any of the issues raised but rather repeat the same thing for the nth time I am going to lock this because I need some peace.

It will stay locked until I see something has improved. You can make changes to your repo as that is in your control.

bbhtt: Opening another PR will lead to a block, you've already run through my patience here in your previous PRs / discourse post.

Kyuyrii:
It will probably take a while before I can make any changes. My computer isn't working.

PR #8566, opened 2026-05-05
Add io.github.cushycush.wflow

cushycush:
Submitting wflow — a Wayland-native automation tool. Built it because Linux still doesn't have a real macOS Shortcuts equivalent: a GUI editor backed by plain-text workflow files, a global-hotkey daemon that fires those workflows on chord, and a community catalog at [wflows.io](https://wflows.io).

## What it does

wflow records, edits, and replays sequences of keystrokes, clicks, shell commands, delays, and notifications. Workflows are KDL files: diffable in git, hand-authorable in `$EDITOR`, shareable as a single file. Qt Quick GUI for visual editing, CLI (`wflow run`) for cron and scripts, both running the same Rust engine. Built on `wdotool-core` (linked in-process, not subprocessed) for input injection on Wayland.

## Permissions

The manifest carries `--talk-name=org.freedesktop.Flatpak` (host-spawn). Want to be upfront about this since it's the bit reviewers care most about.

**Why we need it.** Workflows ARE host commands. The whole point of `shell "git push"` or `shell "hyprctl dispatch exec slack"` is that wflow drives the user's session — an automation tool that can't reach the host shell is a viewer, not a runner. Without host-spawn, every tutorial workflow would silently fail in the sandbox.

**What we narrowed.** Notifications go through `org.freedesktop.portal.Notification`. Input injection goes through the portal's libei `RemoteDesktop` interface (the consent-dialog path). Recording uses the same RemoteDesktop portal. Only `shell` and `clipboard` (`wl-copy`) host-spawn. The remaining surface IS the user-facing feature; reducing it further would make the app non-functional.

Threat model and full surface analysis: https://github.com/cushycush/wflow/blob/main/REVIEW.md

**Precedents.** Same `--talk-name=org.freedesktop.Flatpak` pattern ships in:

- VS Code (`com.visualstudio.code`) — runs build tasks on the host
- GNOME Builder (`org.gnome.Builder`) — same use case
- GNOME Boxes (`org.gnome.Boxes`) — manages host-side VMs via qemu

## Local build verification

Built locally with `flatpak-builder` against the v1.0.2 tag on a Hyprland host. GUI launches, the shell action host-spawns correctly, the Record button shows a graceful "Record can't start" error (Hyprland's `xdg-desktop-portal-hyprland` doesn't ship the RemoteDesktop interface yet, so neither portal nor evdev fallback can attach inside the sandbox — expected behavior, the error message points at the fix on Plasma 6 / GNOME 46+).

Recording works end-to-end on Plasma 6 / GNOME 46+ where the RemoteDesktop portal is available.

## Source pinning

- `tag: v1.0.2`
- `commit: adfec45ef65506d15c69ac2b463a5765e3137947`
- `cargo-sources.json` regenerated from Cargo.lock at that commit (4752 entries)

Happy to iterate on anything.

### DIFF
--- a/cargo-sources.json
+++ b/cargo-sources.json
@@ -1,0 +1,4753 @@
+[
+    {
+        "type": "archive",
+        "archive-type": "tar-gzip",
+        "url": "https://static.crates.io/crates/adler2/adler2-2.0.1.crate",
+        "sha256": "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa",
+        "dest": "cargo/vendor/adler2-2.0.1"
+    },
+    {
+        "type": "inline",
+        "contents": "{\"package\": \"320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa\", \"files\": {}}",
+        "dest": "cargo/vendor/adler2-2.0.1",
+        "dest-filename": ".cargo-checksum.json"
+    },
+    {
+        "type": "archive",
+        "archive-type": "tar-gzip",
+        "url": "https://static.crates.io/crates/aho-corasick/aho-corasick-1.1.4.crate",
+        "sha256": "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301",
+        "dest": "cargo/vendor/aho-corasick-1.1.4"
+    },
+    {
+        "type": "inline",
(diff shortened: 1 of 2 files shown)


hfiguiere:

See also our AI policy

PR #8597, opened 2026-05-07
Add io.github.raspbeguy.Terrain

raspbeguy:

Terrain is a native GNOME GUI for Terraform and OpenTofu.

[Capture vidéo du 07-05-2026 18:02:35.webm](https://github.com/user-attachments/assets/8f34964c-d227-4d56-a326-1dfd8216019c)

- [x] I am the author to the project.


raspbeguy:
Just a small precision about additional permissions:

`--talk-name=org.freedesktop.Flatpak` is needed because the app's whole job is to run `tofu plan`, `apply`, etc. on behalf of the user. By default it shells out to whatever `tofu` or `terraform` binary the user has installed on the host, and optionally to `podman`/`docker` or `bwrap` if the user wants to pin a tool version per workspace or sandbox runs further. None of that is reachable from inside a Flatpak without `flatpak-spawn --host`.

Bundling those tools isn't really an option. Terraform is BUSL-1.1, so we can't redistribute it. OpenTofu is 80 MB+ per version and people pin different versions per project, so picking one would be wrong for everyone. And podman/docker/bwrap are large host components that users already have set up — duplicating them inside the bundle would be confusing and wasteful.

The mitigation is that Terrain has an opt-in "managed binary" mode: it downloads official OpenTofu / Terraform releases (verified against the upstream `_SHA256SUMS`), drops them in `$XDG_DATA_HOME/terrain/binaries/`, and runs them directly from inside the sandbox. No portal involved. Users who'd rather not punch a hole through the sandbox can switch every workspace to managed mode and the talk-name becomes dead weight.

`--filesystem=home` is the other one. Terraform projects live wherever the user keeps their code. `~/code`, `~/repos`, `~/work`, `~/Documents/projects/`, whatever. There's no portal that exposes "the user's source tree" generically, and `--filesystem=xdg-documents` only covers `~/Documents`, which isn't where most people keep IaC repos.

Terrain doesn't sweep the home dir; it only touches paths the user has explicitly registered as a project (the list is in `$XDG_CONFIG_HOME/terrain/config.toml`).

hfiguiere:
This is likely not suitable for flathub

https://docs.flathub.org/docs/for-app-authors/requirements#host-dependent-applications

raspbeguy:
I guess I can live without some features like launching inside containers and rely exclusively on fetching terraform/tofu binaries from the net, thus dropping `--talk-name=org.freedesktop.Flatpak`. What about `--filesystem=home`? Is that a no-go?

CodedOre:
If there is a good argument that warrants the use of `--filesystem=home`, you can [request an linter exception](https://docs.flathub.org/docs/for-app-authors/linter#exceptions).\
But it should only be used when using the portals are not an option.

On that note:

> Terrain doesn't sweep the home dir; it only touches paths the user has explicitly registered as a project (the list is in $XDG_CONFIG_HOME/terrain/config.toml).

I assume the user can set these projects over the UI? In that case, it might be possible to utilize the FileChooser portal to gain selective access to the project folder. You can request access to a folder, and the access is persistent across sessions. 

The document portal has some restrictions, but you should try if you can avoid the static filesystem permission if possible.

raspbeguy:
Oh right, so I don't need it, good news.

Okay so I'll just modify the file without those lines and we should be good

io.github.raspbeguy.Terrain.yml
```diff
@@ -0,0 +1,36 @@
...
+  - '*.la'
+  - '*.a'
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --filesystem=home
+  - --talk-name=org.freedesktop.secrets
```
CodedOre, line 27: Use the portal for the secrets and drop this permission.

raspbeguy: Terrain is written in Go and uses github.com/zalando/go-keyring, which talks D-Bus directly to org.freedesktop.Secret.Service. Sadly, the Go ecosystem doesn't have a portal-aware keyring library that I know of. Is it a blocking issue?

raspbeguy:
bot, build

> Oh right, so I don't need it, good news.
> 
> Okay so I'll just modify the file without those lines and we should be good

Actually I am going to have to do some tests. I'll probably do that tomorrow. Thanks for the review.

raspbeguy:
> If there is a good argument that warrants the use of `--filesystem=home`, you can [request an linter exception](https://docs.flathub.org/docs/for-app-authors/linter#exceptions). But it should only be used when using the portals are not an option.
> 
> On that note:
> 
> > Terrain doesn't sweep the home dir; it only touches paths the user has explicitly registered as a project (the list is in $XDG_CONFIG_HOME/terrain/config.toml).
> 
> I assume the user can set these projects over the UI? In that case, it might be possible to utilize the FileChooser portal to gain selective access to the project folder. You can request access to a folder, and the access is persistent across sessions.
> 
> The document portal has some restrictions, but you should try if you can avoid the static filesystem permission if possible.

I tested with `--filesystem=home` removed.

The folder picker side works as intended: `gtk.FileDialog.SelectFolder()` routes through the FileChooser portal, and the registered project path comes back as a document portal mount (`/run/user/$UID/doc/<id>/<name>`).

However, running a plan against the registered project fails with:

```
Failed to determine current working directory: getwd: no such file or directory
```

The cause: Terrain spawns `tofu` with `cmd.Dir` set to the document portal path. `chdir(2)` succeeds, but tofu then calls `os.Getwd()` during startup, and the kernel's `getcwd(2)` cannot resolve a path inside the xdg-document-portal FUSE mount (it walks `..` inode lookups that the portal doesn't expose). tofu aborts before doing any work.

This isn't fixable app-side: tofu is an external binary that requires a cwd with a resolvable parent-dentry chain, which the document portal intentionally doesn't provide.

So I'd like to request the linter exception as it's a hard requirement of the tool Terrain wraps. What's your opinion?

raspbeguy:
You know what, I'm going to get rid of `--filesystem=home` once and for all by changing my strategy. I'm working on a new release of Terrain that won't ask for an existing path to import a project, but for a git URL so that Terrain will clone the project in its own space, not needing any filesystem portal or any raw access to user files.

So let me finish this feature, I will tag it 0.0.2 and we'll review this submission again.

raspbeguy:
bot, build

io.github.raspbeguy.Terrain.yml
```diff
@@ -0,0 +1,35 @@
...
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --talk-name=org.freedesktop.secrets
+modules:
+  - name: terrain
+    buildsystem: meson
+    sources:
+      - type: archive
+        url: https://github.com/raspbeguy/terrain/archive/refs/tags/v0.0.2.tar.gz
```
bbhtt, line 33: what parts of this did you actually write by yourself and what parts did the LLM write through the process?

raspbeguy: I mostly used AI for the UI part.

bbhtt: was the readme part of the UI?

bbhtt: and it's just not the UI, I can tell just by the comments in your source code.

raspbeguy: Mostly doesn't mean exclusively. As english is not my native language I generated the readme and other doc and then reviewed it myself.

bbhtt: in other words it means "most of it"

raspbeguy: Yes. With human intelligence and review behind.

My intention isn't to lie about the fact that ai was highly involved in this. For the UI there is little to no human review though.

raspbeguy: Documentation says:

>Submissions or changes where most of the code is written by or using AI without any meaningful human input, review, justification or moderation of the code are not allowed.

I say that this doesn't belong to this category.

io.github.raspbeguy.Terrain.yml
bbhtt: you can ask your LLM to write one or contribute support upstream and do the ecosystem some good

raspbeguy: Could be an idea. But that won't happen quite soon though.

raspbeguy:
So what does the "blocked" tag mean? Is that something I can have an action on or does that mean that this submission fate is sealed?

io.github.raspbeguy.Terrain.yml
bbhtt: the only human input i can see is that you are backtracking from "mostly UI" and whatever this is:

> With human intelligence and review behind.

> For the UI there is little to no human review though.

is this not contradictory? is the UI not part of your app?

> I say that this doesn't belong to this category.

I say it does.

raspbeguy: Ok so is there something I can do at this point to unblock the submission?

PR #8607, opened 2026-05-08
Add io.github.ForWard_Technologies_LLC.Pylux

forward-technologies:

Pylux is a remote play and cloud play client for PS consoles
[Screencast from 2026-05-06 23-47-28.webm](https://github.com/user-attachments/assets/1976802b-43df-48ad-8c79-2d9f09d6f9e7)


forward-technologies:
bot, build

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,229 @@
+app-id: io.github.ForWard_Technologies_LLC.Pylux
+runtime: org.kde.Platform
+runtime-version: '6.9'
```
CodedOre: This is not the current version of the runtime.

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,229 @@
...
+sdk: org.kde.Sdk
+base: io.qt.qtwebengine.BaseApp
+base-version: '6.9'
+command: chiaki
+rename-icon: chiaking
+separate-locales: false
```
CodedOre: Why can't the locales be separated?

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,229 @@
...
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=all
+  - --allow=bluetooth
+  - --socket=pulseaudio
+  - --env=QTWEBENGINEPROCESS_PATH=/app/bin/QtWebEngineProcess
+  - --env=LD_LIBRARY_PATH=/app/lib:/app/lib64
```
CodedOre:
```suggestion
```

`/app/lib64` does not exists.

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,229 @@
...
+  - --env=DBUS_FATAL_WARNINGS=0
+  - --env=QML_DISABLE_DISK_CACHE=1
+  - --env=SDL_AUDIODRIVER=pipewire
+  - --talk-name=org.freedesktop.ScreenSaver
+  - --system-talk-name=org.freedesktop.login1
```
CodedOre: Why does this not use the inhibit portal?

forward-technologies:
For tracking, blocked by:  https://github.com/flathub-infra/flatpak-builder-lint/pull/1024

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,229 @@
...
+  - --filesystem=xdg-run/pipewire-0
+  - --filesystem=xdg-run/gamescope-0
+  - --filesystem=host-os:ro
+  - --filesystem=xdg-download
+  - --filesystem=xdg-documents
+  - --filesystem=~/.steam/steam
```
CodedOre: Why these filesystem permissions?\
Why is the FileChooser portal not used?

Also, what reason exists for access to `host-os`? What do you possible need to read from `/usr`?

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,229 @@
...
+
+cleanup-commands:
+  - /app/cleanup-BaseApp.sh
+
+build-options:
+  arch:
+    aarch64:
+      prepend-pkg-config-path: "/app/lib/pkgconfig:/app/lib64/pkgconfig:/app/lib/aarch64-linux-gnu/pkgconfig:/app/lib64/aarch64-linux-gnu/pkgconfig"
+    x86_64:
+      prepend-pkg-config-path: "/app/lib/pkgconfig:/app/lib64/pkgconfig:/app/lib/x86_64-linux-gnu/pkgconfig:/app/lib64/x86_64-linux-gnu/pkgconfig"
```
CodedOre:
```suggestion
```

Once again: `/app/lib64` is not a thing in the Flatpak runtime.

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,229 @@
...
+    cleanup:
+      - /include
+      - /lib/pkgconfig
+    sources:
+      - type: git
+        url: https://github.com/haasn/libplacebo
+        tag: v7.360.1
+        commit: cee9b076f2c63104ccfd497fa79c39a867293ec4
+
+  - name: ffmpeg
```
CodedOre: Ffmpeg is already in the runtime.

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,229 @@
...
+  - name: miniupnpc
+    buildsystem: cmake-ninja
+    subdir: miniupnpc
+    sources:
+      - type: git
+        url: https://github.com/miniupnp/miniupnp
+        tag: miniupnpc_2_3_0
+        commit: f39e4997591a98e7ee6711eabc8b50feeef1bd67
+
+  - name: SDL3
```
CodedOre: SDL3 is already in the runtime.

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,229 @@
...
+      - -DSDL_TESTS=OFF
+      - -DSDL_EXAMPLES=OFF
+      - -DCMAKE_INSTALL_LIBDIR=lib
+    sources:
+      - type: git
+        url: https://github.com/libsdl-org/SDL.git
+        tag: release-3.4.0
+        commit: a962f40bbba175e9716557a25d5d7965f134a3d3
+
+  - name: sdl2-compat
```
CodedOre: Use the shared modules for this.

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,229 @@
...
+    build-commands:
+      - ln -s /usr/lib/$(uname -m)-linux-gnu/libvulkan.so.1 /app/lib/libvulkan.so
+
+  - name: pylux
+    buildsystem: cmake-ninja
+    builddir: true
+    sources:
+      - type: git
+        url: https://github.com/ForWard-Technologies-LLC/Pylux.git
+        commit: 6855bf2386686843bd27a59805deb1f27a537d0e
```
CodedOre: Note about the metainfo file:

The "screenshot" is a promotional image for Android. That's a no. \
Also, do not point the screenshot URL to a branch, use a permalink instead.

0001-vulkan-ignore-frames-without-hw-context.patch
teamcons, line 1: Why do you need a patch

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,229 @@
...
+        url: https://github.com/libusb/hidapi.git
+        tag: hidapi-0.15.0
+        commit: d6b2a974608dec3b76fb1e36c189f22b9cf3650c
+    cleanup:
+      - /lib/*.la
+      - /lib/pkgconfig
+      - /include
+
+  - name: speexDSP
+    buildsystem: simple
```
hfiguiere: this should be `buildsystem: autotools`

io.github.ForWard_Technologies_LLC.Pylux.yml
hfiguiere: no, it's in the runtime too.

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,229 @@
...
+    config-opts:
+      - -DSDL2COMPAT_TESTS=OFF
+      - -DCMAKE_INSTALL_LIBDIR=lib
+    post-install:
+      - mv /app/lib/pkgconfig/sdl2-compat.pc /app/lib/pkgconfig/sdl2.pc
+
+  - name: vulkanso
+    buildsystem: simple
+    build-commands:
+      - ln -s /usr/lib/$(uname -m)-linux-gnu/libvulkan.so.1 /app/lib/libvulkan.so
```
hfiguiere: why this?

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,235 @@
...
+    aarch64:
+      prepend-pkg-config-path: "/app/lib/pkgconfig:/app/lib/aarch64-linux-gnu/pkgconfig"
+    x86_64:
+      prepend-pkg-config-path: "/app/lib/pkgconfig:/app/lib/x86_64-linux-gnu/pkgconfig"
+
+modules:
+  - name: protobuf-compilers
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_INSTALL_LIBDIR=lib
```
hfiguiere:
```suggestion
```

Use the latest flatpak-builder or the one as a flatpak as per the documenation.

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,235 @@
...
+      - /lib/pkgconfig
+    sources:
+      - type: archive
+        url: https://github.com/protocolbuffers/protobuf/releases/download/v21.12/protobuf-cpp-3.21.12.tar.gz
+        sha256: 4eab9b524aa5913c6fffb20b2a8abf5ef7f95a80bc0701f3a6dbb4c607f73460
+
+  - name: libplacebo
+    buildsystem: meson
+    config-opts:
+      - -Dlibdir=lib
```
hfiguiere:
```suggestion
```

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,235 @@
...
+    sources:
+      - type: git
+        url: https://github.com/json-c/json-c.git
+        tag: json-c-0.18-20240915
+        commit: 41a55cfcedb54d9c1874f2f0eb07b504091d7e37
+
+  - name: miniupnpc
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_INSTALL_LIBDIR=lib
```
hfiguiere:
```suggestion
```

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,235 @@
...
+  - name: vulkanso
+    buildsystem: simple
+    build-commands:
+      - ln -s /usr/lib/$(uname -m)-linux-gnu/libvulkan.so.1 /app/lib/libvulkan.so
+
+  - name: pylux
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+      - -DCMAKE_INSTALL_LIBDIR=lib
```
hfiguiere:
```suggestion
```

forward-technologies:
bot, build

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Thank you, I implemented this and am kicking off a build to test. Appreciate the feedback. I will reply to comments and address everything shortly.

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,196 @@
...
+  - --filesystem=xdg-download
+  - --filesystem=~/.steam/steam
+
+cleanup-commands:
+  - /app/cleanup-BaseApp.sh
+
+build-options:
+  arch:
+    aarch64:
+      prepend-pkg-config-path: "/app/lib/pkgconfig:/app/lib/aarch64-linux-gnu/pkgconfig"
```
hfiguiere: this shouldn't be necessary.

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,196 @@
...
+    x86_64:
+      prepend-pkg-config-path: "/app/lib/pkgconfig:/app/lib/x86_64-linux-gnu/pkgconfig"
+
+modules:
+  - name: protobuf-compilers
+    buildsystem: cmake-ninja
+    config-opts:
+      - -Dprotobuf_BUILD_TESTS=OFF
+      - -Dprotobuf_BUILD_SHARED_LIBS=OFF
```
hfiguiere: you should build shared libraries and not static.

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,196 @@
...
+        url: https://github.com/miniupnp/miniupnp
+        tag: miniupnpc_2_3_0
+        commit: f39e4997591a98e7ee6711eabc8b50feeef1bd67
+
+  - name: pylux
+    buildsystem: cmake-ninja
+    builddir: true
+    sources:
+      - type: git
+        url: https://github.com/ForWard-Technologies-LLC/Pylux.git
```
bbhtt, line 127: so you forked chiaki last year from https://github.com/streetpea/chiaki-ng/commit/b05ff3977052c8b3bde314806f376cc160f1f0d1 and added LLM slop to it?

the only change being some random manifest and metainfo updates and one giant commit that's not readable?

Image

And you are using LLM agent to update metainfo URLS

forward-technologies: @bbhtt 

> so you forked chiaki last year from https://github.com/streetpea/chiaki-ng/commit/b05ff3977052c8b3bde314806f376cc160f1f0d1 and added LLM slop to it?

Hi, no, I spent month analyzing the PS Cloud Play protocol and was able to get Playstation cloud play fully functional. This adds full support for PS Cloud streaming (both PS Cloud Game catalog as well as PS Cloud game library - so PS5 cloud game streaming). I am in contact with the chiaki dev and he actually suggested that this be its own different project that I support rather than trying to incorporate it into chiaki. The cloud play changes are quite significant, but the actual streaming protocol is similar enough to remote play that using the familiar chiaki project made sense.

Alternatively, I could simply keep this project closed source and private. I’ve invested a considerable amount of time and effort into building something genuinely new and highly sought after by the community, and I’m going out of my way to share it publicly rather than keeping it to myself.

> And you are using LLM agent to update metainfo URLS

Yes its 2026, of course I am using an agentic programming IDE? I'm not sure what you are getting at here.

forward-technologies: Oh perhaps this was a misunderstanding as well. I squash merged my changes to make things easy to track for myself, checkout the **72k** lines of code change in the commit (would be easy to miss): https://github.com/ForWard-Technologies-LLC/Pylux/commit/74450a424fb2f297bb4ab376c4dad85017be6098 So I understand that at first glance it would like like this was just a few extra commits on top of chiaki rather than a major divergence :)

bbhtt: No I saw it later, still doesn't inspire confidence.

bbhtt: >  I am in contact with the chiaki dev and he actually suggested that this be its own different project that I support rather than trying to incorporate it into chiaki.

Is that public?

> Yes its 2026, of course I am using an agentic programming IDE? I'm not sure what you are getting at here.

How much of it was made through LLM?

forward-technologies: Of the actual Pylux repo itself? I used an LLM mainly for the iOS port (lost steam towards the end, and I am less familiar with swift) but that wouldn't be applicable to the flatpak / linux portion. The changes for Qt (what would be used here for flathub) was mostly just me directly, since I am quite familiar with Qt development.

> Is that public?

Hmm not that I know of, I can ask if he could email someone I guess if that would help? Again, this is a major divergence from Chiaki and warrants its own project.

forward-technologies:
bot, build

io.github.ForWard_Technologies_LLC.Pylux.yml
bbhtt: > The changes for Qt (what would be used here for flathub) was mostly just me directly, since I am quite familiar with Qt development.

Ok I will have a look again later.

forward-technologies: Sounds good, thanks for your time.

I'm going to continue to address the above comments and test things from my end, so apologies for any additional noise.

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: `6.9` is not yet EOL and all our QA + Steam Deck testing for this initial submission was done against it. We'll bump to `6.10` in a follow-up PR shortly after this lands. Is that fine?

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Good catch, thank you — fixed in `9a94c47`.

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Thanks, fixed in `76118cf`.

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Thanks, fixed in `76118cf`.

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Thanks, module removed in `54b188e`. Pylux only links SDL2, which the runtime already provides.

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Thanks, module removed in `54b188e`, using the runtime's SDL2 directly.

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Thanks, both fixed. Replaced the Android promo with two desktop UI captures and the URLs now point at a commit-permalink. Visible in `pylux` module commit `a956aa9e`.

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Thanks, fixed in `76118cf`.

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Thanks, module removed in `54b188e`. The libvulkan.so alias was unnecessary; libvulkan.so.1 from the runtime resolves correctly.

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Thanks, entire build-options block removed in `86d7e05`. Verified Pylux's CMake locates Qt via find_package(Qt6) and only pkg_check_modules for libplacebo / speexdsp / fftw3, all reachable via the default search path.

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Thanks, switched to shared in `86d7e05` (-Dprotobuf_BUILD_SHARED_LIBS=ON). Smoke-tested with the chiaki binary linking libprotobuf.so.32 dynamically.

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Pylux does use the portal for the inhibit path (--talk-name=org.freedesktop.ScreenSaver). The reason org.freedesktop.login1 is also needed is the PrepareForSleep system-bus signal, which the portal does not expose. Pylux subscribes to it so it can cleanly tear down an active streaming session before the host suspends, otherwise the PS5/PS4 console keeps the session open and the user gets stuck reconnecting. This is especially painful on Steam Deck where suspend is the primary way to stop a streaming session.

For this initial PR I have temporarily commented it out so lint passes. I will open a flatpak-builder-lint exception PR with this rationale before this lands.

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Per-permission rationale on the current set:

- xdg-run/pipewire-0: direct PipeWire socket for low-latency audio.
- xdg-run/gamescope-0: Gamescope WSI socket for native fullscreen/HDR on Steam Deck.
- xdg-download: Pylux uses QFileDialog with DontUseNativeDialog for cross-platform consistency, so it does not route through the FileChooser portal. This grants the default save location for export settings and log saves. The dialog default was also moved from DesktopLocation to DownloadLocation to avoid needing xdg-desktop.
- ~/.steam/steam: needed for the "Add as Steam shortcut" feature (reads loginusers.vdf, writes shortcuts.vdf and grid art). Used only on explicit user action.

Removed in 76118cf: host-os:ro (leftover from early Steam Deck testing, verified unused) and xdg-documents (never used).

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: The runtime FFmpeg is built with --disable-decoder=h264,hevc,vc1,vvc so only the libopenh264 software path is available, with no HEVC support. Pylux streams H.264 and H.265 from PS4/PS5 with hardware acceleration (h264_vaapi, hevc_vaapi, h264_vulkan, hevc_vulkan), all of which require the native decoders. The bundled module enables exactly those plus --disable-all for a minimal footprint.

0001-vulkan-ignore-frames-without-hw-context.patch
forward-technologies: 3-line defensive null-check in ff_vk_exec_add_dep_frame(). Pylux's Vulkan decode pipeline hits frames without a populated hw_frames_ctx during decoder reinit, for example when the PS5 host renegotiates resolution mid-stream. Without the guard ffmpeg dereferences NULL and the session crashes. Happy to drop it if there is a preferred upstream commit to reference instead. But this does prevent crashes.

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Implemented, thanks!

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Done thanks!

io.github.ForWard_Technologies_LLC.Pylux.yml
CodedOre: New submissions need to use the latest version if the runtime. 

If you need time, we can wait until you've done the update.

io.github.ForWard_Technologies_LLC.Pylux.yml
CodedOre: So, as I understand, `org.freedesktop.Screensaver` could be covered by the Inhibit portal?

In this case, drop the permission. Portals don't need a permission in the manifest. 

For the other talk name:\
If it is important, why do you comment it out? Just request the linter exception.

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Hmm, all I found in your documentation when setting this up was that it shouldn't be EOL, which 6.9 is not, right? 

"Submissions using an end-of-life runtime, extension or baseapp will not be accepted." 

Can you point me to where it says all dependencies must be on the latest version?

It's only a 3-line change, so I can do it now. I just want to confirm this is truly the policy before treating it as a blocker.

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Doh I'm dumb, yes you are totally right. Fixed!

On login1: I had it commented out because my exception PR was quickly closed without much discussion. Without that exception merged, the linter blocks CI here, so I commented it out to unblock testing and address the other review comments. In hindsight the quick closure may have been partly my fault since the PR still had other questionable permissions at the time. I created a new PR for this here: https://github.com/flathub-infra/flatpak-builder-lint/pull/1024 I will uncomment once that is approved :)

forward-technologies:
bot, build

io.github.ForWard_Technologies_LLC.Pylux.yml
teamcons: > Yes its 2026, of course I am using an agentic programming IDE?

......

io.github.ForWard_Technologies_LLC.Pylux.yml
hfiguiere: I'm not sure what was unclear here. Or maybe I see what is clear: you don't want your package to be accepted.

io.github.ForWard_Technologies_LLC.Pylux.yml
hfiguiere: not if you use the correct runtime.

io.github.ForWard_Technologies_LLC.Pylux.yml
bbhtt: It not being listed is an oversight, I will add it now. It's what we have been following for years.

Also 6.11 beta is out and very near full release, it was already released but recalled due to bugs. So 6.9 will be EOL weeks after that, it doesn't have a lot of time left.

forward-technologies: Ok understood I will update to 6.10 shortly

forward-technologies:
bot, build

io.github.ForWard_Technologies_LLC.Pylux.yml
hfiguiere: so what about now? it has h264, hevc and other patent encumbered codec.

forward-technologies: Thanks for the catch. You're right, runtime FFmpeg with codecs-extra does provide h264/hevc and the vulkan/vaapi hwaccels. I just tested and it's working. However, the reason we still need the bundled FFmpeg module is to carry [this 2 line NULL check patch](https://github.com/ForWard-Technologies-LLC/flathub/blob/442aca06900e7db96b0488343e00ca59a7eb40be/0001-vulkan-ignore-frames-without-hw-context.patch) in [ff_vk_exec_add_dep_frame()](https://github.com/FFmpeg/FFmpeg/blob/51660ad52361befd3c7e55b4d4b5af7e399150af/libavutil/vulkan.c#L790) that prevents a known crash when frames arrive without a populated hw_frames_ctx. The bug is already tracked upstream ([FFmpeg #10847](https://trac.ffmpeg.org/ticket/10847)) and has also been hit by mpv ([#13375](https://github.com/mpv-player/mpv/issues/13375), [#14029](https://github.com/mpv-player/mpv/issues/14029)). I just checked master and n7.1.3 and the fix isn't in either, so it can't be applied to the runtime FFmpeg (at least I don't think there is a way?). Module is otherwise minimal (--disable-all plus just what we need).

I would like to remove it as well, but keeping it until FFmpeg is patched seems like the right call to prevent real crashes. Sound fine?

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Ok I updated to 6.10 and did a full end to end test on several devices. There was one small regression in my code related to an openssl update, but I fixed that and can confirm everything is working. All things considered, a pretty painless update. Closing this comment out as resolved.

forward-technologies: > It not being listed is an oversight, I will add it now. It's what we have been following for years.

Thanks @bbhtt, I'm sure this will save both reviewers and submitters some time. As a submitter, following the documentation exactly (for the runtime portion at least) and still getting rejected, then having to retest everything due to ~~undocumented~~ requirements (not undocumented actually, the documentation directly contradicts what's actually enforced) is frustrating. And as a reviewer, I'm sure you all get tired of responding to the same things repeatedly. Having the docs reflect what's actually enforced would be welcomed

0001-vulkan-ignore-frames-without-hw-context.patch
forward-technologies: Linking relevant discussion here: https://github.com/flathub/flathub/pull/8607#discussion_r3213891126 for tracking

io.github.ForWard_Technologies_LLC.Pylux.yml
bbhtt: I applied the patch to runtime, you can wait for the next release which should be soon.

forward-technologies:
bot, build

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Wow that's awesome thanks, you guys are on top of it. Okay I went ahead and removed the ffmpeg patch in preparation for it being applied in the runtime by default now. I reran some e2e tests on my end and things look good. Considering this resolved and closing out.

forward-technologies:
Status update:

- I believe all comments have been addressed. The app has been fully tested with the suggested changes applied and confirmed functional.
- Still waiting on the exception PR: https://github.com/flathub-infra/flatpak-builder-lint/pull/1024. But thats non-blocking since I've temporarily commented out login1 [here](https://github.com/flathub/flathub/pull/8607/files#diff-191b25b8c9541df7c183c118da02ad736d605c205a03bf832cbbdd3e4d5fa89aR22). Since it's a nice to have feature, I'm happy to merge without it and re-enable it in a follow-up once the exception lands (especially if exception PRs are known to take longer for approval? Not sure if thats the case).
- Waiting on the FFmpeg patch from @bbhtt, but also non-blocking since it's an edge case that will only affect some users occasionally. Happy to merge and let the runtime update go out async.

Let me know if I've missed anything.

bbhtt:
I haven't decided on the comment where I said I will look at thr code again, but you have marked it as resolved.

I will merge the exception for now as soon ss you fix it so that you can have a functional build.

FFMPEG patch will need to wait until runtime release is out in a few weeks

forward-technologies:
> I haven't decided on the comment where I said I will look at the code again, but you have marked it as resolved.

Oh I guess I misunderstood, I assumed you had finished since you moved on to reviewing the rest. Unresolved now

> I will merge the exception for now as soon as you fix it so that you can have a functional build.

Thanks, uncommented login1 now.

> FFMPEG patch will need to wait until runtime release is out in a few weeks

That seems fine from my end. Its a real bug, but it only occurred a total of 2 times in all my testing so its a pretty infrequent edge case. I'm comfortable merging without the patch and just waiting for the runtime update to ship. Or I can ad the patch back temporarily as well. I thought it was causing some contention (flagged in review) and it's a little ugly, so I was quick to remove it. I can  re-add it if thats what you're getting at

forward-technologies:
bot, build

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,160 @@
...
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=all
+  - --allow=bluetooth
+  - --socket=pulseaudio
+  - --env=QTWEBENGINEPROCESS_PATH=/app/bin/QtWebEngineProcess
+  - --env=LD_LIBRARY_PATH=/app/lib:/usr/lib/extensions/vulkan/gamescope/lib
```
bbhtt: you don't need to set /app/lib it is already there by default, but where is `/usr/lib/extensions/vulkan/gamescope/lib` coming from?

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,160 @@
...
+  - name: protobuf-compilers
+    buildsystem: cmake-ninja
+    config-opts:
+      - -Dprotobuf_BUILD_TESTS=OFF
+      - -Dprotobuf_BUILD_SHARED_LIBS=ON
+      - -Dprotobuf_BUILD_EXAMPLES=OFF
+    cleanup:
+      - /include
+      - /lib/cmake
+      - /lib/pkgconfig
```
bbhtt: these should be in toplevel cleanup, no need to repeat.

io.github.ForWard_Technologies_LLC.Pylux.yml
```diff
@@ -0,0 +1,160 @@
...
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DHIDAPI_WITH_LIBUSB=FALSE
+    sources:
+      - type: git
+        url: https://github.com/libusb/hidapi.git
+        tag: hidapi-0.15.0
+        commit: d6b2a974608dec3b76fb1e36c189f22b9cf3650c
+    cleanup:
+      - /lib/*.la
```
bbhtt: this too in toplevel cleanup.

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Yeah agreed, just cleaned that up and resolving

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: :+1:

io.github.ForWard_Technologies_LLC.Pylux.yml
forward-technologies: Good catch, youre right. I added the gamescope path during testing without realizing the loader just opens the .so directly by absolute path so LD_LIBRARY_PATH wasn't doing anything. Removed it and confirmed via VK_LOADER_DEBUG=all and it shows the layer still loads. So I safely removed this, resolving comment

forward-technologies:
bot, build

forward-technologies:
bot, build

forward-technologies:
Status update:
- I applied and resolved your above cleanup comments @bbhtt, thanks for the find.
- I cleaned up a few other leftover scripts/artifacts that were left over from testing
- I did another round of e2e tests with a few device (including steam deck to test the gamescope changes) and confirmed everything is functioning as expected.

At this point, I believe I have addressed al of the open concerns, please let me know if there is anything else I missed when you have the time.

forward-technologies:
Hey guys, just checking in to make sure you have everything you need from me and that I’m not holding anything up on my end.

bbhtt:
There hasn't been a runtime release yet with the ffmpeg crash fix, it is waiting on that.

forward-technologies:
@bbhtt The ffmpeg crash impacts a very small percentage of users (1 in a thousand?), I would prefer this to be non blocking.

forward-technologies:
bot, build

bbhtt:
bot, build

PR #8610, opened 2026-05-08
Add com.revpdf.editor

Pawandeep-prog:

RevPDF is a pdf editor which is offline, no-signup, free. It let's user edit, create, convert PDFs. It can edit text, redact, fill forms, create forms, edit images and so on. It have utility tools like merge, split, compress, OCR, and much more tools. 
[revpdf_recording.webm](https://github.com/user-attachments/assets/c8d1329d-d7c5-4e67-aac9-c2c7e612a576)
- [X] I am an _developer_ to the project.


com.revpdf.editor.yml
```diff
@@ -0,0 +1,48 @@
+app-id: com.revpdf.editor
+runtime: org.freedesktop.Platform
+runtime-version: '24.08'
```
hfiguiere: 25.08 is the latest

com.revpdf.editor.yml
```diff
@@ -0,0 +1,48 @@
...
+command: revpdf_editor
+
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --filesystem=xdg-documents
+  - --filesystem=xdg-download
+  - --env=TMPDIR=/var/tmp
```
hfiguiere, line 14: why do you need TMPDIR?

com.revpdf.editor.metainfo.xml
```diff
@@ -0,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.revpdf.editor</id>
+  <metadata_license>MIT</metadata_license>
+  <project_license>LicenseRef-proprietary</project_license>
```
hfiguiere: what is the license? (missing a link here) are the binaries redistributable?

com.revpdf.editor.metainfo.xml
```diff
@@ -0,0 +1,71 @@
...
+  </description>
+
+  <launchable type="desktop-id">com.revpdf.editor.desktop</launchable>
+
+  <url type="homepage">https://revpdf.com/</url>
+  <url type="bugtracker">https://github.com/Pawandeep-prog/revpdf-release/issues</url>
+
+  <screenshots>
+    <screenshot type="default">
+      <image>https://raw.githubusercontent.com/Pawandeep-prog/revpdf-release/main/images/home.png</image>
```
hfiguiere: URL should NEVER reference the tip of a branch. It's not immutable.

com.revpdf.editor.desktop
hfiguiere, line 1: This file should be (submitted and managed) upstream.

com.revpdf.editor.metainfo.xml
hfiguiere, line 1: This file should be (submitted and managed) upstream.

com.revpdf.editor.png
hfiguiere, line 1: This file should be (submitted and managed) upstream.

com.revpdf.editor.yml
```diff
@@ -0,0 +1,48 @@
...
+      - install -Dm755 revpdf_editor.sh /app/bin/revpdf_editor
+      # Desktop integration
+      - install -Dm644 com.revpdf.editor.desktop /app/share/applications/com.revpdf.editor.desktop
+      - install -Dm644 com.revpdf.editor.png /app/share/icons/hicolor/256x256/apps/com.revpdf.editor.png
+      - install -Dm644 com.revpdf.editor.metainfo.xml /app/share/metainfo/com.revpdf.editor.metainfo.xml
+    sources:
+      # Pre-built release bundle — update url and sha256 for each new release.
+      # To get the sha256: sha256sum revpdf_editor-linux-x86_64.tar.gz
+      - type: archive
+        url: https://github.com/Pawandeep-prog/revpdf-release/releases/download/v4.0.0/revpdf_editor-linux-x86_64.tar.gz
```
bbhtt, line 33: Why is it proprietary? Because you plan to hide your LLM slops?

Pawandeep-prog: Sorry for that,
Part of using llm is true, I used llm in some parts where I had to work with something very new and didn't wanted to screw up the things, for those I took help of llm. Asking it how it could be done, and then learning how it is done.

bbhtt: I'm finding it hard to trust and accept proprietary apps that too with LLM use involved.

com.revpdf.editor.yml
Pawandeep-prog: I set TMPDIR=/var/tmp because RevPDF writes temporary files to the system temp directory in several places, most critically, the undo/redo system stores compressed PDF snapshots via Directory.systemTemp directly. Flatpak's default /tmp is a small in-memory tmpfs, which can be exhausted by large PDFs or long editing sessions. /var/tmp is disk-backed and avoids that issue.

com.revpdf.editor.yml
Pawandeep-prog: updated in the new commit

com.revpdf.editor.metainfo.xml
Pawandeep-prog: i have updated the license URL in the file, to below license
https://github.com/Pawandeep-prog/revpdf-release/blob/main/LICENSE

com.revpdf.editor.metainfo.xml
Pawandeep-prog: fixed

com.revpdf.editor.yml
Pawandeep-prog: I plan to open source the desktop codebase once I separate it from mobile, that's already on my roadmap. Right now both share the same Flutter codebase, and the mobile side relies on small IAPs which is what keeps RevPDF free on desktop. I want to split them cleanly before open sourcing so the IAP logic isn't tangled in the desktop build.
RevPDF already has 40k desktop downloads and 10k+ on mobile with a ~4.5 rating on both Android and iOS, so there's an established user base that's found it trustworthy. Users have also been requesting a Flatpak build for a while, which is what prompted this submission. 
I'm happy to clarify exactly what data sent/ received if that helps with the review. RevPDF is solely built on offline, no-sign up purpose.

Pawandeep-prog:
bot, build

Pawandeep-prog:
bot, build

com.revpdf.editor.yml
```diff
@@ -0,0 +1,49 @@
...
+sdk: org.freedesktop.Sdk
+command: revpdf_editor
+
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --filesystem=xdg-documents
+  - --filesystem=xdg-download
```
bbhtt, line 13: why does these two filesystem permissions not use the portsl? and why are they needed?

com.revpdf.editor.yml
bbhtt: marking this as resolved because there is no way for me to tell but it's not a good look to make a PR ignoring template then a proprietary application.

having tight permissions may be its saving grace.

bbhtt: actually why is metainfo urls being updated using claude?

com.revpdf.editor.yml
Pawandeep-prog: File picking goes through the portal (file_picker -> org.freedesktop.portal.FileChooser). The filesystem permissions aren't for that - they're for everything after: overwriting the currently open file, and re-opening recent files. file_picker returns the raw host path, so those operations bypass the portal and need direct fs access. The correct long-term fix is to use the portal document store for persistent access, but that requires a plugin-level change.

com.revpdf.editor.yml
Pawandeep-prog: I'm using Claude as a dev tool to automate the parts like fetching commit SHAs and recalculating sha256 checksums, the same thing I'd write a shell script for otherwise.

com.revpdf.editor.yml
bbhtt: What "plugin-level" change?

Pawandeep-prog: Removing these permissions would require changing every file operation in the app to go through portal document paths instead. That's the "plugin-level change", not a single call, but a full audit of the file-handling layer. I'll do it properly in the next update. Currently, recent files are something which will be affected, I could remove these permission if it required, I can manage to create a fix for new update.

bbhtt: Where exactly is that plugin is what I am asking. What are you referring as plugin?


Those sentences seem like they were generated with AI.

Pawandeep-prog: I have developed my own revpdf package, where all the file related operations occur. I have to work on that part. 
If for now these lines in yml seems suspicious I can take them out, and prepare a fix in the future.

bbhtt: This is going to be blocked because it is proprietary, you admitted to using LLMs and you are unable to provide clear justifications. All of this raises suspicions.

Pawandeep-prog: so I did not provide justification is the reason for block? I told I need these permission for accessing recently opened files, I could remove if it is causing any suspicion, or I can get back in few days with the fix.

I can share reddit comments where people are asking for revpdf on flathub, 
or atleast tell me what are my options from here?

bbhtt: Reddit comments do not decide a single thing here.

Your options were to provide a human generated response when I asked 3 comments back, you provided an LLM generated response instead, so I'm afraid the options might be exhausted.

Pawandeep-prog: alright, so was the justification not alright to you? I might have formatted the response because I didn't wanted to screw up this, looks like I did.
sorry for the trouble.

PR #8611, opened 2026-05-08
Add io.kinvolk.Headlamp.Tools.oci

dylanmtaylor:

- [x] Please describe the application briefly. Oracle OCI CLI extension for Headlamp Kubernetes IDE. Provides the `oci` command-line tool as an optional extension, enabling Oracle Cloud Infrastructure connectivity from within Headlamp.
**Note:** This extension depends on https://github.com/flathub/io.kinvolk.Headlamp/pull/70 being merged first (adds the `io.kinvolk.Headlamp.Tools` extension point).
- [x] Please attach a video showcasing the application on Linux using the Flatpak. N/A - this is a CLI extension, not a graphical application. See https://github.com/flathub/flathub/pull/2290
* [x] I am a third-party packager. I contacted upstream developers about this submission. **Link:** https://github.com/flathub/io.kinvolk.Headlamp/pull/70#issuecomment-4397336318

### DIFF
--- a/io.kinvolk.Headlamp.Tools.oci.yaml
+++ b/io.kinvolk.Headlamp.Tools.oci.yaml
@@ -5,7 +5,6 @@ runtime-version: stable
 sdk: org.freedesktop.Sdk//25.08
 build-extension: true
 separate-locales: false
-appstream-compose: false
 build-options:
   prefix: /app/tools/oci
   env:
@@ -13,3 +12,10 @@ build-options:
     PYTHONPATH: /app/tools/oci/lib/python3.13/site-packages
 modules:
   - python3-oci-cli.yaml
+  - name: metainfo
+    buildsystem: simple
+    sources:
+      - type: file
+        path: io.kinvolk.Headlamp.Tools.oci.metainfo.xml
+    build-commands:
+      - install -Dm644 io.kinvolk.Headlamp.Tools.oci.metainfo.xml ${FLATPAK_DEST}/share/metainfo/io.kinvolk.Headlamp.Tools.oci.metainfo.xml
--- a/io.kinvolk.Headlamp.Tools.oci.yaml
+++ b/io.kinvolk.Headlamp.Tools.oci.yaml
@@ -1,0 +1,15 @@
+id: io.kinvolk.Headlamp.Tools.oci
+branch: stable
+runtime: io.kinvolk.Headlamp
+runtime-version: stable
+sdk: org.freedesktop.Sdk//25.08
+build-extension: true
+separate-locales: false
+appstream-compose: false
+build-options:
+  prefix: /app/tools/oci
+  env:
+    PATH: /app/tools/oci/bin:/app/bin:/usr/bin
+    PYTHONPATH: /app/tools/oci/lib/python3.13/site-packages
+modules:
+  - python3-oci-cli.yaml
--- a/io.kinvolk.Headlamp.Tools.oci.metainfo.xml
+++ b/io.kinvolk.Headlamp.Tools.oci.metainfo.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="addon">
+  <id>io.kinvolk.Headlamp.Tools.oci</id>
+  <extends>io.kinvolk.Headlamp</extends>
+  <name>OCI CLI for Headlamp</name>
+  <summary>Oracle Cloud Infrastructure CLI for authenticating to OKE clusters</summary>
+  <description>
+    <p>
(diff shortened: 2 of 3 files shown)


hfiguiere:
why not just shipping it with the main package?

dylanmtaylor:
> why not just shipping it with the main package?

The idea is to not have CLIs, etc. for every conceivable cloud environment shipped with Headlamp. Only a small subset of users will need the OCI cli installed for interacting with OKE clusters, so to keep the main padckage  less cluttered I think having an extension of sorts makes sense.

dylanmtaylor:
I wrote this on the main Headlamp flatpak:

I intend to package an OCI plugin for the headlamp flatpak since the configuration Oracle provides requires that binary OOTB to connect. I'm thinking an io.kinvolk.Headlamp.Plugin.oci repository so it's not installed for all users, similar to how VLC has org.videolan.VLC.Plugin.makemkv.

hfiguiere:
Headlamp is a video player?

dylanmtaylor:
> Headlamp is a video player?

No. It is an application used for managing enterprise infrastructure (Kubernetes clusters). Some of which reside on specific cloud providers and require a specific CLI to authenticate and manage them.

hfiguiere:
Then why do you compare it to VLC?

dylanmtaylor:
> Then why do you compare it to VLC?

Because of the same type of flatpak setup where it is an extension/addon to the original application's functionality.

bbhtt:
How big is this and how many other similar such tools exist that may make sense to add to the app? If the answer is small and no many respectively then it belongs in the app.

dylanmtaylor:
> How big is this and how many other similar such tools exist that may make sense to add to the app? If the answer is small and no many respectively then it belongs in the app.

Off the top of my head, it may be needed for CLIs for other cloud providers that provide managed K8s instances, like AWS, Azure, Google Cloud, OCI. It's the OCI cli is fairly small. It's just only needed by a very small subset of users who specifically use the tool with Oracle Kubernetes Engine so shipping it in the main flatpak feels like bloat if you don't need it.

dylanmtaylor:
➜  ~ flatpak list --columns=size,name|sort -g | grep -i OCI | head -n 1
584.9 MB	oci

➜  ~ flatpak list --columns=size,name|sort -g | grep -i headlamp | head -n 1
389.2 MB	Headlamp

I imagine some of this is from other things pulled in to the flatpak and not just the OCI cli. That's significant bloat for users who don't need it.

dylanmtaylor:
The actual CLI is 27MB but I think the large size is because it's pulling in all of Python too. Headlamp isn't a python app so this isn't something that would be there. I suspect it'd at least double the dist size.

dylanmtaylor:
> ❌ [Test build](https://github.com/flathub-infra/vorarbeiter/actions/runs/25590743187) failed.
> 
> Help

This is because this was tested against the version with https://github.com/flathub/io.kinvolk.Headlamp/pull/70 merged in, built locally and not the version currently shipped.

dylanmtaylor:
@bbhtt I'm not sure if you saw my response. I'm still interested in doing this work. Maybe for all of the optional tools for Headlamp - which is currently 

```
aliyunl;aws;az;doctl;gcloud;helm;ibmcloud;k3s;kind;kubectl;kubelogin;minikube;oc;oci
```

yolossn:
https://github.com/flathub/io.kinvolk.Headlamp/pull/70 is merged

bbhtt:
bot, build

io.kinvolk.Headlamp.Tools.oci.yaml
```diff
@@ -0,0 +1,15 @@
+id: io.kinvolk.Headlamp.Tools.oci
+branch: stable
+runtime: io.kinvolk.Headlamp
+runtime-version: stable
+sdk: org.freedesktop.Sdk//25.08
+build-extension: true
+separate-locales: false
+appstream-compose: false
```
bbhtt: you should add an appstream file

yolossn: created a new PR with the appstream file since this PR looks stalled
https://github.com/flathub/flathub/pull/9278

dylanmtaylor:
leave-open - this is a CLI tool/extension and a video is not applicable

io.kinvolk.Headlamp.Tools.oci.yaml
dylanmtaylor: Will pull that into my PR @yolossn

dylanmtaylor:
https://github.com/dylanmtaylor/flathub/commit/0129588ef76e1173e90e475aa5fba4c7489d7c4d

PR #8626, opened 2026-05-10
Add net.shrieker.capsicum

pooza:
## Description

[capsicum](https://github.com/pooza/capsicum) is a Mastodon / Misskey Fediverse client built with Flutter. It supports both Mastodon and Misskey APIs through a unified adapter layer, so the same UI works on both servers and switches between accounts seamlessly. The desktop release adds Linux to the existing macOS / iOS / Android lineup.

- Homepage: https://capsicum.shrieker.net/
- Source: https://github.com/pooza/capsicum
- License: AGPL-3.0-or-later
- Latest release: [v1.24.0](https://github.com/pooza/capsicum/releases/tag/v1.24.0) (2026-05-10)

## Manifest notes

- **Source layout**: prebuilt Flutter Linux release bundle is fetched from a GitHub Releases tarball. The Flathub sandbox is offline, so `flutter pub get` cannot run inside the build — using the pre-bundled output instead.
- **Runtime**: `org.gnome.Platform//49`. GNOME runtime is selected over Freedesktop because the bundled `desktop_webview_window` plugin links against `libwebkit2gtk-4.1.so.0`, which only ships in the GNOME runtime.
- **`finish-args`**: minimum required surface (network, IPC, Wayland + X11 fallback, DRI for GPU rendering, secrets / notifications via portal-aware D-Bus, xdg-pictures / xdg-download for media attachment).

## Verification

- `flatpak-builder --user --force-clean --install-deps-from=flathub --disable-rofiles-fuse build-dir net.shrieker.capsicum.yml` succeeds locally on Debian 13.
- `appstreamcli compose` during the build emits no warnings or errors.
- `appstreamcli validate` against the metainfo file passes (`✔ Validation was successful.`).
- The resulting Flatpak runs and Japanese IME (ibus + mozc) input works correctly.

## Self-review checklist

- [x] App ID matches the reverse-DNS of a domain owned by the submitter (`shrieker.net`).
- [x] AppStream metainfo includes screenshots, OARS rating, license, releases, and developer.
- [x] No use of `--filesystem=host` / `--filesystem=home` or other broad permissions.
- [x] Runtime version is the latest stable GNOME runtime (49).
- [x] All sources referenced by URL with sha256 (no `path:` references).

### DIFF
--- a/net.shrieker.capsicum.yml
+++ b/net.shrieker.capsicum.yml
@@ -1,0 +1,96 @@
+# capsicum Flathub submission manifest (v1.24.0 initial submission)
+#
+# Derived from packaging/linux/flathub/net.shrieker.capsicum.yml
+# (local development manifest) per packaging/linux/flathub/SUBMISSION.md.
+#
+# Source layout:
+# - capsicum bundle: GitHub Releases tarball (v1.24.0 tag asset)
+# - desktop / icons: capsicum repo at v1.24.0 tag
+# - metainfo: capsicum repo at commit 72d54060 (Phase 5b screenshots /
+#   release date 反映 + screenshot URL を SHA 参照に固定)
+
+app-id: net.shrieker.capsicum
+
+runtime: org.gnome.Platform
+runtime-version: '49'
+sdk: org.gnome.Sdk
+
+command: capsicum
+
+finish-args:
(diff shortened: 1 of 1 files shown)


CodedOre:

See also our Generative AI policy:

PR #7415, opened 2025-12-28
Add org.freedesktop.Sdk.Extension.podman

francoism90:
### Please confirm your submission meets all the criteria

- [X] This is a SDK, used in editors like PHPStorm and VSCode (this still has `/app/tools`)
- [X] Please attach a video showcasing the application on Linux using the Flatpak. See screenshot below. I cannot make any videos sorry.
- [X] The Flatpak ID follows all the rules listed in the [Application ID requirements][appid].
- [X] I have read and followed all the [Submission requirements][reqs] and the [Submission guide][reqs2] and I agree to them.
- [ ] I am an _(please keep whichever is applicable and remove the rest)_ author/developer/upstream contributor to the project.
      If not, I contacted upstream developers about this submission. **Link:**

[appid]: https://docs.flathub.org/docs/for-app-authors/requirements#application-id
[reqs]: https://docs.flathub.org/docs/for-app-authors/requirements
[reqs2]: https://docs.flathub.org/docs/for-app-authors/submission

PR #8677, opened 2026-05-15
Add org.freedesktop.Sdk.Extension.podman

francoism90:

- [X] Please describe the application briefly. < This would allow Flatpak applications to use `podman` as generic SDK. I have tested this with PhpStorm and VSCode, but I also think it's useful for other IDEs and Flatpak applications that need Podman integration/interaction. >
- [X] Please attach a video showcasing the application on Linux using the Flatpak. [devcontainer.webm](https://github.com/user-attachments/assets/59e268f7-4456-49a8-8da0-f73b619cf3bd)
[podman-remote.webm](https://github.com/user-attachments/assets/00f9bad8-9dd9-4883-96c5-98ab173de41d)
- [X] I contacted upstream developers about this submission. **Link:** https://github.com/containers/podman/issues/28710


francoism90:
PhpStorm:

Screenshot_20260515_120117
Screenshot_20260515_120222

VScode (devcontainer):

image

francoism90:
bot, build

teamcons:
>     * [x]  Please attach a video showcasing the application on Linux using the Flatpak. < I don't know how to capture, I'll provide an image. >

theres a couple apps for this in flathub tho

francoism90:
@teamcons If it is required, I'm happy to look into this. I'm on KDE, maybe they already offer it themselves. I never done any screen capturing, otherwise I'll use my phone as backup.

teamcons:
If you are literally being told to do it, it probably means it is required, yeah.

francoism90:
@teamcons You learn something new every day, Spectacle can do this just fine. :)

I've added two videos, one with the extension enabled, and the one me using a devcontainer.

teamcons:
Alright thanks. You can attach it to the OG post, else the bot will not pick it up.
Edit and attach files.

Have a nice day.

hfiguiere:
see also #7415

francoism90:
@hfiguiere This time without the AI, and sorry for my comments earlier.

org.freedesktop.Sdk.Extension.podman.yml
```diff
@@ -0,0 +1,49 @@
...
+separate-locales: false
+appstream-compose: false
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.golang
+build-options:
+  no-debuginfo: true
+  strip: true
+  prefix: /usr/lib/sdk/podman
+  append-path: /usr/lib/sdk/golang/bin
+  prepend-path: /usr/lib/sdk/podman/bin
```
bbhtt: you only need one of this

org.freedesktop.Sdk.Extension.podman.yml
```diff
@@ -0,0 +1,49 @@
...
+        tag: v5.8.2
+        commit: 5b263b5f5b48004a87caac44e67349a8266d9ef4
+        x-checker-data:
+          type: json
+          url: https://api.github.com/repos/containers/podman/releases/latest
+          tag-query: .tag_name
+          timestamp-query: .published_at
+          version-query: $tag | sub("^v"; "")
+
+  - name: bundle-setup
```
bbhtt: merge this as post-install to above module.

org.freedesktop.Sdk.Extension.podman.yml
```diff
@@ -0,0 +1,49 @@
...
+  - org.freedesktop.Sdk.Extension.golang
+build-options:
+  no-debuginfo: true
+  strip: true
+  prefix: /usr/lib/sdk/podman
+  append-path: /usr/lib/sdk/golang/bin
+  prepend-path: /usr/lib/sdk/podman/bin
+  prepend-ld-library-path: /usr/lib/sdk/podman/lib
+  prepend-pkg-config-path: /usr/lib/sdk/podman/lib/pkgconfig
+cleanup:
```
bbhtt, line 15: it's go, i doubt any of these are being produced?

org.freedesktop.Sdk.Extension.podman.yml
```diff
@@ -0,0 +1,49 @@
...
+  - "/include"
+  - "/lib/cmake"
+  - "/lib/pkgconfig"
+  - "*.a"
+  - "*.la"
+modules:
+  - name: podman
+    buildsystem: simple
+    build-commands:
+      - make PREFIX=${FLATPAK_DEST} podman-remote
```
bbhtt: why are you building only the remote-client?

org.freedesktop.Sdk.Extension.podman.yml
francoism90: @bbhtt Could you show me what you mean? :)

I assume you meant it like this:

```
build-commands:
  - make PREFIX=${FLATPAK_DEST} podman-remote
post-install:
  - install -Dm755 bin/podman-remote -t ${FLATPAK_DEST}/bin/
```

org.freedesktop.Sdk.Extension.podman.yml
francoism90: Wouldn't that require a lot of deps? https://podman.io/docs/installation#building-from-source

I thought these would be required for the podman bin to work:

```
conmon
containers-common
crun
iptables
netavark
nftables
slirp4netns
```

There is another PR for the VSCode SDK that seems to provide `podman` and `podman-remote`: https://github.com/flathub/com.visualstudio.code.tool.podman/pull/116/changes

I can merge them into this one (without the forked repo - leaving as it is). What are your thoughts about that one? This would provide `podman`, `podman-remote` and `podman-compose`. But I would be actually surprised both bins would work without any deps.

Edit: when you only build podman, it seems to not need all these deps (see comment)

francoism90:
bot, build

org.freedesktop.Sdk.Extension.podman.yml
```diff
@@ -23,8 +23,9 @@ modules:
   - name: podman
     buildsystem: simple
     build-commands:
-      - make PREFIX=${FLATPAK_DEST} podman-remote
+      - make PREFIX=${FLATPAK_DEST} podman podman-remote
```
francoism90, line 27: If I do `make` and `make install`, everything get's pulled. This means docs, quadlets, networking, etc.

francoism90:
bot, build

org.freedesktop.Sdk.Extension.podman.yml
francoism90: I also added podman-compose and version checking. For now I would just point to containers/podman-compose, and hopefully they can fix it in a new release soon. I can comment in the issue if you want.

francoism90:
```
sh-5.3$ podman-compose 
  File "/usr/lib/sdk/podman/bin/podman-compose", line 3, in <module>
    from podman_compose import main
ModuleNotFoundError: No module named 'podman_compose'
```

What's the general recommendation for `PYTHONPATH`? It doesn't seem to find the modules. I can inject it with a patch, but I thought this would be picked up by python itself?

bam80: > There is another PR for the VSCode SDK that seems to provide `podman` and `podman-remote`: https://github.com/flathub/com.visualstudio.code.tool.podman/pull/116/changes

Hi, thanks for working on this.
It was just a quick and dirty hack, we could actually do without full blown podman there (but `podman-compose` is still needed!), see https://github.com/flathub/com.visualstudio.code.tool.podman/pull/116#pullrequestreview-4397163557.

I'm personally doubt if there is a use case for podman in flatpak, as running container _inside_ the flatpak is probably not what you want.
So `podman` to `host-spawn` link and `podman-remote` would probably do the job instead.
Correct me if I'm wrong.

org.freedesktop.Sdk.Extension.podman.yml
bam80: I doubt we need podman, see above:
https://github.com/flathub/flathub/pull/8677#discussion_r3337749596

francoism90: @bam80 Could it be used to build a container inside container? It does also help for completition maybe?

I also agree with your views, because I don't know how useful it is when you don't include the other parts/services.

org.freedesktop.Sdk.Extension.podman.yml
francoism90: @bam80 Hmm seems https://github.com/containers/podman-compose/pull/1435 is the blocker here.

It indeed does work for me when I patch the path, but I don't really like applying patches and other stuff to make things work as well. Hopefully the PR get's merged soon, or we should include your patch set in the PR? I would recommend against cross/fork repo's for this. :)

bam80: We definitely need access to `podman-compose`, in one form or another.
If we don't want to include it here (due the patch you linked, but it doesn't seem as a big problem to me), we could compensate it by calling `podman-compose` from host. It's not very good solution also, as you might just do not have it installed there.

For this, we would need `host-spawn` and `flatpak-spawn`, which you probably want anyway to run `podman`, etc.
How about include them here in the extension?

org.freedesktop.Sdk.Extension.podman.yml
bam80: > build a container inside container

This seems like an odd thing to do, I personally don't need it. But we could add it later if there is a demand.

> It does also help for completition maybe?

It could also play a negative role, as it might make an impression it's needed for things to work, but it isn't.
So I would rather exclude `podman` for now.

org.freedesktop.Sdk.Extension.podman.yml
```diff
@@ -0,0 +1,73 @@
...
+        x-checker-data:
+          type: json
+          url: https://api.github.com/repos/containers/podman/releases/latest
+          tag-query: .tag_name
+          timestamp-query: .published_at
+          version-query: $tag | sub("^v"; "")
+      - type: file
+        path: org.freedesktop.Sdk.Extension.podman.metainfo.xml
+
+  - name: podman_compose
```
bam80, line 46: If we include podman_compose, we also need to patch it for now?
But I don't see the patch below.

org.freedesktop.Sdk.Extension.podman.yml
```diff
@@ -0,0 +1,73 @@
...
+        sha256: 1d8214789a24de455a8b8bd8ae6fe3c6b69a5e3d64aa8a8e5d68e694bbcb285a
+        x-checker-data:
+          type: pypi
+          name: python-dotenv
+      - type: file
+        url: https://files.pythonhosted.org/packages/05/8e/961c0007c59b8dd7729d542c61a4d537767a59645b82a0b521206e1e25c2/pyyaml-6.0.3.tar.gz
+        sha256: d76623373421df22fb4cf8817020cbb7ef15c725b9d5e45f17e189bfc384190f
+        x-checker-data:
+          type: pypi
+          name: pyyaml
```
bam80, line 77: In the future, we will probably need some CI help to handle the boilerplate above.

README.md
```diff
@@ -0,0 +1,83 @@
...
+```bash
+systemctl --user enable podman.socket --now
+flatpak override --user --filesystem=xdg-run/podman:ro app-id
+```
+
+The socket path should be available inside the Flatpak application:
+
+```bash
+/run/user/$USER_ID/podman/podman.sock
+```
```
bam80, line 22: The variant with host-spawn podman(-compose) should also be mentioned here.
It doesn't require the podman.sock.

README.md
```diff
@@ -0,0 +1,83 @@
...
+```bash
+flatpak override --user --filesystem=xdg-run/podman:ro com.visualstudio.code
+```
+
+Open VSCode, run command `Open User Settings (JSON)` and append:
+
+```json
+"dev.containers.dockerPath": "/usr/lib/sdk/podman/bin/podman-remote",
+"dev.containers.dockerSocketPath": "/run/user/$USER_ID/podman/podman.sock",
+"docker.dockerPath": "/usr/lib/sdk/podman/bin/podman-remote"
```
bam80, line 48: podman-compose should be added, too.

org.freedesktop.Sdk.Extension.podman.yml
francoism90: @bam80 That's something the user can opt-in for. As far I'm aware, SDK shouldn't break the sandbox.

org.freedesktop.Sdk.Extension.podman.yml
francoism90: I mean your adjust for the shell path. We could at a simple sed or diff patch.

README.md
francoism90: Doesn't that break the sandbox?

org.freedesktop.Sdk.Extension.podman.yml
bam80: What do you mean by "opt-in"? Without host-spawn/flatpak-spawn utils in the SDK, it would be non-trivial for users to adopt them.
They were designed specifically for flatpak, and here is the good place for them for the reasons outlined above.

It's just luckily we have these utils in VS Code flatpak, but e.g. Qt Creator doesn't have them.
So we need central place for it anyway.

org.freedesktop.Sdk.Extension.podman.yml
bam80: That should work too, the other (and maybe better) alternative is to have a patch to be applied on top.
That looks more declarative to me.

README.md
bam80: It does, still I think it's good alternative here as the socket passthrough is quite cumbersome.
So it worth mentioning, at least.

bam80: The other argument I didn't mention about - it's not easy to force local podman-compose work remotely:
https://github.com/containers/podman-compose/issues/1432
That holds true even if you have podman installed in sdk:
https://github.com/containers/podman-compose/issues/849

By the way, I think we should document [this](https://github.com/containers/podman-compose/issues/849#issuecomment-4200079859) too:
> The other possibility is to set remote = true in [containers.conf](https://github.com/containers/container-libs/blob/main/common/docs/containers.conf.5.md).

As you see, it's yet another hack needed to make it work without breaking the sandbox.
Alternatively, maybe we could link podman to podman-remote to trick podman-compose work remotely.

The point is - there is no good, out of the box way anyway, so we should document all the possibilities.

README.md
```diff
@@ -0,0 +1,83 @@
...
+
+```bash
+flatpak override --user --filesystem=xdg-run/podman:ro com.visualstudio.code
+```
+
+Open VSCode, run command `Open User Settings (JSON)` and append:
+
+```json
+"dev.containers.dockerPath": "/usr/lib/sdk/podman/bin/podman-remote",
+"dev.containers.dockerSocketPath": "/run/user/$USER_ID/podman/podman.sock",
```
bam80: Do we need to replace $USER_ID with the actual UID here?

README.md
francoism90: done

README.md
francoism90: Yes, I can change it to UID.

francoism90:
Add sed, this fixes the podman-compose path.

@bbhtt Hopefully you can take a look. :)

org.freedesktop.Sdk.Extension.podman.yml
francoism90: If you use flatpak host spawn, you bypass the sandbox: https://man7.org/linux/man-pages/man1/flatpak-spawn.1.html

org.freedesktop.Sdk.Extension.podman.yml
francoism90: You can do this, for example to build a Podman container inside your IDE.

org.freedesktop.Sdk.Extension.podman.yml
francoism90: Fixed, see sed.

README.md
francoism90: You shouldn't set that:

```man
remote = false

Indicates whether the application should be running in remote mode. This flag modifies the --remote option on container engines. Setting the flag to true will default podman --remote=true for access to the remote Podman service.
```

A socket is way faster. I think remote mode changes this to only use SSH connections.

bam80: That's needed even for socket - otherwise, podman tries to work locally.

org.freedesktop.Sdk.Extension.podman.yml
```diff
@@ -0,0 +1,78 @@
...
+        path: org.freedesktop.Sdk.Extension.podman.metainfo.xml
+
+  - name: podman_compose
+    buildsystem: simple
+    build-commands:
+      - pip3 install --prefix=${FLATPAK_DEST} --no-index --no-build-isolation --find-links=. .
+    post-install:
+      - |
+        PYTHON_VERSION=$(python3 -c "import sys; print(f'{sys.version_info.major}.{sys.version_info.minor}')")
+        sed -i "2i import sys; sys.path.insert(0, '${FLATPAK_DEST}/lib/python${PYTHON_VERSION}/site-packages')" \
```
bbhtt: why is this sed needed?

bbhtt:
if this is going to be accepted then the vscode extension needs to be archived. So I suggest testing it there to see if it works with vscode.

bbhtt:
Also can you mark the threads as solved when they are done? I haven't followed the discussion.

README.md
```diff
@@ -0,0 +1,85 @@
...
+flatpak override --user --filesystem=xdg-run/podman:ro app-id
+```
+
+The socket path should be available inside the Flatpak application:
+
+```bash
+/run/user/$UID/podman/podman.sock
+```
+
+> Note: replace `$UID` with the current user-id (`id -u`) and make sure the path exists on the host.
```
bam80: There is no need to replace the `$UID` here, as it works perfectly with the variable substitution.
So the "replace" note should be rather moved down to the VS Code Settings clause.

README.md
```diff
@@ -0,0 +1,85 @@
...
+flatpak override --user --filesystem=xdg-run/podman:ro com.visualstudio.code
+```
+
+Open VSCode, run command `Open User Settings (JSON)` and append:
+
+```json
+"containers.composeCommand": "/usr/lib/sdk/podman/bin/podman-compose",
+"dev.containers.dockerComposePath": "/usr/lib/sdk/podman/bin/podman-compose",
+"dev.containers.dockerPath": "/usr/lib/sdk/podman/bin/podman-remote",
+"dev.containers.dockerSocketPath": "/run/user/$UID/podman/podman.sock",
```
bam80:
```suggestion
"dev.containers.dockerSocketPath": "/run/user/<UID>/podman/podman.sock",
```
And move the note about UID replacement from the above to here.
(Assuming you checked $UID won't work here).

README.md
```diff
@@ -0,0 +1,85 @@
...
+}
+```
+
+Other useful `runArgs` may be `--network=systemd-networkname` to allow network integration, and `--security-opt=label=disable` for SELinux to prevent setting labels.
+
+It may be required for certain devcontainers to force the Docker format:
+
+```json
+{
+  "runArgs": ["--userns=keep-id", "--init"],
```
bam80, line 71: Just a question - people complain this is not compatible to docker, can you confirm?

org.freedesktop.Sdk.Extension.podman.yml
bam80: I wonder too.
@francoism90 If you saw my [branch](https://github.com/bam80/podman-compose/tree/patch-1), the only patch needed is [this](https://github.com/containers/podman-compose/compare/main...bam80:podman-compose:patch-1) one:
```diff
   -     mount_src = os.path.realpath(os.path.join(basedir, os.path.expanduser(mount_src)))
   +     mount_src = os.path.abspath(os.path.join(basedir, os.path.expanduser(mount_src)))
```
You can make it with sed, sure, but I would personally prefer applying the patch file.

PS:
At least that was the case with my VS Code extention:
https://github.com/flathub/com.visualstudio.code.tool.podman/pull/116
Maybe with SDK extension the situation is different, I don't know.

org.freedesktop.Sdk.Extension.podman.yml
bam80: That looks pointless - if you build the container inside your IDE, you won't have it available on the host and other flatpaks, isn't it?
And your container will be lost as soon as you uninstall your IDE flatpak with it's data.

README.md
```diff
@@ -0,0 +1,85 @@
+# org.freedesktop.Sdk.Extension.podman
+
+This extension adds Podman support for Flatpak applications.
+
+To use the Podman SDK, set the following environment variable per application:
+
+```bash
+FLATPAK_ENABLE_SDK_EXT=podman
```
bam80, line 8: This apparently is VS Code specific?
https://github.com/flathub/com.visualstudio.code/blob/25f9c9033ccaff29bef83dc9f80a1b869580009c/code.sh#L47
Should make a note if so.

org.freedesktop.Sdk.Extension.podman.yml
francoism90: @bbhtt This solves the path error when you run `podman-compose`. It seems they use a hardcoded path after compiling podman-compose.

@bam80 Could you please suggest your changes as PR on my repo? :)

README.md
francoism90: The user-id can be completly different, for example systemd-user has a really large uid. If you have any suggestion in making this more clear, please let me know. :)

You can make a PR to my repo or a diff here.

README.md
francoism90: @bam80 I replaced everywhere with <UID>, hope it's more clear. :)

README.md
francoism90: That's true, this only works Podman. It's assumed you run podman (or with docker-podman if you really need it).

When you're a docker user, you should omit the `userns=keep-id`.

README.md
francoism90: @bam80 No, this is per application. On PHPStorm you also need to set this (or whatever application you use).

https://github.com/search?q=org%3Aflathub%20FLATPAK_ENABLE_SDK_EXT&type=code

I'm not an expert, but I believe this is a reserved Flatpak env variable? It should also with Zed for example. I don't use Zed, so I cannot test it.

README.md
francoism90: That's the idea, it keeps using the local podman instance.

You should only set this, if you host your podman service somewhere else.

org.freedesktop.Sdk.Extension.podman.yml
bam80: @francoism90 Please see how it's done in my repo, but you are apparently aware of that:
https://github.com/flathub/com.visualstudio.code.tool.podman/pull/116/changes#r3068705597

I can't rework it to the patch myself currently, sorry.

PS:
> This solves the path error when you run podman-compose. It seems they use a hardcoded path after compiling podman-compose.

I didn't face this problem with my VSC extension above. You can install it and make sure yourself.

README.md
bam80: Wait a minute - you use `podman-remote` and set up the socket to be able to talk to the host's podman service ("remote" one), isn't it?
So podman-compose should also talk to it, not any "nested" flatpak podman instance.

org.freedesktop.Sdk.Extension.podman.yml
francoism90: I'm targeting a higher SDK, that could be the issue.

I'll leave it for now, the mount patch, seems to be needed for volumes. I think upstream should fix this.

bbhtt: can you try using a patch for this instead of sed?

francoism90:
image

image

image

@bbhtt I don't know if every `podman` or `podman-compose` command works. The `podman-remote` command works, and I also tested the devcontainers.

I have marked most as solved, as I think most are escaping the sandbox, which is not really covered/purpose of the SDK. Please let me know if you want me to test more, so far I tested VSCode and PHPStorm, it seems more editors could use it now.

org.freedesktop.Sdk.Extension.podman.yml
bam80: > I'll leave it for now, the mount patch, seems to be needed for volumes. I think upstream should fix this.

That patch is actually essential for flatpak - you can't mount volumes in the flatpak without it.
So I suggest just use 2 patches for now - your one (if it really needed) and my.

bbhtt:
Ok that's fine. I had one last comment about the sed.

README.md
bam80: > I'm not an expert, but I believe this is a reserved Flatpak env variable? 

I didn't find it anywhere in the Flatpak doc. So it looks like it's just copied from app to app, so wording here could be:
"For the apps explicitly supporting FLATPAK_ENABLE_SDK_EXT ..."

@bbhtt Do you have some insight here?

README.md
```diff
@@ -0,0 +1,83 @@
...
+
+```bash
+systemctl --user enable podman.socket --now
+flatpak override --user --filesystem=xdg-run/podman:ro app-id
+```
+
+The socket path should be available inside the Flatpak application:
+
+```bash
+/run/user/<UID>/podman/podman.sock
```
bam80:
```suggestion
$XDG_RUNTIME_DIR/podman/podman.sock
```
I believe even if UID differs, the variable expansion should point to the right place here both in the flatpak and host.
Also, it's copy-pasteable variant.

README.md
bbhtt: it's not a Flatpak env var, it comes from ide-flatpak-wrapper https://github.com/flathub-infra/ide-flatpak-wrapper

Apps would be able to use the env var if they are using that which most IDE oriented apps in Flathub do.

README.md
bam80: @francoism90 could you reopen this thread, please? I believe it's essential.
We shouldn't end up with the mess of local and remotes containers.

If you want to have both, you should document the both way, so `remote = true` trick above should be present in the doc for the socket way.

@bbhtt

bbhtt: reopened, can you also post a link to the thread in the comment, if you want it reopened?

Looks like Github stopped doing that in the email so I have to look through every resolved thread to find it.

README.md
bam80: > When you're a docker user, you should omit the userns=keep-id.

Could we explicitly say it here, please? It's not obvious at all.

README.md
francoism90: @bam80 Yes, I haven't tested this scenario, but I actually think you can host your podman instance somewhere else (e.g. cloud). They can talk with each other.

The current only is focused on making it work locally, like when you use Fedora Silverblue. It includes Podman, but no other tools.

org.freedesktop.Sdk.Extension.podman.yml
francoism90: @bbhtt Done! It's much more clear now, and I also add a comment what it fixes. :)

@bam80 The error you're showing is because of it assumes it's remote. Can you test with the default values? I also see tests are still missing in the PR? I think it's better to solve this upstream, and wait till they release a new version (or commit).

README.md
francoism90: I assume it's correct in the README? :)

README.md
```diff
@@ -0,0 +1,85 @@
+# org.freedesktop.Sdk.Extension.podman
+
+This extension adds Podman support for Flatpak applications.
+
+To use the Podman SDK, set the following environment variable per application:
```
bam80, line 5:
```suggestion
To use the Podman SDK in IDE apps from Flathub, set the following environment variable per application:
```
@bbhtt I assume it's more correct then?

bbhtt: seems fine.

README.md
bam80: Almost. I created a new nitpick above. Unfortunately, I can't resolve/unresolve myself here.
@bbhtt

README.md
francoism90: Yeah, I assume you can use the SDK also in none IDE applications later. They just have to expose that env.

README.md
```diff
@@ -0,0 +1,85 @@
...
+"docker.dockerPath": "/usr/lib/sdk/podman/bin/podman-remote"
+```
+
+> Note: Replace <UID> with the user-id that runs the socket.
+
+Restart the editor to apply changes.
+
+### Devcontainers
+
+Update the project `devcontainer.json` file with `runArgs` that apply to Podman:
```
bam80, line 57:
```suggestion
Update the project `devcontainer.json` file with `runArgs`.
NOTE: `--userns` won't work in Docker.
```
Also, could you add a note why these two are needed? I didn't face the need of `--init`, personally.

francoism90: @bam80 Passing --init to devcontainers is highly recommended. To my knowledge this prevents zombie processes and a layer of protection against file corruption. 

Microsoft devcontainers do pass this already, but it shouldn't hurt if it's also passed on top of that.

README.md
bam80: @francoism90 You can, but for non-IDE (non-Flathub?) the `FLATPAK_ENABLE_SDK_EXT` variable is not applicable. And also it's not documented.
This creates a confusion, that is why the nitpick above.

org.freedesktop.Sdk.Extension.podman.yml
bam80: @francoism90 

We have following scenarios here:
1. (default) Using UNIX socket. This is documented here and most commonly used one.
    _This is currently broken in two various ways_ (I tested it):
    1. podman/podman-compose:
        podman won't use socket by default, without `-r` option or **`remote = true` in the containers.conf** file (the latter is also needed for podman-compose), which I'm asking to document here.
    2. podman-compose:
        Upstream **podman-compose without the volume patch breaks volumes mount** in the flatpak containers.
        The volumes are essential, and we are in flatpak here.
        So we don't need to wait when upstream adopts the patch, the patch is sane and makes it possible to use podman-compose at all, so it should be included, too.

The others scenarios are:
2. Fully "local" podman/podman-compose (container is nested to flatpak).
    You implicitly introduced it by including full `podman` binary. I was against that but you do what you want.
    I didn't test this scenario, and can't do that currently.
3. host-spawn podman.
    Most easy to setup one - doesn't require the patched podman-compose and `-r`/`remote = true` option.
    I asked to document it but you rejected it. Well, can do nothing.
    Works nice. The only problem I faced is hard to stop the build process inside the dev container. It's yet to be sure if the problem is specific to this scenario, or common.
5. Fully remote.
    When you use SSH or TCP socket instead of UNIX one, to work over network.
    I didn't test this one, but should be equivalent to the UNIX socket scenario.

_So the most common and useful scenario is currently broken._
_@francoism90 Please make sure to address the most important 2 issues above before this could be accepted. Thanks._

@bbhtt

francoism90: 1. Podman compose is broken yes, but that's not 'our' fault here.

You have submitted the PR upstream to fix it. They also asked for tests and feedback. Both should be solved upstream (in my opinion).

The socket is being used on VSCode over here, it doesn't work without the socket at all. What's your setup? Also if the user need or wants this, it's an opt-in, not something that should be forced. The instructions are as is, they can be differ for everyone.

2. They are offered as is. Some things may be unsupported or won't work at all. The including of the binaries is still useful. For example, it can parse the options and it can also be used to run podman commands in the IDE. Both are untested, may work, but it's not full podman, as these require stuff for networking and mostly containers-common I think.

@bbhtt Also asked me to include it as is. Which I think is perfectly fine and reasonable.

3. host-spawn simply works around the sandbox. You execute the command on the host, which basically is a security risk. @bbhtt should correct me on this, but this warning and usage is already documented over here: https://github.com/flathub/com.visualstudio.code/blob/master/flatpak-warning.txt (you see it on first open)

You can use host-spawn, but it's has nothing to do with this SDK. It's simply a thing that you can do, but should do with caution. To my knowledge, you can simply devcontainers without host spawn over a socket (I'm doing this for 2 years now).

5. This should work, but it's something untested. In most cases the user will mount over SSH (VSCode does this) and run the container(s) locally. You can run it remotely, but I don't think that's very useful for Flatpaks?

bam80:
> host-spawn simply works around the sandbox. You execute the command on the host, which basically is a security risk. @bbhtt should correct me on this, but this warning and usage is already documented over here: https://github.com/flathub/com.visualstudio.code/blob/master/flatpak-warning.txt (you see it on first open)

@francoism90 thank you for pointing me to that doc.
We should probably adjust it for our SDK extension, as the old `com.visualstudio.code.tool.podman` will be swapped to it:

> To get support for additional languages or tools within the Flatpak, you have to install SDK extensions, e.g.
 $ flatpak install flathub org.freedesktop.Sdk.Extension.dotnet
  $ flatpak install flathub org.freedesktop.Sdk.Extension.golang
  **$ FLATPAK_ENABLE_SDK_EXT=dotnet,golang flatpak run com.visualstudio.code**

> Similarly for shells and other tools:
> 
>   $ flatpak install com.visualstudio.code.tool.fish
>   **$ flatpak install com.visualstudio.code.tool.podman**

I made a draft PR:
https://github.com/flathub/com.visualstudio.code/pull/681

org.freedesktop.Sdk.Extension.podman.yml
bam80: > 1. Podman compose is broken yes, but that's not 'our' fault here.
> 
> 
> You have submitted the PR upstream to fix it. They also asked for tests and feedback. Both should be solved upstream (in my opinion).

It should be solved either upstream or here, I wouldn't pass this PR without it, for a good reason.
This problem is hard to debug, I found it myself in a hard way, so wouldn't want anyone to repeat my steps.

But we have already patched downstream with your path patch for our needs.
The second patch is nevertheless important - it literally allows to use podman-compose in our environment.
So I see no problem to apply it downstream, too.

PS:
I just saw the patch was merged upstream:
https://github.com/containers/podman-compose/pull/1465
One problem less, congrats!

org.freedesktop.Sdk.Extension.podman.yml
```diff
@@ -0,0 +1,78 @@
...
+    buildsystem: simple
+    build-commands:
+      - pip3 install --prefix=${FLATPAK_DEST} --no-index --no-build-isolation --find-links=. .
+    post-install:
+      - patch ${FLATPAK_DEST}/bin/podman-compose 10-podman-compose-syspath.patch
+    sources:
+      - type: git
+        url: https://github.com/containers/podman-compose.git
+        tag: v1.5.0
+        commit: f7eeda1a3db10952424af6a5b0501c269ebe3f0d
```
bam80: Let's bump to the latest release with the volume fix:
https://github.com/containers/podman-compose/releases/tag/v1.6.0

```suggestion
        tag: v1.6.0
        commit: 0f6537e9cfa38f6035ac57c1716b6d55dbaf3ca4
```

README.md
```diff
@@ -0,0 +1,85 @@
...
+flatpak override --user --filesystem=xdg-run/podman:ro com.visualstudio.code
+```
+
+Open VSCode, run command `Open User Settings (JSON)` and append:
+
+```json
+"containers.composeCommand": "/usr/lib/sdk/podman/bin/podman-compose",
+"dev.containers.dockerComposePath": "/usr/lib/sdk/podman/bin/podman-compose",
+"dev.containers.dockerPath": "/usr/lib/sdk/podman/bin/podman-remote",
+"dev.containers.dockerSocketPath": "/run/user/<UID>/podman/podman.sock",
```
bam80, line 47: Let me emphasize it the other way why the additional documentation is needed.

Here `dockerSocketPath` influences behavior of `dockerPath` only, when VSC uses it directly.
We can't force podman-compose (`dockerComposePath`) to obey this setting, too (neither VSC can).
Which is a pity, but this is what we have for now.

So we should document all the workarounds we found for this non-obvious problem. Otherwise, podman-compose just breaks.
The one I found and tested is adding `remote = true` to the containers.conf file.

@francoism90 Could you please try VS Code **compose-based Dev Container** yourself, make it work and document the workaround above and maybe what you'll find appropriate, before claiming this PR as ready? Or I can actually add the text myself, if you wish.

This is the last blocker remained to me.
Thanks.

francoism90: Why does or should it break? Enabling remote is an opt-in.

org.freedesktop.Sdk.Extension.podman.yml
```diff
@@ -0,0 +1,78 @@
...
+  - "/lib/pkgconfig"
+  - "/man"
+  - "*.a"
+  - "*.la"
+  - "*.pyc"
+modules:
+  - name: podman
+    buildsystem: simple
+    build-commands:
+      - make PREFIX=${FLATPAK_DEST} podman podman-remote
```
bam80, line 27: The `podman` binary here doesn't seem to work:
```console
sh-5.3$ podman images
Error: no such file or directory
sh-5.3$ podman container list 
Error: no such file or directory
```
That error results every podman command.
I don't know what the problem is, but I think we shouldn't supply `podman` binary in such a broken state.

Considering `podman-remote` fully supports podman capabilities and CLI interface, I still think it's better if we just drop the `podman` binary.

```suggestion
      - make PREFIX=${FLATPAK_DEST} podman-remote
```
PS:
It also would help to make podman-compose work easier.

org.freedesktop.Sdk.Extension.podman.yml
```diff
@@ -0,0 +1,78 @@
...
+    post-install:
+      - install -Dm755 bin/podman -t ${FLATPAK_DEST}/bin/
```
bam80, line 29: per above
```suggestion
```

org.freedesktop.Sdk.Extension.podman.yml
```diff
@@ -0,0 +1,78 @@
...
+        tag: v1.6.0
+        commit: 0f6537e9cfa38f6035ac57c1716b6d55dbaf3ca4
+        x-checker-data:
+          type: json
+          url: https://api.github.com/repos/containers/podman-compose/releases/latest
+          tag-query: .tag_name
+          timestamp-query: .published_at
+          version-query: $tag | sub("^v"; "")
+      - type: file
+        path: patches/10-podman-compose-syspath.patch
```
bam80, line 64: Why we can't just set PYTHONPATH or some other env var instead?

francoism90:
Can we please keep it as is? We just add the bins, and that's it.

I know `podman` or `podman-compose` may need workarounds or some configuration changes. This is not part of the PR. This simply adds the SDK, other adjusts may need to be done upstream.

@bbhtt Sorry for tagging you again, I hope this is fine now. I answered most questions. The patch seems to be needed, the workaround with the env, doesn't work.

bam80:
@francoism90 

> Can we please keep it as is? We just add the bins, and that's it.

Sorry but that's not enough. `podman-compose` doesn't work out of the box in the form we supply it, that's not on the surface, so we should provide documentation how to use it, it's doable.

> The patch seems to be needed, the workaround with the env, doesn't work.

Could you provide the details what's wrong with the PYTHONPATH or other env workarounds?
Patching the code just because of that seems too hacky, we should avoid that wherever possible.

francoism90:
@bam80 Try it out, you'll see it not works. The build commands and repo is available.

I'm not going to document everything, sorry. It's an SDK, it may be up to upstream if it's even supported. It's not up to Flatpak to fix everything.

org.freedesktop.Sdk.Extension.podman.yml
```diff
@@ -0,0 +1,78 @@
+id: "org.freedesktop.Sdk.Extension.podman"
+branch: "25.08"
+runtime: "org.freedesktop.Sdk"
+sdk: "org.freedesktop.Sdk"
+runtime-version: "25.08"
+build-extension: true
+separate-locales: false
+appstream-compose: false
```
bbhtt: you have to drop this

francoism90:
bot, build

org.freedesktop.Sdk.Extension.podman.yml
francoism90: @bbhtt It doesn't seem to build without these.

bbhtt: i meant the appstream one only the locale one is fine to keep.

the build error happens because you are missing build-extension

francoism90: I removed `appstream-compose`

bbhtt:
A patch is fine here as the python version will change when new branches are made.

What about podman-compose doesn't work / needs to be documented? What exactly needs to be documented?

If it's about documentation only I suggest to do it after the submission is accepted.

org.freedesktop.Sdk.Extension.podman.yml
bam80: @bbhtt This thread about broken `podman` is critical, and should be reopened, in my opinion:
https://github.com/flathub/flathub/pull/8677/changes#r3365586519

If podman doesn't work within the flatpak, there is no point to include it here - it brings no more than podman-remote then.

README.md
bam80: > [@francoism90](https://github.com/francoism90) could you reopen this thread, please? I believe it's essential. We shouldn't end up with the mess of local and remotes containers.
> 
> If you want to have both, you should document the both way, so `remote = true` trick above should be present in the doc for the socket way.

@bbhtt this wasn't addressed either, and should be reopened, too:
https://github.com/flathub/flathub/pull/8677/changes#r3349062336

`podman-compose` doesn't work here as we supply it, for local UNIX socket configuration, too.
To make it work, we either:
- (preferable) get rid of real podman binary and replace it with symbolic link to podman-remote. That should make `podman-compose` just work, with no additional tweaking needed.
There is no sacrifice as `podman` binary doesn't work without UNIX socket here, anyway. So `podman-remote` can fully substitute it.
- If we keep `podman` binary, the way to make `podman-compose` work becomes non-trivial, as it requires manual intervention now, so we should document it:
_The workaround I found is to set `remote = true` in [containers.conf](https://github.com/containers/container-libs/blob/main/common/docs/containers.conf.5.md)._
That's needed even for  _local_ UNIX socket, otherwise `podman-compose` tries to use flatpak-nested container and breaks, with no obvious reason and no way to fix via VS Code settings, etc.
_So it's essential to document it in this PR, I believe._

francoism90: Why would podman-remote be a symlink for podman? It's up to the user to configure or overrule this if needed. I don't see why this can be a problem, as previously it wasn't provided at all. 

The workaround is an opt-in, it changes the behaviour of the Podman daemon, it can affect current configurations.

Please tell me what I should include, but it shouldn't be mandatory for the SDK to work. Like I said a couple of times now, the user may need to adjust, which is the same for any SDK: rust, nodejs, etc. 

To be honest I'm kinda done with this discussion. I do understand things can be documented, but like @bbhtt already explained, this can be done after release, it's not a breaking thing (podman + podman-compose weren't even included before).

README.md
```diff
@@ -0,0 +1,82 @@
...
+systemctl --user enable podman.socket --now
+flatpak override --user --filesystem=xdg-run/podman:ro app-id
+```
+
+The socket path should be available inside the Flatpak application:
+
+```bash
+$XDG_RUNTIME_DIR/podman/podman.sock
+```
+
```
bam80, line 23:
```suggestion
WARNING:
podman/podman-compose won't work by default here, because nested containers in Flatpak are problematic.
They only can work in "remote" mode, e.g. to connect via UNIX socket to Podman service on host.

While podman has `--remote` option, podman-compose [doesn't have one](https://github.com/containers/podman-compose/issues/849).
So the only way to make it work is trick them to use remote mode by default.

There are two known possibilities to do so:
- Inside the Flatpak, create [containers.conf](https://github.com/containers/container-libs/blob/main/common/docs/containers.conf.5.md) file with following content:
>   ```console
>   $ cat $XDG_CONFIG_HOME/containers/containers.conf
>   [engine]
>   remote = true
>   ```
- Set the environment variable `CONTAINER_HOST`. You can even set it to an empty value, that will use Podman socket for the remote:
>   ```console
>   $ export CONTAINER_HOST=
>   $ podman-compose ...
>   # or
>   $ CONTAINER_HOST= podman-compose ...
>   ```

The other alternative is to run podman/podman-compose on host via host-spawn or flatpak-spawn in flatpak.
For VS Code flatpak and maybe others, you would also need to share `/tmp` directory to build Dev Containers there:
> ```console
> flatpak override --user --filesystem=/tmp com.visualstudio.code
> ```

```

README.md
```diff
@@ -0,0 +1,82 @@
...
+"containers.composeCommand": "/usr/lib/sdk/podman/bin/podman-compose",
+"containers.containerCommand": "/usr/lib/sdk/podman/bin/podman-remote",
+"dev.containers.dockerComposePath": "/usr/lib/sdk/podman/bin/podman-compose",
+"dev.containers.dockerPath": "/usr/lib/sdk/podman/bin/podman-remote",
+"dev.containers.dockerSocketPath": "/run/user/<UID>/podman/podman.sock",
+"docker.dockerPath": "/usr/lib/sdk/podman/bin/podman-remote"
+```
+
+> Note: Replace <UID> with the user-id that runs the socket.
+
```
bam80, line 52:
```suggestion
Note: For podman-compose, see the WARNING above.

```

README.md
francoism90: Can you please use a title instead? e.g. Interact with host containers

It's not a warning, because it depends on your setup. 

Also add /tmp to the devcontainers section, under the socket command.

bam80: Done:
https://github.com/francoism90/flathub/pull/2

francoism90:
I'm done with this, sorry.

I'm getting crazy with all the things that are not even this SDK related, and no one helping me with it.

If others want it, see my repo, and submit the SDK yourself to here.

PR #8688, opened 2026-05-16
Add io.github.hrniels.Eventix

hrniels:

- [x] Please describe the application briefly. Eventix is an iCalendar event and task manager for Linux desktops.
- [x] Please attach a video showcasing the application on Linux using the Flatpak. https://github.com/hrniels/Eventix/releases/download/v1.0.5/Eventix-Screencast.webm


hrniels:
The 'finish-args-arbitrary-dbus-access' error is expected, because Eventix needs dbus session access for its tray icon. As explained in the README, if the tray icon is not desired, dbus session access can be removed and Eventix will simply not create one. Unfortunately there does not seem to be a way to get proper tray icons (with on-click events) without that permission. If anybody knows a solution to that, please let me know.

Otherwise, how do we proceed here?

io.github.hrniels.Eventix-Import.desktop
CodedOre, line 1: This file is not specific to Flatpak or Flathub. It should therefore come with the source code from the upstream repository.

io.github.hrniels.Eventix.desktop
CodedOre, line 1: This file is not specific to Flatpak or Flathub. It should therefore come with the source code from the upstream repository.

io.github.hrniels.Eventix.metainfo.xml
CodedOre, line 1: This file is not specific to Flatpak or Flathub. It should therefore come with the source code from the upstream repository.

io.github.hrniels.Eventix.json
```diff
@@ -0,0 +1,154 @@
+{
+    "app-id": "io.github.hrniels.Eventix",
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "50",
+    "sdk": "org.gnome.Sdk",
```
CodedOre, line 5: Why the use of the GNOME runtime in specific?\
Would it be possible to use the Freedesktop runtime instead?

io.github.hrniels.Eventix.json
```diff
@@ -0,0 +1,154 @@
...
+        "org.freedesktop.Sdk.Extension.rust-stable",
+        "org.freedesktop.Sdk.Extension.openjdk"
+    ],
+    "command": "eventix-app",
+    "finish-args": [
+        "--share=network",
+        "--share=ipc",
+        "--device=dri",
+        "--socket=x11",
+        "--socket=session-bus",
```
CodedOre: Access to the complete session bus is not permitted. 

Find out which bus names you actually need and only request those:

https://docs.flatpak.org/en/latest/debugging.html#audit-session-or-system-bus-traffic

io.github.hrniels.Eventix.json
```diff
@@ -0,0 +1,154 @@
...
+        "--filesystem=xdg-run/pipewire-0",
```
CodedOre: Why access to the Pipewire socket? Why not use  the Pulseaudio socket?

io.github.hrniels.Eventix.json
```diff
@@ -0,0 +1,154 @@
...
+        "--talk-name=org.kde.StatusNotifierWatcher",
+        "--talk-name=org.freedesktop.Notifications",
+        "--talk-name=org.freedesktop.secrets"
```
CodedOre: Secrets and notifications should run over the portals. Drop these permissions.

io.github.hrniels.Eventix.json
```diff
@@ -0,0 +1,154 @@
...
+            "CARGO_HOME": "/app/cargo-home",
+            "CARGO_NET_OFFLINE": "true"
+        }
+    },
+    "modules": [
+        {
+            "name": "cargo-vendor",
+            "sources": [
+                "cargo-sources.json"
+            ],
```
CodedOre: The cargo sources should be part of the application module, not its own.

io.github.hrniels.Eventix.json
```diff
@@ -0,0 +1,154 @@
...
+            "buildsystem": "simple",
+            "build-commands": [
+                "mkdir -p /app/cargo-vendor",
+                "cp -r cargo/vendor/. /app/cargo-vendor/"
```
CodedOre: Do not install build dependencies into the application folder.

io.github.hrniels.Eventix.json
```diff
@@ -0,0 +1,154 @@
...
+        },
+        {
+            "name": "Eventix",
+            "buildsystem": "simple",
+            "build-commands": [
+                "mkdir -p .cargo",
+                "echo '[source.crates-io]' > .cargo/config.toml",
+                "echo 'replace-with = \"vendored-sources\"' >> .cargo/config.toml",
+                "echo '[source.vendored-sources]' >> .cargo/config.toml",
+                "echo 'directory = \"/app/cargo-vendor\"' >> .cargo/config.toml",
```
CodedOre: Never write files using shell commands. Use a file or inline source instead. 

Also, the `flatpak-cargo-generator` already creates that file. If you would use the generator as expected you would not need this.

io.github.hrniels.Eventix.json
```diff
@@ -0,0 +1,154 @@
...
+                "mkdir -p /app/build-artifacts",
+                "cp target/release/eventix /app/build-artifacts/",
+                "cp target/release/eventix-app /app/build-artifacts/",
+                "cp target/release/eventix-import /app/build-artifacts/",
+                "cp target/release/eventix-keyring /app/build-artifacts/",
+                "install -Dm755 /app/build-artifacts/eventix -t /app/bin/",
+                "install -Dm755 /app/build-artifacts/eventix-app -t /app/bin/",
+                "install -Dm755 /app/build-artifacts/eventix-import -t /app/bin/",
+                "install -Dm755 /app/build-artifacts/eventix-keyring -t /app/bin/",
+                "rm -rf /app/build-artifacts",
```
CodedOre: Why the unnecessary copy to `/app/build-artifacts` instead of directly installing the files?

io.github.hrniels.Eventix.json
```diff
@@ -0,0 +1,154 @@
...
+                "install -Dm644 data/icons/48x48.png /app/share/icons/hicolor/48x48/apps/io.github.hrniels.Eventix.png",
+                "install -Dm644 data/icons/index.theme /app/share/icons/hicolor/index.theme",
+                "install -Dm644 flatpak/io.github.hrniels.Eventix.metainfo.xml -t /app/share/metainfo/",
+                "install -Dm644 LICENSE -t /app/share/licenses/io.github.hrniels.Eventix/eventix",
+                "mkdir -p /app/share/io.github.hrniels.Eventix/locale",
+                "mkdir -p /app/share/io.github.hrniels.Eventix/static",
+                "mkdir -p /app/share/io.github.hrniels.Eventix/icons",
+                "cp -r data/locale /app/share/io.github.hrniels.Eventix",
+                "cp -r data/static /app/share/io.github.hrniels.Eventix",
+                "cp -r data/icons /app/share/io.github.hrniels.Eventix",
```
CodedOre: You should create a buildsystem upstream that serves as a singular source of truth for the installation process and use that here.

io.github.hrniels.Eventix.json
```diff
@@ -0,0 +1,154 @@
...
+                "gtk-update-icon-cache /app/share/icons/hicolor",
+                "rm -rf /app/cargo-vendor",
+                "rm -rf /app/cargo-home"
```
CodedOre: This should not be needed.

io.github.hrniels.Eventix.json
```diff
@@ -0,0 +1,154 @@
...
+            "sources": [
+                {
+                    "type": "git",
+                    "url": "https://github.com/hrniels/eventix.git",
+                    "commit": "746c1e3ba4d2e3c4a85b20d42c4a0887fb4e24c6",
+                    "tag": "v1.0.5",
+                    "disable-submodules": true
+                }
+            ]
+        },
```
CodedOre: All the modules below this should come before the application module.

Your application module should be the last.

io.github.hrniels.Eventix.json
```diff
@@ -0,0 +1,154 @@
...
+            "name": "davmail",
+            "buildsystem": "simple",
+            "build-options": {
+                "env": {
+                    "JAVA_HOME": "/usr/lib/sdk/openjdk",
+                    "PATH": "/usr/lib/sdk/openjdk/bin:/app/bin:/usr/bin"
+                }
+            },
+            "build-commands": [
+                "export PATH=/usr/lib/sdk/openjdk/bin:$PATH",
```
CodedOre: You already set that in the build options. Why do you set it again?

io.github.hrniels.Eventix.json
```diff
@@ -0,0 +1,154 @@
...
+                    "url": "https://github.com/hrniels/davmail.git",
+                    "commit": "3a4751992a44e1c36370ddd525fad8c8534e4239",
+                    "disable-submodules": true
+                },
+                "java-sources.json"
+            ]
+        },
+        {
+            "name": "java-runtime",
+            "buildsystem": "simple",
```
CodedOre: Why do you install your own Java runtime?\
Why do you not use the runtime from the OpenJDK extension?

io.github.hrniels.Eventix.json
```diff
@@ -0,0 +1,154 @@
...
+            "build-commands": [
+                "mkdir -p /app/openjdk",
+                "tar xzf OpenJDK17U-jre_x64_linux_hotspot_17.0.9_9.tar.gz -C /app/openjdk --strip-components=1",
+                "ln -s /app/openjdk/bin/java /app/bin/java"
+            ],
+            "sources": [
+                {
+                    "type": "file",
+                    "url": "https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.9+9/OpenJDK17U-jre_x64_linux_hotspot_17.0.9_9.tar.gz",
```
CodedOre: What happens on aarch64?

io.github.hrniels.Eventix.json
hrniels: I've already done that and there is unfortunately no bus name for the tray icon. If I disable the dbus session permission and run `flatpak run --log-session-bus io.github.hrniels.Eventix | grep "(required 1)"`, I do not get a permission error. The tray icon is just not created.

I'm therefore unsure what to do here. I could of course disable the permission by default and describe in the README that it has to be turned on to get a tray icon. But I think the typical way for flatpak applications is to come with all required permissions set, right?

io.github.hrniels.Eventix.json
hrniels: Because Eventix is using GTK.

io.github.hrniels.Eventix.json
hrniels: That wasn't required at all actually. I've removed that.

io.github.hrniels.Eventix.json
hrniels: Good point, these are indeed not needed.

io.github.hrniels.Eventix.json
hrniels: Ah, I've missed that file. That makes it much easier.

io.github.hrniels.Eventix.json
hrniels: Oops, that (hopefully) made sense some time back, but it is indeed completely pointless now.

io.github.hrniels.Eventix.json
hrniels: What do you mean? I have a buildsystem in my Eventix repo and that is also able to create the flatpak package.

io.github.hrniels.Eventix.json
hrniels: Ah ok, is fixed.

io.github.hrniels.Eventix.json
hrniels: No reason, is removed now.

io.github.hrniels.Eventix.json
hrniels: Because that's only available at build time and extracting only the runtime from there (to avoid the large and unnecessary JDK) and install it to keep it for the runtime seemed more brittle than simply downloading the JRE.

io.github.hrniels.Eventix.json
hrniels: Good point. Since I have no easy way to test aarch64 atm, I've decided to only build for x86-64 for now.

hrniels (commented):
Thanks for the quick and detailed response!

Sorry for the many issues! I've never build a flatpak application before. I hope that I have fixed most of the raised issues. For a few I am not sure what to do.

io.github.hrniels.Eventix.json
hfiguiere: if you need Java use the Java SDK. And don't restrict architectures because of that. Java is suppsed to be compile once, debug everywhere.

io.github.hrniels.Eventix.json
hfiguiere: > Because that's only available at build time

So you didn't read its documentation.

io.github.hrniels.Eventix.json
hfiguiere: which version?

io.github.hrniels.Eventix.json
```diff
@@ -0,0 +1,125 @@
...
+                    "sha256": "c37f729200b572884b8f8e157852c739be728d61d9a1da0f920104876d324733"
+                }
+            ]
+        },
+        {
+            "name": "Eventix",
+            "buildsystem": "simple",
+            "build-commands": [
+                "cargo build --release",
+                "install -Dm755 target/release/eventix -t /app/bin/",
```
hfiguiere: why isn't all of this done a buildsystem upstream?

io.github.hrniels.Eventix.json
hrniels: GTK4

io.github.hrniels.Eventix.json
hrniels: Oh, they indeed document that. Sorry!

io.github.hrniels.Eventix.json
hrniels: True, I've enable aarch64 again.

io.github.hrniels.Eventix.json
hrniels: Because Eventix is exclusively run as a flatpak application. So there is no need to build/package that differently.

io.github.hrniels.Eventix.json
CodedOre: > But I think the typical way for flatpak applications is to come with all required permissions set, right?

Sure. But the keyword here is "required". Access to every bus name of the host system is **not required**.

Find out the bus name for the tray icon, and add the permission for that. Maybe you can find it in the full log, by running the command without grep. Or look into the documentation of the library you use for the tray icon.

io.github.hrniels.Eventix.json
CodedOre: Your buildsystem should be able to build your application and install all its files into a prefix, like `/usr/local`, or in this case, `/app`.

Your manifest should only need to call that buildsystem, not a whole list of commands. 

Notably though, it should not try to get online or build the Flatpak itself.

io.github.hrniels.Eventix.json
hfiguiere: the flatpak manifest isn't a buildsystem.

io.github.hrniels.Eventix.json
bbhtt: then your tray icon is broken, just get rid of it.

io.github.hrniels.Eventix.json
```diff
@@ -0,0 +1,117 @@
...
+                "mkdir -p /app/share/io.github.hrniels.Eventix/static",
+                "mkdir -p /app/share/io.github.hrniels.Eventix/icons",
+                "cp -r data/locale /app/share/io.github.hrniels.Eventix",
+                "cp -r data/static /app/share/io.github.hrniels.Eventix",
+                "cp -r data/icons /app/share/io.github.hrniels.Eventix"
+            ],
+            "sources": [
+                {
+                    "type": "git",
+                    "url": "https://github.com/hrniels/eventix.git",
```
bbhtt, line 88: so this was fully AI generated?

io.github.hrniels.Eventix.json
hrniels: I've looked into that again and fortunately found a way to solve that. ksni has a function `disable_dbus_name`, which then disables owning the bus `StatusNotifierItem-PID-ID`. That in combination with giving access to `org.kde.StatusNotifierWatcher` works (yes, I had that before, but it didn't have an effect without `disable_dbus_name`). So, finally I can remove the dbus session access :)

io.github.hrniels.Eventix.json
hrniels: I've removed that now.

io.github.hrniels.Eventix.json
hrniels: No. I'm working on that since more than 1.5 years and at least 90% of the code is manually written. But of course I am using LLMs as a tool (at least since about 3 months). Primarily to improve the overall quality (tests, documentation, ...), but also to be more productive. In every case with significant human oversight.

hrniels:
This update hopefully resolves the remaining issues. But let me know if there is anything else.

hrniels:
I don't want to push it as you are doing that on a voluntary basis, AFAIK. But can we continue here?

If it's about the AI Slob tag: I fully understand that reviewers like you have a hard time to judge submissions in that regard. But an increasing number of fully AI-generated submissions should not mean that honest submissions that use AI responsibly are no longer possible.

I have read Flathub's AI policy and I clearly do not violate it. If you think otherwise, can you give me concrete reasons?

barthalion:
bot, build

hrniels:
Thanks for unblocking it!

I've added the missing deps for aarch64, so that the build hopefully works now.

bot, build

hrniels:
bot, build

hrniels:
Great, that worked. However, I just noticed that there is one more thing I should fix before publishing it. Currently you have to start the server first and separately from the GUI, which is not that user friendly, especially if the user did not read my README before. I will therefore change that to automatically start the server if it's not already running. Let's better continue here after that's done.

hrniels:
bot, build

hrniels:
bot, build

hrniels:
Ok, for me it looks good now.

io.github.hrniels.Eventix.json
bbhtt: where is the non-AI generated part? Every commit since the past month has been made using copilot.

If it was there, by now it surely has been overwritten by the LLM.

hrniels: First of all, these are not AI-generated commits. These are commits that I personally tagged with "Co-authored by copilot" if it *helped* to build the commit. I thought being transparent about that is a good thing. And I really mean *helped*, because there is not a single commit that was generated by AI and blindly accepted. Everything is carefully designed, reviewed, and edited by myself until I am happy with the results.

Second of all, as I said, I started with agentic coding about 3 months ago. So, if you go further back, everything is manually written. And that's still the large majority of the code. Most of the code I used LLMs for is actually tests.

bbhtt: > because there is not a single commit that was generated by AI and blindly accepted

How can I tell that?

I looked at the files and every file currently has signs of LLM slop which is why I said they got overwritten. 

Also everyone using LLMs will always say "everything is reviewed, fine and happy". The inclusion criteria is not based on your opinion.

bbhtt: Also since every or majority of the files have large signs of LLMs generations, I don't see how it satisfies the criteria of 'meaningful human input'. Maybe it had at some point, but that seems to be lost now and most of it is LLM's input. The human input I can see is letting an LLM assist.

hrniels: Look at the [state of 3 months ago](https://github.com/hrniels/Eventix/tree/78f3c5b679a62eb5ab3998379e045b256d852333). Most of the code was already there.

Edit: 4 months, actually.

bbhtt: 1.0.9 is being Flatpak-ed which is from 2 days ago, so the state from months ago is not relevant.

hrniels: Ok. Then tell me how I can convince you that I used LLMs in line with Flathubs policy. Assuming that they and these checks here make any sense, there needs to be evidence I can provide or a procedure to demonstrate that.

bbhtt: > these checks here make any sense

If you don't think the policies make sense there is no point in opening a submission.

hrniels: I *do* think they make sense. It just seems that there is nothing I can do here to convince you that I adhere to them. That's why I am asking you directly: what piece of evidence or procedure do you accept?

bbhtt: It's your app and submission, you used LLM willingly it's on you to prove. The easiest way would have been to control LLM the generation and not make it obvious, that already failed.

hrniels: That's not answering my question.

But whatever. I built Eventix primarily for myself anyway and just thought, since I put a lot of effort into it, others might find it useful as well. But apparently Flathub is no longer the place for that.

It's just sad. Although I agree with the goal you want to achieve with your AI policy (avoiding low quality submissions I suppose), this case (and there are numerous others as I saw by looking at other PRs) clearly demonstrates that it simply does not work.

Neither can an author prove that LLMs have been used responsibly nor can the reviewers prove that they have not. And that leads to this witch-hunt-like middle ground: whether a submission is accepted or not is *completely* arbitrary (yes, I also saw PRs that are in my view obviously AI generated, but were accepted). You said it yourself: just hide it better and you will be fine. This is clearly not sustainable.

So I at least hope that this stimulates some discussions within Flathub on what you actually want to achieve with the AI policy and how this can be done.

barthalion:
I will make an exception for this submission because I meant to have it merged before we reworded the AI policy.

hrniels:
That's great, thank you!

PR #8689, opened 2026-05-16
Add org.babelserver.spekter

hallyhaa:

Instant, keyboard-driven recursive directory and file diff tool, in the spirit of Meld. It provides side-by-side and unified diff views with synchronized scrolling, inline character-level highlighting, recursive directory comparison with a two-column tree view, a tabbed interface and binary-file detection. It also detects when json files differ only in element ordering and aligns hunk boundaries to structural seams.

  The manifest requests `--filesystem=host:ro`. Spekter is strictly read-only: it reads files solely to compare them and never writes to them. Access to the whole filesystem is needed because the core purpose of the app is comparing directories the user chooses, which can live anywhere. For reference, the closest comparable application already on Flathub, Meld, ships with full read-write `--filesystem=host`. Spekter requests only read-only access.

  This application contains no generative AI features.

https://github.com/user-attachments/assets/232ec93b-38ed-47dc-a0d1-e0f2142261f0

### DIFF
--- a/org.babelserver.spekter.yaml
+++ b/org.babelserver.spekter.yaml
@@ -53,5 +53,5 @@ modules:
       - type: git
         url: https://gitlab.com/hallyhaa/spekter.git
         tag: v1.2.1
-        commit: c59e977afa00fefecca8261de9a4cb4813e6d8e3
+        commit: b578e88eb0db62fcc34cc09e93c24b433a9b7041
       - cargo-sources.json
--- a/org.babelserver.spekter.yaml
+++ b/org.babelserver.spekter.yaml
@@ -1,0 +1,58 @@
+# Flatpak manifest for Spekter.
+#
+# Regenerate cargo-sources.json whenever Cargo.lock changes:
+#   python3 -m venv .gen-venv && .gen-venv/bin/pip install aiohttp tomlkit
+#   curl -O https://raw.githubusercontent.com/flatpak/flatpak-builder-tools/master/cargo/flatpak-cargo-generator.py
+#   .gen-venv/bin/python flatpak-cargo-generator.py ../Cargo.lock -o cargo-sources.json
+#
+# Build and install locally:
+#   flatpak-builder --user --install --force-clean build-dir flatpak/org.babelserver.spekter.yaml
+
+id: org.babelserver.spekter
+
+runtime: org.freedesktop.Platform
+runtime-version: '25.08'
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+
+command: spekter
+
+finish-args:
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --device=dri
+  # Read-only host access: Spekter only reads files to compare them,
+  # never writes them. Meld ships with full host rw access on Flathub.
+  - --filesystem=host:ro
+
+modules:
+  - name: spekter
+    buildsystem: simple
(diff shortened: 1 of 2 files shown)


hallyhaa:
bot, build

hallyhaa:
Hi, and thanks for reviewing. I'd like to request a filesystem access exception for this submission. The linter flags `finish-args-host-ro-filesystem-access` for `--filesystem=host:ro`. Here's my reasoning for requesting the exception:

Spekter is a recursive directory and file diff tool. Its core purpose is comparing two directories or files chosen by the user, which can live anywhere on the filesystem. It needs to read those trees recursively to compute the diff. The requested access is strictly **read-only** (`host:ro`). Spekter never writes or modifies files; it only reads them in order to compare them. For precedent: Meld (`org.gnome.meld`), the closest comparable application already on Flathub, ships with full read-write `--filesystem=host`. Spekter requests strictly less than that.

org.babelserver.spekter.yaml
```diff
@@ -0,0 +1,58 @@
...
+command: spekter
+
+finish-args:
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --device=dri
+  # Read-only host access: Spekter only reads files to compare them,
+  # never writes them. Meld ships with full host rw access on Flathub.
+  - --filesystem=host:ro
```
CodedOre: > Meld ships with full host rw access on Flathub.

**Other apps do not serve as precedent.**\
We review new submissions based on their own merit and the current requirements, which can change over the years.

That aside:\
Has the use of the FileChooser portal been evaluated? Why was it not possible to use those?

org.babelserver.spekter.yaml
```diff
@@ -0,0 +1,58 @@
...
+      env:
+        CARGO_HOME: /run/build/spekter/cargo
+    build-commands:
+      - cargo --offline fetch --manifest-path Cargo.toml --verbose
+      - cargo --offline build --release --verbose -p spekter
+      - install -Dm755 target/release/spekter /app/bin/spekter
+      - install -Dm644 flatpak/org.babelserver.spekter.desktop
+          /app/share/applications/org.babelserver.spekter.desktop
+      - install -Dm644 flatpak/org.babelserver.spekter.metainfo.xml
+          /app/share/metainfo/org.babelserver.spekter.metainfo.xml
```
CodedOre, line 43: Just going to note that both desktop and metainfo file are not specific to Flatpak and are used on Linux for desktop integration in general.

hallyhaa:
> The domain to be used for verification is babelserver.org. If you intend to [verify](https://docs.flathub.org/docs/for-app-authors/verification) this submission, please confirm by uploading an empty `org.flathub.VerifiedApps.txt` file to https://babelserver.org/.well-known/org.flathub.VerifiedApps.txt. Otherwise, ignore this. Please comment if this incorrect.

This should be taken care of now.

org.babelserver.spekter.yaml
hallyhaa: The filesystem-access concern is now resolved. Spekter no longer uses an in-app file browser. File and directory selection goes through the XDG FileChooser portal (via the rfd crate). The manifest's finish-args therefore no longer request `--filesystem` at all. The updated manifest and a regenerated cargo-sources.json (for the new rfd dependency tree) are pushed. The git source is pinned to the v1.2.1 commit containing this change.

hallyhaa:
bot, build

hallyhaa:
bot, build

PR #8694, opened 2026-05-16
Add io.github.ebkr.r2modman

TB516:

r2modman is a simple and easy to use mod manager GUI that allows users to download, install, and update mods for games that use the thunderstore to host mods.
https://github.com/user-attachments/assets/4adb080a-1853-4693-99da-7662d9e7e0ee

@ebkr is the project owner and should be added as a maintainer for the flathub release


io.github.ebkr.r2modman.desktop
hfiguiere, line 1: This file should be (submitted and managed) upstream.

io.github.ebkr.r2modman.metainfo.xml
hfiguiere, line 1: This file should be (submitted and managed) upstream.

io.github.ebkr.r2modman.yaml
```diff
@@ -0,0 +1,63 @@
...
+      - install -Dm644 src/assets/icon/128x128.png
+        /app/share/icons/hicolor/128x128/apps/io.github.ebkr.r2modman.png
+      - install -Dm644 src/assets/icon/256x256.png
+        /app/share/icons/hicolor/256x256/apps/io.github.ebkr.r2modman.png
+      - install -Dm644 io.github.ebkr.r2modman.desktop
+        /app/share/applications/io.github.ebkr.r2modman.desktop
+      - install -Dm644 io.github.ebkr.r2modman.metainfo.xml
+        /app/share/metainfo/io.github.ebkr.r2modman.metainfo.xml
+      - install -Dm755 io.github.ebkr.r2modman /app/bin/io.github.ebkr.r2modman
+finish-args:
```
hfiguiere: this should be above `modules`

TB516:
Making the changes upstream then ill update stuff here!

io.github.ebkr.r2modman
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+if [ -n "$WAYLAND_DISPLAY" ]; then
+  exec zypak-wrapper /app/r2modman \
+    --no-sandbox \
```
bbhtt: if you are using zypak you can drop --no-sandbox.

io.github.ebkr.r2modman.yaml
```diff
@@ -0,0 +1,59 @@
...
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node22
+base: org.electronjs.Electron2.BaseApp
+base-version: "25.08"
+command: io.github.ebkr.r2modman
+finish-args:
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --device=dri
+  - --talk-name=org.freedesktop.Flatpak
```
bbhtt, line 14: why is this needed?

io.github.ebkr.r2modman.yaml
```diff
@@ -0,0 +1,59 @@
...
+  - --share=network
+  - --share=ipc
+  - --filesystem=host
```
bbhtt, line 17: host includes the one below, why is host needed?

io.github.ebkr.r2modman
TB516: Ok! I have the pr updating this in upstream, and im also making it so the script comes from the upstream repo so it aligns with the appstream info and desktop file. Once thats merged ill update stuff here!

io.github.ebkr.r2modman.yaml
TB516: We currently use `flatpak-spawn --host` when we are running on the steam deck in game mode, because the portals dont function in game mode, so calling the steam binary on the host is the only way to get the game to launch from r2. In the future though, we also want to bring feature parity for games that use diret launch (calling a binary directly vs having a storefront launch the game), which would also need `flatpak-spawn --host`.

io.github.ebkr.r2modman.yaml
TB516: Currently, we have `--filesystem=host` beause r2 needs to be able to locate and write to the game install location, wherever that might be, and host filesystem access lets us do that automatically, no matter where the user has their games.

For `--filesystem=~/.var/app/com.valvesoftware.Steam`, we need to be able to access the steam install location, and with flatpak steam, thats not included with host access, or at least, when testing, it wasnt able to detect the flatpak steam install without the permission.

bbhtt: what "game installation" and where is it usually? Can you not use a filechooser to ket the user select a location? Then it will not need host.

io.github.ebkr.r2modman.yaml
bbhtt: Sounds like steam should make portals work. This needs an exception and it won't be granted based on limitations imposed by a single platform.

There's also no guarantee that a steam binary exists on host. It could be a flatpak snap or whatever or may not be just installed at all.

ebkr: That is an unfortunate take and I don’t believe that’s the correct response. This is a requirement for it to function. Steam not supporting xdg portals properly is directly relevant to the tool and it becomes unusable without this.

io.github.ebkr.r2modman.yaml
ebkr: Correct me if I’m wrong, but I believe if you pick a directory where you don’t have access then it’s redirected to a run folder, rather than the real folder. 

There is:
- Automated game directory resolution (by reading Steam’s libraryfolders.vdf)
  - Of which, Steam is also auto detected and can be in a bunch of native locations or the Flatpak location
  - Plus other launchers like Heroic in the future 
- A need to place specific files in game directories. Especially for Proton, because it loads winhttp.dll for example based on Windows DLL resolution

TB516: The 'game installation' is the game's actual files that steam (or another game manager) would install.

How r2 currently works for steam is it looks for the location where steam itself is installed, then checks `libraryfolders.vdf`, which is used to locate the the game installation. Because of this, r2 needs to be able to read and write to wherever that file says the steam library is, which can be anywhere on the system.

A filechooser wouldn't be ideal because it would mean each game would need to be individually identified by the user, and need the user to relocate the game when the permission expires, which breaks the abstraction that is the mod manager, and can lead to user caused issues that aren't present when automatically locating the directories.

io.github.ebkr.r2modman.yaml
TB516: Id like to add that this would still be needed for the 'direct launch' games, which is planned to bring the linux version in line with windows, so while it is needed for the steam deck to work, its not only needed for the steam deck.

TB516: Could this be reconsidered since it isnt steam deck exclusive? We would like to compromise where possible, buit sacraficing/gimping featrures is something we want to avoid.

CodedOre: As I said before in one of your exceptions PRs, `flatpak spawn --host` is the ultimate sandbox escape. 

It would allow any command to be run on the host. So it kind of defeats the whole sandbox. 

Such a permission is only granted in exceptional circumstances.

TB516: R2, as a mod manager, does need some level of sandbox escape in certain situations. We needs to be able to communicate with host binaries sometimes, and afaik, flatpak-spawn --host is the only way to do that, right? If there's some solution to be able to talk with/launch a predefined set of apps on the host, we could do that, but from what I know, flatpak v1 doesn't have anything like that.

Our issue is that it wouldn't be great if the flatpak version was gimped compared to other versions of r2. If someone had purchased a drm free copy of hollow knight silksong, for example, it being unusable from flatpak r2, and needing users to get the AppImage or something else, wouldn't be great. 

Is there anything we can do to further show that this is something that's needed, or any other solution to the issues we're facing? Is there any flexibility on this at all?

TB516:
Are there any updates on the permission requests? If theres any extra info we can provide to help the process, let us know!

io.github.ebkr.r2modman.yaml
bbhtt: > That is an unfortunate take and I don’t believe that’s the correct response. 

What is unfortunate is Steam deck, Steam OS are paid products funded by a corporation and they can't get working portals, something that works in every other distro without doing anything, and for which other unpaid people have to endure the nonsense where I have to repeat the same thing to you across 3 different pull requests and several weeks.

> Id like to add that this would still be needed for the 'direct launch' games, which is planned to bring 

You can ask for it when the feature arrives and it will evaluated it then. There is no need to ask for permissions in advance.

ebkr: Steam OS having issues with portals is still a concern that needs to be handled and is a primary use case of ours. Therefore we still need `flatpak-spawn --host` regardless. This is a clear valid reason for needing it.

Us needing this permission in the future too is an additional supporting reason. Ignoring this, it’s still a requirement.

For all intents and purposes this is both a glorified file manager and also an application launcher.

TB516:
@ebkr decided that he wants to host their own repo for the app so r2 can have the host spawn permissions, so im closing the pr here. Thank you maintainers for the time spent reviewing the submission, and sorry it didn't end up going anywhere!

PR #8716, opened 2026-05-18
Add io.keet.Keet

RaisinTen:

- [x] Please describe the application briefly. Keet is a private and secure P2P chat app.
- [x] Please attach a video showcasing the application on Linux using the Flatpak. [keet-flatpak-demo.webm](https://github.com/user-attachments/assets/ce5e5045-5777-4024-8493-f4648a1a9335)
- [x] I am a developer contributor to the project.

### DIFF
--- a/io.keet.Keet.yml
+++ b/io.keet.Keet.yml
@@ -1,0 +1,43 @@
+id: io.keet.Keet
+runtime: org.freedesktop.Platform
+runtime-version: '25.08'
+sdk: org.freedesktop.Sdk
+base: org.electronjs.Electron2.BaseApp
+base-version: '25.08'
+command: Keet
+separate-locales: false
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=org.freedesktop.Notifications
+  - --persist=.config/Keet
+  - --persist=.config/pear
+  - --filesystem=xdg-download
+  - --env=XCURSOR_PATH=/run/host/user-share/icons:/run/host/share/icons
+modules:
+  - name: Keet
+    buildsystem: simple
+    build-commands:
+      - install -Dm644 flatpak/${FLATPAK_ID}.metainfo.xml /app/share/metainfo/${FLATPAK_ID}.metainfo.xml
+      - install -Dm644 flatpak/${FLATPAK_ID}.desktop /app/share/applications/${FLATPAK_ID}.desktop
+      - install -Dm644 flatpak/${FLATPAK_ID}.png /app/share/icons/hicolor/128x128/apps/${FLATPAK_ID}.png
+      - install -Dm755 flatpak/entrypoint.sh /app/bin/Keet
+      - mkdir -p /app/lib/${FLATPAK_ID}
+      - cp -r ./Keet-*/* /app/lib/${FLATPAK_ID}
+    sources:
+      - type: archive
+        url: https://static.keet.io/downloads/4.15.0/Keet-arm64-flatpak.tar.gz
+        only-arches:
+          - aarch64
+        strip-components: 0
+        sha512: 77e5099f0cc20d2ad5114f266779da1ec85162ec6f7fc39bd311e790520af57a8d2eee4585a051fb170f0b64c70bc55d616d54e471020bd31282da7fe03988bb
+      - type: archive
+        url: https://static.keet.io/downloads/4.15.0/Keet-x64-flatpak.tar.gz
+        only-arches:
+          - x86_64
+        strip-components: 0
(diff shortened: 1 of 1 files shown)


ghost:
If this app is open source, Flathub prefers it to be built from source code.

RaisinTen:
The app is proprietary software and not open source.

ghost:
In that case, you need to add the `proprietary` tag to the manifest file.

```yaml
tags:
  - proprietary
```

io.keet.Keet.yml
```diff
@@ -0,0 +1,43 @@
+id: io.keet.Keet
+runtime: org.freedesktop.Platform
+runtime-version: '25.08'
+sdk: org.freedesktop.Sdk
+base: org.electronjs.Electron2.BaseApp
+base-version: '25.08'
+command: Keet
```
ghost, line 7:
```suggestion
command: Keet
tags:
  - proprietary
```

ghost (changes requested):

ghost:
It would be good to provide the `bugtracker` and `help` URLs in the **metainfo** file (if you are allowed to update it).

See: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-url

io.keet.Keet.yml
```diff
@@ -0,0 +1,45 @@
...
+  - proprietary
+separate-locales: false
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=org.freedesktop.Notifications
```
bbhtt:
```suggestion
```

this uses the portal by default

io.keet.Keet.yml
```diff
@@ -0,0 +1,45 @@
...
+command: Keet
+tags:
+  - proprietary
+separate-locales: false
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
```
bbhtt, line 16: why is pulseaudio needed?

io.keet.Keet.yml
```diff
@@ -0,0 +1,45 @@
...
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=org.freedesktop.Notifications
+  - --persist=.config/Keet
+  - --persist=.config/pear
```
bbhtt: why don't these use the proper xdg locations?

io.keet.Keet.yml
```diff
@@ -0,0 +1,45 @@
...
+  - --filesystem=xdg-download
```
bbhtt: why is this needed?

io.keet.Keet.yml
RaisinTen: This is needed so that the app can also play audio content. For example, when you play a video or hear audio when attending a call.

io.keet.Keet.yml
RaisinTen: The app currently saves files downloaded via the app directly into the user's `~/Downloads` directory, so this permission is needed to allow that.

Edit: Just realized that if I open a file chooser dialog from the app and the user selects `~/Downloads`, the flatpak writes to a path like `/run/user/1000/doc/ece7c321/Downloads/` which mirrors things to the user's actual Downloads folder. We can update that in the next release.

io.keet.Keet.yml
RaisinTen: Will drop both these lines once I update this to use the new flatpak artifacts which should address this comment and https://github.com/flathub/flathub/pull/8716#issuecomment-4475764045.

RaisinTen:
> The domain to be used for verification is keet.io. If you intend to [verify](https://docs.flathub.org/docs/for-app-authors/verification) this submission, please confirm by uploading an empty `org.flathub.VerifiedApps.txt` file to https://keet.io/.well-known/org.flathub.VerifiedApps.txt. Otherwise, ignore this. Please comment if this incorrect.

Done, we've uploaded an empty file to https://keet.io/.well-known/org.flathub.VerifiedApps.txt

RaisinTen:
I've updated the PR and it's ready for a review:

@yakushabb the metainfo file now includes the help and bugtracker URLs.

@bbhtt:
- I've dropped the notifications talk name and made it use proper XDG paths.
- pulseaudio setting: it's needed so that we can hear audio when the app plays audio from media files or while attending calls
- download setting: when users attempt to save files from the app, the app directly writes it to `~/Downloads`. To remove this, we are changing the UI to open a file selector dialog to allow users to select the location. My plan is to drop this setting in the next release of the app (added a TODO for that). Would you be ok with that or is it a hard blocker for acceptance?

Also, an empty file has been uploaded to https://keet.io/.well-known/org.flathub.VerifiedApps.txt for verification.

bbhtt:
bot, build

PR #8727, opened 2026-05-19
Add org.freedesktop.Sdk.Extension.bear

dogtopus:

- [x] Please describe the application briefly. < Bear tracks compiler invocations from arbitrary build systems and generate `compiler_commands.json` for use with several IDEs. This extension allows using Bear in fully sandboxed Flatpak IDEs without the need to allow access to extra binaries and/or manually messing with `PATH`. >
- [X] Please attach a video showcasing the application on Linux using the Flatpak. < Invoking the Bear extension from VSCodium Flatpak >

[demo.webm](https://github.com/user-attachments/assets/d5cc9e3b-7056-43a6-a425-ff007e1d9b1f)

- [x] I contacted upstream developers about this submission. **Link:** https://github.com/rizsotto/Bear/discussions/700

### DIFF
--- a/org.freedesktop.Sdk.Extension.bear.yaml
+++ b/org.freedesktop.Sdk.Extension.bear.yaml
@@ -1,27 +1,30 @@
-id: 'org.freedesktop.Sdk.Extension.bear'
-branch: '25.08'
-runtime: 'org.freedesktop.Sdk'
+id: org.freedesktop.Sdk.Extension.bear
+runtime: org.freedesktop.Sdk
 runtime-version: '25.08'
+sdk: org.freedesktop.Sdk
+branch: '25.08'
 build-extension: true
-sdk: 'org.freedesktop.Sdk'
-sdk-extensions: ['org.freedesktop.Sdk.Extension.rust-stable']
 separate-locales: false
+
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+
 cleanup:
-  - '/share/info'
-  - '/share/man'
+  - /share/info
+  - /share/man
 
 modules:
-  - name: 'bear'
+  - name: bear
     buildsystem: simple
     build-options:
-      append-path: '/usr/lib/sdk/rust-stable/bin'
+      append-path: /usr/lib/sdk/rust-stable/bin
       env:
-        CARGO_HOME: '/run/build/bear/cargo'
+        CARGO_HOME: /run/build/bear/cargo
         CARGO_NET_OFFLINE: 'true'
     sources:
       - type: git
-        url: 'https://github.com/rizsotto/Bear.git'
-        tag: '4.1.3'
+        url: https://github.com/rizsotto/Bear.git
+        tag: 4.1.3
       - 'cargo-deps.json'
     build-commands:
       - 'cargo --offline fetch --manifest-path Cargo.toml --verbose'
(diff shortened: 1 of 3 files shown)


dogtopus:
bot, build

dogtopus:
The aarch64 version doesn't compile and I'm not sure why. I'm putting this into draft mode while I investigate.

Specifically I know why the aarch64 version fails to compile but I can't make sense of it since the x86_64 build should be under the same failing condition but that did build successfully regardless.

bbhtt:
SDK extensions like this need a prospective user who will use it for example existing apps on Flathub. It's in the requirements and we cannot accept it otherwise.

You haven't included something to show any users.

dogtopus:
@bbhtt I imagine a typical user who might need this to be one who

- Uses one of the Flatpak IDEs/editors with sandbox enabled (i.e. no host access, only allow direct access to project folders) and has support for clangd or some other C/C++ language server that make use of `compile_commands.json`.
- Wishes to develop, using such IDE/editor, a C/C++ project that already made use of a build system that does not generate `compile_commands.json` natively (like autotools), and needs to rely on Bear to generate `compile_commands.json`. This project would typically target Flatpak/Freedesktop SDK but this is not a requirement.

For example, one might use the VSCodium Flatpak with [clangd extension](https://open-vsx.org/vscode/item?itemName=llvm-vs-code-extensions.vscode-clangd) installed (which by default looks for `compile_commands.json` under the `builddir` directory), and set up their autotools-based project with a `.vscode/tasks.json` file that looks like this:

```json
{
	"version": "2.0.0",
	"tasks": [
		{
			"type": "shell",
			"label": "Configure and build project",
			"command": "mkdir -p builddir && cd builddir && bear -- ../configure && bear -a -- make -j$(nproc)",
			"group": "build"
		},
		{
			"type": "shell",
			"label": "Clean project",
			"command": "cd builddir && make clean",
			"group": "build"
		}
	]
}
```

The user could then simply use the VSCodium GUI to build the project as they see fit, without the need to configure an external Bear installation by modifying permissions and `PATH` variable, or to modify the tasks file to use absolute paths for invoking Bear.

This also doesn't just limit to VSCodium. Any IDEs or editors that make use of `compile_commands.json` and either allow custom build targets, or make use of a system-wide Bear installation internally (which is probably more common than one would imagine on a first glance since it's [somewhat endorsed by the clangd developers](https://clang.llvm.org/docs/JSONCompilationDatabase.html#supported-systems), but I digress) could make use of this extension.

dogtopus:
The issue listed above should be resolved now and this PR should be ready for a review.

bot, build

bbhtt:
The question is do you have anyone currently wanting to use it. Other apps, other app's users etc. 

I know everything has some theoretical use and it doesn't count otherwise we would be packaging the entire world as Flatpak.

org.freedesktop.Sdk.Extension.bear.yaml
```diff
@@ -0,0 +1,55 @@
+id: org.freedesktop.Sdk.Extension.bear
```
bbhtt, line 1: thread for https://github.com/flathub/flathub/pull/8727#issuecomment-4493471376 (ignore this).

dogtopus:
> The question is do you have anyone currently wanting to use it. Other apps, other app's users etc.
> 
> I know everything has some theoretical use and it doesn't count otherwise we would be packaging the entire world as Flatpak.

I think you got a point. I figured out this setup after encountering that specific problem (i.e. setting up Flatpak IDE + clangd + autotools) and figured that this *could* be useful for many others as well, unlike some of my other Flatpaks that are obviously only specific to my use case. OTOH I'm also not sure whether people would actually be interested in having a setup that's similar to mine, and that I understood that Flatpak workflows aren't generally made for command line tools and can have discoverability issues especially when the package is an extension (at least for gnome-software this seems to be the case).

So in short, I'm the user who's actively using it for some codebases that I work on, but I don't know what's beyond that as I don't have testimonies from other users/IDE projects.

If usefulness for the general public is a concern (i.e. "what do you mean nobody uses it I'm the user" is not enough, which my gut feeling tells me that it isn't), maybe you're right that it's better to leave this outside of Flathub for now, make the manifests available for people to use (maybe also make a note somewhere on the Bear README/documentation, but that'd be a different issue), and maybe consider resubmit if enough people demand for it.

Thanks for your time to explain this to me.

PR #8773, opened 2026-05-24
Add com.alliedbits.ABCalcPlus

emayevski:

 ABCalc is a powerful yet user-friendly calculator for complex expressions, offering multi-format results, persistent history, unlimited memory, and much more.

- [X] Please attach a video showcasing the application on Linux using the Flatpak. https://github.com/user-attachments/assets/7995678a-e6d3-4406-bbf2-98d0953e5aba


com.alliedbits.ABCalcPlus.json
```diff
@@ -30,7 +30,7 @@
         {
           "type": "archive",
           "url": "https://alliedbits.com/downloads/abcalc-1.5.2-linux-x86_64.tar.gz",
-          "sha256": "e37f21af9f27b1f037d524f9e3a955f36e77d401b56d45f691ee95d7b5443275",
+          "sha256": "6bf11bb2143990b31d434985a05bc3e46c631b38cb7d3a8edc906aa4984f14b2",
```
hfiguiere: don't retag. it's bad.

emayevski: I am sorry, but I am new to the process. As I understand it, this is the hash of the archive. If the archive gets updates (I had to update the license text there), what's the approach? Should I just submit with the outdated hash? Also, this is a draft PR, I assumed that it was ok to update it while it was in the Draft status?

hfiguiere: since the archive is different it should have a new name.

emayevski: Thnak you, I will remember this in the future.

emayevski:
The error has been fixed and the domain has been verified.

com.alliedbits.ABCalcPlus.json
```diff
@@ -0,0 +1,58 @@
...
+          "strip-components": 1,
+          "only-arches": [
+            "x86_64"
+          ]
+        },
+        {
+          "type": "script",
+          "dest-filename": "abcalc-launcher",
+          "commands": [
+            "#!/bin/sh",
```
hfiguiere:
```suggestion
```

com.alliedbits.ABCalcPlus.json
```diff
@@ -0,0 +1,58 @@
...
+          ]
+        },
+        {
+          "type": "file",
+          "path": "com.alliedbits.ABCalcPlus.desktop"
+        },
+        {
+          "type": "file",
+          "path": "com.alliedbits.ABCalcPlus.metainfo.xml"
+        }
```
hfiguiere: neither file exist.

com.alliedbits.ABCalcPlus.json
```diff
@@ -0,0 +1,58 @@
...
+  ],
+  "modules": [
+    {
+      "name": "abcalc",
+      "buildsystem": "simple",
+      "build-commands": [
+        "mkdir -p /app/lib/abcalc",
+        "install -m 755 abcalc /app/lib/abcalc/abcalc",
+        "find . -maxdepth 1 -name '*.dll' -exec install -m 644 {} /app/lib/abcalc/ \\;",
+        "find . -maxdepth 1 -name '*.so' -exec install -m 755 {} /app/lib/abcalc/ \\;",
```
hfiguiere: why do you need to "find"?

com.alliedbits.ABCalcPlus.json
```diff
@@ -0,0 +1,58 @@
...
+    "--socket=wayland",
+    "--device=dri"
+  ],
+  "modules": [
+    {
+      "name": "abcalc",
+      "buildsystem": "simple",
+      "build-commands": [
+        "mkdir -p /app/lib/abcalc",
+        "install -m 755 abcalc /app/lib/abcalc/abcalc",
```
hfiguiere:
```suggestion
        "install -Dm755 abcalc /app/lib/abcalc/abcalc",
```

com.alliedbits.ABCalcPlus.json
```diff
@@ -0,0 +1,58 @@
...
+        "install -Dm755 abcalc-launcher /app/bin/abcalc",
+        "install -Dm644 com.alliedbits.ABCalcPlus.desktop /app/share/applications/com.alliedbits.ABCalcPlus.desktop",
+        "install -Dm644 abcalc.png /app/share/icons/hicolor/256x256/apps/com.alliedbits.ABCalcPlus.png",
+        "install -Dm644 com.alliedbits.ABCalcPlus.metainfo.xml /app/share/metainfo/com.alliedbits.ABCalcPlus.metainfo.xml",
+        "install -Dm644 LICENSE /app/share/licenses/com.alliedbits.ABCalcPlus/LICENSE"
+      ],
+      "sources": [
+        {
+          "type": "archive",
+          "url": "https://alliedbits.com/files/abcalc-1.5.2-linux-x86_64.tar.gz",
```
hfiguiere: The way I read the license is that it is not redistributable. So this should be using "extra-data" instead.

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "skip-arches": ["aarch64"]
```
hfiguiere: this should be `only-arches`. Or better, have an aarch64 build so it's not necessary. This this seems to be using a compile-once-debug-everywhere technology, I don't see why not.

com.alliedbits.ABCalcPlus.json
emayevski: There are over 230 DLLs and 19 .so files there. the find commands are way simpler to maintain.

hfiguiere: no it's not. `install` can take many file (and you can use shell globs like virtually anywhere else). I guess why make it simple when you can make it complicated.

com.alliedbits.ABCalcPlus.json
emayevski: This is the end-user license, i.e., it does not talk about distribution at all. I presume that if I submit the software and agree with the T&C of the service or website, I grant the distribution license according to these T&C. 

I can change the entry as you suggest, but is it really necessary?..

com.alliedbits.ABCalcPlus.json
emayevski: Updated in the local file.

com.alliedbits.ABCalcPlus.json
emayevski: Updated in the local file.

flathub.json
emayevski: The application uses Avalonia, which uses SkiaSharp, and SkiaSharp until recently had a bug that prevented its operations on ARM64. As soon as they fix it, we will upgrade to the next major release of Avalonia and provide a version for ARM64. 

As I am reading [the instructions](https://docs.flathub.org/docs/for-app-authors/maintenance#limiting-the-set-of-architectures-to-build-on), I see no difference between skip-arches and only-arches (now  when only two exist). But I see your point; I'll update the file to use only-arches.

com.alliedbits.ABCalcPlus.json
emayevski: The [instructions](https://docs.flathub.org/docs/for-app-authors/requirements#manifest) at the very end of the page are a bit vague - they mention that the files must be included but don't specify where. I will remove the  blocks you highlighted.

hfiguiere: https://docs.flathub.org/docs/for-app-authors/requirements#required-metadata

the "important" box content. it's pretty clear.

also it's pretty clear what a source "file" is and in that specific case the file doesn't exist. (where do you think it finds it?)

emayevski: It may be clear for a person that is related to the process, but for the outsider, the instructions are quite fuzzy. Regarding where it finds the files - the instructions state that the files should be included in the source, which, in our case, is a tar.gz archive, as I understand it (or it may be the flatpak file, depending on how one reads the text). One would assume that files that are not referenced via a URL are local (and local here means whatever package is available for the entity processing the manifest).

Nevermind, I have removed these blocks.

hfiguiere:
blocked on the license question. https://github.com/flathub/flathub/pull/8773#discussion_r3303529521

hfiguiere:
bot, build

emayevski:
Changed the manifest to use extra-data and added the missing files. Hopefully, got it right this time.

com.alliedbits.ABCalcPlus.metainfo.xml
```diff
@@ -0,0 +1,84 @@
...
+        <control>keyboard</control>
+        <control>pointing</control>
+        <control>touch</control>
+    </supports>
+    <requires>
+        <display_length compare="ge">540</display_length>
+    </requires>
+    <custom>
+        <value key="flathub::manifest">https://alliedbits.com/abcalc/flatpak/1.5.2/com.alliedbits.ABCalcPlus.json</value>
+    </custom>
```
hfiguiere:
```suggestion
```

No. this is reserved for cases that is not this one.

emayevski: Just removed it.

com.alliedbits.ABCalcPlus.json
bbhtt: If you are the upstream why is it using extra-data?

com.alliedbits.ABCalcPlus.json
emayevski: This has been figured out and reworked.

com.alliedbits.ABCalcPlus.json
emayevski: > If you are the upstream why is it using extra-data?

Now, I am confused. Should it be extra-data or "archive". The description for "extra-data" says: "Extra data is often used for proprietary applications to avoid redistribution by downloading the binary package provided by the original distributors during installation by the user." , so maybe it is really more appropriate here.

bbhtt: If you are the upstream you automatically have the right to redistribute your work anywhere, that's how licences work and even without a licence.

emayevski: I am fine with both options, actually. I am just looking for a way to finalize the submission process and move to other tasks.

emayevski:
bot, build

emayevski:
bot, build

bbhtt:
https://github.com/flathub/flathub/pull/8773#discussion_r3322489814 was not solved, you marked it as solved then you are asking for a review. I had to spend extra time finding it.

emayevski:
> [#8773 (comment)](https://github.com/flathub/flathub/pull/8773#discussion_r3322489814) was not solved, you marked it as solved then you are asking for a review. I had to spend extra time finding it.

It was solved - you and the other person expressed two opposite opinions. The first reviewer requested that I change "archive" to "extra-data" ([here](https://github.com/flathub/flathub/pull/8773#discussion_r3303529521)), which I did. Then you asked about the reasons, and I explained. As I am reading your comment now, it looks like you are requesting to revert it back from extra-data to archive? I don't know - @hfiguiere started reviewing my submission first, and I don't know which requests have priority.

bbhtt:
@hfiguiere might have missed that you were upstream. If you are upstream there is no need to use extra-data as I said.

Maybe I should put this in docs explicitly...

emayevski:
> @hfiguiere might have missed that you were upstream. If you are upstream there is no need to use extra-data as I said.
> 
> Maybe I should put this in docs explicitly...

Got it, I will rework the submission today. Thank you.

emayevski:
> > @hfiguiere might have missed that you were upstream. If you are upstream there is no need to use extra-data as I said.
> > Maybe I should put this in docs explicitly...
> 
> Got it, I will rework the submission today. Thank you.

The sources were reworked to be "archive".

emayevski:
bot, build

com.alliedbits.ABCalcPlus.json
```diff
@@ -16,65 +16,49 @@
...
-            "rm abcalc.tar.gz"
-          ]
-        },
         {
           "type": "script",
           "dest-filename": "abcalc-launcher",
           "commands": [
-            "export LD_LIBRARY_PATH=/app/extra:$LD_LIBRARY_PATH",
-            "exec /app/extra/abcalc \"$@\""
+            "export LD_LIBRARY_PATH=/app/lib/abcalc:$LD_LIBRARY_PATH",
```
hfiguiere: why not having the libraries in `/app/lib` instead like they should?

com.alliedbits.ABCalcPlus.metainfo.xml
hfiguiere, line 1: this should also be in the archive

com.alliedbits.ABCalcPlus.desktop
hfiguiere, line 1: this should be in the tarball as well.

emayevski:
bot, build

emayevski:
bot, build

com.alliedbits.ABCalcPlus.json
emayevski: Fixed.

com.alliedbits.ABCalcPlus.metainfo.xml
emayevski: Fixed.

com.alliedbits.ABCalcPlus.desktop
emayevski: Fixed.

emayevski:
> @hfiguiere might have missed that you were upstream. If you are upstream there is no need to use extra-data as I said.
> 
> Maybe I should put this in docs explicitly...

Fixed and verified. The submission has been simplified (reduced to one file), it builds and installs fine.

com.alliedbits.ABCalcPlus.json
```diff
@@ -0,0 +1,56 @@
...
+    "--socket=wayland",
+    "--device=dri"
+  ],
+  "modules": [
+    {
+      "name": "abcalc",
+      "buildsystem": "simple",
+      "build-commands": [
+        "install -Dm755 abcalc /app/lib/abcalc",
+        "install -Dm755 abcalc-launcher /app/bin/abcalc",
```
hfiguiere: why is a launcher needed. abcalc should simply be in /app/bin like any binary.

emayevski: > why is a launcher needed. abcalc should simply be in /app/bin like any binary.

Does not work this way (I tried this explicitly) - the .so files are not found. This is a single-file .NET 10 application, and its loading sometimes gets quirky.

bbhtt: What is the problem? You are installing the binary to lib instead of bin.

emayevski: @bbhtt The above comment suggested placing the main executable ("abcalc") to /app/bin in order to get rid of the launcher. This won't work: if "abcalc" sits in /app/bin and the .so are in /app/lib, then abcalc can't locate the .so libraries (some .NET glitches...). 

For this reason, a tiny launcher is placed into /app/bin, and the main exe, "abcalc", is placed in /app/lib. This way, everything works, and there are no problems with this setup.

bbhtt: > then abcalc can't locate the .so libraries (some .NET glitches...)

Yes that is what I am asking. What is the glitch?

emayevski: > > then abcalc can't locate the .so libraries (some .NET glitches...)
> 
> Yes that is what I am asking. What is the glitch?

As I mentioned above - .so modules are not found. 

This is the exact message that occurs when "abcalc" is in bin and libraries are in lib:

```Unhandled exception. System.TypeInitializationException: The type initializer for 'SkiaSharp.SKImageInfo' threw an exception.
 ---> System.DllNotFoundException: Unable to load shared library 'libSkiaSharp' or one of its dependencies. In order to help diagnose loading problems, consider using a tool like strace. If you're using glibc, consider setting the LD_DEBUG environment variable: 
/app/bin/libSkiaSharp.so: cannot open shared object file: No such file or directory
/app/bin/liblibSkiaSharp.so: cannot open shared object file: No such file or directory
/app/bin/libSkiaSharp: cannot open shared object file: No such file or directory
/app/bin/liblibSkiaSharp: cannot open shared object file: No such file or directory

   at SkiaSharp.SkiaApi.sk_colortype_get_default_8888()
   at SkiaSharp.SkiaApi.sk_colortype_get_default_8888()
   at SkiaSharp.SKImageInfo..cctor()
   --- End of inner exception stack trace ---
   at Avalonia.Skia.PlatformRenderInterface..ctor(Nullable`1 maxResourceBytes)
   at Avalonia.Skia.SkiaPlatform.Initialize(SkiaOptions options)
   at Avalonia.SkiaApplicationExtensions.<>c.<UseSkia>b__0_0()
   at Avalonia.AppBuilder.SetupUnsafe()
   at Avalonia.AppBuilder.Setup()
   at Avalonia.AppBuilder.SetupWithLifetime(IApplicationLifetime lifetime)
   at Avalonia.ClassicDesktopStyleApplicationLifetimeExtensions.StartWithClassicDesktopLifetime(AppBuilder builder, String[] args, Action`1 lifetimeBuilder)
   at ABCalc.Desktop.Program.Main(String[] args)
```

The application is installed and runs under a regular user. `flatpak run com.alliedbits.ABCalcPlus` is how the application is launched.

Moreover, I observed the situation where not changing the directory to lib before running the exe would lead to a similar error. But it does not manifest itself at the moment.

bbhtt: > /app/bin/liblibSkiaSharp.so: cannot open shared object file: No such file or directory

This is very broken. Do you know why it is looking for .so relative to the binary instead of libdir?

emayevski: @bbhtt  I can only guess. I assume that while .NET looks for .so modules properly (in the ld_library_path and additionally, in the local directory), this look up part is not compatible with Flatpak. 

Technically, it may be possible to add a resolver class to a .NET application that will suggest other paths to the loader logic, but this is relatively complicated. 

I can try to experiment, but for this, I need to understand - why exactly should the .NET loader look into "../lib" relative to the main exe? I.e., is this "../lib" or "/lib" included in the ld_library_path paths by flatpak, or where does this come from? If a resolver class is to be added, it should use the source of those paths (e.g., look into ld_library_path or other env var), not just dumbly try "../lib" and "/lib" directories.

bbhtt: In general the linker records where libraries should be looked for and the dynamic loader handles them by reading the recorded names and looking through its search paths.

The dynamic loader inside the runtime is configured to look at /app/lib by default.

Or you use dlopen() with the libname.

emayevski: @bbhtt  Well, none of this happens. There is no linker in a strict sense there because a single-file .NET executable is akin to self-extracting archives (some pre-created stub unpacking or loading other modules). I also suspect that the dynamic loader of the OS (or of flatpak) is not involved either - .NET handles loading of all modules on its own (i.e., locates an .so first, then calls dlopen() with a resolved path). This is why it is possible to create a resolver that would point at a custom location. 

dlopen() is probably not an option because (a) the .so modules are referenced by the third-party libraries used by the Avalonia UI framework, and (b) I suspect that early loading of the .so, even if possible (of which I doubt), may not work. the .so may need to get resolved before our code gets control, in which case, a resolver would be the only realistic option.

bbhtt: Can you at least make it look up inside `libdir` then?

emayevski: what is "libdir" in this context? Is it some hardcoded path or an env var, or some OS call? It is possible to hard-code a couple of candidate paths, but this is not _right_. There should be a more generic approach here.

bbhtt: No you can supply one during building. When you are building for a Flatpak app supply /app/lib and so on.

emayevski: @bbhtt I am sorry but I don't quite understand - are you referring to the building of the application or to the building of the flatpak package? When the application is built, there is no linker involved and no way to specify the path (the correct approach here is to write a custom resolver right in the .NET code).

bbhtt: You don't need a linker or whatever. You cannot pass some variable or argument when you are building the application?

You can do `LIBDIR=/app/lib your_dotnet_build_cmd` then it reads that variable and embeds the resultant path for searching the libraries.

emayevski: Setting LIBDIR does nothing in the .NET case.

.NET linkage is not exactly linkage in a C/C++ meaning. When the application is built, there is an IL assembly generated, optionally amended by some pre-existing native code provided by .NET. This native code is a launcher, but it may not even be present (one can run the managed code by calling "dotnet run {appname}"). 

After that, when a single-file publishing is employed (and it is, because without it, loading fails in even more complicated ways), the toolkit packs multiple assemblies into a single file and adds a stub that does the unpacking and loading. 

It _may_ be possible to patch the native stubs and .so's (in the case they reference other .so's), however, even this may not help depending on how references to the native libraries are set up. The .NET loader uses its fixed algorithms to check some custom set of paths (or a custom resolver I've been speaking about). 

So, realistically, a custom resolver is the .NET way to control the paths. It is a legitimate way used in many scenarios, but it seems to be overkill here, and then we are back to the question of how to determine which paths to check.

emayevski:
bot, build

com.alliedbits.ABCalcPlus.json
```diff
@@ -0,0 +1,47 @@
+{
+  "app-id": "com.alliedbits.ABCalcPlus",
+  "runtime": "org.freedesktop.Platform",
+  "runtime-version": "25.08",
+  "sdk": "org.freedesktop.Sdk",
+  "command": "abcalc",
+  "finish-args": [
+    "--share=network",
```
bbhtt: why does it need network?

emayevski: For checking for updates.

bbhtt: How does that work? It is updated via Flatpak.

emayevski: Grabs changes.txt from the site and parses it.

bbhtt: So for displaying update notification or displaying changelog or both?

emayevski: The update window displays a part of the change log relevant to the user (what has changed since the user's version) and offers an option to go to the site for an update.

com.alliedbits.ABCalcPlus.json
```diff
@@ -0,0 +1,47 @@
...
+    "--share=ipc",
+    "--socket=fallback-x11",
+    "--socket=wayland",
```
bbhtt: Not working on wayland DE afaik avalonia doesn't support native wayland.

```
flatpak run com.alliedbits.ABCalcPlus
Unhandled exception. System.Exception: XOpenDisplay failed
   at Avalonia.X11.AvaloniaX11Platform.Initialize(X11PlatformOptions options)
   at Avalonia.AvaloniaX11PlatformExtensions.<>c.<UseX11>b__0_0()
   at Avalonia.AppBuilder.SetupUnsafe()
   at Avalonia.AppBuilder.Setup()
   at Avalonia.AppBuilder.SetupWithLifetime(IApplicationLifetime lifetime)
   at Avalonia.ClassicDesktopStyleApplicationLifetimeExtensions.StartWithClassicDesktopLifetime(AppBuilder builder, String[] args, Action`1 lifetimeBuilder)
   at ABCalc.Desktop.Program.Main(String[] args)
```

com.alliedbits.ABCalcPlus.json
hfiguiere: > and offers an option to go to the site for an update.

which is an anti-pattern with flatpak.

com.alliedbits.ABCalcPlus.json
emayevski: From [the article on the avalonia site](https://avaloniaui.net/blog/bringing-wayland-support-to-avalonia):

```There's a technology called XWayland that acts as a compatibility layer, running X11 applications on Wayland systems. It's installed by default on virtually every desktop Linux distribution.

Think of XWayland like Windows' compatibility mode, your application thinks it's talking to X11, but behind the scenes, XWayland is translating those calls to Wayland. You might miss out on some Wayland-specific benefits (like better HiDPI scaling or reduced latency), but your application will run.
```

So, if the application did not work on your system, this means that the compatibility layer is not there. I don't think we can do anything on our side in regard to this, can we? Or, am I missing something, and we need to change the flags?

com.alliedbits.ABCalcPlus.json
emayevski: @hfiguiere 

In my businesses, we make updates of our products as soon as some bug is fixed or a new feature is added. This means _frequently_. And if updates are accepted to FlatHub with the same speed as the first submission, then we better have an update mechanism that lets users grab a fresh copy when it is released, not in a couple of months.

com.alliedbits.ABCalcPlus.json
hfiguiere: No this not what it means. It means that you falsely claim with the chosen permission that you support wayland when you don't. 

Maybe you missed that part:

https://docs.flatpak.org/en/latest/sandbox-permissions.html#permissions-guidelines (it is linked directlyfrom our documentation)

What you posted above is just mansplaining.

emayevski: @hfiguiere I admit we didn't know that Avalonia does not support Wayland. I will update the submission.

emayevski:
bot, build

PR #8794, opened 2026-05-25
Add org.praat.Praat

LinuxSBC:

Praat is a widely used tool in the field of Linguistics. It allows for a wide range of computational linguistic analyses, both with a GUI and through scripts in its custom language.

https://github.com/user-attachments/assets/87516c1f-ce8b-4b1a-8623-ab45f4d44eff

- [x] I am an upstream contributor to the project. If not, I contacted upstream developers about this submission. **Link:**  https://github.com/praat/praat.github.io/issues/2992

This Flatpak build requests --filesystem=home permissions, since Praat scripts often need to access other files without using the xdg-document-portal. I have requested an exception for this: https://github.com/flathub-infra/flatpak-builder-lint/pull/1044


org.praat.Praat.yml
```diff
@@ -0,0 +1,33 @@
...
+  - --filesystem=home
+
+modules:
+  - name: praat
+    buildsystem: simple
+    build-commands:
+      - make -j15
+      - install -Dm0755 praat -t ${FLATPAK_DEST}/bin/
+      - install -Dm0644 main/praat-480.svg ${FLATPAK_DEST}/share/icons/hicolor/scalable/apps/org.praat.Praat.svg
+      - sed -i 's/^Icon=praat$/Icon=org.praat.Praat/' main/praat.desktop
```
hfiguiere: Never use sed to change desktop files. It has no failure mode. Use the command `desktop-file-edit`.

org.praat.Praat.metainfo.xml
hfiguiere, line 1: This file should be (submitted and managed) upstream.

org.praat.Praat.yml
```diff
@@ -18,16 +18,11 @@ modules:
     buildsystem: simple
     build-commands:
       - make -j15
```
hfiguiere: what if the builder only has 2 cores??????

screenshot-1.png
hfiguiere, line 1: screenshots belong upstream too

LinuxSBC:
bot, build

Edit: Sorry, I forgot there's not yet an exception for it. I just wanted to make sure it also builds successfully on the Flathub infra, but it should; I've tried it on multiple computers and Github Actions before.

org.praat.Praat.yml
```diff
@@ -0,0 +1,26 @@
...
+
+modules:
+  - name: praat
+    buildsystem: simple
+    build-commands:
+      - make
+      - make install PREFIX=${FLATPAK_DEST}
+    sources:
+      - type: git
+        url: https://github.com/praat/praat.github.io
```
bbhtt, line 24: there is no CI in the repo

org.praat.Praat.yml
```diff
@@ -0,0 +1,26 @@
...
+        commit: a12f3c875986f3812b022ee9cd001ca4b60ab6e8
```
bbhtt: why not using a release?

org.praat.Praat.yml
```diff
@@ -0,0 +1,26 @@
...
+  - --socket=pulseaudio
+  - --socket=cups
+  - --persist=.praat-dir
+  - --filesystem=home
+
+modules:
+  - name: praat
+    buildsystem: simple
+    build-commands:
+      - make
```
bbhtt: `make -j $FLATPAK_BUILDER_N_JOBS` was the solution not removing `-j` unless it is broken with parallel build.

LinuxSBC: Thanks, sorry. I'll fix that.

org.praat.Praat.yml
LinuxSBC: Are you asking why the SHA doesn't point to a release tag? It usually does, I'm just doing this so that it builds until an [upstream PR](https://github.com/praat/praat.github.io/pull/3269) for the AppStream metadata is accepted.

org.praat.Praat.yml
LinuxSBC: Yes. Is that a problem?

bbhtt: yes

LinuxSBC: Okay, I'm sorry. Just to be sure I'm understanding properly, are you saying that I need to set up a Github Actions runner in upstream Praat to build it automatically on each commit?

I'm somewhat reluctant to change things too much about the upstream build process because they've been doing it for over 30 years in the way they're accustomed to, but I can see if they'd be comfortable with that.

Or are you referring to this PR repo? Because my understanding was that Flathub handles that automatically, such as with the flathubbot here. In case you're referring to the lack of external-data-checker use, I realized it's a good idea to add regardless, so I've done so.

org.praat.Praat.yml
LinuxSBC: But I've now switched it back to a release tag. It no longer builds due to the upstream PR not being accepted yet, but I can fix that as soon as it is.

LinuxSBC:
This manifest currently targets version 6.4.67 of Praat, which doesn't yet have a few upstream changes I ended up needing to merge, so it does not build in its current form. If I target it at the latest master commit, it does build, but I realize that's not suitable for merging and causes problems for the external-data-checker.

I'm in active communication with the primary developers of Praat, and they'd probably be willing to get a release out for this, but I only want to do that once I know that no further upstream changes will be needed.

Please just let me know when you think no more upstream changes will be required, and I can see if Praat can make a release.

org.praat.Praat.yml
bbhtt: Upstream repo, do they have any CI testing things?

LinuxSBC: They do not. I can try to get that set up to build it and run the existing test scripts on push in Github Actions; would that be sufficient?

bbhtt: yes

LinuxSBC: Okay, thanks. I'll see what I can do.

LinuxSBC:
I'm sorry I took so long. Praat upstream now has CI, and since it has had a new release, the YAML file now correctly matches the commit hash and tag. What else needs to be done at this point?

LinuxSBC:
bot, build

PR #8813, opened 2026-05-27
Add com.axantum.xecrets.ez

xecrets:

Password protect files and text with strong AES-256 cryptography, YubiKey support, generate strong passwords, create Shamir's Secret Shares for password and seed backup and much more. Runs on Linux, macOS and Windows. 100% compatible with AxCrypt.
- [X] Please attach a video showcasing the application on Linux using the Flatpak. [Xecrets Ez Flatpak demo.webm](https://github.com/user-attachments/assets/b61f5c85-75ba-49bf-8fcd-8045057e3f83)


com.axantum.xecrets.ez.yml
```diff
@@ -0,0 +1,110 @@
+app-id: com.axantum.xecrets.ez
+runtime: org.freedesktop.Platform
+runtime-version: "25.08"
+sdk: org.freedesktop.Sdk
+command: XecretsEz
+
```
ghost, line 8:
```suggestion
command: XecretsEz
tags:
  - proprietary
```

ghost (changes requested):

xecrets:
bot, build

com.axantum.xecrets.ez.yml
```diff
@@ -0,0 +1,111 @@
...
+        strip-components: 1
+        x-checker-data:
+          type: html
+          url: https://www.axantum.com/rss/releases
+          version-pattern: 'Xecrets Ez release ([0-9]+\.[0-9]+\.[0-9]+) for macOS, Linux and Windows'
+          url-template: https://www.axantum.com/build/linux/XecretsEz-Linux-x64-$version.tar.gz
+        only-arches:
+          - x86_64
+      - type: archive
+        url: https://www.axantum.com/build/linux/XecretsEz-Linux-arm64-2.3.1511.tar.gz
```
ghost: This URL does not exist and returns 404.

com.axantum.xecrets.ez.yml
```diff
@@ -0,0 +1,111 @@
...
+      - install -D -m 0644 XecretsEz.png /app/share/icons/hicolor/256x256/apps/com.axantum.xecrets.ez.png
+      - install -D -m 0644 xecrets-ez-16x16.png /app/share/icons/hicolor/16x16/mimetypes/com.axantum.xecrets.ez-application-vnd.xecrets-encrypted.png
+      - install -D -m 0644 xecrets-ez-32x32.png /app/share/icons/hicolor/32x32/mimetypes/com.axantum.xecrets.ez-application-vnd.xecrets-encrypted.png
+      - install -D -m 0644 xecrets-ez-48x48.png /app/share/icons/hicolor/48x48/mimetypes/com.axantum.xecrets.ez-application-vnd.xecrets-encrypted.png
+      - install -D -m 0644 xecrets-ez-128x128.png /app/share/icons/hicolor/128x128/mimetypes/com.axantum.xecrets.ez-application-vnd.xecrets-encrypted.png
+      - install -D -m 0644 xecrets-ez-256x256.png /app/share/icons/hicolor/256x256/mimetypes/com.axantum.xecrets.ez-application-vnd.xecrets-encrypted.png
+      - update-mime-database /app/share/mime
+    sources:
+      - type: archive
+        url: https://www.axantum.com/build/linux/XecretsEz-Linux-x64-2.3.1511.tar.gz
```
ghost: This too.

xecrets: Sorry, my bad... Fixed now.

xecrets:
bot, build

README.md
```diff
@@ -0,0 +1,90 @@
...
+access was sandboxed to a few specific locations. It can also be used to overwrite files, which is
+more secure than just deleting them from the file system, and that is another reason why host file
+system access is needed.
+
+Network access is required in order to access files on network shares, and also for the single instance
+logic to work properly since it is implemented with named pipes and sockets.
+
+Both x11 and Wayland was added after actual compatibility issues were discovered when running in
+certain environments, for example a virtual machine under UTM. Just fallback x11 did not work
+unfortunately.
```
hfiguiere: gotta have to fix that then.

com.axantum.xecrets.ez.yml
```diff
@@ -0,0 +1,111 @@
...
+  - --socket=x11
+  - --filesystem=host
+  - --socket=pcsc
+  - --device=dri
+
+modules:
+  - name: pcsc-lite
+    buildsystem: meson
+    config-opts:
+      - --libdir=lib
```
hfiguiere:
```suggestion
```

com.axantum.xecrets.ez.yml
```diff
@@ -0,0 +1,111 @@
...
+      - -Dlibudev=false
+      - -Dlibsystemd=false
+      - -Dpolkit=false
+    post-install:
+      - rm /app/sbin/pcscd
+      - rmdir /app/sbin || true
```
hfiguiere: this should be a cleanup rule for `/sbin`

com.axantum.xecrets.ez.yml
```diff
@@ -0,0 +1,111 @@
...
+      - type: file
+        path: linux-desktop/xecrets-ez-16x16.png
+      - type: file
+        path: linux-desktop/xecrets-ez-32x32.png
+      - type: file
+        path: linux-desktop/xecrets-ez-48x48.png
+      - type: file
+        path: linux-desktop/xecrets-ez-128x128.png
+      - type: file
+        path: linux-desktop/xecrets-ez-256x256.png
```
hfiguiere: none of these files exist

```suggestion
```

README.md
xecrets: Not sure what you mean there. I don't think it's possible for me to fix that Avalonia UI under UTM with flatpak doesn't work with 'fallback x11'. If I misunderstand, please advise!

hfiguiere: Then we can wait you fix it to accept this.

com.axantum.xecrets.ez.yml
xecrets: That's needed. It won't work otherwise, it'll place it in the wrong place and won't be found IIRC. If there's an alternate solution, please advise.

hfiguiere: No it's not. It is just that you use the wrong version of flatpak builder (ie your distro didn't bother to update it). The doc clearly recommend to use the one we provide as a flatpak.

README.md
xecrets: What do you mean? I can change to 'fallback x11' - with the result that it won't work in some environments. That's no good? Avalonia UI is the probably most used cross-platform framework available, but it's way out of my league to fix that!

com.axantum.xecrets.ez.yml
xecrets: They're in the archive, it works fine to build the flatpak with this manifest. Is there something I don't understand about the build process here?

xecrets:
The build fails with the following warnings:

'finish-args-contains-both-x11-and-wayland' error found in linter manifest check.
'finish-args-host-filesystem-access' error found in linter manifest check.

Both are noted during validation before submission, and the rationale explained in the README.

I respectfully ask for an exception for these, with the background stated there, please.

hfiguiere:
`xecrets.axantum.com` doesn't resolve.

xecrets:
> `xecrets.axantum.com` doesn't resolve.

No, it doesn't. I was not aware that it should. I was under the impression that 'axantum.com' should resolve (which it does). I can add that to the DNS and have it redirect to www.axantum.com (like axantum.com does), but I must have missed or misunderstood the requirements somehow, can you please give me a pointer to the appropriate section? Thanks!

hfiguiere:
But you acknowledged

> The Flatpak ID follows all the rules listed in the [Application ID requirements](https://docs.flathub.org/docs/for-app-authors/requirements#application-id).

So you should have been aware.

com.axantum.xecrets.ez.yml
xecrets: I got the manifest code for pcscd from Yubico, is that wrong too? I kind of assumed it would be correct there. See https://github.com/flathub/com.yubico.yubioath/blob/master/com.yubico.yubioath.yml .

Do you mean that I should change it to:

```
- rm /sbin/pcscd
- rmdir /sbin || true
```

Please advise.

xecrets:
> So you should have been aware.

I am trying here, but there's a lot of text, and it's easy to miss some aspect. So, in this case, I guess it's a matter of "I read but did not understand fully". Sorry. Probably you are referring to:

_The domain from the ID is determined by reversing its components and demangling them to form a FQDN, which is then matched against [PSL](https://publicsuffix.org/). If a match is not found in PSL, all component except the last from the previous final form is taken as the domain._

Ok, no problem. I can make 'xecrets.axantum.com' to resolve, but... I now also read:

_A dash - is only allowed in the last component._

So..., perhaps the cleanest fix is to change the App ID to `com.axantum.xecrets-ez`. Is that acceptable and preferred?

I appreciate all the help and feedback in getting this right.

com.axantum.xecrets.ez.yml
xecrets: > They're in the archive, it works fine to build the flatpak with this manifest. Is there something I don't understand about the build process here?

My bad, this is a learning process... It was a misunderstanding and an artifact of my build process which does things to the manifest depending on the target environment. Will fix and update the PR. Thank you.

com.axantum.xecrets.ez.yml
hfiguiere: it's wrong.

com.axantum.xecrets.ez.yml
xecrets: Fixed.

README.md
xecrets: Removed Wayland.

com.axantum.xecrets.ez.yml
xecrets: Fixed by writing a small wrapper script setting LD_LIBRARY_PATH appropriately.

com.axantum.xecrets.ez.yml
xecrets: Fixed. Sorry about that, thanks for pointing it out.

xecrets:
bot, build

com.axantum.xecrets.ez.yml
xecrets: Fixed.

xecrets:
> ❌ [Test build](https://github.com/flathub-infra/vorarbeiter/actions/runs/26578492057) failed.

I believe I have addressed all reviewer comments.

The build now fails due to the 'finish-args-host-filesystem-access' error found in linter manifest check.

Please consider granting an exception to this rule for this app, as the app is not really a suitable candidate for file system sandboxing as described in the README and previously.

com.axantum.xecrets.ez.yml
```diff
@@ -0,0 +1,89 @@
...
+      - install -D -m 0644 linux-desktop/xecrets-ez-32x32.png /app/share/icons/hicolor/32x32/mimetypes/$FLATPAK_ID-application-vnd.xecrets-encrypted.png
+      - install -D -m 0644 linux-desktop/xecrets-ez-48x48.png /app/share/icons/hicolor/48x48/mimetypes/$FLATPAK_ID-application-vnd.xecrets-encrypted.png
+      - install -D -m 0644 linux-desktop/xecrets-ez-128x128.png /app/share/icons/hicolor/128x128/mimetypes/$FLATPAK_ID-application-vnd.xecrets-encrypted.png
+      - install -D -m 0644 linux-desktop/xecrets-ez-256x256.png /app/share/icons/hicolor/256x256/mimetypes/$FLATPAK_ID-application-vnd.xecrets-encrypted.png
+      - update-mime-database /app/share/mime
+    sources:
+      - type: script
+        dest-filename: XecretsEz.sh
+        commands:
+          - export LD_LIBRARY_PATH="/app/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
```
bbhtt, line 69: you should not need to set LD_LIBRARY_PATH, just install libraries to `/app/lib`.

README.md
bbhtt, line 1: you can drop this file. this is not a diary.

README.md
```diff
@@ -0,0 +1,78 @@
...
+
+Host file system access is required because of the nature of the application. It is a cross-platform
+application intended to encrypt and decrypt files in all locations on the system, including external
+devices and network shares. It would not be natural, and it would cripple it quite severaly if file
+access was sandboxed to a few specific locations. It can also be used to overwrite files, which is
+more secure than just deleting them from the file system, and that is another reason why host file
+system access is required.
+
+Also some features are likely to break and make the app less useful, notably the ability to copy
+and paste files to encrypt/decrypt/open them. This won't work seamlessly with portals.
```
bbhtt: what do you mean by copy paste? what does it have to do with portals?

com.axantum.xecrets.ez.yml
```diff
@@ -0,0 +1,89 @@
...
+sdk: org.freedesktop.Sdk
+command: XecretsEz
+tags:
+  - proprietary
+
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=x11
+  - --filesystem=host
```
bbhtt: why does it not use filechoosers to select the file to encrypt decrypt?

com.axantum.xecrets.ez.yml
xecrets: That would be ideal, but apparently .NET 10 will not find native libraries if LD_LIBRARY_PATH is not set. See https://learn.microsoft.com/en-us/dotnet/standard/native-interop/best-practices "Troubleshoot P/Invoke failures", "On Linux, verify LD_LIBRARY_PATH or rpath". Unfortunately without setting LD_LIBRARY_PATH it does not work. With LD_LIBRARY_PATH it does. I see no harm in setting LD_LIBRARY_PATH. If there is another way to make the Yubico SDK find the library, please help me out!

bbhtt: that framework is broken then.

README.md
xecrets: Exactly. So, when a user copies and pastes one or more files to the app, with the intention that the app should open/encrypt/decrypt the file, it will not work because the paths inside the sandbox for files that need access via portals are not the same as outside the sandbox where the copy/paste lives. At least that is what I understand, thus copy/paste will be broken and there's no simple way to request access through a portal. If I am wrong, please let me know! The effect being that the app is partially crippled, and I will have to implement flathub-specific code to disable this very popular feature.

bbhtt: It is about drag and drop then? There exists a portal for it and Portals are in no way `Flathub` specific -  they work on every Linux distribution. Maybe you should research portals.

com.axantum.xecrets.ez.yml
xecrets: It does use file and folder pickers, but it also supports other ways to select files including double-click to open, as well as a recent files list and perhaps primarily copy/paste. On Windows and macOS also Drag and Drop, but that is unfortunately broken/not implemented on Linux by the cross-platform framework used - Avalonia UI. But the main reason for `--filesystem=host` is because the use case for the app is to be able to encrypt/decrypt/wipe any file, and also recursively in directory hierarchies - without further UI complications. The driving force behind publishing on Flathub is to make it easier for users to install, not to make it harder to use. It doesn't quite fit the "document model" like a text editor. It *can* be squeezed into such a model, but it will lead to a degraded user experience, and in the end the migration of some users from the Flathub distributed version to a plain flatpak where full permissions are granted.

Many users will also be users who have been using this app and it's predecessors on Windows and macOS for years and even decades, but who want to bring it with them when they migrate or also start using Linux. I would prefer to be able to offer them finally a fully featured solution that is equivalent to their expectations, and not a somewhat limited experience on Linux.

README.md
xecrets: It was intended as initial extended information as a service for the reviewers, not as a diary. As all of what it says has been re-iterated in the comments here, it probably serves no purpose anymore and perhaps never did. Not even as a diary. Sure, I'll drop it.

xecrets: Deleted.

com.axantum.xecrets.ez.yml
bbhtt: double click to open does not require host and recent files menu works with portals.

README.md
xecrets: It's not just about Drag and Drop as of now. As mentioned, this is not supported on Linux. As mentioned, it's about the app not really fitting the restricted document-centric model of portals. There's a partial fit, but the full functionality of the app will be crippled if file system access is restricted.

One comparison might be https://github.com/flathub/org.kde.okular (which by the way has a similar README diary as the one that was requested to be deleted here).

bbhtt: have you actually tried to see if it works with portals? You are constantly saying "full functionality will be crippled" and I don't see evidence of where you tried.

Some of the usecases you mentioned do not need `host` permissions.

Past examples don't count. The review is about your app not Okular.

com.axantum.xecrets.ez.yml
xecrets: Well, be that as it may... If so, what's the harm of setting LD_LIBRARY_PATH to the accepted value `/app/lib`? You and I might be annoyed that it's necessary, but simple harmless workarounds is not a problem, is it?

com.axantum.xecrets.ez.yml
xecrets: Recent files menu works so-so with portals. This is how a recent file list can look like when the file is accessed via the portal. Note the user id etc inserted in the path seen by the app. It kind of works, but it's what I call a degraded user experience.

Double-click will open the file - but once again, if the file is outside of the sandbox, the path seen and subsequently possibly exposed to the user will be mangled.

Recent files via portals

README.md
xecrets: Hi!

Copy/paste is completely broken for files outside the sandbox. The app does not see the file list on the clipboard, there is nothing to paste.

But much worse, and essentially a risk for data loss by the non-understanding user, is that the app works by reading a file, transforming it (encrypting/decrypting), and then writing it back, first to a temporary copy so as not to risk data loss if the write fails or is aborted, then deleting the original and finally renaming the temp to a new name. For example an encrypted "readme.md" becomes "readme-md.axx".

But when accessed outside of the sandbox, the names are completely mangled, like this:
Encrypted and decrypted files via portal

The files contain the correct data, but the final rename appears to fail, leaving this result. This does not happen inside the sandbox, and thus of course not with host file system permissions. It's probably caused by a failure in the final renaming step, leaving the temporary with the portal name there instead.

Portal access is not transparent, and access outside the sandbox becomes fraught with complications that in the end may affect the end user, even causing effective data loss (not all, or even many, of the app users will understand how to see and rename the file to restore it's usefulness).

This is of course not the final and complete list of issues, but it's a sample.

For the app to be safe to use, I would have to add code forbidding the user to access files outside the sandbox, as well as disabling the recent files list - or at least removing files from outside the sandbox. As I understand it there is no reliable way to determine if a given path is inside or outside the sandbox. Path inspection is a fragile hack. So that seems like a bit of a catch-22, apart of course from the fact that app authors should not be required to do major changes to an app to detect and limit it's operation inside a Flatpak sandbox, when it's just not that type of app.

I am of course open to reasonable suggestions and to learn what I don't know, but these are my current conclusions. The app is not a good fit for a limited file system sandbox, but a very good fit for Flathub as long as it's allowed to operate as it's intended to operate, and does operate in other environments.

com.axantum.xecrets.ez.yml
bbhtt: The portal has an API to get the actual host path and not the mangled one.

xecrets: Yes, that might enable prettier display of recent files located outside the sandbox, *if* I make the application portal aware *and* implement persistence of both paths when running in a sandbox for paths outside the sandbox. 

Unfortunately it does nothing for all the other much more problematic issues.

I am of course willing to make minor adjustments to the app to handle running inside a sandbox, but I do not have the resources to completely rewrite the app, and even if I could, the user experience will be at the best changed, at the worst degraded. Please remember that this is a cross-platform app running on macOS and Windows and Linux outside of Flatpak sandboxes as well, and it should behave in a similar way with a similar feature set in all situations, with as much as possible of the code being unaware of the platform.

I understand and respect the intention of Flatpak and Flathub, but this app has some use cases that complicate things.

- The user selects a folder and the app must enumerate files recursively
- It creates encrypted/decrypted files next to originals. Doesn't work outside sandbox.
- It processes multiple files. Repeated chooser prompts just wont work in practice.
- Portal document paths are not useful as persisted paths, requires significant rewrite just for Flathub just for recent files.
- It persists file locations internally for later re-encryption when files are opened for view/editing. This won't work well either.
- It can rename encrypted files to random names to hide information in the file name. Doesn't work outside the sandbox.

I am sure there are more use cases too, but I don't have an exhaustive list.

bbhtt: Can you post shorter replies?

What is the problem here? You said double click works and recent files menu works minus mangled paths which can be fixed by using the API.

> if I make the application portal aware and implement persistence of both paths when running in a sandbox for paths outside the sandbox.

The requirements make it mandatory when the portal API exists and fits the usecase. So it's not a question of 'if'. You don't want to do so, from the beginning and the entire reasoning seems to be based on "this is cross platform, i have been doing this since XYZ years".

Persistence of paths is guaranteed by the Portal and Flatpak you don't have to implement anything there.

xecrets: Broken use cases:

Core encryption/decryption: "file.txt" <=> "file-txt.axx"
Anonymous rename: "file-txt.axx" => "axhjuxgx.axx"
Collision avoidance: "file.txt" => "file-txt.axx" (exists already) => "file-txt (1).axx"

No API I know of. Request through save dialog not possible.

bbhtt: Ok you can open an exception PR.

com.axantum.xecrets.ez.yml
```diff
@@ -0,0 +1,93 @@
...
+runtime: org.freedesktop.Platform
+runtime-version: "25.08"
+sdk: org.freedesktop.Sdk
+command: XecretsEz
+tags:
+  - proprietary
+
+finish-args:
+  - --share=ipc
+  - --share=network
```
bbhtt: why does it need network?

com.axantum.xecrets.ez.yml
```diff
@@ -0,0 +1,93 @@
...
+      - install -D -m 0644 linux-desktop/xecrets-ez-256x256.png /app/share/icons/hicolor/256x256/mimetypes/$FLATPAK_ID-application-vnd.xecrets-encrypted.png
+      - update-mime-database /app/share/mime
+    sources:
+      - type: script
+        dest-filename: XecretsEz.sh
+        commands:
+          - export LD_LIBRARY_PATH="/app/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
+          - exec /app/bin/XecretsEz-bin "$@"
+      - type: archive
+        url: https://www.axantum.com/build/linux/XecretsEz-Linux-x64-2.3.1531.tar.gz
```
bbhtt: i remember you said somewhere along the lines of this software has existed for a long time and I couldn't find a way to verify that as far as I remember.

com.axantum.xecrets.ez.yml
xecrets: It needs an abstract socket that is system-global to ensure single-instancing when co-existing with a plain tarball 'installation'.

com.axantum.xecrets.ez.yml
xecrets: https://www.axantum.com/axcrypt-history

Open source components:
[xecrets-net](https://github.com/axantum/xecrets-net) - First github commit Nov 7, 2012
[xecrets-cli](https://github.com/xecrets/xecrets-cli) - First github commit Jun 13, 2023
Predecessor:
[AxCrypt 1.x](https://github.com/xecrets/xecrets-file-classic) - First github commit Dec 18, 2011 (Before then, Subversion. Released in 2001)

Proprietary current iteration GUI front-end to xecrets-cli, [Xecrets Ez](https://www.axantum.com/release-notes), released Aug, 2023.

If you wish to view Xecrets Ez repository, let me know and I'll invite you.

bbhtt:
bot, build

com.axantum.xecrets.ez.yml
bbhtt: Using abstract socket for single instancing is an odd choice and since there is no network namespace this also grants the entirety of network access.

Single instancing is done by registering a well known name on the session bus which is the app id. That's how GTK and Qt does it.

xecrets: >  ... odd choice ...

Using sockets on Linux is nearly identical with macOS for this purpose. Cross-platform apps try to minimize the amount of platform specific code.

> Single instancing is done by registering a well known name on the session bus ...

That's one Desktop Linux-specific way.

Are you requiring the app to remove `--share=network` from the manifest for approval?

bbhtt: Yes ideally.

xecrets: > Yes ideally.

Interpreted as a "_Yes, you must remove it to be approved_". So now it is. It would have been nice to have it approved for now, and trust I will fix it in the next release. But let's move forward with this.

bbhtt: at some point your are going to have to write some platform specific code and doing so is better when the mutiplatform code is requiring to open unnecessary sandbox holes.

otherwise you may as well stuff the .exe inside WINE.

bbhtt:
bot, build

xecrets:
bot, build

xecrets:
bot, build

com.axantum.xecrets.ez.yml
bbhtt: an external source will be ideal to verify this. otherwise it is a bit circular.

bbhtt:
I noticed the website showing a paid version, how is that supposed to work in regards to the flatpak here? Is it the paid one or the free one being packaged here?

You removed network access so can licences still be activated or the paid version is something different entirely?

com.axantum.xecrets.ez.yml
bbhtt: Nevermind I saw the GitHub repo missed that before.

xecrets:
> Is it the paid one or the free one being packaged here?

Same bits. Use a license key text string/file to activate paid features.

> You removed network access so can licences still be activated [...]

Yes, paste the license key string (or a file) to the app. The procedure is shown in the video that was part of the PR.

The app has never required network *access* - it only needed network *permissions* in the Flatpak sandbox to use abstract sockets. This is no longer the case.

PR #8821, opened 2026-05-28
Add de.gonicus.gonnect.plugins.Matrix

cajus:

- [x] Please describe the application briefly. This flatpak extends de.gonicus.gonnect app to handle matrix chat. It contains a plugin executable that is started and controlled by the main application. Please note that the current "main" de.gonicus.gonnect app has the interface definition in its manifest, but does not do anything with it yet. Chat will be available with the next major version, so you can't test this plugin with what's on flathub yet.
- [x] Please attach a video showcasing the application on Linux using the Flatpak. N/A - as this is a plugin for de.gonicus.gonnect.plugins interface and has no GUI by itself.
   

### DIFF
--- a/de.gonicus.gonnect.plugin.Matrix.yml
+++ b/de.gonicus.gonnect.plugin.Matrix.yml
@@ -1,0 +1,45 @@
+id: de.gonicus.gonnect.plugin.Matrix
+runtime: de.gonicus.gonnect
+runtime-version: 'stable'
+branch: "2"
+
+sdk: org.kde.Sdk//6.10
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+
+build-extension: true
+separate-locales: false
+
+build-options:
+  prefix: /app/plugins/matrix
+  prepend-path: /app/plugins/matrix
+  env:
+    RUSTFLAGS: "-C link-arg=-fuse-ld=mold"
+
+command: gouda_matrix
+
+modules:
+  - name: gouda-matrix
+    buildsystem: simple
+
+    build-options:
+      append-path: /usr/lib/sdk/rust-stable/bin
+      env:
+        CARGO_HOME: /run/build/gouda-matrix/cargo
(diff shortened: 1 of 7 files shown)


de.gonicus.gonnect.plugin.Matrix.yml
```diff
@@ -0,0 +1,45 @@
+id: de.gonicus.gonnect.plugin.Matrix
```
bbhtt, line 1: why can't this be in the main Flatpak?

cajus: The main flatpak is the base application and is primary a VoIP client. It can be extended by various chat, video conferencing, teams like office integration, etc. modules. Matrix is the first of them we're providing here.

These plugins are separate projects that dock into the plugin interface. They are not released together with the main application and have their own agenda.

Putting everything together in one big flatpak doesn't fit the release process of the main application, would (depending on the module) drastically increase the complexity of the manifests, the build time and the installed size on the user side - even if the user is just using GOnnect as a simple VoIP client.

bbhtt: How big is this and are there other plugins planned?

cajus: What do you mean by "this"? The matrix plugin seems to be around 72MB installed size. A potential LibreOfficeKit plugin would be definitively > 200MB - but I can't give any exact values here, as there may be a big difference between rpm based and flatpak runtime based builds. There are more plugins coming, but the first one is the one we're talking about currently. As the chat interface is abstracted, you can i.e. install a Rocket.Chat plugin instead of a Matrix one. From the user experience point of view: install what you want to use, and it's in there automatically.

If your question aims at "that's small, put it all in one": as said, these are different code bases that are loosely coupled, being maintained by different people and having different release cycles. That doesn't feel good to have that all in one. Not from a maintenance point of view, and not from "resources" (i.e. build time, network resources for people who don't want it) point of view.

PR #8882, opened 2026-06-04
Add page.codeberg.grinka.Transition

OleksiiGrinka:

Transition is a simple multimedia to audio converter application written in rust using libadwaita & relm4
- [x] Please attach a video showcasing the application on Linux using the Flatpak. https://github.com/user-attachments/assets/bf73cf20-ba1f-4b19-9dcf-4b6af96a6270

### DIFF
--- a/page.codeberg.grinka.Transition.json
+++ b/page.codeberg.grinka.Transition.json
@@ -44,8 +44,8 @@
                 {
                     "type" : "git",
                     "url" : "https://codeberg.org/grinka/transition.git",
-                    "tag": "v1.3",
-                    "commit": "869d19d84c65d8879b4aeea8bb5d5a6b3a8fbd00"
+                    "tag": "v2.0",
+                    "commit": "c37f2f67127671c3a71f6e2d56ff552480665262"
                 },
                 "cargo-sources.json"
             ]
--- a/page.codeberg.grinka.Transition.json
+++ b/page.codeberg.grinka.Transition.json
@@ -1,0 +1,53 @@
+{
+    "id" : "page.codeberg.grinka.Transition",
+    "runtime" : "org.gnome.Platform",
+    "runtime-version" : "50",
+    "sdk" : "org.gnome.Sdk",
+    "command" : "transition",
+    "sdk-extensions" : [
+        "org.freedesktop.Sdk.Extension.rust-stable"
+    ],
+    "finish-args" : [
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--device=dri",
+        "--socket=wayland",
+        "--filesystem=xdg-music:create",
+        "--filesystem=xdg-videos:create",
+        "--filesystem=xdg-download:create"
+    ],
+    "modules" : [
+        {
+            "name" : "Transition",
+            "buildsystem" : "simple",
+            "build-options" : {
+                "append-path" : "/usr/lib/sdk/rust-stable/bin",
+                "env" : {
+                    "CARGO_HOME": "/run/build/Transition/cargo",
+                    "CARGO_NET_OFFLINE": "true",
+                    "RUST_BACKTRACE" : "1",
+                    "RUST_LOG" : "transition=debug"
+                }
+            },
+            "build-commands": [
+                "install -Dm0644 cargo/config .cargo/config.toml",
+                "cargo --offline fetch --manifest-path Cargo.toml --verbose",
+                "cargo build --offline --release --all-features",
+                "install -Dm0755 target/release/transition ${FLATPAK_DEST}/bin/transition",
+                "install -Dm0644 data/icons/${FLATPAK_ID}.svg ${FLATPAK_DEST}/share/icons/hicolor/scalable/apps/${FLATPAK_ID}.svg",
+                "install -Dm0644 data/icons/page.codeberg.grinka.Transition-symbolic.svg ${FLATPAK_DEST}/share/icons/hicolor/symbolic/apps/${FLATPAK_ID}-symbolic.svg",
+                "install -Dm0644 data/page.codeberg.grinka.Transition.desktop ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop",
+                "install -Dm0644 data/page.codeberg.grinka.Transition.service ${FLATPAK_DEST}/share/dbus-1/services/${FLATPAK_ID}.service",
+                "install -Dm0644 data/page.codeberg.grinka.Transition.metainfo.xml ${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.metainfo.xml"
+            ],
+            "sources" : [
+                {
+                    "type" : "git",
+                    "url" : "https://codeberg.org/grinka/transition.git",
+                    "tag": "v1.1"
+                },
+                "cargo-sources.json"
+            ]
+        }
+    ]
+}
--- a/page.codeberg.grinka.Transition.json
+++ b/page.codeberg.grinka.Transition.json
@@ -12,9 +12,9 @@
         "--socket=fallback-x11",
         "--device=dri",
         "--socket=wayland",
-        "--filesystem=xdg-music:create",
-        "--filesystem=xdg-videos:create",
-        "--filesystem=xdg-download:create"
+        "--filesystem=xdg-music",
+        "--filesystem=xdg-videos",
+        "--filesystem=xdg-download"
     ],
     "modules" : [
         {
@@ -44,7 +44,8 @@
                 {
                     "type" : "git",
                     "url" : "https://codeberg.org/grinka/transition.git",
-                    "tag": "v1.1"
+                    "tag": "v1.2",
+                    "commit": "1746ad2d7481014e8a7b7a3155850c011f3fdd67"
                 },
                 "cargo-sources.json"
             ]
--- a/page.codeberg.grinka.Transition.json
+++ b/page.codeberg.grinka.Transition.json
@@ -44,8 +44,8 @@
                 {
                     "type" : "git",
                     "url" : "https://codeberg.org/grinka/transition.git",
-                    "tag": "v1.2",
-                    "commit": "1746ad2d7481014e8a7b7a3155850c011f3fdd67"
+                    "tag": "v1.3",
+                    "commit": "869d19d84c65d8879b4aeea8bb5d5a6b3a8fbd00"
                 },
                 "cargo-sources.json"
             ]
--- a/cargo-sources.json
+++ b/cargo-sources.json
@@ -1,0 +1,2452 @@
+[
+    {
+        "type": "archive",
+        "archive-type": "tar-gzip",
+        "url": "https://static.crates.io/crates/adler2/adler2-2.0.1.crate",
+        "sha256": "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa",
+        "dest": "cargo/vendor/adler2-2.0.1"
+    },
+    {
+        "type": "inline",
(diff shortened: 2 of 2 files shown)


page.codeberg.grinka.Transition.json
```diff
@@ -0,0 +1,53 @@
...
+                "install -Dm0644 data/icons/page.codeberg.grinka.Transition-symbolic.svg ${FLATPAK_DEST}/share/icons/hicolor/symbolic/apps/${FLATPAK_ID}-symbolic.svg",
+                "install -Dm0644 data/page.codeberg.grinka.Transition.desktop ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop",
+                "install -Dm0644 data/page.codeberg.grinka.Transition.service ${FLATPAK_DEST}/share/dbus-1/services/${FLATPAK_ID}.service",
+                "install -Dm0644 data/page.codeberg.grinka.Transition.metainfo.xml ${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.metainfo.xml"
+            ],
+            "sources" : [
+                {
+                    "type" : "git",
+                    "url" : "https://codeberg.org/grinka/transition.git",
+                    "tag": "v1.1"
```
bbhtt: you have to add commit to the tag as well.

page.codeberg.grinka.Transition.json
```diff
@@ -0,0 +1,53 @@
...
+                "install -Dm0644 data/icons/${FLATPAK_ID}.svg ${FLATPAK_DEST}/share/icons/hicolor/scalable/apps/${FLATPAK_ID}.svg",
+                "install -Dm0644 data/icons/page.codeberg.grinka.Transition-symbolic.svg ${FLATPAK_DEST}/share/icons/hicolor/symbolic/apps/${FLATPAK_ID}-symbolic.svg",
+                "install -Dm0644 data/page.codeberg.grinka.Transition.desktop ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop",
+                "install -Dm0644 data/page.codeberg.grinka.Transition.service ${FLATPAK_DEST}/share/dbus-1/services/${FLATPAK_ID}.service",
+                "install -Dm0644 data/page.codeberg.grinka.Transition.metainfo.xml ${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.metainfo.xml"
+            ],
+            "sources" : [
+                {
+                    "type" : "git",
+                    "url" : "https://codeberg.org/grinka/transition.git",
```
bbhtt, line 46: repo is missing any CI, format, lint build CI should be added at minimum. once you have fixed the issues create a new tag and update it here too.

page.codeberg.grinka.Transition.json
```diff
@@ -0,0 +1,53 @@
...
+        "org.freedesktop.Sdk.Extension.rust-stable"
+    ],
+    "finish-args" : [
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--device=dri",
+        "--socket=wayland",
+        "--filesystem=xdg-music:create",
+        "--filesystem=xdg-videos:create",
+        "--filesystem=xdg-download:create"
```
bbhtt: why are these needed? (and remove :create).

bbhtt (changes requested):

page.codeberg.grinka.Transition.json
OleksiiGrinka: Transition creates folder for converted files, I think convertsing something in Music, Videos and Downloads is a common usecase

page.codeberg.grinka.Transition.json
OleksiiGrinka: I think flathub docs don't mention CI, but I'll do it

page.codeberg.grinka.Transition.json
CodedOre: Has the use of the FileChooser portal been evaluated? What is the reason it can't be used?

OleksiiGrinka: Opening files using folder chooser works even without permissions but when using file chooser app wouldn't be able to create output folder without permissions

OleksiiGrinka:
bot, build

OleksiiGrinka:
bot, build

OleksiiGrinka:
Tested build locally, everything works as expected

OleksiiGrinka:
Okay it's not :p, adding files via file chooser don't work even in Music, Videos & Downloads, I guess :create is needed?

page.codeberg.grinka.Transition.json
CodedOre: Wouldn't it be possible to have the user select an output folder via the FileChooser portal to have a writeable location for the converted files?

OleksiiGrinka:
Okay it's not :create, I will try to fix in code

OleksiiGrinka:
Okay I tested again but on my PC, for some reason it works on PC but on my laptop it's not? I guess my xdg folders on laptop are broken

page.codeberg.grinka.Transition.json
bbhtt: hasn't been answered unresolved the thread.

OleksiiGrinka: > Wouldn't it be possible to have the user select an output folder via the FileChooser portal to have a writeable location for the converted files?

No, because input may have files from different folders and I find it inconvenient to output them into one big pile

OleksiiGrinka: > Wouldn't it be possible to have the user select an output folder via the FileChooser portal to have a writeable location for the converted files?

Want to expand a bit on why this IMO not the answer

1. Outputting all tracks into one big pile is bad (A lot of sorting work for users that just want mass convert the library)
2. I thought about using album tag for creating folders within output folder but that still wont preserve structure of input (e.g. CD 1, 2, 3...)  and also what to do with tracks that don't have an album tag? I see only trowing them into the pile as a solution which isn't good
3. I also thought about just throwing all files without accessible parent dir into the pile but I don't see how program would know if something is flatpak fake path or not (also there's gonna be a lot tracks from different albums in CD 1, 2, 3 etc)

@CodedOre

OleksiiGrinka: @bbhtt so... um... should I resolve conversation? It's been a week since I provided an answer why those permissions are needed and there no reply from codedore about this

CodedOre: Right, two questions:

- Where exactly are the converted files saved to? Considering you previously said "Transition creates folder for converted files". Is this a subfolder of the folder where the original files are in?
- How does the application deal if the user selects a file outside of the permitted folders, like Desktop, Documents, external drives?

OleksiiGrinka: > Right, two questions:
> 
>     * Where exactly are the converted files saved to? Considering you previously said "Transition creates folder for converted files". Is this a subfolder of the folder where the original files are in?
> 
>     * How does the application deal if the user selects a file outside of the permitted folders, like Desktop, Documents, external drives?

1. Files are saved to new directory which is created in a place where files are
2. It's gonna abort converting files with inaccessible directory, printing into logs that it doesn't have enough permissions to create directory

CodedOre: > 2. It's gonna abort converting files with inaccessible directory, printing into logs that it doesn't have enough permissions to create directory

So, from the perspective of the user it will just do nothing without explanation?

Most users will likely just expect the program to be broken, not look into the logs.

OleksiiGrinka: > > 2. It's gonna abort converting files with inaccessible directory, printing into logs that it doesn't have enough permissions to create directory
> 
> So, from the perspective of the user it will just do nothing without explanation?
> 
> Most users will likely just expect the program to be broken, not look into the logs.

I believe logs are pretty accessible (Program shows toast with log button on it, if window is not focused notification about finishing conversion will be sent), it's not some obscure thing that you can only see in terminal

OleksiiGrinka: Okay I thinked a bit and maybe I will try to implement output folder selection to see how it will work & look like, maybe the usecase I was thinking of doesn't work very well with making new directories in input directories...

OleksiiGrinka: Sooo I created 2.0 branch (not finished yet) with choose output folder change and it works,

but in Transition there is another way to output files called Replace Mode which just creates new file and deletes old one in place where original file is, and there is an issue in sandbox when you add files & convert them

On Fedora Linux replace mode produces fully working files but with weird prefix and suffix:

image

It happens when you try to move (rename) or copy file too.

~~On Arch Linux (clean gnome installation with archinstall) it seems like it can't create file at all? but app produces no errors~~*

So I'm really not sure what to do here :p

@CodedOre 

*: I forgot that weird prefix have dot in beginning a that I had show hidden files false :p, so arch does the same thing that fedora does

CodedOre: Well, you're not meant to rename a file path given by the document portal. 

If you selected only a file with the FileChooser portal, you only have permission for that one file path. That is expected behavior on the side of the portals.

OleksiiGrinka: > Well, you're not meant to rename a file path given by the document portal.
> 
> If you selected only a file with the FileChooser portal, you only have permission for that one file path. That is expected behavior on the side of the portals.

sooo, I can't do much here within sandbox?

CodedOre: Without a `filesystem` permission? No.

Allowing to change files other than the one selected over the FileChooser portal has been discussed before (see discussions about an Neighboring Files Portal), but there has been no portal implemented yet.

OleksiiGrinka: > Without a `filesystem` permission? No.
> 
> Allowing to change files other than the one selected over the FileChooser portal has been discussed before (see discussions about an Neighboring Files Portal), but there has been no portal implemented yet.

So should I leave above permissions in place (considering that I want to release app with choose output folder feature)? I believe that replace mode won't be used much on individually added files but I can see why would someone do that.

Also Neighboring Files Portal sounds interesting, I would look into this, thanks!

OleksiiGrinka: 2.0 got merged (output folder selection and other changes), so what to do with permissions? I was thinking about how can I fix replacing added files and only come up with warning dialog which gonna ask you for permission to parent folder of files, but I'm not sure if adding this is necessary because I think that replace mode gonna be used for big conversions (i.e. selecting folder), and before I was thinking that someone could use it to replace specific videos in videos folder but does it make much sense to replace videos in place where videos is with audio files? 

Anyways gonna do as you decide. v2.0 tag would be created after translation catch up and after this thing would be fixed

@CodedOre

OleksiiGrinka: > 2.0 got merged (output folder selection and other changes), so what to do with permissions? I was thinking about how can I fix replacing added files and only come up with warning dialog which gonna ask you for permission to parent folder of files, but I'm not sure if adding this is necessary because I think that replace mode gonna be used for big conversions (i.e. selecting folder), and before I was thinking that someone could use it to replace specific videos in videos folder but does it make much sense to replace videos in place where videos is with audio files?
> 
> Anyways gonna do as you decide. v2.0 tag would be created after translation catch up and after this thing would be fixed
> 
> @CodedOre

So I tried to implement warning dialog and it turns out that it's doesn't work as I thought it would (The idea was basically to show folder chooser before file chooser and hope that it'll allow program to see parent folder but it doesn't work that way and it's too clunky anyway...)

Anyways I'm back and ready to do stuff again (didn't even realise that 2 weeks passed already)

bbhtt: So you updated to 2.0, are these still needed?

OleksiiGrinka: > So you updated to 2.0, are these still needed?

yes, replace mode with added files need sandbox holes, sadly

bbhtt:
bot, build

PR #8892, opened 2026-06-05
Add app.privatepractice.meshapp

smikme:

- [x] MeshApp is a desktop client for Meshtastic mesh networks. It provides chat, node management, telemetry, map view, connection management, and Linux BLE/serial support.
- [x] Please attach a video showcasing the application on Linux using the Flatpak. https://disk.yandex.ru/i/iU9ZGaDp4pTe_Q.
- [x] The Flatpak ID follows all the rules listed in the Application ID requirements.
- [x] I have read and followed all the Submission requirements and the Submission guide and I agree to them.
  Additional maintainers: N/A


teamcons:
>     * [x]  Please attach a video showcasing the application on Linux using the Flatpak. N/A for now; local Flathub-style build was tested successfully and screenshots are provided in AppStream metadata.

**_Please attach a video showcasing the application on Linux using the Flatpak._**

app.privatepractice.meshapp.yml
```diff
@@ -0,0 +1,63 @@
+app-id: app.privatepractice.meshapp
```
teamcons, line 1: do you own https://privatepractice.app/ ?

app.privatepractice.meshapp.yml
```diff
@@ -0,0 +1,63 @@
...
+  - --filesystem=xdg-documents:create
+  - --filesystem=xdg-download:create
+  - --env=PATH=/app/jre/bin:/app/bin:/usr/bin
+  - --env=JAVA_HOME=/app/jre
+
+modules:
+  - name: openjdk
+    buildsystem: simple
+    build-commands:
+      - /usr/lib/sdk/openjdk25/install.sh
```
teamcons, line 24: is this required with the extension?

app.privatepractice.meshapp.yml
```diff
@@ -0,0 +1,63 @@
...
+
+  - name: meshapp
+    buildsystem: simple
+    build-options:
+      env:
+        GRADLE_USER_HOME: /run/build/meshapp/gradle-home
+        JAVA_HOME: /usr/lib/sdk/openjdk25/jvm/openjdk-25
+        PATH: /app/bin:/usr/bin:/usr/lib/sdk/openjdk25/bin
+    build-commands:
+      - |
```
teamcons, line 34: why not use a script?

flathub.json
```diff
@@ -0,0 +1,5 @@
+{
+  "only-arches": [
+    "x86_64"
```
teamcons: Why?

app.privatepractice.meshapp.yml
smikme: Yes, I am the owner of privatepractice.app. The application ID `app.privatepractice.meshapp` is based on that domain. The project repository is hosted at: https://git.privatepractice.app/covox/meshapp

app.privatepractice.meshapp.yml
smikme: The application was developed in Java and use jdk

app.privatepractice.meshapp.yml
smikme: I don’t fully understand the question.

flathub.json
smikme: I haven’t tested the application on other architectures. The BlueZ bridge was tested only on x86

teamcons: Then please remove this. Even if untested, leave the benefit of the doubt.

app.privatepractice.meshapp.yml
teamcons: you can use a script instead of writing a ton of lines of random commands in the manifest (which is.. not good)

though an actual buildsystem handling the various files would be better

app.privatepractice.meshapp.yml
```diff
@@ -0,0 +1,63 @@
...
+        set -eu
+        gradle-dist/bin/gradle --offline --no-daemon installDist \
```
hfiguiere:
```suggestion
      - gradle-dist/bin/gradle --offline --no-daemon installDist \
```

flathub.json
hfiguiere: what's the point of using Java then?

app.privatepractice.meshapp.yml
smikme: believe this optimization is excessive. The number of commands is relatively small, and their purpose is already clear.

flathub.json
smikme: Because the vast majority of the application is written in Java. It’s a large, feature-rich application, but some low-level or platform-specific functionality simply can’t be implemented in pure Java, so there is a small native component where necessary.

smikme: https://git.privatepractice.app/covox/meshapp

bbhtt:
Was the app written with the help of LLMs?

smikme:
Please attach a video showcasing the application on Linux using the Flatpak.
https://disk.yandex.ru/i/iU9ZGaDp4pTe_Q

smikme:
> Was the app written with the help of LLMs?

No, the LLM was only used for translation and localization.

bbhtt:
Haven't looked at the code yet but parts of the readme definitely also look LLM generated to me

smikme:
> Haven't looked at the code yet but parts of the readme definitely also look LLM generated to me
The translations aren’t perfect, but the commit history clearly shows that the application wasn’t developed using an LLM.

smikme:
> Haven't looked at the code yet but parts of the readme definitely also look LLM generated to me

    It’s an older, well-established application with a substantial development history.

smikme:
> > ```
> > * [x]  Please attach a video showcasing the application on Linux using the Flatpak. N/A for now; local Flathub-style build was tested successfully and screenshots are provided in AppStream metadata.
> > ```
> 
> **_Please attach a video showcasing the application on Linux using the Flatpak._**

https://disk.yandex.ru/i/iU9ZGaDp4pTe_Q

bbhtt:
You didn't answer it, I said "readme" you posted about translations again.

smikme:
> You didn't answer it, I said "readme" you posted about translations again.

I think there's a misunderstanding here.
The README was written by hand. An LLM was only used later to translate it, together with the rest of the documentation. The project itself, its architecture, implementation, and documentation structure were not generated by an LLM.
The translation quality is admittedly average. If you feel the README could be improved and are willing to help rewrite it, I'd be happy to review and merge contributions.

app.privatepractice.meshapp.yml
teamcons: Oh ok

app.privatepractice.meshapp.yml
teamcons: Nice

app.privatepractice.meshapp.yml
teamcons: Alright

app.privatepractice.meshapp.yml
smikme: I honestly don't understand the motivation behind disabling instruction enforcement.
Instructions exist to guide the model's behavior and ensure predictable results. Disabling their enforcement seems likely to create more problems than it solves. Could you elaborate on the use case you're trying to address?

teamcons: What model are you talking about? This is just cleaning out some nonsense

smikme: “execute build commands strictly; if a command fails or a variable is not set, fail the Flatpak build immediately.”

This is standard practice for Flatpak manifests, ensuring that the build does not continue after a hidden error.

teamcons: .. yes?

teamcons: You do not need those two lines.

smikme: - | is YAML syntax used to define a multi-line command block

smikme: If you insist on removing it, then the Gradle command should be updated accordingly:

gradle-dist/bin/gradle --offline --no-daemon installDist -PmeshappJavaVersion=25 -PmeshappJavaReleaseVersion=25 -PmeshappJavaFxPlatform=linux

CodedOre: > - | is YAML syntax used to define a multi-line command block

That's not the contentious point. 

The point is that `set -eu` is not necessary. 

A Flatpak manifest is not a simple bash script. If a command fails, the build will stop. There is no need to set that manually.

smikme: Ok. If you want to remove set -eu, that’s fine, but the command should either stay multi-line for readability, or the Gradle invocation should be rewritten as I suggested:

gradle-dist/bin/gradle --offline --no-daemon installDist -PmeshappJavaVersion=25 -PmeshappJavaReleaseVersion=25 -PmeshappJavaFxPlatform=linux

to single line

smikme: I’ve updated it and converted it into a single command block.

flathub.json
hfiguiere: still not resolved. yes you could try to build it for aarch64. They have bluetooth too.

smikme: 1. I currently have no plans to support BLE/Serial functionality (the native libraries) on Linux aarch64.

2. Additionally, I don't have the ability to build or test those native libraries for that architecture.

3. The JVM is architecture-specific. It must be built for either x86_64 or aarch64; there is no single universal JVM binary that works across both architectures.

hfiguiere: The JVM comes from https://github.com/flathub/flathub/pull/8892/files#diff-2c13b5f0f721981cb1a7abf90109b41537ec012a6388d08dcb5d64bdd59b47efR25 . I don't see what the problem is.

smikme: Do you understand how a Java application is packaged?
The application consists of a JVM (built for a specific architecture) bundled with the package, along with the application itself, which includes Java code and some native C/C++ components.
Both the JVM and the native libraries are architecture-specific. This is not like Apple’s Universal Binary approach, where a single package can contain binaries for multiple architectures.

This is a mature, stable application with 73 releases behind it. Please give Linux users a small gift and let them install it through Flatpak instead of having to do everything manually.

hfiguiere: If you ship your own JVM you are doing it wrong. So yes I do understand.

Now if you don't build these native components from source here then you are doing it wrong too.

smikme: All components are built by Gradle for a selected target architecture.

It is not possible to build a single universal package that works for both aarch64 and x86_64.

hfiguiere: > All components are built by Gradle for a selected target architecture.

so what's the problem?

 
> It is not possible to build a single universal package that works for both aarch64 and x86_64.

It seem you don't even understand the clearly documented concept behind flatpak. I don't see what is the universal package your are talking about.  There would be a flatpak for each architecture.

smikme: At the moment, my codebase only supports aarch64 on macOS. Linux aarch64 is not supported.

In addition, none of the roughly 5,000 users of the application have ever requested Linux aarch64 support.

I have not seen any requirement that applications must support this architecture. If Linux aarch64 support is expected, then where is the line drawn? Why not require support for RISC-V, MIPS, or other architectures as well?

I'm genuinely struggling to understand what is being asked of me. Are you requiring me to add and maintain support for an architecture that the application does not currently support and for which I have neither build infrastructure nor user demand?

hfiguiere: > Why not require support for RISC-V, MIPS, or other architectures as well? 

Because we don't support other architectures. Only the two most common 64 bits architectures: x86_64 and aarch64.

smikme: More importantly, even if I were willing to add Linux aarch64 support, users would still be left without a properly functioning UI.

The reality is that JavaFX support on Linux aarch64 is currently not mature enough for production use. Building the application for that architecture is one thing; providing a stable and usable graphical application is another. At this point, I cannot reasonably claim to support Linux aarch64 when the underlying JavaFX stack itself remains problematic on that platform.

smikme: When Linux aarch64 support becomes practical and the underlying JavaFX stack is mature enough, I will be happy to add support for it. Alternatively, another distributor or maintainer is free to provide and maintain such a build.

For now, I prefer not to advertise support for an architecture that I cannot properly build, test, or support.

smikme: There are a lot of practical issues involved that cannot be solved by simply enabling another build target.

Architecture support is not a checkbox. It requires working dependencies, reliable tooling, testing infrastructure, and ongoing maintenance.

smikme: This issue will not be solved by simply enabling the aarch64 build.
The application will not function properly on Linux aarch64 with the current dependency stack. So what should we do here?

smikme: meshapp-architecture

bbhtt:
You mean the English readme is the translated version? It didn't just translate it then.

flathub.json
smikme: I have added aarch64, and it will probably even build successfully, but it will not actually work properly.
So this would only produce a misleading package: technically built, but not really supported or usable.

smikme:
> You mean the English readme is the translated version? It didn't just translate it then.

If you want the text to read naturally, you need either a native speaker or someone willing to edit the machine translation.
The English README is a translated version. The goal was to make the documentation accessible, not to produce native-level technical writing.
https://git.privatepractice.app/covox/meshapp/src/branch/main/src/main/resources/i18n/messages_en.properties
The translation files are probably not masterpieces either, but they're good enough to be useful.

flathub.json
bbhtt, line 1: drop this file

app.privatepractice.meshapp.metainfo.xml
bbhtt, line 1: this should be in upstream repo

app.privatepractice.meshapp.metainfo.xml
```diff
@@ -0,0 +1,56 @@
...
+    <name>Konstantin A. Smirnov</name>
+  </developer>
+  <launchable type="desktop-id">app.privatepractice.meshapp.desktop</launchable>
+  <description>
+    <p>MeshApp is a desktop client for Meshtastic devices and mesh networks.</p>
+    <p>It supports TCP, serial or USB and Bluetooth Low Energy connections, chat, node monitoring, telemetry charts, packet debugging and device configuration.</p>
+  </description>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://git.privatepractice.app/covox/meshapp/raw/branch/main/docs/screenshots/chat-b.jpg</image>
```
bbhtt: links should be pinned to commit, same below

app.privatepractice.meshapp.yml
```diff
@@ -0,0 +1,80 @@
...
+
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=x11
+  - --device=all
+  - --allow=bluetooth
+  - --system-talk-name=org.bluez
+  - --filesystem=xdg-documents:create
+  - --filesystem=xdg-download:create
```
bbhtt: why is this needed?

app.privatepractice.meshapp.yml
```diff
@@ -0,0 +1,80 @@
...
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.openjdk25
+command: MeshApp
+
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=x11
+  - --device=all
```
bbhtt, line 13: why is device=all needed?

app.privatepractice.meshapp.yml
```diff
@@ -0,0 +1,80 @@
...
+  - name: meshapp
+    buildsystem: simple
+    build-options:
+      env:
+        GRADLE_USER_HOME: /run/build/meshapp/gradle-home
+        JAVA_HOME: /usr/lib/sdk/openjdk25/jvm/openjdk-25
+        PATH: /app/bin:/usr/bin:/usr/lib/sdk/openjdk25/bin
+    build-commands:
+      - |
+        case "$(uname -m)" in
```
bbhtt: FLATPAK_ARCH and the cases are x86_64 / aarch64

flathub.json
smikme: Which file are you talking about?

app.privatepractice.meshapp.metainfo.xml
smikme: It’s just a static file.

app.privatepractice.meshapp.yml
smikme: Data import/export and file upload functionality.

app.privatepractice.meshapp.yml
smikme: It is not possible to know in advance which device will be used for the connection.
The application interacts with hardware, and the connection method depends on the specific device available to the user.

flathub.json
bbhtt: the file I commented on flathub.json.

app.privatepractice.meshapp.metainfo.xml
bbhtt: yes still.

app.privatepractice.meshapp.yml
bbhtt: dows this go through filechoosers? you can't use the portal?

app.privatepractice.meshapp.yml
smikme: Ok. Not a problem. Added

app.privatepractice.meshapp.yml
smikme: Yes, through a file chooser. Unfortunately, using the portal is not practical in this scenario.

app.privatepractice.meshapp.metainfo.xml
smikme: The version is set during the build.
I cannot reference a file from the future, because I have to specify the exact commit it belongs to.

app.privatepractice.meshapp.yml
smikme: For example 
image

app.privatepractice.meshapp.metainfo.xml
bbhtt: No idea what you mean.

The normal process is add new version tags in metainfo + do release tasks + git commit then git tag then release

You have to follow this sequence.

app.privatepractice.meshapp.yml
bbhtt: Why is it not practical?

app.privatepractice.meshapp.metainfo.xml
bbhtt: FYI this is a blocker, the requirements are clear about where it is supposed to be.

app.privatepractice.meshapp.yml
```diff
@@ -0,0 +1,80 @@
...
+    buildsystem: simple
+    build-options:
+      env:
+        GRADLE_USER_HOME: /run/build/meshapp/gradle-home
+        JAVA_HOME: /usr/lib/sdk/openjdk25/jvm/openjdk-25
+        PATH: /app/bin:/usr/bin:/usr/lib/sdk/openjdk25/bin
+    build-commands:
+      - |
+        case "$FLATPAK_ARCH" in
+          x86_64|amd64)
```
bbhtt: x86_64 only that's what I said above where is amd64 coming from?

app.privatepractice.meshapp.yml
```diff
@@ -0,0 +1,80 @@
...
+            meshapp_javafx_platform=linux
+            meshapp_protoc_classifier=linux-x86_64
+            ;;
+          aarch64|arm64)
```
bbhtt: same as above

app.privatepractice.meshapp.yml
```diff
@@ -0,0 +1,80 @@
...
+            meshapp_javafx_platform=linux-aarch64
+            meshapp_protoc_classifier=linux-aarch_64
+            ;;
+          *)
+            echo "Unsupported Flatpak build architecture: $FLATPAK_ARCH" >&2
```
bbhtt: drop this these are the only two arches it is getting built for

flathub.json
bbhtt: Not done why did you mark it as resolved?

smikme: I really don’t understand your point here.

app.privatepractice.meshapp.metainfo.xml
smikme: Okay, then I think we’re done here.

I don’t have the time or the desire to continue arguing over minor details. I’m not particularly interested in publishing the application on Flathub myself; this submission was made in response to requests from users.

I’ve done my best to explain the technical constraints, limitations, and maintenance implications. If those explanations are not sufficient, I’d rather withdraw the submission than continue debating the same points.

Thank you for your time.

app.privatepractice.meshapp.yml
smikme: Because the application interacts with external hardware and allows users to import/export arbitrary files.

The access targets are not known ahead of time, so the application uses a standard file chooser to let the user explicitly select files and locations. Reworking these workflows around portals would add considerable complexity while providing little benefit for this particular use case.

flathub.json
bbhtt: Just remove the `flathub.json` file, I am having trouble understanding the problem here.

app.privatepractice.meshapp.yml
smikme: changed

app.privatepractice.meshapp.metainfo.xml
smikme: This workflow is not compatible with the project’s CI/CD pipeline, so it cannot be automated as part of the build process.

For that reason, I added the file manually.

flathub.json
smikme: ok

smikme:
Hi! I was hoping that at least the new 2.1.20 release would be available on Flathub by now. Is there any chance of that happening?

app.privatepractice.meshapp.yml
bbhtt: > The access targets are not known ahead of time, so the application uses a standard file chooser to let the user explicitly select files and locations. 

Portaled filechoosers also lets you view and select arbitrary files. There is no requirement to know them ahead of time. So I am not sure what is the problem here.

smikme: The problem is that the application isn’t adapted for this approach and doesn’t use calls that the portal can intercept.
There are no plans to adapt it (for now), since the application isn’t developed specifically for Linux. For it, Linux is just one of the supported platforms.

bbhtt: > When a suitable XDG portal exists, is supported across the ecosystem, and covers the application's use case, or can be implemented in the application itself — using the portal becomes **mandatory** instead of static permissions.

https://docs.flathub.org/docs/for-app-authors/requirements#permissions

smikme: This project is far too large to impose blockers like this. I have no idea how to implement it with the resources currently available and within a reasonable timeframe. It’s not an application that’s still being actively developed from scratch — it’s already a mature application that people are using today.

smikme: It’s hard to imagine users being asked to grant access to the application’s configuration every single time they start the app.

smikme: The --filesystem=xdg-download:create permission can indeed be removed. It’s not required.

smikme: I’ll replace it with --filesystem=home.

bbhtt: then it went further away from a resolution, it will not be accepted until you use the portals or use a proper reason for not using them.

it's cross platform, it's old, mature etc. are not acceptable. there are plenty of mature and cross platform applications using portals.

smikme: The application creates its database and configuration files under ~/.meshapp/, and it needs read/write/create access to that directory. This is a requirement imposed by the application itself.

bbhtt: That's another thing to fix. You should use XDG compliant location instead of dumping things in home.

smikme: This is not something that can be fixed. This is required for compatibility reasons.

smikme: Dear reviewer, the majority of the application’s users are on platforms that do not support XDG portals. Approximately 80% use Windows (which has no XDG portals), 17% use macOS (which also has no XDG portals), and only about 3% are Linux users who may choose to install the application from Flathub.

Given this distribution, the application was not designed around XDG portals, and supporting them would require significant platform-specific development effort for a very small subset of users.

bbhtt: This excuses will not work as I said in the last comment. Also until this is actually fixed you can stop requesting reviews there is nothing new to review

smikme:
I’m quite surprised and disappointed by this outcome. However, it seems I have no choice but to maintain my own Flatpak repository.

PR #8925, opened 2026-06-09
Add io.github.j0ck4.Wren

j0ck4:
This adds **Wren**, a native GTK4 / libadwaita client for managing WireGuard VPN tunnels.

- Source: https://github.com/j0ck4/wren-project (GPL-3.0-or-later)
- Built offline from vendored Cargo sources (`cargo-sources.json`); no network access at build time.

### Permissions justification

Wren is a GUI front-end for the host's `wg-quick`. Bringing a WireGuard tunnel up or down needs root and manipulates the host network namespace, so it cannot run inside the sandbox and there is no portal for it. Hence:

- `--talk-name=org.freedesktop.Flatpak` — runs `wg-quick up/down` on the host via `flatpak-spawn --host`. This is the core requirement; a WireGuard manager cannot work without host access.
- `--system-talk-name=org.freedesktop.PolicyKit1` — `wg-quick` is invoked through `pkexec`, so the user authenticates via polkit instead of the app holding privileges.
- `--talk-name=org.kde.StatusNotifierWatcher` — optional system-tray icon.
- `--talk-name=org.freedesktop.Notifications` — connect/disconnect notifications.
- `--filesystem=xdg-config/autostart:create` — optional “start at login” entry.
- Display sockets only; **no `--share=network`** — the app itself does no networking, all traffic is handled by the host's `wg-quick`.

I understand this grants host access and that the app will be flagged accordingly; it is inherent to a WireGuard tunnel manager. Happy to adjust per review.

### DIFF
--- a/cargo-sources.json
+++ b/cargo-sources.json
@@ -1,0 +1,2088 @@
+[
+ {
+  "type": "archive",
+  "archive-type": "tar-gzip",
+  "url": "https://static.crates.io/crates/aho-corasick/aho-corasick-1.1.4.crate",
+  "sha256": "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301",
+  "dest": "cargo/vendor/aho-corasick-1.1.4"
+ },
+ {
+  "type": "inline",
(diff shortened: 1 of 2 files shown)


j0ck4:
bot, build

j0ck4:
bot, build

bbhtt:
You can open slop PRs elsewhere.

PR #8926, opened 2026-06-09
Add io.github.trystan_sa.rproc

Trystan-SA:
[rproc](https://github.com/Trystan-SA/rproc) is a resource and process monitor for Linux (Slint, software renderer): live CPU / memory / disk / network / GPU charts, the full process tree (sort, filter, kill), systemd services, and startup entries.

- **Manifest** pins the upstream `v0.3.6` tag (with commit).
- **`cargo-sources.json`** generated from the pinned `Cargo.lock` with `flatpak-cargo-generator.py`.
- `flatpak-builder-lint appstream` passes; the manifest builds (also built green in upstream CI).

### Requested permissions / linter exceptions

The manifest lint flags three permissions, all intrinsic to a system monitor :

- `--filesystem=host:ro`, read-only host access so the monitor can report system-wide disk/mount usage and read host metadata (e.g. `/etc/os-release`, `/proc/mounts`). Strictly read-only.
- `--talk-name=org.freedesktop.systemd1` and `--system-talk-name=org.freedesktop.systemd1`, the Services and Startup tabs enumerate and toggle systemd user and system units over D-Bus.

### DIFF
--- a/cargo-sources.json
+++ b/cargo-sources.json
@@ -1,0 +1,7574 @@
+[
+    {
+        "type": "archive",
+        "archive-type": "tar-gzip",
+        "url": "https://static.crates.io/crates/ab_glyph/ab_glyph-0.2.32.crate",
+        "sha256": "01c0457472c38ea5bd1c3b5ada5e368271cb550be7a4ca4a0b4634e9913f6cc2",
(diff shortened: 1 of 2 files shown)


hfiguiere:

See also our AI policy

PR #8928, opened 2026-06-09
Add org.amule.aMule

got3nks:
## About aMule

aMule is the all-platform P2P file-sharing client based on the eDonkey2000 + Kad protocols, actively maintained since 2003 at https://github.com/amule-org/amule by @mrjimenez and contributors. This submission packages aMule 3.0.0 pinned to a master-tip commit (`843bd120`) so the bundle includes post-release fixes — notably the libbfd opt-out, the locale catalog lookup-prefix fix, and WebSocket hardening landed since the tag.

## App-id

The manifest uses `org.amule.aMule`, which has been aMule's AppStream `<id>` since the project first shipped a metainfo file. The original project lived at amule.org (2003–2015); the current `amule-org` GitHub organization is the active maintainer continuation. We don't currently control amule.org DNS, though we're reaching out to former maintainers about reclaiming it.

Requesting acceptance under the established app-id under Flathub's continuity guidelines. Happy to provide additional attestation, or to rename to `io.github.amule_org.amule` (reverse-DNS of `amule-org.github.io`, which we do control) if the historical claim isn't sufficient.

## `--filesystem=home` rationale

aMule is a P2P client where users mark arbitrary folders as shared (uploads) and pick arbitrary destinations for Incoming / Temp / completed-files directories (downloads). Restricting to `xdg-download` alone silently breaks the most common configuration — "share my Music folder, save downloads to ~/Movies". This matches the precedent set by qBittorrent and Transmission on Flathub for the same architectural reason.

## Build & test evidence

Built locally with `flatpak-builder --user --force-clean --arch=aarch64 --ccache` against GNOME 50 Platform+SDK on Ubuntu 24.04 ARM64. Bundle installs, launches, and runs translated UI strings correctly. The manifest's two non-default choices (`-DENABLE_BFD=NO`, `separate-locales: false`) are explained inline where they appear.

`flatpak-builder-lint` on the manifest is clean except `finish-args-home-filesystem-access` (per the rationale above). Repo lint adds `appstream-screenshots-not-mirrored-in-ostree` / `appstream-external-screenshot-url` — the screenshot URLs return HTTP 200, and Flathub's build infrastructure mirrors them automatically on first publish.

## Maintenance

The `x-checker-data` block on the amule source drives automated update PRs against this listing on every new aMule release tag. Maintenance permissions on the resulting `flathub/org.amule.aMule` repo will be extended to other amule-org maintainers post-merge.

Happy to make any changes the reviewer suggests, including the app-id rename if needed.

### DIFF
--- a/org.amule.aMule.yaml
+++ b/org.amule.aMule.yaml
@@ -1,0 +1,268 @@
+# Flathub-strict aMule manifest.
+#
+# This file is the staging copy of what gets submitted to
+# github.com/flathub/flathub once the listing is approved, the dedicated
+# github.com/flathub/org.amule.aMule repo becomes the canonical home and
+# this file is kept in sync with it.
+#
+# Differences from packaging/linux/flatpak/org.amule.aMule.yaml.in:
+#  - Concrete version pins (no ${...} template variables — Flathub CI
+#    doesn't run envsubst).
+#  - amule source: tag + commit instead of branch (Flathub forbids
+#    mutable refs).
+#  - cryptopp source: commit added next to the tag.
+#  - All other modules identical (they were already archive + sha256
+#    pinned).
+#
+# To refresh after a new aMule release tag:
+#   1. Resolve the tag's commit SHA:
+#        gh api repos/amule-org/amule/git/refs/tags/<tag> --jq '.object.sha'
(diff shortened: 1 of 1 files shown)


hfiguiere:

See also our AI policy

PR #8935, opened 2026-06-10
Add ru.linux_gaming.PortProtonQt

Boria138:

- [x] Please describe the application briefly. (Software for playing Microsoft Windows games and launchers, replace for ru,linux_gaming,PortProton)
- [x] Please attach a video showcasing the application on Linux using the Flatpak.   https://github.com/user-attachments/assets/16d0c5a9-cfa2-4be3-9325-c5851c4926ea
- [x] I am an  author


ru.linux_gaming.PortProtonQt.yaml
```diff
@@ -0,0 +1,330 @@
...
+    LLVM_INSTALL_DIR: /usr/lib/sdk/llvm21
+
+cleanup-commands:
+  - /app/pyside_cleanup.sh
+
+finish-args:
+  - --share=ipc
+  - --socket=wayland
+  # --fallback-x11 forces Wine to run in native Wayland mode it is still too unstable for this behavior
+  - --socket=x11
```
CodedOre, line 24: The Wayland socket should always be used with the fallback socket. 

If Wine's Wayland mode is really too unstable, only use the X11 socket for now.

ru.linux_gaming.PortProtonQt.yaml
```diff
@@ -0,0 +1,330 @@
...
+  # For Wine crash handling
+  - --allow=devel
+  # For wine mono and gecko work
+  - --allow=per-app-dev-shm
+  # Wine uses UDisks2 to enumerate disk drives
+  - --system-talk-name=org.freedesktop.UDisks2
+  # HDR10 support through gamescope
+  - --filesystem=xdg-run/gamescope-0:ro
+  - --filesystem=xdg-desktop:create
+  - --filesystem=xdg-data/applications:create
```
CodedOre: If access to the desktop and `xdg-data/applications` is to install desktop files, then you should use the dynamic launcher portal instead. 

Also, why does your application require an desktop folder to exists?

And, as an sidenote: These folders are already accessible over the `home` permissions. Doesn't change you should use the portal.

ru.linux_gaming.PortProtonQt.yaml
```diff
@@ -0,0 +1,330 @@
...
+  - --filesystem=home
+  - --filesystem=/mnt
+  - --filesystem=/media
+  - --filesystem=/run/media
```
CodedOre, line 40: Why access to all files on home and external drives?

ru.linux_gaming.PortProtonQt.yaml
```diff
@@ -0,0 +1,330 @@
...
+  - --device=all
+  # Mangohud, OBSVkCapture and GameScope
+  - --env=PATH=/app/bin:/usr/bin:/app/utils/bin:/usr/lib/extensions/vulkan/MangoHud/bin/:/usr/lib/extensions/vulkan/gamescope/bin:/usr/lib/extensions/vulkan/OBSVkCapture/bin
+  # For Steam Deck HDR support
+  - --env=LD_LIBRARY_PATH=/usr/lib/extensions/vulkan/gamescope/lib
+  - --env=GST_PLUGIN_SYSTEM_PATH=/app/lib/gstreamer-1.0:/app/lib32/gstreamer-1.0:/usr/lib/extensions/gstreamer-1.0:/usr/lib/x86_64-linux-gnu/codecs-extra/lib/gstreamer-1.0:/usr/lib/i386-linux-gnu/codecs-extra/lib/gstreamer-1.0:/usr/lib/x86_64-linux-gnu/gstreamer-1.0:/usr/lib/i386-linux-gnu/gstreamer-1.0
+  # These are for notifications/tray icons
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=org.freedesktop.Notifications
```
CodedOre: For notifications, you should always use the portal instead.

ru.linux_gaming.PortProtonQt.yaml
```diff
@@ -0,0 +1,330 @@
...
+  # Inhibit ScreenSaver on Game Active
+  - --talk-name=org.freedesktop.ScreenSaver
```
CodedOre: This should use the inhibit portal instead.

ru.linux_gaming.PortProtonQt.yaml
```diff
@@ -0,0 +1,330 @@
...
+
+  org.freedesktop.Platform.Compat.i386.Debug:
+    directory: lib/debug/lib/i386-linux-gnu
+    version: '25.08'
+    no-autodownload: true
+
+  org.freedesktop.Platform.codecs_extra.i386:
+    directory: lib/i386-linux-gnu/codecs-extra
+    version: 25.08-extra
+    autodelete: false
```
CodedOre, line 86: Why do you set `autodelete` to `false` here?

ru.linux_gaming.PortProtonQt.yaml
```diff
@@ -0,0 +1,330 @@
...
+    add-ld-path: lib
+
+modules:
+  - ru.linux_gaming.PortProtonQt.pypi-deps.yaml
+
+  - name: pyside-setup
+    buildsystem: simple
```
CodedOre, line 93: It probably makes more sense to use the Pyside base app instead of the Wine one. 

Would the "system" Wine even be used by your launcher?

ru.linux_gaming.PortProtonQt.yaml
```diff
@@ -0,0 +1,330 @@
...
+    modules:
+      - name: perl
+        buildsystem: simple
+        build-commands:
+          - ./Configure -des -Dprefix=$FLATPAK_DEST -Dman1dir=none -Dman3dir=none
+          - make
+          - make install
+        post-install:
+          # Fix wrong permissions
+          - chmod -R u+w $FLATPAK_DEST/lib/perl5
```
CodedOre, line 159: Why do you say a prefix being read-only on the final system having no write permissions is "wrong"?

ru.linux_gaming.PortProtonQt.yaml
```diff
@@ -0,0 +1,330 @@
...
+          - /bin/prove
+          - /bin/ptar
+          - /bin/ptardiff
+          - /bin/ptargrep
+          - /bin/shasum
+          - /bin/splain
+          - /bin/streamzip
+          - /bin/xsubpp
+          - /bin/zipdetails
+          - '*.pod'
```
CodedOre, line 191: Is Perl used by your application at runtime?

If not, you should remove it fully on cleanup:


```suggestion
        cleanup:
          - "*"
```

ru.linux_gaming.PortProtonQt.yaml
```diff
@@ -0,0 +1,330 @@
...
+      - type: inline
+        dest-filename: ld.so.conf
+        contents: |
+          /app/lib32
+          /app/lib/i386-linux-gnu
+      - type: file
+        path: pyside_cleanup.sh
+
+  - name: PortProtonQt
+    buildir: true
```
CodedOre:
```suggestion
```

ru.linux_gaming.PortProtonQt.yaml
Boria138: In that case, the application itself will run in Xwayland, which is also undesirable, since it is Wine that is unstable, not PortProtonQt; it can be run in native Wayland, and there is a setting in the launcher to enable native Wayland—anyone who wants to can enable it. I don’t use the fallback because it forces Wayland, and there’s no way to control it; in the context of WiNE launchers, two sockets are a much more stable solution

ru.linux_gaming.PortProtonQt.yaml
Boria138: > Would the "system" Wine even be used by your launcher?

Yes

barthalion:
What's the difference between this and https://flathub.org/en/apps/ru.linux_gaming.PortProton specifically?

Boria138:
> What's the difference between this and https://flathub.org/en/apps/ru.linux_gaming.PortProton specifically?

The ability to run natively on Wayland; the previous project used yad as its GUI, so running natively on Wayland wasn't possible. No more automatic updates—the very thing I got flak for last time—better integration with portals, and much more. Essentially, the previous project is already EOL.

ru.linux_gaming.PortProtonQt.yaml
Boria138: > Is Perl used by your application at runtime?

This is a dependency for exiftool; it is used, for example, to automatically create shortcuts with GOG or EGS, so perl cannot be removed

barthalion:
We've got a strict anti-LLM policy described in the submission requirements and yet: https://git.linux-gaming.ru/Linux-Gaming/PortProtonQt/src/branch/main/AGENTS.md

Are you using LLM to develop this?

Boria138:
> We've got a strict anti-LLM policy described in the submission requirements and yet: https://git.linux-gaming.ru/Linux-Gaming/PortProtonQt/src/branch/main/AGENTS.md
> 
> Are you using LLM to develop this?

I read the guidelines on neural networks, and they specify that exceptions can be made if a project is sufficiently mature. The scripts inside PortProton have been around for over ten years and don’t contain a single line of neural network code. The GUI itself is a bit more complicated; a neural network was used there, but with AGENTS.md and automated tests to rule out false positives (and human verification was also performed). Similarly, this isn’t the first version of PPQT; everything was developed and refined over the course of a year.

Boria138:
I wrote the manifest myself, based on the old PortProton manifest as well as other examples; exiftool, for instance, was taken from a different application, as noted in the manifest

ru.linux_gaming.PortProtonQt.yaml
Boria138: I copy it from WINE

Boria138:
bot, build

ru.linux_gaming.PortProtonQt.yaml
```diff
@@ -0,0 +1,326 @@
...
+    ORIGLIBDIRS: /lib:/lib64:/lib/x86_64-linux-gnu
+    LLVM_INSTALL_DIR: /usr/lib/sdk/llvm21
+
+cleanup-commands:
+  - /app/pyside_cleanup.sh
+
+finish-args:
+  - --share=ipc
+  - --socket=wayland
+  # --fallback-x11 forces Wine to run in native Wayland mode it is still too unstable for this behavior
```
hfiguiere, line 23: then remove the wayland socket.

ru.linux_gaming.PortProtonQt.yaml
Boria138: Снимок экрана_20260610_174953
There are builds downloaded separately, Proton versions found on Steam, and the system-level WINE; different versions of WINE or Proton can be useful for different tasks.

hfiguiere:
> > We've got a strict anti-LLM policy described in the submission requirements and yet: https://git.linux-gaming.ru/Linux-Gaming/PortProtonQt/src/branch/main/AGENTS.md
> > Are you using LLM to develop this?
> 
> I read the guidelines on neural networks, and they specify that exceptions can be made if a project is sufficiently mature. The scripts inside PortProton have been around for over ten years and don’t contain a single line of neural network code. The GUI itself is a bit more complicated; a neural network was used there, but with AGENTS.md and automated tests to rule out false positives (and human verification was also performed). Similarly, this isn’t the first version of PPQT; everything was developed and refined over the course of a year.

The guidelines are not on "neural networks" but on generative AI (aka chatbots). You have everything setup to redirect the major planet-burning-chatbots to instructions that don't tell them to get away.

As for "mature" you don't inherit the mature status from forking a project that is no longer maintained.

barthalion:
If this is meant to be a continuation of PortProton, I suggest using the beta branch and eventually replacing the old version with this. It doesn't have to be submitted separately, and appending toolkit-dependent suffix isn't something users are going to care about.

Boria138:
> If this is meant to be a continuation of PortProton,

That’s right—this is a continuation of PortProton, with a switch to PySide6 to fix all the fundamental issues that have lingered due to the less-than-ideal combination of yad and bash. Just a moment ago, someone accused me of not being able to call this a mature project because of the fork, but this isn’t a fork—it’s a continuation, same code, same developers, but a new, more suitable stack. The plan is for PortProtonQt to completely replace PortProton. We can’t do that right now because the user bases are incomparable, and it’s simply wrong to change something people are used to without asking. I think the beta branch is the better solution in this regard. Thanks for the suggestion.

Boria138:
I submitted a PR to the beta branch

https://github.com/flathub/ru.linux_gaming.PortProton/pull/158

Is that okay?

PR #8966, opened 2026-06-12
Add com.wcpos.main

kilbot:
## WCPOS — Point of Sale for WooCommerce

WCPOS is a desktop point-of-sale client for [WooCommerce](https://woocommerce.com) stores: manage orders, products and customers, and take payments in person, with offline support and receipt/USB printer integration.

- Homepage: https://wcpos.com
- Source / issues: https://github.com/wcpos/electron
- Developer: Paul Kilmurray (wcpos.com / github.com/wcpos)

## Why extra-data

WCPOS cannot be built from source on Flathub's buildbot: it bundles the proprietary, license-key-gated `rxdb-premium` package, and the renderer bundle is produced from a private monorepo. The manifest therefore uses the `extra-data` pattern: it downloads the prebuilt `.deb` we publish to GitHub releases and unpacks it on the user's machine at install time (same pattern as other proprietary-component apps, e.g. `com.google.Chrome`). `x-checker-data` is wired so flatpak-external-data-checker can open version-bump PRs automatically per release.

## Permissions notes

- `--device=all`: required for raw USB receipt printers and cash drawers — the app drives ESC/POS USB devices directly via libusb (the `usb` Node native module). The device set is arbitrary retail printer hardware, so the USB portal's fixed-device model doesn't fit yet; happy to revisit when portal-based USB enumeration covers this case.
- `--share=network`: network ESC/POS printers and store sync.

## Validation

Validated in our repo's CI ([workflow run](https://github.com/wcpos/electron/actions/runs/27437862900), [workflow source](https://github.com/wcpos/electron/blob/main/.github/workflows/flathub-validate.yml)):

- `flatpak-builder-lint manifest` and `appstream` pass (one note below)
- Full `org.flatpak.Builder` build, exported to a local OSTree repo, installed from the host — exercising the `extra-data` download and `apply_extra.sh` end-to-end, asserting the unpacked Electron binary lands in `/app/extra`
- `StartupWMClass=wcpos` verified against the running app's visible window (xprop under Xvfb)

## Known linter note

`appid-url-not-reachable`: wcpos.com currently returns 403 to `python-requests` user agents (WAF bot rule) while serving browsers and curl normally. We are relaxing that rule on the site; if it lingers, we'd appreciate guidance on the exceptions process. The domain is ours (it serves the app's homepage, docs and the WordPress plugin) and we intend to verify the app via the website method after acceptance.


hfiguiere:

See also our AI policy

PR #8976, opened 2026-06-14
Add it.belloworld.mercurygram

drizzt:

- [X] Please describe the application briefly. Mercurygram Desktop, an unofficial,
privacy and security focused fork of Telegram Desktop. It is the desktop
counterpart of the Mercurygram Android client, built by rebasing Mercurygram
patches on top of upstream Telegram Desktop.
- [X] Please attach a video showcasing the application on Linux using the Flatpak. <https://github.com/user-attachments/assets/9b0ea8cb-ef0d-4192-9da6-17d582eaa0ef>

### DIFF
--- a/it.belloworld.mercurygram.yml
+++ b/it.belloworld.mercurygram.yml
@@ -1,0 +1,397 @@
+id: it.belloworld.mercurygram
+runtime: org.gnome.Platform
+runtime-version: '50'
+sdk: org.gnome.Sdk
+command: Mercurygram
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --device=all
+  - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.gnome.Mutter.IdleMonitor
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=com.canonical.AppMenu.Registrar
+  - --talk-name=com.canonical.indicator.application
+  - --talk-name=org.ayatana.indicator.application
+  - --talk-name=org.sigxcpu.Feedback
+  - --unset-env=QT_PLUGIN_PATH
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.llvm22
+build-options:
+  append-path: /usr/lib/sdk/llvm22/bin
+  append-ld-library-path: /usr/lib/sdk/llvm22/lib
+  cflags: -g1
+  cxxflags: -g1
+  env:
+    CMAKE_BUILD_TYPE: None
+    CMAKE_PREFIX_PATH: /app
+modules:
+  - name: pkg-config
+    buildsystem: simple
+    build-commands:
+      - install -d $FLATPAK_DEST/bin
+      - |
+        cat <<EOF > $FLATPAK_DEST/bin/pkg-config
+        #!/bin/sh
+        unset PKG_CONFIG_ALLOW_SYSTEM_LIBS
+        exec /usr/bin/pkg-config "\$@"
+        EOF
+      - chmod +x $FLATPAK_DEST/bin/pkg-config
+    cleanup:
+      - '*'
+
+  - name: implib
+    buildsystem: simple
+    build-commands:
+      - |
+        install -d $FLATPAK_DEST/lib
+        mkdir build
+        cd build
+        implib() {
+          LIBFILE=$(basename $1)
+          LIBNAME=$(basename $1 .so)
+          ../implib-gen.py -q $1
+          gcc $CFLAGS -c -o $LIBFILE.tramp.o $LIBFILE.tramp.S
+          gcc $CFLAGS -c -o $LIBFILE.init.o $LIBFILE.init.c
+          ar rcs $FLATPAK_DEST/lib/$LIBNAME.a $LIBFILE.tramp.o $LIBFILE.init.o
+        }
+        implib /usr/lib/$FLATPAK_ARCH-linux-gnu/libgtk-3.so
+        implib /usr/lib/$FLATPAK_ARCH-linux-gnu/libgdk-3.so
+    sources:
+      - type: git
+        url: https://github.com/yugr/Implib.so.git
+        commit: ecf7bb51a92a0fb16834c5b698570ab25f9f1d21
+    cleanup:
+      - '*'
+
+  - name: jinja2
+    buildsystem: simple
+    build-commands:
+      - pip3 install --exists-action=i --no-index --find-links="file://${PWD}" --prefix=${FLATPAK_DEST}
+        flit-core --no-build-isolation
+      - pip3 install --exists-action=i --no-index --find-links="file://${PWD}" --prefix=${FLATPAK_DEST}
+        "Jinja2" --no-build-isolation
+    sources:
+      - type: file
+        url: https://files.pythonhosted.org/packages/69/59/b6fc2188dfc7ea4f936cd12b49d707f66a1cb7a1d2c16172963534db741b/flit_core-3.12.0.tar.gz
+        sha256: 18f63100d6f94385c6ed57a72073443e1a71a4acb4339491615d0f16d6ff01b2
+        x-checker-data:
+          type: pypi
+          name: flit-core
+      - type: file
+        url: https://files.pythonhosted.org/packages/7e/99/7690b6d4034fffd95959cbe0c02de8deb3098cc577c67bb6a24fe5d7caa7/markupsafe-3.0.3.tar.gz
+        sha256: 722695808f4b6457b320fdc131280796bdceb04ab50fe1795cd540799ebe1698
+        x-checker-data:
+          type: pypi
+          name: MarkupSafe
+      - type: file
+        url: https://files.pythonhosted.org/packages/df/bf/f7da0350254c0ed7c72f3e33cef02e048281fec7ecec5f032d4aac52226b/jinja2-3.1.6.tar.gz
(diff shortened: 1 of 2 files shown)


hfiguiere:
> I'd like to request an exception (exactly like you did for org.telegram.desktop) for the following rule:

No.

> Doing a video will leak my contacts or phone number and I don't like to do that.

Then we can close this.

drizzt:
> > I'd like to request an exception (exactly like you did for org.telegram.desktop) for the following rule:
> 
> No.

so why telegram.desktop has the exception?

hfiguiere:
> > > I'd like to request an exception (exactly like you did for org.telegram.desktop) for the following rule:
> > 
> > 
> > No.
> 
> so why telegram.desktop has the exception?

Because the rule didn't exist. It would be rejected today. This is the problem with exceptions: there is always someone feeling entitled to ask for one.

drizzt:
> > > > I'd like to request an exception (exactly like you did for org.telegram.desktop) for the following rule:
> > > 
> > > 
> > > No.
> > 
> > 
> > so why telegram.desktop has the exception?
> 
> Because the rule didn't exist. It would be reject today. This is the problem with exceptions: there is always someone feeling entitled to ask for one.

oh ok, I didn't know it's a new rule... I'll refactor it to use it.belloworld.mercurygram instead (exactly like the Android App)

bbhtt:
What features does this fork have that telegram or no other telegram-forks have and why can't those features be contributed back to an existing fork or upstream telegram?

drizzt:
> What features does this fork have that telegram or no other telegram-forks have and why can't those features be contributed back to an existing fork or upstream telegram?

Of course, I thought about sending my PR to the original Telegram Desktop, but it [seems](https://github.com/telegramdesktop/tdesktop/issues/118#issuecomment-72011592) the authors doesn't like it since they want to keep Telegram mobile-centric. The target of Mercurygram (both mobile and desktop) is to avoid too bigger (bloat) features that make difficult to keep the fork aligned to upstream and introduces instability, but to focus on privacy features. See [mobile Mercurygram README](https://github.com/Mercurygram/Mercurygram#privacy--anti-tracking-mercurygram-only) for the features I have on Mercurygram and that I'll gradually migrate to Mercurygram Desktop (of course only the migrable features, since some features are strictly Android-related).

In the past, I used to use Kotatogram, but it's not updated since 4 years (maybe you should consider removing it, since keeping outdated software that may have security issue may not be a good idea) then I tried the other forks you currently have in flathub, but they are too bloated or they are updated not so frequently when upstream release a new version.

drizzt:
I updated the branch to use v6.9.3.2 tag that removed .desktop and I opened a [PR](https://github.com/flathub-infra/flatpak-builder-lint/pull/1060) to flatpak-builder-lint for lint exceptions.

it.belloworld.mercurygram.yml
```diff
@@ -0,0 +1,397 @@
...
+sdk: org.gnome.Sdk
+command: Mercurygram
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --device=all
+  - --talk-name=org.freedesktop.Notifications
```
bbhtt, line 13: why doesn't this use the portal?

it.belloworld.mercurygram.yml
```diff
@@ -0,0 +1,397 @@
...
+  - --talk-name=org.gnome.Mutter.IdleMonitor
```
bbhtt, line 14: why doesn't this use the portal?

it.belloworld.mercurygram.yml
```diff
@@ -0,0 +1,397 @@
...
+runtime-version: '50'
+sdk: org.gnome.Sdk
+command: Mercurygram
+finish-args:
+  - --share=ipc
+  - --share=network
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --device=all
```
bbhtt, line 12: why do you need device=all?

it.belloworld.mercurygram.yml
```diff
@@ -0,0 +1,397 @@
...
+        x-checker-data:
+          type: json
+          url: https://api.github.com/repos/xiph/rnnoise/releases/latest
+          tag-query: .tag_name
+          timestamp-query: .published_at
+          version-query: $tag | sub("^[vV]"; "")
+          url-query: '"https://github.com/xiph/rnnoise/releases/download/\($tag)/rnnoise-\($version).tar.gz"'
+      - type: shell
+        commands:
+          - sed -i 's/os_support.h/common.h/g;s/OPUS_CLEAR/RNN_CLEAR/g' src/vec{,_neon}.h
```
bbhtt: this should be a patch. why is this needed?

it.belloworld.mercurygram.yml
```diff
@@ -0,0 +1,397 @@
...
+      - '*'
+
+  - name: tdesktop
+    buildsystem: cmake-ninja
+    builddir: true
+    build-options:
+      cflags: -fpch-preprocess
+      cxxflags: -fpch-preprocess
+      env:
+        CCACHE_SLOPPINESS: pch_defines,time_macros
```
bbhtt, line 381: why is this needed?

bbhtt (changes requested):

drizzt:
@bbhtt it's exactly the same file (with the needed modification for Mercurygram) that are currently accepted for Telegram Desktop and for Forkgram (that you accepted 2 month ago: https://github.com/flathub/flathub/pull/8183#issuecomment-4273979120). why are you asking all that changes on my fork, when the other forks have it?

it.belloworld.mercurygram.yml
drizzt: **TL;DR** it's necessary to reduce ccache misses

It's a very recent upstream addendum: https://github.com/flathub/org.telegram.desktop/commit/13094b0724f5ab86a878483dabc3b544d33e109e.

The reason is that Telegram Desktop, and so my fork, builds with precompiled headers, with the -fpch-preprocess flags.
When ccache caches a PCH-using compile, it can't see inside the PCH to detect #define changes or time-macro usage, so it refuses to cache unless you tell it to relax those two checks.

See https://ccache.dev/manual/latest.html#_precompiled_headers for more details.

bbhtt:
Why don't you answer the review instead? I don't remember reviews from 2 months ago and this isn't a review for forkgram.

it.belloworld.mercurygram.yml
```diff
@@ -0,0 +1,397 @@
...
+          timestamp-query: .published_at
+          version-query: $tag | sub("^[vV]"; "")
+          url-query: '"https://github.com/xiph/rnnoise/releases/download/\($tag)/rnnoise-\($version).tar.gz"'
+      - type: shell
+        commands:
+          - sed -i 's/os_support.h/common.h/g;s/OPUS_CLEAR/RNN_CLEAR/g' src/vec{,_neon}.h
+    cleanup:
+      - '*'
+
+  - name: qt
```
hfiguiere, line 263: why do you not use the KDE runtime?

it.belloworld.mercurygram.yml
drizzt: Telegram Desktop, and so my fork, uses webcam for videocalls and for webcam is necessary to have device=all (https://docs.flatpak.org/en/latest/sandbox-permissions.html#device-access)

it.belloworld.mercurygram.yml
drizzt: Replaced the sed with the upstream rnnoise fix, commit 372f7b4 by the rnnoise author (https://github.com/xiph/rnnoise/commit/372f7b4b76cde4ca1ec4605353dd17898a99de38).

The 0.2 release fails to build because several headers include a non-existent os_support.h and use OPUS_CLEAR (https://github.com/xiph/rnnoise/issues/222).

Should I also send a PR to the Telegram Desktop manifest with that, since they are still using sed?

it.belloworld.mercurygram.yml
drizzt: IIRC there is no idle-detection portal.
org.gnome.Mutter.IdleMonitor is the only interface for last-input/away on a Wayland/GNOME session.
It's the same used for lots of browsers + Telegram Desktop and its forks

it.belloworld.mercurygram.yml
drizzt: All the [upstream notification manager](https://github.com/telegramdesktop/tdesktop/blob/dev/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp) is based on org.freedesktop.Notifications directly.
The portal spec has no inline reply (the text-input-in-notification field) and Telegram Desktop, and so my forks, uses this for quick reply from the notification.
I could probably send a PR upstream to migrate to notification portals, but we'll loose the quick reply from the notification feature and probably they won't accept this.

it.belloworld.mercurygram.yml
drizzt: I could probably do that, but the battle tested implementation (aka upstream) likes to pin a specific Qt version and statically link to it. Changing that could have some side effect that upstream build doesn't have.
I'd like to do that, so we can avoid build-time for Qt libraries and to be more policy-friendly, but it'll take some time to do that (like extensive testing, checking if something breaks, etc).
If possible, can we defer that for the next builds so that right now I can have my Telegram Fork pushed exactly like the other Telegram Desktop forks (so with GNOME runtime) and so I can work to migrate it to KDE runtime (by opening PR upstream, etc).
permanent divergence from upstream telegram is what I'd like to avoid, whenever is possible

bbhtt:
bot, build

drizzt:
bot, build

bbhtt:
GitHub is having an outage I will try again later.

PR #8997, opened 2026-06-15
Add io.github.GaimsDevSoftware.LinuxPop

GaimsDevSoftware:
## New app: LinuxPop (`io.github.GaimsDevSoftware.LinuxPop`)

LinuxPop is a PopClip-style floating popup of context-aware actions for
selected text — search, open URLs/paths, send to an AI assistant,
encode/decode, translate, run commands — plus a clipboard manager and a
no-code custom-button wizard. It runs on **X11** and on **KDE Plasma 6 /
Wayland** (native gtk-layer-shell popup, KWin scripting, KGlobalAccel).

- Upstream: https://github.com/GaimsDevSoftware/linuxpop (tag `v0.9.0`)
- Built locally end-to-end with `org.flatpak.Builder` against
  `org.gnome.Platform//49` — builds and runs (tray, hotkeys, popup,
  clipboard recording all verified in-sandbox).

### finish-args that need reviewer attention (requesting exceptions)
Two permissions trip the linter; both are required for the native
KDE/Wayland features and have no portal equivalent:

1. `--talk-name=org.kde.KWin` — used for the global cursor position
   (`workspace.cursorPos` via KWin scripting) so the popup appears at the
   selection, and to read the active window for the user's block-list. No
   XDG portal exposes the global cursor position.
2. `--filesystem=/tmp` — KWin's `loadScript` takes a **file path**, and a
   sandbox-private `/tmp` gives KWin a path it can't open (popup then lands
   at 0,0). Shared `/tmp` lets host-side KWin read the short-lived script,
   which is written and unloaded immediately. Happy to switch to a
   centred-popup fallback for the sandboxed build if `/tmp` isn't
   acceptable.

Screenshots are referenced by URL on the upstream `main` branch (Flathub
will mirror them). Thanks for reviewing!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "shared-modules"]
+	path = shared-modules
+	url = https://github.com/flathub/shared-modules.git
--- a/io.github.GaimsDevSoftware.LinuxPop.yml
+++ b/io.github.GaimsDevSoftware.LinuxPop.yml
@@ -1,0 +1,288 @@
+# Flatpak manifest for LinuxPop.
+#
+# Build locally with:
+#     flatpak-builder --user --install --force-clean build-dir \
+#         packaging/flatpak/io.github.GaimsDevSoftware.LinuxPop.yml
+#
+# Then run:
+#     flatpak run io.github.GaimsDevSoftware.LinuxPop
+#
+# Notes:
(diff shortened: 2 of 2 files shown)


teamcons:
Unable to follow the most base directions, removes the template, posts AI slop.
Will not be merged.

PR #9008, opened 2026-06-16
Add io.github.filegram.Filegram

stan220:
[Filegram](https://github.com/filegram/filegram-desktop) is a disk space
analyzer that visualizes disk usage as an interactive treemap.

- Built from source from the `v0.2.5` release tag (commit pinned).
- Rust app (iced/wgpu); built offline with the `rust-stable` SDK extension and
  vendored `cargo-sources.json` (513 crates, no git dependencies).
- `--filesystem=host` is required for a disk-usage analyzer: it reads the tree
  it scans and moves files to the trash (same scope as Baobab).
- AppStream metainfo passes `appstreamcli validate`.

Manifest source of truth lives at
https://github.com/filegram/filegram-desktop/tree/master/packaging/flatpak

### DIFF
--- a/cargo-sources.json
+++ b/cargo-sources.json
@@ -1,0 +1,6664 @@
+[
+    {
+        "type": "archive",
+        "archive-type": "tar-gzip",
+        "url": "https://static.crates.io/crates/ab_glyph/ab_glyph-0.2.32.crate",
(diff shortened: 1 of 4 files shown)


bbhtt:
We don't accept AI slop PRs per the generative AI policy.

PR #9023, opened 2026-06-17
Add io.github.ashbuk.dabri

AshBuk:

- [x] Please describe the application briefly. Dabri is a minimalist, privacy-first application for offline voice recognition directly into any active window 
- [x] Please attach a video showcasing the application on Linux using the Flatpak. https://github.com/user-attachments/assets/b798f350-c6a8-4408-83eb-51eef392a731


AshBuk:
bot, build

io.github.ashbuk.dabri.yml
```diff
@@ -0,0 +1,156 @@
+app-id: io.github.ashbuk.dabri
+runtime: org.gnome.Platform
+runtime-version: "50"
+sdk: org.gnome.Sdk
+command: dabri
+
+build-options:
+  libdir: /app/lib
+
```
hfiguiere:
```suggestion
```

io.github.ashbuk.dabri.yml
```diff
@@ -0,0 +1,156 @@
...
+  - shared-modules/libayatana-appindicator/libayatana-appindicator-gtk3.json
+
+  - name: golang
+    buildsystem: simple
+    build-commands:
+      - mkdir -p /app/lib
+      - cp -a go /app/lib/go
+    sources:
+      - type: archive
+        url: https://go.dev/dl/go1.25.3.linux-amd64.tar.gz
```
hfiguiere: What happen on aarch64?

*hint* this is not how you use Go on to build flatpak.

io.github.ashbuk.dabri.yml
```diff
@@ -0,0 +1,156 @@
...
+        sha256: 0335f314b6e7bfe08c3d0cfaa7c19db961b7b99fb20be62b0a826c992ad14e0f
+        strip-components: 0
+
+  - name: ffmpeg
```
hfiguiere: this is in the runtime.

io.github.ashbuk.dabri.yml
```diff
@@ -0,0 +1,156 @@
...
+    buildsystem: cmake-ninja
+    sources:
+      - type: archive
+        url: https://github.com/KhronosGroup/SPIRV-Headers/archive/refs/tags/vulkan-sdk-1.4.321.0.tar.gz
+        sha256: 5bbea925663d4cd2bab23efad53874f2718248a73dcaf9dd21dff8cb48e602fc
+
+  - name: whisper-cpp
+    buildsystem: simple
+    build-commands:
+      - cmake -B build -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON -DCMAKE_PREFIX_PATH=/app -DCMAKE_INSTALL_RPATH=/app/lib/dabri -DCMAKE_BUILD_WITH_INSTALL_RPATH=ON -DGGML_NATIVE=OFF -DGGML_AVX=ON -DGGML_AVX2=ON -DGGML_FMA=ON -DGGML_F16C=ON -DGGML_AVX512=OFF -DGGML_VULKAN=ON
```
hfiguiere: this should be `buildsystem: cmake-ninja`

io.github.ashbuk.dabri.yml
```diff
@@ -0,0 +1,156 @@
...
+  - name: dabri
+    buildsystem: simple
+    build-options:
+      append-path: /app/lib/go/bin
+      env:
+        CGO_ENABLED: "1"
+        CGO_CFLAGS: -I/app/include/dabri
+        CGO_LDFLAGS: -L/app/lib/dabri -lwhisper -lggml -lggml-cpu -lggml-vulkan
+        LD_LIBRARY_PATH: /app/lib/dabri
+        PKG_CONFIG_PATH: /app/lib/pkgconfig:/usr/lib/pkgconfig
```
hfiguiere:
```suggestion
```

this is properly set

io.github.ashbuk.dabri.yml
```diff
@@ -0,0 +1,156 @@
...
+    buildsystem: meson
+    sources:
+      - type: archive
+        url: https://github.com/atx/wtype/archive/refs/tags/v0.4.tar.gz
+        sha256: da91786d828b6a6e29b884bc510473939eda052658ebef87d7bdeafa6a8746f9
+
+  - name: ydotool
+    buildsystem: simple
+    build-commands:
+      - sed -i '/add_subdirectory(manpage)/d;/add_subdirectory(Daemon)/d' CMakeLists.txt
```
hfiguiere: this should be a patch

io.github.ashbuk.dabri.yml
```diff
@@ -0,0 +1,156 @@
...
+      - cmake -B build -G Ninja -DCMAKE_INSTALL_PREFIX=/app -DCMAKE_BUILD_TYPE=Release -DCMAKE_POLICY_VERSION_MINIMUM=3.5
```
hfiguiere: this should be `buildsystem: cmake-ninja`

io.github.ashbuk.dabri.yml
```diff
@@ -0,0 +1,156 @@
+app-id: io.github.ashbuk.dabri
+runtime: org.gnome.Platform
+runtime-version: "50"
```
hfiguiere: why does it need the GNOME runtime?

io.github.ashbuk.dabri.yml
AshBuk: Hi, thanks for fast review! whisper.cpp here is built with x86-specific flags (-DGGML_AVX=ON -DGGML_AVX2=ON -DGGML_FMA=ON -DGGML_F16C=ON) - so it doesn't compile on aarch64, x86_64 only. Is this ok?

io.github.ashbuk.dabri.yml
AshBuk: the window and the tray (libayatana-appindicator) both need GTK3 (so needs GNOME runtime)

hfiguiere: no it doesn't.

AshBuk: Right, the tray is fyne.io/systray (pure DBus, no GTK). GNOME runtime is for the GTK3 settings window (gotk3).
I'll look into your feedback. I'll get back with the updates, thanks!

hfiguiere: gtk3 doesn't require to have the GNOME runtime.

io.github.ashbuk.dabri.yml
```diff
@@ -0,0 +1,156 @@
...
+  libdir: /app/lib
+
+finish-args:
+  - --share=network
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --device=dri
+  - --talk-name=org.freedesktop.Notifications
```
CodedOre: You should use the portal for notifications instead.

io.github.ashbuk.dabri.yml
AshBuk: Ah right, GTK3 is actually in the freedesktop runtime 🙃️

io.github.ashbuk.dabri.yml
CodedOre: You know you can set architecture-specific configuration options?

That should not make an aarch64 build impossible.

hfiguiere: I don't see the relation between build flags and incorrectly using the Go toolchain. 

Here the question was "how will the x86_64 binary you install (and not cleanup) will run on aarch64 when we build the aarch64?".

AshBuk: @CodedOre  this setup is fairly hard-won and tuned for x86_64, so for now I'd prefer to ship only-arches: ["x86_64"]. Is that OK for Flathub?

@hfiguiere  You're right about the Go toolchain, already fixed and will be commited.

hfiguiere: > prefer to ship only-arches: ["x86_64"]. Is that OK for Flathub?

we prefer that you actually also build for aarch64. There is no valid reason to not.

io.github.ashbuk.dabri.yml
AshBuk: You're right - the runtime's ffmpeg ships with libpulse, the audio backend just uses that. Dropped.

io.github.ashbuk.dabri.yml
AshBuk: Portal notifications frontend isn't available on all compositors (wlroots/Hyprland expose only ScreenCast/GlobalShortcuts). But I made this portal-first (with a fallback to org.freedesktop.Notifications, hence the talk-name)

AshBuk:
bot, build

AshBuk:
Hey @hfiguiere @CodedOre,  I went through the review and reworked the manifest accordingly.
Already tested on upstream. The manifest got leaner, and the flatpak bundle itself shrank from 18 to 10 MB. Thanks!

io.github.ashbuk.dabri.yml
bbhtt: Sounds like those compoistors' issue? It works on majority of the desktop environments

io.github.ashbuk.dabri.yml
```diff
@@ -0,0 +1,139 @@
...
+    buildsystem: simple
+    build-options:
+      append-path: /usr/lib/sdk/golang/bin
+      env:
+        CGO_ENABLED: "1"
+        CGO_CFLAGS: -I/app/include/dabri
+        CGO_LDFLAGS: -L/app/lib/dabri -lwhisper -lggml -lggml-cpu -lggml-vulkan
+        LD_LIBRARY_PATH: /app/lib/dabri
+    build-commands:
+      - APP_VERSION="$(python3 -c 'import xml.etree.ElementTree as ET; print(ET.parse("io.github.ashbuk.dabri.appdata.xml").getroot().find("releases/release").attrib["version"])')" && go build -mod=vendor -tags systray,gtk -trimpath -ldflags "-s -w -X github.com/AshBuk/dabri/v2/internal/version.Version=${APP_VERSION} -linkmode=external -extldflags '-Wl,-rpath,/app/lib/dabri'" -o dabri ./cmd/dabri
```
bbhtt: what is happening here? why are you executing a Python script?

io.github.ashbuk.dabri.yml
```diff
@@ -0,0 +1,139 @@
...
+      - type: archive
+        url: https://github.com/KhronosGroup/SPIRV-Headers/archive/refs/tags/vulkan-sdk-1.4.321.0.tar.gz
+        sha256: 5bbea925663d4cd2bab23efad53874f2718248a73dcaf9dd21dff8cb48e602fc
+
+  - name: whisper-cpp
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DBUILD_SHARED_LIBS=ON
+      - -DCMAKE_PREFIX_PATH=/app
```
hfiguiere:
```suggestion
```

io.github.ashbuk.dabri.yml
```diff
@@ -0,0 +1,139 @@
...
+        url: https://github.com/ReimuNotMoe/ydotool/archive/refs/tags/v1.0.4.tar.gz
+        sha256: ba075a43aa6ead51940e892ecffa4d0b8b40c241e4e2bc4bd9bd26b61fde23bd
+      - type: patch
+        path: packaging/flatpak/ydotool-drop-manpage-daemon.patch
+
+  # SPIRV-Headers cmake config - required by whisper.cpp >= v1.8.5 Vulkan backend
+  - name: spirv-headers
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_INSTALL_LIBDIR=lib
```
hfiguiere:
```suggestion
```

io.github.ashbuk.dabri.yml
```diff
@@ -0,0 +1,139 @@
...
+    buildsystem: meson
+    sources:
+      - type: archive
+        url: https://github.com/atx/wtype/archive/refs/tags/v0.4.tar.gz
+        sha256: da91786d828b6a6e29b884bc510473939eda052658ebef87d7bdeafa6a8746f9
+
+  - name: ydotool
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_INSTALL_LIBDIR=lib
```
hfiguiere:
```suggestion
```

io.github.ashbuk.dabri.yml
```diff
@@ -0,0 +1,139 @@
...
+
+  - name: whisper-cpp
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DBUILD_SHARED_LIBS=ON
+      - -DCMAKE_PREFIX_PATH=/app
+      - -DCMAKE_INSTALL_LIBDIR=lib/dabri
+      - -DCMAKE_INSTALL_INCLUDEDIR=include/dabri
+      - -DCMAKE_INSTALL_RPATH=/app/lib/dabri
```
hfiguiere: I don't see why this should be installed in that directory...

io.github.ashbuk.dabri.yml
AshBuk: It read the app version. Fair point thats a bit hack though. I'll drop it and read the version from the installed metainfo at runtime instead.

io.github.ashbuk.dabri.yml
AshBuk: dropped, ydotool only installs binaries so libdir doesn't matter here

io.github.ashbuk.dabri.yml
AshBuk: Kept only this one. Without it whisper doesn't link (my Flatpak CI failed on exactly that): non-Debian SDK > GNUInstallDirs defaults to lib64, libs land in /app/lib64.

io.github.ashbuk.dabri.yml
AshBuk: dropped

io.github.ashbuk.dabri.yml
AshBuk: dropped the /dabri subdir and the rpath, whisper moved to /app/lib

io.github.ashbuk.dabri.yml
hfiguiere: no
you are using the wrong version of flatpak-builder.
you don't need it.

hfiguiere: https://docs.flathub.org/docs/for-app-authors/submission#build-and-install

AshBuk: changed it, checking now

AshBuk:
bot, build

io.github.ashbuk.dabri.yml
AshBuk: Agreed it's the compositors' gap. But wlroots compositors are a good chunk of this app's users. It ships ydotool exactly for them. so I'd rather keep notifications working there than drop them until those portals catch up. The talk-name is minimal and **the portal stays the default path**; happy to drop it once Notification is common on wlroots.

io.github.ashbuk.dabri.yml
AshBuk: good news: dropped

io.github.ashbuk.dabri.yml
CodedOre: AFAIK wlroots compositors can use `xdg-desktop-portal-gtk` to implement the notification portal. I mean they probably rely on it already for the FileChooser portal. 

Not having a implementation for the notification portal would sound like an bug that should be resolved by whoever has setup the desktop, I.e. the admin or distribution. 

We cant always carve out exceptions just because of that.

AshBuk: > Not having a implementation for the notification portal would sound like an bug that should be resolved by whoever has setup the desktop, I.e. the admin or distribution.

It's not all that exotic, I wouldn't call it a bug 🙂
Although since it's a matter of principle the fallback was removed, now it's portal only.

AshBuk:
bot, build

io.github.ashbuk.dabri.yml
```diff
@@ -0,0 +1,124 @@
...
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --device=dri
+  - --talk-name=com.canonical.AppMenu.Registrar
+  - --talk-name=org.kde.StatusNotifierWatcher
+
+cleanup:
+  - /include
+  - /lib/pkgconfig
+  - /lib/cmake
```
hfiguiere, line 22:
```suggestion
  - /lib/cmake
  - /share/cmake
  - /share/bash-completion
  - /share/pkgconfig
```

AshBuk:
bot, build

bbhtt:
You removed the actual checklist items and replaced them with the answers instead of putting them in the designated area...

I fixed it.

bbhtt:
On first launch it seems download a model then start in tray. There is no visual feedback for this and tray will not work for systems without it. Can you make it show a visual indicator of the download and a first time window saying it is launching into tray or something?

Also would be nice if you can bundle the model it is downloading in the manifest. Then I imagine it will not need network access? Or, at least on first launch it will work without downloading anything.

AshBuk:
> On first launch it seems download a model then start in tray. There is no visual feedback for this and tray will not work for systems without it.

Yeah, GNOME doesn't have a system tray out of the box, that's documented behavior. 

> Can you make it show a visual indicator of the download and a first time window saying it is launching into tray or something?

agree

> Also would be nice if you can bundle the model it is downloading in the manifest. Then I imagine it will not need network access? Or, at least on first launch it will work without downloading anything.

There's a trade-off. Network access is still required to download Whisper models and to allow switching between them - we just can't bundle all of them

bbhtt:
But you can bundle the one it is downloading on first launch? It seems to be ~180 MB or something. I think that will be better UX wise too.

> agree

Yes so I suggest setting up a first launch window.

AshBuk:
> But you can bundle the one it is downloading on first launch? It seems to be ~180 MB or something. I think that will be better UX wise too.

The small model  was bundled (2.2.0 tag), I'll update the manifest once it lands on upstream Flathub, thanks

AshBuk:
bot, build

AshBuk:
> On first launch it seems download a model then start in tray. There is no visual feedback for this and tray will not work for systems without it. Can you make it show a visual indicator of the download and a first time window saying it is launching into tray or something?
> 
> Also would be nice if you can bundle the model it is downloading in the manifest. Then I imagine it will not need network access? Or, at least on first launch it will work without downloading anything.

Hi @bbhtt! New manifest with the small model bundled (no network required by default). The launch window now clearly shows it minimizes to tray, and (for GNOME) a tooltip on the App Indicator requirement.

AshBuk:
@hfiguiere @CodedOre  Hi! Just a ping, is everything okay (this has been waiting for a while)? Can it be published, or is anything else required from my side?

PR #9026, opened 2026-06-18
Add io.github.pebois.gnome-ai-chat

pebois:
Please add GNOME AI Chat to Flathub.


CodedOre:

See also our Generative AI policy:

PR #9036, opened 2026-06-19
Add io.github.eltonfabricio10.bigtube

eltonfabricio10:
New app: BigTube — a GTK4/libadwaita multimedia downloader (yt-dlp front-end), written in Rust. Crates are vendored via cargo-sources.json so the build is fully offline; bundles lame + ffmpeg for audio extraction/conversion.


hfiguiere:
Since you removed the template and failed to follow the basic requirements there is no reason to review this PR.

See also our AI policy

PR #9038, opened 2026-06-19
Add org.linphone.linphone

Ferenc-:

- [x] Please describe the application briefly.  **Open source softphone for voice and video over IP calling and instant messaging.**
[Screencast From 2026-06-19 23-39-17.webm](https://github.com/user-attachments/assets/ec5ddba7-2bfe-487c-a07b-3e5832e107f1)

- [x] I am an _(please keep whichever is applicable and remove the rest)_ author/developer/upstream contributor to the project. If not, I contacted upstream developers about this submission. **Link:** https://github.com/BelledonneCommunications/linphone-desktop/issues/820#issuecomment-4754789524


org.linphone.linphone.metainfo.xml
hfiguiere, line 1: This file should be (submitted and managed) upstream.

org.linphone.linphone.yml
```diff
@@ -0,0 +1,495 @@
...
+  - --talk-name=org.freedesktop.PowerManagement
+  - --talk-name=org.freedesktop.ScreenSaver
+  - --talk-name=org.gtk.Notifications
+  - --talk-name=org.gnome.SessionManager
+cleanup:
+  - /bin/pystache
+  - /bin/pystache-test
+  - /bin/vsyasm
+  - /bin/yasm
+  - /bin/ytasm
```
hfiguiere: why aren't these cleanup in the yasm module?

org.linphone.linphone.yml
```diff
@@ -0,0 +1,495 @@
...
+  - /lib/python3.13
+modules:
+  # Development tools from:
+  # https://github.com/BelledonneCommunications/linphone-desktop/blob/master/docker-files/bc-dev-ubuntu-22-04-lts
+  ## Python dependencies
+  - name: python3-pystache-six
+    buildsystem: simple
+    build-commands:
+      - ls -lah
```
hfiguiere:
```suggestion
```

org.linphone.linphone.yml
```diff
@@ -0,0 +1,495 @@
...
+cleanup:
+  - /bin/pystache
+  - /bin/pystache-test
+  - /bin/vsyasm
+  - /bin/yasm
+  - /bin/ytasm
+  - /lib/python3.13
+modules:
+  # Development tools from:
+  # https://github.com/BelledonneCommunications/linphone-desktop/blob/master/docker-files/bc-dev-ubuntu-22-04-lts
```
hfiguiere, line 20: if these are build-only then there should be a cleanup "*" rule for it.

org.linphone.linphone.yml
```diff
@@ -0,0 +1,495 @@
...
+      - /share/Speex
+      - /share/decaf
+      - /share/libjpeg-turbo
+      - /share/mbedtls
+    config-opts:
+      - -DCMAKE_BUILD_PARALLEL_LEVEL=10
+      - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+      - -DCMAKE_INCLUDE_PATH=/app/include
+      - -DCMAKE_LIBRARY_PATH=/app/lib
+      - -DCMAKE_LIBRARY_PATH=/app/lib
```
hfiguiere:
```suggestion
```

shouldn't be needed.

org.linphone.linphone.yml
```diff
@@ -0,0 +1,495 @@
...
+      - /share/Mediastreamer2
+      - /share/Ortp
+      - /share/PostQuantumCryptoEngine
+      - /share/SQLITE3
+      - /share/Speex
+      - /share/decaf
+      - /share/libjpeg-turbo
+      - /share/mbedtls
+    config-opts:
+      - -DCMAKE_BUILD_PARALLEL_LEVEL=10
```
hfiguiere: let flatpak builder handle this.
```suggestion
```

org.linphone.linphone.yml
```diff
@@ -0,0 +1,495 @@
...
+        # Either use BC's codebase directly:
+        # url: https://gitlab.linphone.org/BC/public/external/bv16-floatingpoint/-/archive/c4a341a3277e8adcc42427c4fd80a457fe934f2a/bv16-floatingpoint-c4a341a3277e8adcc42427c4fd80a457fe934f2a.tar.bz2
+        # sha256: f28d6f0ae6d27d5f86d3cdf73c6c8d05a22096e6082aaaefa89390d1892b139b
+        # Or a faster more reliable mirror:
+        url: https://github.com/Ferenc-/linphone-bv16-floatingpoint/archive/c4a341a3277e8adcc42427c4fd80a457fe934f2a.zip
+        sha256: 07fb1ca72bd9eeb18e8298d899fac0ca9d956b9f9d623a3958befb5279b4250a
+      - type: archive
+        dest: external/linphone-sdk/external/dav1d
+        url: https://github.com/videolan/dav1d/archive/refs/tags/1.5.3.tar.gz
+        # NOTE: dav1d is part of the SDK but it's lagging behind the latest
```
hfiguiere: what do you need that isn't in the runtime

org.linphone.linphone.yml
```diff
@@ -0,0 +1,495 @@
...
+        url: https://github.com/open-source-parsers/jsoncpp/archive/refs/tags/1.9.7.tar.gz
+        sha256: 830bf352d822d8558e9d0eb19d640d2e38536b4b6699c30a4488da09d5b1df18
+        x-checker-data:
+          type: anitya
+          project-id: 7483
+          stable-only: true
+          url-template: https://github.com/open-source-parsers/jsoncpp/archive/refs/tags/$version.tar.gz
+      - type: archive
+        dest: external/linphone-sdk/external/libjpeg-turbo
+        # NOTE: libjpeg-turbo is part of the SDK but it's lagging behind the latest
```
hfiguiere, line 329: what do you need that isn't in the runtime

org.linphone.linphone.yml
```diff
@@ -0,0 +1,495 @@
...
+          type: anitya
+          project-id: 364436
+          stable-only: true
+          url-template: https://github.com/open-quantum-safe/liboqs/archive/refs/tags/$version.tar.gz
+      - type: archive
+        dest: external/linphone-sdk/external/libvpx
+        # Reproducible alternative to:
+        # url: https://chromium.googlesource.com/webm/libvpx/+archive/v1.16.0.tar.gz
+        url: https://github.com/webmproject/libvpx/archive/refs/tags/v1.16.0.tar.gz
+        # NOTE: libvpx is part of the SDK but it's lagging behind the latest
```
hfiguiere: what do you need that isn't in the runtime

org.linphone.linphone.yml
```diff
@@ -0,0 +1,495 @@
...
+        sha256: 7a479a3c66b9f5d5542a4c6a1b7d3768a983b1e5c14c60a9396edc9b649e015c
+        x-checker-data:
+          type: anitya
+          project-id: 11083
+          stable-only: true
+          url-template: https://github.com/webmproject/libvpx/archive/refs/tags/$version.tar.gz
+      - type: archive
+        dest: external/linphone-sdk/external/libxml2
+        url: https://github.com/GNOME/libxml2/archive/refs/tags/v2.15.3.tar.gz
+        # NOTE: libxml2 is part of the SDK but it's lagging behind the latest
```
hfiguiere: what do you need that isn't in the runtime?

org.linphone.linphone.yml
```diff
@@ -0,0 +1,495 @@
...
+        url: https://github.com/SOCI/soci/archive/refs/tags/v4.1.2.tar.gz
+        sha256: c0974067e57242f21d9a85677c5f6cc7848fba3cbd5ec58d76c95570a5a7a15b
+        x-checker-data:
+          type: anitya
+          project-id: 11081
+          stable-only: true
+          url-template: https://github.com/SOCI/soci/archive/refs/tags/v$version.tar.gz
+      - type: archive
+        dest: external/linphone-sdk/external/speex
+        url: https://github.com/xiph/speex/archive/refs/tags/Speex-1.2.1.tar.gz
```
hfiguiere: turns out this one version is in the runtime.

org.linphone.linphone.yml
```diff
@@ -0,0 +1,495 @@
...
+        sha256: 0f28ea2d7a0a3824d326c8bc4be26a3bb01dde289663beb6f15b241f80cbd3ae
+        x-checker-data:
+          type: anitya
+          project-id: 5182
+          stable-only: true
+          url-template: https://github.com/apache/xerces-c/archive/refs/tags/v$version.tar.gz
+      - type: archive
+        dest: external/linphone-sdk/external/zlib
+        url: https://github.com/madler/zlib/releases/download/v1.3.2/zlib-1.3.2.tar.xz
+        # NOTE: Zlib is part of the SDK but it's lagging behind the latest
```
hfiguiere: what do you need that isn't in the runtime?

org.linphone.linphone.yml
```diff
@@ -0,0 +1,495 @@
...
+        dest: external/linphone-sdk/postquantumcryptoengine
+        # Same as all the releases in the 5.4.X series only 5.4.106 is missing
+        url: https://github.com/BelledonneCommunications/postquantumcryptoengine/archive/refs/tags/5.4.102.tar.gz
+        sha256: c4bd907140a34562d4e6de306d00804fd56d3e6cba56dd6d987433c4b9ac3b41
+      - type: archive
+        dest: external/linphone-sdk/external/aom
+        # Reproducible alternative to:
+        # url: https://aomedia.googlesource.com/aom/+archive/v3.13.3.tar.gz
+        url: https://github.com/m-ab-s/aom/archive/refs/tags/v3.13.3.tar.gz
+        # NOTE: AOM is currently part of the SDK but it's lagging behind the latest
```
hfiguiere: that's actually the version in the runtime.

org.linphone.linphone.yml
```diff
@@ -0,0 +1,495 @@
...
+          - patches/sdk/external/gsm/Add-CMakeLists.txt.patch
+          - patches/sdk/external/liboqs/CMake-subdir.patch
+          - patches/sdk/external/libyuv/FindTurboJpeg-and-CMake-integration.patch
+          - patches/sdk/external/mbedtls/CMake-Integration-And-HW-Accelerated-SHA256-on-ARMv8A.patch
+          - patches/sdk/external/openldap/Add-CMake-Enable-MbedTLS-And-Fix-Compilation.patch
+          - patches/sdk/external/opus/Add-CMake-Fix-CPU-support.patch
+          - patches/sdk/external/srtp/Enhanced-FindMbedTLS.cmake.patch
+          - patches/sdk/external/speex/Add-CMake.patch
+          - patches/sdk/external/xerces-c/Fix-C++-minimum-level.patch
+  - name: metainfo
```
hfiguiere: this shouldn't be a separate module.

org.linphone.linphone.yml
```diff
@@ -0,0 +1,495 @@
...
+      - -DLINPHONESDK_PLATFORM=Desktop
+      # While this seems redundandt, without this `bctoolbox` is not buildable from archive, as cmake would try to use git to read the repo
+      - -DLINPHONESDK_VERSION=5.4.106 # Also Linphone-desktop needs this
+      # While this seems redundandt, without this `bctoolbox` is not buildable from archive, as cmake would try to use git to read the repo
+      - -DLINPHONEAPP_VERSION=6.1.2
+      - -DGIT_BRANCH_NAME=6.1.2
+      # Option added by `CMake-flatpak-packaging.patch`
+      - -DFLATPAK_ID=org.linphone.linphone
+    post-install:
+      - desktop-file-edit --set-key="Icon" --set-value="${FLATPAK_ID}" ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop
```
hfiguiere: I don't see where it renames the icon. Either the desktop file is just wrong upstream, or the icon won't be found.

org.linphone.linphone.yml
```diff
@@ -0,0 +1,495 @@
+app-id: org.linphone.linphone
+runtime: org.kde.Platform
+runtime-version: "6.10"
+sdk: org.kde.Sdk
+command: linphone
+finish-args:
+  - --device=dri
```
hfiguiere:
```suggestion
```
it's included in `all`

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "shared-modules"]
+	path = shared-modules
+	url = git@github.com:flathub/shared-modules.git
```
hfiguiere: this should be https

org.linphone.linphone.yml
Ferenc-: These can't be cleaned up in the yasm module, because these are needed to build the main module.

hfiguiere: cleanup doesn't do what you think it does.

org.linphone.linphone.yml
Ferenc-: Removed.

org.linphone.linphone.yml
Ferenc-: Upstream has the `cmake/install/install.cmake` file which installs the icon with the right name.
The `patches/linphone-desktop/CMake-flatpak-packaging.patch` ensures that the cmake file also uses the `FLATPAK_ID` for the right name.

org.linphone.linphone.yml
Ferenc-: Removed.

org.linphone.linphone.yml
Ferenc-: Removed.

patches/linphone-desktop/CMake-flatpak-packaging.patch
```diff
@@ -0,0 +1,49 @@
...
+ 		set(CPACK_PACKAGE_FILE_NAME "${CPACK_PACKAGE_NAME}-${PACKAGE_VERSION}")
+ 		set(PACKAGE_EXT "AppImage")
++]]
+ 		configure_file("${CMAKE_SOURCE_DIR}/cmake/install/linux/linphone.desktop.cmake" "${CMAKE_BINARY_DIR}/cmake/install/linux/${EXECUTABLE_NAME}.desktop" @ONLY)
+-		install(FILES "${CMAKE_BINARY_DIR}/cmake/install/linux/${EXECUTABLE_NAME}.desktop" DESTINATION "${CMAKE_INSTALL_DATADIR}/applications" PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
+-		install(FILES "${CMAKE_SOURCE_DIR}/Linphone/data/image/logo.svg" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/scalable/apps/" RENAME "${EXECUTABLE_NAME}.svg")
+-		set(ICON_DIRS 16x16 22x22 24x24 32x32 64x64 128x128 256x256)
++		install(FILES "${CMAKE_BINARY_DIR}/cmake/install/linux/${EXECUTABLE_NAME}.desktop" DESTINATION "${CMAKE_INSTALL_DATADIR}/applications" RENAME "${FLATPAK_ID}.desktop" PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
++		install(FILES "${CMAKE_SOURCE_DIR}/Linphone/data/image/logo.svg" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/scalable/apps/" RENAME "${FLATPAK_ID}.svg")
```
hfiguiere: these two are unecessary. There is `rename-desktop-file` and `rename-icon` that takes care of it. If you want to fix it upstream for a future release, just submit the patch.

Ferenc-:
bot, build

org.linphone.linphone.metainfo.xml
Ferenc-: I agree, in an ideal world it should be.
The problem is that, AFAIK such a file used to exist upstream, I believe with an upstream flatpak manifest but all that has been removed as it was considered too much maintenance.
If this PR passes submission criteria, then I will try to submit  upstream,
this way we could have a stronger argument for it.

.gitmodules
Ferenc-: Fixed.

org.linphone.linphone.metainfo.xml
hfiguiere: If it was removed then it should be put back. This is not flatpak specific at all. It predates flatpak.

org.linphone.linphone.yml
Ferenc-: I stand corrected. It's fixed now.

org.linphone.linphone.yml
Ferenc-: Removed.

org.linphone.linphone.yml
Ferenc-: Done.

org.linphone.linphone.metainfo.xml
Ferenc-: I know. It's not me who needs to be convinced. I will try to upstream.

org.linphone.linphone.yml
```diff
@@ -0,0 +1,482 @@
...
+finish-args:
+  # NOTE: As long as /dev/video is directly needed and neither the Camera portal nor device=usb helps (also used for USB microphones)
+  - --device=all
+  - --share=ipc
+  - --share=network
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --socket=wayland
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=org.freedesktop.Notifications
```
bbhtt: why doesn't this use the portal?

org.linphone.linphone.yml
```diff
@@ -0,0 +1,482 @@
...
+  - --talk-name=org.freedesktop.PowerManagement
+  - --talk-name=org.freedesktop.ScreenSaver
+  - --talk-name=org.gtk.Notifications
```
bbhtt: gtk notifications is supposed to use the portal.

org.linphone.linphone.yml
```diff
@@ -0,0 +1,482 @@
...
+  - --talk-name=org.gnome.SessionManager
```
bbhtt: the rest of this org.freedesktop.PowerManagement and org.freedesktop.ScreenSaver I assume is for idle inhibition. There is a portal for it.

org.linphone.linphone.yml
Ferenc-: Right. Fixed.

bbhtt:
This seems a ridiculous amount of patches.

org.linphone.linphone.yml
hfiguiere: this is not resolved. see https://github.com/flathub/flathub/pull/9038/files#r3445198567

Ferenc-:
> This seems a ridiculous amount of patches.

I agree, this more than ideal, and I have already tried to reduce this number.
Do you happen to have any concrete suggestions, regarding how would you reduce them?

bbhtt:
I fixed the PR body.

>  Do you happen to have any concrete suggestions, regarding how would you reduce them?

I don't know. Most of them are adding cmakelists or patching that. Why are they needed?

hfiguiere:
for starter some of the patches are to build things that are in the runtime. Maybe remove all of these since you haven't answered why you needed to build it vs the one in the runtime.

patches/linphone-desktop/CMake-flatpak-packaging.patch
Ferenc-: I am checking [the docs here](https://docs.flatpak.org/en/latest/manifests.html#file-renaming), it says:
> The best way to do this is to rename these files directly in the application’s source.  If renaming exported files to use the application ID is not possible
[...]
Note that this renaming method can introduce internal naming conflicts, and that renaming files in tree is therefore the most reliable approach.

I might have misinterpreted this.
In any case now rename-desktop-file and rename-icon are utilized.

hfiguiere: what it is saying that doing it upstream is best. So submitting the patches upstream IS a good idea. But there is no point of having them here.

org.linphone.linphone.yml
Ferenc-: > I don't see where it renames the icon

Now we have `rename-icon: linphone` at the top of the manifest.

org.linphone.linphone.yml
Ferenc-: Removed except `SreenSaver` which is directly used by `ScreenSaverDBus.cpp` and currently is not portal aware. Also kept `StatusNotifierWatcher`, because currently without it the status icon won't get displayed.
Currently it appears that these 2 can't be eliminated without feature loss or even more patches.
If they absolutely need to go then I guess then there will be feature loss...

org.linphone.linphone.yml
Ferenc-: See above.

org.linphone.linphone.yml
Ferenc-: `Notifications` removed, for the rest see below.

org.linphone.linphone.yml
Ferenc-: Removed.

org.linphone.linphone.yml
Ferenc-: Removed.

org.linphone.linphone.yml
Ferenc-: Removed.

org.linphone.linphone.yml
Ferenc-: Removed

org.linphone.linphone.yml
Ferenc-: Removed

org.linphone.linphone.yml
Ferenc-: Removed.

org.linphone.linphone.yml
hfiguiere: still here.

don't resolve thread when they are not resolved.

org.linphone.linphone.yml
hfiguiere: not resolved

org.linphone.linphone.yml
hfiguiere: not resolved.

org.linphone.linphone.yml
hfiguiere: not resolved.

org.linphone.linphone.yml
hfiguiere: not resolved.

org.linphone.linphone.yml
hfiguiere: not resolved.

org.linphone.linphone.yml
hfiguiere: not resolved.

org.linphone.linphone.yml
Ferenc-: https://github.com/Ferenc-/flathub/blob/c95159be631979ab2d518bf467b51fa251f54f9d/org.linphone.linphone.yml

patches/sdk/external/libjpeg-turbo/CMakeLists.txt
hfiguiere, line 1: this is no longer needed, is it?

org.linphone.linphone.yml
hfiguiere: still not

org.linphone.linphone.yml
```diff
@@ -0,0 +1,495 @@
...
+        sha256: d7a0654783a4da529d1bb793b7ad9c3318020af77667bcae35f95d0e42a792f3
+        x-checker-data:
+          type: anitya
+          project-id: 5303
+          stable-only: true
+          url-template: https://github.com/madler/zlib/releases/download/v$version/zlib-$version.tar.gz
+      - type: patch
+        paths:
+          - patches/linphone-desktop/Own-name-match-Flatpak-ID.patch # https://docs.flathub.org/docs/for-app-authors/linter#finish-args-own-name-cpt
+          - patches/linphone-desktop/Add-missing-English-translation.patch
```
hfiguiere, line 387: majority of these patches shouldn't exist. If linphone build system is so broken it should be fixed first before attempting to package it.

patches/sdk/external/libjpeg-turbo/CMakeLists.txt
Ferenc-: Removed in https://github.com/flathub/flathub/pull/9038/commits/576d955cf7ff70eabf33991b42b32d6f433e986a

hfiguiere: (when you actually remove libjpeg-turbo from here)

org.linphone.linphone.yml
Ferenc-: Should I close this PR already, so we don't waste more time?

hfiguiere: we can wait there is no urgency.

Ferenc-: > we can wait there is no urgency.

Upstreaming patches in this project is extremely slow, and a good portion of these patches are unlikely to be accepted.

BTW which patches do you think must be eliminated before FH submission could be accepted, and why?

* Own-name-match-Flatpak-ID -> Unlikely to be accepted because this breaks backwards compatibilty and this is flatpak specific, and flatpak has been explicitly dropped upstream some time ago.
* Add-missing-English-translation -> There is an open PR for this upstream already waiting since a while, even if this is super straightforward change. If this is a blocker, then I can simply remove this.
* CMake-build-from-archive  -> Unlikely to be accepted, because this is needed because of the flathub build system.
* CMake-flatpak-packagin ->  This is obviously unacceptable upstream, but any high quality equivalent is just as unlikely to be accepted upstream, since this is flatpak specific, which has been dropped before.
* Fix-use-of-function-deleted-in-SOCI-4.1.2-and-potential-stack-use-after-scope -> Obviously unacceptable upstream at this point, because upstream uses a much older version of SOCI where this doesn't apply. There is a completely different 3PP lifecycle there. Theoretically this could be eliminated with downgrading in the flatpak.
* Find-dependencies -> This might be upstreamable after some reworking, but I personally wouldn't wait for it.

The rest are external and not strictly  linphone specific problems:
* gsm/Add-CMakeLists.txt -> GSM is an ancient historical project, it is a miracle that the original sources are (most of the time) still available, nobody should think that upstreaming there is an option. One way of eliminating this patch is to get the sources from the linphone fork, which already has this.
* liboqs/Export-OQS_KEM_hqc_128-and-256_length -> This is complex OQS topic, where I don't think it makes sense for me to even speculate about upstreaming possibilities. One way of eliminating is to use the linphone fork, but that is actually lagging behind upstream for reasons. In any case if this is a blocker then I gladly downgrade to the fork.
* libyuv/FindTurboJpeg-and-CMake-integration -> Contributing to this google project is not exactly the easiest, and most of the devs there don't use any of the build systems that flatpak would support out of the box, which could further reduce, the priority of CMake patches. Again this can be downgraded to a fork.
* xerces-c/Fix-C++-minimum-level.patch -> This has already been upstreamed, just not released. If needed we can point to an unrealeased commit.

hfiguiere: > * Fix-use-of-function-deleted-in-SOCI-4.1.2-and-potential-stack-use-after-scope -> Obviously unacceptable upstream at this point, because upstream uses a much older version of SOCI where this doesn't apply. There is a completely different 3PP lifecycle there. Theoretically this could be eliminated with downgrading in the flatpak.

Then why do you not use the one that is compatible?

hfiguiere: for the other why don't you use the linphone fork? It seems that many patches are about building and using a different version

hfiguiere: > CMake-build-from-archive -> Unlikely to be accepted, because this is needed because of the flathub build system.

I don't see why this is needed either.

patches/sdk/external/libyuv/CMakeLists.txt
hfiguiere, line 1: so how does linphone build this is you claim it's no upstreamable?

Ferenc-: From a fork.

hfiguiere: Then why don't you use that fork? It seems that all the problem are related to use different versions, or ignoring the runtime.

org.linphone.linphone.yml
Ferenc-: > Then why do you not use the one that is compatible?

Every lib can have vulnerabilities, and this one also had a CVE before, future ones won't be fixed on older version but on top of latest.
If that is all cool with you, and the the rest of the blockers are also resolvable, then I will downgrade for sure.

Ferenc-: > for the other why don't you use the linphone fork? I

Same reason as above. And if the submission only depends on this then I will downgrade for sure.

Ferenc-: > I don't see why this is needed either.

Last time I have checked, the linphone GitLab instance was way too much rate limited, and unrealiable for flatpak to build from it, a single failed request out of hundreads is enough to make the build fail.
Current GitHub mirror, are exact mirrors meaning carrying references to the previously mentioned GitLab instance, so hardly a help. An alternative would be a fork, that doesn't have GitLab references, but that is probably more maintenance effort than this whole flatpak.

patches/sdk/external/libyuv/CMakeLists.txt
Ferenc-: Same as https://github.com/flathub/flathub/pull/9038#discussion_r3484333304, just with more CVEs.
Plus if the worry is maintenance, then forks are usually much less maintained, so after a while they will also need more patches just to compile with latest flatpak runtimes.
Anyhow, if it makes sense then it can be downgraded for sure.

hfiguiere: take that with upstream. They are the one making the poor choices.

PR #9221, opened 2026-07-05
Add org.notepadng.Notepadng

kravich:

> Notepad++-like text editor based on Scintilla editor component

https://github.com/user-attachments/assets/55da7807-22b7-4887-88cb-72d1679d1e4d

  **Link**: https://github.com/kravich/notepadng

### DIFF
--- a/org.notepadng.Notepadng.yaml
+++ b/org.notepadng.Notepadng.yaml
@@ -1,0 +1,56 @@
+id: org.notepadng.Notepadng
+runtime: org.kde.Platform
+runtime-version: 6.10
+sdk: org.kde.Sdk
+command: notepadng
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+cleanup:
+  - /include
+  - /lib/cmake
+  - /lib/pkgconfig
+  - /share/man
+
+modules:
+  - name: libqscintilla
+    buildsystem: qmake
+    subdir: src
+    cleanup:
+      - /mkspecs
+      - /qsci
+      - /lib/qt5
+      - /share/qt5/qsci
+    sources:
+      - type: archive
+        url: https://www.riverbankcomputing.com/static/Downloads/QScintilla/2.14.1/QScintilla_src-2.14.1.tar.gz
+        sha256: dfe13c6acc9d85dfcba76ccc8061e71a223957a6c02f3c343b30a9d43a4cdd4d
+      - type: patch
+        path: patches/qscintilla-lib-paths.patch
+
+  - name: uchardet
+    buildsystem: cmake-ninja
+    config-opts:
+      # uchardet declares compatibility with cmake 3.1, but newer runtime provides cmake 4.0+,
+      # which removed compatibility with eveything below cmake 3.5. Use compatibility option
+      - -DCMAKE_POLICY_VERSION_MINIMUM=3.5
+      - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+      - -DBUILD_BINARY=OFF
+      - -DBUILD_STATIC=OFF
+    sources:
+      - type: archive
+        url: https://www.freedesktop.org/software/uchardet/releases/uchardet-0.0.8.tar.xz
+        sha256: e97a60cfc00a1c147a674b097bb1422abd9fa78a2d9ce3f3fdcc2e78a34ac5f0
+
+  - name: notepadng
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+      - -DBUILD_VARIANT=Flatpak
+    sources:
+      - type: git
+        url: https://github.com/kravich/notepadng.git
+        tag: 0.1.0
+        commit: 6a8c03ef99d6adf52d4929af399d561a621d66d7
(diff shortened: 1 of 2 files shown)


kravich:
bot, build

org.notepadng.Notepadng.yaml
```diff
@@ -0,0 +1,60 @@
...
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+
+modules:
+  - name: libqscintilla
+    buildsystem: simple
+    build-commands:
+      - qmake src/qscintilla.pro
```
bbhtt: there is a qmake buildsystem instead of simple have you tried using that?

org.notepadng.Notepadng.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - make -j $FLATPAK_BUILDER_N_JOBS
+      - INSTALL_ROOT=./prefix make install
+      - mv ./prefix/usr/lib/x86_64-linux-gnu/* ./prefix/usr/lib && rm -r ./prefix/usr/lib/x86_64-linux-gnu
```
bbhtt: why is the libdir wrong? It is supposed to be `lib` not the arch triplet.

org.notepadng.Notepadng.yaml
```diff
@@ -0,0 +1,60 @@
...
+      # which removed compatibility with eveything below cmake 3.5. Use compatibility option
+      - -DCMAKE_POLICY_VERSION_MINIMUM=3.5
+      - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+      - -DBUILD_BINARY=OFF
+      - -DBUILD_STATIC=OFF
+    cleanup:
+      - /include
+      - /lib/cmake
+      - /lib/pkgconfig
+      - /share/man
```
bbhtt: move these to toplevel cleanup then you don't need to repeat.

flathub.json
```diff
@@ -0,0 +1,3 @@
+{
+  "skip-arches": ["aarch64"]
```
bbhtt: why disable aarch64?

CodedOre:
This is an fork of an application that is present on Flathub:

https://flathub.org/de/apps/com.notepadqq.Notepadqq

What does the fork changes from the application. Do note that a fork without substantial changes won't be allowed on Flathub under the duplicate submissions policy:

https://docs.flathub.org/docs/for-app-authors/requirements#duplicate-submissions

kravich:
> What does the fork changes from the application.

The problem with original Notepadqq is that it, being a desktop application, is based on Qt5 Webkit + CodeMirror web editing component, and with RPC between editor and UI. This adds a lot of complexity for no reason. Eventually, some update of Qt broke this fragile machinery, making it impossible to use, and Notepadqq still remains unusable since 2020, and original author had no time to fix this until recently.

My fork removes web-based editor component and all inevitable asynchrony, and integrates Scintilla editor component instead, which is the best editor library in my opinion and is a core of Notepad++ also. This gives a lot of Notepad++-like look and feel, and allows to use native N++ themes also, which was made intentionally.

Last month, the author of Notepadqq revived development of original project, but my fork was already implemented at this point, and I belive it provide value on its own, being more maintainable and close to what me and other people wait from Notepad++-like editor on Linux.

org.notepadng.Notepadng.yaml
kravich: This is actually hard.
When I use conventional
```
    buildsystem: qmake
    subdir: src
    make-install-args:
      - INSTALL_ROOT=/app
```
qscintilla installs its libraries to /app/usr/lib/x86_64-linux-gnu, which is a complete nonsense.
That's why I came up with this "manual" solution. I'll try to research this situation some more to understand if there is better solution.

flathub.json
kravich: The reason of failing aarch64 build is x86_64-linux-gnu triplet in 'simple' install instructions of  libqscintilla module. When I replace it with 'aarch64-linux-gnu', everything builds successfully for aarch64.

This is actually related to previous issue with simple buildsystem instead of qmake. If I find how to use qmake buildsystem, this may resolve naturally as no arch triplet will present in build instructions. Otherwise, this triplet should be parametrized somehow.

org.notepadng.Notepadng.yaml
hfiguiere: SonicPi doesn't do all of that

https://github.com/flathub/net.sonic_pi.SonicPi/blob/master/net.sonic_pi.SonicPi.json#L298

hfiguiere: it just patches qmake files.

hfiguiere: possibly there is a better way

kravich: Ahaa, so, somebody already solved this! Since other developers also can't install qscintilla without quirks, I believe there is no clean solution at all, and I like approach with a patch, since it's more declarative. I'll adopt SonicPi approach

kravich:
bot, build

org.notepadng.Notepadng.yaml
kravich: Fixed

org.notepadng.Notepadng.yaml
kravich: Fixed using the same approach as SonicPi

org.notepadng.Notepadng.yaml
kravich: Yep, now cleanup commands are deduplicated.
I've left some module-specific cleanup for libqscintilla module, and added some more paths to it to cleanup leftovers (mostly python-related)

flathub.json
kravich: After libqscintilla module build rework, there is no hardcoded arch triplet in paths, and aarch64 builds successfully on my machine

PR #9254, opened 2026-07-08
Add io.github.p5k369.grawji

p5k369:
**grawji** is a GTK4/libadwaita app that develops Fujifilm RAF files through the camera's own conversion engine over USB (PTP). The camera body renders every preview and export, so results are identical to Fujifilm's own X RAW Studio.

Upstream: https://github.com/p5k369/grawji (GPL-3.0-or-later, first release v0.2.0)

**Permission notes:**

- `--device=usb` + `--device=dri`: the camera is driven over raw USB (PTP vendor commands via libusb), the UI wants GPU rendering. `--require-version=1.16.0` is set for the usb device type.
- `--filesystem=host:ro`: **requesting a linter exception** for `finish-args-host-ro-filesystem-access`. grawji is a photo-library app with its own folder-tree browser (same workflow as darktable or digiKam). Users keep RAW files on SD cards, external drives, network mounts and arbitrary photo-library layouts, so portal-scoped file access does not fit the browsing workflow. Access is strictly read-only; all writes (JPEG exports) go through the file-chooser portal, which grants write access only to the user-chosen target.

The manifest builds offline (build backend vendored as pinned wheels, cleaned up after the build) and was test-built locally including screenshot mirroring; the appstream file passes `flatpak-builder-lint appstream`.

### DIFF
--- a/io.github.p5k369.grawji.yaml
+++ b/io.github.p5k369.grawji.yaml
@@ -1,0 +1,130 @@
+# Flatpak manifest for grawji.
+#
+# Build and install locally:
+#   flatpak-builder --user --install --force-clean build flatpak/io.github.p5k369.grawji.yaml
+# Run:
+#   flatpak run io.github.p5k369.grawji
+#
+# The GNOME runtime supplies GTK4, libadwaita and PyGObject. The USB stack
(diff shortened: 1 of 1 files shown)


p5k369:
bot, build

PR #9279, opened 2026-07-10
Add io.github.lucascouts.ClaudeDesktopFlatpak

lucascouts:
## io.github.lucascouts.ClaudeDesktopFlatpak

Unofficial packaging of **Claude Desktop**, Anthropic's desktop app for Claude.

- **Not a fork.** This ships Anthropic's *genuine, unmodified* Linux binary. Anthropic distributes it only as a `.deb` for Debian/Ubuntu; there is no upstream Flatpak, so this third-party submission does not duplicate an upstream-provided Flatpak.
- **No redistribution.** The proprietary binary is fetched from Anthropic's official server (`downloads.claude.ai`) at install time via an `extra-data` source — nothing proprietary is hosted here. A checker (`x-checker-data`) tracks new releases.
- **Unofficial note** is present in the metainfo `<description>`, as required for unofficially packaged proprietary apps.

### Requested `finish-args` exceptions
The linter flags two permissions that are core to the app's purpose (Claude Code sessions and MCP servers):

- `--filesystem=home` — Claude Code reads/writes the user's project files; portal-only access would break its core workflow.
- `--talk-name=org.freedesktop.Flatpak` — lets users launch local MCP servers on the host via `flatpak-spawn --host`. Remote MCP servers work without it.

Both are documented for users in the metainfo and README. Happy to narrow these if the reviewers prefer a more restricted default.

### Checklist
- [x] Built and installed locally with `org.flatpak.Builder`; app launches and signs in.
- [x] `flatpak-builder-lint manifest` / `repo` clean except the two exceptions above.
- [x] Screenshot hosted at a stable raw URL; AppStream validates.
- [x] `x86_64` and `aarch64` via `only-arches`.


hfiguiere:
This will not be accepted regardless.

PR #9289, opened 2026-07-11
Add io.github.recol.dlss_updater

Recol:

- [X] Please describe the application briefly. DLSS Updater scans the games installed on a Linux system (Steam/Proton, Heroic, Lutris, custom folders), detects the upscaling and frame-generation DLLs they ship (DLSS / XeSS / FSR / DirectStorage), and updates them with automatic backups and one-click restore. It also generates per-title Proton upscaler launch options. Upstream: https://github.com/Recol/DLSS-Updater (AGPL-3.0).

https://github.com/user-attachments/assets/21ff6ff7-9ed2-4767-b55e-ae3430841a91

- [X] The Flatpak ID follows all the rules listed in the [Application ID requirements][appid]. (Note: the linter's calculated repo URL check needs a manual exception — the upstream repository name contains a dash, `DLSS-Updater`, which cannot be represented in the ID; see notes below.)

---

**Build notes**
- Everything is built from source, offline: CPython 3.14 (free-threaded) from the python.org tarball, all Python dependencies as pinned wheels, and the Flet UI framework's Flutter client compiled from the flet-dev/flet tag (offline sources generated with [flatpak-flutter](https://github.com/TheAppgineer/flatpak-flutter); the input manifest `flatpak-flutter.yml` is included for regeneration).
- The Flutter client is trimmed (`patches/flet-client-trim.patch`) to remove upstream flet extensions this app does not use — notably the ones that ship prebuilt binaries or download artifacts mid-build (rive, media_kit).
- x86_64 only (`flathub.json`): the free-threaded cp314t wheel ecosystem does not consistently ship aarch64 manylinux wheels yet.
- The app source is pinned to the V4.3.1 release commit.

**Linter exceptions requested**
- `finish-args-home-filesystem-access` (plus `/mnt`, `/media`, `/run/media`): the app's core function is scanning user game libraries — which live in arbitrary user-chosen folders across drives — and patching DLL files in-place inside the game installations it finds. Portals are not practical for recursive discovery and in-place modification across many game directories.
- `appid-url-not-reachable`: the linter probes `github.com/recol/dlss_updater`; the actual repository is https://github.com/Recol/DLSS-Updater (dash in the name). I will verify the app via GitHub login after acceptance.


Recol:
Requesting PR to be reopened, doesn't appear to be a manual command.

CodedOre:
The original PR description and seemingly the manifest itself has been created by probabilistically automated means. 

This is in violation of the Generative AI policy: \

Recol:
@CodedOre No "automated means" i.e. LLM usage etc was used here, the build scripts are automated via a build pipeline, this does not violate any guidelines, and is required due to the underlying GUI framework having pulled in dependencies i.e. it being Flutter, please see: https://github.com/Recol/DLSS-Updater/blob/main/build_flathub.sh

CodedOre:
Right, considering that the project has existed for a bit, I'll entertain this discussion. \
Keep in mind that we receive a lot of slop created in just a few days recently, so we tend to be skeptical nowadays. 

1. The script you've linked: What does it do and why is it so elaborate? Using `flatpak-flutter` and some of the Flatpak Builder tools do not require running the Flatpak build steps manually?
2. The manifest in this PR seems to have no relation to that script. Where does this come from?
3. If you've done this submission yourself, why delete the template? It's there for a reason?

Adding to that Github reports that Claude has been a contributor. Have you used it and in which capacity?

Recol:
> Right, considering that the project has existed for a bit, I'll entertain this discussion. Keep in mind that we receive a lot of slop created in just a few days recently, so we tend to be skeptical nowadays.
> 
> 1. The script you've linked: What does it do and why is it so elaborate? Using `flatpak-flutter` and some of the Flatpak Builder tools do not require running the Flatpak build steps manually?
> 2. The manifest in this PR seems to have no relation to that script. Where does this come from?
> 3. If you've done this submission yourself, why delete the template? It's there for a reason?
> 
> Adding to that Github reports that Claude has been a contributor. Have you used it and in which capacity?

1: `build_flathub.sh` isn't part of the submission — it's just my local dev pipeline in the upstream repo (same idea as the [build_flatpak.sh](https://github.com/Recol/DLSS-Updater/blob/main/build_flatpak.sh) I've used for the self-hosted bundle for ages). It builds the Flet client on the host so I can iterate quickly instead of doing a full in-sandbox Flutter build every cycle. It's long mostly because the app runs on free-threaded Python 3.14, so I generate the offline wheel sources myself (stock flatpak-pip-generator targets the SDK's Python, which doesn't work here).
2. The manifest is normal `flatpak-flutter` output — generated from the `flatpak-flutter.yml` that's in the PR (the header of the generated file says so). No manual build steps in the submission itself. The only custom bit is the trim patch: the app doesn't use any of Flet's optional extensions, so the client is cut down to core — which conveniently also drops `rive_native` (prebuilt .so's) and `media_kit` (downloads libmpv mid-build), neither of which would fly on your builders anyway.
3. That was my mistake, i wrote my own description for the first submission as i didn't realise the checker wants the literal template, this is because of the number of build caveats that are needed here, you have my apologies.

Claude was only used as part of a test branch which was eventually deleted entirely to test out it's capabilities, none of that made it into any sort of upstream code/documentation/etc, hope that makes sense.

flatpak-flutter.yml
CodedOre, line 1: Quite a few of the comments in this manifest are verbose. You could reduce them.

flatpak-flutter.yml
```diff
@@ -0,0 +1,95 @@
...
+  - --share=ipc
+  - --share=network
+  # Game libraries live anywhere the user put them; the app scans and
+  # patches DLLs inside them. home:rw needs a linter exception at review
+  # (finish-args-home-filesystem-access) - it also covers the app's own
+  # config/cache dirs, which is why they aren't listed separately.
+  - --filesystem=home:rw
+  - --filesystem=/mnt:rw
+  - --filesystem=/media:rw
+  - --filesystem=/run/media:rw
```
CodedOre, line 30: Access with read/write is the default. 

As for the linter, see the [documentation](https://docs.flathub.org/docs/for-app-authors/linter#exceptions).

flatpak-flutter.yml
```diff
@@ -0,0 +1,95 @@
...
+  - --filesystem=~/.local/share/dlss-updater:create
```
CodedOre, line 31: No.

Flatpak applications should store their internal data in the sandboxed folders by following the XDG base directory specification.

flatpak-flutter.yml
```diff
@@ -0,0 +1,95 @@
...
+  - --env=LD_LIBRARY_PATH=/app/lib64:/app/lib
```
CodedOre, line 32: If you're using `/app/lib64`, you're doing something wrong. Libraries should only reside in `/app/lib`.

Use the official Flatpak build of Flatpak Builder, as detailed in the submission guide. With that, libraries installed by understood buildsystems with definitely land in `/app/lib`.

flatpak-flutter.yml
```diff
@@ -0,0 +1,95 @@
...
+      - cp main.py /app/share/dlss-updater/main.py
+      - install -Dm755 launcher.sh /app/bin/dlss-updater
+      - |
+        mkdir -p /app/share/metainfo
+        sed 's/io\.github\.recol\.dlss-updater/io.github.recol.dlss_updater/g' \
+            io.github.recol.dlss-updater.appdata.xml \
+            > /app/share/metainfo/io.github.recol.dlss_updater.metainfo.xml
+      - |
+        sed -e 's/\r$//' -e 's/io\.github\.recol\.dlss-updater/io.github.recol.dlss_updater/g' \
+            flatpak/io.github.recol.dlss-updater.desktop > /tmp/desktop.tmp
```
CodedOre, line 85: Do not use `sed` in a Flatpak manifest, its not failsafe. 

Ideally fix this upstream, otherwise use a patch or the rename options in the manifest.

modules/cpython.json
CodedOre, line 1: This should be part of the main manifest file.

modules/cpython.json
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "cpython-freethreaded",
+    "buildsystem": "autotools",
```
CodedOre, line 3: Why the need to rebuild python? 

Why can't the version in the runtime be used?

CodedOre:
Alright, I reopen it for now.

flatpak-flutter.yml
Recol: Alright, the reason this is here is to ensure continuity with the bundled flatpak install, otherwise it means Flathub installs start with a fresh config/db, i'll strip this out.

modules/cpython.json
Recol: The runtime ships GIL-locked Python `3.12`/`3.13`; the app is architected on free-threaded CPython 3.14 (3.14.3 to be exact) with the GIL being unlocked, its scanner/updater do true parallel multi-threaded work with the GIL disabled, all its native wheels are cp314t, and it's the same interpreter shipped on every platform, there is no way in which this can be done which doesn't compromise concurrency and need a rewrite.

Recol:
Alright well not sure how to proceed here, a bit of a shame, i understand your concerns about AI slop, but that doesn't mean everything is, i'll shutdown future work on this, clearly not worth my time.

hfiguiere:
github wouldn't lie, would it?

image

(non withstanding the PR text really has all the characteristic of AI slop, including asking for exceptions that if it did understand the documentation it wouldn't request)

barthalion:
@Recol You seem to have deleted the repo so I can't re-open it. I find this project useful and existing for long enough to grant it an exception. Please open a new PR.

PR #9290, opened 2026-07-11
Add io.github._66HEX.Frame

66HEX:

- [x] Please describe the application briefly.  Frame is a FFmpeg GUI with controls over image/video/audio/subtitles/filters/metadata settings. Frame has built-in preview panel so users have better control over output
- [x] Please attach a video showcasing the application on Linux using the Flatpak. [< Please insert the video here >](https://github.com/user-attachments/assets/e3791f97-5758-42a4-a4cb-a97d920b0c78)

### DIFF
--- a/io.github._66HEX.Frame.yml
+++ b/io.github._66HEX.Frame.yml
@@ -12,7 +12,6 @@ finish-args:
   - --socket=pulseaudio
   - --device=dri
   - --talk-name=org.freedesktop.Notifications
-  - --env=FRAME_USE_SYSTEM_MEDIA_TOOLS=1
 cleanup:
   - /include
   - /lib/pkgconfig
@@ -30,13 +29,22 @@ modules:
       - mkdir -p .cargo
       - cp cargo-vendor/cargo-config.toml .cargo/config.toml
       - cargo build --release --locked --offline --manifest-path frame-app/Cargo.toml
-      - install -Dm755 target/release/frame /app/bin/frame
+      - install -Dm755 target/release/frame /app/lib/frame/frame
       - |
-        sed \
-          -e 's|$APP_NAME|Frame|g' \
-          -e 's|$APP_CLI|frame|g' \
-          -e 's|$APP_ICON|io.github._66HEX.Frame|g' \
-          frame-app/resources/frame.desktop.in > io.github._66HEX.Frame.desktop
+        install -d /app/bin
+        printf '%s\n' \
+          '#!/bin/sh' \
+          'export FRAME_USE_SYSTEM_MEDIA_TOOLS=1' \
+          'exec /app/lib/frame/frame "$@"' \
+          > /app/bin/frame
+        chmod +x /app/bin/frame
+      - cp frame-app/resources/frame.desktop.in io.github._66HEX.Frame.desktop
+      - desktop-file-edit
+        --set-name=Frame
+        --set-key=Exec --set-value=frame
+        --set-key=TryExec --set-value=frame
+        --set-icon=io.github._66HEX.Frame
+        io.github._66HEX.Frame.desktop
       - install -Dm644 io.github._66HEX.Frame.desktop /app/share/applications/io.github._66HEX.Frame.desktop
       - |
         sed \
@@ -49,12 +57,10 @@ modules:
       - install -Dm644 frame-app/resources/app-icons/128x128.png /app/share/icons/hicolor/128x128/apps/io.github._66HEX.Frame.png
       - install -Dm644 frame-app/resources/app-icons/128x128@2x.png /app/share/icons/hicolor/256x256/apps/io.github._66HEX.Frame.png
       - install -Dm644 frame-app/resources/app-icons/icon.png /app/share/icons/hicolor/512x512/apps/io.github._66HEX.Frame.png
-      - install -Dm644 LICENSE /app/share/licenses/frame/LICENSE
     sources:
       - type: archive
         url: 'https://github.com/66HEX/frame/releases/download/0.31.0/frame-0.31.0-source.tar.gz'
         sha256: '1f2c8015bc17b978bb3f6d2bd32c33aefb08940013f4a4634a73b69c1532b3b4'
-        strip-components: 1
       - type: archive
         url: 'https://github.com/66HEX/frame/releases/download/0.31.0/frame-0.31.0-cargo-vendor.tar.gz'
         sha256: '8ee652053a8cca22c605d1dabc39a9ca6bc460e8c7473c5505f8b4030c812769'
--- a/io.github._66HEX.Frame.yml
+++ b/io.github._66HEX.Frame.yml
@@ -11,7 +11,7 @@ finish-args:
   - --socket=fallback-x11
   - --socket=pulseaudio
   - --device=dri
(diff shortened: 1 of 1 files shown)


io.github._66HEX.Frame.yml
```diff
@@ -0,0 +1,61 @@
...
+        CARGO_HOME: /run/build/frame/cargo
+        CARGO_NET_OFFLINE: 'true'
+        FRAME_USE_SYSTEM_MEDIA_TOOLS: '1'
+    build-commands:
+      - mkdir -p .cargo
+      - cp cargo-vendor/cargo-config.toml .cargo/config.toml
+      - cargo build --release --locked --offline --manifest-path frame-app/Cargo.toml
+      - install -Dm755 target/release/frame /app/bin/frame
+      - |
+        sed \
```
hfiguiere: Never use sed to change desktop files. It has no failure mode. Use the command `desktop-file-edit`.

hfiguiere: a `.in` is supposed to be processed by the build system, not by an obscure command.

You should have a proper build system.

io.github._66HEX.Frame.yml
```diff
@@ -0,0 +1,61 @@
...
+command: frame
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+finish-args:
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
+  - --device=dri
+  - --talk-name=org.freedesktop.Notifications
```
teamcons: Why do you need this?

io.github._66HEX.Frame.yml
```diff
@@ -0,0 +1,61 @@
...
+  - --env=FRAME_USE_SYSTEM_MEDIA_TOOLS=1
```
teamcons: Why do you need this? It doesnt even make sense as a finish-args

io.github._66HEX.Frame.yml
```diff
@@ -0,0 +1,61 @@
...
+          -e "s|@FRAME_FLATHUB_VERSION@|0.31.0|g" \
+          -e "s|@FRAME_FLATHUB_RELEASE_DATE@|2026-07-11|g" \
+          packaging/flathub/io.github._66HEX.Frame.metainfo.xml.in > io.github._66HEX.Frame.metainfo.xml
+      - install -Dm644 io.github._66HEX.Frame.metainfo.xml /app/share/metainfo/io.github._66HEX.Frame.metainfo.xml
+      - install -Dm644 frame-app/resources/app-icons/32x32.png /app/share/icons/hicolor/32x32/apps/io.github._66HEX.Frame.png
+      - install -Dm644 frame-app/resources/app-icons/64x64.png /app/share/icons/hicolor/64x64/apps/io.github._66HEX.Frame.png
+      - install -Dm644 frame-app/resources/app-icons/128x128.png /app/share/icons/hicolor/128x128/apps/io.github._66HEX.Frame.png
+      - install -Dm644 frame-app/resources/app-icons/128x128@2x.png /app/share/icons/hicolor/256x256/apps/io.github._66HEX.Frame.png
+      - install -Dm644 frame-app/resources/app-icons/icon.png /app/share/icons/hicolor/512x512/apps/io.github._66HEX.Frame.png
+      - install -Dm644 LICENSE /app/share/licenses/frame/LICENSE
```
hfiguiere:
```suggestion
```

io.github._66HEX.Frame.yml
```diff
@@ -0,0 +1,61 @@
...
+    sources:
+      - type: archive
+        url: 'https://github.com/66HEX/frame/releases/download/0.31.0/frame-0.31.0-source.tar.gz'
+        sha256: '1f2c8015bc17b978bb3f6d2bd32c33aefb08940013f4a4634a73b69c1532b3b4'
+        strip-components: 1
```
hfiguiere:
```suggestion
```

io.github._66HEX.Frame.yml
66HEX: Frame uses native desktop notifications to notify when an active conversion batch finishes. This goes through notify-rust / org.freedesktop.Notifications

io.github._66HEX.Frame.yml
66HEX: Frame uses `FRAME_USE_SYSTEM_MEDIA_TOOLS=1` to force `ffmpeg`/`ffprobe` resolution through the Flatpak runtime instead of looking for bundled release binaries. I moved it from `finish-args` into the `/app/bin/frame` wrapper, so it is now only part of Frame's launch environment

66HEX:
bot, build

io.github._66HEX.Frame.yml
teamcons: Yeah, but dont portals handle it? You dont need the permission if portals

66HEX: I checked this in the Flatpak build. Frame currently uses the notify-rust crate for desktop notifications, and that crate talks directly to org.freedesktop.Notifications rather than using the notification portal

If I remove --talk-name=org.freedesktop.Notifications, conversion-finished notifications fail with org.freedesktop.DBus.Error.ServiceUnknown, so I cannot simply drop this permission with the current implementation

If using the portal is required for Flathub, I can rework the notification backend to use org.freedesktop.portal.Notification, but that would be a larger refactor of the notification system rather than a manifest-only change

bbhtt: If you can, yes that would be better.

io.github._66HEX.Frame.yml
bbhtt: No this was fine in finish-args if you need it. If this is the only reason for the wrapper, you should move it back.

io.github._66HEX.Frame.yml
```diff
@@ -0,0 +1,67 @@
...
+        --set-key=Exec --set-value=frame
+        --set-key=TryExec --set-value=frame
+        --set-icon=io.github._66HEX.Frame
+        io.github._66HEX.Frame.desktop
+      - install -Dm644 io.github._66HEX.Frame.desktop /app/share/applications/io.github._66HEX.Frame.desktop
+      - |
+        sed \
+          -e "s|@FRAME_FLATHUB_VERSION@|0.31.0|g" \
+          -e "s|@FRAME_FLATHUB_RELEASE_DATE@|2026-07-11|g" \
+          packaging/flathub/io.github._66HEX.Frame.metainfo.xml.in > io.github._66HEX.Frame.metainfo.xml
```
bbhtt: this should be fixed upstream. either fill up when tagging or have a buildsystem fill it up, not sed.

66HEX: fixed

io.github._66HEX.Frame.yml
66HEX: Sure, I reverted my approach

io.github._66HEX.Frame.yml
66HEX: I reworked the notification system; starting with version 0.31.1, Flatpak distributions use portals for notifications ;)

66HEX:
bot, build

PR #5248, opened 2024-05-16
Add org.freedesktop.Sdk.Extension.fish

mdmundo:
## fish

_The user-friendly command line shell as a freedesktop SDK extension_

### Please confirm your submission meets all the criteria

- [x] Please describe your application briefly.
- [x] I have read the [App Requirements][reqs] and [App Maintenance][maint] pages.
- [x] My pull request follows the instructions at [App Submission][submission].
- [x] I have [built][build] and tested the submission locally.
- [x] I am using only the minimal set of permissions. *(If not, please explain each non-standard permission.)*
- [x] All assets referenced in the manifest are redistributable by any party.  If not, the unredistributable parts are using an extra-data source type.
- [x] I am an author/developer/upstream contributor of the project. If not, I contacted upstream developers about submitting their software to Flathub. **[Link](https://github.com/fish-shell/fish-shell/issues/9432#issuecomment-2115845544)** 
- [x] Any additional patches or files have been submitted to the upstream projects concerned. *(If not, explain why.):* _no patches_ 
- [x] The domain used for the application ID is controlled by the application developers either directly or through the code hosting (e.g. GitHub, GitLab, SourceForge, etc.). The [application id guidelines][app-id] are followed. _do not apply, it is an SDK extension._

[reqs]: https://docs.flathub.org/docs/for-app-authors/requirements
[maint]: https://docs.flathub.org/docs/for-app-authors/maintanance
[submission]: https://docs.flathub.org/docs/for-app-authors/submission
[build]: https://docs.flathub.org/docs/for-app-authors/submission/#before-submission
[app-id]: https://docs.flathub.org/docs/for-app-authors/requirements#application-id

PR #9307, opened 2026-07-12
Add org.gpxsee.GPXSee

joklaps:

- [x] Please describe the application briefly. GPXSee is a GPS log file viewer and analyzer that supports all common GPS log file formats.
- [x] Please attach a video showcasing the application on Linux using the Flatpak. https://github.com/user-attachments/assets/558ad83f-7d67-414a-8287-8cad30bdc3cb
- [x] I contacted upstream developers about this submission. **Link:** https://github.com/tumic0/GPXSee/issues/143#issuecomment-4947361212


joklaps:
The video was / is present.  Automatic validation failed on the video.

teamcons:
@CodedOre this one has been closed despite video. Do you think you could reopen it please?

org.gpxsee.GPXSee.yml
```diff
@@ -0,0 +1,32 @@
+app-id: org.gpxsee.GPXSee
+runtime: org.kde.Platform
+runtime-version: "6.10"
```
teamcons, line 3: 6.11 is last

org.gpxsee.GPXSee.yml
```diff
@@ -0,0 +1,32 @@
...
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=cups
+  - --share=network
+  - --filesystem=home
+rename-desktop-file: gpxsee.desktop
+rename-icon: gpxsee
+rename-mime-file: gpxsee.xml
+rename-appdata-file: gpxsee.appdata.xml
```
teamcons, line 15: Why do you need al the renames? 

The buildsystem should handle installing the correct files in the correct places, not the manifest

org.gpxsee.GPXSee.yml
```diff
@@ -0,0 +1,32 @@
...
+sdk: org.kde.Sdk
+command: gpxsee
+finish-args:
+  - --device=dri
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=cups
+  - --share=network
+  - --filesystem=home
```
teamcons: Why do you need the three last permissions exactly ? Home is very wide reaching, cups you can use the portal, and im not sure why network ? ?

hfiguiere:
I don't know why the bot cloesd, but don't edit the description.

org.gpxsee.GPXSee.yml
```diff
@@ -0,0 +1,32 @@
...
+rename-desktop-file: gpxsee.desktop
+rename-icon: gpxsee
+rename-mime-file: gpxsee.xml
+rename-appdata-file: gpxsee.appdata.xml
+modules:
+  - name: gpxsee
+    buildsystem: qmake
+    post-install:
+      - lrelease gpxsee.pro
+      - install -Dm644 metainfo/org.gpxsee.GPXSee.metainfo.xml /app/share/metainfo/gpxsee.appdata.xml
```
hfiguiere: why do you insstall it with the wrong name? Why isn't it installed by the build system?

org.gpxsee.GPXSee.yml
```diff
@@ -0,0 +1,32 @@
...
+      - sed -i 's/^Icon=gpxsee$/Icon=org.gpxsee.GPXSee/' /app/share/applications/gpxsee.desktop
```
hfiguiere: Never use sed to change desktop files. It has no failure mode. Use the command `desktop-file-edit`.

Also `rename-icon` fix that too. That's what it is designed for.

org.gpxsee.GPXSee.yml
joklaps: That's correct. I tried building against 6.11, but this version has an ssl error. It is logged and I will keep an eye on it, if it gets solved I will upgrade: https://invent.kde.org/packaging/flatpak-kde-runtime/-/work_items/91

GPXSee needs to download map tiles over https call, and the handshake fails because of this issue.

org.gpxsee.GPXSee.yml
joklaps: The reason I use the renames is this is a community build of an upstream application. I try to reuse as much as possible of the files of the upstream project. The author of the upstream allows to build the flatpak package, but will not support it in the upstream project. So I try to work around it as best as possible. I double checked and remove as much as possible.

org.gpxsee.GPXSee.yml
joklaps: Thank you for pointing this out. 

- I have removed the cups socket. I tested it and it is handled by the portal.
- The network is needed, the application loads map tiles over the internet and needs to resolve hostnames
- The filesystem home is because the application interacts with a lot of files the user has, created by other applications, or downloaded from the internet to show mapping data.

org.gpxsee.GPXSee.yml
joklaps: I fixed this, it has better logic now.

org.gpxsee.GPXSee.yml
joklaps: I worked around it to not use sed anymore. Indeed, the rename-icon already did that. Thank you.

org.gpxsee.GPXSee.yml
```diff
@@ -0,0 +1,30 @@
...
+    post-install:
+      - lrelease gpxsee.pro
+      - rm -f /app/share/metainfo/gpxsee.appdata.xml
+      - install -Dm644 metainfo/org.gpxsee.GPXSee.metainfo.xml /app/share/metainfo/org.gpxsee.GPXSee.metainfo.xml
+    sources:
+      - type: archive
+        url: https://github.com/tumic0/GPXSee/archive/refs/tags/16.11.tar.gz
+        sha256: 1542387d670d870829e1d52d66f7ea3ac6628a05fcaa39a78963ab7a18ebc70b
+      - type: git
+        url: https://github.com/joklaps/GPXSee-flathub
```
hfiguiere: why not using this https://github.com/tumic0/GPXSee/blob/16.11/pkg/linux/gpxsee.appdata.xml ?

joklaps: There are 2 issues I couldn't find a solution for when using the original gpxsee.appdata.xml
- The upstream author wants a clear disclaimer that he will not support issues with flatpak installs (that is upto the community that builds this package). So I added this in bold to the &lt;description&gt; tag.
- The &lt;releases&gt; information is not in the original appdata file. So I added it.

This is the file I added to support the build and will keep up-to-date with future releases: https://github.com/joklaps/GPXSee-flathub/blob/main/org.gpxsee.GPXSee.metainfo.xml

hfiguiere: this is what patches are for.

joklaps: Ok, thanks. I'll look into it. Where do I host the patch file(s). Also future versions. Are they allowed to be in the flatpak repository?

joklaps: Patch for appdata / metainfo is implemented

org.gpxsee.GPXSee.yml
bbhtt: > The filesystem home is because the application interacts with a lot of files the user has, created by other applications, or downloaded from the internet to show mapping data.

What kind of files and how does this work?

gpxsee-appdata.patch
bbhtt, line 1: was this submitted upstream?

gpxsee-appdata.patch
```diff
@@ -0,0 +1,252 @@
...
++        <keyword>map</keyword>
++        <keyword>maps</keyword>
++        <keyword>gps</keyword>
++    </keywords>
++
++    <url type="homepage">https://www.gpxsee.org</url>
++    <url type="contribute">https://github.com/tumic0/GPXSee</url>
++    <url type="translate">https://hosted.weblate.org/projects/gpxsee/translations/</url>
++
++  <content_rating type="oars-1.1">
```
bbhtt: this is just `<content_rating type="oars-1.1" />`

org.gpxsee.GPXSee.yml
joklaps: GPXSee is primarily a viewer for GPS data, maps, and related geospatial files. It needs filesystem access because users typically open files that already exist on their computer, rather than creating them inside the application.

Examples of the files GPXSee commonly uses include:
- gpx (GPS Exchange Format)
- fit (Garmin FIT)
- tcx (Training Center XML)
- kml / kmz (Google Earth)

gpxsee-appdata.patch
joklaps: No, the patch cannot be submitted upstream. This is a community build of the project, the original author allows it but doesn't support it in the upstream project.

org.gpxsee.GPXSee.yml
CodedOre: But why is the FileChooser portal not used instead?

Using the portal would allow you to get files the user selects without a filesystem permission.

gpxsee-appdata.patch
joklaps: I updated this as suggested.

org.gpxsee.GPXSee.yml
joklaps: I did some more tests. Without the file home permission the export functionality isn't working as the end user expects perhaps? I see these path containing "/run/user/1000/doc/hEhaRwM_hEtMtlgHN7_5Rw". Also recent files appear on these paths. 
image

joklaps:
About the build validation. I know the gpxsee.org domain doesn't validate with the ssl check. I know www.gpxsee.org does.  I didn't add the "www" as it is not part of the domain, and I checked a list of other flatpaks, nobody adds the www subdomain. So I decided not to add the www when submitting the application id. Please advice how to continue? Can somebody please manually validate the domain?

_The domain must be directly related to the project or the application being submitted and the author or the developer or the project must have control over the domain. The corresponding URL must be reachable over HTTPS. In some edge cases, if it is not reachable, an exception can be granted after manually checking._

org.gpxsee.GPXSee.yml
CodedOre: The `/run/user/$UID/doc/$DOCID` path is expected. That is where the Documents portal mounts the files and folders to which access were granted for. \
There exists an API to get the host path from that ([`org.freedesktop.portal.Documents.GetHostPaths`](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Documents.html#org-freedesktop-portal-documents-gethostpaths)) if you wish to display it. But file operations must use the path as coming from the portal. \
If you take the path given to you without modifying it, it will work with the file on the host.

If opening and saving files works with the portals and the only issue is that it shows the path from the Documents portal, then there is no need for a filesystem permission.

joklaps: I understand your reasoning, and from a technical /security perspective it makes sense. My concern is more about the end-user experience.

- This is a community build. The upstream author is not interested in making Flatpak-specific changes, so using org.freedesktop.portal.Documents.GetHostPaths to translate the displayed paths is not something that will not be implemented. I try to work with this restriction to still give the best possible experience to the end-user.
- I will remove the home filesystem permission, but I'm concerned that the target audience for this application is not always very technical. Seeing paths like /run/user/$UID/doc/... instead of their actual file locations is likely to be confusing, and I expect it will result in a number of support requests and complaints.

bbhtt:
Have you contacted upstream about the domain issue? It's something they should fix.

gpxsee-appdata.patch
bbhtt: Why don't they want the patch?

org.gpxsee.GPXSee.yml
bbhtt: > The upstream author is not interested in making Flatpak-specific changes

Where did they say this?

joklaps: As link in the original PR comment. https://github.com/tumic0/GPXSee/issues/143

This issue is a few years old. There is a request for a flatpak, author will not do it. Author states he will not support it in the upstream project the last comment.

gpxsee-appdata.patch
joklaps: As link in the original PR comment. Please read this issue: https://github.com/tumic0/GPXSee/issues/143

This issue is a few years old. There is a request for a flatpak, author will not do it. Author states he will not support it in the upstream project the last comment. I'm trying to do it as a community effort.

hfiguiere: if only you hadn't reindented the whole file it would be easy to see the changes. Please make this happen. **A patch is really meant to be minimalist.**

joklaps: Updated as requested

gpxsee-appdata.patch
```diff
@@ -0,0 +1,147 @@
...
+ 
+@@ -43,88 +48,52 @@
+ 	<screenshots>
+ 		<screenshot type="default">
+ 			<image>https://www.gpxsee.org/gallery/linux.png</image>
++			<caption>Main window showing a track on the map</caption>
+ 		</screenshot>
+ 		<screenshot>
+ 			<image>https://www.gpxsee.org/gallery/gear_ratio.png</image>
++			<caption>Gear ratio graph</caption>
```
hfiguiere, line 23: these two caption should be sumitted upstream

gpxsee-appdata.patch
```diff
@@ -0,0 +1,147 @@
...
+ 
+-	<categories>
+-		<category>Graphics</category>
+-		<category>Viewer</category>
+-		<category>Maps</category>
+-		<category>Geoscience</category>
+-		<category>Geography</category>
+-		<category>DataVisualization</category>
+-	</categories>
++	<launchable type="desktop-id">gpxsee</launchable>
```
hfiguiere: no this should be
```suggestion
+	<launchable type="desktop-id">gpxsee.desktop</launchable>
```

gpxsee-appdata.patch
```diff
@@ -0,0 +1,147 @@
...
+ 
+-	<url type="homepage">https://www.gpxsee.org</url>
++	<icon type="stock">org.gpxsee.GPXSee</icon>
```
hfiguiere:
```suggestion
```

you never need this

gpxsee-appdata.patch
joklaps: I understand, but the upstream author stated they will not support flatpak builds. https://github.com/tumic0/GPXSee/issues/143  This is a community build, allowed but not supported by the upstream.

gpxsee-appdata.patch
```diff
@@ -0,0 +1,147 @@
...
++		<url type="details">https://github.com/tumic0/GPXSee/releases/tag/16.11</url>
++		<description>
++		<p>Changes</p>
++			<ul>
++			<li>Removed Qt-private dependency.</li>
++			<li>Improved KMZ files loading.</li>
++			</ul>
++		</description>
++	</release>
++</releases>
```
hfiguiere: also these changes should submitted. if upstream doesn't want them that mean they don't even want to support linux.

gpxsee-appdata.patch
```diff
@@ -0,0 +1,147 @@
...
+@@ -1,12 +1,17 @@
+-<?xml version="1.0" encoding="UTF-8"?>
+-<component type="desktop">
+-	<id>gpxsee.desktop</id>
++<?xml version="1.0" encoding="UTF-8" ?>
++<component type="desktop-application">
++	<id>org.gpxsee.GPXSee</id>
++	<developer id="com.github.tumic0">
++		<name>Martin Tůma</name>
++	</developer>
```
hfiguiere, line 11: developer tag should be upstream.

Also `com.github.tumic0` isn't valid. it should at worst `io.gihub...` or use the same domain as the app-id

gpxsee-appdata.patch
hfiguiere: What is unclear about this **not being** flatpak specific???

org.gpxsee.GPXSee.yml
tumic0: > > The upstream author is not interested in making Flatpak-specific changes
> 
> Where did they say this?

In case any one would doubt tumic0/GPXSee#143 says exactly this, I will say it here and now - Yes, I will not add a single line of code to GPXSee for Flatpak. I already did that for another crippled Linux (called Android) and I regret it every day I look into the source code that is now full of obscure workarounds "solving" the disability to work properly with the filesystem.

GPXSee is designed to work extensively with the filesystem. It does/may access a lot more of files (CRS files, styles, maps, POIs.... where all of them may be symlinks to some common storage for multiple apps) than the user selects in the file dialogs and all of that will be broken or at least limited when used in a Flatpak. I'm even convinced that the essential functionality of opening a file and then browsing other files in that directory will not work (like it does not work on Android).

Due to the GPL3 license used I can't ban anyone from making a Flatpak of GPXSee or even distribute it on Flathub, but I can and will discourage anyone from using such a crippled version of the application.

gpxsee-appdata.patch
tumic0: > if upstream doesn't want them that mean they don't even want to support linux.

How did you get to such conclusion? Linux is the primary platform for GPXSee. Not being interested in any "age rating system" crap is far from not being interested in Linux...

hfiguiere: appstream is not a flatpak/flathub thing. Distros use it too for their packages. They just don't enforce it.

But hey if it's crap we can just close this PR and move on.

tumic0: The current appstream XML file as provided in the GPXSee git repository is fine for a dozen of Linux and BSD distributions (Arch, Fedora, Gentoo, OpenSUSE, OpenBSD, ...).

If there are changes that make sense/improve the appstream info for "regular" distributions we may discuss them, and I may accept them. But precisely because of the file is used by __all__ regular distributions, I won't make any random changes to it just because some Flatpak people want them...

> we can just close this PR and move on.

That would be the best result for me, but I doubt your users (especially the PR author) would be that happy with that...

tumic0:
> About the build validation. I know the gpxsee.org domain doesn't validate with the ssl check

This should be fixed now.

joklaps:
bot, build

joklaps:
bot, build

gpxsee-appdata.patch
joklaps: I made all the requested changes in the patch file. I believe all of the reviewers' comments have been addressed by applying this patch. However, I'm not in a position to determine which of these changes can safely be incorporated into the upstream appstream file.

gpxsee-appdata.patch
joklaps: This was changed to io.github

bbhtt:
Seems like upstream has strong onions on Flatpak packaging 

> Flatpack is a hack/workaround for non-opensource applications where the authors are not willing/able to prepare standard packages or a workaround for bureaucratic distributions like Debian, where adding a package takes several years.

and clearly doesn't want to support it

> Building a Flatpak package for yourself and distribute it on FlatHub are two completely different things. While I have no objections against the first, in the second case I really would like that there is clearly stated on FlatHub that this package is not "official" and people should not come here for any support.

I don't think we would want to publish such an app even as a community build. This will only lead to hostility and headaches on both sides later on.

joklaps:
Well, this was quite a disappointing first experience. As someone willing to put effort into supporting Flathub with a well-known package, I expected a more constructive process. Every issue that was raised was addressed, either by making the requested changes or fixing the underlying problems, and the patch file took care of the AppStream XML.

hfiguiere:
@bbhtt there are plenty of open source software packaged here with hostile upstream. Don't be that person.

bbhtt:
Just because there is doesn't make it the right thing to do now. It's not "our" software.

They are pretty clear that they don't want to support any issues arising from Flatpak package, and clearly make it unofficial apart from using strong tones in the upstream issue and coming here to call requested changes "random".

So what happens when Flatpak specific bugs arise or changes that are needed to be made? A packager without upstream collaboration would lead to the eventual decay of this Flathub listing. That would be a disservice to Flathub users and would leave them without recourse.

I would like to avoid yet another drama from this. There has been enough of those. So I will absolutely be that person, whatever that means.

hfiguiere:
We can decide to EOL the package if it is too complicated to fix. Given the number of proprietary software that can NEVER be fixed that we accept, this is a very low risk in comparison. 

It's not even a pile a fragile javascript on top of the Electron bloat, a pile of Flutter whose design is suited for Windows 95 or Java.

Really.

bbhtt:
It's a case where they are explicit about their wishes. I wouldn't have accepted anything if I knew upstream had reservations about it beforehand. In fact we rejected certain things multiple times before when upstream said they don't want it on Flathub. See for example fish extension or whatever https://github.com/flathub/flathub/pull/5248#issuecomment-2115934840

I try to avoid the distro game of packaging whatever found in the sand and it is also a way of respecting their wishes to me.

gpxsee-appdata.patch
```diff
@@ -0,0 +1,142 @@
...
++	<developer id="io.github.tumic0">
++		<name>Martin Tůma</name>
++	</developer>
+ 	<metadata_license>MIT</metadata_license>
+ 	<project_license>GPL-3.0</project_license>
+ 
+ 	<name>GPXSee</name>
+ 	<summary>GPS log file viewer and analyzer</summary>
+ 	<description>
++		<p>**This is a community-built package of GPXSee. It is not an official package and is not supported by the author of GPXSee. Please do not contact the GPXSee author for support regarding this Flatpak package.**</p>
```
barthalion: This is GPL licenced software. This disclaimer doesn't belong here.

barthalion:
In the end, it's released under the GPL3 licence, so I would say fuck it, let's publish it. I assume joklaps will continue maintaining it as it seems fairly uncomplicated given all dependencies come from the runtime.

bbhtt:
You cannot in good conscience ignore the social contracts involved as it is not just about the licence alone.

What exactly is the point of asking in the template to contact upstream, having a verification system or being upstream-first, then if we are going to ignore them altogether when time comes? 

Especially in this case where it is an open source piece of software, as far as I can see in large parts, maintained by one person and they have bern very explicit even in the review comments here.

This laissez-faire approach is not correct.

bbhtt:
> We can decide to EOL

This doesn't solve the problem that people will be stuck on whatever last version we had and eventually that will lead to bad reviews, support requests or whatever when they encounter issues unfixed or features unavailable in Flathub vs. fixed elsewhere

As I said, not having it solves multiple classes of problems on both ends.

gpxsee-appdata.patch
```diff
@@ -0,0 +1,142 @@
+--- GPXSee-16.11/pkg/linux/gpxsee.appdata.xml	2026-07-14 20:37:22.882125124 +0200
++++ GPXSee-16.11/pkg/linux/gpxsee.appdata.xml	2026-07-14 20:37:23.123124507 +0200
+@@ -1,12 +1,16 @@
+-<?xml version="1.0" encoding="UTF-8"?>
+-<component type="desktop">
+-	<id>gpxsee.desktop</id>
++<?xml version="1.0" encoding="UTF-8" ?>
++<component type="desktop-application">
++	<id>org.gpxsee.GPXSee</id>
```
hfiguiere: remove this app-id change. `rename-appdata-file` with do the right thing and it upsets the upstream developer for no reason.

joklaps:
bot, build

gpxsee-appdata.patch
barthalion: The frontend now carries an explicit disclaimer for unverified apps. This should be dropped.

joklaps: I changed the description as described here: https://docs.flathub.org/docs/for-app-authors/requirements#appstream

If the package gets accepted, I see it is possible to point the "report an issue" url towards the flathub repository of the package.

barthalion: That doc is outdated. Please remove the disclaimer altogether.

joklaps: Done

gpxsee-appdata.patch
joklaps: fixed it

AsciiWolf:
bot, build org.gpxsee.GPXSee

PR #9362, opened 2026-07-16
Add com.dygma.bazecor

mean-ui-thread:

- [X] Please describe the application briefly. **This is the flatpak for Dygma Bazecor app, for configuring programmable Dygma keyboards such as the Dygma Defy and the Dygma Raise.**
- [X] Please attach a video showcasing the application on Linux using the Flatpak.  [**video link here**](https://github.com/user-attachments/assets/ef64683e-e4cb-4618-a024-bb7089e00d5b)
  - [X] The application has a meaningful development history, evidence of real-world use, and a clear commitment to ongoing maintenance, as required by the [development history requirements][history].
  - [X] I have disclosed any AI-generated material included in the application or its Flathub packaging, as required by the [Generative AI policy][ai]. **Affected parts and approximate extent:** None
  - [X] I have not used AI tools or agents to generate or automate this submission pull request or its review interactions.
- [X] I am a (new) developer/upstream contributor to the project. If not, I contacted upstream developers about this submission. **Link:** https://discord.com/channels/895297925578637362/1519724469705375857/1519729031942242424

- AlexDygma
- MiquelDygma
- gonza2206

[history]: https://docs.flathub.org/docs/for-app-authors/requirements#insufficient-development-history
[ai]: https://docs.flathub.org/docs/for-app-authors/requirements#generative-ai-policy


com.dygma.bazecor.yml
```diff
@@ -0,0 +1,118 @@
...
+tags:
+  - floss
+finish-args:
+  - --allow=per-app-dev-shm
+  - --device=all
+  - --env=ELECTRON_OZONE_PLATFORM_HINT=auto
+  - --env=XCURSOR_PATH=/run/host/user-share/icons:/run/host/share/icons
+  - --filesystem=/run/udev:ro
+  - --filesystem=~/Dygma
+  - --filesystem=host-etc:ro # This flag requires explicit permission to be granted by flathub.
```
mean-ui-thread, line 16: I do not know what the process is to formally request permission to use `host-etc`, so I am going to at leave explain it here in this comment.

Dygma UX design have a requirement to check:
1. if `/etc/udev/rules.d/60-dygma.rules` is present, and
3. if `/etc/udev/rules.d/60-dygma.rules` is up-to date.

If the file is either missing or if it is too old, this dialog prompt appears to the user at launch:

Image

in order for Bazecor to access `/run/host/etc/udev/rules.d/60-dygma.rules` to meet their UX requirements, we need `--filesystem=host-etc:ro`

And in case you are curious about what happens when the user clicks on the Install button, it will (obviously) fail, but will also provide the bash commands to create or update their rule files to the latest.

It is still under review in my PR, but the proposal currently looks like this

Image

But will most likely be changed a little bit more to be a bit more generic and look less like a catastrophic failure. I'm just waiting for Miquel to get back to me to let me know what their UX would like to see.

Thanks!

forge.config.ts.patch
```diff
@@ -0,0 +1,38 @@
+diff --git a/forge.config.ts b/forge.config.ts
+index 9a6a54db..4d1c6ff5 100644
+--- a/forge.config.ts
++++ b/forge.config.ts
+@@ -10,6 +10,7 @@ import rendererConfig from "./webpack.renderer.config";
+ import mainConfig from "./webpack.main.config";
+ 
+ const packagerConfig: ForgePackagerOptions = {
++  download: { unsafelyDisableChecksums: true },
```
mean-ui-thread, line 9: FYI, without this line, Electron-Forge (the builder tool that Dygma selected for building their electron app) tries to access the internet to retrieve the SHA256 of the electron version that was selected, probably to check if the version currently installed by `yarn install --offline` matches what it expects. Of course in a sandboxed build, that's not possible, so that's why I'm disabling the sha256 check that way via a patch.

forge.config.ts.patch
```diff
@@ -0,0 +1,38 @@
...
++
+       /**
+        * Serialport, usb and uiohook-napi are problematic libraries to run in Electron.
+        * When Electron app is been built, these libraries are not included properly in the final executable.
+@@ -101,7 +112,7 @@ const config: ForgeConfig = {
+       };
+ 
+       fs.writeFileSync(path.resolve(buildPath, "package.json"), JSON.stringify(packageJson));
+-      spawnSync("npm", ["install", "--omit=dev"], {
++      spawnSync("yarn", ["install", "--production", "--offline"], {
```
mean-ui-thread, line 35: Dygma made this workaround to call `npm install --omit=dev` inside of the build to package up 3 native modules (`uiohook-napi`, `usb`, and `serialport`) that electron-forge wasn't automatically installing.

https://github.com/Dygmalab/Bazecor/blob/development/forge.config.ts#L97-L101

I really don't know enough about electron and electron to know if that's the right or wrong way to proceed, but their work-around does not work in a sandboxed build. 

Since they are using `yarn` and have setup an offline-mirror for yarn to get this build-from-source going, I had to ensure that their work-around was using the same offline mirror as our build. That's why I'm copying `.yarnrc` and `yarn.lock` into this internal folder, and that did the trick!

Image

com.dygma.bazecor.yml
```diff
@@ -0,0 +1,118 @@
...
+modules:
+  # Bazecor's linux build (or WSL on Windows) requires libudev for compiling Bazecor.
+  # In their Windows build instructions: https://github.com/Dygmalab/Bazecor#development-on-windows,
+  # They instruct WSL users (same as Linux), to install the `libudev-dev`. Unfortunately, since the flatpak
+  # build system is required to be sandboxed, it is not possible to allow it consume the system installed
+  # libudev library and other development files. After searching across flathub repositories for any
+  # existing projects with a similar build requirements, I stumbled upon this flathub app, borrowed and
+  # adapted their solution:
+  # https://github.com/flathub/com.github.jkotra.eovpn/blob/f90222a30251c1988026f43d0fe598c5075eb520/com.github.jkotra.eovpn.yml#L54-L79
+  # If there is a better way to make libudev available to Bazecor, please let me know. Thanks!
```
Erick555: libudev is in flatpak runtimes, what's the problem then?

```
❯ flatpak run --command=ls org.freedesktop.Sdk//25.08 /usr/lib/x86_64-linux-gnu |grep libudev
libudev.so
libudev.so.1
libudev.so.1.7.10
```

com.dygma.bazecor.yml
Erick555: Why they don't scan directly for `/dev/<device>` access rather than for udev rule? Simple logic tells that if device is inaccessible then udev rule doesn't exist so it could show the same dialog for users.

Accessing path with hundreds of files to check for one file existence (not even using it) can be questionable.

com.dygma.bazecor.yml
```diff
@@ -0,0 +1,118 @@
...
+runtime: org.freedesktop.Platform
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node24
+command: Bazecor
+separate-locales: false
+tags:
+  - floss
+finish-args:
+  - --allow=per-app-dev-shm
```
Erick555: what's the rationale for this?

com.dygma.bazecor.yml
mean-ui-thread: Thank you @Erick555 for that feedback 

> Why they don't scan directly for /dev/<device> access rather than for udev rule?

I cannot speak for them and would have to ask them, but I believe the rationale behind that decision (and that's just my personal guess) is that they probably want to ensure that the udev rule file is up to date, and to have it automatically updated when they release a new product. For example, they are about to release a new product yet to be announced:

image

and once it is released, then they would like everyone's udev file to be updated with that new entry on the next release of Bazecor, so that if those users buy the new keyboard product, it should just work. That's what I think their rationale might be.

> Accessing path with hundreds of files to check for one file existence (not even using it) can be questionable.

I don't disagree with you at all. But it's not my call to make this big change in their code unfortunately. 

If you think that overwall, what Dygma wants isn't a good valid reason to be granted permission to access `host-etc:ro`, then Dygma will probably need to figure this out unless I can find some time to do it for them. I just volunteering for the flatpak part, I didn't really want to get involved doing any electron/typescript/nodejs stuff since I don't have any experience with any web tech (I'm just a C/C++ guy!)

Another alternative would be to disable that udev rule installation prompt dialog entirely using another patch file, but I'm pretty sure Dygma won't agree to that. They are already expressed concerns about the poor user experience from asking the user to copy/paste commands manually into a terminal.

If you are certain that `--filesystem=host-etc:ro` will be rejected, then I'll remove it right now, and then I guess I will try torturing myself with more typescript code... 

Thanks!

com.dygma.bazecor.yml
mean-ui-thread: Hi @Erick555 

Great catch!  I had literally no idea that it was already available in the latest org.freedesktop.Sdk//25.08. Even the `/usr/include/libudev.h` header is there.  Everything is already there. I removed eudev completely and it was compiling just fine this time.

That is so strange... I recall having... I think it was a missing header, something like that which prompted me to build and install what was missing to get my build up and running. Could it be because I originally did not specify any runtime version and the default was older than `25.08`? I'm not sure.

This discovery makes me very happy. The leaner my yaml is, the easier it will be for me to maintain down the line!  Thanks!

com.dygma.bazecor.yml
mean-ui-thread: Hi @Erick555 

That was probably accidentally left behind from when I was trying to stop my Electron app from crashing in Wayland (and it did not work). I believe the fix in the end was to replace `--socket=x11` by `--socket=fallback-x11`. 

I know for sure that Electron uses `/dev/shm` (shared memory IPC) to communicate between the main loop and the renderer, but from what I can read from flatpak's documents, this flag is to allow multiple instances of the same app to use the same shared memory instance, which we really don't need for bazecor. I can try removing it and see if the runtime is still stable in wayland.... just a sec...

mean-ui-thread: I am pleased to report that Bazecore still works without this flag on Wayland, and does not crash! Thank you @Erick555

com.dygma.bazecor.yml
mean-ui-thread: @Erick555 Just in case we won't be getting the permission to get the permission to use `host-etc:ro`, I put the upstream PR on hold until we know what the plan will be: 

https://github.com/Dygmalab/Bazecor/pull/1123/changes#r3600258849

com.dygma.bazecor.yml
Erick555: > Could it be because I originally did not specify any runtime version and the default was older than 25.08? I'm not sure.

libudev was added in runtime in 19.08 so you would need to install freedesktop-sdk 18.08 to not have it.

com.dygma.bazecor.yml
Erick555: Personally I think I wouldn't allow this permission for this usecase but I'm not deciding about that on flathub so I don't know what will happen.

com.dygma.bazecor.yml
```diff
@@ -0,0 +1,79 @@
...
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node24
+command: Bazecor
+separate-locales: false
+finish-args:
+  - --device=all
+  - --env=ELECTRON_OZONE_PLATFORM_HINT=auto
+  - --env=XCURSOR_PATH=/run/host/user-share/icons:/run/host/share/icons
+  - --filesystem=/run/udev:ro
+  - --filesystem=~/Dygma
```
Erick555: What's the purpose of this folder? Is it shared with something outside of flatpak?

In general flatpaks should have their private files in `~/.var/app/<app_id>`. In such case this could be replaced with `--persist=Dygma` that will put it into `~/.var/app/com.dygma.bazecor/Dygma`

com.dygma.bazecor.yml
```diff
@@ -0,0 +1,79 @@
...
+  - --filesystem=host-etc:ro # This flag requires explicit permission to be granted by flathub.
+  - --filesystem=xdg-documents
```
Erick555: it isn't obvious why keyboard config app(?) needs access to whole Documents folder. Electron should support file chooser portal so users should be able to open document of their choice without this permission.

com.dygma.bazecor.yml
mean-ui-thread: @Erick555 I completely understand. I will ask directly to other flathub maintainers on matrix.org and see what say. Thanks!

com.dygma.bazecor.yml
mean-ui-thread: Hi @Erick555, This is a bit of a long story, and I'm going to do my best to explain it the best I can. It is directly linked to your other question about filesystem access to `xdg-documents`

The purpose of this `$HOME/Dygma` folder is essentlally the default backup folder to store Dygma Keyboard Configuration for new Bazecor users For the past ~7 years, new users running Bazecor for the first time would get a new config created, and by default, the backup folder would be set to `$HOME/Dygma/Backup`.  It is the same for Windows and MacOS.  It is user-configurable and can be changed to a different folder in the app's settings, but the majority of people leave this backup folder as-is.

However, this created an issue for flatpak: unless the `~/Dygma` was already present before launch, it wouldn't be able to bind-mount it. If it is missing, the only way to allow Bazecor to create this folder  would be to allow read/write access to the users home folder:

```yaml
 - --filesystem=xdg-home
```

And I was told that getting`--filesystem=xdg-home` approved by flathub would be impossible. 

Unfortuately, we cannot penalize legacy users with pre-existing backups to lose access to their already-existing `$HOME/Dygma/Backups`. This is why we would like to grant access to `--filesystem=~/Dygma` so that legacy users who wants to switch from `.appimage` to `.flatpak` would retain access to their pre-existing keyboard configuration backups. However, I understand that users who have moved this backup folder to a different name (example: changed their settings manually from $HOME/Dygma/Backups to `$HOME/dygma-backups` would be penalize for switching to flatpak, but there is not much that we can do about that since `--filesystem=xdg-home` would mean a guaranteed app submission rejection from flathub.

Your suggestion to use `--persist=Dygma` is a very good suggestion, but that would also make legacy user lose access to their pre-existing `$HOME/Dygma` unfortunately. But also, because flatpak's persistant storage stores the folder in a nested hidden folder structure, it would be very difficult for other released bundle formats (appimage, future .deb, and future .rpm bundles) to access this backup folder.

After many cognitive back-and-forth, the compromize was to change the default folder from `$HOME/Dygma` to `$HOME/Documents/Dygma`. This is why we are adding the following access
```yaml
 - --filesystem=xdg-documents
```
And we are making the same change upstream. This way, Bazecor for new users will be able to create the new `$HOME/Documents/Dygma/Backups` folder.

This way, with `--filesystem=~/Dygma` and `--filesystem=xdg-documents` combined:

Pro:
1. For legacy users, flatpak app will still access their pre-existing `${HOME}/Dygma/Backups` folder
2. For new users, flatpak app will be able to create `${HOME}/Documents/Dygma/Backups`
3. AppImage app bundle can access both `${HOME}/Dygma/Backups` and `${HOME}/Documents/Dygma/Backups`
4. RPM app bundle can access both `${HOME}/Dygma/Backups` and `${HOME}/Documents/Dygma/Backups`
5. DEB app bundle can access both `${HOME}/Dygma/Backups` and `${HOME}/Documents/Dygma/Backups`

Con:
1. Legacy user who manually set their backup folder to another location other than `$HOME/Dygma` (and not under their ~/Documents folder) would lose access to their backup.

It is the best compromize that we were able to come up with. 

I hope that my thorough explanation makes sense.

Thanks!

image

com.dygma.bazecor.yml
mean-ui-thread: Hi @Erick555 . The explanation for it is part of the thorough explanation I wrote for `--filesystem=~/Dygma` 

https://github.com/flathub/flathub/pull/9362#discussion_r3603737819

It was easier to combine the two explanations in a single one because there is a valid reason why both mounts would be required: they are tied together.

Thanks!

mean-ui-thread: > Electron should support file chooser portal so users should be able to open document of their choice without this permission.

I'm unsure about what you mean by that. Could you please clarify? What's a "file chooser portal"?

com.dygma.bazecor.yml
mean-ui-thread: > libudev was added in runtime in 19.08 so you would need to install freedesktop-sdk 18.08 to not have it.

Then I really don't know how this happened. I definitely never tried installing freedesktop-sdk 18.08 manually, that's for sure. But I do remember struggling quite a lot with the built-in flatpak builder from `@electron-forge/maker-flatpak` that I originally tried, and remember strugging with an outdated zypak version and many `clang++ not found` errors. Maybe my incorrect presumption of missing libudev libary and missing headers came from these old failed experiments. No idea. This was at least couple of months ago. My memory isn't as good as it used to be.

But I'm so glad that this isn't needed anymore. Thank you for your help @Erick555 Much appreciated! :smile:

com.dygma.bazecor.yml
Erick555: > However, this created an issue for flatpak: unless the ~/Dygma was already present before launch, it wouldn't be able to bind-mount it

You can pre-create folder and bind-mount it after:
```
$ file Dygma
Dygma: cannot open `Dygma' (No such file or directory)
$ flatpak run --filesystem=~/Dygma:create --command=file org.freedesktop.Platform//25.08 Dygma
Dygma: directory
$ file Dygma
Dygma: directory
```
So for your use-case you need either `--filesystem=~/Dygma:create` or `--filesystem=xdg-documents/Dygma:create` or both. You shouldn't need `--filesystem=xdg-documents` or `--filesystem=home`

com.dygma.bazecor.yml
Erick555: File chooser portal allows for flatpak apps to open arbitrary file or folder on filesystem without having permission to access them. User is able to click _Open file_ or whatever the function is named in app and flatpak will show dialog window with all files on filesystem to choose like it wasn't sandboxed (but it is). This way apps don't need permission while keeping functionality to open files. 

Electron framework should support this. 

Note that this requires user interaction through opening file dialog window, it doesn't allow accessing file in automate way unless they were opened by user before.

https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.FileChooser.html

Erick555: So for example user would be able to load backup by opening file dialog window and choosing appropriate file from any directory without having any` --filesystem=~/Dygma` but app wouldn't be able to do that on user behalf. This has some caveats like the chosen file can't depend on more files (like multifile archive).

I don't know if that fits for Dygma workflow.

mean-ui-thread: For the dialog windows, Dygma Bazecor React app appears to calling built-in feature of Electron called `open-dialog` with an `openDirectory` property.

```typescript
    const options = {
      title: i18n.keyboardSettings.backupFolder.title,
      buttonLabel: i18n.keyboardSettings.backupFolder.windowButton,
      properties: ["openDirectory"],
    };

    const resp = await ipcRenderer.invoke("open-dialog", options);
```

https://www.electronjs.org/docs/latest/api/dialog


and from a quick google search, it looks like electron might already be using `xdg-desktop-portal` under the hood!

do I need to use a different permission flag? because when I remove both `--filesystem=~/Dygma` and `--filesystem=xdg-documents`, it breaks.

Erick555: In what sense it breaks? Do you open file dialog and it doesn't allow you to chose a backup file?

mean-ui-thread: > In what sense it breaks? Do you open file dialog and it doesn't allow you to chose a backup file?

The app crashes because the user settings still points to a backup folder that has become unavailable, but this is most likely a bug or oversight from the Bazecor implementation.

```
(node:2332731) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, scandir '/home/mean-ui-thread/Dygma/Backups/'
    at Object.readdirSync (node:fs:1506:26)
    at t.readdirSync (node:electron/js2c/node_init:2:13703)
    at u (/tmp/.mount_BazecoOeJGIH/usr/lib/bazecor/resources/app/.webpack/main/index.js:8:256746)
    at t.default (/tmp/.mount_BazecoOeJGIH/usr/lib/bazecor/resources/app/.webpack/main/index.js:8:257601)
    at /tmp/.mount_BazecoOeJGIH/usr/lib/bazecor/resources/app/.webpack/main/index.js:8:155099
    at Generator.next (<anonymous>)
    at /tmp/.mount_BazecoOeJGIH/usr/lib/bazecor/resources/app/.webpack/main/index.js:8:154607
    at new Promise (<anonymous>)
    at n (/tmp/.mount_BazecoOeJGIH/usr/lib/bazecor/resources/app/.webpack/main/index.js:8:154352)
    at App.<anonymous> (/tmp/.mount_BazecoOeJGIH/usr/lib/bazecor/resources/app/.webpack/main/index.js:8:155026)
```

~I think I found the flag I needed:~

```yaml
- --talk-name=org.freedesktop.portal.FileChooser
```

~This fixes everything, but now I am getting `finish-args-portal-talk-name`~

```json
{
    "errors": [
        "finish-args-portal-talk-name",
    ],
    "info": [
        "finish-args-portal-talk-name: finish-args has talk-name access to XDG Portal busnames"
    ],
    "message": "See https://docs.flathub.org/linter for details and exceptions"
}
```

~First, did I use the correct permission flag? If so, how easy is it request permission from flathub to use this "portal"? Personally, it feels equally as dangerous as using `--filesystem=xdg-home` but maybe that's just me?~

EDIT: works without it (see below)

com.dygma.bazecor.yml
mean-ui-thread: ~HI @Erick555 My mind is blown!!! I had no idea about `:create`! This solves everything!!! Thank you!~

It works fine without it now. I have no idea what's going on (see my comment about removing `xdg-documents` here https://github.com/flathub/flathub/pull/9362#discussion_r3609098955 ). It used to crash, now it doesn't. Can't reproduce it anymore.  I'm happy to have fewer of these `--filesystem=...` flags if I don't need them. Thank you @Erick555

com.dygma.bazecor.yml
mean-ui-thread: > In what sense it breaks? Do you open file dialog and it doesn't allow you to chose a backup file?

After trying it a 2nd time after deleting all user setting (to simulate a new user launching Bazecor for the first time ever), without any ~/ dygma or xdg-documents, filesystem flags, without any talk-name flags, it did not crash but the back folder is pointing to a very strange place, but the dialog shows the correct location.

image

Is this expected?

mean-ui-thread: Very confusing. 

Tried it again, and got better results after doing roughtly the same thing:

First, deleted flatpak user settings to start from a clean slate.
```bash
rm -rf ~/.var/app/com.dygma.bazecor
```
rebuilt from scratch, deleting all previous built artefact to make sure nothing can invludence anything else I already have
```bash
git clean -fdx
flatpak run --command=flathub-build org.flatpak.Builder --force-clean --install com.dygma.bazecor.yml
flatpak run com.dygma.bazecor
```
but this time, everything looks good
image

I think my computer is gaslighting me now. I can't even reproduce the missing folder runtime crash anymore.

com.dygma.bazecor.yml
mean-ui-thread: @Erick555  bad news: I should have done more testing than just assuming that because it built without any errors that it was good to go, but after removing eudev, the app can no longer find the keyboards:

image

I'm going to try to revert the removal to see if that's truly the `eudev` package that was responsible. libudev 1.7.10 is quite old, and eudev was giving us libudev 3.2.14

mean-ui-thread: @Erick555 yup. After adding `eudev` module back, Bazecor can now see my keyboard 
image

Something's missing with the built-in libudev. I'll do some more investigation, but I might have to add that module back if I can't figure this out.

Sorry about that.

mean-ui-thread: @Erick555 I think the error log speaks for itself. This is the error generated when using the built-in libudev

```
16:32:13.855 › Error while finding keyboards Error: spawn udevadm ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:285:19)
    at onErrorNT (node:internal/child_process:483:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
```
the `udevadm` binary is missing. It's not called directly by Bazecor, but either the Electron module calls it, or libudev itself calls it.

when using `eudev` (embedded udev), it still doesn't add any `udevadm` to the flatpak, but the scan works flawlessly with it:

```
16:26:26.446 › SerialPort devices find: [
  {
    path: '/dev/ttyACM0',
    manufacturer: 'DYGMA',
    serialNumber: '2ED8EB4B128838AF',
    pnpId: 'usb-DYGMA_DEFY_2ED8EB4B128838AF-if00',
    vendorId: '35ef',
    productId: '0012'
  },
  { path: '/dev/ttyS0' },
  { path: '/dev/ttyS1' },
   ...
   ...
   ...
   ...
]
```

I am going to add it back, but I'll refine the comment above it explaining why `eudev` is needed afterall.

Thanks!

mean-ui-thread: Module added back, comment updated to :

```yaml
  # The built-in libudev from org.freedesktop.Sdk is missing a necessary udevadm binary. Without it, it fails
  # to perform a system-wide scan to discover every Dygma devices that are currently connected via USB or
  # bluetooth. Until we find a way to perform system-wide device scans via the built-in libudev provided by
  # org.freedesktop.Sdk, we will keep using this fallback solution.
```

com.dygma.bazecor.yml
Erick555: `--talk-name=org.freedesktop.portal.FileChooser` is granted automatically and shouldn't be added in permissions.

> The app crashes because the user settings still points to a backup folder that has become unavailable, but this is most likely a bug or oversight from the Bazecor implementation.

Since the Dygma flatpak doesn't exist yet I think we may assume there won't be anything set in config unless user would copy it manually from non-flatpak installation.

com.dygma.bazecor.yml
Erick555: @mean-ui-thread


> libudev 1.7.10 is quite old, and eudev was giving us libudev 3.2.14

libudev shipped in runtime is part of systemd project. 1.7.10 is from systemd 257, originally  released 1.5 years ago. Latest version is 1.7.14. eudev project uses different versioning so they aren't comparable.

Maybe someone could debug this further to understand why systemd libudev doesn't work while eudev libudev does. It may be something fixable in the runtime.

mean-ui-thread: Sorry @Erick555. You were right and I was wrong. `eudev` is not the same thing as `libudev`. It's pretty much a rewrite from a `systemd` fork. It is API-compatible, and more like a replacement. Their verions cannot be compared. You are correct: `eudev` project uses different versioning so they aren't comparable.

From the logs, it looks like (and that's just my own hypothesis) that electron attempts under the hood to spawn an instance of `udevadm`, and fails because the binary is missing from the sdk. When using the `eudev` replacement isntead, it still does not provide any `udevadm` binaries, but electron does not try to spawn an instance of `udevadm` anymore. I haven't checked the implementation of electron, but my guts tell me that spawning `udevadm` might be a fallback mechanism, and that fallback mechanism seems no longer needed with `eudev`. But that's just my personal theory. I don't have enough brain power left to look deeper into it.

Perhaps you might know a way to get `udevadm` binary installed in the flatpak, instead of replacing the build-in libudev by eudev? Maybe that might be worth trying first?

com.dygma.bazecor.yml
mean-ui-thread: Hi @Erick555 

> --talk-name=org.freedesktop.portal.FileChooser is granted automatically and shouldn't be added in permissions.

That explains why it works out-of-the-box without any issues.

> Since the Dygma flatpak doesn't exist yet I think we may assume there won't be anything set in config unless user would copy it manually from non-flatpak installation.

That's also what I thought, and I tried closing the app, moving `~/Dygma` to `~Dygma2`, relaunched the app, tried to manually create a new backup, and the folder was recreated with the one new backup file in there. I simply cannot reproduce the crash anymore. Having no more crash is both a good thing, but also a scary thing because I cannot explain it.

Thank you again for all of your help!

com.dygma.bazecor.yml
mean-ui-thread: @Erick555 I was wrong again: I couldn't find it because it's not installed in `/usr/bin`, but in `/app/bin` 🤦 

```b
(.venv) %m%# which udevadm
/app/bin/udevadm
(.venv) %m%# 
```
that is why it is working. `eudev` essentially provides that missing `udevadm`

mean-ui-thread: Another thing I discovered where I was wrong again: Where I thought it was Electron API, it was just plain nodejs module called `SerialPort`. I found the `udevadm` call by grepping into the downloaded `node_modules`, and found it in `node_modules/@serialport/bindings-cpp/dist/linux-list.js`. 

Learning something every day.

Erick555: > that is why it is working. eudev essentially provides that missing udevadm


Ok, then building eudev in the app should be the proper way

mean-ui-thread:
Looks like the checklist has changed recently. Instead of re-opening a new PR, I have updated the old checklist with the new checklist. Let's try to re-open it. Thanks!

teamcons:
> /review

this doesnt do shit

teamcons:
bot, build

mean-ui-thread:
bot, build

mean-ui-thread:
bot, build

mean-ui-thread:
> 'runtime-update-available-to-org.freedesktop.Platform-26.08' warning found in linter manifest check. Details: Please consider updating to the latest runtime version

FYI, I was aware of this warning when testing my yaml file locally with the linter. I did try to upgrade from 25.08 to 26.08 but it generated two build errors: It couldn't find the Electron2 Base, and couldn't find node24. Perhaps it is too soon to upgrade and perhaps these extensions will become available in 26.04 at a later time?  I can always upgrade my SDK version at a later time, right?

Erick555:
yes, you can upgrade it later

teamcons:
I will be closing my eyes on two things, because this PR is cursed, and i want this to be finally over but:
-Usually we expect the runtime to be bumped before merging. Here this PR just is a little bit older, and we've done enough back and forth
-We prefer people to keep their personal build scripts locally, to avoid unnecessary files lingering around and confusing people. Though thats not dramatic

PR #9385, opened 2026-07-18
Add org.catacombing.charon

chrisduerr:

- [X] Please describe the application briefly. Charon is a mobile-optimized Wayland maps and navigation application.

https://github.com/user-attachments/assets/ae84ef1b-6f9b-4248-99f7-eb2f7aeae514

- [X] I am an _(please keep whichever is applicable and remove the rest)_ author to the project.


chrisduerr:
Should be ready for review now.

I originally omitted the video hoping I could let the builder build it for me, but since this was immediately closed I decided to just spend the 2 hours compiling it myself on weak arm hardware.

The reason the video is so choppy is because the hardware is terrible. I promise my software isn't that bad.

modrana_countries_provided.json
teamcons, line 1: What is this file?

barthalion:
bot, build

org.catacombing.charon.json
```diff
@@ -0,0 +1,139 @@
...
+        "org.freedesktop.Sdk.Extension.rust-stable",
+        "org.freedesktop.Sdk.Extension.llvm22"
+    ],
+    "command": "charon",
+    "finish-args": [
+        "--socket=wayland",
+        "--share=network",
+        "--device=dri",
+        "--system-talk-name=net.hadess.SensorProxy",
+        "--system-talk-name=org.freedesktop.ModemManager1"
```
teamcons, line 16: what are these for?

tile_metadata.json
teamcons, line 1: What is this file?

last_commit_position.h
teamcons, line 1: What is this for?

org.catacombing.charon.json
chrisduerr: For GPS location and heading.

last_commit_position.h
chrisduerr: This is a patch necessary to build `gn`. It tries to get details with `git` during build but fails when built from an archive, so instead you have to build it with `--no-last-commit-position` to omit that and manually provide the version info yourself. It has no impact on the output executable.

tile_metadata.json
chrisduerr: The tile metadata contains metadata about the offline tiles I serve on https://catacombing.org/tiles/.

chrisduerr:
> Error: 'finish-args-only-wayland' error found in linter manifest check. Details: finish-args has only wayland socket but no x11 or fallback-x11 socket

I've added an exception to the failing lint, which apparently I had to make for my other application too (org.catacombing.kumo): https://github.com/flathub-infra/flatpak-builder-lint/pull/1080

tile_metadata.json
teamcons: It should be upstream, not on flathub

org.catacombing.charon.json
```diff
@@ -0,0 +1,139 @@
...
+            "path": "./modrana_countries_provided.json"
+                },
+                {
+                    "type": "file",
+            "path": "./tile_metadata.json"
+                },
+                {
+                    "type": "git",
+                    "url": "https://git.sr.ht/~undeadleech/charon",
+                    "commit": "cb27c2c2d413a94e1da9b046736da8d539ce581f"
```
teamcons, line 134: a tagged releasing alongside the commit maybe?

last_commit_position.h
teamcons: Ah yes. I oversaw it in the manifest. Awkward. Thank you.

org.catacombing.charon.json
chrisduerr: My plan is to make a release once I know the flatpak will actually build properly for aarch64 and x86, just to avoid having to make multiple releases in case there's some other issues.

Personally I would have just done so once it's on flathub/org.catacombing.charon, but I can also tag it once I've got confirmation from the builder that everything is ready to go.

modrana_countries_provided.json
chrisduerr: This is the metadata of geocoding and routing data served by modrana.org.

tile_metadata.json
chrisduerr: Upstream doesn't need it though, because upstream can build online.

This is in flathub because it's required as an asset at build time, but cannot be downloaded during the build (since it's offline).

The reason why I'm not just pulling it from the server directly is just because the server doesn't keep archives, thus making it non-reproducible if I wanted to build a previous version. This way I can build offline without automatically always pulling the latest version from the server.

It's essentially the exact same as a build dependency I statically compile into my project, thus it's not appropriate to include it upstream directly.

modrana_countries_provided.json
CodedOre: Why is this file part of the submission?

Why is it not provided with the application from upstream?

CodedOre:
There are still quite a few desktops which run on X11, either exclusively or by default. 
Also, it seems to me that the application is intended for phones only?

The latter as well, but especially the first makes this inpermissble under the "Environment-locked application" policy:
https://docs.flathub.org/docs/for-app-authors/requirements#environment-locked-applications

chrisduerr:
> There are still quite a few desktops which run on X11, either exclusively or by default.
> Also, it seems to me that the application is intended for phones only?

The same applies to my Kumo application, which got accepted in the past (https://flathub.org/en/apps/org.catacombing.kumo).

Both of these applications work just fine on my desktop and I test them on the desktop regularly. I just don't use them on my desktop. Sure applications that focus on mobile Linux first are niche, but it wouldn't exactly help with fixing that to prevent their packaging entirely.

As far as Wayland exclusivity is concerned: Pretty much every major system out there defaults to Wayland at this point. We're at a point where even I have switched from i3 to Sway after many years of having remaining issues. Supporting mainly phones might make it a bit niche, but supporting only Wayland on flathub shouldn't make a significant difference.

> The latter as well, but especially the first makes this inpermissble under the "Environment-locked application" policy:
https://docs.flathub.org/docs/for-app-authors/requirements#environment-locked-applications

This text explicitly points out applications that function under highly specific environments, using ***a particular desktop environment or distribution*** as an example. I've tested this application on x86 Arch (Sway), aarch64 Arch (Catacomb), and aarch64 postmarketOS (Phosh). It **functions** just fine under all of these environments. While this also clearly states that exceptions are possible, I don't think this applies in the first place.

modrana_countries_provided.json
chrisduerr: Because it's usually pulled online during the build process.

chrisduerr: Because it's usually pulled online during the build process.

CodedOre:
@chrisduerr, I run a short test of the application, and I would say there is some point to it running into issues with the policy.

The "particular desktop environment or distribution" part is an example for restricted environments. Another would be people using a desktop or mobile device.

As of now, the application has no obvious way to control the window (can't close, move or resize it) and controls seem to expect a touchscreen to be available. With purely mouse and keyboard stuff like zooming is unavailable.

This would be a prime example of the application being restricted to a mobile device environment.

I would now consider missing X11 to be a minor point as it's true that Wayland support is relatively widespread. But I still find it not ideal to have an application that will perform very not well on normal desktops, considering that is still what most Flathub user would use.

teamcons:
I got a review request too but codedore is the one calling the shots. The app adapting somewhat itself depending on mobile or desktop would be nice

chrisduerr:
> As of now, the application has no obvious way to control the window (can't close, move or resize it)

That's completely unrelated to mobile though, CSDs are just not a good idea. This works just fine on Sway or any other desktop environments with keyboard shortcuts/SSDs.

> With purely mouse and keyboard stuff like zooming is unavailable.

Mouse wheel zooming is part of the upcoming release, though you can still zoom with a mouse even without mouse wheel zooming. I've been testing it that way on desktop for months.

> I still find it not ideal to have an application that will perform very not well on normal desktops

This seems like a very biased perception of a "normal desktop". It's essentially just gnome.

CodedOre:
> That's completely unrelated to mobile though, CSDs are just not a good idea. This works just fine on Sway or any other desktop environments with keyboard shortcuts/SSDs.
>
> […]
>
> This seems like a very biased perception of a "normal desktop". It's essentially just gnome.

Granted, I am using GNOME. And sure, some other Wayland compositors to provide SSD. 

But GNOME still is an commonly used desktop. And SSD is to my knowledge an optional protocol, not a Wayland standard. Whether its a good idea or not is not important here. Importan is that the application will look broken on quite a few systems.

